PR #2, opened 2016-06-10
Beta - dummy PR

tobiasKaminsky:


AndyScherzinger:
Should we then unpublish the beta from f-droid?

PR #9, opened 2016-06-12
Support animated GiF-Images

AndyScherzinger:
Add functionality to show animated gif files in slideshow (image preview)

rebased original PR done by @tobiasKaminsky so all kudos belongs to him.
Please review @tobiasKaminsky @przybylski

### DIFF
--- a/src/com/owncloud/android/ui/preview/ImageViewCustom.java
+++ b/src/com/owncloud/android/ui/preview/ImageViewCustom.java
@@ -4,12 +4,17 @@ import android.annotation.SuppressLint;
 import android.content.Context;
 import android.graphics.Bitmap;
 import android.graphics.Canvas;
+import android.graphics.Movie;
 import android.os.Build;
 import android.util.AttributeSet;
 import android.view.View;
 import android.widget.ImageView;
 
+import com.owncloud.android.datamodel.OCFile;
 import com.owncloud.android.lib.common.utils.Log_OC;
+
+import java.io.FileInputStream;
+import java.io.InputStream;
 
 public class ImageViewCustom extends ImageView {
 
@@ -23,7 +28,12 @@ public class ImageViewCustom extends ImageView {
     private int mBitmapHeight;
     private int mBitmapWidth;
 
-    
+    private Movie gifMovie;
+    private int movieWidth, movieHeight;
+    private long movieDuration;
+    private long movieRunDuration;
+    private long lastTick;
+
     public ImageViewCustom(Context context) {
         super(context);
     }
@@ -39,18 +49,54 @@ public class ImageViewCustom extends ImageView {
     @SuppressLint("NewApi")
 	@Override
     protected void onDraw(Canvas canvas) {
-
         if(IS_ICS_OR_HIGHER && checkIfMaximumBitmapExceed(canvas) || IS_VERSION_BUGGY_ON_RECYCLES ) {
             // Software type is set with two targets:
             // 1. prevent that bitmaps larger than maximum textures allowed are shown as black views in devices
             //  with LAYER_TYPE_HARDWARE enabled by default;
-            // 2. grant that bitmaps are correctly dellocated from memory in versions suffering the bug fixed in
+            // 2. grant that bitmaps are correctly de-allocated from memory in versions suffering the bug fixed in
             //  https://android.googlesource.com/platform/frameworks/base/+/034de6b1ec561797a2422314e6ef03e3cd3e08e0;
             //
             setLayerType(View.LAYER_TYPE_SOFTWARE, null);
         }
 
-        super.onDraw(canvas);
+        if(gifMovie == null){
+            super.onDraw(canvas);
+        } else {
+            long nowTick = android.os.SystemClock.uptimeMillis();
+            if (lastTick == 0) {
+                movieRunDuration = 0;
+            } else {
+                movieRunDuration += nowTick -lastTick;
+                if(movieRunDuration > movieDuration){
+                        movieRunDuration = 0;
+                }
+            }
+
+            gifMovie.setTime((int) movieRunDuration);
+
+            float scale;
+            if(gifMovie.height() > getHeight() || gifMovie.width() > getWidth()) {
+                scale = (1f / Math.min(canvas.getHeight() / gifMovie.height(),
+                        canvas.getWidth() / gifMovie.width())) + 0.25f;
+            } else {
+                scale = Math.min(canvas.getHeight() / gifMovie.height(),
+                                 canvas.getWidth() / gifMovie.width());
+            }
+
+            canvas.scale(scale, scale);
+            canvas.translate(((float) getWidth() / scale - (float) gifMovie.width()) / 2f,
+                    ((float) getHeight() / scale - (float) gifMovie.height()) /2f);
+
+            gifMovie.draw(canvas, 0, 0);
+
+            lastTick = nowTick;
+            invalidate();
+        }
+    }
+
+    @Override
+    protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
+        setMeasuredDimension(movieWidth, movieHeight);
     }
 
     /**
@@ -60,13 +106,9 @@ public class ImageViewCustom extends ImageView {
      */
     @SuppressLint("NewApi")
 	private boolean checkIfMaximumBitmapExceed(Canvas canvas) {
-        Log_OC.v(TAG, "Canvas maximum: " + canvas.getMaximumBitmapWidth() + " - " + canvas.getMaximumBitmapHeight());
-        if (mBitmapWidth > canvas.getMaximumBitmapWidth()
-                || mBitmapHeight > canvas.getMaximumBitmapHeight()) {
-            return true;
-        }
-        
-        return false;
+        return mBitmapWidth > canvas.getMaximumBitmapWidth()
+                || mBitmapHeight > canvas.getMaximumBitmapHeight();
+
     }
     
     @Override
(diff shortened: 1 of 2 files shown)


src/com/owncloud/android/ui/preview/ImageViewCustom.java
```diff
@@ -39,18 +48,58 @@ public ImageViewCustom(Context context, AttributeSet attrs, int defStyle) {
...
-            // 2. grant that bitmaps are correctly dellocated from memory in versions suffering the bug fixed in
+            // 2. grant that bitmaps are correctly de-allocated from memory in versions suffering the bug fixed in
             //  https://android.googlesource.com/platform/frameworks/base/+/034de6b1ec561797a2422314e6ef03e3cd3e08e0;
             //
             setLayerType(View.LAYER_TYPE_SOFTWARE, null);
         }
 
-        super.onDraw(canvas);
+        if(mGifMovie == null){
+            super.onDraw(canvas);
```
przybylski: Is it possible that not calling onDraw in this case would lead to some instability on different vendors? ImageView and View might do some low level calls to hardware to invoke redrawing.

src/com/owncloud/android/ui/preview/ImageViewCustom.java
```diff
@@ -39,18 +48,58 @@ public ImageViewCustom(Context context, AttributeSet attrs, int defStyle) {
...
+                        mMovieRunDuration = 0;
+                }
+            }
+
+            mGifMovie.setTime((int) mMovieRunDuration);
+
+            float scale;
+            if(mGifMovie.height() > getHeight() || mGifMovie.width() > getWidth()) {
+                scale = (1f / Math.min(canvas.getHeight() / mGifMovie.height(),
+                        canvas.getWidth() / mGifMovie.width())) + 0.25f;
```
przybylski: Why 0.25? Can this magic number have a name?

src/com/owncloud/android/ui/preview/ImageViewCustom.java
```diff
@@ -39,18 +48,58 @@ public ImageViewCustom(Context context, AttributeSet attrs, int defStyle) {
...
+                mMovieRunDuration += nowTick - mLastTick;
+                if(mMovieRunDuration > mMovieDuration){
+                        mMovieRunDuration = 0;
+                }
+            }
+
+            mGifMovie.setTime((int) mMovieRunDuration);
+
+            float scale;
+            if(mGifMovie.height() > getHeight() || mGifMovie.width() > getWidth()) {
```
przybylski: It would be nice to have this if/else extracted to some function like `getScaleToViewFactor(Movie)` or something like this.

src/com/owncloud/android/ui/preview/ImageViewCustom.java
```diff
@@ -60,24 +109,35 @@ protected void onDraw(Canvas canvas) {
...
+          InputStream gifInputStream = new FileInputStream(file.getStoragePath());
+          setLayerType(View.LAYER_TYPE_SOFTWARE, null);
+          setFocusable(true);
+
+          mGifMovie = Movie.decodeStream(gifInputStream);
+          mMovieWidth = mGifMovie.width();
+          mMovieHeight = mGifMovie.height();
+          mMovieDuration = mGifMovie.duration();
+        } catch (Exception e) {
+            e.printStackTrace();
```
przybylski: Maybe OC_Log here ? :)

src/com/owncloud/android/ui/preview/ImageViewCustom.java
```diff
@@ -60,24 +109,35 @@ protected void onDraw(Canvas canvas) {
...
      * but without keeping another reference to the {@link Bitmap}
      */
-    public void setImageBitmap (Bitmap bm) {
+    public void setImageBitmap(Bitmap bm) {
         mBitmapWidth = bm.getWidth();
         mBitmapHeight = bm.getHeight();
         super.setImageBitmap(bm);
     }
 
+    public void setGifImage(OCFile file){
```
przybylski: Is is necessary for this function to be public?

przybylski: Also GIF since it's abbreviation

src/com/owncloud/android/ui/preview/PreviewImageFragment.java
```diff
@@ -504,7 +504,12 @@ private void showLoadedImage(LoadImage result) {
                     imageView.setBackground(backrepeat);
                 }
 
-                imageView.setImageBitmap(bitmap);
+                if (result.ocFile.getMimetype().equalsIgnoreCase("image/gif")){
```
przybylski: missing space between ){

src/com/owncloud/android/ui/preview/ImageViewCustom.java
```diff
@@ -39,18 +48,58 @@ public ImageViewCustom(Context context, AttributeSet attrs, int defStyle) {
...
+            mGifMovie.draw(canvas, 0, 0);
+
+            mLastTick = nowTick;
+            invalidate();
+        }
+    }
+
+    @Override
+    protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
+        if (mGifMovie == null){
```
przybylski: missing space between ){

src/com/owncloud/android/ui/preview/ImageViewCustom.java
tobiasKaminsky: I have no idea :/

src/com/owncloud/android/ui/preview/ImageViewCustom.java
tobiasKaminsky: This is just a little offset that ensure that nothing gets cropped.

src/com/owncloud/android/ui/preview/ImageViewCustom.java
tobiasKaminsky: Yes, as it is used in PreviewImageFragment

tobiasKaminsky:
@przybylski thanks for the code review. 
I have commented/implemented your ideas

src/com/owncloud/android/ui/preview/ImageViewCustom.java
przybylski: I would rather keep calls to super unless you are absolutely sure that it is safe to avoid those.

src/com/owncloud/android/ui/preview/ImageViewCustom.java
```diff
@@ -39,18 +49,66 @@ public ImageViewCustom(Context context, AttributeSet attrs, int defStyle) {
...
+        if (movie.height() > getHeight() || movie.width() > getWidth()) {
+            float offset = 0.25f;
+            scale = (1f / Math.min(canvas.getHeight() / movie.height(),
+                    canvas.getWidth() / movie.width())) + offset;
+        } else {
+            scale = Math.min(canvas.getHeight() / movie.height(),
+                    canvas.getWidth() / movie.width());
+        }
+
+        return scale;
```
przybylski: Don't be afraid of early return ;)

src/com/owncloud/android/ui/preview/ImageViewCustom.java
```diff
@@ -60,24 +111,35 @@ protected void onDraw(Canvas canvas) {
...
      */
-    public void setImageBitmap (Bitmap bm) {
+    public void setImageBitmap(Bitmap bm) {
         mBitmapWidth = bm.getWidth();
         mBitmapHeight = bm.getHeight();
         super.setImageBitmap(bm);
     }
 
+    public void setGIFImage(OCFile file) {
+      try {
```
przybylski: There is some inconsistency here, setGIFImage takes OCFile and setImageBitmap takes bitmap. Latter is fine but former not quite. Altho I am not sure what would be the best way to resolve it.
Maybe `setGIFImageFromPath(String)` ?

AndyScherzinger: I'll change it to ```setGIFImageFromStoragePath(String storagePath)``` if that is fine with you

przybylski: Sounds good to me

przybylski:
LGTM +1

src/com/owncloud/android/ui/preview/ImageViewCustom.java
AndyScherzinger: @przybylski @tobiasKaminsky DONE, see https://github.com/nextcloud/android/pull/9/commits/d00cf73feee805f7ebc2c316a3490a9e4e7e14d3 I also reformatted the (new) code and added some javaDoc

AndyScherzinger:
I'll test it now on my device...

przybylski:
LGTM

tobiasKaminsky:
After a final test by @AndyScherzinger it LGTM

AndyScherzinger:
LGTM

AndyScherzinger:
@tobiasKaminsky I found one small issue which might not be related to the animated gif itself. When I view animated gifs that are way smaller than my display resolution I can see the image_fail.png in the background behind the gif itself. Any idea as to why that happens? try for example the following gif.
giphy2

przybylski:
This is probably related with calling super.onDraw

przybylski:
What I mean by that is because `ImageViewCustom` extends `ImageView` so it requires image to be set if image is suppose to be displayed. If image is unavailable then `ImageView` sees that as an error, so the natural thing in onDraw is to draw image_fail resource no canvas

AndyScherzinger:
Hah! You are right! So leave it out, I guess...

przybylski:
@AndyScherzinger you probably don't see a problem here but try to open any non gif image

przybylski:
Haha :D it made this pull request as looks good to me even without telling it :D

AndyScherzinger:
Ouch.. Can see anything anymore...

przybylski:
The ideal implementation would so something like:
* extend View not ImageView
* mimic few additional methods to support transformations from extending classes
* don't scale, inheriting classess should handle them (or allow scaling as an option)
* draw only bitmap or gif frame

but the current code is ok, just call `super.onDraw` when `mGifMovie` is null

AndyScherzinger:
Thanks @przybylski added what you described https://github.com/nextcloud/android/pull/9/commits/abe2b24f48cbff3389e0148daf67fd62211f414c
> but the current code is good enough, just call super.onDraw when mGifMovie is null

anyone fancying opening an issue (maybe plus PR) to work on the proposed solution by @przybylski ?

AndyScherzinger:
@tobiasKaminsky It's yours to hit the merge button :)

tobiasKaminsky:
:tada: my first merge, thanks for this great feeling :+1:

AndyScherzinger:
Well very much deserved it! :+1:

PR #1, opened 2016-06-06
first pass of changing to Nextcloud

jancborchardt:
Please review @nextcloud/android @LukasReschke :)

(The Travis stuff and screenshots we can change later.)

PR #35, opened 2016-06-19
Feature uploadsubfolders

Spacefish:
My InstantUpload folder is pretty crowded with thousands of pictures.. so i programmed a feature, which upload the pictures into subfolders based on year / month.
My folder got too so big, such that opening it takes quite a lot of time..


res/xml/preferences.xml
```diff
@@ -25,9 +25,13 @@
 		<com.owncloud.android.ui.CheckBoxPreferenceWithLongTitle android:key="instant_uploading"
 	                        android:title="@string/prefs_instant_upload"
 	                        android:summary="@string/prefs_instant_upload_summary"/>
-         <com.owncloud.android.ui.PreferenceWithLongSummary
+        <com.owncloud.android.ui.PreferenceWithLongSummary
 							android:title="@string/prefs_instant_upload_path_title"
 							android:key="instant_upload_path" />
+		<com.owncloud.android.ui.CheckBoxPreferenceWithLongTitle
+							android:title="@string/prefs_instant_upload_path_use_subfolders_title"
```
tobiasKaminsky: Please only lines with 80 chars, this makes review on github a lot easier.

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
```diff
@@ -83,6 +83,7 @@ private void handleNewPictureAction(Context context, Intent intent) {
         String file_path = null;
         String file_name = null;
         String mime_type = null;
+        String date_taken = null;
```
tobiasKaminsky: Why is this a string? It seems to be used as unix_timestamp so it should be long?

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
```diff
@@ -116,6 +117,8 @@ private void handleNewPictureAction(Context context, Intent intent) {
         file_path = c.getString(c.getColumnIndex(Images.Media.DATA));
         file_name = c.getString(c.getColumnIndex(Images.Media.DISPLAY_NAME));
         mime_type = c.getString(c.getColumnIndex(Images.Media.MIME_TYPE));
+        // date_taken = c.getString(c.getColumnIndex(Images.Media.DATE_TAKEN)); does not seem to work for all cameras
```
tobiasKaminsky: remove comment. The time the broadcast is received should be precious enough, although on the end of a month it could maybe load to a wrong date, very unlikely.

src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -117,11 +120,30 @@ public static final String getLogPath()  {
...
-    public static String getInstantUploadFilePath(Context context, String fileName) {
+    /**
+     * Returns the InstantUploadFilePath on the owncloud instance
+     *
+     * @param context
+     * @param fileName
+     * @param dateTaken: Time in milliseconds since 1970 when the picture was taken.
+     * @return
+     */
+    public static String getInstantUploadFilePath(Context context, String fileName, String dateTaken) {
```
tobiasKaminsky: :+1: javadoc :+1:

src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -117,11 +120,30 @@ public static final String getLogPath()  {
...
         SharedPreferences pref = PreferenceManager.getDefaultSharedPreferences(context);
         String uploadPathdef = context.getString(R.string.instant_upload_path);
         String uploadPath = pref.getString("instant_upload_path", uploadPathdef);
-        String value = uploadPath + OCFile.PATH_SEPARATOR +  (fileName == null ? "" : fileName);
+        String subFolders = "";
+        if(dateTaken != null && com.owncloud.android.db.PreferenceManager.instantPictureUploadPathUseSubfolders(context)) {
```
tobiasKaminsky: what if dateTaken is null? Then the file would never uploaded.
This should either result in an error that leads to a notification for the user or the file should be uploaded to the "plain" instant path.

src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -117,11 +120,30 @@ public static final String getLogPath()  {
...
+            try {
+                SimpleDateFormat formatter = new SimpleDateFormat("yyyy" + OCFile.PATH_SEPARATOR + "MM" + OCFile.PATH_SEPARATOR, Locale.ENGLISH);
+                subFolders = formatter.format(new Date(Long.parseLong(dateTaken)));
+                Date d = new Date(Long.parseLong((dateTaken)));
+            }
+            catch(RuntimeException ex) {
+                // don´t use a subfolder if we can´t parse the date
```
tobiasKaminsky: Same here: use "plain" upload path.

tobiasKaminsky:
Thanks for this feature!
Can you implement this also for video? Maybe it is not your use case, but it is inconsistent to have it only for images. But I think it should be two independent settings.

src/com/owncloud/android/utils/FileStorageUtils.java
Spacefish: oh, that´s  a typical, had an idea, made it work, changed it later and introduced a bug ^^ will fix

res/xml/preferences.xml
AndyScherzinger: @tobiasKaminsky 120 is fine with me ;) Since this is still enough for the files changed tab :D So I would vote to go for 120 line length. @przybylski

Spacefish:
implemented your feedback. It now works for videos as well.
I am a littlebit unsure if "use subfolders" is the correct name for the prefence setting.. Maybe we should use "Organize uploads into folders based on date" or something like this?
I am not an native english speaker, so i am unsure if this sounds correct.

AndyScherzinger:
pinging @jancborchardt for input regarding the mechanism and UX. I am unsure if and how this should work. I do like the idea of having more than simply one hard coded folder to where all my photos get uploaded but I am also not sure how to solve this since would use hierarchies as in folders for the year with subfolders in them. Since it is an optional setting I don't care too much about arguing about if and how to do this. Besides that I usually group my photos based on events like a vacation within the corresponding year's folder like "2016/05-Rome" which is not something we can implement since this is something for Google Photos server farm...

AndyScherzinger:
@jancborchardt In general my question here is probably how to handle and deal with additional app settings being introduced. #notAnExpert

tobiasKaminsky:
@AndyScherzinger this has been discussed on owncloud, ref: https://github.com/owncloud/android/issues/1421

I think it is a good approach. If we see that another mechanism for subfolder is needed we can extend this setting.
@jancborchardt ok for you?

AndyScherzinger:
Perfectly fine with me :)

res/xml/preferences.xml
przybylski: Longer line would be great I would vote to stick to how android code is written, which is 100 lines.
https://source.android.com/source/code-style.html#limit-line-length

res/values/strings.xml
```diff
@@ -307,6 +307,8 @@
 
     <string name="error__upload__local_file_not_copied">%1$s could not be copied to %2$s local folder</string>
     <string name="prefs_instant_upload_path_title">Upload path</string>
+    <string name="prefs_instant_upload_path_use_subfolders_title">Use subfolders</string>
+    <string name="prefs_instant_upload_path_use_subfolders_summary">Store in subfolders based on year and month</string>
```
przybylski: year & month: plural or singular ?

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
```diff
@@ -116,6 +117,7 @@ private void handleNewPictureAction(Context context, Intent intent) {
         file_path = c.getString(c.getColumnIndex(Images.Media.DATA));
         file_name = c.getString(c.getColumnIndex(Images.Media.DISPLAY_NAME));
         mime_type = c.getString(c.getColumnIndex(Images.Media.MIME_TYPE));
+        date_taken = System.currentTimeMillis();
```
przybylski: why do you separate declaration and initialization?

src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -117,11 +120,45 @@ public static final String getLogPath()  {
...
+    /**
+     * Returns the a string like 2016/08/ for the passed date. If date is 0 an empty
+     * string is returned
+     *
+     * @param date: date in microseconds since 1st January 1970
+     * @return
+     */
+    private static String getSubpathFromDate(long date) {
+        try {
+            if(date == 0) {
```
przybylski: this check should be outside try/catch to avoid catch block initialization on corner case

src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -117,11 +120,45 @@ public static final String getLogPath()  {
...
+                return "";
+            }
+            SimpleDateFormat formatter = new SimpleDateFormat(
+                    "yyyy" + OCFile.PATH_SEPARATOR + "MM" + OCFile.PATH_SEPARATOR, Locale.ENGLISH);
+            return formatter.format(new Date(date));
+        }
+        catch(RuntimeException ex) {
+            return "";
```
przybylski: Please log what went wrong

src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -117,11 +120,45 @@ public static final String getLogPath()  {
...
+     * @param dateTaken: Time in milliseconds since 1970 when the picture was taken.
+     * @return
+     */
+    public static String getInstantUploadFilePath(Context context, String fileName, long dateTaken) {
         SharedPreferences pref = PreferenceManager.getDefaultSharedPreferences(context);
         String uploadPathdef = context.getString(R.string.instant_upload_path);
         String uploadPath = pref.getString("instant_upload_path", uploadPathdef);
-        String value = uploadPath + OCFile.PATH_SEPARATOR +  (fileName == null ? "" : fileName);
+        String subPath = "";
+        if(com.owncloud.android.db.PreferenceManager.instantPictureUploadPathUseSubfolders(context)) {
```
przybylski: space after `if` keyword

src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -131,11 +168,16 @@ public static String getInstantUploadFilePath(Context context, String fileName)
...
+    public static String getInstantVideoUploadFilePath(Context context, String fileName, long dateTaken) {
         SharedPreferences pref = PreferenceManager.getDefaultSharedPreferences(context);
         String uploadVideoPathdef = context.getString(R.string.instant_upload_path);
         String uploadVideoPath = pref.getString("instant_video_upload_path", uploadVideoPathdef);
-        String value = uploadVideoPath + OCFile.PATH_SEPARATOR +  (fileName == null ? "" : fileName);
+        String subPath = "";
+        if(com.owncloud.android.db.PreferenceManager.instantVideoUploadPathUseSubfolders(context)) {
+            subPath = getSubpathFromDate(dateTaken);
+        }
+        String value = uploadVideoPath + OCFile.PATH_SEPARATOR + subPath
```
przybylski: do early return please

src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -131,11 +168,16 @@ public static String getInstantUploadFilePath(Context context, String fileName)
...
      * @return String: video file path composed
      */
-    public static String getInstantVideoUploadFilePath(Context context, String fileName) {
+    public static String getInstantVideoUploadFilePath(Context context, String fileName, long dateTaken) {
         SharedPreferences pref = PreferenceManager.getDefaultSharedPreferences(context);
         String uploadVideoPathdef = context.getString(R.string.instant_upload_path);
         String uploadVideoPath = pref.getString("instant_video_upload_path", uploadVideoPathdef);
-        String value = uploadVideoPath + OCFile.PATH_SEPARATOR +  (fileName == null ? "" : fileName);
+        String subPath = "";
+        if(com.owncloud.android.db.PreferenceManager.instantVideoUploadPathUseSubfolders(context)) {
```
przybylski: NOTE: @AndyScherzinger line like this is a example why PreferenceManager name for helpers is wrong ;)

AndyScherzinger: Hmmm, Yeah then we should probably rename unfortunately there is no way to enforce the usage of our helper class.
@Spacefish you should use com.owncloud.android.db.PreferenceManager all the time and add missing methods :)

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
Spacefish: cause it´s done like that in the code above it, and i don´t want to break the codeflow

src/com/owncloud/android/utils/FileStorageUtils.java
Spacefish: i don´t get it.. Please explain.

res/values/strings.xml
Spacefish: both, as year and month refers to a single subfolder, but multiple subfolders are created.

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
przybylski: IMO declarasion should be also initialization. I would even fix wrong other stuff. @AndyScherzinger @tobiasKaminsky you agree?

res/xml/preferences.xml
AndyScherzinger: @tobiasKaminsky what do you think? I am with @przybylski on this, so I also vote for 100 :100:

AndyScherzinger:
@Spacefish just FYI I did rebase your branch, so it's best if you pull again before you do any changes. Keeping you on par with master :)

Spacefish:
@AndyScherzinger  probably that´s why "git rebase master" did turn up as "already up to date" 💃

Spacefish:
any chance getting this merged?

src/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
AndyScherzinger: I would leave it like that until we are done with most of the open PRs and then open refactoring tasks

AndyScherzinger:
@Spacefish absolutely yes just waiting for the approval of @przybylski and @tobiasKaminski and a final test of the branch. Did you test what happens when you activate this for photos and videos with the same folders and naming?

AndyScherzinger:
And probably also some feedback from @jancborchardt

Spacefish:
Yes, the subfolders get created and the uploads are put into them.. I always used the same folder for videos and pictures.

AndyScherzinger:
Pinging @LukasReschke any idea how to proceed? In case Tobias and Bartosz give a thumbs up and my test is successful I would merge so this feature would be part of the next release. Any opinion on this? Else Jan blocks the merge...

res/xml/preferences.xml
tobiasKaminsky: 100 is also fine for me. I do not know who said 80 back on owncloud, but if 100 is still visible in changes tab on github everything is fine.

tobiasKaminsky:
This is a general question how to proceed.
I would like to have every PR first in beta and then after ~2 weeks it should get merged in master and therefore is ready for the next release.

AndyScherzinger:
Hmm, I would rather go for test and merge on master and put that into the Google Play beta and after two weeks release to stable. This is totally aside from the "real" beta you publish on f-droid.

res/xml/preferences.xml
przybylski: I think I introduced 80 chars limit. I don't recall why :D

AndyScherzinger: I think Github would also work with 120, not sure though...

tobiasKaminsky:
@przybylski is the code now fine for you?

src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -117,12 +123,46 @@ public static final String getLogPath()  {
...
+     * @param date: date in microseconds since 1st January 1970
+     * @return
+     */
+    private static String getSubpathFromDate(long date) {
+        if (date == 0) {
+            return "";
+        }
+        try {
+            SimpleDateFormat formatter = new SimpleDateFormat(
+                    "yyyy" + OCFile.PATH_SEPARATOR + "MM" + OCFile.PATH_SEPARATOR, Locale.ENGLISH);
```
przybylski: IMO it should be Locale.getDefault()

src/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -117,12 +123,46 @@ public static final String getLogPath()  {
...
 
-    public static String getInstantUploadFilePath(Context context, String fileName) {
+    /**
+     * Returns the a string like 2016/08/ for the passed date. If date is 0 an empty
+     * string is returned
+     *
+     * @param date: date in microseconds since 1st January 1970
+     * @return
+     */
+    private static String getSubpathFromDate(long date) {
```
przybylski: I think that passing `date` as a parameter isn't needed. You can get current time inside of this function. Pushing it here all the way doesn't makes too much of a sense.

jancborchardt:
Just a question regarding the first iteration – is this similar to how Dropbox and Google Drive or other solutions do it, so we don’t do something entirely different which people will not be used to?

AndyScherzinger:
@jancborchardt they don't offer this at all because they don't have to due to a smart clustering. The UI is taking care of displaying clusters of photos as in time, location, people, etc.
So they have a smart, analyzing server side and an (unknown) nice API to access the photo buckets based on these criteria. Nextcloud is a storage solution so we live and breathe folder structures as the only available element to provide structures so it makes sense to offer this but you will probably never see it in the other, mentioned apps.

jancborchardt:
> they don't offer this at all because they don't have to due to a smart clustering. The UI is taking care of displaying clusters of photos as in time, location, people, etc.

Right, like iOS does as well in photos. We could do this as well instead of imposing a hard structure via folders, right?

AndyScherzinger:
Theoretically, yes. This would then mean that the client (web, Android, etc.) would need to be implement such a view the tricky part is we have an one-for-all approach, we don't just do media we do any kind of line, so it is hard to decide when to cluster (images) and when not to.
Even then "having a folder per month" might still be legit simply because people might want to have this folder structure, but I can't say because I do grouping topic based most of the time and I create a folder per year and change the instant upload folder, so I also kind of use this feature, but atm by hand.

Spacefish:
A indexing solution on the server with an API would be nice. However this is far off IMHO and not related to the android app currently, server infrastructure has to be in place beforehand.
I would image the server to do automatic tagging of photos, based on GPS location and "time based clusters". Such that photos from one trip, end up in the same cluster.
To be performant, one could use elastic search and such to make this tags / metainformation easily queriable / searchable. Putting the pictures in subfolders would not hurt this concept at all.

src/com/owncloud/android/utils/FileStorageUtils.java
Spacefish: I introduced the parameter, i might later on want to upload existing photos (upload all media on the device, which hasn´t been uploaded already). This would require us to read the timestamp from the pictures and use this timestamp for folderpath. Thats why i made it into a parameter on the method.

jancborchardt:
> the tricky part is we have an one-for-all approach, we don't just do media we do any kind of line, so it is hard to decide when to cluster (images) and when not to.

I’d argue that this is what we’re here to solve though. :) cc @LukasReschke @schiessle for the server.

@Spacefish the difficult part about such a feature is that when it’s in, it’s hard to remove. As soon as we have a structure, it will likely need to be in forever. And especially here I’m very hesitant to include it since it adds a lot of complexity and levels for when you just want to have a few photos synced.

jancborchardt:
A giant drawback of this as well is that when you have a vacation which is on the end+beginning of a month or year (such as new year festivities), the pictures will be in completely different folders. That makes no sense at all. ;)

Spacefish:
Yes, there are pro and cons, the solution to put pictures in subfolders based on month, is the same that shotwell choose to order their imported pictures by default.
We can continue bikeshedding ( http://bikeshed.com/ ) or merge the feature. The time required to talk about things like whether we should should pass date/time to a helper, use System.CurrentTime directly, whether a comment is to obvious, or the max length of lines, required was much higher than the actual development time.
Please either merge it or close the PR.

jancborchardt:
> We can continue bikeshedding or merge the feature. The time required to talk about things […] was much higher than the actual development time.

Sorry about that, we really value your contribution. But this is not a simple bikeshedding discussion – while it might help your specific case, it’s something which will actually impair the experience for people who just have a few pictures.

You might have also read recently »[Rule #1 of open-source: no is temporary, yes is forever.](https://twitter.com/solomonstre/status/715277134978113536)« Meaning that if we »just merge the feature« right now, then it will be incredibly hard to remove it in the future should we reconsider. Discussion time vs development time is never an argument for any sensible software. ;)

With Nextcloud we really want to provide a product which is very easy and fun to use. That often means fussing over details, and discussing stuff. Since this feature wasn’t discussed or specced prior to this implementation, we of course have to weigh it now.

jancborchardt:
(Btw @Spacefish I just saw you are based in Stuttgart – you should meet up with @karlitschek and @schiessle some time as they’re there too! Probably time to get a Stuttgart Nextcloud meetup going. ;)

tobiasKaminsky:
As it is an **optional** feature I think it is not such a problem.
It as in enhancement and an opportunity for the user and not something they are forced to...

> A giant drawback of this as well is that when you have a vacation which is on the end+beginning of a month or year (such as new year festivities), the pictures will be in completely different folders.

This is a very rarely scenario and on the opposite site I could say: what if the folder gets > 10k images and is not managable by the user. This feature is, at least for me, a slightly enhancement and is not that strict as the user still can move the files around, like @AndyScherzinger and me are doing it: regular photos in year/month and special images are moved to a meaningful subfolder. 

BTW: As this is the second PR (after #4) that is heavily discussed after developing we should *really* keep this schema: feature request/idea -> discuss -> implement -> testing -> merging. Of course it can happen that a requirement changes, but the features should be so small that it happens rarely.

So long speech, I would merge this, but by default the behaviour remains the same as now.
And on the long term we can try to enhance this by an intelligent server-side sorting system.
(As said before this is nothing that the app should care about)

AndyScherzinger:
I second @tobiasKaminsky's vote here. The standard way is still the standard behavior and in this case I don't see any problems since this feature isn't a one way street and could be removed whenever we come up and ship a smarter processing and wouldn't run into (technical) problems.

AndyScherzinger:
@jancborchardt we would like to get some final feedback / vote because this is the last open PR for the beta release :rocket:

LukasReschke:
Seconding @AndyScherzinger and @tobiasKaminsky here. LGTM.

jancborchardt:
Ok, let’s try this out. Can’t be that bad since it’s optional and apparently lots of people want it. Not going to be against that then ;) :+1: Wording is fine too.

AndyScherzinger:
Thanks for the feedback @jancborchardt 

@LukasReschke next up we will prepare the master branch for a Play store release for the beta channel :) @tobiasKaminsky and I will prepare everything for you to simply build and deploy :)

LukasReschke:
>  @LukasReschke next up we will prepare the master branch for a Play store release for the beta channel :) @tobiasKaminsky and I will prepare everything for you to simply build and deploy :)

🚀

jancborchardt:
@Spacefish @AndyScherzinger @tobiasKaminsky sorry about the longer discussion btw – while the basic points regarding design of course stand, I must have missed that this is an optional subsetting. :)

AndyScherzinger:
No worries, thanks for the feedback. All new features are now on master and we are preparing the beta release :)

PR #36, opened 2016-06-19
Video thumbnail

tobiasKaminsky:
Enable video thumbnail.
Needed:
'enabledPreviewProviders' => array(
'OC\Preview\Movie'
)
in config.php

@jancborchardt we need a video thumbnail.


jancborchardt:
Nice! :) But does this mean that every single person needs to add this config option to their config? Can’t we enable that by default? cc @LukasReschke

Spacefish:
@jancborchardt you need to activate this in your server, but this patch is needed to even be able to see videothumbs in the android app +1
Enabling the Video Preview Provider by default would be nice, but we should keep security in mind. There are multiple bugs lately in ffmpeg which allowed code execution due to malformed container / bytestream headers.. Just think of the stagefright bug in android.

src/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -720,6 +720,7 @@ private void updateLayout() {
                 registerLongClickListener();
             } else {
                 switchToListView();
+//                switchToGridView();
```
przybylski: if unused, please remove

tobiasKaminsky:
@jancborchardt not every user, but every admin.
I have talked to @LukasReschke on IRC an he says that this would be a security problem as the previews are currently not generated in a sandbox.

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -389,6 +404,51 @@ public static ThumbnailGenerationTask getBitmapWorkerTask(ImageView imageView) {
...
+
+        // compute visual center of play button, according to resized image
+        int x1 = resizedPlayButton.getWidth();
+        int y1 = resizedPlayButton.getHeight() / 2;
+        int x2 = 0;
+        int y2 = resizedPlayButton.getWidth();
+        int x3 = 0;
+        int y3 = 0;
+
+        double ym = ( ((Math.pow(x3,2) - Math.pow(x1,2) + Math.pow(y3,2) - Math.pow(y1,2)) *
```
przybylski: a lot of zeros in the calculations so this probably can be simplified ;)

jancborchardt:
> not every user, but every admin.

@tobiasKaminsky yeah, this is still a bit bad cause this results in different UX for different people depending on the server, and this should ideally not happen. Since it’s a security issue, is it possible to generate the previews client-side as well?

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
tobiasKaminsky: I doubt so, as this is a mainly pythagoras and he needs three coordinates. 
To simplify it, I have chosen them to be mostly zero.

tobiasKaminsky:
@jancborchardt when the previews are generated in a sandbox the generation of videos can be enabled by default.
The fact that it is a security risk we should not enable preview on client. Beside the security risk, it is also very difficult as the video has to be downloaded prior to generate a preview locally.
So I think we should do it this way and provide previews for those (home) servers that enable the preview generation and for the others we should provide a good video thumbnail.

jancborchardt:
Okidoke, sounds good

tobiasKaminsky:
So this can go in after we have a dummy video thumbnail @jancborchardt (the third reminder)

AndyScherzinger:
How should the dummy look like? We do have an video icon which we could use as long as we do not have created a thumbnail of the video,  can we not?

jancborchardt:
@tobiasKaminsky aaah I didn’t know that was a request! I assumed you already have the dummy video thumbnail in the filetypes, because it’s here: https://github.com/nextcloud/server/blob/master/core/img/filetypes/video.svg

jancborchardt:
And for when a thumbnail is created, didn’t we already have the play icon overlayed on top of it for the grid view? cc @AndyScherzinger

AndyScherzinger:
@jancborchardt as far as I recall we do, yes.

tobiasKaminsky:
@AndyScherzinger can you make video thumbnails for me?
@jancborchardt yes, play button as overlay on video thumbnails, both in list and grid view

AndyScherzinger:
@tobiasKaminsky absolutely, do you need an icon which is shown when there is no thumbnail or a "play icon" overlay which is put on top of a video thumbnail? (can do both, whatever you need :smile:)

tobiasKaminsky:
Placeholder when there is no video based thumbnail available.

AndyScherzinger:
Erm...  It is all in place :D
https://github.com/nextcloud/android/blob/master/res/drawable-hdpi/file_movie.png

AndyScherzinger:
These are already in use for all videos atm since we didn't have video thumbnails yet.  :)

tobiasKaminsky:
You are right, then we are done here

AndyScherzinger:
@tobiasKaminsky can you rebase and resolve the conflict? I can also give it a try but I guess it is saver when you do it since it is your PR. I'll test drive it afterwards. Code changes look good to me so far :)

AndyScherzinger:
The placeholders work fine for me, the real video thumbnails need to be checked by someone else :(

device-2016-07-02-181725

tobiasKaminsky:
This is already working :)
But we have to get this running on your pc ;-)

AndyScherzinger:
Works, see screenshot. But only if the server already created the thumbnail. At first I didn't log into the web view and opened the folder which seems to trigger the thumbnail generation. After I did that the Android client was able to fetch the thumbnails.

device-2016-07-14-223419

AndyScherzinger:
:shipit:

AndyScherzinger:
@przybylski do you have the time to do a review? I already did one :)

przybylski:
@AndyScherzinger I already did a review some time ago, if nothing changed since than 👍

AndyScherzinger:
Sorry @przybylski I missed that :(

@tobiasKaminsky can you update it to latest master for another test round before merging?

tobiasKaminsky:
@AndyScherzinger is updated

AndyScherzinger:
Sweet, will do a test tomorrow or beginning of next week.

AndyScherzinger:
Test successful, can imho be merged. :+1:

AndyScherzinger:
LGTM is broken again... grrr

AndyScherzinger:
:+1:

PR #43, opened 2016-06-20
Thumbnail during upload

tobiasKaminsky:


AndyScherzinger:
Checked the code LGTM :+1:

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -157,6 +157,14 @@ public ThumbnailGenerationTask(ImageView imageView, FileDataStorageManager stora
             mAccount = account;
         }
 
+        public ThumbnailGenerationTask(FileDataStorageManager storageManager, Account account){
```
przybylski: you might wanna have a look on NonNull annotation in android :)

tobiasKaminsky: > Denotes that a parameter, field or method return value can never be null. 

But it is the other way, the parameter should never be null, but it can.

przybylski: So handling null parameter via throwing exception seems a bit odd

tobiasKaminsky: It is adapted from ownCloud ;)
This should not happen as far as I have seen the code, so I would leave it this way.

przybylski:
👍

AndyScherzinger:
:+1:

tobiasKaminsky:
Please review again.
It was not working if "file is kept in original folder" is used.

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -177,6 +186,9 @@ protected Bitmap doInBackground(Object... params) {
                 }
 
                 mFile = params[0];
+                if (params.length == 2){
```
przybylski: This is wrong on so many levels... but i'll let it slide since it is adopted from ownCloud

przybylski:
👍

AndyScherzinger:
Code LGTM - unfortunately I can't test it right now since instant upload is broken for Android N 
(Thanks Google...)

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
tobiasKaminsky: As I am always eager to learn/code better, please help me and show a better solution.
First I wanted to create a new inner class that holds both values and can be passed to the task. But it was too much work  ;-)

AndyScherzinger: So, to be merged to master or wait for @przybylski's coding ninja skills?

przybylski:
Push it, Ill make a PR

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
przybylski: Push it, Ill make a PR

AndyScherzinger:
LGTM

AndyScherzinger:
Merged 🚀

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
AndyScherzinger: Sweet! Always a pleasure pushing things forward with you guys! 💯

PR #51, opened 2016-06-22
Two level thumbcache + async I/O

Spacefish:
Just compare to the normal version, while scrolling in a folder

fixed a bug, which retriggered thumbnail loading, after a thumbnail was loaded and the view was scrolled at the end of the view (beginning and and aswell as backscroll).

All I/O is now done in a AsyncTask, even cache I/O on the flash, as it really introduced stutter, because flash read access aswell as JPEG->Bitmap Decoding is done on the UI Thread.

I added a bitmap in memory cache for thumbnails (LRUCache), this cache is queried in the AsyncTask aswell (should not be).

I think final solution would be

Check Memory Bitmap Cache
  Hit -> Use the bitmap directly on UI Thread
  Miss -> Start AsyncTask and query JPEG DiskCache
     Hit -> Decode and set the save bitmap in memcache aswell as set it on the UI Thread
     Miss -> generate the thumbnail, save in both caches and update bitmap in UI Thread

Otherwise scrolling large folders with a lot of Thumbs is a pain / looks stuttery.

### DIFF
--- a/src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -140,7 +140,20 @@ public class ThumbnailsCacheManager {
         return null;
     }
 
-    public static class ThumbnailGenerationTask extends AsyncTask<Object, Void, Bitmap> {
+    private static void waitForCacheInit() {
+        synchronized (mThumbnailsDiskCacheLock) {
+            // Wait while disk cache is started from background thread
+            while (mThumbnailCacheStarting) {
+                try {
+                    mThumbnailsDiskCacheLock.wait();
+                } catch (InterruptedException e) {
+                    Log_OC.e(TAG, "Wait in mThumbnailsDiskCacheLock was interrupted", e);
+                }
+            }
+        }
+    }
+
+    public static class ThumbnailGenerationTask extends AsyncTask<Object, Bitmap, Bitmap> {
         private final WeakReference<ImageView> mImageViewReference;
         private static Account mAccount;
         private Object mFile;
@@ -155,11 +168,13 @@ public class ThumbnailsCacheManager {
                 throw new IllegalArgumentException("storageManager must not be NULL");
             mStorageManager = storageManager;
             mAccount = account;
+            waitForCacheInit();
         }
 
         public ThumbnailGenerationTask(ImageView imageView) {
             // Use a WeakReference to ensure the ImageView can be garbage collected
             mImageViewReference = new WeakReference<ImageView>(imageView);
+            waitForCacheInit();
         }
 
         @Override
@@ -167,6 +182,24 @@ public class ThumbnailsCacheManager {
             Bitmap thumbnail = null;
 
             try {
+                mFile = params[0];
+
+                // try cache
+                String cacheKey = null;
+                if (mFile instanceof OCFile) {
+                    cacheKey = ((OCFile) mFile).getRemoteId();
+                } else if (mFile instanceof File) {
+                    cacheKey = String.valueOf(((File) mFile).hashCode());
+                }
+                Log_OC.d(TAG, "CacheKey is " + cacheKey);
+                if (cacheKey != null) {
+                    thumbnail = getBitmapFromDiskCache(cacheKey);
+                }
+                if (thumbnail != null) {
+                    Log_OC.d(TAG, "Early cache hit in ThumbnailGenerationTask");
+                    return thumbnail;
+                }
+
                 if (mAccount != null) {
                     OwnCloudAccount ocAccount = new OwnCloudAccount(mAccount,
                             MainApp.getAppContext());
@@ -174,15 +207,14 @@ public class ThumbnailsCacheManager {
                             getClientFor(ocAccount, MainApp.getAppContext());
                 }
 
-                mFile = params[0];
-                
                 if (mFile instanceof OCFile) {
                     thumbnail = doOCFileInBackground();
-                }  else if (mFile instanceof File) {
+                } else if (mFile instanceof File) {
                     thumbnail = doFileInBackground();
-                //} else {  do nothing
+                    //} else {  do nothing
                 }
 
+                addBitmapToCache(cacheKey, thumbnail);
                 }catch(Throwable t){
                     // the app should never break due to a problem with thumbnails
                     Log_OC.e(TAG, "Generation of thumbnail for " + mFile + " failed", t);
@@ -190,6 +222,10 @@ public class ThumbnailsCacheManager {
                         System.gc();
                     }
                 }
+
+            //Bitmap preview = Bitmap.createScaledBitmap(thumbnail, 10,10, true);
+
+            // publishProgress(preview);
 
             return thumbnail;
         }
@@ -206,7 +242,14 @@ public class ThumbnailsCacheManager {
                         tagId = String.valueOf(mFile.hashCode());
                     }
                     if (String.valueOf(imageView.getTag()).equals(tagId)) {
-                        imageView.setImageBitmap(bitmap);
+                        final ThumbnailsCacheManager.AsyncDrawable asyncDrawable =
+                                new ThumbnailsCacheManager.AsyncDrawable(
+                                        imageView.getResources(),
+                                        bitmap,
+                                        this
+                                );
+                        imageView.setImageDrawable(asyncDrawable);
+                        // imageView.setImageBitmap(bitmap);
                     }
                 }
             }
@@ -247,6 +290,8 @@ public class ThumbnailsCacheManager {
             OCFile file = (OCFile)mFile;
 
             final String imageKey = String.valueOf(file.getRemoteId());
+
+            Log_OC.d(TAG, "doOCFileInBackground CacheKEy " + imageKey);
 
             // Check disk cache in background thread
             Bitmap thumbnail = getBitmapFromDiskCache(imageKey);
@@ -341,6 +386,7 @@ public class ThumbnailsCacheManager {
 
             final String imageKey = String.valueOf(file.hashCode());
 
+            Log_OC.d(TAG, "doFileInBackground CacheKEy " + imageKey);
             // Check disk cache in background thread
             Bitmap thumbnail = getBitmapFromDiskCache(imageKey);
 
@@ -367,6 +413,7 @@ public class ThumbnailsCacheManager {
         if (bitmapWorkerTask != null) {
             final Object bitmapData = bitmapWorkerTask.mFile;
             // If bitmapData is not yet set or it differs from the new data
+
             if (bitmapData == null || bitmapData != file) {
                 // Cancel previous task
                 bitmapWorkerTask.cancel(true);
@@ -392,7 +439,7 @@ public class ThumbnailsCacheManager {
     }
 
     public static class AsyncDrawable extends BitmapDrawable {
-        private final WeakReference<ThumbnailGenerationTask> bitmapWorkerTaskReference;
+        private final ThumbnailGenerationTask bitmapWorkerTaskReference;
 
         public AsyncDrawable(
                 Resources res, Bitmap bitmap, ThumbnailGenerationTask bitmapWorkerTask
@@ -400,11 +447,11 @@ public class ThumbnailsCacheManager {
 
             super(res, bitmap);
             bitmapWorkerTaskReference =
-                    new WeakReference<ThumbnailGenerationTask>(bitmapWorkerTask);
+                    bitmapWorkerTask;
         }
 
         public ThumbnailGenerationTask getBitmapWorkerTask() {
-            return bitmapWorkerTaskReference.get();
+            return bitmapWorkerTaskReference;
         }
     }
 }
--- a/src/com/owncloud/android/ui/adapter/DiskLruImageCache.java
+++ b/src/com/owncloud/android/ui/adapter/DiskLruImageCache.java
@@ -21,15 +21,20 @@ package com.owncloud.android.ui.adapter;
 
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.lang.ref.SoftReference;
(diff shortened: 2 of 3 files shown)


AndyScherzinger:
Wow! I get the idea but this is not my area of expertise so calling @tobiasKaminsky and @przybylski to the rescue :helicopter:

src/com/owncloud/android/ui/adapter/DiskLruImageCache.java
```diff
@@ -43,6 +48,13 @@
     private static final int CACHE_VERSION = 1;
     private static final int VALUE_COUNT = 1;
     private static final int IO_BUFFER_SIZE = 8 * 1024;
+    private static final int RUNTIME_MAX_MEMORY = (int) (Runtime.getRuntime().maxMemory() / 1024);
+    private static final int MEMORY_LRU_CACHE_SIZE = RUNTIME_MAX_MEMORY / 8;
```
przybylski: I don't think that this is the best way to determine the size of cache, on some devices it might lead to throwing `IllegalArgumentException` in `LruCache` constructor.

przybylski:
It is great to see someone is caring about performance :)
But when you are introducing intermediate cache layer what is the additional memory footprint on this for smaller and larger directories? It wouldn't be very nice if nextCloud app would be killed often by the system to save memory.

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -167,22 +167,39 @@ protected Bitmap doInBackground(Object... params) {
             Bitmap thumbnail = null;
 
             try {
+                mFile = params[0];
+
+                // try cache
+                String cacheKey = null;
+                if (mFile instanceof OCFile) {
+                    cacheKey = ((OCFile) mFile).getRemoteId();
```
przybylski: Maybe OCFile should implement hashCode ? i think that remote id is unique enough to identify OCFile instance. And this code could be a bit simplified then :)

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -167,22 +167,39 @@ protected Bitmap doInBackground(Object... params) {
...
+                } else if (mFile instanceof File) {
+                    cacheKey = String.valueOf(((File) mFile).hashCode());
```
przybylski: Object has hashCode method, there is no need to cast it to `File`

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -191,6 +208,10 @@ protected Bitmap doInBackground(Object... params) {
                     }
                 }
 
+            //Bitmap preview = Bitmap.createScaledBitmap(thumbnail, 10,10, true);
```
przybylski: Please remove dead code

src/com/owncloud/android/ui/adapter/DiskLruImageCache.java
```diff
@@ -43,6 +48,13 @@
...
+    private LruCache<String, Bitmap> memoryLruCache = new LruCache<String, Bitmap>(MEMORY_LRU_CACHE_SIZE) {
```
przybylski: String comparison even for HashMap is a bit inefficient. Have you considered using Long?

src/com/owncloud/android/ui/adapter/DiskLruImageCache.java
```diff
@@ -76,6 +88,11 @@ public void put( String key, Bitmap data ) {
         DiskLruCache.Editor editor = null;
         String validKey = convertToValidKey(key);
         try {
+            // store in memoryLruCache
```
przybylski: comments shouldn't be obvious :)

Spacefish:
This is just an first shot to try if i can improve app performance.. I used my own patch during the day and noticed some drawbacks..

1.) Android only runs CPU_NUM + 1 AsyncTasks at once, the rest get queued (up to 128 in Queue). This leads to "waves" of thumbnail refresh, as for example 5 Tasks download pictures from the owncloud server and no Task is availiable to fetch the thumbnail from the memory or disk cache.

2.) The Bitmaps takes so much memory, that the cache only holds like 1,5 sites of thumbnails

The Code isn´t optimal.. As pointed out OCFile and File could do with a common interface to expose a getCacheKey() function or just both implement hashCode correctly..
The whole Thumnail generation process could do with a rewrite.. The View shouldn´t spawn the ThumbnailAsyncTask, instead it should call ThumbnailGenerator and pass a callback, the ThumbnailGenerator should return a handle (to cancel the generation if required).

I would change the whole process.. 
Having 2 Threads with queues, which do: Thumbnail Download and the other Thumbnail Resizing / Generation and Disk Cache. Work is scheduled and the threads do the work in sequence. This way a Download does not block a generation or disk cache query.
Memory cache should be queried synchronously.

So the process would look like:

ThumbManager queries memory cache
 -> exists: return thumb
 -> does not exist: dispatch work to disk cache thread
Disk cache thread queries disk cache in background
 -> exists: set in memcache and call callback on UI
 -> does not exist: dispatch work to download thread
Download thread downloads image and dispatches message to cache thread, cache thread caches image and dispatches message to UI

To minimize memory impact i would borrow a common concept from 3d Game design. Texture Streaming. Having for example 3 memory caches with different quality versions of the picture.
Cache 1: Full size (10 pictures)
Cache 2: Reduzed size 32x32px (250 pictures)
Cache 3: Color of picture RGB value (10.000 pictures)

caches are queried in order, if cache 1 misses the process above is initiated in the background, however Cache 2 and Cache 3 is checked in the UI thread directly after dispatching the message.
For the user the thumbnail is loaded in a low quality version and once the background work finishes is replaced by the full quality version..
Like this: https://www.youtube.com/watch?v=3jdItPONjHE

tobiasKaminsky:
As I am not an expert with async tasks and handling UI I would leave this to you.
It would really cool if you can enhance this :)

tobiasKaminsky:
@Spacefish what do you think? Would it be a good idea to reduce the quality of the jpg thumbnails?

Spacefish:
i don´t think so, the resolution of the thumb should match the real display resolution, everything above this is wasted.. Currently the images are written to the disk cache as jpeg with quality 70, decoding and encoding the jpeg takes some cpu time. I don´t think the resolution is such big of a deal for the cache. The bitmaps in memory take far more space.. :(

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -82,11 +84,22 @@
...
     {
         if(file == null || fileIcon == null)
             return;
 
+        synchronized (fileIcon) {
+            // don´t do work twice if state is already present or queued!
+            if (imageViewStates.containsKey(fileIcon) && imageViewStates.get(fileIcon) == file.getRemoteId()) {
+                return;
+            }
+            imageViewStates.remove(fileIcon);
```
przybylski: https://docs.oracle.com/javase/7/docs/api/java/util/Map.html#put(K,%20V)
`If the map previously contained a mapping for the key, the old value is replaced by the specified value.`

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -66,12 +73,209 @@
...
+        protected FileDataStorageManager storageManager;
+        protected Account account;
+
+        @Override
+        public int compareTo(DiskCacheRequest c) {
+            return this.priority.compareTo(c.priority);
+        }
+    }
+
+    private static PriorityBlockingQueue<DiskCacheRequest> diskCacheRequestQueue = new PriorityBlockingQueue<>();
```
przybylski: Since this is a part of class it should be declared at the top of the class.

przybylski: And follow naming convention

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -66,12 +73,209 @@
...
+                                @Override
+                                public void run() {
+                                    final ImageView imageView = dcr.fileIcon;
+                                    imageView.setImageBitmap(thumbCopy);
+                                }
+                            });
+                        }
+                    }catch(Throwable t){
+                        // the app should never break due to a problem with thumbnails
+                        Log_OC.e(TAG, "Generation of thumbnail for " + dcr.file + " failed", t);
```
przybylski: In some extreme cases this will throw another OutOfMemoryError.

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
przybylski: All elements have the same priority, does it needs to be priority queue instead of other types?

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -66,12 +73,209 @@
...
+            imageViewStates.put(fileIcon, file.getRemoteId());
+        }
+
+        fileIcon.setImageBitmap(ThumbnailsCacheManager.mDefaultImg);
+
+        DiskCacheRequest dcr = new DiskCacheRequest();
+        dcr.file = file;
+        dcr.fileIcon = fileIcon;
+        dcr.storageManager = storageManager;
+        dcr.account = account;
```
przybylski: those assignments should be done in `DiskCacheRequest` constructor

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
przybylski: Since you require elements to be unique maybe you can implement inheriting queue which would take care of element uniqueness instead of putting this logic in this class, which doesn't belong here.

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -66,12 +73,209 @@
...
+
+        return resultBitmap;
+    }
+
+    private static Bitmap getOCThumb(OCFile file, OwnCloudClient client, Account account) {
+        final String imageKey = String.valueOf(file.getRemoteId());
+
+        Bitmap thumbnail = null;
+
+        Log_OC.d(TAG, "NOT FOUND DOWNLOADING!");
```
przybylski: This comment isn't 100% clear to me at first sight

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -66,12 +73,209 @@
...
+        // Not found in disk cache
+        if (file.needsUpdateThumbnail() || true) {
```
przybylski: soooo, always true?

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -66,12 +73,209 @@
...
+            int px = Math.round(r.getDimension(R.dimen.file_icon_size_grid));
+
+            // Download thumbnail from server
+            OwnCloudVersion serverOCVersion = AccountUtils.getServerVersion(account);
+            if (client != null && serverOCVersion != null) {
+                if (serverOCVersion.supportsRemoteThumbnails()) {
+                    GetMethod get = null;
+                    try {
+                        String uri = client.getBaseUri() + "" +
+                                "/index.php/apps/files/api/v1/thumbnail/" +
```
przybylski: Is this URL constant? Does it depends on server configuration etc? @LukasReschke

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -66,12 +73,209 @@
...
+
+                            // Add thumbnail to cache
+                            if (thumbnail != null) {
+                                addBitmapToCache(imageKey, thumbnail);
+                            }
+                        } else {
+                            client.exhaustResponse(get.getResponseBodyAsStream());
+                        }
+                    } catch (Exception e) {
+                        e.printStackTrace();
```
przybylski: This should be handled by Log_OC

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
tobiasKaminsky: This is as long as the api exists contant.
In an additional PR (resized images) I am using a different API based on the server version.

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
Spacefish: That's just copy paste of old Code to try it out. Plan is to completely refactor ThumbsManager and introduce an ThumbImageView which subclasses ImageView to  automatically talk to ThumsManager. This ThumbImageView can then be used inside the different lists without them containing Code starting Tasks for Thumbgeneration and so on. (Better separation of concerns and Handling of the differrent cases in just one Place instead of 3)

przybylski: Thats even better! :)

tobiasKaminsky: This sounds awesome!

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
przybylski: But this is true for most of the cases in ThumbnailsCacheManager so no need to worry.

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
Spacefish: i created issue #77 so we can discuss this before implementation.

iskradelta:
Hey how is it going with this PR?

The preview/thumbnail generation is very slow with the beta app - the requests and updates to view is done serially, this one has merge conflicts with master, is it worth to merge it/resolve-conflicts?

tobiasKaminsky:
I have updated the branch to my best knowledge as I did not wrote the code.
Please have a look at the discussion and decide if you want to continue on it.
Maybe @Spacefish will shed us some light on this?

tobiasKaminsky:
With #2871 we will have a proper library for handling our image cache, so this is not needed.

PR #62, opened 2016-06-23
HotFix master branch: Set fixed lib version

AndyScherzinger:
For some reason jitpack is not refreshing the snapshot build so the branch won't build thus I created a lib release 1.0.1 and set the dependency to 1.0.1 instead of -SNAPSHOT

ATM This is necessary in order to get master to compile again. Sorry for any inconvenience :(

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ dependencies {
     /// dependencies for app building
     compile name: 'touch-image-view'
 
-    compile 'com.github.nextcloud:android-library:-SNAPSHOT'
+    compile 'com.github.nextcloud:android-library:1.0.1'
     compile "com.android.support:support-v4:${supportLibraryVersion}"
     compile "com.android.support:design:${supportLibraryVersion}"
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
FYI the display name change in the lib simply adds the possibility to retrieve the full name, we haven't utilized this yet but I will asap for the drawer implementation but that isn't a blocking issue so we can still release the app whenever we think it is ready and can ship full name display any time later :)

przybylski:
+1 we always should depend on stable version not repos snapshot

AndyScherzinger:
That's true! Absolutely :+1:

tobiasKaminsky:
:+1: 
We should keep the version mostly the same!

AndyScherzinger:
@tobiasKaminsky why keep it mostly the same? We could create new versions for new features which would only be used on a feature branch and push the new version/dependency with the feature PR to master. Not sure I got your point :S

tobiasKaminsky:
Yeah, I think I was confused. Please forget my idea ;)

PR #64, opened 2016-06-25
use hashCode() instead of getRemoteId() on OCFile

Spacefish:
It´s possible to generate cacheKeys in the same maner for File and OCFile isntances that way


przybylski:
LGTM

tobiasKaminsky:
I am not sure if this is still valid: https://github.com/owncloud/android/issues/817
Can you have a look on this as the upload behaviour has changed.

(and to avoid that the PR has to be redone/changed a lot, I suggest to first open an issue where we can discuss it and after that implement it).

Spacefish:
i think we can close this

PR #74, opened 2016-06-26
actual -> current

mabkenar:
Longman Dictionary says: Do not use actual to mean 'at the present time'. Use current or present
http://www.ldoceonline.com/dictionary/actual
http://english.stackexchange.com/questions/29654/using-actual-to-signify-current

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -140,7 +140,7 @@ Examples for different versions:
 * 8.12.2 ```80120200```
 * 9.8.4-rc18 ```90080418```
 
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
:+1:

tobiasKaminsky:
What is the second change about? I see no difference.

mabkenar:
@tobiasKaminsky There should be no other change. Only two replacement of actual -> current in line 143 and that's all. I am also confused why GitHub shows a (null) change in line 175. I checked, the lines are char-to-char the same.

AndyScherzinger:
@tobiasKaminsky should be fine then imho.

tobiasKaminsky:
It remains strange ;-)
:+1:

PR #80, opened 2016-06-27
RC 1.1.0 preparation

AndyScherzinger:
@tobiasKaminsky @przybylski @LukasReschke @jancborchardt feel free to :+1: -> merge -> GH tag/release -> build -> upload to Google Play Beta Channel :)

rc tag would be: ```rc-Mayor.Minor.Hotfix-RCIncrement``` --> ```rc-1.1.0-01```

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -3,7 +3,8 @@
   ownCloud Android client application
 
   Copyright (C) 2012  Bartek Przybylski
-  Copyright (C) 2012-2015 ownCloud Inc.
+  Copyright (C) 2012-2016 ownCloud Inc.
+  Copyright (C) 2016 Nextcloud
 
   This program is free software: you can redistribute it and/or modify
   it under the terms of the GNU General Public License version 2,
@@ -19,8 +20,8 @@
 -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.owncloud.android"
-    android:versionCode="10000100"
-    android:versionName="1.0.1" >
+    android:versionCode="10010000"
+    android:versionName="1.1.0" >
 
     <uses-sdk
         android:minSdkVersion="14"
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -20,7 +20,7 @@
 -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.owncloud.android"
-    android:versionCode="10010000"
+    android:versionCode="10010001"
     android:versionName="1.1.0" >
 
     <uses-sdk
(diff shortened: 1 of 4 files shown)


LukasReschke:
👍

AndyScherzinger:
:+1:

AndyScherzinger:
awaiting 2nd :+1:

tobiasKaminsky:
Typo "hostes" -> "hosted
What about the version schema?

>beware, that beta releases for an upcoming version will always use the minor and hotfix version of the current release not the upcoming so that the version code of the upcoming stable release will always be higher so the current beta release can be updated to the latest hot fix release.

1 001 00 00 is 1.1.0-rc0, but what is then stable 1.1.0?

AndyScherzinger:
If everything goes according to plan the same, since Lukas can then mark is as stable in the store. Else we will have to bump the rc increment (shipping bugfixes) so the final/stable could be 10010006 (as in having had 5 release candidates and then a final release with changes) or 10010006 (as in having had 5 release candidates and then declaring the last rc one final)

>beware, that beta releases for an upcoming version will always use the minor and hotfix version of the current release not the upcoming so that the version code of the upcoming stable release will always be higher so the current beta release can be updated to the latest hot fix release.

That is something that I would change like mentioned above I made the former mistake because then from the version code it gives you the impression that the feature set is closer to the old/last minor version which is misleading since a RC is rather close to the next/upcoming minor version.

Typo has been fixed

AndyScherzinger:
@tobiasKaminsky Both ways are fine. with the RCs always having the former minor version number we make sure the stable of the upcoming minor release has always a higher version code and always ends with 00 but when you look at the version code "some years later" it might be confusing since 10010003 is the 3rd RC for 1.2.0 on the other hand with 00 being the RC0 you never know looking at the last two digits if it is a RC or a stable version. :( So both have their ups and downs.

tobiasKaminsky:
I know this is not urgent and we can discuss it later, but this is how they do it:
https://www.jayway.com/2015/03/11/automatic-versioncode-generation-in-android-gradle/

> So, version “1.2.3-SNAPSHOT” would then result in version code 1020300, whereas version “1.2.3-RC4” would result in version code 1020304. Final releases always gets assigned 99 as their last code, to make sure that they always have a higher version code than Release Candidates.

AndyScherzinger:
Nice! This looks perfect!!!

AndyScherzinger:
@tobiasKaminski can you +q so I can simply merge when I am done with the changelog? :)

AndyScherzinger:
I will also update the MD for the jayway version number schema description.

tobiasKaminsky:
LGTM and I hope you do it in a right way ;) ;)

AndyScherzinger:
I'll ping you on IRC when I am done on the branch so you can have a final look :)

AndyScherzinger:
fine with you @tobiasKaminsky ?

tobiasKaminsky:
Yes, fine :+1:

AndyScherzinger:
:tada:

PR #87, opened 2016-07-01
check if correct adapter is used

tobiasKaminsky:


AndyScherzinger:
LGTM

src/third_parties/in/srain/cube/GridViewWithHeaderAndFooter.java
```diff
@@ -412,6 +412,11 @@ public void setNumColumns(int numColumns) {
         }
     }
 
+    public boolean isCorrectAdapter(){
+        return getAdapter() == null ||
+               (getAdapter() != null && getAdapter() instanceof HeaderViewGridAdapter);
```
przybylski: You don't need to do nullchecking: http://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.20.2

przybylski:
But is checking just for an adapter type a root cause of this crash? Maybe something else is leading to it.

src/third_parties/in/srain/cube/GridViewWithHeaderAndFooter.java
tobiasKaminsky: Right, removed.

tobiasKaminsky:
https://github.com/nextcloud/android/blob/d6ecf06f39bcf72e2fcf0d3c3d158f14e98259fc/src/third_parties/in/srain/cube/GridViewWithHeaderAndFooter.java#L182
here it is checked if mAdapter instanceof HeaderViewGridAdapter and otherwise it results in an exception.
The root cause is that we switch between ListAdapter and HeaderViewGridAdapter. Sometimes loading a directory takes too long and therefore it gets a race conflict and the old adapter is still in use, but the new "addFooterView" method is already called.

przybylski:
👍 but i still thinks this isn't the way to fix it

tobiasKaminsky:
We have some time till release, so feel free to fix it in a better way.
I also think that is not a solution, but a workaround...

tobiasKaminsky:
@przybylski I suggest to put this in RC2, but try to get a better solution?

AndyScherzinger:
Sounds good to me :+1:

przybylski:
ok to me 👍

AndyScherzinger:
review check seems to have a hickup...

AndyScherzinger:
:+1:

AndyScherzinger:
:+1:

AndyScherzinger:
merged, will be shipped with 1.1.0 RC2

PR #89, opened 2016-07-02
New launcher icon

AndyScherzinger:
As wished and created by @jancborchardt a revamped launcher icon which has material design compliant margin.

Only thing we are missing now for the next update after this on is tinted and shaded edges.

pinging @jancborchardt @LukasReschke @tobiasKaminsky @przybylski for review.


tobiasKaminsky:
:+1:

mar1u5:
Personally I dont like it, because the other one was much more simple. Keep in mind Material Design :grin:

Would like to hear something from @jancborchardt about this :grin:

AndyScherzinger:
This icon has been done and requested to be added by @jancborchardt :grin:

mar1u5:
cc @nextcloud/designers

enoch85:
:+1:

przybylski:
I'm voting 👎 they don't follow material design guidelines

jancborchardt:
We talked about it and think it looks way nicer with the background than the bland blue. It will also be like that for the iOS app and the favicon of the web interface. You can already see the background in the log in view of the web interface, our website and basically all our social media profiles. :)

Sizing- and radius-wise it follows material guidelines – the previous icon did not. I left out the tinted and shaded edges intentionally because they look strange with that background as it’s not a defined object.
You can check out how it looks like roughly from the demo images @MorrisJobke prepared: https://cloud.morrisjobke.de/s/cvBHuY3n8UvvOl9

AndyScherzinger:
How should be go on now? We have vote for and against. Personally I would also vote for merging because as mentioned by Jan all other applications of the eco system will use it. So I value consistency over simplicity here. 

Who is now able to decide if it gets merged or not?

przybylski:
With my super limited graphics skill and this website: https://romannurik.github.io/AndroidAssetStudio/icons-launcher.html
I made this:
ic_launcher
[ic_launcher.zip](https://github.com/nextcloud/android/files/344953/ic_launcher.zip)

Henni:
I remember a discussion where the idea was mentioned to change the login background on the web with every new release.
In that case this icon would only make sense if the app release cycle were bound to the server release cycle, which it currently isn't (correct me if I'm wrong).

AndyScherzinger:
In general the release cycles differ while we will still try to support new features of a server release within the app. Keeping the background and icon in sync with the server release isn't possible due to the fact that the app will always work with several versions of the server, so we would then show a new background even though the server might be one or several versions behind.

MorrisJobke:
I like it 👍

mar1u5:
Because of the many :+1: I think we could do this :wink:

AndyScherzinger:
So merge then? Just asking Since for now it seems this is the last remaining issue which thus is blocking the upcoming release.

mar1u5:
Lets ask the community in the forum and maybe make a poll :wink:

MorrisJobke:
> Lets ask the community in the forum and maybe make a poll 😉

If we do this for every PR -> 🙈 

I would say, that we should try it. We can revert it anyway if the feedback is negative, right? ;)

mar1u5:
Lets do this :muscle:

MorrisJobke:
@AndyScherzinger As this is your repo you should press the merge button.

przybylski:
Noooooo!
640px-rejtan_at_sejm_of_1773_by_jan_matejko _1866
(source https://en.wikipedia.org/wiki/Rejtan_(painting) )

AndyScherzinger:
0_o

przybylski:
I still think that it doesn't follow material design, but I got outvoted, so I as well can make a scene :D

jancborchardt:
Come on, let's move forward. We don't discuss that much over code changes either (well often we do ;D) but thisbis a bit out of proportion. ;)

Merging so we don't get stuck.

Icantcodeatall:
Wow the new icon looks older than the older one.

jancborchardt:
Well, people _really_ like the background on the log in screen, so it kinda makes sense to use it other places too.

Icantcodeatall:
Actually I don't think it makes sense to use the background of the login screen on the icon, its two separate things.
Imho it makes the icon looks less clean, especially on the smaller sizes, and not really material-design compliant if it makes any sense.
But maybe the people will think as you, we will see !

williambargent:
I'm not a fan either, could we not just have the logo on a transparent background. Does it need to be inside a box, quite a few apps are doing this.

jancborchardt:
@williambargent that's against the guidelines as well.

Folks – on stuff like this everyone has different opinions. Simply making a decision here and moving on prevents us from losing focus. We just need to be pragmatic and solve real problems. Your time is better spent reviewing pull requests or fixing issues. :)

jancborchardt:
Folks – on stuff like this everyone has different opinions. Simply making a decision here and moving on prevents us from losing focus. We just need to be pragmatic and solve real problems. Your time is better spent reviewing pull requests or fixing issues. :)

Icantcodeatall:
Maybe sticking to the former material-design compliant icon would have been more pragmatic then mhh ?
But you are right, let's focus on things that actually matter and let the public opinion be the judge for this particular "issue".

Issue #25, opened 2016-06-15
Instant upload for the not-currently active account


AndyScherzinger:
copied from the forum https://help.nextcloud.com/t/android-app-one-app-instance-for-each-account/605
As far as I understand the request the issue is that instant upload only works for the currently active account:

> Hello,
> sorry if this is the wrong place to post about the Android app. I've tried to use the mail address
> apps@nextcloud.com (proposed by the app itself) but got the answer "no such user".
> So this is my feature wish:
> 
> I've two owncloud accounts as a normal user (two
> more as admin) and would like to have different instances of the app
> each with it's own account assigned. Is that possible? At this time the
> automatic photo/video upload works only on the active account. I would
> like to assigned it to only one account and have this always active, so every
> uploaded photo will be stored under the same account. The other accounts should be
> active also at the same time.
> 
> Thank you
> Jürgen

Looping in @jancborchardt for UX aspects and @przybylski and @tobiasKaminsky for general discussion/approval of the feature request

bes1002t:
mh I don't know whether this is an good idea or not. In my opinion we should implement an account management system in our android application. That would be much cleaner than creating a new instance for every account.

tobiasKaminsky:
Multiple instances of the app are not possible.
I remember that there is a PR in ownCloud that, at least, shows to which account the instant upload is set.
Technical background: the instant upload is not always to the current accout, but to the account which was in use when the instant upload is defined.

So my proposal is to make the account to which instant upload is bound to easier to choose.
@jancborchardt for designing.

tobiasKaminsky:
https://github.com/owncloud/android/pull/1637
@LukeOwncloud maybe you want to help us here? I promise that your PR will be handled much faster :+1:

jgrete:
Hello,

I'm the OP from help.nextcloud.com.
I've learned multiple instances of an app are not possible. OK, an account managment would do also.
The following are just my thoughts:
- the active account should be visible all the time (maybe next to "Nextcloud" ?)
- choosing the active account is misplaced in the settings and would be better placed in the left  
  menu at toplevel as a dropdown, so changing an account is easier
- the account used for uploading should be marked as such in the dropdown
- if the upload account is not the active acount no instant upload would happen, wouldn't it?
  That was the reason many of my instant uploads didn't work. Is there a better solution?

As I said - just my thoughts :-)

Thank you very much
Jürgen

LukeOwncloud:
[...]

PR #98, opened 2016-07-06
Instant upload in the context of multi account support

AndyScherzinger:
Work in progress PR for #25

### DIFF
--- a/src/com/owncloud/android/ui/activity/Preferences.java
+++ b/src/com/owncloud/android/ui/activity/Preferences.java
@@ -809,6 +809,7 @@ public class Preferences extends PreferenceActivity
         SharedPreferences.Editor editor = appPrefs.edit();
         editor.putString(PreferenceKeys.INSTANT_UPLOAD_PATH, mUploadPath);
                 .putString(INSTANT_UPLOAD_PATH_ACCOUNT, mUploadPathAccount);
+        editor.putString("instant_upload_path_account", mUploadPathAccount);
         editor.commit();
     }
 
@@ -833,6 +834,7 @@ public class Preferences extends PreferenceActivity
         SharedPreferences.Editor editor = appPrefs.edit();
         editor.putString(PreferenceKeys.INSTANT_VIDEO_UPLOAD_PATH, mUploadVideoPath);
                 .putString(INSTANT_VIDEO_UPLOAD_PATH_ACCOUNT, mUploadVideoPathAccount);
+        editor.putString("instant_video_upload_path_account", mUploadVideoPathAccount);
         editor.commit();
     }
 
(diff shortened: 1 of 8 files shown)


src/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -560,23 +566,33 @@ private AppCompatDelegate getDelegate() {
...
     /**
-     * Save the "Instant Upload Path" on preferences
+     * Save the "Instant Upload Path" and corresponding account on preferences
      */
     private void saveInstantUploadPathOnPreferences() {
         SharedPreferences appPrefs =
                 PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
         SharedPreferences.Editor editor = appPrefs.edit();
         editor.putString("instant_upload_path", mUploadPath);
+        editor.putString("instant_upload_path_account", mUploadPathAccount);
```
przybylski: putString returns Editor so you can chain but thats nothing big

src/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -587,17 +603,20 @@ private void loadInstantUploadVideoPath() {
         SharedPreferences appPrefs =
                 PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
         mUploadVideoPath = appPrefs.getString("instant_video_upload_path", getString(R.string.instant_upload_path));
-        mPrefInstantVideoUploadPath.setSummary(mUploadVideoPath);
+        mUploadVideoPathAccount = appPrefs.getString("instant_video_upload_path_account",
```
przybylski: Could you please use constant for preference name

AndyScherzinger:
Instant Upload is brokoen on Android N, can some else test if this is working? pinging @przybylski @tobiasKaminsky @Spacefish :)

AndyScherzinger:
@tobiasKaminsky imho obsolete as soon as #358 is done but keeping it open for now. (Right?)

tobiasKaminsky:
Is this something still up to date as of the big rewrite of @mario?

mario:
I have to check what is this.

mario:
Looks outdated, can be deleted.

I think we need to respect account disabled for Auto Upload, but dont think
this is something that is respected across the app at all atm, and the rest
of this is useless since instant upload disappears.

AndyScherzinger:
This one is obsolete as soon as auto upload is active for all Android versions!
Thanks to @mario's efforts with auto upload 🚀

tobiasKaminsky:
> This one is obsolete as soon as auto upload is active for all Android versions!

I thought that auto Upload is only possible for >= 5 as of JobScheduler?

AndyScherzinger:
AFAIK it'll work on earlier versions too since @mario integrated Evernote job library which will fallback to an AlarmManager implementation to mimic JobScheduler so that should then be solved too. Right @mario?

mario:
It will use GCM network manager on 4 starting Nextcloud 1.5 :)

1.4.2 enables it only for 6 and 7.

tobiasKaminsky:
Thanks for explanation --> closing

Issue #109, opened 2016-07-07
Confirmation dialog before removing an account


AndyScherzinger:
Review from the Google Play store. User wished to have a confirmation dialog before an account is removed. 
If I remember correctly we had this discussion at some point and decided not to. But, yes delete is a tricky one. Any input on this @jancborchardt ?

TASK: Show an undo toast after removing an account.

jancborchardt:
As usual, we should offer undo here. :) It’s similar as a confirmation dialog, with the exception that confirmation dialogs are often mindlessly clicked away. Undo, where the action is directly visible (account gone) but actually only carried out in the interface, is much more error-friendly.

AndyScherzinger:
I also though about adding an undo action here. Will have to check how to implement such a thing because this might be a delete & recreate, or a hide from list and delete in case undo hasn't been clicked whatever is technically possible.

Thanks for the feedback! :)

jancborchardt:
> or a hide from list and delete in case undo hasn't been clicked whatever is technically possible.

This is how undo generally works – just lying by removing it from the interface, and then actually deleting it when no undo has been clicked or the next action is performed.

tobiasKaminsky:
In contributing.md is said that we should use label "accepted" for issues and 1...5 only for PRs.
I also suggest to write a summary in the first post if a conclusion is made.
(I will demonstrate it on this issue).

przybylski:
the undo is problematic on a lot of technical levels.
* If we cheat to delete account how long should we postpone until we really delete it? 
* If user closes app immediately after "deleting" account we won't be able to respond to that and account will exists.
* Removing may fail if we quietly ignore it, account will exist after "deletion".

If user won't read warning dialog than it isn't our fault.

jancborchardt:
> * If we cheat to delete account how long should we postpone until we really delete it?

What’s the default for that on Android, like for deleting pictures in Gallery? 5 seconds?

> * If user closes app immediately after "deleting" account we won't be able to respond to that and account will exists.

Isn’t there an exit-app hook?

> * Removing may fail if we quietly ignore it, account will exist after "deletion".

Sure, but that’s exactly the same with a confirmation dialog. Warning needs to be shown either way when the deletion fails. Undo does not mean ignoring an error, but actually the opposite: Only requiring people to take action when there is something wrong.

> * Toaster doesn't provide a way to get feedback when it is dismissed, but even if it did, app might be closed by then making dismiss notification null pointer vulnerable

[...]

PR #116, opened 2016-07-10
Add confirmation dialog on account deletion

przybylski:
https://github.com/nextcloud/android/issues/109 

also i did a small adjustment on UI since delete and password update button gave no visual feedback on press.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -105,6 +105,7 @@
     <string name="about_title">About</string>
     <string name="change_password">Change password</string>
     <string name="delete_account">Remove account</string>
+    <string name="delete_account_warning">Delete account %s? Deleting cannot be undone.</string>
     <string name="create_account">Create account</string>
     <string name="upload_chooser_title">Upload from &#8230;</string>
     <string name="uploader_info_dirname">Folder name</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -105,7 +105,7 @@
     <string name="about_title">About</string>
     <string name="change_password">Change password</string>
     <string name="delete_account">Remove account</string>
-    <string name="delete_account_warning">Delete account %s? Deleting cannot be undone.</string>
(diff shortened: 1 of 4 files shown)


AndyScherzinger:
:+1:

AndyScherzinger:
Will have to change the latest commit. I can add whatever icon we want to use. Using @android is bad because then the app loads the OS icons which depends on the OS version and we wouldn't want ICS icons for example.

What is the icon you where aiming for @przybylski?

przybylski:
recycle bin

AndyScherzinger:
DONE :) https://github.com/nextcloud/android/pull/116/commits/d9716c0fe10b84ccf57bb48de4d735ec46c123e4

AndyScherzinger:
Work fine on my phone. Can you also review @tobiasKaminski?

tobiasKaminsky:
@AndyScherzinger please use autocompletion as you type my last name very often wrong and thus I do not get a notification ;-)
Besides that please have a look at https://github.com/nextcloud/android/issues/109#issuecomment-232146608

AndyScherzinger:
Ups,sorry. The typos are from my phone,  so no auto complete :(

AndyScherzinger:
@tobiasKaminsky can you :+1: if this is fine with you, so the merge will be allowed by the plugin? :)

res/values/strings.xml
```diff
@@ -109,6 +109,7 @@
     <string name="about_title">About</string>
     <string name="change_password">Change password</string>
     <string name="delete_account">Remove account</string>
+    <string name="delete_account_warning">Delete account %s? Deleting cannot be undone.</string>
```
tobiasKaminsky: Very minor thing: I would do a linebreak for the second sentence.

tobiasKaminsky:
LGTM

res/values/strings.xml
AndyScherzinger: DONE: https://github.com/nextcloud/android/pull/116/commits/63bf982a15cc82dff3cf485c7220d15f828757e6

PR #124, opened 2016-07-13
Sort in upload files activity

AndyScherzinger:
yet another feature already implemented a while ago by @tobiasKaminsky as oc 1058

I rebased it to our latest master (not tests yet).

> Add the feature to sort in the uploader, in order to avoid a long scroll when selecting the most recent files for uploading.

Putting this up for discussion if we want to integrate this. It would imho help people especially for manual photo uploads.

cc @jancborchardt @LukasReschke and @przybylski


jancborchardt:
Yes, for sure. Actually in the uploader I would say the default layout should be sort by recency, cause you are most likely to upload recently modified files.

AndyScherzinger:
There actually isn't a default (now). The uploader is now able to change the sorting (like in the standard file list) and the sorting is shared between them, so the last sorting chosen by the user is respected by the file list, so if you change the sorting in the uploader your local file list will adopt that sorting (and vice versa).

jancborchardt:
Ok, I guess that’s the best course of action for now. :+1:

tobiasKaminsky:
LGTM

PR #125, opened 2016-07-13
Switch offline files

AndyScherzinger:
rebased and adapted PR originally done by @tobiasKaminsky to switch between all and offline files via the drawer menu.

please review @przybylski @tobiasKaminsky 

cc @LukasReschke and @jancborchardt for feature review in general even though this is an oldie PR :)


src/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -525,7 +522,7 @@ private void sendLocalBroadcast(
 
 
     private void fetchFavoritesToSyncFromLocalData() {
-        List<OCFile> children = mStorageManager.getFolderContent(mLocalFolder);
+        List<OCFile> children = mStorageManager.getFolderContent(mLocalFolder, true);
```
AndyScherzinger: @tobiasKaminsky wasn't sure what to put in here, decided to use true, but might also be false. You might know/recall :)

tobiasKaminsky: As far as I understand all files within the folder (childrens) should be fetched, therefore changing it to "false".

AndyScherzinger:
Mergable from my point of view 

LGTM

AndyScherzinger:
@przybylski @tobiasKaminsky what do you think? I tested it on my device and it works (for me)

src/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -208,19 +204,15 @@ public boolean onNavigationItemSelected(MenuItem menuItem) {
...
+                                refreshDirectory();
                                 break;
-                            // TODO activate when On Device branch is merged
-                            // case R.id.nav_on_device:
-                            //     menuItem.setChecked(true);
-                            //     mCheckedMenuItem = menuItem.getItemId();
-                            //     MainApp.showOnlyFilesOnDevice(true);
-                            //     refreshDirectory();
-                            //     break;
+                             case R.id.nav_on_device:
```
przybylski: show on device and show all items are mutually exclusive, so cannot be checked simultaneously. And yet, from code point of view it looks like they can. But I can be mistaken.

AndyScherzinger: Hi @przybylski they are mutually exclusive. Can you go a bit more into details why they could be checked simultaneously? The menu in general only allows one item to be in the checked state: https://github.com/nextcloud/android/blob/master/res/menu/drawer_menu.xml#L24

przybylski: This convinces me :) I haven't checked the xml file. Sorry

przybylski:
👍

src/com/owncloud/android/ui/activity/DrawerActivity.java
AndyScherzinger: No worries, I had to check the menu myself too :+1:

AndyScherzinger:
@tobiasKaminsky can you have a final look for the LGTM thingy :wink:

jancborchardt:
Nice, good stuff! :) We should push forward more with the offline stuff. The limited offline capability of the app was always a shortcoming – especially when you are on a train, plane or with bad mobile connection.

AndyScherzinger:
@jancborchardt that part is still the same (at the moment), this PR basically filters out all non-offline-available files, so you only see files already on your phone so you don't accidentally download a file.

jancborchardt:
Yup yup, I know – the sidebar entry. :) Good stuff

AndyScherzinger:
@tobiasKaminsky can you check to code and if fine with you, merge? :rocket:

jancborchardt:
(Btw, let’s keep the thumbs up emoji as an indicator for review. We use that since ages now. ;) And l g t m is not really understandable to many people cause it’s a really bad abbreviation.)

AndyScherzinger:
OK! :) but for the plugin you can go for LGTM, :+1: or :shipit: to approve (plugin enforced to MAINTAINERS responding with one of these) in order to be able to merge... :dancer:

jancborchardt:
Ah k, was just confused cause someone changed the +1 in my comment to LGTM (cause I didn’t write that)

AndyScherzinger:
Yeah, that was me...  :grin:

tobiasKaminsky:
LGTM

PR #128, opened 2016-07-14
corrected icon sizes for hdpi and mdpi

AndyScherzinger:
...like the title says, replaced all icons with wrong sizes, since on material design icons are 24dp sizes which wasn't 100% correct for some hdpi and mdpi icons, so they get resized at runtime. 

Besides that, removed all unused icons.

Please review and merge @LukasReschke @przybylski @jancborchardt @tobiasKaminsky 

PS: it is just icon work and sizes, so your choice to actually really look at the image diff ;)
Now all icons in all menus are shipped and no OS icon links are there anymore so it will always be nice, material style icon work :rocket:


jancborchardt:
Looks good I guess. :) If sizing them down is correct

AndyScherzinger:
@jancborchardt yes, the size should be 24dp according to material guidelines while we do you 24dp some icons haven't been 24dp but 36dp which leads to some "oversized" icons in cases where the OS doesn't resize on the fly and when it resizes the images get slightly blurred.

tobiasKaminsky:
I trust you :+1:

Issue #129, opened 2016-07-14
Provider authorities should be an absolute package name


AndyScherzinger:
As already filed in the owncloud issue tracker https://github.com/owncloud/android/issues/1735 forks will have collisions with authorities, e.g. ```UsersAndGroupsSearchProvider``` as well as the ```ManageSpaceActivity```.

The later has been fixed (for us) by using the full qualified name of the class but I think it would be best, if we move these two (and any other that may need to be moved) to a ```com.nextcloud.android``` package so both apps behave well and let a user run them in parallel - being good citizens.

cc for approval: @nextcloud/android

jancborchardt:
Sounds good, yeah.

przybylski:
ok to me

AndyScherzinger:
Sweet let's when we find the time to do this. We either can get it into 1.2.0 or should maybe do a 1.2.1 depending on how long it'll take us to implement 1.3.0

tobiasKaminsky:
Merged

PR #130, opened 2016-07-14
#129 Add authorities prefixes to providers

przybylski:


przybylski:
#129

tobiasKaminsky:
Have you tested it with an owncloud app in parallel?
I think we have to modify also the package name of the class, or?

przybylski:
I haven't but It shouldn't be a problem, BUT ill check.

tobiasKaminsky:
I tried it, but it did not work. I had to change authorities to
com.nextcloud.android.providers.UsersAndGroupsSearchProvider

AndyScherzinger:
I've tested it with @przybylski's latest commit and it is working. :+1:

AndyScherzinger:
@przybylski @tobiasKaminsky - one minor thing though which we can postpone for now. The authority changes are now using ```com.nextcloud.android``` which nicely matches the app's id except for the newly introduced document provider which is packaged in ```org.nextcloud```. Not sure if we want to change this, but using com.nextcloud.android as a parent package for new thing seems consistent naming wise.

AndyScherzinger:
@tobiasKaminsky can you also do a final test/approval ?

tobiasKaminsky:
@AndyScherzinger regarding com.nextcloud vs. org.nextcloud we should discuss in a new issue.

tobiasKaminsky:
I have tested it with both apps installed in parallel.
It works :+1:

PR #137, opened 2016-07-19
Log.wtf() should not be used unless an assertion fails

AndyScherzinger:
bugfix based on oC's 1603 with adaption for the new drawer implementation.

> This will prevent crashing when running on a userdebug phone. Log#wtf
should only be used when the program can not go on because a critical
assertion fails. These all don't seem to meet that criteria.

> Fixes 1603

please check and merge @nextcloud/android


AndyScherzinger:
Changes are MIT licensed btw.

Issue #111, opened 2016-07-07
Support for "files drop" feature


schiessle:
Nextcloud allows you to create public link shares where the users don't see the file listing but can only upload files. This is how it looks like in the web interface:

![filedrop](https://cloud.githubusercontent.com/assets/1589737/16657742/d103b50e-4463-11e6-8721-294b79818db0.png)

If a public link is created and the "Allow editing" permission was set a additional check box shows up, "Hide file listing". If activated it removes the "read" permissions, so it is a simple permissions change which can already be done by the OCS API. Of course this option only exists for folders.

tobiasKaminsky:
How is this "remove read permission" done:
https://doc.owncloud.org/server/8.2/developer_manual/core/ocs-share-api.html#create-a-new-share
1 = read; 2 = update; 4 = create; 8 = delete; 16 = share; 31 = all

are those values summed up like chmod (1, 2, 4)?
Or can I explicitly remove read?

AndyScherzinger:
Besides the necessary app change we might then also need a library change to support this.

tobiasKaminsky:
I do not think as we can currently edit the permission value mentioned above.
So it is only a question to which value I have to change it and not how I can do it ;)

tobiasKaminsky:
So this is working :)
I have just to do some minor UI things.

AndyScherzinger:
Happy to help with any kind of UI implementation necessary :)

AndyScherzinger:
@schiessle can you clearify https://github.com/nextcloud/android/issues/188 with @jancborchardt 
and/or @LukasReschke this matter blocks this issue since the implementation is done and tested but this open question remains.

schiessle:
@AndyScherzinger As others say at #188 I think it is OK to show such a warning

PR #141, opened 2016-07-21
Files drop

tobiasKaminsky:
according to #111 

TODO:

- [x] check for capabilities: https://github.com/nextcloud/server/pull/888 
- [x] show compatibility pop up #188

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -20,8 +20,8 @@
 -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.owncloud.android"
-    android:versionCode="10020099"
-    android:versionName="1.2.0">
+    android:versionCode="10030001"
+    android:versionName="1.3.0 RC1">
 
     <uses-sdk
         android:minSdkVersion="14"
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ dependencies {
     /// dependencies for app building
     compile name: 'touch-image-view'
 
-    compile 'com.github.nextcloud:android-library:1.0.6'
+    compile 'com.github.nextcloud:android-library:1.0.7'
     compile "com.android.support:support-v4:${supportLibraryVersion}"
     compile "com.android.support:design:${supportLibraryVersion}"
     compile 'com.jakewharton:disklrucache:2.0.2'
--- a/res/layout/share_file_layout.xml
+++ b/res/layout/share_file_layout.xml
@@ -191,6 +191,40 @@
                 </RelativeLayout>
 
                 <RelativeLayout
+                    android:id="@+id/shareViaLinkHideListPermissionSection"
+                    android:layout_width="match_parent"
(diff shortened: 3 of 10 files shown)


AndyScherzinger:
This is an incompatible feature, right?
So we need some kind of feature detection for this if we want the app to still work well with oC backends. (which of course we could decide not to, but for now I would try to stay compatible).

tobiasKaminsky:
This is nextcloud app for nextcloud server so it is fine for me ;-)
What is your opinion? @jancborchardt @przybylski

tobiasKaminsky:
IMPORTANT: Do we need a server version check for nextcloud? When was this feature introduced...? @schiessle

mar1u5:
It was introduced with 9.0.50 ...

tobiasKaminsky:
Ah, great :)
For me there is no need to have a version check against ownCloud.
If user should ever use the nextcloud app with an ownCloud server they should know that things can be broken.

mar1u5:
I agree with you @tobiasKaminsky

AndyScherzinger:
Found one (minor) issue:

1. choose folder to share
2. activate "share link"
3. activate "allow editing"
4. activate "hide file listing"
5. deactivate "share link"

Issue: all "sub options" get hidden except for "hide file listing" which is still visible and set to active. This is just a UI thing though. In the background the folder got properly un-shared!

Spacefish:
I would display a warning during the initial account adding inside the app,
which tells the user, that owncloud might be incompatible and that he may
want to upgrade his server to next cloud.. however we loose all the users
which may use the next cloud app but can't control the server installation
(coorporate or educational users mainly)

Am 02.08.2016 17:11 schrieb "Andy Scherzinger" <notifications@github.com>:

AndyScherzinger:
Issue https://github.com/nextcloud/android/pull/141#issuecomment-236935431 has been fixed

AndyScherzinger:
found it and fixed it in https://github.com/nextcloud/android/pull/141/commits/0faa786b2eae1df7307b79e9f34be5decc45a180

AndyScherzinger:
So tested and code reviewed: :+1:

AndyScherzinger:
@tobiasKaminsky do you think we can get a compatibility pop up done before the release so we can also ship this one? as in implement #188

tobiasKaminsky:
@AndyScherzinger ready for test. 
You will need https://github.com/nextcloud/android-library/pull/16

AndyScherzinger:
android lib has been merged, published via jitpack and referenced in this PR, see my latest commit.

Can test it right now since Android Studio went nuts and corrupts the installation on the phone atm... :(

AndyScherzinger:
Got it installed but never reached the check you implemented (not sure what is happening on my phone...). So my tests are negative at the moment :(

Will give it another try tomorrow.

AndyScherzinger:
Code review and tests successful, waiting for the build to complete.

PR #143, opened 2016-07-21
Pinch zoom

tobiasKaminsky:
pinch in/out to get more/less columns in grid view.

TODO
- [x] does not remember column numbers if apps closes
- [x] column number is resetted during rotation
- [ ] looks ugly right after zooming out until refresh


src/com/owncloud/android/ui/fragment/ExtendedListFragment.java
```diff
@@ -224,6 +274,10 @@ public void onSaveInstanceState(Bundle savedInstanceState) {
         savedInstanceState.putIntegerArrayList(KEY_TOPS, mTops);
         savedInstanceState.putInt(KEY_HEIGHT_CELL, mHeightCell);
         savedInstanceState.putString(KEY_EMPTY_LIST_MESSAGE, getEmptyViewText());
+
+        SharedPreferences.Editor editor = PreferenceManager
+                .getDefaultSharedPreferences(getContext()).edit();
+        editor.putFloat(GRID_COLUMNS, mScale).apply();
```
AndyScherzinger: should extract that to a helper function in our own/existing PreferenceManager implementation.

AndyScherzinger:
rebased to today's master.

Open issue: proper persistence/loading of grid column number. It gets lost during rotation and also during app restart.

AndyScherzinger:
I could fix two of the issues, while:
- [ ] looks ugly right after zooming out until refresh

remains open and I haven't found a way to fix this :( from what I read the ```invalidaViews()``` call should do the job)

tobiasKaminsky:
Updated.
@AndyScherzinger maybe you have time to have a look into this refreshing thing?

AndyScherzinger:
@tobiasKaminsky is this PR still valid? Looking at the fact that we plan to move on to recycler view I am unsure if it'll then need a slightly different implementation and also if that resolved the render issue still present? What do you think?

tobiasKaminsky:
I doubt that we will move to recycler view within the next months :/
But I do not not know how much changes needs to be done.
Storing column size, and such, should stay the same.
So I think, we can integrate it and then transfer it...

tobiasKaminsky:
The remaining refreshing "bug" is only happening when you swipe in/out. 
If you then scroll it will vanish.
As I tried it several times to fix it, I vote for adding this now.
When we switch to recycler view this bug will vanish.

Are you ok with this?

AndyScherzinger:
I am fine with this! Could you also fix the last remaining codacy issue? :)

AndyScherzinger:
👍 for merge

tobiasKaminsky:
@mario are you fine with merging this?

AndyScherzinger:
@tobiasKaminsky build fails...

```
/drone/src/github.com/*****/android/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.java:222: error: cannot find symbol
            if (getResources().getConfiguration().orientation == ORIENTATION_LANDSCAPE) {
                                                                 ^
  symbol:   variable ORIENTATION_LANDSCAPE
location: class ExtendedListFragment
```

AndyScherzinger:
👍 re-approved, awaiting @mario's approval/merge 🚀

mario:
@tobiasKaminsky @AndyScherzinger it says that one task still isn't done? 

" looks ugly right after zooming out until refresh"

AndyScherzinger:
> "looks ugly right after zooming out until refresh"

best way to fix this is using a recycler view and so Tobias and I think it is fine to leave it like this until we move to recycler view.

mario:
@AndyScherzinger ok. Please come to IRC too, I need you for a brief minute.

mario:
👍

AndyScherzinger:
Will be there in ~15 minutes

PR #144, opened 2016-07-21
Resize cache

tobiasKaminsky:


src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -107,7 +109,19 @@ protected Void doInBackground(File... params) {
...
+                        SharedPreferences appPrefs =
+                                PreferenceManager.getDefaultSharedPreferences(MainApp.getAppContext());
+                        // due to backward compatibility
+                        Integer cacheSize = CACHE_SIZE_MB * 1024 * 1024;
+                        try {
+                            cacheSize = appPrefs.getInt("pref_cache_size", cacheSize);
+                        } catch (ClassCastException e) {
+                            String temp = appPrefs.getString("pref_cache_size",
+                                    cacheSize.toString());
+                            cacheSize = Integer.decode(temp) * 1024 * 1024;
```
przybylski: too bad that decode can throw ;)

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -159,6 +173,31 @@ public static Bitmap getBitmapFromDiskCache(String key) {
...
+     * Sets max size of cache
+     * @param maxSize in MB
+     * @return
+     */
+    public static boolean setMaxSize(long maxSize){
+        if (mThumbnailCache != null){
+            mThumbnailCache.setMaxSize(maxSize * 1024 * 1024);
+            return true;
+        } else {
+            return false;
```
przybylski: why in else and not simply after `if` bracket?

src/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -159,6 +173,31 @@ public static Bitmap getBitmapFromDiskCache(String key) {
...
+        }
+    }
+
+    /**
+     * Shows max cache size
+     * @return max cache size in MB.
+     */
+    public static long getMaxSize(){
+        if (mThumbnailCache == null) {
```
przybylski: as a presonal preference I think you should first check for expected case `mThumbnailCache != null` and than go to fallback behavior (getting data from cache manager)

src/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -145,6 +146,27 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
             });
         }
 
+        final Preference pCacheSize = findPreference("pref_cache_size");
+        if (pCacheSize != null){
+            final SharedPreferences appPrefs =
+                    PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
+            Long cacheSize = ThumbnailsCacheManager.getMaxSize();
+            pCacheSize.setSummary(cacheSize + " Mb");
```
przybylski: Mb is suggesting that this is megaBIT is is right?

src/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -145,6 +146,27 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
...
+            pCacheSize.setOnPreferenceChangeListener(new OnPreferenceChangeListener() {
+                @Override
+                public boolean onPreferenceChange(Preference preference, Object newValue) {
+                    int size = Integer.decode((String) newValue);
+                    if (ThumbnailsCacheManager.setMaxSize(size)){
+                        appPrefs.edit().putInt("pref_cache_size", size);
```
przybylski: Without commit/apply this makes no sense

tobiasKaminsky:
Due to our recent discussion about a more intelligent mechanism for cache size I postpone this (and maybe it will never be merged)

tobiasKaminsky:
As discussed in another issue this should be done in a real smart way by the app itself and not bother an inexperienced user.

Issue #145, opened 2016-07-21
Upload -> no possibility to go back


tobiasKaminsky:
When I select "Uploads" in main menu, tapping on "All files" or "Uploads" doesn't do anything.
Other than that, switching between the two works ok.

tobiasKaminsky:
@AndyScherzinger this is yours ;)

AndyScherzinger:
I'll have a look. 

Big Thank you for reporting this!!! Sorry for any inconvenience caused!

AndyScherzinger:
Any info on device and Android version?

tobiasKaminsky:
I have reproduced it with an emulator and android 6.

michalmiddleton:
There issue was on Android 6.0.1 on Nexus 6p. Sorry for not being specific..

AndyScherzinger:
No worries @michalmiddleton, This is going to be fixed in the next release candidate :+1:

tobiasKaminsky:
This is now in RC2, available soon.

PR #146, opened 2016-07-21
#145 Let user go back from upload activity

przybylski:
I am not happy with the way of solving this, but I see it as a hotfix


tobiasKaminsky:
Let us wait if @AndyScherzinger has a better approach as he has written the new drawer part.

AndyScherzinger:
here is my approach #148

przybylski:
my approach sux, @AndyScherzinger's is much better, closing

PR #148, opened 2016-07-22
#145 Alternative fix for drawer navigation fix in upload activity

AndyScherzinger:
fix for #145 - alternative implementation approach to #146

pinging @przybylski and @tobiasKaminsky for feedback

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -20,8 +20,8 @@
 -->
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.owncloud.android"
-    android:versionCode="10020001"
-    android:versionName="1.2.0 RC1">
+    android:versionCode="10020002"
+    android:versionName="1.2.0 RC2">
 
     <uses-sdk
         android:minSdkVersion="14"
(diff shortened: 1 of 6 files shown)


przybylski:
better, one thing tho, showFiles(boolean) does set parameter in MainApp, maybe move that to super implementation (DrawerActivity) ?

AndyScherzinger:
@przybylski DONE, see https://github.com/nextcloud/android/pull/148/commits/58f86500ad8938e6dab73f18b0c581f679aa4fb5 - I moved it to FileActivity though, so the Drawer implementation can still declare the method abstract to force others to implement it and FileActivity is a super class of the using classes for Upload and Display so I found that to be a nice place to put it. :) Fine with you? :smiley: :+1:

przybylski:
👍

tobiasKaminsky:
:+1: 
Then we should release a RC2 within the next few days?

AndyScherzinger:
Yes we should. Can you guys take care of it?

tobiasKaminsky:
LGTM
The only remaining "problem" is that you do not get back to the folder you had opened before switching to "uploads".
But this is a separate issue.

PR #152, opened 2016-07-23
filter for oc file list and local file list

tobiasKaminsky:


jancborchardt:
Can you put up a screenshot? :)

tobiasKaminsky:
Here you go

filter1
filter

src/com/owncloud/android/ui/adapter/FilterableListAdapter.java
```diff
@@ -0,0 +1,27 @@
+/**
+ *   Nextcloud Android client application
+ *
+ *   Copyright (C) 2016 Tobias Kaminsky
+ *   Copyright (C) 2016 Nextcloud.
+ *
+ *   This program is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License version 2,
```
AndyScherzinger: Let's move to GPL3 for new stuff :)

tobiasKaminsky: Feel free :)
I have no idea how to do that

jancborchardt:
Nice! I assume when clicking the search icon, the input is automatically focused and the keyboard opens, right?

Also, instead of the x on the right it seems standard to change the menu icon on the left to a »back arrow« icon. Other apps like Signal, Youtube and Pocket Casts do that.

AndyScherzinger:
@jancborchardt both actually :smile: back arrow to close the search and x to clear the input (like you mentioned e.g. in the youtube app).

tobiasKaminsky:
@AndyScherzinger can you do the drawer part? Pleaaaaase :)

src/com/owncloud/android/ui/adapter/FilterableListAdapter.java
AndyScherzinger: Replace the 2 with a 3 :dancer:

AndyScherzinger:
Sure Thing! Can you give me a hint where the search I field get triggered to be shown and to be hidden?

jancborchardt:
@AndyScherzinger ah right, both back arrow and x as soon as something has been typed in, of course. :)

AndyScherzinger:
@jancborchardt close ;p x and arrow would be shown as soon as the search gets opened (widget behavior), which is like every where else. 

Open issue which I will have to see how quick I get a hold of it is theming optimization of the search widget. It is okay for now but there minor things I like to optimize :D

jancborchardt:
@AndyScherzinger if you check Youtube, the x only gets shown when something is put it. ;) Which makes total sense since the x is used for deleting content of the field. Isn’t needed when nothing is put in yet.

AndyScherzinger:
@jancborchardt true, I overlooked that...

AndyScherzinger:
@jancborchardt just a how to progress question. Would you be okay depending on how fast we can solve the "x" issue if we get close to a release of 1.3.0 that we postpone the "x" issue, release the filtering and open an issue for it? Wouldn't to it now since we will still try to solve this right away of course. It is just an idea to not block a feature completely.

jancborchardt:
@AndyScherzinger sure, that’s just a very small detail which shouldn’t block the feature. :)

src/com/owncloud/android/ui/adapter/FilterableListAdapter.java
AndyScherzinger: DONE!

AndyScherzinger:
@jancborchardt and @tobiasKaminsky I fixed the drawer part (arrow instead of hamburger) and also fixed the "x" issue :dancer: 

device-2016-07-28-144657
device-2016-07-28-144646

AndyScherzinger:
So since all the known changes have been done, ready for review I guess pinging @przybylski 

from my side: :+1:

AndyScherzinger:
@jancborchardt one issue to be discussed (non blocking) is the number of menu icons now visible in the upload view:
device-2016-07-28-145917
device-2016-07-28-150259

Not sure if we want to hide any... I would guess you want to hide them (all)... They are all quite new actions. You couldn't select all before 1.1.0, you couldn't sort in 3 different ways yet and 1.3.0 in a couple of weeks will add the filter functionality. So showing the icons helps promoting the functionality and I am not sure if hiding any action in the 3-dot menu is a general way to go (less cluttered Vs. less clicks).

jancborchardt:
Well, both select all and sorting are not really that important I would say. They could go into a 3-dot-menu. Together with a »Show hidden files« action possibly, see https://github.com/nextcloud/android/issues/165 :)

AndyScherzinger:
Okay :( I made such a nice icon transition for selectall/deseslectall :(

Some times you have to let go... :cry:

jancborchardt:
@AndyScherzinger yeah, sorry about that. ;) On the bright side, search is a kick-ass feature!

tobiasKaminsky:
@jancborchardt not "search", but only filter the current folder...
A true search is something I would like to do on conference, where all the /core guys can help me with the search api :)

jancborchardt:
@tobiasKaminsky yeah, that’s what I meant ;D

AndyScherzinger:
Merge? :smiley:

tobiasKaminsky:
As we do not have any other dev who can code review, I guess so.

PR #153, opened 2016-07-23
djvu is not an image

tobiasKaminsky:


src/com/owncloud/android/datamodel/OCFile.java
```diff
@@ -580,8 +580,14 @@ public boolean isVideo() {
...
-        return ((mMimeType != null && mMimeType.startsWith("image/")) ||
-                getMimeTypeFromName().startsWith("image/"));
+        String mimeType;
+        if (mMimeType != null) {
+            mimeType = mMimeType;
+        } else {
+            mimeType = getMimeTypeFromName();
+        }
+
+        return (mimeType.startsWith("image/") && !mimeType.contains("djvu"));
```
AndyScherzinger: This changes the logic from above. if mMimeType is null the "image/" check wouldn't have been done before. Without looking into the code, does getMimeTypeFromName work if mMimeType is null?

tobiasKaminsky: I do not unterstand what you mean :/
We have either mMimeType defined or we get it by guessing from name.
This is assigned to mimeType and checked if image and if it is not djvu.

tobiasKaminsky: Now I got you.
getMimeTypeFromName is only using mRemotePath, which is always set.

AndyScherzinger: Maybe I didn't mean anything. I was just wondering what getMimeTypeFromName does since I wouldn't have been called by the old code since Java would have stopped checking the statement if mMimeType is null but with the new code the get... Method gets called. So just wanted to make sure we wouldn't run into NPE or else.

tobiasKaminsky: (mMimeType != null && mMimeType.startsWith("image/")) || getMimeTypeFromName().startsWith("image/")

the second part after || will be called if mMimeType is null or if mMimeType is not null but it is not an image, or?

AndyScherzinger: Erm, not sure right now. But like you said the called are save NPE wise  :)

PR #166, opened 2016-07-28
Refresh when app starts

AndyScherzinger:
oC 1729

>### Actual behaviour
Sometimes when the application starts the root folder is not refreshed, and a manual refresh is required. This is verified, because execution will not enter the if(hasWindowFocus()) statement in startSyncFolderOperation method in FileDisplayActivity.java. The issue cannot be reproduced every time. Increasing the delay applied to the runnable, for example to DELAY_TO_REQUEST_OPERATIONS_LATER + 350 fixes the problem. 

>### Expected behaviour
The application should refresh the root folder when it is started.
 
>### Steps to reproduce
1. Compile the application
2. Start the owncloud application a good number of times, at least 10.

cc @tobiasKaminsky @przybylski

### DIFF
--- a/src/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/src/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -1,20 +1,23 @@
 /**
- *   ownCloud Android client application
+ *   Nextcloud Android client application
  *
  *   @author Andy Scherzinger
+ *   Copyright (C) 2016 Andy Scherzinger
+ *   Copyright (C) 2016 Nextcloud
  *   Copyright (C) 2016 ownCloud Inc.
  *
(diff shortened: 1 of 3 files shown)


AndyScherzinger:
cc @tobiasKaminsky

tobiasKaminsky:
Why not change the constant directly?

AndyScherzinger:
Will check if the constant is also used in other places. If so, we can just add another one.

tobiasKaminsky:
If neccessary a new constant is cleaner!

AndyScherzinger:
DONE, created a new constant, see https://github.com/nextcloud/android/pull/166/commits/25f4cdf68bd2a61ca6ad183f92a6d98dcd2883d2

tobiasKaminsky:
:+1:

PR #183, opened 2016-08-01
Streaming

tobiasKaminsky:


AndyScherzinger:
Hi @tobiasKaminsky, I rebased to latest master and adapted the code to my best knowledge. Can you have a look at the branch? The streaming isn't working for me. I do get the pop up (text warning the using that password might be exposed). I choose yes and then chrome pops up and asks my if I want to download the video. So some detection for streaming client might have been broken during my rebase (sorry if that is the case) or I do have a special issue simply already being on Android N Preview.

tobiasKaminsky:
I think this is normal as the "streaming external" is only providing a direct link to the video file.
The streaming is then handled by a video player app, e.g. vlc.
If no such video player is installed the default is chrome.

AndyScherzinger:
@tobiasKaminsky just installed vlc for android. Still chrome is launched and starts a download :(

tobiasKaminsky:
This seems to be a problem with only VLC.
Please try mx player.

For tracking purposes: https://trac.videolan.org/vlc/ticket/15741

I merged it into beta.

tobiasKaminsky:
Opening a https video stream link in firefox on android presents way more recipients than our system.
Maybe we can compare the intent of both to get vlc (and other) working.

AndyScherzinger:
Where is the firefox source available (Android app)?

tobiasKaminsky:
https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions/Simple_Firefox_for_Android_build#Get_the_source

AndyScherzinger:
VLC: still not working  (with latest release 2.0.6): https://trac.videolan.org/vlc/ticket/15741#comment:8
MX: works like a charm :)

aronovgj:
For me MX is not working properly. It says "Kann diesen Link nicht abspielen".

tobiasKaminsky:
We need a proper solution to not expose the real credentials to the public/other apps.
My idea currently is to create a one-time password for the file and pass this to the other apps.

AndyScherzinger:
So delete the branch I guess (for housekeeping) :)

fromage9747:
Just would like to confirm. This feature is still under development? I ask this as I experience the video file download instead of stream.

tobiasKaminsky:
Video streaming, with another approach, will be in next release 3.3.

johninstockie:
That's good to hear as this has been a problem for many of us.  Any idea when 3.3 is planned for?

tobiasKaminsky:
As RC2 looks good so far, Wednesday next week.

johninstockie:
Yes, this is what I am seeing too, I guess more to it

Issue #196, opened 2016-08-08
Shared link cannot be copied again


tobiasKaminsky:
### Actual behaviour
- no text/link to copy

### Expected behaviour
- Link should be copied
 
### Steps to reproduce
1. Create a shared link
2. Access it later via "get link"

### Environment data
Android version:

Device model: 

Stock or customized system:

Nextcloud app version:

Nextcloud server version:

#### Web server error log
```
Insert your webserver log here
```

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

tobiasKaminsky:
This fixes  oc 1562
It will be soon in beta. 
@dva-re it would be cool, if you can test it?

AndyScherzinger:
any feedback @dva-re

tobiasKaminsky:
Merged.
If any problems occur please reply

PR #197, opened 2016-08-08
search for the right ocShare with public link

tobiasKaminsky:
fix for #196


src/com/owncloud/android/ui/activity/ShareActivity.java
```diff
@@ -289,7 +289,18 @@ private void onCreateShareViaLinkOperationFinish(CreateShareViaLinkOperation ope
...
             // Create dialog to allow the user choose an app to send the link
             Intent intentToShareLink = new Intent(Intent.ACTION_SEND);
-            String link = ((OCShare) (result.getData().get(0))).getShareLink();
+
+            // if share to user and share via link multiple ocshares are returned,
+            // therefore filtering for public_link
+            String link = "";
+            for (Object object : result.getData()) {
+                OCShare shareLink = (OCShare) object;
+                if (shareLink.getShareType().name().equalsIgnoreCase("PUBLIC_LINK")) {
```
AndyScherzinger: Not sure if name() or getShareType() can be null but would then rather go for ```"PUBLIC_LINK".equalsIgnoreCase(shareLink.getShareType().name())``` and would also extract ```"PUBLIC_LINK"``` to a constant.

tobiasKaminsky: Done, thanks for the hint

PR #203, opened 2016-08-08
Remember last share

AndyScherzinger:
@tobiasKaminsky is this PR worth keeping it or should we delete the branch (throw it away). Just asking for housekeeping reasons and after checking the (Google) apps on my phone I couldn't find any app using this (anymore).


tobiasKaminsky:
Throw it away :)
As far as I remember it did not worked after all... ;)

Issue #195, opened 2016-08-04
Show quota if set on navigation drawer


tobiasKaminsky:
Idea per email, seen on cirrus

![screenshot_2016-08-03-13-32-52](https://cloud.githubusercontent.com/assets/5836855/17406966/deff16fe-5a65-11e6-8fe4-bee8d4e45f14.jpg)

AndyScherzinger:
Nein items thumbs up, but for that we should integrate the existing library pr first :)

tobiasKaminsky:
What library PR?
This will probably be in 1.4.0, so no hurry ;)

AndyScherzinger:
There is a nextcloud android library PR to retrieve quota info :)  so that one should imho be used. 
I would love to do the UI part and would hope you could do the server communication and caching part if that would be fine with you. :) 

1.4.0 sounds fine.

tobiasKaminsky:
As far as I have read the code it is just a quick "/ocs/v1.php/cloud/users/" + user + "?format=json" and parsing the returning values.
But a proper caching, e.g. query only once an hour, is a good idea.
I would keep the UI very close to the screenshot.

AndyScherzinger:
Makes sense. As for the caching, we should also refresh after uploads are finished because a user would want to have the actual value.

AndyScherzinger:
This is the library PR I was referring to: https://github.com/nextcloud/android-library/pull/10

AndyScherzinger:
Cc @skymania :)

tobiasKaminsky:
Then we should do it with the library :+1:

AndyScherzinger:
PR with initial UI work has been opened: #204

tobiasKaminsky:
PR merged

PR #169, opened 2016-07-28
Beta

tobiasKaminsky:
Dummy PR

PR #204, opened 2016-08-08
Display quota if configured/available in navigation drawer

AndyScherzinger:
for initial discussion please see #195 

Open issue:
* quota string formatting. For now I went with ```%1$s of %2$s used``` which could re-use the given formatter for (might need some work or a slightly different string) so %1$s and %2$s would be the sizes with the units calculated like the one in the file lists e.g. "1MB of 3,0GB used", another idea (like dropbox) would be to mix: "10% of 3,0GB used" - both have their advantages and disadvantages.

DONE
* coloring of the progress bar: "just blue" or turning yellow-ish and red-ish when hitting certain thresholds like 85% and 95%? And if so which yellow/red colors should be used.

As a sidenote, for now I had to add a "invisible" menu item for the bottom padding since I cannot customize the layouting of the menu-list itself which is imho okay until AppCompat provides more options to customize the drawer itself.

cc @tobiasKaminsky for further development and @jancborchardt regarding the quota string/format and colors to be used.

device-2016-08-08-172117

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ dependencies {
     /// dependencies for app building
     compile name: 'touch-image-view'
 
-    compile 'com.github.nextcloud:android-library:1.0.2'
+    compile 'com.github.nextcloud:android-library:1.0.4'
     compile "com.android.support:support-v4:${supportLibraryVersion}"
     compile "com.android.support:design:${supportLibraryVersion}"
     compile 'com.jakewharton:disklrucache:2.0.2'
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ dependencies {
     /// dependencies for app building
     compile name: 'touch-image-view'
 
-    compile 'com.github.nextcloud:android-library:1.0.4'
+    compile 'com.github.nextcloud:android-library:1.0.6'
     compile "com.android.support:support-v4:${supportLibraryVersion}"
     compile "com.android.support:design:${supportLibraryVersion}"
     compile 'com.jakewharton:disklrucache:2.0.2'
--- a/res/layout/drawer.xml
+++ b/res/layout/drawer.xml
@@ -35,7 +35,6 @@
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:layout_gravity="bottom"
-            android:background="#F5F5F5"
             android:clickable="false"
             android:orientation="vertical"
             android:paddingBottom="@dimen/standard_half_padding"
--- a/res/layout/drawer.xml
+++ b/res/layout/drawer.xml
@@ -35,6 +35,7 @@
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:layout_gravity="bottom"
+            android:background="@color/white"
             android:clickable="false"
             android:orientation="vertical"
             android:paddingBottom="@dimen/standard_half_padding"
--- a/res/layout/drawer.xml
+++ b/res/layout/drawer.xml
@@ -24,9 +24,43 @@
         android:layout_width="wrap_content"
         android:layout_height="match_parent"
         android:layout_gravity="start"
+        android:layout_weight="1"
         android:fitsSystemWindows="true"
-        app:theme="@style/NavigationView_ItemTextAppearance"
         app:headerLayout="@layout/drawer_header"
-        app:menu="@menu/drawer_menu"/>
+        app:menu="@menu/drawer_menu"
+        app:theme="@style/NavigationView_ItemTextAppearance">
+
+        <LinearLayout
+            android:id="@+id/drawer_quota"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_gravity="bottom"
+            android:background="#F5F5F5"
+            android:clickable="false"
+            android:orientation="vertical"
+            android:paddingBottom="@dimen/standard_half_padding"
+            android:paddingLeft="@dimen/standard_padding"
+            android:paddingRight="@dimen/standard_padding"
+            android:paddingTop="@dimen/standard_half_padding"
+            android:visibility="gone">
+
+            <ProgressBar
+                android:id="@+id/drawer_quota_ProgressBar"
+                style="?android:attr/progressBarStyleHorizontal"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:indeterminate="false"
+                android:indeterminateOnly="false"
+                android:text="@string/drawer_quota"
+                />
+
+            <TextView
+                android:id="@+id/drawer_quota_text"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:text="@string/drawer_quota"/>
+        </LinearLayout>
+
+    </android.support.design.widget.NavigationView>
 
 </merge>
--- a/res/menu/drawer_menu.xml
+++ b/res/menu/drawer_menu.xml
@@ -67,4 +67,15 @@
             android:icon="@drawable/ic_settings"
             android:title="@string/actionbar_settings"/>
     </group>
+
+    <!--
+      dummy group/element as a workaround to see
+      the whole menu in case of quota being displayed
+    -->
+    <group>
+        <item
+            android:enabled="false"
+            android:orderInCategory="200"
+            android:title=""/>
+    </group>
 </menu>
--- a/res/values/colors.xml
+++ b/res/values/colors.xml
@@ -52,4 +52,10 @@
     <!-- special transparent action bar colors for image preview -->
     <color name="owncloud_blue_transparent">#201D2D44</color>
     <color name="owncloud_blue_dark_transparent">#40162233</color>
+
+    <!-- level colors for info notifications/visualisations -->
+    <color name="infolevel_info">@color/color_accent</color>
+    <color name="infolevel_warning">#fdd835</color>
+    <color name="infolevel_critical">#e57373</color>
+
 </resources>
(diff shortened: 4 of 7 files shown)


AndyScherzinger:
@tobiasKaminsky please see https://github.com/nextcloud/android/pull/204/commits/c264f38c18e79be3c8984b2abf75dbcc2d226df9#diff-154af4ff2566e264dcca122072d20ae7R507 for an easy starting point, for now I have marked them private since I am not yet sure if everything could be handled within the DrawerActivity.

AndyScherzinger:
@tobiasKaminsky UI part is done, see below. Can you help me out with the caching part and also check the code if I implemented the server calls correctly (first time for me in this project)
device-2016-08-09-190834

Weird thing I realized is that as soon as I activated the 5 GB for the test account (tiger), my personal account even though it is set to unlimited shows a quota of 1 TB ?! Id this on purpose? Not sure who to ask to pinging @LukasReschke @nickvergessen @MorrisJobke and @rullzer ... Sorry guys for the noise.

tobiasKaminsky:
I'll do it tomorrow, wednesday.
Great work!

AndyScherzinger:
@jancborchardt is this fine with you color and threshold wise?

tobiasKaminsky:
+1 for color.
Great

AndyScherzinger:
Thanks @tobiasKaminsky :)  Did you already have a chance to to look into the caching aspect?

jancborchardt:
Awesome! Looks good. I’d only change 2 things:
- [ ] wording/values: use the same unit for both current and max value, likely MB or GB. Also do not use decimal values for MB or KB as it’s too little anyway.
- [ ] remove the darker grey background, it’s not needed.

AndyScherzinger:
@jancborchardt @nextcloud/designers Thanks for the Feedback, just a minor issue to clarify:

> wording/values: use the same unit for both current and max value, likely MB or GB. Also do not use decimal values for MB or KB as it’s too little anyway.

This is only for the quota right? For formatting I used the implementation that we use to display the individual file size for each file in the list views, which shows 1 decimal for MB, see:
https://github.com/nextcloud/android/blob/quotaDisplay/src/com/owncloud/android/utils/DisplayUtils.java#L69

So should we keep 1 decimal for MB for files but none for MB displaying quota or remove the MB decimal everywhere? (Bytes and KBytes don't have a decimal anywhere)

tobiasKaminsky:
@AndyScherzinger thinking about caching I think it is not necessary.
The overhead is so small in comparison to a refresh/listing of a directory and the risk is very high to show a deprecated value.
So I would leave it the way it is now. Hope you are okay with this?

tobiasKaminsky:
Ping me, when I should do a CR

AndyScherzinger:
@tobiasKaminsky you can start right away. Only remaining changes I'll do are:

* remove grey background color
* remove decimal for MB (probably a separate beautifier method like the one we have simply with another scales-Array :)

AndyScherzinger:
@jancborchardt @nextcloud/designers 

> wording/values: use the same unit for both current and max value, likely MB or GB. Also do not use decimal values for MB or KB as it’s too little anyway.

That doesn't make sense! If you used 2MB and have a Limit of 1GB you wouldn't want to display ```0,002GB of 1GB used``` Thus I went for no digits for B,KB,MB and 1 for GB and TB but strongly advice to drop that requirement of same unit. Only other smart thing I can think of is going for percentages lieke ```2% of 1GB used```

AndyScherzinger:
device-2016-08-14-201710

AndyScherzinger:
@nextcloud/designers here is a little problem imho with the white background for the quota...

device-2016-08-14-202221

Now the quota seems to be part of the menu (which it is not) and the user might not get that he can scroll (see right image)..

AndyScherzinger:
As for the "unlimited quota" detection I opened an issue for core: https://github.com/nextcloud/server/issues/879

AndyScherzinger:
CR ping to @tobiasKaminsky :)

tobiasKaminsky:
CR: :+1: 
Really awesome job!

AndyScherzinger:
Should we put it in 1.3.0 then or postpone this until the "unlimited quota" is solved?
Any opinions on this @tobiasKaminsky @jancborchardt --> Problem is that we can't detect unlimited quota and the server say 1TB quota, see nextcloud/server#879 even if there is no quota. Other than that we could (as a workaround) not display the quota information if the quota is >=1TB _and_ more less than 80% are in use. Else we have to postpone this until the server matter is resolved and that might mean Nc10 (or later) with the issue that we need to block this feature on earlier server versions :(

AndyScherzinger:
corresponding server PR: https://github.com/nextcloud/server/pull/890 - open issue is if it gets backported, which I am hoping for so we do only have to implement some handling of missing "unlimited flag" which would then only show up as described in the earlier comments for early Nc versions.

nickvergessen:
Patch was merged on the server and will be included in Nextcloud 10 RC1 and later

AndyScherzinger:
Thanks @nickvergessen! Plus is has also been merged to stable9 for 9.0.54 :rocket:

Bugsbane:
Personally I agree that we shouldn't be using different units of measurement as many users will have no idea how many Kb are in a Mb, or Mb in a Gb etc. The suggestion to use a percentage (eg. 2% of 1Gb) seems a good one.

The only possible issue I can see with using percentages is that when you have a large quota, the difference in free space between say 99% and 99.99% (which would also show as 99% used) could be significant to the user, especially if they need to upload a large file. Could we just show 2 decimal places for the percentage, only for where the actual percent used is >99% and <100%?

As to the quota usage hiding that users can scroll the sidebar, I guess I don't really see how it's a new issue? It seems to me that they might not have realized they could scroll even before the quota being visible, and that this doesn't really change that. Maybe I'm misunderstanding something here...

AndyScherzinger:
> As to the quota usage hiding that users can scroll the sidebar, I guess I don't really see how it's a new issue? It seems to me that they might not have realized they could scroll even before the quota being visible, and that this doesn't really change that. Maybe I'm misunderstanding something here...

Good point, yes the same issue is present already and thinking about it would also apply to any kind of list. My mistake...

> The suggestion to use a percentage (eg. 2% of 1Gb) seems a good one.

> Could we just show 2 decimal places for the percentage, only for where the actual percent used is >99% and <100%?

What do you think @jancborchardt @nextcloud/designers ? Changing the display of this kind of information into any direction is easy. We just need to decide which way to go :)

MorrisJobke:
> What do you think @jancborchardt @nextcloud/designers ? Changing the display of this kind of information into any direction is easy. We just need to decide which way to go :)

Maybe also show the space left then? Let's say under 500 MB we show something like:

"450 MB free (94% of 7,5 GB)"

For some people it's also hard to judge properly on percentages. On the other side it's also quite abstract to handle this in MB/KB, because often people even don't know how big the images on their phone are ... So maybe percentage is okay.

AndyScherzinger:
Personally I would be fine with just going with
* percent-used% of total-space unit

While as a rule we could "glue" it to the coloring rules
* INFO (blue): ```< 90``` - scale 0 --> e.g. ```89% of 7,5 GB```
* WARNING (yellow): ```>= 90``` and ```< 95``` - scale 0 -->```89% of 7,5 GB```
* CRITICAL (red): ```>= 95``` - scale 1 -->```96,2% of 7,5 GB```

I understand the issue with scale 2 for large total-space scenarios but would still say 99,x is good enough since you are already very close to your limit. So if you have 1TB as total space and 99,9% are in use you do have a problem already simply since you are a power user storage wise...

So how about my coloring / scale approach, it might not be the perfect solution (if there is any) but it is better than not having anything :smile:

jancborchardt:
We don’t need to add % because the bar already signifies that.

Regarding the white background I think it’s fine @AndyScherzinger. :) If it’s really a problem, maybe it’s possible to make the background slightly transparent so it’s clear there’s items below?

AndyScherzinger:
@jancborchardt the background shouldn't be a problem. The drawer and the scroll-able menu can be considered "standard behavior" since all Google apps and any other app implementing a drawer menu I can think of support scrolling, so I think the user is expecting this anyways :) 
As for the quota formatting, we can than just keep it like this and wait for any user feedback after the release.

I'll postpone this PR until 9.0.54 has been released. To be done then:

- [ ] update the library to also retrieve the quota on/off flag
- [ ] show/hide the quota UI elements for unlimited quota configuration

jancborchardt:
Sounds good yes :)

Bugsbane:
> We don’t need to add % because the bar already signifies that.

Good point.

AndyScherzinger:
@jancborchardt one issue to be decided. How should the app behave for Servers _not_ supporting the new Quota (NC < 9.0.54 and oC both not detecting unlimited quota) ?

* always show a quota, which for unlimited just means showing the total size
* never show a quota since "unlimited" cannot be detected

I am fine either way. I just need to know and since I am already running Nc10 this isn't a matter for my personal scenarios.

AndyScherzinger:
@jancborchardt PLEASE take a look at my last comment https://github.com/nextcloud/android/pull/204#issuecomment-243747037 - development wise this PR is implemented, tested and code reviewed so this one is ready to go for the next release!

AndyScherzinger:
added to beta branch #169

jancborchardt:
@AndyScherzinger 

> If unlimited can't be detected (this fact is propagated) than the total size will be used. If the quota is unlimited, then no quota will be shown.

Sounds good I guess!

- appropriate units are fine instead of forcing the same, you are right
- ah right, so only no decimals for kB, sorry. One decimal for MB and GB is fine.

:+1:  :)

AndyScherzinger:
And merged to master to be included in the next release :rocket:, also put to the beta branch with the latest changes. Thanks everyone! :clap: 

* @tobiasKaminsky for the code review
* @jancborchardt @Bugsbane @MorrisJobke for the UI/UX discussions
* @nickvergessen for implementing the quota changes in core for Nc10 and 9.0.54 - **lightspeed**

PR #69, opened 2016-06-25
Resized images

tobiasKaminsky:
What it does, copied from ownCloud and heavily discussed there, so here just a CR:
- Clicking on an image loads a reduced version to the cache.
- The width and height is determinated by the screen.
- the benefit is a very small, but yet usable preview of the image.
- it can be used to send via whatsapp, or as a file to another app (like k9) with the correct filename.
- resized images are not treated as downloaded files. If the user wants to have the real file, he has to long-click and select download.
- clicking on an image shows first a square thumbnail (if available) until the resized image is loaded

**ToDo**
- [x] test with disabled API (gallery app)
  - test if thumbnails api can be used -> not possible right now
- [x] Unsupported image formats https://github.com/owncloud/android/pull/1599#issuecomment-229400874
  - [x] svg
  - [x] tiff
- [x] Bad preview of landscape images https://github.com/owncloud/android/pull/1599#issuecomment-229406324 [WIP]
- [x] sharing a non-downloaded image on Android N fails (at least with whatsapp)
- [ ] sharing a non-cached resized image fails -> first download resized version
- [ ] a resized image should load immediately (first on queue or dismiss all other current downloads)
- [x] ~~download of a resized image also with a nice download animation on the bottom?~~
- [x] snackbar is sometimes beneath virtual keys
- [ ] ~~if resized version show "sending resized image" otherwise "send" (as it is  now)~~
- [x] test with disabled data
  - [x] show resized image if available
  - [x] show thumbnail if resized image not available
  - [x] show (something smart) if both is not available -> default image icon
- [x] on air plane mode: snackbar is shown all the time as resized download files -> prior check of connection

Assumptions:
- users do not want to download the whole image, instead just like streaming have only a resized version
- if an user wants to have many images in original quality, he can synchronize the folder

This leads us to
- always show a resized version
- if an user zooms in (> 2x) a snackbar "download full image: yes" will be shown (on every time)
- (to have the old behaviour: add setting "click on image downloads full image" (as checkbox))

Update: Talked to Jan @conf:
- behvaviour setting can be removed
- otherwise approved

PR #207, opened 2016-08-08
Store resized images on deletion

AndyScherzinger:
making @tobiasKaminsky's PR visible for code reviews and conflict tracking


tobiasKaminsky:
This is already part of #69, so closing

Issue #190, opened 2016-08-03
show/hide toolbar on scroll down/up


AndyScherzinger:
Since we now have the toolbar in place (which replaced the legacy action bar) we can start implementing further material design thing like show/hide of the toolbar as described here: http://android-developers.blogspot.de/2014/10/material-design-on-android-checklist.html

> App bar
> Where appropriate, upon scrolling down, the app bar can scroll off the screen, leaving more vertical space for content. Upon scrolling back up, the app bar should be shown again.

If this general behavior is approved I would start working on an implementation on the main list which shows the files on server/on device as a first implementation. If that is done we can then think about adopting this also on other screens like file picker for upload I probably wouldn't implemented it on all screens but may on all list or just most lists...not sure though.

any feedback/comments? @tobiasKaminsky @jancborchardt @przybylski ?

You can see this behavior in Chrome on Android for example.

tobiasKaminsky:
This is a really cool feature :+1: 
But let's wait for @jancborchardt ;)

jancborchardt:
Yeah, this is also in place in Firefox and the Play Store. :) Let’s test out if it works nicely for us as well.

joshtrichards:
Duplicate of #6238 
Fixed by #6262

PR #212, opened 2016-08-14
show/hide toolbar on scroll

AndyScherzinger:
initial work / prototyping of #190


tobiasKaminsky:
Is this still valid if we (you) switch to recyclerview?

AndyScherzinger:
Yes :) 
I will work on this as soon as I moved to recyclerview which will make this PR more reliable with the event handling :)

tobiasKaminsky:
Shall we try to integrate this now, as recycler view will take quite some time?

AndyScherzinger:
No, this one needs the recyclerview to work properly. to not spam the notifications on GH all the time I could close the PR for now but would keep the branch. What do you think?

tobiasKaminsky:
I do not mind the notifications, but think it results in a lot of duplicate work as you need to rebase it from time to time...

AndyScherzinger:
Or I don't and just keep it as a code reference for later when I can actually implement it :)

AndyScherzinger:
removed since the scroll implementation would be different for a recycler view and has been added as a separate PR for the auto upload already, thus closing

Issue #210, opened 2016-08-12
Use log in background image for background in top of sidebar


jancborchardt:
In apps like Play Store and Pocket Casts you can see a very nice background image in the header of the sidebar (where you have the account switcher or app name). We should use the clouds background for that:
https://github.com/nextcloud/server/blob/master/core/img/background.jpg

PR to be found here: #213

AndyScherzinger:
Sounds good, I also thought about this. Will do :)

AndyScherzinger:
release with 1.3.0

Issue #54, opened 2016-06-22
Server based theming capabilities


fictitiousexistence:
Will the nextcloud app eventually be able to visually look like the mobile web browser version? This way it can be themed just like our own installs are. Or Will its always have it's own design and look?

AndyScherzinger:
It will imho always have the same design and look which aligns with the Nextcloud theme in general but might differ here and there due to platform specific guidelines. Any theme changes on the server won't be reflected in the client as in the Android app, or a desktop client since it is a complete independent implementation and it is also a native one so no mobile web views.

Sorry for not having any better news here, cc @jancborchardt just in case.

mar1u5:
Would be a nice idea taking colors etc. from the new theming app from the instance which is currently active in the app 🎨 https://github.com/nextcloud/server/pull/59 😉 

cc @nextcloud/designers

AndyScherzinger:
@mar1u5 you rather need to include @nextcloud/android since this need some research if and how to do this besides the fact that is is way harder in the client think  multi user / server account switching scenario...

I am sure it is possible I am just not sure it is worth the effort right now since you need to implement the theming, you need an API to retrieve theming information, need to store and cache and need to think of a balanced way to refresh your caches, and you will have to recalculate colors since the server theme won't know anything about them, e.g. 700-shaded variation of the primary color is used for the systembar...

c0fe:
I honestly would like to see this implemented, at least on the web based portion where if you are sharing public documents/files you can just have your own theme installed whichever you prefer. This would be fairly unique.

AndyScherzinger:
@c0fe I think the web based part is already working on this, right @jancborchardt ?

jancborchardt:
@LukasReschke @schiessle @karlitschek how about an API for logo, color, name, slogan etc so we can push that out to the clients? That would be pretty cool if the app automatically changes to the values provided by the server.

(Would also be a _lot_ of work for the case where the web interface is themed though. ;)

mar1u5:
@jancborchardt or we rewrite the **ownBrander** (which will be done in parts with the theming app) from the EE to theme the clients 😉

LukasReschke:
> @jancborchardt or we rewrite the ownBrander (which will be done in parts with the theming app) from the EE to theme the clients 😉

Oh god. You don't want to write THAT. The theming API stuff seems way more easy and reasonable 🙈

AndyScherzinger:
[...]

PR #213, opened 2016-08-14
background image for background in top of sidebar

AndyScherzinger:
added the background image to the drawer header as requested in #210 

device-2016-08-14-223504


AndyScherzinger:
please review @tobiasKaminsky @nextcloud/designers

mar1u5:
How complicated would it be to fetch the picture from the theming app 😁

AndyScherzinger:
@mar1u5 if the server side offers an interface for it, it is technically possible. As discussed in #54 we can't support launcher icons and coloring would be a problem here and there but for the background image here, it would be possible, yes.  I which version of Nc server would this feature/api be available?

Bugsbane:
This is a good step forward, but as @mar1u5 suggested, it really should use the theming app image to be consistent.

mar1u5:
I think @juliushaertl is the right person to give Information about the APIs :grin:

juliushaertl:
@mar1u5 @AndyScherzinger By now there is no public api for getting theming values, but that shouldn't be much work. I guess such an API would absolutely make sense, as other external apps might use theming values as well.

tobiasKaminsky:
This PR is :+1: 
For setting the background image via public theming API please open a  new issue and link it to the corresponding server issue.

MorrisJobke:
You rock! This is amazing :) I like it 👍

jancborchardt:
Looks slick as hell! :tada:  Now we only need to add it to the connection screen as well @AndyScherzinger ;) Wanna play around with that as well?

AndyScherzinger:
@jancborchardt Connection screen as in the login screen? Do we really want to do this? I do like the simplicity of the uni-colored background and not having the image. How would we design it? simply putting in the image isn't enough since all the input fields are "just lines" (as in lovely native!)

AndyScherzinger:
@jancborchardt please see the screenshots, as before strongly advising against it...

- Yes - this can be done
- Yes - it is ugly as hell
- No - we shouldn't do this
- No - this is not a webapp on a 24" screen

device-2016-08-17-164111

device-2016-08-17-164132

@nextcloud/android please say NO - you can do it

enoch85:
I'll say YES actually. Anyone else?

Maybe just make the background a bit darker??

AndyScherzinger:
@enoch85 Did you open the image on an actual phone or just on a PC?

enoch85:
@AndyScherzinger Just looked at the pics here in this PR.

AndyScherzinger:
Please look at it on an Android phone. Imho this really far away from a material design experience.

Bugsbane:
To me it seems that if the admin has chosen to theme their install by replacing the Nextcloud mountains image, that we shouldn't still be pushing that image on them anywhere.

If we're concerned about readability issues though, like what happens if they choose an image that's almost all white, then possibly we could opt to go with the chosen theme colour instead of the theme image? We've already talked elsewhere about limiting the theme colour to only the darker colours so we don't make the web header unreadable (although personally I'd prefer that we just automatically flip text colour from white to black if the chosen background is too light).

jancborchardt:
@AndyScherzinger yeah, good point. It just looks very off in this case, and distracts from simply logging in. Let’s not do it. :)

AndyScherzinger:
> If we're concerned about readability issues though, like what happens if they choose an image that's almost all white, then possibly we could opt to go with the chosen theme background colour instead of the theme image?

That is the point (or one point)... we can't because that would mean (while technically possible) we would have to do a color analysis of the background image being used and we will run into large problems with no solution yet which is the next demand that will come up which is the logo... And the transparency issue is yet unresolved, like with PNG avatars, we simply calculate a background color (white), so no transparency here!

> @AndyScherzinger yeah, good point. It just looks very off in this case, and distracts from simply logging in. Let’s not do it. :)

:heart: :heart: :heart: :heart: :heart: :heart:

tobiasKaminsky:
We should discuss this in a new issue ;)

Bugsbane:
> we can't because that would mean (while technically possible) we would have to do a color analysis of the background image being used

I wasn't suggesting Nextcloud automatically uses background colour where the image is nearly all white. I was suggesting using the users chosen background colour, period, and thus avoiding problems *such as* when their image is almost all white.

Using their chosen background *colour* instead of image, ensures that we're using their theme, not ours, while still keeping things readable (if we only allow darker colours to be chosen, anyway)

AndyScherzinger:
As mentioned by tobias, let's discuss this in a new issue while being aware that the theming capabilities are targeting the web interface _not_ native mobile interfaces (!). that being that the theming on Nc server doesn't have a input field for a system bar coloring present on Lollipop and later, plus you cannot change the color of the folder icons nor the check boxes nor the radio buttons at the moment, at least not consistently. This will be possible at some point (as soon as Google get vector support right and backwards compatible).

So...
1. let's discuss this in a new issue
2. let's postpone theming until this can be done easily and decide if we want to do such things (also keeping in mind day/night mode introduces in the Android 6 or 7...)

Issue #202, opened 2016-08-08
NPE in CheckAvailableSpaceTask.onPostExecute


AndyScherzinger:
Reported on the Play store developer console (unfortunately no comments)
 
### Environment data
Android version: 6

Device model: Nexus 5 (hammerhead) 

Stock or customized system: - 

Nextcloud app version: 1.2.0

Nextcloud server version: - 

#### Nextcloud log (data/nextcloud.log)
```
java.lang.NullPointerException: Attempt to invoke virtual method 'void android.support.v4.app.DialogFragment.dismiss()' on a null object reference
	at com.owncloud.android.ui.activity.UploadFilesActivity$CheckAvailableSpaceTask.onPostExecute(UploadFilesActivity.java:440)
	at com.owncloud.android.ui.activity.UploadFilesActivity$CheckAvailableSpaceTask.onPostExecute(UploadFilesActivity.java:400)
	at android.os.AsyncTask.finish(AsyncTask.java:651)
	at android.os.AsyncTask.-wrap1(AsyncTask.java)
	at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:668)
	at android.os.Handler.dispatchMessage(Handler.java:102)
	at android.os.Looper.loop(Looper.java:148)
	at android.app.ActivityThread.main(ActivityThread.java:5417)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
```

PR #215, opened 2016-08-15
NPE in CheckAvailableSpaceTask.onPostExecute (and unnecessary auto-boxing)

AndyScherzinger:
NPE stack trace, see #202 

cc @tobiasKaminsky


tobiasKaminsky:
:+1:

PR #236, opened 2016-09-01
DAVdroid login screen prefilled launch

AndyScherzinger:
Ability to launch DAVdroid via the drawer menu including a prefilling of server URL and username, password could also be transfer but unencrypted, so for security reasons the user has to type in the password which is imho the right trade-off.

Please see attached a screenshots to comment on for @jancborchardt / cc @jospoortvliet 
device-2016-09-01-185248

Open issues:

- [x] show a popup with f-droid and play to be routed to the stored to install DAVdroid if not installed (atm I just show a toast message)
- [x] check if there is a nicer way to retrieve the account base URI then to create an oC/Nc-client (which has to be done asynchronous). Any idea @tobiasKaminsky ?


childnode:
what is the exact use once it is set up? Is this really an option a normal user wants to keep in the "main menu"? For me, it's more or less a "setting"

This also might be more interesting right upon you setup the nextCloud app.
(get asked if you also want to create / setup calendar / contact sync)
This should be optional and not be asked for, if there is already a DavDroid account set up for this server + user. I'm not that in what is possible with the [AccountManager](https://developer.android.com/reference/android/accounts/AccountManager.html) and what isn't but I thought you can also reuse another set-up account. So, the other way round, if someone has already setup DavDroid and wants to setup nextCloud, nextCloud might just ask for permission to use this for simplified / automatic nextCloud setup?!

AndyScherzinger:
@childnode Not sure what would be possible with the AccountManager (as in shared data across applications) either but I would also agree that the drawer menu seems to be to prominent especially looking at the fact that with DAVdroid's actual capability we can only launch/prefill the Login screen. So I would also vote for putting it on the settings screen, if we keep this direction (Nc to DAVDroid). 
As for the scenario you described I am not able to tell which scenario is rather the case, having the Nc app setup first or the DAVdroid app.

If the AccountManager allows access to the Nc accounts that seems to be best (if technically possible) but would mean that DAVdroid would have to do the implementation.

Another idea which would need to be researched would be inter-app communication as in DAVdroid doing up/download up calendar and contacts via the Nextcloud app.

Still as a first step to offer at least something / some help I think we can work with this but put it into the settings screen.

any comments @jancborchardt @jospoortvliet - just implemented what has been asked for.

childnode:
> So I would also vote for putting it on the settings screen, if we keep this direction (Nc to DAVDroid).
As for the scenario you described I am not able to tell which scenario is rather the case, having the Nc app setup first or the DAVdroid app.

Thinking about the "nextCloud product suite" it is a valid case to me to see Nc first being installed aside the server. Then the user wants to know how to setup the rest like contacts and calendar. 

> Still as a first step to offer at least something / some help I think we can work with this but put it into the settings screen.

totally agree 👍 

For the "davDroid first" scenario: this would be a nice feature to make setting up but using secure passwords as easy as possible, nevermind the most users will be experienced users then.

But: let's start and see what other users say if they taste blood✌️nice idea anyway

AndyScherzinger:
@nextcloud/designers and @jospoortvliet can we go on as discussed in this issue by putting it into the settings screen?

AndyScherzinger:
Update, item has been put into the setting's more section.
device-2016-09-07-175315

@tobiasKaminsky please review and also let's discuss: I would like to include this in the 1.3.0 RC2 (if you are fine with this) plus remaining open issue 
>check if there is a nicer way to retrieve the account base URI then to create an oC/Nc-client (which has to be done asynchronous). Any idea @tobiasKaminsky ?.

I tested the following scenarios:
* no app store app: browser launched with f-droid DAVdroid page
* only Play store installed: play store launches
* only f-droid installed: f-droid launched
* both installed: Android asks you to choose between Play store and f-droid
* app installed: DAVdroid login view is launched with server address and user name pre-filled, after finishing the setup you automatically return to Nextcloud with a Toast shown "successfully setup" in case this is the setups result propagated by DAVdroid.

Tested on Nexus5X device running Android 7.0 and emulator 6.0

tobiasKaminsky:
Code wise it is :+1:

AndyScherzinger:
Great! Thanks for the code review @tobiasKaminsky 
I'll see if I find the time to merge it this weekend and do a final test on it.

AndyScherzinger:
Latest Build ran successfully after master rebase and LGTM has been reset to 1

AndyScherzinger:
:+1: plugin has hiccups .... again :rage:

PR #298, opened 2016-09-29
Add open URL file feature

zmatsuo:
Add open URL file feature.
*.url file is Internet short cut for windows.

### DIFF
--- a/src/com/owncloud/android/files/FileOperationsHelper.java
+++ b/src/com/owncloud/android/files/FileOperationsHelper.java
@@ -110,7 +110,7 @@ public class FileOperationsHelper {
                 }
             }
 
-            if (br != null) {
+            if (fr != null) {
                 try {
                     fr.close();
                 } catch (IOException e) {
--- a/src/com/owncloud/android/files/FileOperationsHelper.java
+++ b/src/com/owncloud/android/files/FileOperationsHelper.java
@@ -30,6 +30,7 @@ import android.content.pm.PackageManager;
 import android.content.pm.ResolveInfo;
 import android.net.Uri;
 import android.os.Parcelable;
+import android.support.annotation.Nullable;
 import android.support.v4.app.DialogFragment;
 import android.webkit.MimeTypeMap;
 import android.widget.Toast;
@@ -50,8 +51,13 @@ import com.owncloud.android.ui.activity.FileActivity;
 import com.owncloud.android.ui.activity.ShareActivity;
 import com.owncloud.android.ui.dialog.ShareLinkToDialog;
 
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  *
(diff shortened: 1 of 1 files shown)


src/com/owncloud/android/files/FileOperationsHelper.java
```diff
@@ -71,42 +77,112 @@ public FileOperationsHelper(FileActivity fileActivity) {
...
     }
 
+    /**
+     * Windows internet shortcut file .url
+	 * Ubuntu internet shortcut file .desktop
+	 */
+    @Nullable
+    private String getUrlFromUrlFile(String file) {
+        String url = null;
+        Pattern p = Pattern.compile("^URL=(.+)$");
```
AndyScherzinger: The pattern should not be compiled within the method since pattern compilation is a time/cpu consuming and thus for performance reasons it is better to only do this once either by making this a member variable or even a static one.

src/com/owncloud/android/files/FileOperationsHelper.java
```diff
@@ -71,42 +77,112 @@ public FileOperationsHelper(FileActivity fileActivity) {
...
+                Matcher m = p.matcher(line);
+                if (m.find()) {
+                    url = m.group(1);
+                    break;
+                }
+            }
+            br.close();
+            fr.close();
+        } catch (IOException ex) {
+            return null;
```
AndyScherzinger: please add a Log statement here, so in case this issue occurs users would be able to supply us with a log fiel where we could see this issue has occurred. :)

src/com/owncloud/android/files/FileOperationsHelper.java
```diff
@@ -71,42 +77,112 @@ public FileOperationsHelper(FileActivity fileActivity) {
...
+        return url;
+    }
+
+    /**
+     * mac internet shortcut file .webloc
+	 */
+    @Nullable
+    private String getUrlFromWeblocFile(String file) {
+        String url = null;
+        Pattern p = Pattern.compile("<string>(.+)</string>");
```
AndyScherzinger: Same here. The pattern should not be compiled within the method since pattern compilation is a time/cpu consuming and thus for performance reasons it is better to only do this once either by making this a member variable or even a static one.

src/com/owncloud/android/files/FileOperationsHelper.java
```diff
@@ -71,42 +77,112 @@ public FileOperationsHelper(FileActivity fileActivity) {
...
+
+        try {
```
AndyScherzinger: please extract two get* methods for the different URL file types into 1 method since except for the used pattern the code is identical. So there should only be one method which has the pattern as a input parameter :)

zmatsuo:
Hi @AndyScherzinger 
Thank you for your review.
The modification has been committed.

src/com/owncloud/android/files/FileOperationsHelper.java
```diff
@@ -67,46 +73,90 @@
...
+
+            String line;
+            while ((line = br.readLine()) != null) {
+                Matcher m = pattern.matcher(line);
+                if (m.find()) {
+                    url = m.group(1);
+                    break;
+                }
+            }
+            br.close();
```
AndyScherzinger: One small thing here I found after looking at the changes again.
```br``` and ```fr``` should be closed in a finally {} block with null checks so it should probably look somethign like this (quite some some checks to be bullet proof):

```
... 
        FileReader fr = null;
        BufferedReader br = null;

        try {
            fr = new FileReader(storagePath);
            br = new BufferedReader(fr);

            String line;
            while ((line = br.readLine()) != null) {
                Matcher m = pattern.matcher(line);
                if (m.find()) {
                    url = m.group(1);
                    break;
                }
            }
        } catch (IOException e) {
			Log_OC.d(TAG, e.getMessage());
            return null;
        } finally {
            if (br != null) {
                try {
                    br.close();
                } catch (IOException e) {
                    Log_OC.d(TAG, "Error closing buffered reader for URL file", e);
                }
            }

            if (br != null) {
                try {
                    fr.close();
                } catch (IOException e) {
                    Log_OC.d(TAG, "Error closing file reader for URL file", e);
                }
            }
        }
...
```

AndyScherzinger:
@zmatsuo I found another minor issue, sorry for being late with this one, but I also posted a code snippet how I think it should be resolved. Thank you for your patience :+1: You can also provide us with write permission on a PR when you create it (not sure if it can be done afterwards, probably not), so next time I can do code review changes myself if it is to much of a hassle :smiley:

AndyScherzinger:
@tobiasKaminsky please also review. I successfully tested the branch (for windows-based .URL files).

zmatsuo:
@AndyScherzinger Thank you.
I understand repository/Setting/Collaborators and close() in finally block.
The modification has been pushed.

tobiasKaminsky:
LGTM
@AndyScherzinger as you are blocking the review, please merge it.

AndyScherzinger:
Code looks good :+1: just fixed one null check.

Thank you very much for your contribution @zmatsuo ! :tada: - your feature will be part of the next stable release 1.4.0 :)

AndyScherzinger:
@zmatsuo btw you are now a Nextcloud member and can open branches in the Nextcloud Android repo :grin:

zmatsuo:
I have wrote PR #327. I will open branch in the Nextcloud Android repo nexttime.
Thank you.

Issue #185, opened 2016-08-01
Upload view: switch to Grid view


tobiasKaminsky:
When choosing a file from within the app it should be possible to switch to grid view.

AndyScherzinger:
Good call, especially for the camera folder a quite useful feature! :+1: 
Should put that in the three dot overflow menu imho.

tobiasKaminsky:
And the view should be saved and restored as the next view, just like everywhere else.

AndyScherzinger:
fine with me, yes :+1:

AndyScherzinger:
Dev hint: The grid view implementation/switching can be found here for the general file listing and is already in place (for this view, not active for upload views):

https://github.com/nextcloud/android/blob/master/src/com/owncloud/android/ui/activity/FileDisplayActivity.java#L597

PR #301, opened 2016-10-01
add grid view option when choosing a file from within the app

apicellaj:
Implemented a grid view option for choosing a file from the app as mentioned in #185

### DIFF
--- a/src/com/owncloud/android/ui/activity/UploadFilesActivity.java
+++ b/src/com/owncloud/android/ui/activity/UploadFilesActivity.java
@@ -151,7 +151,7 @@ public class UploadFilesActivity extends FileActivity implements
         actionBar.setDisplayShowTitleEnabled(false);
         actionBar.setNavigationMode(ActionBar.NAVIGATION_MODE_LIST);
         actionBar.setListNavigationCallbacks(mDirectories, this);
-        
+
         // wait dialog
         if (mCurrentDialog != null) {
             mCurrentDialog.dismiss();
@@ -182,6 +182,8 @@ public class UploadFilesActivity extends FileActivity implements
         getMenuInflater().inflate(R.menu.upload_files_picker, menu);
         MenuItem selectAll = menu.findItem(R.id.action_select_all);
         setSelectAllMenuItem(selectAll, mSelectAll);
+        MenuItem switchView = menu.findItem(R.id.action_switch_view);
+        switchView.setTitle(isGridView() ? R.string.action_switch_list_view : R.string.action_switch_grid_view);
         return super.onCreateOptionsMenu(menu);
     }
 
@@ -229,13 +231,11 @@ public class UploadFilesActivity extends FileActivity implements
             case R.id.action_switch_view: {
                 if (isGridView()) {
                     item.setTitle(getString(R.string.action_switch_grid_view));
-                    item.setIcon(ContextCompat.getDrawable(getApplicationContext(),
-                            R.drawable.ic_view_module));
+                    item.setIcon(R.drawable.ic_view_module);
                     mFileListFragment.switchToListView();
                 } else {
                     item.setTitle(getApplicationContext().getString(R.string.action_switch_list_view));
-                    item.setIcon(ContextCompat.getDrawable(getApplicationContext(),
-                            R.drawable.ic_view_list));
+                    item.setIcon(R.drawable.ic_view_list);
                     mFileListFragment.switchToGridView();
                 }
                 return true;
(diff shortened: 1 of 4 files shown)


AndyScherzinger:
Hi @apicellaj 

thank you very much for your PR and welcome to Nextcloud. I did a small test and found two (minor) issues (for code review please see code comments):

* grid/list state information gets lost during screen rotation: switch to grid view and rotate device, the menu still say "grid view" instead of "list view"
* The grid view doesn't use the right layout file and grid width, see screenshot 1, while it should use the one the "regular" file listing view would use, see screenshot 2

Screenshot 1
device-2016-10-04-230925

Screenshot 2
device-2016-10-04-231417

src/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -223,6 +226,20 @@ public void onClick(DialogInterface dialog, int which) {
...
                 break;
             }
+            case R.id.action_switch_view: {
+                if (isGridView()) {
+                    item.setTitle(getString(R.string.action_switch_grid_view));
+                    item.setIcon(ContextCompat.getDrawable(getApplicationContext(),
+                            R.drawable.ic_view_module));
+                    mFileListFragment.switchToListView();
+                } else {
+                    item.setTitle(getApplicationContext().getString(R.string.action_switch_list_view));
```
AndyScherzinger: ```getApplicationContext()``` can be removed

apicellaj:
Hi @AndyScherzinger 

Thanks for the feedback and the warm welcome. The two issues seem to be resolved but let me know if you would like anything changed.

AndyScherzinger:
@apicellaj thanks for the code changes! :+1: 

@tobiasKaminsky please have a look, now the grid view for local files would look like this:
device-2016-10-05-103545

--> The local files grid view does distinguish between grid-item and grid-image (image/video) which it can't at the moment since we would have to dynamically detect the mime type. That is possible but we do not have a implementation for that. So I would say it is out of scope and rather open another issue for this matter and declare this one feature complete. What do you think?

AndyScherzinger:
Slight update, with some customization of the code:
```
@Override
    public View getView(int position, View convertView, ViewGroup parent) {
        View view = convertView;
        File file = null;
        boolean isGridView = true;
        LayoutInflater inflator = (LayoutInflater) mContext
                .getSystemService(Context.LAYOUT_INFLATER_SERVICE);

        if (mFiles != null && mFiles.length > position && mFiles[position] != null) {
            file = mFiles[position];
        }

        // Find out which layout should be displayed
        ViewType viewType;
        if (parent instanceof GridView) {
            String mimeType = MimetypeIconUtil.getBestMimeTypeByFilename(file.getName());
            if (file != null && (MimetypeIconUtil.isImage(mimeType) || MimetypeIconUtil.isVideo(mimeType))) {
                viewType = ViewType.GRID_IMAGE;
            } else {
                viewType = ViewType.GRID_ITEM;
            }
        } else {
            viewType = ViewType.LIST_ITEM;
            isGridView = false;
        }

        // create view only if differs, otherwise reuse
        if (convertView == null || convertView.getTag() != viewType) {
            switch (viewType) {
                case GRID_IMAGE:
                    view = inflator.inflate(R.layout.grid_image, parent, false);
                    view.setTag(ViewType.GRID_IMAGE);
                    break;
                case GRID_ITEM:
                    view = inflator.inflate(R.layout.grid_item, parent, false);
                    view.setTag(ViewType.GRID_ITEM);
                    break;
                case LIST_ITEM:
                    view = inflator.inflate(R.layout.list_item, parent, false);
                    view.setTag(ViewType.LIST_ITEM);
                    break;
            }
        }

        if (file != null) {
            if(!ViewType.GRID_IMAGE.equals(viewType)) {
                TextView fileName = (TextView) view.findViewById(R.id.Filename);
                String name = file.getName();
                fileName.setText(name);
            }
            ....
```

leads to this:
device-2016-10-05-111101

@tobiasKaminsky I would apply these changes _after_ we merged this PR since I did some refactorings in the background for MimeType utils and ViewType etc. etc.

AndyScherzinger:
@tobiasKaminsky what do you think? I'd say merge + another PR later on. :+1:

tobiasKaminsky:
@apicellaj thank you very much for your contribution!

AndyScherzinger:
Thank you very much for your contribution @apicellaj :tada: It will be included in the next stable-feature release :rocket:

apicellaj:
Great! Thanks again for your feedback

Issue #107, opened 2016-07-07
Sort favorite files first in the list


jancborchardt:
Favoriting is not very integrated at the moment. Only an indicator, we can do better here.

Better integration without needing to switch to a separate view would be to always sort the favorited items of a the current folder up top.

AndyScherzinger:
Dev-hint: This should be done for FileDisplayActivity.java so the sorting then needs to incorporate the chosen way of sorting (alphabetical, date, size) as a second order criteria and favorite (yes/no) as a first order criteria.

ryanwr:
I'll work on this

AndyScherzinger:
implemented, reviewed and tested, part of the next stable release 1.4.0

jancborchardt:
@ryanwr @AndyScherzinger awesome! :) Now we should get this in the web interface as well, cc @nextcloud/designers for https://github.com/nextcloud/server/issues/1802

PR #302, opened 2016-10-02
Sorts favourite files first in list

ryanwr:
Shows files marked as favourites first, closes issue #107

### DIFF
--- a/src/com/owncloud/android/utils/FileStorageUtils.java
+++ b/src/com/owncloud/android/utils/FileStorageUtils.java
@@ -247,6 +247,8 @@ public class FileStorageUtils {
             files = FileStorageUtils.sortOCFilesBySize(files);
             break;
         }
+
+        files = FileStorageUtils.sortOCFilesByFavourite(files);
        
         return files;
     }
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@tobiasKaminsky can you (also) have a look? Code looks good to me. To be discussed if (as implemented in this PR by @ryanwr) there should be 2 chained sortings or if the sorting for the favorites should be incorporated in the 3 existing sorting implementations. That way you don't have to do a second sorting which makes the sorting implementations more complex but would save us 1 complete/additinal loop especially when it comes to very large file listings.

tobiasKaminsky:
I doubt that there is a real performance loss as there are very very few favorites files I guess.
So I would vote for keeping it this way.
Or is someone eager doing some performance tests? ;-)

AndyScherzinger:
Then I am fine doing it like this :+1: 
I'll will test it again (probably tomorrow) for all 3 sorting-options and if it works correctly will then merge it.

AndyScherzinger:
Test were successful - so merging to master.

Thanks @ryanwr for your contribution! :tada: This will be part of the next stable release 1.4.0

jancborchardt:
@ryanwr very cool contribution! :) Added you to the Nextcloud organization and Android team :)

ryanwr:
Thanks! Happy to help

PR #304, opened 2016-10-03
Fix Polish translations

agilob:
4 translations in "Uploads" view made no sense and were completely out of context.


tobiasKaminsky:
Thanks for your contribution and welcome!
The translation is done via transifex, please translate it there.

AndyScherzinger:
@agilob I just fixed the strings in transifex based on your pull request (so the translations will be fixed in the next release) :) Nevertheless, if you have the time and interest in registering at https://www.transifex.com and help us out with the Polish translations that would be awesome since we are having some (15) string which aren't translated for Polish language at the moment. So this would really help us having it 100% translated for Polish :rocket:

agilob:
Thanks @AndyScherzinger

AndyScherzinger:
You are very welcome! :smiley: 

here you can find the remaining open strings for the Android app https://www.transifex.com/nextcloud/nextcloud/translate/#pl/android/97057962?translated=no

PR #307, opened 2016-10-04
To Check: Move files

AndyScherzinger:
Is this branch still relevant @tobiasKaminsky ?


tobiasKaminsky:
> When sharing a file with owncloud it is now possible to choose between copy and move, just like in the upload from within our app.
But in despite of this the default is always copy (as e.g. no one wants to move whatsapp images to owncloud as this might break the app).
Of course it can happen that a file is moved during sharing that cannot be moved (e.g. a cached image), then it will just copy it and fails silently.

Quoting myself from oC. The question is to we want to have this?

tobiasKaminsky:
If yes, we should enhance this and check if a source file is move-able and only show the option if possible.

tobiasKaminsky:
@AndyScherzinger @mario what is your opinion here?
I vote for closing as sharing always means "copying". If someone really wants to share a bunch of files to nextcloud, then he can always delete them afterwards.

mario:
Well, then close ... :D

AndyScherzinger:
I also vote close and delete the branch :)

PR #321, opened 2016-10-09
Incorporated Checkstyle, findbugs, pmd and lint

djubreel:
Incorporated checkstyle, findbugs, pmd and lint for code quality. Most rules are not set yet as that would be needing a lot of changes in the code base so the rules can be added bit by bit and corresponding changes effected in codebase.  @AndyScherzinger

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,9 @@ buildscript {
 }
 
 apply plugin: 'com.android.application'
+apply plugin: 'checkstyle'
+apply plugin: 'pmd'
+apply plugin: 'findbugs'
 
 ext {
     supportLibraryVersion = '23.4.0'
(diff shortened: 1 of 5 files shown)


AndyScherzinger:
Awesome work @djubreel :+1: 

@tobiasKaminsky please also have a look and if you are okay with it please merge. I did a dry run locally (gradlew.bat check) and I also set ```ignoreFailures = true``` in build.gradle so the build runs through and all checks get executed and reported (plus the build doesn't fail that way). Imho this is ready to merge. Then we should start new issues to resolve the reported issue (mainly findbugs and pmd so the build doesn't fail) and then activate this in our CI builds.

tobiasKaminsky:
Code is :+1: 
Can we merge it? Or do we have to change "ignoreFailures" to true before merging?

tobiasKaminsky:
One minor thing:
- This should be mentioned in some readme/documentation

AndyScherzinger:
I would say let's wait with the "documentation" until the fails are fixes, then activate it in the builds and put it into the documentation.

We can merge right away since the checks will fail if you execute them, but they aren't activated in our builds just yet.

AndyScherzinger:
@djubreel The PR has been merged! :tada: Thank you very much for this contribution!!! :rocket:

Issue #165, opened 2016-07-28
Hide hidden folders & files in uploads view


jancborchardt:
See https://github.com/nextcloud/android/pull/152#issuecomment-235889677

We should hide those by default. Maybe we can add a 3-dot menu with a checkbox entry »Show hidden files« to show them as we do in the web interface.

AndyScherzinger:
Some questions do get a better understanding of this... details ;)
* What characterizes a hidden file? 
* Everything starting with a dot? 
* Does this also apply for hidden folders (in case we have something like that) too?
* What is meant by uploads view: view to choose what to upload or view to choose where to upload to? (unfortunately both are called upload views at the moment, plus we have the upload view showing you the files that have been uploaded (in the drawer menu)...)

jancborchardt:
Starting with dots, yes. What’s the logic on the server @Morris @schiessle?
And yes, both folders and files. And any view I would say. Even when browsing your own files. Then the »Show hidden files« could be in the 3-dot-menu in any view.

jancborchardt:
(Off by default, hidden files should be hidden of course.)

AndyScherzinger:
@jancborchardt so to be consistent with the server or rather with the web UI should we than just also put this into the settings instead of the menu on the different views?

schiessle:
@AndyScherzinger just to avoid confusions. It is not in the global setting menu (admin or personal settings) but in the files app setting at the bottom left:

![menu](https://cloud.githubusercontent.com/assets/1589737/17218747/34b29f1c-54e9-11e6-954d-6e062615a4a7.png)

I think it is important that the setting is at a place where people can access it easily to toggle the state without leaving the current folder.

AndyScherzinger:
I'd say in the settings screen. Putting it in the menu seems like a too prominent place for an action you probably only use once, never or every couple of weeks/ months. @jancborchardt ?

pybizzle:
I'm not sure if this is expected behavior or not but the hidden files show up if you preform a search from any location via the web interface. I would expect most people to not want this to happen, is there a way to prevent hidden files/folders from being listed in a search?  I understand that the file will only be visible within your own searches unless it was shared for some reason, however for sensitive files it would be safer for them not to be listed in any search.

AndyScherzinger:
changes have been merged and will be released with 1.4.0

PR #328, opened 2016-10-13
add preference to show hidden files

apicellaj:
Add setting to allow showing hidden files as mentioned in issue #165

### DIFF
--- a/src/com/owncloud/android/ui/activity/Preferences.java
+++ b/src/com/owncloud/android/ui/activity/Preferences.java
@@ -86,6 +86,7 @@ public class Preferences extends PreferenceActivity {
     private Uri mUri;
 
     private CheckBoxPreference pCode;
+    private CheckBoxPreference mShowHiddenFiles;
     private Preference pAboutApp;
     private AppCompatDelegate mDelegate;
 
@@ -163,6 +164,19 @@ public class Preferences extends PreferenceActivity {
                 }
             });
         }
+
+        mShowHiddenFiles = (CheckBoxPreference) findPreference("show_hidden_files");
+        mShowHiddenFiles.setOnPreferenceClickListener(new OnPreferenceClickListener() {
+            @Override
+            public boolean onPreferenceClick(Preference preference) {
+                SharedPreferences appPrefs =
+                        PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
+                SharedPreferences.Editor editor = appPrefs.edit();
+                editor.putBoolean("show_hidden_files_pref", mShowHiddenFiles.isChecked());
+                editor.commit();
+                return true;
+            }
+        });
 
         PreferenceCategory preferenceCategory = (PreferenceCategory) findPreference("more");
 
--- a/src/com/owncloud/android/ui/adapter/LocalFileListAdapter.java
+++ b/src/com/owncloud/android/ui/adapter/LocalFileListAdapter.java
@@ -346,6 +346,6 @@ public class LocalFileListAdapter extends BaseAdapter implements FilterableListA
                 ret.add(file);
             }
         }
-        return ret.toArray(new File[0]);
+        return ret.toArray(new File[ret.size()]);
     }
 }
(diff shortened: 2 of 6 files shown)


AndyScherzinger:
@apicellaj thanks (again) for yet another sweet PR! :tada: This is awesome

@tobiasKaminsky can you give it a try since I done have a dev environment until next Sunday/Monday? Also can you check if "hidden" is the activated default? I already had a quick look at the code.

@jancborchardt @schiessle at the moment the on/off switch is implemented as a general setting in the preference screen which I would prefer in contrast to the discussion we had in the original issue since putting it in the file action menu seem to be to prominent to me. What do you think? @tobiasKaminsky?

jancborchardt:
Nice @apicellaj! :) Added you to the Nextcloud organization.

@AndyScherzinger good point. The main focus of https://github.com/nextcloud/android/issues/165 was that in the uploads view the hidden files need to be hidden. Putting that setting in the settings section with the rest of advanced stuff like Passcode lock sounds better than so present in the top bar. :)

AndyScherzinger:
@jancborchardt thanks for the feedback! :) - btw. the hidden files will also be hidden for the standard/server listing view in the actual PR's implementation which is absolutely fine imho (and what we also want).

Won't be able to test drive the PR, so looking forward to @tobiasKaminsky's feedback. @apicellaj @tobiasKaminsky without having had a closer look at the code to answer one follow up question: When I use search (which at the moment is basically a filter - ) the hidden files shall be shown. So this will need to be tested.

src/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -164,6 +165,19 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
...
 
+        mShowHiddenFiles = (CheckBoxPreference) findPreference("show_hidden_files");
+        mShowHiddenFiles.setOnPreferenceClickListener(new OnPreferenceClickListener() {
+            @Override
+            public boolean onPreferenceClick(Preference preference) {
+                SharedPreferences appPrefs =
+                        PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
+                SharedPreferences.Editor editor = appPrefs.edit();
+                editor.putBoolean("show_hidden_files_pref", mShowHiddenFiles.isChecked());
+                editor.commit();
```
apicellaj: Thanks for the feedback! I used `commit()` because `saveInstantUploadPathOnPreferences` and `saveInstantUploadVideoPathOnPreferences` used `commit()` too. Should all three be changed to `apply()`?

src/com/owncloud/android/ui/adapter/LocalFileListAdapter.java
```diff
@@ -325,4 +332,20 @@ public void filter(String text){
...
+     * @return                  Non-hidden files as an array
+     */
+    public File[] filterHiddenFiles(File[] files) {
+        List<File> ret = new ArrayList<>();
+        for (File file: files) {
+            if (!file.isHidden()) {
+                ret.add(file);
+            }
+        }
+        return ret.toArray(new File[0]);
```
apicellaj: I thought I read somewhere that `ret.toArray(new File[0]);` used to be slightly better for the JVM since `list.size()` caused unnecessary overhead. I'm not sure if that's still the case since the [java docs](https://docs.oracle.com/javase/7/docs/api/java/util/Collection.html#toArray()) don't seem to indicate any overhead so I can change that to `ret.toArray(new File[ret.size()]);`.

src/com/owncloud/android/ui/activity/Preferences.java
AndyScherzinger: @tobiasKaminsky what's is your opinion? I'd say move on to ```apply()``` even though this isn't something originally introduced by this PR.

AndyScherzinger:
* Test drove the changes :+1: Works like a charm
* Reviewed the code changes: :+1:

AndyScherzinger:
:tada: @apicellaj Thanks for this sweet PR! :tada: 
It's been merged and we go into the next release! :rocket:

Issue #297, opened 2016-09-28
Fix logging in beta app


AndyScherzinger:
All recent issue reports when asking for the beta app's log reply with the fact that the log is empty which seems to be an issue and need to be solved in order to gain the necessary information to solve the actual issues that got reported.

At least installed via Android Studio (thus with debug flag) my logged are filled, same if I install it from f-droid.

Any idea @tobiasKaminsky ?

---

TODO:
improve logging on standard beta client installations to find
- [ ]  "UNKNOWN_ERROR" https://github.com/nextcloud/android/issues/288#issuecomment-251504857
- [ ] Unexpected exception https://github.com/nextcloud/android/issues/287#issuecomment-250549099

tobiasKaminsky:
We need to collect the device and android version and see if there is something in common.

radegand:
Hi,
I've reported issue #288 and I have Nexus 5X on Android 7.0, build number NRD90R. I can provide more logs if someone points me to what's the best way to get them.
Thanks.

AndyScherzinger:
added https://github.com/nextcloud/android/pull/169/commits/59355d38d5de3babcff3369ebe7d15be7dfef719 to gain the result code for issue #288

AndyScherzinger:
added https://github.com/nextcloud/android-library/commit/c1ebf063043b076caba0cd51f31e3f3fa2fc2151 to log the exception type and message for issue #287 (ven though log suggest this is related to chunking)

iskradelta:
Hello friends,

in MainApp the check is if (BuildConfig.DEBUG) but DEBUG is not set to true when checking the built with baksmali - but we have BUILD_TYPE set to DEBUG - which seems reasonable - so the fix is a simple "DEBUG".equalsIgnoreCase(BuildConfig.BUILD_TYPE) which seems simpler than to figure out why gradle doesnt set the variable DEBUG or somehow force it to do it when build type is debug... Im new to this.

tobiasKaminsky:
@AndyScherzinger 
@iskradelta found the error and we added it to beta.
Please test and close it.

AndyScherzinger:
@tobiasKaminsky this might still be an issue in some way, see #979 :/

hechi:
Hey still no logs for me :(
Nextcloud Android App: 20170509
![screenshot_20170512-164745](https://cloud.githubusercontent.com/assets/3808354/26003481/12aedc50-3733-11e7-9385-5eeacfe3c99f.png)

[...]

PR #335, opened 2016-10-14
Begin debug logs when build type is debug

iskradelta:
Previous check was not true hence no logs
being made ever. #297


tobiasKaminsky:
This also does not work as the build process uses the build type RELEASE.
For simplicity I suggest to enable this on beta at any time and just remove the check.

It would be great if you can change it this way.

iskradelta:
I see, nice catch! Pushed a new commit now, without the if statement, this is only on the beta branch, I guess the master branch should have the "if build type is debug"?

tobiasKaminsky:
:+1:

tobiasKaminsky:
Will be in the next beta.

Issue #336, opened 2016-10-15
android app crashes when trying to sort a directory of photos to upload


koreywithak:
### Actual behaviour

When trying to sort a folder of photos on my android device while in the upload screen of the nextcloud app, the app crashes.
### Expected behaviour

It should sort the folder without crashing
### Steps to reproduce
1. Click upload in app
2. Find the DCIM folder on phone to get to camera photos
3. Click sort in the app
### Environment data

Android version: 6.0.1

Device model: Nexus 6

Stock or customized system: stock

Nextcloud app version: 1.3.1

Nextcloud server version: 10.0.1

Don't have time to upload logs right now. Hopefully I can add later.
#### Web server error log

```
Insert your webserver log here
```
#### Nextcloud log (data/nextcloud.log)

```
Insert your Nextcloud log here
```

AndyScherzinger:
Hi @koreywithak thank you for the report. No logs needed since I can reproduce this crash.

koreywithak:
Okay great. Glad to help.

AndyScherzinger:
corresponding PR #346

AndyScherzinger:
Hi @koreywithak, the fix has been merged and will be shipped with the next release! :rocket:

koreywithak:
Awesome. Pretty fast and organized resolution. Thanks!

AndyScherzinger:
Thanks! I consider 27 days until merge quite long but that is due to the upcoming feature release and Hacktoberfest.

PR #346, opened 2016-10-17
Use correct array of options

AndyScherzinger:
Resolves: #336

> android app crashes when trying to sort a directory of photos to upload

@tobiasKaminsky please review - This is a work in progress since imho the implementation is broken for a simply reason. OCFIleList and LocalFileList both use the same preference for the sorting choice. So if you choose BySize in the main screen and then open upload via the FAB it'll load "by Size" and will try to sort this way. Unfortunately the local file sorting by size also sorts folders, by recursively calculating their size which takes around one minute on my phone... So imho we either need to 
- remove folders from the local file sorting as in sort by size with folders on top sorted alphabetically or
- split the sorting choice into two preferences and don't offer sorting by size at all.

What do you think @tobiasKaminsky - I vote for ignore folders during size sorting (locally)

### DIFF
--- a/res/values/attrs.xml
+++ b/res/values/attrs.xml
@@ -32,4 +32,9 @@
 		<item>@string/menu_item_sort_by_date</item>
 		<item>@string/menu_item_sort_by_size</item>
 	</string-array>
+
+	<string-array name="menu_items_sort_by_options_local" translatable="false">
+		<item>@string/menu_item_sort_by_name</item>
+		<item>@string/menu_item_sort_by_date</item>
+	</string-array>
 </resources>
--- a/res/values/attrs.xml
+++ b/res/values/attrs.xml
@@ -32,9 +32,4 @@
 		<item>@string/menu_item_sort_by_date</item>
 		<item>@string/menu_item_sort_by_size</item>
 	</string-array>
-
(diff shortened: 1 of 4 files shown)


tobiasKaminsky:
What about a async folder computation which updates the list one after one?

AndyScherzinger:
That would be an option but seems very expensive as is battery consumption to be able to sort by size including folders and their subfolders. So for local files we could still go for folders first alphabetically and files by size. 

What do you think @jancborchardt? Local complete would be nice and consistent but there is a high price to pay for imho.

AndyScherzinger:
To speed up things for now and fix the bug (crash!), I'd say lets not support sort by size for the local view and fall back to alphabetical sorting in that case.

Agreed?

jancborchardt:
Sorting method by size is not that crazily important anyway – so if it causes an issue it’s fine to not have it.

AndyScherzinger:
@tobiasKaminsky ready to review :smiley: Sort by size is now also supported in the manual upload view, while folders are sorted alphabetically at the top. Tested on my Nexus5X

AndyScherzinger:
pushed to beta branch

tobiasKaminsky:
:+1:

Issue #285, opened 2016-09-21
[TOP] Enhance sync folders


tobiasKaminsky:
As discussed on conf this is a collaboration between @AndyScherzinger and me.
Purpose: 

* [x] Increment 1 - instant upload - #358 - release 1.4.0 :heavy_check_mark: 
  - replace current instant upload
  - full support for Android 6,7

* [x] Increment 2 - enhancement auto upload - #797 - release 2.0.0 :heavy_check_mark: 
  - support video folders
  - allow any kind of folder pair (custom folders)
  - show jobs in upload pane
    - track jobs
    - get job result to update upload pane
    - if job fails: allow user to restart
  - automatic retry
  - upload updated files

* [x] Increment 3 - upload existing files :heavy_check_mark: 
  - add "upload/sync all existing images/videos/files" as an option to the auto upload folder three-dot menu.

* [x] Increment 4 - folder upload :heavy_check_mark: 
  - allow folder hierarchies
  - smart UI to semi-automatically detect and notify about potential new media folders the user can add/active for instant upload #2406

* [x] Increment 5 - sync one way
  - sync - one way
  - free up space

* [ ] Increment 6 - sync two way
  - sync - two way
  - ref https://github.com/nextcloud/android/issues/19

aparcar:
Does this include a constant sync of a `todo.txt` file on change locally and periodically checked on server?

tobiasKaminsky:
@aparcar currently this will only replace instant upload.
But this approach is designed to also handle a real two way sync and a constant sync of files.

AndyScherzinger:
This issue/implementation is to be considered the top prio since we now already have > 1,300 Android 7 installations according to the play store :exclamation:

tobiasKaminsky:
Camera should be top of the list as this is the most wanted "folder", I guess.
At least I was searching a while where it is ;-)

tobiasKaminsky:
Uhh, and caching is not working properly, right?

tobiasKaminsky:
I know it is kind of reduntant, but I think that it should be possible to enable/disable sync also in the setting of a folder (and not only in the folder sync view).
[folder sync looks fantastic! :+1: :tada: ]

AndyScherzinger:
Yeah, we can put camera on top, need to think how to put this into the comparator. Doable :)
As for the caching, don't know actually. The queries need to be tweaked also since they need to order by date.

AndyScherzinger:
[...]

PR #358, opened 2016-10-26
Auto Upload (Android 7+)

AndyScherzinger:
New instant upload implementation for any device running Android 5 or later - called **Auto upload**

This new implementation supports "V1" of the specified feature set, which focuses on 
- bringing the instant upload feature to Android 7+
- supporting Android Doze
- supporting instant upload for any kind of media folder (at the moment this means Images/Videos)

for any future increments, please see #285

_This is a collaboration between @AndyScherzinger and @tobiasKaminsky coding-wise._

@nextcloud/designers please have a look at the actual screenshots (below), hereby requesting feedback/discussion. In short we mimic Google Photos to some extend.

display media folder in a sectioned grid (--> headers). Each folder shows its name, up to 8 thumbnails, remaining images of the folder for >8 files in the respective folder, on/off toggle icon, three dot for custom preferences (per folder, beware though that we have a defaulting so you don't need to configure anything).
device-2016-10-26-144206
_The sorting is weighted and alphabetical, meaning they are sorted alphabetically (while a folder named "Camera" wins...) and weighted by active/inactive. Thus active one will be at the top and the Camera folder will always be at the top of the active or the inactive cluster.
The thumbnails shown per folder are always the latest ones (as in date taken, modification-date) and they get cached_

Preferences are adopted from the actual/known instant upload capabilities as is choose the upload folder, wifi only, charging only, created sub-folders, upload behavior, enable/disable.
device-2016-10-26-144227

The new instant upload is so cool it got it's own main menu item :yum: 
device-2016-10-26-144449
_Also due to the fact that this feature has at least for evolutionary increments to be possibly implemented see #285 and will transform into a general folder sync feature in the future_

---

**Tasks Tobi**
- [x] Use DB for hardcoded DCIM folder to implement FolderObserver, JobScheduler
- [x] Use DB to start/stop/resume all DB entries for FolderObserver
- [x] Adapt UploadPane to new system
- [x] Fallback system for android versions that do to support JobScheduler (< 5)
- [x] "null bug" see https://github.com/nextcloud/android/pull/358#issuecomment-256663141
- [x] Using subfolders when uploading screenshots doesn't work. see https://github.com/nextcloud/android/pull/358#issuecomment-256665297
- [x] Depending on the upload behavior (like "just upload") the instant upload should not copy the uploaded files into the app (at least they should be removed after they have been uploaded, else instant upload is not respecting its own preferences)

**Tasks Andy**
- [x] Folder sync overview recycler view with header implementation
  - [x] Performance tuning, async setting of the image
- [x] Folder sync custom settings UI
- [x] Media query implementation for standard options for media folders (sync deactivated by default)
- [x] Smart sorting for sync folder list (weighted, mostly alpha-numerical, active/disabled)
- [x] Show message if no (really no) media folder is available (e.g. fresh created emulator)
- [x] Fix state (enabled/disabled) persistence bug
- [x] Always display the latest media (add order by clause to media queries)
- [x] Hide menu item below Android 5
- [x] Hide (legacy) instant upload preferences on Android 5+
- [x] Feature pop-up with a hint to the new folder sync functionality and deactivation of the legacy instant upload
- [x] Remove legacy shared preferences for Android 5+
- [x] Further UI tweaking of the settings screen
- [x] Ability to disable a folder in its settings
- [x] Camera uploads lead to "new" Camera folder in the instant upload screen, hiding the original one (suspicion: wrong distinct in the media query + filter needed for Nc internal folders containing media)
  - [x] Filter Nextcloud app folders from being displayed


AndyScherzinger:
@tobiasKaminsky the UI works (at least for my tests) and the Upload works too (at least tried for Camera and Instagram for Upload always and upload on Wifi only). So depending on your tasks (moved them to this PR, see at the top) this can be integrated into the beta client. What do you think?

res/values-sw600dp/dims.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  Nextcloud Android client application
+
+  Copyright (C) 2016 Nextcloud
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License version 2,
+  as published by the Free Software Foundation.
+
```
tobiasKaminsky: GPL 3?

res/xml/folder_sync_preferences.xml
```diff
@@ -0,0 +1,92 @@
...
+<!--
+  ownCloud Android client application
+
+  Copyright (C) 2012  Bartek Przybylski
+  Copyright (C) 2012-2013 ownCloud Inc.
+
+  This program is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License version 2,
+  as published by the Free Software Foundation.
+
```
tobiasKaminsky: GPL 3?

src/com/owncloud/android/datamodel/MediaFolder.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package com.owncloud.android.datamodel;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Created by scherzia on 22.09.2016.
+ */
+public class MediaFolder {
```
tobiasKaminsky: remove dummy author code?

res/xml/folder_sync_preferences.xml
AndyScherzinger: DONE

res/values-sw600dp/dims.xml
AndyScherzinger: DONE

tobiasKaminsky:
let us do a "final" test and get it ready for beta :+1:

AndyScherzinger:
feel free :wink: -> I tested it based on the latest commit: https://github.com/nextcloud/android/pull/358/commits/235b01b4bf0a3a80728f712cd7bcdd8b215edfe2

tobiasKaminsky:
IRC please  ;-)

eppfel:
So, is the accessible from the standard ui e.g inside a folder, can I open open the instant upload preferences from the three button menu?

Why is there an cloud icon next to the local folder in the preferences view? 

Awesome work, now I have to add something to the whatsnew

AndyScherzinger:
@eppfel 

> So, is the accessible from the standard ui e.g inside a folder, can I open open the instant upload preferences from the three button menu?

No, it is its own view, die "normal" file list you see in the app is the server side, while the view shown above is actually a view on all media folders on your phone

> Why is there an cloud icon next to the local folder in the preferences view?

it is the enabled/disabled icon, but yeah I also wasn't happy with it. Suggestion: use a standadr Android switch element - what do you think?

> Awesome work, now I have to add something to the whatsnew

Looking forward to this too :smiley:

eppfel:
> "normal" file list you see in the app is the server side, while the view shown above is actually a view on all media folders on your phone

Hihi, I'm a dummy, but this confusion, could be room for improvement: We have remote view, files/folders that are mirrored and InstantUpload. We should think about, how to make it more clear and why not, say, I select my "Photos" folder on Nextcloud and in the menu jump to InstantUpload preferences, where the local folder is not selected? Too big of a deal? I know this goes towards sync and maybe is further down the roadmap...

> Suggestion: use a Standard Android switch element - what do you think?

Disabling makes all other settings ineffective, so it has nothing to do with the local folder in particular, right? Just the settings are stored, but not active? Therefore, I feel this should be a distinct option. Which boils down to yes 😁

AndyScherzinger:
> select my "Photos" folder on Nextcloud and in the menu jump to InstantUpload preferences, where the local folder is not selected? Too big of a deal? I know this goes towards sync and maybe is further down the roadmap...

Sync is on the roadmap as in #285 v2 while this PR just resolves v1 ;) The question to me is flow/navigation. Personally I want to "sync a local folder with the server" not a "remote folder with a device folder" as in it is more about upload than download, so I would (again personally) expect to start with the local folder and say "sync this with the server, use the following remote folder).  Down the road the new view would be extended to support specialized "file views" as in media (images/videos/audio) and evolve to any folder pair over time. So imho the actual solution is fine for instant uploads for images (I say images because this also includes screenshots for example, the media query API gives me any image folder present on the phone).

> Disabling makes all other settings ineffective, so it has nothing to do with the local folder in particular, right? Just the settings are stored, but not active? Therefore, I feel this should be a distinct option. Which boils down to yes

It has absolutely nothing to do with the local folder, right. It just gives you the power to enable/disable the sync for this local folder. Nevertheless all settings can still be changed and will be saved but not activated. The local folder is something you can't change, it is fixed but I felt it would be nice to show it (even though it is a technical thing), since users might have several folders on the phone with the same name, but with different locations (Google Photos for example then just shows both "Camera" folders but you don't have a clue which one is which/where except for actually looking at the thumbails...).

So like this?
2016-10-27 10_47_59-syncedfolders - c__dev_src_android_nextcloud_syncedfolders - syncedfolders -

eppfel:
> So imho the actual solution is fine

Agreed. Maybe I'll think about wording or other ways to differentiate the three ways better...

> So like this?

The toggle is fine. Maybe combine the title with the local folder item. Just greying out is a weak indicator for an  immutable property. Something like:

---

**Instant Upload Preferences**
for the folder *Camera*
in `/storage/emulated/0/DCIM/Camera`

---


Or shorter?

---
**Instant Upload Preferences**
for *Camera* in `/storage/emulated/0/DCIM/Camera`

---

AndyScherzinger:
Went for the long version since you might have long Names/Paths, like Whatsapp (the path-line is allowed to be 2 lines is size:
device-2016-10-27-135634

short path like camera:
device-2016-10-27-135530

Do we want to do this styling of the text? Pro is makes it easier to distinguish the information, Con it feels very "hectic" to me.

AndyScherzinger:
Hi @eppfel Here is the changed settings screen. Not sure about the different styles/typefaces:
device-2016-10-27-135634
device-2016-10-27-135530

eppfel:
> Do we want to do this styling of the text? Pro is makes it easier to distinguish the information, Con it feels very "hectic" to me.

It does. I wanted just to propose some highlighting, but these 4 font style are a bit much. Maybe we could even leave out for folder .... Although it is kind of technical, the name is in the path.

AndyScherzinger:
> Maybe we could even leave out "for folder ...." Although it is kind of technical, the name is in the path.

So here are the next 3 variants:

* Monospace font for path
device-2016-10-27-150331
device-2016-10-27-150320

* Folder name in path made bold, standard font
device-2016-10-27-150606
device-2016-10-27-150554

* no styling at all
device-2016-10-27-150753
device-2016-10-27-150809

eppfel:
Nice, thx for the variants!!! 

> Folder name in path made bold, standard font

I like this one. You realize it is a path due to the slashes, so the monospace font is not really needed, but the bold name helps. I think this is an improvement. What do you think?

AndyScherzinger:
As actions speak louder than words... https://github.com/nextcloud/android/pull/358/commits/a72056161cb7cb95f90b61e132ade127e546a29e 
--> just committed *Folder name in path made bold, standard font* :roller_coaster:

enoch85:
I'm happy to announce that it works on Android 7! Just tried the latest beta. :D

AndyScherzinger:
Awesome Feedback @enoch85 :tada: Hoping others will report back positive too :smiley: 
Did you try anything else besides the standard/camera folder like other app's media or screenshots?

AndyScherzinger:
aaaah, the "null bug", yeah found out about that one too yesterday. @tobiasKaminsky will probably look into it today :)

enoch85:
Great! 

Also, using subfolders when uploading screenshots doesn't work.

When is this expected to reach stable?

AndyScherzinger:
Well, it got released to the beta channel now effectively and I'd say it depends on how many issues will be reported by the beta users/testers (maybe 1-2 weeks, also depending on @tobiasKaminsky's point of view), then we will release the RC version (which is basically the "Google Play Beta Programm") and leave it there for 1-2 weeks and if no issues get reported back, we release the final version. Looking at the release though (https://github.com/nextcloud/android/milestone/7) we already have 38 closed PRs/issues which is A LOT, this release needs quite some testing as in beta/RC release time. 

So the fuzzy answer would be within the next weeks, if everything goes well, before December, sometime in November also depending on @tobiasKaminsky since stable releases are a joint decision.

tobiasKaminsky:
> Also, using subfolders when uploading screenshots doesn't work.

Please give me some details:
you have enabled "using subfolders" in the settings for "screenshots"? Note, that you have all options separate for each folder.

iskradelta:
This is great! \o/

AndyScherzinger:
@tobiasKaminsky you did only commit your fix to the beta branch https://github.com/nextcloud/android/pull/169/commits/fa856582812a0837c997d3ca8cf0463e1cf1c290 Was this on purpose for testing before cherry picking it to the feature branch?

@tobiasKaminsky do you now how I get the Nextcloud apps installation path? Right now the implementation seems to copy the files during instant upload and that leads to a folder inside the app which gets picked up by my query:

* I need to filter Nextcloud app folders
* depending on the upload behavior (like "just upload") should not copy the uploaded files to the app

iskradelta:
@AndyScherzinger It is weird why anyone would want to copy files for-upload to the app, but that was done and PR #362 removes it.

AndyScherzinger:
@iskradelta well it has been a requested feature, I think @tobiasKaminsky is using it for example, so yes while it is an edge case we support it since it is a configurable option and the default is "just upload, leave the file where it originates".

jancborchardt:
Woah, awesome work! :tada: :)

Really just 2 details:
- [ ] We always capitalize only the first word, so in the navigation as well as header »Instant upload« with »upload« lowercase. I know in the folder name it’s uppercase and I guess it’s hard to change there?
- [ ] For the preferences with the toggle, I would actually super-simplify it: Just show the main folder as the header »Camera«, »WhatsApp«, »Instagram« etc. and the toggle to the right of it. Maybe »Instant upload preferences« as grey second line. But the important thing (folder name) needs to be the header. It’s visible that it’s preferences.

AndyScherzinger:
> We always capitalize only the first word, so in the navigation as well as header »Instant upload« with »upload« lowercase. I know in the folder name it’s uppercase and I guess it’s hard to change there?

The Navigation isn't an issue, the FolderName could be changed too, since we haven't release a RC, question is should the folder be named ```Instantupload``` then? This is "just" the default and the user can (of course) use any folder(name) - this is configurable! :grin: - I am fine either way. 

> For the preferences with the toggle, I would actually super-simplify it: Just show the main folder as the header »Camera«, »WhatsApp«, »Instagram« etc. and the toggle to the right of it. Maybe »Instant upload preferences« as grey second line. But the important thing (folder name) needs to be the header. It’s visible that it’s preferences.

can be implemented for sure, just be aware that in case you have more than one media folder with the same name that you can't distinguish the two except for the shown thumbsnails. If that is fine, consider it done. Else the only way to distinguish them would be t display the full path as it is doesn't at the moment. It is an edge case though since most people probably don't have several image folders with the same name and if so the content differs. 

What do you think @jancborchardt @eppfel?

enoch85:
@tobiasKaminsky 

> you have enabled "using subfolders" in the settings for "screenshots"? Note, that you have all options separate for each folder.

Yes, it's enabled in the screenshots view.

eppfel:
>> For the preferences with the toggle, I would actually super-simplify it: Just show the main folder as the header »Camera«, »WhatsApp«, »Instagram« etc. and the toggle to the right of it. Maybe »Instant upload preferences« as grey second line. But the important thing (folder name) needs to be the header. It’s visible that it’s preferences.

> can be implemented for sure, just be aware that in case you have more than one media folder with the same name that you can't distinguish the two except for the shown thumbsnails. If that is fine, consider it done. Else the only way to distinguish them would be t display the full path as it is doesn't at the moment. It is an edge case though since most people probably don't have several image folders with the same name and if so the content differs.

I thought about having the folder as header, too, but was unsure, because I would not want to leave out the path for the reasons @AndyScherzinger mentioned. And Android users actually are still forced to "understand" these folder structures in other apps, so completely hiding this information is not a good way. I understand that important info should be the focus, but by switching it and having "preferences" the secondary line, we contradict the model of the rest of this view: Primary: Property (dark+semibold), Secondary: value(light+regular).
So, secondary (lightgrey) should be "enabled" or "disbaled" ...

mh, no suggestion right now

enoch85:
@tobiasKaminsky Sorry, I was wrong. It works!

enoch85:
Updated with some more screenshots here: https://cloud.techandme.se/index.php/apps/gallery/s/uhDKTfvHTVKtjkm#

The % is off by several thousands in the progressbar when uploading.

@oparoz Is there a bug in the gallery? It seems like it's just spinning in the subfolders.

enoch85:
Found another bug. When the original file gets deleted the pictures that are being uploaded becomes corrupt.

enoch85:
Ok here it goes...

#### Steps to reproduce
1. Sync the screenshots folder
2. Choose to delete or move the photo to ownCloud (not translated btw)
3. Take a screenshot

#### Expected behaviour
The photo should be uploaded and not corrupt

#### Actual Behaviour
The photo gets corrupt and the progressbar shows several thousands of precent (%).

doits:
Just tried the latest beta 20161027 and have been outside, taking some pictures with google camera app. Instant upload is enabled for camera folder on Wifi only. After returning to Wifi it only uploaded the last picture I shot. Idle for 30+min now, so does not look like it wants to upload the others. Android 7.1.

AndyScherzinger:
@tobiasKaminsky I am not mixing it up, "keep in original folder" for my camera folder still leads to a file copy to the nextcloud directory on my phone creating a camera folder too and this is the issue I am refering to. So there seems to be a bug as in *not respecting the upload behavior in the new instant upload implementation*.

AndyScherzinger:
@eppfel @jancborchardt for the "Instant upload Preferences" screen we know need it in a version where you can choose local and remote folder to configure any kind folder pair (e.g. since S7 edge users are mounting the sd card as an USB drive, Samsung's camera saves the photos there, and Android doesn't seem to care, so Media queries don't find these images...) other than that it'll give you let's say "Auto upload" (since instant upload is technically wrong, since it is not instant anymore due to Android Doze).

eppfel:
Ah, ok. So we basically take your original version @AndyScherzinger, but make the local folder item mutable?
And should we change the name to Auto Upload then? Or is Instant Upload a well known principal that users will recognize?

AndyScherzinger:
@eppfel @jancborchardt using my original version and making it mutable would be best, yes. Question is then, what to put in the headline. For "existing" items it is easy, since we can do what Jan suggested, showing the local folder's name. When creating a new one we do not know the local folder since it hasn't been chosen yet. So there need to be 2 variants (create new one, change existing one).

As for the name discussion: Instant upload is by now well now since that is the name we went along with in the past, same for oC. With respect to Android Doze and when allowing any kind of folder and file(type) it is not instant anymore and not limited to the camera(photos) thus we are having an auto upload. So imho we should change it to "Auto upload" since that is what it is and even if it might be strange in the beginning it is not confusing anyone or raising wrong expectations (already had beta users mentioning, that the camera uploads aren't "instant", which it isn't....by design...).

fossibaer76:
Hi,

today I tested the latest beta (20161027). I only have the "error on local copy bug", everything else seems to working fine.

Greets
Fossi

tobiasKaminsky:
> So there seems to be a bug as in not respecting the upload behavior in the new instant upload implementation. @AndyScherzinger 

Can you open a new issue, describe it exactly, assign it to me and I will fix it tomorrow.

AndyScherzinger:
@tobiasKaminsky Done, see #368

enoch85:
With latest beta from 20161102 the "Auto upload" stopped working on Nexus5x (Android 7). Tested both with Screenshots and regular camera uploads (both the camera folder and the app folder).

AndyScherzinger:
@enoch85 did you try to disable/enable the sync for the two folders? (Just an idea what could be the problem though). I ran the branch on my phone and just tested screenshots successfully but the branch is already having new code changes. I'll check the release tag code version next.

enoch85:
@AndyScherzinger 

Oops, forgot to disable "Only upload via Wifi", now it uploads the file but it's still corrupted, but there are still some issues with the %. Same bug as here: https://github.com/nextcloud/android/pull/358#issuecomment-256930362

doits:
Still having the problem with beta-20161105 that it only uploads the last picture taken. To reproduce:

1. Set upload to wifi only
2. Take some pictures without wifi
3. Enable wife

Result: It only uploads the last picture.

AndyScherzinger:
@doits I can reproduce the log crash and opened a PR fixing the issue #372

I could reproduce the issue with just the last / one image being uploaded following the steps you descibed.

AndyScherzinger:
Small update as for the release schedule, discussed with Tobias. Instant Upload is currently broken on Android 7+. This implementation is now already available in the beta release for some time but we are still facing issues when it comes to, folders on external storage (as in USB mounted memory), uploads not seem to get triggered on all the devices out there, Auto upload screen showing image folders within Nextcloud (e.g. photos you downloaded - which should be hidden of course).

So to not postpone the release candidates much longer for the next stable release we discussed to:
* fix as many of the issues mentioned 
* activate it only for Android7+ (in RC/stable)
* move on from there fixing issues (via >= 1.4.x releases)

tobiasKaminsky:
:+1:
Stable enough :tada:

AndyScherzinger:
Yeah! :rocket:

enoch85:
Just tested the 20161123 beta version and there are still some issues with Android 7.

- When auto-uploading pictures it first say that it fails, and then it retries and it suceeds
- When deleting a photo in the Instanupload folder it says it can't be deleted, but it's deleted anyway (haven't tested other folders or files, but it should be the same behaviour for other files as well)
- When choosing Move to Nextcloud folder or delete in the Auto-upload menu it doesn't upload and  [this]https://github.com/nextcloud/android/pull/358#issuecomment-256930362) still happens

What's what I've found so far. Do you guys have a Android phone with 7 installed? You should see the same behaviour.

> Stable enough 

Naah... But close :)

AndyScherzinger:
Thanks for the tests @enoch85 we'll look into it! :) The stable enough is meant as-in "stable enough for a release candidate" :D - I do have a Nexus5X running the latest (today's) Android 7.1.1 beta so I should be able to reproduce your findings.

as for the first fails, yeah that doesn't look nice. The initial fail happens every now and then since Android might just kill the process at the initial upload and later Tanks to Doze the retry sets in.

cc @tobiasKaminsky

enoch85:
Ok, just want to confirm that Auto upload now works as expected on Android 7 Nexus 5x.

jancborchardt:
You rock @tobiasKaminsky @AndyScherzinger :metal: :)

sunny7773:
gb

Issue #351, opened 2016-10-19
 Local storage being used for uploads from sdcard 


tobiasKaminsky:
If you upload say 10 files from SD card the android app will copy the 10 files to the local storage on the device before uploading.

Files can be found in /nextcloud/user@//...

This is extremely inefficient and wastes storage space on local device. This part of the upload process needs to be re-coded.

This only applies to files uploaded manually, auto upload does not do this.

https://help.nextcloud.com/t/local-storage-being-used-for-uploads-from-sdcard/3102

iskradelta:
Maybe a reason why this was done in the first place, was to upload the file even if it has been deleted, moved or renamed after the upload has started,  especially if upload is chunked or wifi is lost meanwhile, is this the desired behavior?

tobiasKaminsky:
Is this still happening?
@mario as you have a device with an external sdcard?

tobiasKaminsky:
Verified, not happening anylonger.

ankitsejwal:
Hi @tobiasKaminsky, I would like to report that this issue still persists. I am new to NextCloud and I uploaded around 35GB of files to NextCloud (External drive) on my Openmediavault using NextCloud Android app. I faced two issues:
Android phone storage got full (issue #2532) and
OMV disk space was full as well (see attached screenshot)

I was able to empty the phone through clearing data from NextCloud folder using a file manager. 
> However, I can't find NextCloud directory in my Openmediavault, can you please help me find the NextCloud directory as my sd card storage is now full and the uploads have stopped.


The scenario in short:
Uploading photos to NextCloud using Android app, on a NextCloud attached with External storage

What's happening:
Files are being uploaded to External drive :) but,
Phone and SD card on my Raspberry Pi (running OMV) ran out of storage :(

>Ideal scenario:
Upload photos to External storage attached to NextCloud without building unnecessary data onto phone or OMV SD card.

Thanks

![Screen Shot 2019-05-23 at 4 48 45 am](https://user-images.githubusercontent.com/13817482/58205322-9e9d1c00-7d21-11e9-93d1-9ff7eaddcd05.png)

PR #362, opened 2016-10-27
Avoid local filecopy on manual upload

iskradelta:
https://github.com/nextcloud/android/issues/351

### DIFF
--- a/src/com/owncloud/android/operations/UploadFileOperation.java
+++ b/src/com/owncloud/android/operations/UploadFileOperation.java
@@ -308,7 +308,7 @@ public class UploadFileOperation extends SyncOperation {
 
             /// check if the file continues existing before schedule the operation
             if (!originalFile.exists()) {
-                Log_OC.d(TAG, mOriginalStoragePath.toString() + " not exists anymore");
+                Log_OC.d(TAG, mOriginalStoragePath + " not exists anymore");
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@tobiasKaminsky can you have a look since you are the master of upload behavior :smiley:

tobiasKaminsky:
> https://github.com/nextcloud/android/pull/358#issuecomment-256783195 @AndyScherzinger wrote
> @iskradelta well, it has been a requested feature. I think @tobiasKaminsky is using it for example. So yes while it is an edge case we support it since it is a configurable option and the default is "just upload, leave the file where it originates".

@AndyScherzinger you are mixing two things up:
- instant upload allows keep file, delete move, move file (prior to these three options there was also "copy")
- copy file during upload: this is always done if the file is not on the same filesystem

The latter case is fixed by @iskradelta in this PR.
We now need to figure out why this initially was implemented this way and if we can remove if now without problems.
@iskradelta I hope I remember correctly that you came up with this: how did you experienced this? This should only be a temporary file which will be deleted after a successful upload?

Again many thanks @iskradelta for your great work here!

AndyScherzinger:
@tobiasKaminsky can you do a code review here (or is your last comment already the review result? - not sure, so I ask :) - So this can be merged to beta :)

PR #372, opened 2016-11-06
fix crash during log sending when exposing URI on Android 7+

AndyScherzinger:
as reported in https://github.com/nextcloud/android/pull/358#issuecomment-258614965 and reproduced by me, sending the log file crashes the app since from Android 7 on exposing file:// URIs is prohibited.

This PR fixes the issue.

@tobiasKaminsky please review and merge since the master is already targeting Android N :smiley:


AndyScherzinger:
pushed to beta branch

tobiasKaminsky:
:+1:

PR #389, opened 2016-11-19
in case of sharedPref not accessible fallback to "do nothing"

tobiasKaminsky:
#388


tobiasKaminsky:
I know that this does not solve the cause of the problem, but the crash reveals a wrong handling of null setting. It should never lead to a crash, but should work with a (maybe wrong) default value.

AndyScherzinger:
:+1:

PR #395, opened 2016-11-22
Updated federated sharing edit options

AndyScherzinger:
Update federated sharing options checking if re-sharing is allowed.

Depends on https://github.com/nextcloud/android-library/pull/25 and feedback from @schiessle in which server version this is available (to be set in the library)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -183,7 +183,7 @@ dependencies {
     compile name: 'touch-image-view'
     compile 'com.android.support:multidex:1.0.1'
 
-    compile 'com.github.nextcloud:android-library:1.0.16'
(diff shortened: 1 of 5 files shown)


schiessle:
@AndyScherzinger the permissions where introduced in 9.1

LukasReschke:
> @AndyScherzinger the permissions where introduced in 9.1 

Don't we have a capability for that? - Hardcoding versions seems error prone :)

AndyScherzinger:
@LukasReschke I don't know, for this feature I simply did a cherry pick

tobiasKaminsky:
:+1:

AndyScherzinger:
Then also the usual steps open, releasing new lib version and updating dependency here :)

PR #401, opened 2016-11-23
Fix DavDroid integration (append necessary addition to mURI)

justin-sleep:
DavDroid integration doesn't work as is, because the mURI isn't actually what DavDroid needs. I've appended "/remote.php/dav", as that is the URI necessary for DavDroid to successfully find the share.

### DIFF
--- a/src/com/owncloud/android/authentication/AccountUtils.java
+++ b/src/com/owncloud/android/authentication/AccountUtils.java
@@ -40,6 +40,7 @@ public class AccountUtils {
     private static final String TAG = AccountUtils.class.getSimpleName();
 
     public static final String WEBDAV_PATH_4_0_AND_LATER = "/remote.php/webdav";
+    public static final String DAV_PATH = "/remote.php/dav";
     private static final String ODAV_PATH = "/remote.php/odav";
     private static final String SAML_SSO_PATH = "/remote.php/webdav";
     public static final String STATUS_PATH = "/status.php";
--- a/src/com/owncloud/android/ui/activity/Preferences.java
+++ b/src/com/owncloud/android/ui/activity/Preferences.java
@@ -476,9 +476,7 @@ public class Preferences extends PreferenceActivity
         if (getPackageManager().resolveActivity(davDroidLoginIntent, 0) != null) {
             // arguments
             if (mUri != null) {
-                davDroidLoginIntent.putExtra("url", mUri.toString()
-                        + AccountUtils.getWebdavPath(AccountUtils.getServerVersion(account), MainApp.getAuthTokenType())
-                        );
+                davDroidLoginIntent.putExtra("url", mUri.toString() + AccountUtils.DAV_PATH);
             }
             davDroidLoginIntent.putExtra("username", AccountUtils.getAccountUsername(account.name));
             //loginIntent.putExtra("password", "...");
--- a/src/com/owncloud/android/ui/activity/Preferences.java
+++ b/src/com/owncloud/android/ui/activity/Preferences.java
@@ -476,7 +476,7 @@ public class Preferences extends PreferenceActivity
         if (getPackageManager().resolveActivity(davDroidLoginIntent, 0) != null) {
             // arguments
             if (mUri != null) {
-                davDroidLoginIntent.putExtra("url", mUri.toString());
+                davDroidLoginIntent.putExtra("url", mUri.toString() + "/remote.php/dav");
             }
             davDroidLoginIntent.putExtra("username", AccountUtils.getAccountUsername(account.name));
             //loginIntent.putExtra("password", "...");
--- a/src/com/owncloud/android/ui/activity/Preferences.java
+++ b/src/com/owncloud/android/ui/activity/Preferences.java
@@ -476,7 +476,9 @@ public class Preferences extends PreferenceActivity
         if (getPackageManager().resolveActivity(davDroidLoginIntent, 0) != null) {
             // arguments
             if (mUri != null) {
-                davDroidLoginIntent.putExtra("url", mUri.toString() + "/remote.php/dav");
+                davDroidLoginIntent.putExtra("url", mUri.toString()
+                        + AccountUtils.getWebdavPath(AccountUtils.getServerVersion(account), MainApp.getAuthTokenType())
+                        );
             }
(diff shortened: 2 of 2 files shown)


AndyScherzinger:
Hi @justin-sleep thanks for this contribution! 🎉 
@rfc2822 what do you think? The actual integration as discussed when implemented just provides the main URL and DAVdroid does all the detection magic for the DAV url.
Other than that I think if we change that we should ask the server for the DAV url (which can be retrieved via API afaik.

rfc2822:
DAVdroid doesn't do any magic ;) It uses [well-known URLs (and services) for CalDAV/CardDAV](https://tools.ietf.org/html/rfc6764) if available. If those are not available, there's no way for DAVdroid to detect the `/remote.php/dav/` path. So I would vote for adding it: if the DAV path is known, there's no reason to drop it and rely on other mechanisms like well-known URLs, which may not be set up [correctly] and thus cause difficulties.

tobiasKaminsky:
@justin-sleep do you have any special server base url?

justin-sleep:
@tobiasKaminsky I'm not sure what would be considered special. My Nextcloud instance is positioned at "domain.com/cloud".

justin-sleep:
Rather than hardcoding the WebDAV path, I've made the change to retrieve it from the server.

AndyScherzinger:
@justin-sleep looks good to me! :)
@tobiasKaminsky what do you think? I am not using DAVdroid (or any other), so for mass testing I'd say we push it to the next beta release.

justin-sleep:
Turns out that getWebDavPath will return /remote.php/**web**dav, rather than /remote.php/dav. When DAVdroid gets that it... doesn't work.  Whoops!  Out of habit, I had accidentally tested my own branded fork of the app yesterday, and just noticed this mysterious unbranded "Nextcloud" app on my device :)

I've done a preliminary once-over of the relevant code, and it seems that nextcloud/server@c79496 was never accounted for in AccountUtils. I've added a simple string there that contains the path we need, then referenced that when calling DAVdroid.  I realize this is pretty close to just hardcoding it in, but as far as I can tell we don't need to check for authentication token type, as DAVdroid will need the same URL regardless.

AndyScherzinger:
@MorrisJobke any idea who can help with this path issue? @justin-sleep I just check my "standard, unzip and install" installation on my server and the path is ```/remote.php/webdav/``` !

MorrisJobke:
We added a general `remote.php/dav` (in 10 or 9 AFAIK). The `remote.php/webdav` is since then only an internal redirect to `remote.php/dav/files/USERNAME/...`

Under `remote.php/dav/` are all other endpoints browsable:

```
$ cadaver http://localhost/server/remote.php/dav
Authentication required for Nextcloud on server `localhost':
Username: admin
Password: 
dav:/server/remote.php/dav/> ls
Listing collection `/server/remote.php/dav/': succeeded.
Coll:   addressbooks                           0  Jan  1  1970
Coll:   calendars                              0  Jan  1  1970
Coll:   comments                               0  Jan  1  1970
Coll:   files                                  0  Jan  1  1970
Coll:   principals                             0  Jan  1  1970
Coll:   public-calendars                       0  Jan  1  1970
Coll:   systemtags-relations                   0  Jan  1  1970
Coll:   systemtags                             0  Jan  1  1970
Coll:   uploads                                0  Jan  1  1970
dav:/server/remote.php/dav/> 
```

That means that `remote.php/dav` is a valid endpoint and should be used to autodetect endpoints ;)

MorrisJobke:
> @MorrisJobke just checked, the server tells me webdav while the documentation says dav... --> https://docs.nextcloud.com/server/10/user_manual/files/access_webdav.html

Both work for backwards compatibility ;)

AndyScherzinger:
@MorrisJobke @rfc2822 @justin-sleep so we should then just provide the link ```http(s)://myawesomeserver/dav``` ? And again since when is this available? (which is important to decide if there is the need for some version checks for backwards compatibility.

MorrisJobke:
> so we should then just provide the link http(s)://myawesomeserver/dav ?

No, `/remote.php/dav/`. Since 9.0 https://github.com/owncloud/core/pull/18688

justin-sleep:
Are there any other changes you'd like me to make?

AndyScherzinger:
Hi @justin-sleep,

sorry for the late reply, been more or lees offline for the weekend. So according to @MorrisJobke your used path is the right one. So this fix works and depending on @tobiasKaminsky's feedback we can merge this after 1.4.0 has been released. 

Other than that I think for a long term fix we should  use ```AccountUtils.getWebdavPath(AccountUtils.getServerVersion(account)``` while this one should then return ```...remote.php/dav/``` instead of ```...remote.php/webdav/```.

mario:
What's happening with this PR? Can it be scheduled for 1.4.2 if it has been tested?

tobiasKaminsky:
We should include @AndyScherzinger 
This way we have the right path in one central place.

mario:
@AndyScherzinger thoughts? :)

AndyScherzinger:
I tested it and it is fine. I am okay with having it merged the way it is now, while we should then rewrite it for e.g. 1.5.0 release since:

* https://github.com/nextcloud/android/pull/401/commits/66b6436d2b6f48783dd7931c6fd1c20b0c81fbce is actually the right fix
* while we then also need https://github.com/nextcloud/android/pull/401/commits/b5ca2cdc9008937f1ef72cc79195584fa3a95637#diff-20c923e713598571723283b67ce38a06R43 and 
* https://github.com/nextcloud/android/pull/401/commits/b5ca2cdc9008937f1ef72cc79195584fa3a95637#diff-20c923e713598571723283b67ce38a06L184 should return https://github.com/nextcloud/android/pull/401/commits/b5ca2cdc9008937f1ef72cc79195584fa3a95637#diff-20c923e713598571723283b67ce38a06R43

AndyScherzinger:
👍

mario:
👍

tobiasKaminsky:
@AndyScherzinger can you then please open a new issue containing your enhancement ideas?

AndyScherzinger:
sure thing! (just haven't found the time yet)

AndyScherzinger:
@tobiasKaminsky @mario here is the new issue for further improvement: #629

PR #402, opened 2016-11-24
minor layout optimization so the German "skip" (überspringen) fits in…

AndyScherzinger:
… a single line

@tobiasKaminsky please review and merge. It is a very simple layout optimization to give some more space (no left/right padding) to the skip button on the what's new screen since "ÜBERSPRINGEN" would end up having a line break before the "N" so this PR gives the text as much space as possible.


tobiasKaminsky:
:+1:

Issue #445, opened 2016-12-12
Minor Usability improvements


eppfel:
I looked closely to the current app and spotted some things we could improve for usability:

# Missing space between avatar and filename :white_check_mark: 
`papercut` - see #449

![screenshot_20161124-194840](https://cloud.githubusercontent.com/assets/1899308/21112965/13410e06-c0a8-11e6-8d94-2d72c02acb93.png)

# Add icon to make state more recognize :white_check_mark: 
Reading the text takes time, an icon is easily remembered and then you would spot on first glimpse, that this is an empty folder. see #449

![screenshot_20161124-195142](https://cloud.githubusercontent.com/assets/1899308/21112456/216f2744-c0a6-11e6-8e88-eb580a74e939.png)

## Some folders have opaque white background in multiselect
![screenshot_20161124-195506](https://cloud.githubusercontent.com/assets/1899308/21112623/c214f89a-c0a6-11e6-9bd4-5fb6108b72d5.png)

## Wrong top-left icon after menu swipe :white_check_mark: 
When you are in an folder and swipe from the left edge of the screen the menu opens. After closing the menu again, the icon has changed from `back` to `menu`, but the action is still `back`. see #449

![screenshot_20161212-195635](https://cloud.githubusercontent.com/assets/1899308/21112835/93afcbf0-c0a7-11e6-81cd-86d3c1c6541b.png)

#  Trim filenames in respect to each other
In this view you can not distinguish the files, because they follow the same pattern `P3_[Date]_[Content]_[Owner].[extesnion]`. Not so easy to fix, I guess and not a use-case that happens often: Therefore Low priority

![screenshot_20161124-195114](https://cloud.githubusercontent.com/assets/1899308/21112234/4ebc597a-c0a5-11e6-89ae-8d76418a1df8.png)

### Environment data
Android version: 6.0.1

Device model: Sony E5823

Nextcloud app version: Latest Beta (30.11.16)

AndyScherzinger:
Great eye for details @eppfel!!! :+1:

AndyScherzinger:
3 out of 5 have been fixed by see #449

eppfel:
- [x] Missing space between avatar and filename https://github.com/nextcloud/android/pull/449/commits/7012be9baa2396b6064479a0ee8fd72433225be3
- [x] Add icon to make state more recognize https://github.com/nextcloud/android/pull/449/commits/246acd94bbd11f5e78988103934302ca14e1307f
[...]

PR #449, opened 2016-12-13
Minor Usability improvements

AndyScherzinger:
This is a PR for #445 solving the first 3 issues:

- [x] Missing space between avatar and filename https://github.com/nextcloud/android/pull/449/commits/7012be9baa2396b6064479a0ee8fd72433225be3
- [x] Add icon to make state more recognize https://github.com/nextcloud/android/pull/449/commits/246acd94bbd11f5e78988103934302ca14e1307f
- [x] Wrong top-left icon after menu swipe https://github.com/nextcloud/android/pull/449/commits/1d420261fcbca95d0ab44ea534c02c5569714313
- [ - ] Some folders have opaque white background in multiselect (not sure if/how this can be fixed) - _postponed_
- [ -  ] Trim filenames in respect to each other (not sure if/how this can be fixed) - _postponed_

So for the last two issues I need some help from @nextcloud/designers :)


AndyScherzinger:
# Missing space between avatar and filename https://github.com/nextcloud/android/pull/449/commits/246acd94bbd11f5e78988103934302ca14e1307f
device-2016-12-13-172744

# Add icon to make state more recognize https://github.com/nextcloud/android/pull/449/commits/246acd94bbd11f5e78988103934302ca14e1307f
Basically did this now for: Uploads, Move/Copy, Empty remote folder, and loading :)
device-2016-12-13-172619
device-2016-12-13-154723
device-2016-12-13-172336
device-2016-12-13-172357
device-2016-12-13-172417
device-2016-12-13-172433

AndyScherzinger:
For the code review you just need to take a look at:
* https://github.com/nextcloud/android/pull/449/commits/246acd94bbd11f5e78988103934302ca14e1307f
* https://github.com/nextcloud/android/pull/449/commits/7012be9baa2396b6064479a0ee8fd72433225be3
* https://github.com/nextcloud/android/pull/449/commits/1d420261fcbca95d0ab44ea534c02c5569714313

anything else are just organize imports/shortenings for resolving codacy warnings 😁

AndyScherzinger:
@eppfel 

> Wrong top-left icon: Can I test this easily?

Yes it'll be in the next beta release (all of the mentioned changes...)

1. exactly, the action will have to be another PR (and some more specifics and what shall be done)
2. exactly, empty folder in the cloud
3. empty folder in the cloud (no subfolders) _when_ choosing a destination for uploading a file shared via another app with Nextcloud (only action then is to choose the folder or like it says to create a sub-folder)
4. exactly, choosing a file to be manually uploaded so it shows your local FS and you ended up in an empty folder(Sorry for the folder name "Crash Me")
5. exactly, ups... same as 3 yes.... my mistake
6. shown when you refresh a cloud folder which is or has been empty (when you enter an folder it'll be displayed as empty, if you at some point in time navigate to it again we will check the server for changes). So yes, the Nothing in here isn't necessary. We can show any text you like us to show :) we can also put the Loading... as the "headline" and not have the smaller text-line below.

jancborchardt:
I think @eppfel already covered most of it. I mainly had two thoughts:

- »You can add a folder!« – How? Via the 3-dot-menu? Would be cool if it would also be possible by tapping the big folder icon with +
- »No files in here« and »Loading« in one screen? That should never happen. Either it’s loading or there are no files. And »No files in here« should only ever show when there are absolutely no files in that folder, not locally and not on the server – and not when there’s just a loading delay or a connection error.

And awesome work as usual ;)

eppfel:
I would not make the icon able to tap and use the `+` action button in all cases for consistency, but jep we should offer that option.

So, let's get this into the beta, so I can get a better idea, of the several cases. I review in #445 then.

AndyScherzinger:
> I would not make the icon able to tap and use the `+` action button in all cases for consistency, but jep we should offer that option.

I would suggest to not do that since the main action is to pick a folder (target when you want to upload files) and the FAB is Android's UI element for the main action and in this case there is none since it is a "simple" folder picker, else we have a secondary action button (cancel), a primary action button (the blue choose) and might confuse users when there is then also a blue FAB to create a folder.

The loading screen has been fixed by simply just stating "Loading" 
device-2016-12-30-175205

tobiasKaminsky:
:+1:

Issue #457, opened 2016-12-14
SVG images can not be shown


jancborchardt:
We should be able to show SVG images. ;)

AndyScherzinger:
Seems doable but would introduce a new library to actually be able to show SVGs, like https://github.com/BigBadaboom/androidsvg (APL2) with ~ 168.5 KB

What do you think @tobiasKaminsky ? I don't mind the minor size increase + web can do it ;)

PR #465, opened 2016-12-20
show SVG images

AndyScherzinger:
This PR adds support for viewing SVG images as requested in #457.

It adds a library to support proper rendering (since standard Android can't properly open/display e.g. inkscape svg files): https://github.com/nextcloud/android/pull/465/files#diff-c197962302397baf3a4cc36463dce5eaR51


AndyScherzinger:
@tobiasKaminsky can you take a look at this and also merge it to the beta? 
I tried but it only works for downloaded files, for the non-downloaded files it seems your thumbnail/preview logic gets triggered and nothing happens. Unfortunately I couldn't find the implementation which does that. Any help/hints are highly welcome.

jancborchardt:
Wooo SVG support! :+1:

mario:
What's the status of this PR?

AndyScherzinger:
basically done, only thing is that it isn't compatible with @tobiasKaminsky's preview-thumbnail-fetching PR since for SVG we need to download the file to be able to render it on the client (no serverside SVG preview rendering available).

So this can be merged but then Tobias' PR would need to take care of not previewing SVG or Tobias' gets merged first than this one need to take care of this part.

tobiasKaminsky:
We should get #69 merged soon. Otherwise this PR will stay there forever :/

AndyScherzinger:
Closed in favor of #805

PR #472, opened 2016-12-23
Auto upload: delay job, check file

tobiasKaminsky:
- delay job execution by minimal 10s
- on job execution: check if file exists prior to real upload. If file does not exist anymore, just ignore it


AndyScherzinger:
👍

PR #510, opened 2017-01-05
Properly encode the share link password

AndyScherzinger:
cherry pick of oC 1869 for reported issue oC 1842
______
### Actual behaviour
Example
When I put _Hello%20World_ as password the client doesn't encode the value and sends it without any change to the server.
So, when I navigate to the url returned by the server and introduce _Hello%20World_ as password I get password authentication error... The password that I should introduce to get access is _Hello World_

Other example
If I choose _Hello&World_ as password -> The server takes _Hello_ as password 

### Expected behaviour
The password should be exactly what I typed
 
### Steps to reproduce
The password parameter in https://doc.owncloud.org/server/9.1/developer_manual/core/ocs-share-api.html#update-share is sent by the android client to the Server without x-www-form-encoding 
The windows desktop client is sending the password in the right way , so the server doesn't get messed up

### DIFF
--- a/src/com/owncloud/android/ui/dialog/SharePasswordDialogFragment.java
+++ b/src/com/owncloud/android/ui/dialog/SharePasswordDialogFragment.java
@@ -1,20 +1,20 @@
 /**
- *   ownCloud Android client application
- *   @author masensio
- *   Copyright (C) 2015 ownCloud Inc.
+ * ownCloud Android client application
  *
- *   This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Not needed as far as I have tested it.

mario:
Um, @tobiasKaminsky are you sure? If you are, then please close & delete branch.

tobiasKaminsky:
@AndyScherzinger should confirm this (and then he can delete the branch)

PR #513, opened 2017-01-06
Android automatic refactor - Recycle

luiscruz:
Hi,

This pull request has the changes generated while applying the rule "Recycle".

Some resources (e.g., ```Cursor``` instances) should be closed when they are no longer necessary. 

Please consider them and let me know if you agree with them.

Best,
Luis

### DIFF
--- a/androidTest/java/com/owncloud/android/datamodel/OCFileUnitTest.java
+++ b/androidTest/java/com/owncloud/android/datamodel/OCFileUnitTest.java
@@ -104,6 +104,9 @@ public class OCFileUnitTest {
         // Read the data from the parcel
         parcel.setDataPosition(0);
         OCFile fileReadFromParcel = OCFile.CREATOR.createFromParcel(parcel);
+		if (parcel != null) {
+			parcel.recycle();
+		}
 
         // Verify that the received data are correct
         assertThat(fileReadFromParcel.getRemotePath(), is(PATH));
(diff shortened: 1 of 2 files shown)


przybylski:
Hi!
Thanks for your PR.
We will look into it.

androidTest/java/com/owncloud/android/datamodel/OCFileUnitTest.java
```diff
@@ -104,6 +104,9 @@ public void writeThenReadAsParcelable() {
         // Read the data from the parcel
         parcel.setDataPosition(0);
         OCFile fileReadFromParcel = OCFile.CREATOR.createFromParcel(parcel);
+		if (parcel != null) {
+			parcel.recycle();
```
przybylski: You should use spaces instead of tabs.

androidTest/java/com/owncloud/android/datamodel/OCFileUnitTest.java
```diff
@@ -104,6 +104,9 @@ public void writeThenReadAsParcelable() {
         // Read the data from the parcel
         parcel.setDataPosition(0);
         OCFile fileReadFromParcel = OCFile.CREATOR.createFromParcel(parcel);
+		if (parcel != null) {
```
przybylski: This check makes no sense. there are already calls to the parcel, if it would be null then much worse things would happen than don't call recycle

AndyScherzinger:
any update/comment @luiscruz ?

AndyScherzinger:
I close this PR since we never got any replies. Please re-open if willing to work on it or discuss.

PR #514, opened 2017-01-06
Android automatic refactor - Viewholder

luiscruz:
Hi (again),

This pull request has the changes generated while applying the rule "ViewHolder".

When developing list views, it is very common to repeat work in every item. Something very common is retrieving a field with the API method ```findViewById```. There is a ViewHolder pattern that lets you reuse the work made in the previous computed item.

Please consider them and let me know if you agree with them.

Best,
Luis


mario:
@luiscruz this tool sounds like fun. Open source?

jancborchardt:
@luiscruz I added you to the Nextcloud organization and Android team by the way! :) After accepting, make sure to set your visibility from Private to Public at https://github.com/orgs/nextcloud/people

AndyScherzinger:
@luiscruz Thanks for the PR and sorry for the late reply. I haven't had the change to test drive it yet but looking at the code change it seems fine while I have two comments/issues:

* LogListAdapter - will be removed by another PR since it is unused code (not and issue, just mentioning it even though I am sure that such a detection would be out of scope for your analyzer)
* LogListAdapter & ShareUserListAdapter - seem fine while I found one optimization though: both changed code fragments keep the following code:
```
View view = convertView;
...
return view;
```

in both cases view doesn't get changed anywhere and even then it would be a call by reference, so the right change would be to not instantiate "view" and simply do a ```return convertView;```

could you do this code change? 😸

AndyScherzinger:
any update/comment @luiscruz ?

AndyScherzinger:
closing due to inactivity, please re-open if still valid

PR #515, opened 2017-01-06
Autorefactor obsoletelayoutparam

luiscruz:
Hi (again),

This pull request has the changes generated while applying the rule "ObsoleteLayoutParam".

While developing your application's views you might be specifying attributes in a view's artefact that are not necessary due to the nature of its parent. In this PR, those attributes were replaced by a comment.

Unfortunately the tool is not yet ready to change the file in a specific line and file comparison might be confusing when whitespace is not ignored.
Please consider the changes and let me know if you agree with them.

Best,
Luis

### DIFF
--- a/res/layout/list_fragment.xml
+++ b/res/layout/list_fragment.xml
@@ -1,131 +1,130 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version='1.0' encoding='utf-8'?>
+  <!--
+
   ownCloud Android client application
+    
 
   Copyright (C) 2012  Bartek Przybylski
+
   Copyright (C) 2015 ownCloud Inc.
+    
 
   This program is free software: you can redistribute it and/or modify
+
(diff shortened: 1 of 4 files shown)


AndyScherzinger:
@mario @tobiasKaminsky what do you think? Looking at the diff it is basically a massive reformatting and a hand of removed, obsolete layout params so I would vote to create a separate PR that manually removed the parameters since this PR already has conflicts with the master. :/

tobiasKaminsky:
@AndyScherzinger  good idea 👍

tobiasKaminsky:
@AndyScherzinger is your idea still valid to create a new PR from this, e.g. are the proposed changes still meaningful?
Then I'll have a look into it.

AndyScherzinger:
Imho that would still be a valid approach.
Best solution though will be to refractor most of our layout files at some point to be in line with the layout guidelines from the last io conference... But that'll take time I guess, so I vote to simply close this PR.

PR #516, opened 2017-01-07
round thumbnail size to power of 2: 384 => 256

tobiasKaminsky:
NC11 only supports thumbnails with power of 2: 128, 256, 512.
So if a mobile wants to have 384 this will result in 512px.
This PR rounds it down to power of 2: 384 => 256 instead of 512.


mario:
@AndyScherzinger this should go into 1.4.2 actually. It's a bugfix.

AndyScherzinger:
👍

mario:
👍

AndyScherzinger:
👍

PR #544, opened 2017-01-16
fix quota display

AndyScherzinger:
fixing the fetching of quota information from server via https://github.com/nextcloud/android-library/pull/28 (v 1.0.10) 
- Big Thanks to @mario for the library patch 🎉 

@tobiasKaminsky has also been merged to beta (20170117) and should imho be part of 1.4.1. What do you think?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,7 +41,7 @@ dependencies {
     /// dependencies for app building
     compile name: 'touch-image-view'
 
-    compile 'com.github.nextcloud:android-library:1.0.9'
+    compile 'com.github.nextcloud:android-library:1.0.10'
     compile "com.android.support:support-v4:${supportLibraryVersion}"
     compile "com.android.support:design:${supportLibraryVersion}"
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
shipped with beta 20170117 via the android-library dependency

AndyScherzinger:
@mario do you have a server v9 running with which this patch could be tested?

mario:
@AndyScherzinger no, but I will get it running for you later today :)

AndyScherzinger:
I just need two users then one with unlimited and one with limited quota to check if this still works for the two given scenarios :)

Thanks a lot!!!! ❤️

AndyScherzinger:
Just testdrove this PR against @mario's Nc9 installation and the quota is displayed as expected. @tobiasKaminsky @mario can be merged after review to be part on 1.4.1 RC4 :)

tobiasKaminsky:
I'll trust you ;)
:+1:

Issue #541, opened 2017-01-16
NoClassDefFoundError: android...JobScheduler


AndyScherzinger:
Via Google Play Developer Console.

* Android version: Android 4.1
* Device:  Galaxy Win (delos3geur)
* Reported for: 1.4.1 RC2, 1.4.1 RC3

To me it seems a Android version check is missing for legacy Android versions:

```
java.lang.NoClassDefFoundError: android.app.job.JobScheduler
	at com.owncloud.android.datamodel.UploadsStorageManager.getPendingJobs(UploadsStorageManager.java:388)
	at com.owncloud.android.datamodel.UploadsStorageManager.getCurrentAndPendingUploads(UploadsStorageManager.java:384)
	at com.owncloud.android.ui.adapter.ExpandableUploadListAdapter$1.refresh(ExpandableUploadListAdapter.java:150)
	at com.owncloud.android.ui.adapter.ExpandableUploadListAdapter.loadUploadItemsFromDb(ExpandableUploadListAdapter.java:620)
	at com.owncloud.android.ui.adapter.ExpandableUploadListAdapter.<init>(ExpandableUploadListAdapter.java:185)
	at com.owncloud.android.ui.fragment.UploadListFragment.onStart(UploadListFragment.java:103)
	at android.support.v4.app.Fragment.performStart(Fragment.java:2106)
	at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1146)
	at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1290)
	at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1272)
	at android.support.v4.app.FragmentManagerImpl.dispatchStart(FragmentManager.java:2154)
	at android.support.v4.app.FragmentController.dispatchStart(FragmentController.java:212)
	at android.support.v4.app.FragmentActivity.onStart(FragmentActivity.java:610)
	at android.support.v7.app.AppCompatActivity.onStart(AppCompatActivity.java:178)
	at com.owncloud.android.ui.activity.BaseActivity.onStart(BaseActivity.java:189)
	at com.owncloud.android.ui.activity.FileActivity.onStart(FileActivity.java:181)
	at android.app.Instrumentation.callActivityOnStart(Instrumentation.java:1178)
	at android.app.Activity.performStart(Activity.java:5213)
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2087)
[...]

PR #547, opened 2017-01-17
prevent NoClassDefFound

tobiasKaminsky:
ref: https://github.com/nextcloud/android/issues/541

### DIFF
--- a/src/com/owncloud/android/datamodel/UploadsStorageManager.java
+++ b/src/com/owncloud/android/datamodel/UploadsStorageManager.java
@@ -29,6 +29,7 @@ import android.database.Cursor;
 import android.net.Uri;
 import android.os.Build;
 import android.os.PersistableBundle;
+import android.support.annotation.RequiresApi;
 
 import com.owncloud.android.db.OCUpload;
 import com.owncloud.android.db.ProviderMeta.ProviderTableMeta;
@@ -392,6 +393,7 @@ public class UploadsStorageManager extends Observable {
         }
     }
 
+    @RequiresApi(Build.VERSION_CODES.LOLLIPOP)
     private List<OCUpload> getPendingJobs() {
         JobScheduler js = (JobScheduler) mContext.getSystemService(Context.JOB_SCHEDULER_SERVICE);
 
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
Tested with android 4.1

src/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -381,7 +382,11 @@ private OCUpload createOCUploadFromCursor(Cursor c) {
         );
 
         // add pending Jobs
-        return getPendingJobs().toArray(uploads);
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            return uploads;
+        } else {
+            return getPendingJobs().toArray(uploads);
```
AndyScherzinger: haven't used it for a while, will have to see if I can test this later, but wouldn't the list elements then override the array items already in there? 
What I would have thought:
```
List<OCUpload> result= getPendingJobs();
Collections.addAll(result, uploads);
result.toArray(result.size())
```

src/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -381,7 +382,11 @@ private OCUpload createOCUploadFromCursor(Cursor c) {
         );
 
         // add pending Jobs
-        return getPendingJobs().toArray(uploads);
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
```
AndyScherzinger: JobScheduler is API level 21 (Lollipop): https://developer.android.com/reference/android/app/job/JobScheduler.html

src/com/owncloud/android/datamodel/UploadsStorageManager.java
tobiasKaminsky: Right. I will change it.

AndyScherzinger:
@tobiasKaminsky review done, tested successfully, I added another commit for the cancellation, please see https://github.com/nextcloud/android/pull/547/commits/fa4439bc4619e00f984a878af11be055ff675479 and merge in case you are fine with it 👍

tobiasKaminsky:
:+1:

PR #550, opened 2017-01-18
Improvements search

mario:
Before I start explaining what I've done, I'm aware that there is an ongoing work on the search api on the server side. This PR implements local search/filtering and adds a bunch of fixes to the overall user experience. 

- search across files (either all or only On Device)
- search results are limited to 100 results, and you get a message stating that
(if you need more, your query probably isn't concise enough anyway)
- search is now nicely restored across orientation changes
- footer and FAB are hidden while searching
- FAB is closed after you close search, even if it was previously open

- search the entire local filesystem
- search results are limited to 100 results, and you get a message stating that
(if you need more, your query probably isn't concise enough anyway)
- search is now nicely restored across orientation changes

Both operations are really quick, but just in case - there's a loading/refresh thing
rotating while searching.

There are a few bugs with the local files screen that I've discovered in the process, so I'll make sure to open issues for them in the near future.

Thank you for considering this PR and don't be too harsh!


mario:

Issue #456, opened 2016-12-14
PNG files have full-screen checkerboard background


jancborchardt:
That makes them look very very strange. Instead, the background should be dark as with JPG. The checkerboard background should only be as large as the size of the image, directly behind it, as fallback if there’s transparency in the image.

przybylski:
But isn't this widely accepted way of presenting png with transparency ? What if I have image with only black shapes on transparent background ?

tobiasKaminsky:
@jancborchardt means the background if the image is not full sized like here: https://github.com/nextcloud/android/pull/560#issuecomment-273851857

PR #560, opened 2017-01-19
Bugfix 456

mario:
This PR fixes issue #456, and in addition, fixes a bug where we tried to fetch EXIF tags for PNG files that don't really have them.

I now toggle full screen on view click as well, because some images are too small and people might be having trouble clicking on them.

### DIFF
--- a/res/layout/preview_image_fragment.xml
+++ b/res/layout/preview_image_fragment.xml
@@ -31,7 +31,8 @@
     android:id="@+id/top"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:background="#000000">
+    android:background="#000000"
+    android:animateLayoutChanges="true">
 
     <ProgressBar
         android:id="@+id/progressWheel"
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
You can add your next PR/branch to our repository. 
With having all feature branches in our repo it is easier to collaborate or test the branch.

AndyScherzinger:
without having the branch at hand with the wrap_content change, what happens for a square png with a pinch-to-zoom, will it then still use the whole screen?

mario:
@AndyScherzinger not at the moment. I'll take a look and see if I can adapt this for the pinch-to-zoom situation as well.

mario:
Attaching an image so you folks can see how it looks. Pinch-to-zoom works, and zooming happens inside the actual image dimensions (aka does not expand). Let me know if it should!
screenshot_20170119-190341

screenshot_20170119-190616

AndyScherzinger:
> Pinch-to-zoom works, and zooming happens inside the actual image dimensions (aka does not expand). Let me know if it should!

@mario it did before, so I'd say yes, but than the actual solution you implemented wouldn't work anymore since it just did the view content wrapping.

mario:
@AndyScherzinger should be fixed now.

mario:

Issue #211, opened 2016-08-13
Search inside of text files


jancborchardt:
For plain text and markup files like txt notes, code, markdown, rtf and anything like that it would be great to be able to search for specific words in the document when you have it open.

My main use-case is looking for content of Notes txt files when I want to send an email or look something up.
Together with »In-App Text Editor #126« it would be a big step to make the app usable for notes.

PR #562, opened 2017-01-20
Bugfix 454 & 211

### LINKED ISSUE #454: Search icon shown when opening text file
@nextcloud/android the search icon is shown in the header when you open text files but it does nothing. I guess it’s not intended to show it there? ;)

In any case Search inside of text files #211 would be good :)

mario:
This fixes #454 and #211. 

Note that it looks like I can't push to nextcloud/android - so please look into that as well! :)

### DIFF
--- a/src/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -170,9 +170,7 @@ public class FileDisplayActivity extends HookActivity
             mWaitingToSend = (OCFile) savedInstanceState.getParcelable(
                     FileDisplayActivity.KEY_WAITING_TO_SEND);
             mSearchOpen = savedInstanceState.getBoolean(FileDisplayActivity.KEY_IS_SEARCH_OPEN, false);
-            if (savedInstanceState.getString(FileDisplayActivity.KEY_SEARCH_QUERY) != null) {
-                mSearchQuery = savedInstanceState.getString(FileDisplayActivity.KEY_SEARCH_QUERY);
-            }
+            mSearchQuery = savedInstanceState.getString(FileDisplayActivity.KEY_SEARCH_QUERY);
         } else {
             mWaitingToPreview = null;
             mSyncInProgress = false;
--- a/src/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -141,6 +141,14 @@ public class FileDisplayActivity extends HookActivity
 
     private Collection<MenuItem> mDrawerMenuItemstoShowHideList;
 
+    public static final String KEY_IS_SEARCH_OPEN = "IS_SEARCH_OPEN";
+    public static final String KEY_SEARCH_QUERY = "SEARCH_QUERY";
+
+    private String mSearchQuery;
+    private boolean mSearchOpen;
+
+    private SearchView mSearchView;
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         Log_OC.v(TAG, "onCreate() start");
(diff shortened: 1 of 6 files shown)


src/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -146,12 +169,55 @@ public void onStart() {
...
+
+        if (mSearchView != null) {
+            mSearchView.clearFocus();
+        }
+        return true;
+    }
+
+    @Override
+    public boolean onQueryTextChange(final String newText) {
+        if (getActivity().getClass().equals(FileDisplayActivity.class)) {
```
przybylski: you should use `getActivity() instanceof FileDisplayActivity`, so you can use other classes which are subclass of FileDisplayActivity

src/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -146,12 +169,55 @@ public void onStart() {
...
+            FileDisplayActivity fileDisplayActivity = (FileDisplayActivity) getActivity();
+            fileDisplayActivity.setSearchQuery(newText);
+        }
+        mHandler.removeCallbacksAndMessages(null);
+        mHandler.postDelayed(new Runnable() {
+            @Override
+            public void run() {
+                if (newText != null && !newText.isEmpty()) {
+                    String coloredText = StringUtils.searchAndColor(mOriginalText, newText);
+                    mTextPreview.setText(Html.fromHtml(coloredText.replace("\n", "<br \\>")));
```
przybylski: This may produce NPE according to your `searchAndColor` function

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -160,6 +169,10 @@ protected void onCreate(Bundle savedInstanceState) {
             mSyncInProgress = savedInstanceState.getBoolean(KEY_SYNC_IN_PROGRESS);
             mWaitingToSend = (OCFile) savedInstanceState.getParcelable(
                     FileDisplayActivity.KEY_WAITING_TO_SEND);
+            mSearchOpen = savedInstanceState.getBoolean(FileDisplayActivity.KEY_IS_SEARCH_OPEN, false);
+            if (savedInstanceState.getString(FileDisplayActivity.KEY_SEARCH_QUERY) != null) {
```
przybylski: Bundle method `getString` will return null if no mapping is available. You don't initiate it with anything else in `onCreate` so I don't see the point of this check.

src/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -88,8 +101,7 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
 
 
         View ret = inflater.inflate(R.layout.text_file_preview, container, false);
-
-        mTextPreview = (TextView) ret.findViewById(R.id.text_preview);
+            mTextPreview = (TextView) ret.findViewById(R.id.text_preview);
```
przybylski: Wrong indentation

src/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -113,19 +126,29 @@ public void onCreate(Bundle savedInstanceState) {
             mAccount = args.getParcelable(FileDisplayActivity.EXTRA_ACCOUNT);
         }
 
+        if (args.containsKey(FileDisplayActivity.EXTRA_SEARCH_QUERY)) {
```
przybylski: Same as above, it will return null when no mapping to key is available. Since you are doing it in `onCreate` you don't need such detailed checks. If you want to handle resume checks then you should move grabbing arguments to `onResume`/`onStart` and store values in `onPause`

src/com/owncloud/android/utils/StringUtils.java
```diff
@@ -0,0 +1,36 @@
+package com.owncloud.android.utils;
```
przybylski: Please add copyright header

mario:
@przybylski thank you for your review - all of the issues mentioned should now be corrected.

tobiasKaminsky:
Please only one issue per PR.
This makes a PR a little bit smaller and therefore easier to review (and to track the changes).

tobiasKaminsky:
> Note that it looks like I can't push to nextcloud/android - so please look into that as well! :)

You should have write access to nextcloud/android...
What is the error? (you can also ping me on irc)

mario:
The issues are very related @tobiasKaminsky - in fact I did not remove search at all, but instead implemented search.

mario:

PR #566, opened 2017-01-20
PNG: remove fullscreen checkerboard background

mario:
This PR fixes issue #456, and in addition, fixes a bug where we tried to fetch EXIF tags for PNG files that don't really have them.

I now toggle full screen on view click as well, because some images are too small and people might be having trouble clicking on them.

### DIFF
--- a/src/com/owncloud/android/ui/preview/PreviewImageFragment.java
+++ b/src/com/owncloud/android/ui/preview/PreviewImageFragment.java
@@ -150,17 +150,19 @@ public class PreviewImageFragment extends FileFragment {
         View view = inflater.inflate(R.layout.preview_image_fragment, container, false);
         mImageView = (TouchImageViewCustom) view.findViewById(R.id.image);
         mImageView.setVisibility(View.GONE);
-        mImageView.setOnClickListener(new OnClickListener() {
+
+        view.setOnClickListener(new OnClickListener() {
             @Override
             public void onClick(View v) {
                 ((PreviewImageActivity) getActivity()).toggleFullScreen();
             }
-
         });
+
         mMessageView = (TextView)view.findViewById(R.id.message);
         mMessageView.setVisibility(View.GONE);
         mProgressWheel = (ProgressBar)view.findViewById(R.id.progressWheel);
         mProgressWheel.setVisibility(View.VISIBLE);
+
         return view;
     }
 
@@ -441,8 +443,10 @@ public class PreviewImageFragment extends FileFragment {
                             Log_OC.e(TAG, "File could not be loaded as a bitmap: " + storagePath);
                             break;
                         } else {
-                            // Rotate image, obeying exif tag.
-                            result = BitmapUtils.rotateImage(result, storagePath);
+                            if (ocFile.getFileName().endsWith("jpg") || ocFile.getFileName().endsWith("jpeg")) {
+                                // Rotate image, obeying exif tag.
+                                result = BitmapUtils.rotateImage(result, storagePath);
+                            }
                         }
 
                     } catch (OutOfMemoryError e) {
@@ -507,6 +511,7 @@ public class PreviewImageFragment extends FileFragment {
             if (imageView != null) {
                 Log_OC.d(TAG, "Showing image with resolution " + bitmap.getWidth() + "x" +
                         bitmap.getHeight());
+
 
                 if (result.ocFile.getMimetype().equalsIgnoreCase("image/png")) {
                     Drawable backrepeat = getResources().getDrawable(R.drawable.backrepeat);
--- a/src/com/owncloud/android/ui/preview/PreviewImageFragment.java
+++ b/src/com/owncloud/android/ui/preview/PreviewImageFragment.java
@@ -81,7 +81,6 @@ public class PreviewImageFragment extends FileFragment {
     private TouchImageViewCustom mImageView;
     private TextView mMessageView;
     private ProgressBar mProgressWheel;
-    private RelativeLayout mRelativeLayout;
 
     public Bitmap mBitmap = null;
 
@@ -90,9 +89,6 @@ public class PreviewImageFragment extends FileFragment {
     private boolean mIgnoreFirstSavedState;
 
     private LoadBitmapTask mLoadBitmapTask = null;
-
-    private boolean weZoomedAlready;
-
 
     /**
      * Public factory method to create a new fragment that previews an image.
--- a/src/com/owncloud/android/ui/preview/PreviewImageFragment.java
+++ b/src/com/owncloud/android/ui/preview/PreviewImageFragment.java
@@ -158,6 +158,13 @@ public class PreviewImageFragment extends FileFragment {
             }
         });
 
+        mImageView.setOnClickListener(new OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                ((PreviewImageActivity) getActivity()).toggleFullScreen();
+            }
+        });
+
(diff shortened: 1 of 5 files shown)


tobiasKaminsky:
When testing it with a rather small png it is (in emulator) not really working.
Also it "jumps" between original size (very small) to fullscreen without any step between it.

screenshot_1484927941

owncloud1

mario:
@tobiasKaminsky 

First screenshot shows an original image and the fact that transparency is only directly behind the image, and everything else is black.
screenshot_20170120-171043

The second screenshot (once you start zooming) expands to full screen (match_parent) so you could actually move around the image, even if outside the current viewable area.

screenshot_20170120-171133

I am aware of the "jumping", and the reason for that is explained above. (though I could potentially implement animations and stuff, we still need to go full screen I'd say).

mario:
I'll see if I can come up with a more clever solution to this.

jancborchardt:
Nice @mario! And yes, we should definitely not have any jumping there :) everything needs to be smooth

mario:
@tobiasKaminsky feel free to test this again.

mario:
One more commit pending - will be up tomorrow.

mario:
Please review once you get the time - thanks!

src/com/owncloud/android/ui/preview/PreviewImageFragment.java
```diff
@@ -525,18 +527,21 @@ private void showLoadedImage(LoadImage result) {
...
-                    Drawable d = new BitmapDrawable(getResources(), bitmap);
-                    layers[1] = d;
-                    LayerDrawable layerDrawable = new LayerDrawable(layers);
-                    layerDrawable.setLayerHeight(0, (int) convertDpToPixel(bitmap.getHeight(), getActivity()));
-                    layerDrawable.setLayerHeight(1, (int) convertDpToPixel(bitmap.getHeight(), getActivity()));
-                    layerDrawable.setLayerWidth(0, (int) convertDpToPixel(bitmap.getWidth(), getActivity()));
-                    layerDrawable.setLayerWidth(1, (int) convertDpToPixel(bitmap.getWidth(), getActivity()));
-                    imageView.setImageDrawable(layerDrawable);
-
+                    if (getResources() != null) {
```
AndyScherzinger: now you call getResources twice, you could alternatively just check if ```r != null``` :)

src/com/owncloud/android/ui/preview/PreviewImageFragment.java
```diff
@@ -598,6 +603,30 @@ private void finish() {
         container.finish();
     }
 
+    private void toggleImageBackground() {
+        if (getFile() != null && getFile().getMimetype().equalsIgnoreCase("image/png")) {
+            if (getActivity() != null && (getActivity() instanceof PreviewImageActivity)) {
```
AndyScherzinger: the two if statements can be combined into a single one

src/com/owncloud/android/ui/preview/PreviewImageFragment.java
```diff
@@ -598,6 +603,30 @@ private void finish() {
...
+                PreviewImageActivity previewImageActivity = (PreviewImageActivity) getActivity();
+                if (getResources() != null) {
```
AndyScherzinger: same applies for this statement, it can also be combined with the top ones, since you don't do anything if one of the checks fails.

AndyScherzinger: Is the null check necessary? Just asking :)

src/com/owncloud/android/ui/preview/PreviewImageFragment.java
AndyScherzinger: Like in the other comment is the null check necessary?

src/com/owncloud/android/ui/preview/PreviewImageFragment.java
```diff
@@ -572,6 +600,34 @@ private void finish() {
...
+            }
+
+            layerDrawable.setDrawableByLayerId(layerDrawable.getId(0), layerOne);
+
+            mImageView.setImageDrawable(layerDrawable);
+        }
+    }
+
+
+    private static float convertDpToPixel(float dp, Context context) {
```
AndyScherzinger: Just looking at the diff it seems this method is new and only used in one place where you always need to cast to int, thus I'd say simply return an int.

src/com/owncloud/android/ui/preview/PreviewImageFragment.java
mario: This is already combined?

AndyScherzinger: 👍

AndyScherzinger: ...most likely a comment on an older commit :)

tobiasKaminsky:
Now I do not see any checkerboard on a png file.
Is it only a problem of me and my emulator?

mario:
@tobiasKaminsky no it's not - not this time. This behaviour is by design, and mimics popular gallery apps that actually show white instead of checkboard UNTIL you click on the image. So please, go ahead and click on the image :)

tobiasKaminsky:
Aha, yeah, I get it.
But I am not sure if a regular user is used to it, @jancborchardt?

mario:
Yea, @jancborchardt approved this change, and even suggested it in the first place.

jancborchardt:
Yep, tbis is also how Wikipedia does it. Checkerboard is unusual for anyone, and only really needed when there are white parts in an image with transparency. So it shouldn't be the default :)

:+1:

AndyScherzinger:
Two issues:

1) open a PNG and click on the image (shows the toolbar): the toolbar has the drawer menu icon set but should have the back arrow icon set
device-2017-01-26-131606

2) after that rotate the screen: toolbar has been removed while it shouldn't be "lost" via rotation
device-2017-01-26-131801

mario:
@AndyScherzinger I will fix the first issue, but to me the second sounds like a desired behaviour? How is it currently in the app?

mario:
@AndyScherzinger it does not look like I can reproduce the problems you're seeing?

mario:
OK, found the first issue somehow. The second issue is fixed in the another PR related to images. I will now fix the backarrow, and leave the second issue after we're done with this PR.

mario:
Thank you for finding this bug btw.

mario:
Just FYI, once this is merged I'll rebase the other PR on master and make it ready for testing.

AndyScherzinger:
Awesome @mario! Ans also sorry for my report. I just checked the actual master and both issues I mentioned are the current behavior (even though imho both are valid issues).

I also checked Google Photos and is behaves as I described, a device rotation doesn't change the Toolbar state, so it stays visible and visible/invisible is only toggled/changed by the click.

AndyScherzinger:
What I meant to say is that the PR is fine with me but the other two issues should be solved (at some point) but aren't part of this PR.

mario:
@AndyScherzinger since two separate branches of are related to the same thing, I am working on merging them and will present them as one PR - but since I know that is evil, please go ahead and review the code on both and if both is fine we'll merge the integration branch.

AndyScherzinger:
Code review on this one: 👍

AndyScherzinger:
@mario Which is the other PR you are referring to?

mario:
@AndyScherzinger PR #576

jancborchardt:
@mario will the branch still be used or not? If not please delete it so we do not have stale branches floating around. ;)

PR #578, opened 2017-01-23
Grey shared indicator

### LINKED ISSUE #577: Share icon in the file list should be grey and have some information
My wife and me always get confused by the black share icon. I always try to click it to get information from whom the file was shared or expect it to allow me to share it.
But it just opens the file/folder.

If you only want to indicate that the file is shared, the icon should be gray scaled and not black

AndyScherzinger:
This PR resolves #577 replacing the black link/user shared icons with grey ones.

Please check the screenshots, the grey is fine in the list view but isn't optimal in the grid view. A way to fix this would be to use the "old" black ones for the grid view for images (all other file types are fine).

What do you think @eppfel ?

device-2017-01-23-171309
device-2017-01-23-171259


nickvergessen:
Looks better 👍

AndyScherzinger:
cc @eppfel and @jancborchardt for design feedback :)

eppfel:
It's an improvement. The overlay isn't perfect, but don't have any better idea now.

AndyScherzinger:
@mario @tobiasKaminsky please review :)

mario:
@AndyScherzinger not much to review :D I don't like the icon in the grid view, but it's otherwise fine.

AndyScherzinger:
Thanks @mario for the review, not much to check but Tobias and I agreed on "no merges without reviews" when the fork started. About the icon in the grid view, is it about the fact that there are icons in the grid view in general or about the grey coloring? As I said, we could also use the black version for the grid and the grey version for the list view.

eppfel:
> About the icon in the grid view, is it about the fact that there are icons in the grid view in general or about the grey coloring?

Just overlays are problematic. Blend modes are better in these situations, but I don't want to bloat this PR. What could also work, is white icon on dark images and black icon on bright ones. But that's als complicated.

>As I said, we could also use the black version for the grid and the grey version for the list view.

No, not in general.

Still a go 🚀 from my side...

AndyScherzinger:
Thanks for the feedback @eppfel then let's go with the actual state of the PR and see how we can improve this in a future iteration, e.g. semi-transparent white-ish with a minimal dark outer glow. Image check for coloring might be tricky since this add processing (thus battery) just for an "overlay").

So just waiting for @tobiasKaminsky's review :)

jancborchardt:
Cause this is a short-term solution to #577 (cause it will still be clicked), please check out the long term plan at: File actions in the list #579

AndyScherzinger:
Since this is a more then minor change (basically just icons...) but helps with UX I would vote for it to be released with 1.4.2 where we "just" release the permission fix (that caused some disturbance).

What do you think @tobiasKaminsky @mario ?

mario:
Of course, do it.

mario:
Btw I would argue that silly PRs like this can get in automatically. We
should exercise trust among each other and contributors that have write
access rather than spending time on formalities - with git, it becomes
really easy to revert anything.

AndyScherzinger:
We can discuss such an approach while I still tend to favor reviews on all PRs which leads to simple PRs taking a little longer to be merged but takes away the subjective part of deciding (and defining) if a PR can be considered simple. It also shows potential new contributors that we take this seriously and that there are no shortcuts for maintainers, certain people etc. Else we might run into discussion why a PR has been merged directly. As for the revert issue, we can't. By design the master is protected so a "revert" would need to be done via a PR! That is also how the server team handles development (protecting the master).

It is a formality, true but on the other we also experienced bugs introduced via very simply PRs that haven't been thoroughly looked at and were shipped to the user base. I'd say the formality costs a little bit of time but saves us from embarrassing issues to some extend.

mario:
@AndyScherzinger thank you for a very detailed explanation of the process.

jancborchardt:
Totally agree with @AndyScherzinger on the process. This is also how we do it in the server and other apps. The pull request review process is important so everyone – the existing contributors, the volunteers, the full-timers, as well as people entirely new to the project – know what’s going on. And there’s always mistakes made even in the smallest changes. It’s just good practice. Saying »we can revert anything anyway« will just lead to people being unhappy and prevents us from learning from each other. :)

PR #576, opened 2017-01-23
Improvement loading error views

mario:
This fixed bug #455, and bears the stamp of approval of @jancborchardt xD

PR #590, opened 2017-01-27
Integration

mario:
This PR integrates PR #576 and #566 so it would be easier to review and test. If it's too complicated, please use the separate ones for code review, but this one for testing.


tobiasKaminsky:
In the PR I see the string "Image can't be shown", but I do not see it here:
screenshot_1485631370


At least in emulator right after clicking on an image I see this grey header. In master the image is shown directly in fullscreen and only after a click image name and overflow menu are shown (also without an header).
screenshot_1485631358

When clicking on the image the header disappears, after clicking again the header is visible again, but not with the grey bar.
screenshot_1485631364

mario:
I will check both issues. Anything else? Code-wise? Can you upload this image so I can test?

mario:
(Also, header is here on purpose and shown on purpose - the only thing that shouldn't be here is the grey color)

mario:
@AndyScherzinger @tobiasKaminsky ui issue fixed.

mario:
@jancborchardt say yes to all this, go go go :p

tobiasKaminsky:
So I have tested 
- error image
- png with trasparent background
- regular images
- rotation on all cases

Anything else that was changed and I can/should test?

mario:
Loading text files and loading videos is now nicer.

AndyScherzinger:
Have it this way then, this branch also has conflicts with the master and needs rebasing+conflicts resolved ;P

mario:
You can also test error videos.

mario:
So demanding, do you ask this much of every new contributor? :P

AndyScherzinger:
No, I usually do it myself (for mostly all branches we have), but with this many commits and conflicts the risk is just to high to break the whole thing and thus testing/review doesn't make sense since the could will differ after the rebase and it might also be broken. So no reason to test/review 😉

src/com/owncloud/android/ui/preview/PreviewImageFragment.java
```diff
@@ -363,34 +394,17 @@ private void openFile() {
...
     }
 
-    
+
     private class LoadBitmapTask extends AsyncTask<OCFile, Void, LoadImage> {
 
         /**
          * Weak reference to the target {@link ImageView} where the bitmap will be loaded into.
-         *
+         * <p>
```
mario: No idea, but fixing.

mario:
@AndyScherzinger should be fine now.

mario:
Actually, I have a clear integration branch - gonna open a PR based on it now. @tobiasKaminsky please approve that other one when I open it :P

PR #595, opened 2017-01-28
Update beta build process desc

tobiasKaminsky:


mario:
Why does this includes more than one commit (it should only include "add description for automatically beta build process")?

tobiasKaminsky:
Cause I had messed with my branch and accidently uploaded the wrong files, which I then revert.
And now I have no idea how to revert it :/

mario:
@tobiasKaminsky I'll close this PR - please kill the branch and create a new one with a proper commit :) While I could tell you how to revert individual commits, we would still end up with "Reverted xyz" which I'd like to avoid in this situation :)

PR #602, opened 2017-01-29
Integration of pullapprove

AndyScherzinger:
This is the PR for integration https://pullapprove.com

It demands:
* Any PR to be reviewed by at least 2 devs
* Any PR labeled with design to be reviewed by at least 1 designer (Jan, Felix)
* Approval can be approved via commenting: :shipit:, 👍, Approved or LGTM
* Pushes to the PR resets the review process for the code review
* Reopening of a PR resets the review process
* Authors can't approve their own PR

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -40,4 +40,4 @@ groups:
     required: 1
     users:
       - jancborchardt 
-      - eppfel 
+      - eppfel
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1,0 +1,43 @@
+version: 2
+
+# General settings to apply
+always_pending:
+  title_regex: '(WIP|wip)'
+  labels:
+    - 1. to develop
+    - 2. developing
+  # custom message that will be used for the GitHub status
+  explanation: 'This PR is a work in progress...'
+
+# Group settings to apply to all groups by default, optionally being overridden later
+group_defaults:
+  author_approval:
+    ignored: true
+  approve_by_comment:
+    enabled: true
+    approve_regex: '^(Approved|:shipit:|:\+1:|LGTM)'
+    reject_regex: '^(Rejected|:-1:)'
+  reset_on_push:
+    enabled: true
+  reset_on_reopened:
+    enabled: true
+
+groups:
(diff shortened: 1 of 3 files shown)


mario:
Will this mean that if I make a positive review, and someone pushes new things my review won't get reset? Because I think it should.

AndyScherzinger:
at the moment it doesn't but we can activate that. The reason why I didn't activate it is, that I don't know what happens if the PR gets rebased, because then the commits change (hash wise) but that shouldn't reset the review.

mario:
You can still end up in trouble even if just rebasing :)

AndyScherzinger:
SOLD! I activated the review reset for the code review while the designer review won't be reset. This way we don't overdo it and as devs we will make sure that a designer's approval is still valid in case of further code pushes.

.pullapprove.yml
```diff
@@ -0,0 +1,43 @@
...
+  design-review:
+    conditions:
+      labels:
+      - design
+    reset_on_push:
+      enabled: false
+    required: 1
+    users:
+      - jancborchardt 
+      - eppfel
```
mario: Add a newline.

mario: Add a newline here please.

AndyScherzinger:
Thanks @mario ❤️

tobiasKaminsky:
Let us see what happen 👍

AndyScherzinger:
see regex, the 👍 must be at the beginning of a line ;)

AndyScherzinger:
@tobiasKaminsky looks good :)
2017-01-29 20_56_31-nextcloud_android

mario:
@AndyScherzinger please do this for library as well, minus the designers.

AndyScherzinger:
https://github.com/nextcloud/android-library/pull/32

AndyScherzinger:
Already did ;)

AndyScherzinger:
cc: @MorrisJobke @LukasReschke we replaced LGTM with pullapprove which works an is more flexible :)

@jancborchardt @eppfel the plugin now enforces an approval for all PRs labeled with design :)

PR #605, opened 2017-01-30
Don't restore from unknown state for migration. Fixes #597

### LINKED ISSUE #597: Can not change storage directory to external SD card
### Actual behaviour
- the apps stops, if the user attempts to change storage path to external SD

### Expected behaviour
- change storage path to external SD
 
### Steps to reproduce
1. Storage path in preferences is shown as "unknown"
2. click on "Storage path", two directories are shown now
  a) /storage/emulated/0/Android/0/data/com.nextcloud.client/files
  b) /storage/756C-1905/Android/0/data/com.nextcloud.client/files
3. click on the second directory (external SD card)

### Environment data
Android version: 7.1

Device model: Sony Xperia Z

Stock or customized system: customized (Lineage OS 14.1-20170110-UNOFFICIAL-yuga)

Nextcloud app version: 14.1

Nextcloud server version: 11.0.1

#### Web server error log
```
192.168.1.15 - sony [28/Jan/2017:21:44:48 +0100] "GET /nextcloud/ocs/v1.php/cloud/users/sony?format=json HTTP/1.1" 200 232 "-" "Mozilla/5.0 (Android) ownCloud-android/1.4.1"
192.168.1.15 - sony [28/Jan/2017:21:44:49 +0100] "GET /nextcloud/index.php/avatar/sony/384 HTTP/1.1" 200 31 "-" "Mozilla/5.0 (Android) ownCloud-android/1.4.1"
```

```
(There is no nextcloud.log? Logcat log is available)
```

przybylski:
fixes #597

### DIFF
--- a/src/com/owncloud/android/ui/activity/StorageMigration.java
+++ b/src/com/owncloud/android/ui/activity/StorageMigration.java
@@ -175,7 +175,7 @@ public class StorageMigration {
                                      String source,
                                      String target,
                                      ProgressDialog progressDialog,
-                                     StorageMigrationProgressListener listener) {
+                                     StorageMigrationProgressListener listener) throws SecurityException {
             mContext = context;
             mStorageSource = source;
             mStorageTarget = target;
@@ -242,6 +242,12 @@ public class StorageMigration {
         }
 
         protected void restoreAccountsSyncStatus(boolean oldSync[]) {
+            // If we don't have the old sync statuses, then
+            // probably migration failed even before saving states,
+            // which is weird and should be investigated.
+            // But its better than crashing on ArrayOutOfBounds.
+            if (oldSync == null)
+                return;
             for (int i = 0; i < mOcAccounts.length; ++i) {
                 ContentResolver.setSyncAutomatically(mOcAccounts[i], mAuthority, oldSync[i]);
             }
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
👍

tobiasKaminsky:
👍

tobiasKaminsky:
@mario 
@AndyScherzinger go for 1.4.2?

@przybylski we do not require GH review system any longer. We can still use it for adding comments to the code review. But PullAprove replaced LGTM and this is the only blocker (despite CI)

AndyScherzinger:
I am fine with 1.4.2 since it is a bugfix

mario:
If I'm correct, this just hides a crash? But go for it for now, and please open an issue to investigate what's mentioned in a comment in this commit.

przybylski:
@mario this is related with data migration (swapping storage). I know exactly what was causing this. If you want I can elaborate :)

mario:
@przybylski would appreciate it :)

przybylski:
When we prepare for migration from one storage point to another one we first check the availability of target path, cleanup if needed (after user approval) and than starting migration.
Just before the migration we need to make sure that no synchronization is currently in progress, and that we restore everything to the state as it was before. So after the initial cleanup we save the current state of synchronization, copy data, and restore synchronization status of accounts.
If something goes wrong while copying data we can restore. Beforehand it doesn't make any sense to restore to the same state, which might trigger additional synchronization. In this bug it was causing problems even before saving current sychronization state, eg. not not enough space, insufficient permission etc. It caused a background task to crash, and user was clueless what happened. It should be more informative now after crash handling.

mario:
Thank you @przybylski.

PR #621, opened 2017-02-03
reformatting Code

tobiasKaminsky:
I used this option.
As you can see *no* rearrange code as this was changing way too much (in my opinion).

screenshot_20170203_154906


AndyScherzinger:
@tobiasKaminsky Where do I find these settings? BTW: The reformatting also changed the translation files.

mario:
I actually like "Rearrange code", I just don't like <p> stuff that happens with reformatting :-/

tobiasKaminsky:
"Re-arrange code" destroyed the order of comments, e.g. a comment was not right above the variable it should describe.
What do you not like on "reformatting"? Theoretically we can define our own code/style guide, but then we must make (somehow) sure that this is used by AndroidStudio automatically.

mario:
@tobiasKaminsky understood. I forgot to clarify what I don't like - I don't like <p> :)

tobiasKaminsky:
...? What? Am I missing a part of your answer?

mario:
Closing for now, but we need to do reformatting eventually for sure.

tobiasKaminsky:
👍 should be done when most of the outstanding PRs are merged ;-)

Issue #532, opened 2017-01-11
Feature: unlock with fingerprint


laurentdinclaux:
### Actual behaviour
I can unlock the app with a pin

### Expected behaviour
I would like to unlock it using my fingerprints

PR: #624

joseandroidengineer:
I've worked with adding fingerprint features so I'll give it a whirl!

AndyScherzinger:
Sounds awesome @joseandroidengineer!

DanielYWoo:
Today I setup my own nextcloud and give it a try, it looks cool, but this feature is missing ;-)
It will be very convenient on iPhone.

jshatch:
The owncloud app has had this request in for a year and nothing was done.  I'm excited to see how this progresses here.

DanielYWoo:
@jshatch That's exactly why I dumped ownCloud!

tobiasKaminsky:
@joseandroidengineer if I can assist you while building this feature, please just ping me here or on irc.

Flole998:
That pull request should add that feature.

JSoko:
Fingerprint lock works an SGS7 nougat with beta 20170211 👍

AndyScherzinger:
will be shipped with the upcoming version 1.5.0

PR #624, opened 2017-02-04
Added fingerprint support

Flole998:
In response to #532 this adds Fingerprint support.


Flole998:
Using Pincode and Fingerprint Unlock is technically possible and I have tested that. So we just need to know if it is also wanted.

In Android it is not possible to directly access the Fingerprint Scanner. You can only rely on the System doing it. In case you don't have any Fingerprints setup, it will show you a message and unset the checkmark that no fingerprints are setup. If the device doesn't have a fingerprintscanner the checkbox is hidden completely.

res/values-de-rDE/strings.xml
```diff
@@ -28,6 +28,8 @@
     <string name="prefs_accounts">Konten</string>
     <string name="prefs_manage_accounts">Konten verwalten</string>
     <string name="prefs_passcode">PIN gesperrt</string>
+    <string name="prefs_fingerprint">Fingerabdruck gesperrt</string>
+    <string name="prefs_fingerprint_notsetup">Es sind keine Fingerabdrücke eingerichtet.</string>
```
mario: Like @tobiasKaminsky, please rollback changes to this file entirely. I'll review everything one of these days.

AndyScherzinger:
I don't use either but I think there should be some fallback, so I'd say if the fingerprint mechanism is activated then the user must also set a PIN.

Flole998:
@AndyScherzinger To prevent users from becoming locked out? To disable fingerprint all you have to do is disable it in the Android Options as that will delete all Fingerprints from the system. Nextcloud will just start up then and the fingerprint setting will become disabled.

AndyScherzinger:
@Flole998 Thanks for the details, not an expert on Android fingerprint API. 
@tobiasKaminsky @jancborchardt Than I think we are fine with not having fingerprint AND pin both activated at the same time.

tobiasKaminsky:
As fingerprint is more "secure" than pin I would also vote for hiding the pin entry possibility if fingerprint is activated.

mario:
While I see a point for fingerprint + pin, I'd vote for supporting 1 option at a time. So if you use fingerprint, you don't use PIN. And vice versa.

Flole998:
Seems like you guys want only 1 method possible at the same time. I think using a check box is not the right approach then, if i set a check box i don't want my device to disable it cause I checked another one. A radio button popup as seen on the storage config is more appropriate in my opinion. Also i think "details" should be renamed to "security" as it contains pin, fingerprint and hiding files, which are security options.

mario:
@Flole998 I'll check the actual UI of this PR tomorrow and let you know how to approach the UX changes. Sounds good? :)

Also, what's your opinion? Would you prefer to leave both options at the same time?

Flole998:
Alright, sound great.

As i already said above, technically it is possible and has been tested to leave both on at the same time, that would result in both being required to access the app. I dont see any reason to not give the user that option. If someone is really paranoid they might want to use both at the same time, while for most users it would be useless but again: They don't have to enable it, it's just an additional option.

mario:
Please accept my apologies, got sidetracked with fixing auto upload :( Will do this first thing Monday!

mario:
Android 5.1:

02-15 03:54:25.688 2170-2170/com.nextcloud.android.beta D/Error: ERR: stack=java.lang.NoClassDefFoundError: Failed resolution of: Landroid/hardware/fingerprint/FingerprintManager;
                                                                     at com.owncloud.android.ui.activity.FingerprintActivity.isFingerprintCapable(FingerprintActivity.java:257)
                                                                     at com.owncloud.android.ui.activity.Preferences.onCreate(Preferences.java:216)
                                                                     at android.app.Activity.performCreate(Activity.java:5990)
                                                                     at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1106)
                                                                     at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2278)
                                                                     at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2387)
                                                                     at android.app.ActivityThread.access$800(ActivityThread.java:151)
                                                                     at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1303)
                                                                     at android.os.Looper.loop(Looper.java:135)
                                                                     at android.app.ActivityThread.main(ActivityThread.java:5254)
                                                                     at java.lang.reflect.Method.invoke(Method.java:372)
                                                                     at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
                                                                     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
                                                                  Caused by: java.lang.ClassNotFoundException: Didn't find class "android.hardware.fingerprint.FingerprintManager" on path: DexPathList[[dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-touch-image-view_c8dffc8b90fcf3f6583cf9e0f7a2f818f4736e83-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-support-annotations-24.2.1_34fe3faf5e52f4f087d54f64861ffb04d7db2100-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-slice_9-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-slice_8-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-slice_7-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-slice_6-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-slice_5-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-slice_4-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-slice_3-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-slice_2-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-slice_1-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-slice_0-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-slf4j-api-1.7.4_64612b92d52d59a748b72eb4964a8efaa7a0e40e-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-parceler-api-1.1.6_66939c28e42bc1692af1095bd5a000a865256c32-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-jcl-over-slf4j-1.7.4_464b1651e89a21a6da6832c36529c2de9a5bbdff-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-jackrabbit-webdav-2.12.4_ff66740b8a0872fc718aac43275dfae9737fe00f-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-internal_impl-24.2.1_f4f7a8f1842d3507974837d92c568e8f38919b03-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-internal_impl-24.2.1_d2c56de47c7e8f8459e122f9e08a5edabab1e8fe-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-internal_impl-24.2.1_cf1cefe11bf46d4ef24adbcc3b0b33eb49bd8e32-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-internal_impl-24.2.1_b7fcf0666312744f9764b8047c1ef990a6a14cef-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-internal_impl-24.2.1_2e0f6537e29e2efcbbb55800bc067de32bdfbf19-classes.dex", dex file "/data/data/com.nextcloud.android.beta/files/instant-run/dex/slice-disklrucache-2.0.2_e03082df31a5979e81f7fc0559b90e02

mario:
@Flole998 can you look into that error maybe? :)

jasonbayton:
@Flole998 so _you've_ been making my app crash ;)

Fingerprints aren't necessarily more secure than PIN and certainly not password. Here's some reading with a few external links to research: http://security.stackexchange.com/questions/144428/how-secure-is-a-fingerprint-sensor-versus-a-standard-password

In every implementation of this I've ever come across on mobile, there's *always* a PIN/password fallback. Fingerprints can be temperamental on occasion and no one wants to be locked out of their NC app because they couldn't fall back to another auth method.

Flole998:
@jasonbayton I'm sorry. I haven't considered that some vendors might remove the fingerprint manager for whatever reason. I will fix that.

If you haven't seen an implementation where there's no fallback you could check out LastPass, over there it's not possible to bypass the fingerprint protection. Also you are not forever locked out, by disabling the fingerprint protection in Android settings it gives you a warning and then removes all fingerprints. That enables you to open nextcloud app again and will disable the fingerprint setting then, so after re-enabling it in Android preferences you would have to re-enable it in nextcloud settings aswell.

mario:
@Flole998 I think the problem in his case is that he uses Android 5.x, and FingerPrint manager was introduced in 6.0. Before that, Samsung used their own system for accessing fingerprint data (though I'm pretty sure J3 doesn't have fingerprint functionality).

jasonbayton:
> @mario I think the problem in his case is that he uses Android 5.x, and FingerPrint manager was introduced in 6.0. Before that, Samsung used their own system for accessing fingerprint data (though I'm pretty sure J3 doesn't have fingerprint functionality).

Yes and yes. Not so much removed as replaced with something else I guess. 

> @Flole998 by disabling the fingerprint protection in Android settings it gives you a warning and then removes all fingerprints.

So if I'm in a rush to get some files out of NC and my sensor isn't working (greasy fingers, hardware issue, software glitch, whatever) I have to: 

* open settings
* disable fingerprint, deleting all stored fingerprints on my phone
* do my task in the now unlocked NC app

Then later: 

* reopen settings
* re-enable fingerprint
* re-add my four fingers 
* re-enable it in the NC app (as well as any other apps I have it enabled in I expect)

That isn't particularly straightforward. A fallback would be much simpler.

Flole998:
Fallback to what? A 4 digit PIN?

jasonbayton:
Or passcode, yes, as is typically standard behaviour across mobile devices (lastpass excluded as you pointed out).

tobiasKaminsky:
@jasonbayton this is a very good idea:
During setup the fingerprint the user has to enter a password which acts as a failback. It should be offered after the third false attempt @jancborchardt 

What is the decision about allowing both, fingerprint and pin?

mario:
@tobiasKaminsky with "password" you mean 4 digit pin, right? :)

jasonbayton:
> @mario @tobiasKaminsky with "password" you mean 4 digit pin, right? :)

Does it matter too much? If a user wants to add an alphanumeric, 16 character password.. why not? Who knows what people keep on their NC servers? ;)

mario:
And you want to protect it with an easily hackable protection on an insecure operating system? Really? :P

Flole998:
Yeah and add safetynet check so nextcloud doesn't run on unlocked and rooted devices ;)

jasonbayton:
I'll pretend I didn't see that. :P

mario:
I do agree that PIN can serve as a fallback though, but that is easily "hackable" using basics of social engineering, but who am I to judge how good your pins are :P

jasonbayton:
10,000 possible combinations in 4 digits, a million with 6. It's not _that_ insecure. With access to the device you can crack the fingerprint also. All EMM platforms offer a sane default of 10 attempts before the device wipes, which negates the brute force aspect of a PIN (or to a lesser extent, password). I don't expect NC to erase its own app data on the 10th failure though in combination with the MDM controls mentioned in https://github.com/nextcloud/server/issues/3456 we could well see improvements longer-term.

tobiasKaminsky:
> @tobiasKaminsky with "password" you mean 4 digit pin, right? :)

@mario no, I meant a real password as my intention was that this feature is independent from the pin code. And passwords allow "safer" ciphers than a forced 4 digit pin (of course you can "abuse" a password and use only a 4 digit one)

Flole998:
What do you guys think about adding another checkbox to the settings menu that can only be enabled when fingerprint setting and PIN are both enabled and that setting could switch between "Require PIN and Fingerprint" and "Use PIN as Fallback"? As there are a lot of different opinions on how it should, i think that would be the best option so the user can choose.

jasonbayton:
I was thinking about that earlier; if PIN can be set as a fallback, it could just as easily be required for 2-factor authentication. In that instance I'd say having to reset and reconfigure the fingerprint if you can't get it working would be acceptable (because really there's no other way around that, it'd otherwise be no different to fallback to PIN).

Do checkboxes offer the best user experience? Would a dropdown be more suitable where there are more than 2 options?

Flole998:
Considering that if Fingerprint is not supported on that device the settings will be hidden, you might end up with a dropdown with just 1 item in it. Also you would have to have a lot of options in that dropdown then:
- No Protection
- Pin Only
- Pin and Fingerprint
- Pin and Fingerprint as Fallback
- Only Fingerprint

I think 3 checkboxes are easier in that case, considering 2 would disappear when Fingerprint is not supported. In the dropdown the last 3 items will be removed leaving just the first 2 items which would look weird to users not having Fingerprint support.

jasonbayton:
Ah OK, I hadn't considered that. Checkboxes it is!

jshatch:
Why can't the fallback be entering your account password?

jasonbayton:
I guess because the app passcode is a whole separate thing to account info. 

Would it even be possible to verify against the account password?

jshatch:
I was thinking more along the lines of this:  Even if you are using an app specific password and 2FA, the app should still be able to call into the server with your main password and validate it is correct.  You wouldn't have to worry about 2FA, if the password is accepted, let them into the app with the stored credentials.  Probably more code than the scope of this PR, but I think it could work.

tobiasKaminsky:
Using the account password should be possible.

mario:
What's happening with this PR folks?

Flole998:
I would suggest we merge it and then create a new issue/feature request to discuss bypassing the fingerprint.

AndyScherzinger:
@Flole998 can you rebase this branch against the latest master? Then I can do a test drive.

jancborchardt:
> During setup the fingerprint the user has to enter a password which acts as a failback. It should be offered after the third false attempt @jancborchardt

Sounds good.

> What is the decision about allowing both, fingerprint and pin?

Nah, I would handle it like Android phone unlock which uses one main method, right?

AndyScherzinger:
> Nah, I would handle it like Android phone unlock which uses one main method, right?

Both actually, if you use fingerprint, then PIN is the backup

tobiasKaminsky:
> >   Nah, I would handle it like Android phone unlock which uses one main method, right?
> Both actually, if you use fingerprint, then PIN is the backup

Yes, but the pin on android should then we should use the account password and not use the 4-digit pin as a fallback, agreed?

So, to summarize:
- fingerprint disables our 4-digit pin and vice vera
- failback after three wrong fingerprints: use account password

AndyScherzinger:
@tobiasKaminsky sounds fine to me. What I don't know is which PIN would actually be the fallback since I never implemented fingerprint, so it might well be the OS PIN?

tobiasKaminsky:
I just had a quick look into the API, but I do not see any generic fallback system, so I expect that we can do this on ourselves, prompting for the account password after the third false attempt.

AndyScherzinger:
> prompting for the account password after the third false attempt.

sounds good to me

Flole998:
I am somehow not able to rebase this PR, either someone else tries this or this PR can be closed. I am not spending anymore time on this, it took almost 2 months now since this PR was created, that's definitely too long for a simple change that was already discussed in an issue before.

AndyScherzinger:
I also did push the branch to the Nc repo as a copy, https://github.com/nextcloud/android/tree/fingerprint

Flole998:
@AndyScherzinger I think it is fine that after 5 tries it locks itself, even though this was not wanted during development.

I also did a check of the rebased work and for me it is also working fine. Therefor I also think that this could be merged now.

src/main/res/values-de-rDE/strings.xml
```diff
@@ -596,7 +596,6 @@
     <string name="welcome_feature_3_text">Schützen Sie Ihre Fotos</string>
 
     <string name="whats_new_skip">Überspringen</string>
-
```
Flole998: This might cause an issue because of Transifex. The line should not be removed.

Flole998:
@AndyScherzinger Yeah that seems to be a leftover from the previous layout and can safely be removed.

tobiasKaminsky:
According to https://github.com/nextcloud/android/pull/624#issuecomment-291937183 I am fine with this approach.
I will test and review it tomorrow.

@mario are you fine with this?

mario:
Will test tomorrow

AndyScherzinger:
@mario @tobiasKaminsky the PR has been rebased to latest master from this morning 🚀

src/main/java/com/owncloud/android/authentication/PassCodeManager.java
```diff
@@ -112,4 +125,18 @@ private boolean passCodeIsEnabled() {
         return (appPrefs.getBoolean(PassCodeActivity.PREFERENCE_SET_PASSCODE, false));
     }
 
+    private boolean fingerprintShouldBeRequested(){
+        if ((System.currentTimeMillis() - mTimestamp) > PASS_CODE_TIMEOUT &&
+                mVisibleActivitiesCounter <= 0
+                ){
```
tobiasKaminsky: can be in fewer lines.
Line length should be max 120.

src/main/java/com/owncloud/android/ui/activity/FingerprintActivity.java
```diff
@@ -0,0 +1,356 @@
...
+     */
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.fingerprintlock);
+    }
+
+    private void startFingerprint() {
+        TextView fingerprinttext = (TextView) findViewById(R.id.scanfingerprinttext);
+
+        FingerprintManager fingerprintManager = (FingerprintManager) MainApp.getAppContext().getSystemService(Context.FINGERPRINT_SERVICE);
```
tobiasKaminsky: line wrap at 120

src/main/java/com/owncloud/android/ui/activity/FingerprintActivity.java
```diff
@@ -0,0 +1,356 @@
...
+     *
+     * @param savedInstanceState    Previously saved state - irrelevant in this case
+     */
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.fingerprintlock);
+    }
+
+    private void startFingerprint() {
+        TextView fingerprinttext = (TextView) findViewById(R.id.scanfingerprinttext);
```
tobiasKaminsky: camelCase

src/main/java/com/owncloud/android/ui/activity/FingerprintActivity.java
```diff
@@ -0,0 +1,356 @@
...
+        FingerprintManager fingerprintManager = (FingerprintManager) MainApp.getAppContext().getSystemService(Context.FINGERPRINT_SERVICE);
+
+        if (ActivityCompat.checkSelfPermission(this, Manifest.permission.USE_FINGERPRINT) != PackageManager.PERMISSION_GRANTED) {
+            return;
+        }
+        KeyguardManager keyguardManager = (KeyguardManager) getSystemService(KEYGUARD_SERVICE);
+
+        if (!keyguardManager.isKeyguardSecure()) {
+            return;
+        }else{
```
tobiasKaminsky: whitespaces between } and else
you can also use code formatting by android studio for new files. But please do not use it with new files as this will result in many changes.

src/main/java/com/owncloud/android/ui/activity/FingerprintActivity.java
```diff
@@ -0,0 +1,356 @@
...
+            return;
+        }
+
+
+        try {
+            keyStore.load(null);
+            keyGenerator.init(new
+                    KeyGenParameterSpec.Builder(KEY_NAME,
+                    KeyProperties.PURPOSE_ENCRYPT |
+                            KeyProperties.PURPOSE_DECRYPT)
```
tobiasKaminsky: same here, fewer lines with meaningful new lines will make it easier to read

src/main/java/com/owncloud/android/ui/activity/FingerprintActivity.java
```diff
@@ -0,0 +1,356 @@
...
+            return true;
+        } catch (KeyPermanentlyInvalidatedException e) {
+            return false;
+        } catch (KeyStoreException | CertificateException | UnrecoverableKeyException | IOException | NoSuchAlgorithmException | InvalidKeyException e) {
+            return false;
+        }
+    }
+
+
+    private void fingerprintresult(boolean fingerok) {
```
tobiasKaminsky: camelcase

src/main/java/com/owncloud/android/ui/activity/FingerprintActivity.java
```diff
@@ -0,0 +1,356 @@
...
+        this.update(MainApp.getAppContext().getString(R.string.fingerprint_unknown), false);
+    }
+
+
+    @Override
+    public void onAuthenticationSucceeded(FingerprintManager.AuthenticationResult result) {
+        this.update("Fingerprint Authentication succeeded.", true);
+    }
+
+
```
tobiasKaminsky: replace double empty lines with only one

src/main/java/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -191,7 +238,7 @@ public boolean onPreferenceClick(Preference preference) {
             }
         });
 
-        PreferenceCategory preferenceCategory = (PreferenceCategory) findPreference("more");
+        preferenceCategory = (PreferenceCategory) findPreference("more");
```
tobiasKaminsky: re-assigning an existing category? First one should be on its own, called "detailCategory" or similar.

AndyScherzinger:
@mario @tobiasKaminsky I fixed all code review comments you had so far, so you can do a final check/review :) - Hope that is fine with you @Flole998 :)

mario:
@AndyScherzinger I will review & merge today. Can you please open all the relevant issues in the mean time? Thanks.

AndyScherzinger:
@mario thanks! :) I have added the enhancements/alignments in #859

mario:
@Flole998 thank you for your contributions! :) We hope you'll continue working on this and other things! :)

PR #609, opened 2017-01-31
Filtering improvements

mario:
This improves filtering:

- add rotation support
- improves performance in case of many files (results limited to 100)
- no more duplicates
- add a few usability improvements (like hiding FAB & footer when searching)

PR #661, opened 2017-02-15
New filtering magic

mario:
This is basically a rewrite of the original PR #609 - unfortunately, I got too tangled into the other one and couldn't figure out what's happening where, so decided to start from scratch. This is the result.

Same functionality, just simpler and working nicer.

Gotta do one more thing on this PR, but that's gonna be done in an hour or so.


mario:
Ready, closing the other one. Thanks @AndyScherzinger and @tobiasKaminsky in advance!

mario:
@jancborchardt was just being silly I'm sure :P Since all the icons in the empty views are light blue,  I believe the search icon in the search empty view should also be light blue and not grey! :)

jancborchardt:
@mario as I said in the other issue: only folders are blue. The normal icons are grey in the empty view. Favorites are yellow. As in the web interface.

mario:
OK @jancborchardt. @AndyScherzinger, please update the branch with grey icons instead of blue - thanks! :)

jancborchardt:
Things are not just arbitrarily colored. And the search icon is only white or black (for light themes) or grey (in emptycontent views). ;)

jancborchardt:
@mario am busy at a trade fair atm – the icon is in https://github.com/nextcloud/server/blob/master/core/img/actions/search.svg :)

tobiasKaminsky:
BUG: filtering in a subfolder

content of a subfolder
screenshot_1487180692

filtering for 123 is showing folders many times and wrong file?:
screenshot_1487180700

clicking x leads to another folder
screenshot_1487180707

tobiasKaminsky:
Rejected
(mainly writing this to test the new pullaprove setting)

mario:
@tobiasKaminsky I will check all the bugs but your concern about the wrong file is wrong - the file indeed contains "123" in its filename.

I also don't understand the "X" thing? How to reproduce?

tobiasKaminsky:
Is was filtering in folder "11" (first screenshot) there is no IMG_20160401_123...
X is closing filter input query on UI.

mario:
@tobiasKaminsky this is ready for re-review.

mario:
@AndyScherzinger @tobiasKaminsky can either of you replace the icon with a gray one, and the referencing line of code for me please? :)))

I would appreciate it very much! :)

Otherwise I think you should hurry up with reviews:P

tobiasKaminsky:
BUG:

I have this folder structure:
filtering in 11, with this content:
 - 123 (folder)
   - IMG_20160925_105934.jpg
   - IMG_20160925_124718.jpg
 - 1.jpg
 - IMG_6507.jpg

filtering for "1" results in
- 11 (it finds the current folder as result)
- 123 👍 
- 1.jpg 👍 
- IMG_20160925_105934.jpg 👎 (as it is in a subfolder)
- IMG_20160925_124718.jpg 👎 (as it is in a subfolder)

initial:
screenshot_1487346436

filtering for "1"
screenshot_1487346443

tobiasKaminsky:
BUG:

Search for something that does not exist, should be "no results found".
screenshot_1487346592

tobiasKaminsky:
I currently suspend the testing as there seems to be still major problems, maybe depending on the rewrite?
Please test it for yourself extensively. At least the bugs I have here and in #609 should be solved before me or @AndyScherzinger should check again.

mario:
So I tested this before I asked you to review and it worked ... what in the world happened...

mario:
Dunno where are my commits o.O But I've made new ones that work.

tobiasKaminsky:
BUG: Same folder as above, filtering again for "1" leads now to 
screenshot_1487361311

but it should also show the folder "123".

tobiasKaminsky:
Just an idea, do we want to name it "filter" instead of search? in dolphin (KDE file manager) there is also a distinction between search and filter.
When we add real search, it should renamed to search, obviously.

tobiasKaminsky:
When I entered a filter query and I click on the "left arrow" the file list is moved to the parent folder, which is the normal behaviour of the "left arrow". But while I am in "filter mode" either the "left arrow" should be invisible (preferred) or act as removing the filter string (like the x button on the left).
Currently it is very non-intuitive.

mario:
Ah so it's missing the folder - will investigate.

I wouldn't do any renames, but I'll look into the back arrow thing.

tobiasKaminsky:
> I wouldn't do any renames

Just saying "no" in a discussion, instead of explaining why, is not very productive, especially when there are reasons for the other opinion.

mario:
1.5 will feature real "search" as you call it, and will technically be a combination of "filtering" and "search", yet will be called "search" because that's what user is more familiar with.

I think we should avoid changing something that we'll change in 1.5 anyway.

tobiasKaminsky:
Good point, just thought a user maybe expects a real (deep) search.
But then leave it the way 👍

mario:
@tobiasKaminsky so I thought about your back arrow thing, and here are my conclusions:

- we have X for clearing the query
- therefore, I didn't want to have two ways to do one action
- so, the back button and the home button now just close the search (and therefore clear the query as well)
- that leaves you in the current folder

Also, the lack of folders in filtering results is now fixed. (as soon as I push)

AndyScherzinger:
> @AndyScherzinger @tobiasKaminsky can either of you replace the icon with a gray one, and the referencing line of code for me please? :)))

DONE! :)

mario:
@AndyScherzinger thank you, now test and approve :P Same for two other PRs targeting 1.4.2 :P

AndyScherzinger:
device-2017-02-18-153936

* Is it on purpose that the search shows the hamburger menu in the root folder
* why is the overflow menu again gone?

mario:
@AndyScherzinger 
a) didn't really think about it, I can make it a back arrow. Thoughts?
b) that's intended - no switching between views when you search. Overflow menu appears once you long click on one of the files.

tobiasKaminsky:
> Overflow menu appears once you long click on one of the files.

I guess @AndyScherzinger means not the overflow menu of a file, but the overflow menu for the folder view. But, as we discussed, this is, intended.

tobiasKaminsky:
"Bug"/ENHANCEMENT:
If I filter something, then open one item of the filtered list (e.g. show details, or view an image) and then go back the folder list is not filtered any longer.
If this is too much work @mario then I create a new issue for this and we add it in the next version.

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -369,7 +376,7 @@ protected void onAccountSet(boolean stateWasRecovered) {
             if (!stateWasRecovered) {
                 Log_OC.d(TAG, "Initializing Fragments in onAccountChanged..");
                 initFragmentsWithFile();
-                if (file.isFolder()) {
+                if (file.isFolder() && TextUtils.isEmpty(searchQuery)) {
```
tobiasKaminsky: To distinguish the, later implemented, search from the current filter, all search related things (currently only the UI) should be called "search", but all other should be named "filter".

src/com/owncloud/android/ui/adapter/FileListListAdapter.java
```diff
@@ -443,47 +454,79 @@ public void setSortOrder(Integer order, boolean ascending) {
...
+
+        @Override
+        protected FilterResults performFiltering(CharSequence constraint) {
+            FilterResults results = new FilterResults();
+            Vector<OCFile> ocFileVector = new Vector<>();
+
+            if (!TextUtils.isEmpty(constraint)) {
+                for (int i = 0; i < mFilesAll.size(); i++) {
+                    OCFile currentFile = mFilesAll.get(i);
+                    if (currentFile.getRemotePath().replace(currentFile.getFileName(), "").equals(currentDirectory.getRemotePath()) &&
```
tobiasKaminsky: max 120 line length

src/com/owncloud/android/ui/adapter/FileListListAdapter.java
```diff
@@ -443,47 +454,79 @@ public void setSortOrder(Integer order, boolean ascending) {
...
+        }
+        return mFilesFilter;
+    }
+
+    private class FilesFilter extends Filter {
+
+        @Override
+        protected FilterResults performFiltering(CharSequence constraint) {
+            FilterResults results = new FilterResults();
+            Vector<OCFile> ocFileVector = new Vector<>();
```
tobiasKaminsky: maybe "filteredFiles"?

src/com/owncloud/android/ui/fragment/ExtendedListFragment.java
```diff
@@ -151,21 +161,111 @@ public boolean isGridEnabled(){
...
+
+
+        LinearLayout searchBar = (LinearLayout) searchView.findViewById(R.id.search_bar);
+        searchBar.setLayoutTransition(new LayoutTransition());
     }
 
-    public boolean onQueryTextChange(String query) {
-        mAdapter.filter(query);
+    public boolean onQueryTextChange(final String query) {
+        performSearch(query, false);
```
tobiasKaminsky: performFilter to distinguish. Also later we will first do a filtering first, and then a real search via search api.

src/com/owncloud/android/ui/fragment/ExtendedListFragment.java
```diff
@@ -415,15 +523,17 @@ public void setMessageForEmptyList(@StringRes int headline, @StringRes int messa
...
-     * Set message for empty list view.
-     */
-    public void setEmptyListMessage() {
-        setMessageForEmptyList(
-                R.string.file_list_empty_headline,
-                R.string.file_list_empty,
-                R.drawable.ic_list_empty_folder
-        );
+    public void setEmptyListMessage(boolean isSearch) {
+        if (!isSearch) {
```
tobiasKaminsky: Please change the if else content for better readability. 
It is easier if the test is not negate: if (isSearch) { }

src/com/owncloud/android/ui/interfaces/ExtendedListFragmentInterface.java
```diff
@@ -0,0 +1,28 @@
+/**
+ * Nextcloud Android client application
+ *
+ * @author Mario Danic
+ * Copyright (C) 2017 Mario Danic
```
tobiasKaminsky: additional Nextcloud GmbH?

tobiasKaminsky:
@mario CR is done. If the bugs are vanished, I will test again. :+

src/com/owncloud/android/ui/activity/FileDisplayActivity.java
mario: Search will be a combination of local filtering + API call for search. Technically, it's still search. Even filtering is searching. The view where you enter query is called SearchView, so the query should be called searchQuery :)

src/com/owncloud/android/ui/fragment/ExtendedListFragment.java
mario: See my previous comment. It will be *one* call for the app.

mario:
@tobiasKaminsky commented. As for your request - once you do filtering/search the search view is closed and search query null-ified for a reason - you've found what you wanted. This has been discussed with @jancborchardt.

tobiasKaminsky:
> once you do filtering/search the search view is closed and search query null-ified for a reason - you've found what you wanted. This has been discussed with @jancborchardt.

I have found what I wanted? Well, then it is strange at least for me. Real scenario I had today: I have a folder with lots of funny images and I wanted to show a gif to some people. So I filtered for "gif" and as there are several nearly identical, I opened one (out of 5) and looked at it. Of course it was the wrong one ;-) So to select one of the remaining 4 gifs, I had to filter again...

mario:
My commits seem to disappear recently. Pushed now, and updated formatting and variable names.

There's another bug with the back arrow that I'm working on (cc @AndyScherzinger), so will be ready tomorrow or something.

mario:
Should be all fine now. I'm open to opinions about keeping the filter open cc @AndyScherzinger

AndyScherzinger:
>    Overflow menu appears once you long click on one of the files.

> I guess @AndyScherzinger means not the overflow menu of a file, but the overflow menu for the folder view. But, as we discussed, this is, intended.

@mario @tobiasKaminsky what is intended? To not show the menu or the folder during filter? So no "switch to grid view" or "change sorting" while filtering? Both seem helpful to me e.g. when searching for images you might want to sort differently or switch to grid view.

mario:
@AndyScherzinger @tobiasKaminsky menu entry is back. :) Be happy :P

AndyScherzinger:
@mario not happy ;) You just need to fix the layout file, like we did in #609 ... https://github.com/nextcloud/android/pull/609/commits/5006e310b0a005120a2966a1204f6eb23543ba91 (!)

mario:
@AndyScherzinger no, that "fix" is flawed aka doesn't work. (it fixes the layout, but breaks restoring search state, etc etc)

AndyScherzinger:
Oh, didn't know that! Thanks for the mention :)

ekeitho:
FileListListAdapter::swapDirectory will hit an outOfMemoryException since if I click search, write something, hit back. Repeat this, swapDirectory keeps getting called and mFilesAll keeps appending the original list of items and is never cleared.

So currently I have only 5 files and doing this 17 times, I now have 85 items in mFilesAll.

screen shot 2017-02-19 at 6 25 09 pm

mario:
Thanks @ekeitho - same problem exists in master, but since this PR targets 1.4.2 anyway, I have merged it in! :)

ekeitho:
No problem @mario glad to help! :)

tobiasKaminsky:
Is this intended to have the current folder beside the search query? At least it is something new...?

screenshot_1487620092

tobiasKaminsky:
screenshot_1487620925

mario:
@tobiasKaminsky should be fixed.

tobiasKaminsky:
screenshot_1487621415

tobiasKaminsky:
👍

AndyScherzinger:
Not sure if these are considered bugs (fab is shown, folder name is shown and search string is kind of in the center and also if you pull-to refresh it will remove your filtering after the refresh is done). All tested on an Nexus5X Emulator:
device-2017-02-20-224827

AndyScherzinger:
👍 since all above mentioned things are intended 😸

tobiasKaminsky:
👍  again

mario:
@AndyScherzinger you just need to approve once again :)

AndyScherzinger:
👍 merge?

jancborchardt:
Sorry, just reading through backlog of issues …

> once you do filtering/search the search view is closed and search query null-ified for a reason - you've found what you wanted. This has been discussed with @jancborchardt.

@mario hm … don’t fully remember discussing this. I tend to agree with Tobias here that when you go back from viewing a file then the state should be the same, filtered. Otherwise it’s confusing because the view you are going back to is not the same.


(And just a sidenote:)

> Otherwise I think you should hurry up with reviews:P

Keep in mind basically everyone is working on this in their spare time. Even though you might have meant it jokingly, let’s keep it chill ok? :)

mario:
You're the one who brought this up. But I'll "fix" this in a PR later today.

jancborchardt:
@mario then sorry for confusing it. :( That’s why it’s important to test and reevaluate what we thought was good in the design phase. :)

tobiasKaminsky:
@mario as I was the last few days working on my flat, do you have fixed what @jancborchardt has mentioned? Otherwise I can open an issue just for tracking)

mario:
@tobiasKaminsky no, Andy agreed it's fine to leave this for 1.5 as it would create a lot of additional work if I did it now (due to two other PRs that touch search).

mario:
There's an issue for that, I've opened it.

PR #670, opened 2017-02-20
fix potential out of memory exception

ekeitho:
in regards to #661


mario:
Hi - this seems to be a problem in master as well. While I prefer creating a new Vector<> for performance reasons, this will do for now! :)

Thank you very much! :)

Issue #694, opened 2017-02-28
App crashes when loading an huge directory


taaem:
### Actual behaviour
I have a directory on my Nextcloud instance with ~3000 pictures (all in the same dir), when I open that folder from the app it takes ages and nothing happens. After a while I tried to refresh via pull down, I don't think that is actually related to the crash. Then the app crashes and when trying to open the app again it straight crashes again until I reboot.

### Expected behaviour
Load the images.
 
### Steps to reproduce
1. Have a folder with many images
2. Open from the Nextcloud app
(3. Pull to refresh)

### Environment data
Android version: 7.1.1

Device model: OnePlus X E1003

Stock or customized system: Customized

Nextcloud app version: 23.2.2017 Nightly (latest from fdroid)

Nextcloud server version: 11.0.2

#### Web server error log
```
192.168.1.145 - [28/Feb/2017:12:41:39 +0100] "PROPFIND /remote.php/webdav/SofortUpload/WhatsApp%20Images/ HTTP/1.1" 207 882 "-" "Mozilla/5.0 (Android) nextcloud-android/20170223"
192.168.1.145 - [28/Feb/2017:12:41:51 +0100] "PROPFIND /remote.php/webdav/SofortUpload/WhatsApp%20Images/ HTTP/1.1" 207 3031975 "-" "Mozilla/5.0 (Android) nextcloud-android/20170223"
192.168.1.145 - [28/Feb/2017:12:45:41 +0100] "GET /ocs/v1.php/apps/files_sharing/api/v1/shares?path=%2FSofortUpload%2FWhatsApp+Images%2F&reshares=true&subfiles=true HTTP/1.1" 200 0 "-" "Mozilla/5.0 (Android) nextcloud-android/20170223"
```

#### Nextcloud log (data/nextcloud.log)
```
02-28 12:42:09.752  6526 11087 I ReadRemoteFolderOperation: Synchronized /SofortUpload/WhatsApp Images/: Operation finished with HTTP status code 207 (success)
02-28 12:42:09.753  6526 11087 D RefreshFolderOperation: Synchronizing /SofortUpload/WhatsApp Images/
02-28 12:42:09.759  6526 11087 D RefreshFolderOperation: Remote folder /SofortUpload/WhatsApp Images/ changed - starting update of local data 
02-28 12:43:01.871  6526 11087 D FileDataStorageManager: Sending 3839 operations to FileContentProvider
[...]

PR #704, opened 2017-03-02
Taaem fix

mario:
This fixes an issue with really really big folders, and has been tested by @taaem.

Ideally in the future, we'll use persistent storage (sqlite) for this instead of the Data Holder I just wrote, but this works.

IMHO, important enough to be in 1.4.2.

### DIFF
--- a/src/com/owncloud/android/operations/RefreshFolderOperation.java
+++ b/src/com/owncloud/android/operations/RefreshFolderOperation.java
@@ -526,8 +526,8 @@ public class RefreshFolderOperation extends RemoteOperation {
         }
 
         DataHolderUtil dataHolderUtil = DataHolderUtil.getInstance();
-        String dataHolderItemId;
-        dataHolderUtil.save(dataHolderItemId = dataHolderUtil.nextItemId(), result);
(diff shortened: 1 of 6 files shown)


taaem:
For reference issue #694. Without this the app gets unusable and is force closing all the time.

mario:
This would resolve issue #694

AndyScherzinger:
👍 Looks good

One question though, since I have never seen this in the wild. any reason for writing it this way?
```
String dataHolderItemId;
dataHolderUtil.save(dataHolderItemId = dataHolderUtil.nextItemId(), result);
```
compared to?
```
String dataHolderItemId = dataHolderUtil.nextItemId()
dataHolderUtil.save(dataHolderItemId, result);
```

mario:
> One question though, since I have never seen this in the wild. Any reason for writing it this way?

Not really, left-overs from testing. So please re-approve :)

AndyScherzinger:
👍 looking good 🌹

tobiasKaminsky:
👍 
Ready for 1.4.2 as it is a bug fix

PR #709, opened 2017-03-03
[WiP] Upgrade to Gradle 3.3, Build Tools 25.0.2

AndyScherzinger:
Work in progress to start all checks and test drive against Drone Images

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,7 +29,7 @@ configurations.all {
 }
 
 ext {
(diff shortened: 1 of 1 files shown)


mario:
@AndyScherzinger can we kill this PR?

AndyScherzinger:
Kind of still valid (build tools...) but let's kill it since we are on 25.x so that is fine for now :)

PR #728, opened 2017-03-08
Nightly only: remove gcm

tobiasKaminsky:
To get nightly working again on fdroid, I have removed GCM and all pending uploads will be deleted on reboot.


mario:
👍

PR #732, opened 2017-03-11
Remove tabs and line breaks from string source

M1dgard:
They were shown in Transifex and appeared to be just artefacts of word wrap and indentation of the XML source, so this commit removes them.

I noticed while translating that the source messages on Transifex are not exactly the same as those in the file I edited. Should the messages be edited in `en-rGB` or `en-rUS` as well perhaps?


AndyScherzinger:
👍 Thanks for the fix @M1dgard! No fix shoudl only be done in the base strings.xml like you did, all other resource files get automatically pulled from transifex only.

@mario @tobiasKaminsky can we merge this?

mario:
👍

mario:
Thanks for the contribution @M1dgard !

AndyScherzinger:
sweet! 🎉

PR #679, opened 2017-02-24
Branding

mario:
This PR adds support for Google Firebase Analytics in a custom build, and changes project structure to allow easier branding.

PR #767, opened 2017-03-22
Firebase

tobiasKaminsky:
As discussed this is a clean start, based on #679, for analytics/firebase.
This will NEVER be merged to the official app, but instead will be used for custom builds.


tobiasKaminsky:
@mario I guess this is obsolete? If so, please close.

mario:
Yup, #901 handles this.

PR #771, opened 2017-03-23
Add support for web login

mario:
Just needs updating for new structure. Thanks @AndyScherzinger

### DIFF
--- a/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -8,16 +8,16 @@
  * Copyright (C) 2012  Bartek Przybylski
  * Copyright (C) 2015 ownCloud Inc.
  * Copyright (C) 2017 Mario Danic
- * <p>
+ *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2,
  * as published by the Free Software Foundation.
- * <p>
+ *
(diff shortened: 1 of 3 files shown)


mario:
@AndyScherzinger can you update this and merge please?

AndyScherzinger:
@mario sure, how would I have to resolve the conflict?

> **Conflicting files**
src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java

AndyScherzinger:
👍 code is fine by me, default internal app login successfully tested -> merging

aleister09:
@mario I found an issue 

When you're adding another account with Web-Login and introduce the same account the AuthenticatorActivity is not closing

mario:
@aleister09 I'm not sure I understand the second part of the sentence?

The first is known and is because only single-account is enabled (as desired).

mario:
(You should not be able to add another account anyway?)

aleister09:
I'm able to add another account but if I add the same account the AuthenticatorActivity is not closing 

I saw the code and it is because the account is already registered but the app doesn't show a message to notify that the account is already registered,

mario:
Interesting. Will evaluate. I assume you come to the web view screen via the change password for the current account option?

aleister09:
No, I enable the Web-Login option so in my first login and when I add another account I go to web view screen

MorrisJobke:
As I understood it should be for now only one login.

PR #772, opened 2017-03-23
Fix snapchat etc

mario:
This fixes Snapchat auto upload and various other folders.


AndyScherzinger:
👍 code wise fine and tested by community members

Issue #770, opened 2017-03-23
1.4.2 Video upload folder destination incorrect 


AndyScherzinger:
see https://help.nextcloud.com/t/1-4-2-video-upload-folder-destination-incorrect/10432

There is an option to have it upload to ```<folder>/<year>/<month>``` (folder being user selected) this is broken in 1.4.2

In this case my destination folder for my videos is "video".

In 1.4.2. it creates a folder called ```<folder>+<year>/<Month>```  i.e.  ```video2017/03```  when the path should have been ```video/2017/03```

-----

I haven't tested this personally but know that we fixed this issue pre-1.4.2 final for images, but am not sure if this is then also automatically fixed for video uploads (but would think so).

AndyScherzinger:
probably fixed by #774 - waiting for the user feedback on the Android version in use.

DaveBurkhardt:
![screenshot_2017-03-19-17-04-25](https://cloud.githubusercontent.com/assets/23338339/24314429/8c1aa13c-10af-11e7-88d6-14792fdd4b1f.png)
![screenshot_2017-03-24-16-27-38](https://cloud.githubusercontent.com/assets/23338339/24314430/8dca2868-10af-11e7-856d-89d83a2d560b.png)

Validation images.

mario:
@DaveBurkhardt thanks. weather doesn't help though :D

mario:
This will either be fixed in the next release or we'll replace instant upload with auto upload which doesn't have separation between images and videos. Thanks for reporting.

DaveBurkhardt:
@mario haha, sorry!  Fixed above

mario:
@DaveBurkhardt all good! :)

AndyScherzinger:
PR has been merged, to be shipped with 1.5.0

PR #774, opened 2017-03-24
Fix video upload path

mario:
Only applies to instant upload on Android 4 and 5. Considering we might not have time to enable Auto Upload for 1.5, I think it's worth to have it.

### DIFF
--- a/src/main/java/com/owncloud/android/utils/FileStorageUtils.java
+++ b/src/main/java/com/owncloud/android/utils/FileStorageUtils.java
@@ -200,7 +200,7 @@ public class FileStorageUtils {
         if (com.owncloud.android.db.PreferenceManager.instantVideoUploadPathUseSubfolders(context)) {
             subPath = getSubpathFromDate(dateTaken);
         }
-        return uploadVideoPath + subPath + (fileName == null ? "" : fileName);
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@mario what gets fixed here, just instant upload mentioned on Play:
* ```\dirFile``` instead of  ```\dir\File```

we also have a path building issue with auto upload and videos! --> #770 
* ```\<folder><year>/<Month>``` instead of ```\<folder>\<year>\<Month>``` : ```\video2017\03``` instead of ```\video\2017\03```

AndyScherzinger:
@mario Did you test this one?

mario:
No, but I believe the original report is wrong as auto upload uses
different method and does NOT separate videos and images.

Also, the user is using Android 5 so no auto upload.

mario:
Hmm just re-read what you wrote. Will check, but same method is used for
video and images

AndyScherzinger:
@mario let me ask on the forum, since the user is also just mentioning videos specifically with "choosing a path for my videos" which might also mean instant upload and would also fit the same line of code you fixed.

mario:
@AndyScherzinger he said 5.0 :)

AndyScherzinger:
👍 approved :D
@mario sweet, than the fix should work :)

AndyScherzinger:
merging with one approval due to vacation of Tobias

Issue #781, opened 2017-03-27
add icons to website/twitter on user info screen


AndyScherzinger:
As described in https://github.com/nextcloud/android/pull/585#issuecomment-289394805 also 

* use icons for the entries: website and twitter
* remove http:// and https:// from the shown URL
* make sure the twitter handle starts with an ```@```

jancborchardt:
Cool! :) 

@schiessle maybe on the server we should also make sure that the Twitter handle starts with an `@` and also for the placeholder use `@…` to make sure people know which format we want it in.

AndyScherzinger:
PR has been opened for review: #782

mario:
Closing, merged

PR #585, opened 2017-01-26
Show user information

mario:
This solves issue #554. Let me know if you have any questions.

P.S. Please don't complain about the Gradle upgrade, thank you :)

PR #782, opened 2017-03-27
add icons to website/twitter on user info screen, cleanup - #781 

AndyScherzinger:
PR implementing #781:

> As described in https://github.com/nextcloud/android/pull/585#issuecomment-289394805 also 
>* use icons for the entries: website and twitter
>* remove http:// and https:// from the shown URL
>* make sure the twitter handle starts with an ```@``` 

@jancborchardt @mario please review, changes have been made according to the 3 points specified. Here is the screenshot how it looks with this PR:
device-2017-03-27-174325


src/main/java/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -150,6 +154,50 @@ public static String unixTimeToHumanReadable(long milliseconds) {
...
+     * @param url to be beautified url
+     * @return beautified url
+     */
+    public static String beautifyURL(String url) {
+        if (url == null) {
+            return "";
+        }
+
+        if (url.length() >= 7) {
+            if (url.substring(0,7).equalsIgnoreCase(HTTP_PROTOCOLL)) {
```
mario: Fix this and I approve :)

src/main/java/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -150,6 +154,50 @@ public static String unixTimeToHumanReadable(long milliseconds) {
...
+                return url.substring(HTTP_PROTOCOLL.length());
+            }
+        }
+
+        if (url.length() >= 8) {
+            if (url.substring(0,8).equalsIgnoreCase(HTTPS_PROTOCOLL)) {
```
mario: This too.

src/main/java/com/owncloud/android/utils/DisplayUtils.java
mario: Btw. why not use startsWith?

AndyScherzinger: because "Https://nextcloud.com" wouldn't be caught (case sensitive) so I went super bulletproof ;)

mario: toLowerCase :P

But fine. Just fix ifs :)

AndyScherzinger: > toLowerCase 

won't work either... "Https://nextcloud.com/Downloads" might not be the same as "https://nextcloud.com/downloads" while for simply detecting a protocol prefix it would also work.

src/main/java/com/owncloud/android/utils/DisplayUtils.java
AndyScherzinger: both done :)

AndyScherzinger:
@mario fixed the nested if statements

jancborchardt:
@AndyScherzinger I would even cut the words »Website« and »Twitter«. :)

jancborchardt:
(Also: wow this was quick! Good to see you remotely participate in the hackweek. ;D)

AndyScherzinger:
> I would even cut the words »Website« and »Twitter«. :)

sure thing, should it then be headline style as in larger font size and black coloring or like it is secondary text style with smaller font size and grey coloring (also as in this is considered secondary information)?

jancborchardt:
@AndyScherzinger how about we don’t separate the blocks (cause it will look a bit strange if you have just one website or one Twitter separately listed there) but use grey coloring? :)
(Font size should be the same though, coloring change is enough.)

Quick mockup:
android-profile-background
(On the right I experimented with removing the borders – that looks too un-interactive then. But also, can you move the icons & text so it’s flush with the name in the header? (like it’s on the right image – just a detail :)

Issue #486, opened 2016-12-29
Auto upload: wifi/charging only for specific filetypes


tobiasKaminsky:
Back on old instant upload we distinguished between upload on type and "situation"
- video / wifi only
- video / charging only
- images /wifi only
- images / charging only
all settings could be changed interdependently.

Now with auto upload it is "folder based" and only options are
- Upload pictures via wifi only (which is misleading as it applies for all file types)
- charging only

So my idea is to extend this to have a selectable pre-defined "mime-type" list:
- pictures
- videos
- other
The menu setting would then be 
- Upload via wifi only
- Upload when charging only
Each setting has a second option line where the selected items are be shown:
- Upload via wifi only: videos
- Upload when charging only: videos, other

@AndyScherzinger @eppfel

eppfel:
This sounds a little complex, especially because the auto Upload settings are already numerous. I would first think about the use-cases for the old options used in Instant Upload. Maybe it is more related to file size, than to file type?
And to achieve what you suggest one could easily separated file types in separate folders on their android system.

Anyway It would be a powerful feature to sync distinctive mime types, I just see bigger issues now. 

The misleading description about "pictures"  should be fixed independently.

AndyScherzinger:
Yes, this sounds like a meetup would help a lot! 👍

kvoit:
As I suggested and motivated in #587, I think the current possibilities are insufficient. I understand that the new auto upload feature is meant to be more general and I don't think we need complex rules. However, a size threshold for mobile upload would be help a lot: If file >10MB, wait for Wifi to upload. I don't think this would make it too complicated.

jmclaren7:
It seems to me that upload conditions based on file size would be a first step in addressing the issue of conserving data (upload only on wifi). However, it wont help keep media organized. Most of my users and myself organize videos in separate folders and that feature was a must have of the old app. 

I haven't made myself familiar with this project but maybe this concept would minimize the work needed:

- When scanning for media directories we should note ones that contain both photos and videos. 
- Display the list of media directories as we currently do (combined videos and photos). ie a folder called "Camera" would be listed.
- Append to the list any directories that contained both types but apply a filter to only work with photos and display on the list as "Camera (Photos only)".
- Append to the list any directories that contained both but apply a filter to only work with videos and display on the list as "Camera (Videos only)".

[...]

Issue #270, opened 2016-09-19
Sync on certain rules


tobiasKaminsky:
Sync images and videos during
- a certain period of time (week day and time)
- when connected to determined Wireless Network(s)
- etc.

All possible combinations should be allowed.

AndyScherzinger:
Hmm, not sure if it is worth the complexity being added to the app just to a the feature "generic task automation" what we could do though (did that for another app once) would be to write a tasker/locale integration and then you could basically do just anything :wink:

matzedings:
A Tasker plugin would be incredibly useful! Are there any plans to implement that, and if so, do you have a release estimation?

aronovgj:
+1 on tasker plugin. Actually just wanted to open an issue. Tasker plugins are also compatible with apps such as Automagic. And I constantly have problems with third party apps which offer syncing.

AndyScherzinger:
@aronovgj what would you like to trigger via the Tasker Plugin? Just asking to collect the use cases to be implemented.

aronovgj:
Syncing folders to or from NC or in both directions using filters by name, file type, regex, size, date and so on

It would also be nice to get information via the plugin about files or folders. For example if file x has changed or if a new file is in folder y sync it. Things like when and how often to sync and if on wifi or not can be determined by tasker without plugins

I might come up with further ideas later.

Bullnados:
Perhaps also move files i.e pictures to a folder by date, geolocates or still after one week, month etc. and rename folder to month, city/place,

tobiasKaminsky:
If someone wants to do this please discuss it first with us as I am currently unsure how to handle this.
(though the initial request is made by me, the real idea was sent via mail)

PR #797, opened 2017-04-03
Add further media types to auto upload

AndyScherzinger:
as discussed in #285 and numerous other issues, this PR adds further media types to the auto upload screen, starting with videos. For now video and images are split and shown twice for folders contain both e.g. the camera folder. We could also merge them into one (if we want that which is subject to discussion).

Audio is also on the agenda but I would leave it out for now until it is clear how to implement this. The reason for that is that we are querying the media store and then display all folders that contain images (and are visible to the media store, as-in indexed). So for music in case each album has it's won folder we would display all those folders which might be a lot while it does makes sense do handle it this way for video and images...

here is a screenshot of the work in progress and btw thumbnail generation and caching has been extended for video. If a format can't be read or the file is somehow corrupt we just show the standard icon:
device-2017-06-09-172344

**OPEN ISSUES**
- [x] add delete capability for custom folders (delete button on the bottom left in the folder preference dialog)
- [x] hide delete for non custom folders
- [x] hide delete for unpersisted custom folders
- [x] block save button in preference screen as long as not all necessary data has been set (grey out "save" text until set)
- [x] fix filter for local folder picker (upload activity re-use) to only show folders but no files
- [x] make Jobs aware of the folder type and only upload all file types for custom folders, video-only for video folders and image-only for image folders
- [x] make sure a job gets removed/terminated when a custom folder setup gets deactivated or deleted
- [x] rewrite auto upload
- [x] adapt to customer-hack-flag which hides the per-folder setting, which is needed for custom folders and also the flag needs to be ignored for custom folders since the user (IMHO) will like to have full control over this type of auto upload folder!
- [ ] add check to prevent several custom folder definitions which define syncs on different levels of one file tree since this will either lead to duplicate uploads or even errors (in case delete after upload)
- [x] check why successful uploads disappear from uploads view after visiting that screen a few times
- [x] lock files while uploading
- [x] show only current user uploads
- [x] automatically retry manual uploads
- [x] remove uploads screen menu
- [x] show "Setup custom folder" even if no media folders are detected
- [x] Make it "Set up a custom folder" where it says "Setup a custom folder" on the work in progress screen


src/main/java/com/owncloud/android/utils/BitmapUtils.java
```diff
@@ -111,6 +111,24 @@ private static int calculateSampleFactor(Options options, int reqWidth, int reqH
...
+     * scales a given bitmap depending on the given size parameters.
+     *
+     * @param bitmap the bitmap to be scaled
+     * @param px     the target pixel size
+     * @param width  the width
+     * @param height the height
+     * @param max    the max(height, width)
+     * @return the scaled bitmap
+     */
+    public static Bitmap scaleBitmap(Bitmap bitmap, float px, int width, int height, int max) {
```
AndyScherzinger: fixed via https://github.com/nextcloud/android/pull/797/commits/43a6f40ef5abebda42444db0328745c9a41a600f

mario:
Hi @AndyScherzinger. First of all nice work.

a) Any specific questions?
b) No need to really I think
c) Sure as we might want to operate on that later on - integer please (0, 1, 2, etc)

AndyScherzinger:
Thanks for the feedback @mario than I know how to proceed. 

1. Will keep video and images separately 
2. Will extend the db model to save the folder type 

After that maybe one of you two can then work with the folder type for images and video so no videos get uploaded for an image folder and vice versa. 
After that I'll then add custom folders but probably in a new PR. 

Fine with you @mario @tobiasKaminsky?

mario:
@AndyScherzinger I'm unsure if that's desirable - I would just upload whatever ends up being in the folder. But you tell me what you prefer cc @tobiasKaminsky

AndyScherzinger:
I am also fine either way it is just that people asked for the separation of images and video. If we don't want to do this then I need to merge folders like camera, since it shows up twice since it is found in both media stores

mario:
Yea, I was silly - just ignore me as I said :D Is late and I have a headache.

AndyScherzinger:
So not an issue we just need to decide it. If we split it than it'll work like before since probably on camera folders put both into one directory..

mario:
Like in your screenshot seems fine.

tobiasKaminsky:
The only drawback I currently see is, that an user (like m) who wants to have video and images uploaded in the same directory, both on wifi only, have to set it up twice...
So I still would go for an approach like I mentioned here https://github.com/nextcloud/android/issues/486#issuecomment-291078817 but this is also net perfect, especially for the space constraints @AndyScherzinger mentioned.

I fear that this will only satisfy few users and the rest is still "complaining". So we should (at least plan) a broader approach as now audio and non-media-files are missing, and we do not have a file sized rule.

mario:
I would maybe ignore audio anyway, and just go for "Photos", "Videos" and "Others" type, where custom folders would only have "Others".

AndyScherzinger:
@tobiasKaminsky This is the PR we were talking about which might (at some point) implement the custom folders depending on my time (or anyone else`s)

mario:
@AndyScherzinger so ... list of tasks? :D

AndyScherzinger:
Here is the list of open task for the moment to get this PR done:

- Implement action behind the "create custom folder" element
  - let the user choose the local folder with a picker
  - let the user choose the remote folder with a picker (and always appending the local folder, as it is done for images/videos to prevent collisions when sync will come in the future and a user might make the mistake to configure 2 local folders to the same remote folder with 2-way activated...)
  - let the user configure the upload behavior (showing the auto upload settings dialog)
- upon persistence of a new custom folder item, refresh the recyclerview after the item is persisted
- put the "star folder icon" in front of custom folders (images and video already have their icons, star folder icon 18dp is also already in the branch)
- **important**: implement image/video distinction/behavior since now e.g. for the camera folder, it is shown twice, 1 for images and 1 for video, which would lead to the ability to activate it twice for any folder that contains videos and images
- **discussion needed**: How do we deal with the scenario above? (the camera folder), at the moment you can now configure image/video separately, so you could then do images on cellular and video on wifi only. Open issue is with today's naming convention (local folder added on the remote side) images and video would still end up in the same folder. In my personal opinion this is just fine. It is basically 1 folder so it ends up in one folder (even though some users asked for separation). So I'd advice to keep it this way else we introduce a "custom folder choice per file type, size, date, whatever" which is wrong in my opinion since I don't consider this a "file app functionality" but a rule/workflow engine function which wouldn't be client side but server side.
- **discussion needed**: the "create custom folder" element is above the list at the moment and stays there. Imho it should be integrated into the list, to be "scrolled away" (or the appbar). A FAB would also work but  a) be a bit to prominent imho and b) hurt the design since it would be positioned in the same area as the actions of the items (three-dot and cloud). So for a start we should probably look into the first 2 options and choose the one easier to implement.

mario:
So 1 addition:

- apart from separating folders to image/video we should add a third type which is "other", just in case.

As for two points worth discussing:

- yes, I'm fine with them ending up in the same folder
- yup, scrolling away is fine - so it should basically be a recyclerview element

AndyScherzinger:
@mario would "other" be equivalent to "custom" or another type?

mario:
It would be "all other types", but yes - same as custom.

src/main/java/com/owncloud/android/datamodel/MediaFolder.java
```diff
@@ -27,6 +27,11 @@
  * Business object representing a media folder with all information that are gathered via media queries.
  */
 public class MediaFolder {
+    public static final Integer CUSTOM = 0;
+    public static final Integer IMAGE = 1;
+    public static final Integer VIDEO = 2;
+    public static final Integer IMAGE_VIDEO = 3;
```
AndyScherzinger: @mario atm we have custom, image, video and (to be deleted) image_video. Would that be sufficient?

mario: Yes, perfect. Basically if we have IMAGE from media providers, we'd split it to image, video and custom.

mario:
@AndyScherzinger do you think you'll have time to work on the open stuff for this PR?

src/main/java/com/owncloud/android/datamodel/MediaFolder.java
AndyScherzinger: Well, no ;) We have IMAGE from media provider (query), VIDEO from (different) media provider (query) and would then retrieve custom straight from our database 😁

AndyScherzinger: image and video is already implemented like described

AndyScherzinger:
> @AndyScherzinger do you think you'll have time to work on the open stuff for this PR?

I might yes but likely not this week, so the question is about the timeline of this PR. Maybe next week.

tobiasKaminsky:
After my bug hunting session I can also assist here :-D

AndyScherzinger:
> After my bug hunting session I can also assist here :-D

That would be awesome!

mario:
What happens if I drop docx file in there?

AndyScherzinger:
> What happens if I drop docx file in there?

It'll get uploaded! The implementation I did so far is "just UI". So you can see and activate video and image folders but the services don't care about the content (!)

mario:
Yea I meant ... how do I enable uploading docx (or any other file) in a
folder we get from media providers?

AndyScherzinger:
> Yea I meant ... how do I enable uploading docx (or any other file) in a folder we get from media providers?

When someone implements the file type filters for image/video, then you can't (and shouldn't)... 
In order to do that I think one should then define a customer folder (which won't have a type filter) and not use the media provider detected folders.

src/main/java/com/owncloud/android/datamodel/MediaFolder.java
```diff
@@ -30,7 +30,7 @@
     public static final Integer CUSTOM = 0;
     public static final Integer IMAGE = 1;
     public static final Integer VIDEO = 2;
-    public static final Integer IMAGE_VIDEO = 3;
+    public static final Integer LEGACY = 3;
```
AndyScherzinger: We can remove this one, since this feature is still in development so we do not have legacy data :)

mario:
- [X] Put star folder icon next to custom folders
- [ ] Implement migration path from legacy folders to new type of folders
- [X] Implement "Setup a custom folder" as RV item
- [ ] Implement custom folder setup procedure
- [ ] Rewrite auto upload

AndyScherzinger:
Awesome @mario ! :)

> Implement migration path from legacy folders to new type of folders

Since we "only" had ```IMAGE``` before I'd say the migration is really simple :D - In the DB Upgrade, where we add the type column, we just update all existing rows setting the value of the type-column to ```1```. Since they are all of type ```IMAGE```, so there is no legacy or anything, right?

mario:
I don't think the migration is THAT simple. I've set the default type to 3 (legacy) and that should be split out in two upon migration - if we have both video and image for that folder, enable them both and set the same settings as what legacy was set to.

So we will need a preferences entry and a custom method running on startup if this entry is not set.

mario:
@AndyScherzinger do you think you can do this migration magic? :)

AndyScherzinger:
Argh, you are right.... Oh this migration will be hell since we need to do this during upgrade time since users might not open the screen while during DB upgrade we would then have to call some media provider magic :( So the DB Upgrade code would reference some auto upload code 😢

mario:
I'm not sure I understand the issue? Migration should basically be triggered in onCreate of MainApp.

AndyScherzinger:
I'd say migration should be triggered wihtin the DB-version upgrade routine, would it not?

mario:
It can be, but it's not mandatory for sure.

AndyScherzinger:
It's not mandatory, that is true, while I'd vote for just putting it here: https://github.com/nextcloud/android/pull/797/files#diff-f9acf3921f3be4a83ad881a769f5231eR954

mario:
@AndyScherzinger ok, do it :) Go go go!

mario:
@AndyScherzinger I have implemented initial migration stuff. Could you please test it out and see if it works? Thanks!

AndyScherzinger:
@mario I will test it, yes!

AndyScherzinger:
@mario -> straight crash and burn (can't start the app anymore...):
```
05-05 17:26:37.210 31190-31190/? E/AndroidRuntime: FATAL EXCEPTION: main
Process: com.nextcloud.client, PID: 31190
java.lang.RuntimeException: Unable to create service com.owncloud.android.services.observer.SyncedFolderObserverService: java.lang.IllegalStateException: Couldn't read row 0, col -1 from CursorWindow.  Make sure the Cursor is initialized correctly before accessing data from it.
    at android.app.ActivityThread.handleCreateService(ActivityThread.java:3308)
    at android.app.ActivityThread.-wrap5(ActivityThread.java)
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1617)
    at android.app.ActivityThread.main(ActivityThread.java:6290)
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
 Caused by: java.lang.IllegalStateException: Couldn't read row 0, col -1 from CursorWindow.  Make sure the Cursor is initialized correctly before accessing data from it.
    at android.database.CursorWindow.nativeGetLong(Native Method)
    at android.database.CursorWindow.getLong(CursorWindow.java:511)
    at android.database.CursorWindow.getInt(CursorWindow.java:578)
    at android.database.AbstractWindowedCursor.getInt(AbstractWindowedCursor.java:69)
    at android.database.CursorWrapper.getInt(CursorWrapper.java:122)
    at com.owncloud.android.datamodel.SyncedFolderProvider.createSyncedFolderFromCursor(SyncedFolderProvider.java:335)
    at com.owncloud.android.datamodel.SyncedFolderProvider.getSyncedFolders(SyncedFolderProvider.java:100)
    at com.owncloud.android.services.observer.SyncedFolderObserverService.onCreate(SyncedFolderObserverService.java:66)
    at android.app.ActivityThread.handleCreateService(ActivityThread.java:3298)
    at android.app.ActivityThread.-wrap5(ActivityThread.java) 
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1617) 
    at android.app.ActivityThread.main(ActivityThread.java:6290) 
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886) 
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776) 
```

We change the db structure as it seems _after_ the observer queries for rows, thus the query fails leading to the death of the app :(

AndyScherzinger:
probably found the issue, testing...

AndyScherzinger:
@mario ypou are breaking the build: https://drone.nextcloud.com/nextcloud/android/4225 due to java8 compile targeted code ;)

mario:
As a note, this now requires Android Studio 2.4 to build.

AndyScherzinger:
@mario is it technically possible to downgrade the PR so it just needs Android Studio 2.3 which is the latest stable release and move for Jack usage for the time being to support java8 language features / compile targeting? 

What do you think @tobiasKaminsky?

AndyScherzinger:
@mario adding to my previous comment. The idea is to separate the two "issues" as in 1 PR for adding the media types (video/image/custom + db upgrade routine) and another stacked PR that adds further improvements/rewrite of the auto upload which would also include the java8 bump while I am not sure if going for AS2.4. preview makes sense or if we should use AS2.3 and Jack simply since we don't know if f-droid will be able to build it (yet) :( So it bears the risk of blocking the f-droid releases which I would like to be supported for 1.5 again.

What do you think @mario  @tobiasKaminsky ?

mario:
I am against separation because I did ask before I started working on this if it's fine to merge.

As for F-Droid build: there's no reason why it wouldn't be able to build it. It's just different gradle/and all versions. But if you prefer Jack *shrug*, feel free to modify this PR :)

AndyScherzinger:
I just wanted to discuss the matter ;) since after thinking about the AS Preview and alpha/beta gradle versions and Java8 support that this might lead to issues with f-droid (the build pipeline isn't always supporting the latest tool versions)and also that when merged to master it would force all devs to upgrade to a preview version.

AndyScherzinger:
@mario @tobiasKaminsky for whatever reason I don't know the newly introduced Drone jobs for the flavors fail for gplay and modified (which is a good thing because now we can see this in the CI builds 👍 )
https://drone.nextcloud.com/nextcloud/android/4331/2
https://drone.nextcloud.com/nextcloud/android/4331/3

Any idea why that is?

mario:
I will check. I fixed this

mario:
Ah. I will look into it - I know the solution. Probably due to rebase or
something.

mario:
Should work everywhere now :)

AndyScherzinger:
@tobiasKaminsky @mario the DroneCI build keeps failing again for "modified" https://drone.nextcloud.com/nextcloud/android/4433/3

AndyScherzinger:
@mario can you make the "setup a custom folder" item clickable? I tried but failed :/

mario:
@AndyScherzinger foldersync activity, method onMessageEvent - enjoy :)

AndyScherzinger:
@mario @tobiasKaminsky I made further progress on the UI parts and added all open issues to the main description

@mario can you also document all open issues from your side for the upload/job implementation part please? :)

Again: I am not sure anymore if the auto upload rewrite and the media/custom folder extension (which has been the original focus of this PR) should be mixed since this PR already has quite a lot of changes which might make it hard to keep it up to date with the master since my guess is that it'll still take some time to finish this one and further, probably mayor changes will be merged to master before this one :/ But I am fine if you folks decide to keep it that way!

AndyScherzinger:
This PR is now hard blocked by Google --> https://stackoverflow.com/questions/44301207/android-plugin-is-too-old-2-4-0-alpha7

due to the Java8 / AS2.4 move and it's 40 day limit on non-final tooling the build is dead (and no I won't set an env var to workaround!). So I can't even build/install this PR anymore...

cc @mario @tobiasKaminsky

mario:
@AndyScherzinger I fixed the build.

AndyScherzinger:
@mario I'll see that I test drive it later today <3

AndyScherzinger:
@mario  the build didn't work anymore, upgrading from 2.4+ to 3.0 didn't help either. I can't run the branch in AS 3.0 canary 4 anymore :(

mario:
@AndyScherzinger I've had to revert your gradle changes because indeed it won't build with them. With the current branch, just set daily override and it'll work.

I'd appreciate if you can look into the following issue: Add custom folder entry doesn't appear in auto uploads. I'd assume it's due to some rebase or something, but if you could check that would be awesome...

mario:
@AndyScherzinger nevermind fixed :D

mario:
- [ ] fix updates of synced folders

@AndyScherzinger I am doubtful about:

AndyScherzinger:
@mario that is a good question, I know from @tobiasKaminsky that he implemented such a Check somewhere in the App.

mario:
@AndyScherzinger considering I'm semi-away, do you think you could look into updating synced folders? E.g. when changing any of the available options for a synced folder (especially if it's not persisted previously) there is strange behaviour all around ...

Would appreciate if you could do this today so I can check it while I still have decent amount of time to look into it...

AndyScherzinger:
I am not home / with a keyboard at reach. Maybe @tobiasKaminsky can take a look?

tobiasKaminsky:
> Do you really think we should worry about this? If yes, then we also need to make sure that if you enable images for a folder and add a custom folder (with same path) that the images don't get uploaded twice.

I think this is really important to check. Even if it is a users mistake to set up upload for the same folder twice, we should prevent this, or at least warn the user.

AndyScherzinger:
@mario just tested the UI couldn't find any strange behavior? What did you see that I or @tobiasKaminsky should take a look?

Non-persisted custom folders don't show a switch, don't show a delete button, disable save button until local and remote folder have been chosen.

AndyScherzinger:
Issue found customer folders aren't shown when initializing UI (loading data), I'll have a look

mario:
@AndyScherzinger try changing some preferences (like wifi-only) for a non-persisted folder. That causes issues as well (not actually persisting the preference).

AndyScherzinger:
Thanks for the hint! Will have a look

AndyScherzinger:
@tobiasKaminsky @mario not sure if I understood all issues correctly but fixed anything I found regarding list content and preferences per entry.

mario:
@AndyScherzinger thanks. If you have time, could you also look into why Succesful uploads are removed from the "Uploads" view once you visit the "Uploads" view several times?

mario:
@AndyScherzinger small issue with your code. If there are no media folders (like in an emulator) then "Setup a custom folder" won't be shown :(

AndyScherzinger:
> @AndyScherzinger small issue with your code. If there are no media folders (like in an emulator) then "Setup a custom folder" won't be shown :(

@mario couldn't wrap my head around it yet, my initial implementaiton showed "no media folders" but now that diesn't get shown anymore but an empty screen ?! 0_o

mario:
@AndyScherzinger it shows "No media folder", but shouldn't it show the "Setup a custom folder" element on the top anyway? :)

tobiasKaminsky:
> it shows "No media folder", but shouldn't it show the "Setup a custom folder" element on the top anyway? :)

I am on it

tobiasKaminsky:
To get this PR bigger once more: we have FolderSyncActivity but all other stuff is named SyncedFolder.
To have this consistent I would like to change the name of the activity to SyncedFolderActivity?

mario:
@tobiasKaminsky sure.

AndyScherzinger:
@tobiasKaminsky sure, while it would have to be SyncedFolder**s**Activity

tobiasKaminsky:
After discussing with @AndyScherzinger we decided to let the "add custom folder" on top of the screen.

mario:
@tobiasKaminsky pure curiosity ... where was this discussed? You call me out for not communicating, and then you appear with a "decision" :)

Interesting...

mario:
Oh and clearly ... cc @MorrisJobke to the rescue, as usual.

AndyScherzinger:
We discussed this on Whatsapp today. While I am fine either way my prefered solution would be an scroll-able toolbar integration. Problem is that I am not sure if I find the time to implement this in time for a release and also if you folks would agree with going for a scroll-able toolbar solution.

mario:
@AndyScherzinger sorry but this is not acceptable. Since @tobiasKaminsky calls me out for not communicating all the time, this kind of decision-making process is a joke from your side.

Please also remember that I and you actually did it the way @tobiasKaminsky did it now and discarded this way afterward TOGETHER.

AndyScherzinger:
I am sorry for any miscommunication that might have happened (if it so happened!). 

From what I understood you guys talked about how to proceed about this matter today, decided that Tobias talks to me + decision to be made then and implemented then. If so than I'd say you guys agreed on this proceeding and then executed this proceeding, no?

MorrisJobke:
> @AndyScherzinger sorry but this is not acceptable. Since @tobiasKaminsky calls me out for not communicating all the time, this kind of decision-making process is a joke from your side.

Did you said before, that you are on vacation and @tobiasKaminsky and @AndyScherzinger should get to a decision on their own.

Just go and enjoy your vacation and rest a bit. 😉 I guess this was just a misunderstanding.

If Andy and Tobi think it is okay to keep it in because it's easier code wise then just keep it like that.

I guess in the future you should not try to work against each other but together. Either technical solution in a form of a PR is good and if there are real technical issues in here then do a code comment otherwise. 😉 

@tobiasKaminsky @AndyScherzinger Maybe also just write down the arguments for the decision and then it is clear to everybody and we can together move forward. Not everybody always needs to participate in a decision and also decisions could be reverted in the future. 

Keep calm and code \o/

mario:
No, it was not a misunderstanding. Annoyed by always being called out for
things he does.

I said that Tobias and Andy should discuss this in this PR so the whole
community can see. Instead it is done behind closed doors.

And I said this because Tobias feels it is unacceptable to say what he did
is wrong. Children game.

tobiasKaminsky:
I read your sentence in the wrong way: instead of discussing it *in* this PR, I read "discuss this PR".
So, sorry for that!

MorrisJobke:
> I said that Tobias and Andy should discuss this in this PR so the whole
> community can see. Instead it is done behind closed doors.

Come on - it was one decision that they did in private and it's also not about this specific one here. Yes this was not perfect, but this is what happens from time to time. In all other cases that was done properly.

Lets not start kindergarten here and blame everybody on every tiny step they do. You want to fix things together and that should be the main goal. And this can only happen if people in the public also know why things are done. And this @AndyScherzinger also explained why the decision was made:

> Problem is that I am not sure if I find the time to implement this in time for a release and also if you folks would agree with going for a scroll-able toolbar solution.


Now: Move on and code things \o/

AndyScherzinger:
As mentioned above and cited by Morris I would now try to implement the Toolbar version which will basically give us a very similar UI behavior to what @mario implemented with the recycler view while not having to pack it into the recycler view but the Toolbar with scroll-relation so it'll disappear when scrolling down and re-appears when scrolling up. In my personal opinion it gives us the nice scroll-aware behavior @mario achived with the recycler implementation but doesn't add a single/non auto upload content item to the recycler view.

Since it is the weekend and everybody is probably afk I'll try to do an implementation and will do it on a different branch so in case you are against such an implementation we don't have to revert anything. As for the UI look it'll basically look like the original implementation by @mario with a light grey bar at the top below the blue toolbar in case it is visible in the current moment.

*fingerscrossed*

AndyScherzinger:
PR against this branch has been opened #1195 making it scroll-able and part of the toolbar. Hope this solution fits everyone's point of view 😃

tobiasKaminsky:
@AndyScherzinger cool idea. :+1:

AndyScherzinger:
@tobiasKaminsky sweet! So can you take a look at #1195 ? :)

tobiasKaminsky:
Is this now ready for dev version? So we can ping all users who experienced problems with the current auto upload to get a feedback if this resolves it?

mario:
No

AndyScherzinger:
@mario lib release has been done 1.0.23

AndyScherzinger:
As per the open issues list in the top description all issues are fixed except one which we can fix down the road to release 1.5.0. @mario and I discussed merging this PR to master and then do any changes that will then still need to be done like bugfixes etc. should be done in separate PRs since maintaining this PR get a bit time consuming due to the amount of code changes in it.

Things to be done before a merge to master in my opinion are:
1. drone build needs to work b/c else we loose all QA metrics currently in place and all new or rebased branches will fail in CI too
2. ship in the dev build where the build containing this PR needs to have been published by f-droid (this is imho a blocking issue if f-droid doesn't build/publish it)
3. as per discussion/feedback of @tobiasKaminsky bring back the menu/actions in the uploads activity (I suppose they have been removed in this PR? If not than forget about this item here)

What do you think @mario @tobiasKaminsky ? Other than that any issues/work remaining that needs to be done for this PR except the things I mentioned?

mario:
All good, except we have to wait for Google to fix the build :P

AndyScherzinger:
@mario do you have some details/links what Google needs to do?

mario:
@AndyScherzinger sure. Release a non-alpha gradle 3.0.0?

I reverted the menu removal from uploads. If you want to satisfy lint & codacy, would be great.

mario:
(The same is probably true for f-droid and dev btw)

AndyScherzinger:
Ah :) probably yeah. I can check codacy (and probably lint too) sure :)

AndyScherzinger:
@tobiasKaminsky can you look into the codacy matter? because for whatever reason it is not referncing https://www.codacy.com/app/Nextcloud/android/pullRequest?prid=590865 but https://www.codacy.com/app/tobias_2/android/pullRequest?prid=590866 which looks like your personal codacy config...

@mario checkde the "official" codacy report and we are fine there (already took care of that) 🥇

MorrisJobke:
> drone build needs to work b/c else we loose all QA metrics currently in place and all new or rebased branches will fail in CI too

Drone should work again. But it's currently quite slow, because 2 nodes are not available currently-

MorrisJobke:
> Drone should work again. But it's currently quite slow, because 2 nodes are not available currently-

There are test failures now 🙈

AndyScherzinger:
@MorrisJobke Nope, the build is *not* working if you look at the logs, it fails due to gradle plugins https://drone.nextcloud.com/nextcloud/android/266/1 so the "test" fails because it can't even build the app!

tobiasKaminsky:
> @tobiasKaminsky can you look into the codacy matter? because for whatever reason it is not referncing https://www.codacy.com/app/Nextcloud/android/pullRequest?prid=590865 but https://www.codacy.com/app/tobias_2/android/pullRequest?prid=590866 which looks like your personal codacy config...

This is not valid any longer, or? As codacy is "green"?

tobiasKaminsky:
> Release a non-alpha gradle 3.0.0?

Do we have an estimation when this will be done?
For which part is this needed? Maybe we can circumvent the need of gradle 3.0?

AndyScherzinger:
Unfortunately it seems we can't since the build demand 3.0.0, see build error log but the build keeps failing with the 3.0.0 alpha. Maybe it is worth looking into getting the build back to work with the alpha.

mario:
We currently build with gradle 2.5 non-stable which will be replaced with 3.0.0 alpha. Unfortunately 3.0.0 alpha is completely broken so we can't build with that. We will have to wait until Google releases 3.0.0, but I'd still appreciate reviews and approval of this code base and functionality.

mario:
@AndyScherzinger please review & test this today before you disappear :P

AndyScherzinger:
I'll be having it on my phone for the upcoming week but won't be able to do a code review or testing today :(

@tobiasKaminsky please review and test too 🤖

AndyScherzinger:
@mario I rebased to latest master (just a second ago) and installed it on my phone :)

mario:
@AndyScherzinger thanks!

mario:
So ...

- I disabled the chunking hack and will properly fix it in 1.5.1
- I implemented near-instant image & video uploads on 7+ (!!!!). Near instant means a few seconds. 

Yay yay and yay! :D

mario:
Personally, I think our current Auto Upload is too complicated and that most users would prefer the simple old "Upload photos" and "Upload videos" functionality. However, we have what we have with a really nice UI, so we'll live with that :)

To make things a bit easier, I've introduced an expert mode with a checkbox in settings. Currently, it adds the following:

- Logs
- Manual upload control in uploads screen

What I'm proposing additionally is to add Custom folders support only in Expert mode, because it's a relatively complex feature, likely to be used only by a select "few" (where few is more than 2,3 of course, otherwise we wouldn't do it - what I mean is, more technical users).

Please say your thoughts on this. 

cc @jancborchardt @AndyScherzinger @tobiasKaminsky

mario:
Also cc @nextcloud/designers about my last comment.

jancborchardt:
Yeah, the basic default should be something simple. Just "Back up everything" from the phone.

Detail folders etc can always be managed further, but we need to make it very seamless. Also since Auto-Upload seems to be something people often report bugs with, we should keep it as simple as possible.

mario:
@jancborchardt luckily I fixed all the bugs now :)

Ok, so since we can't have "Back up photos" & "Back up videos" (photos & videos) right now, I'll just keep Custom folders (ones where you can choose any folder on the filesystem) out of reach via expert mode.

Agreed?

comradekingu:
Switching on the fly would be nice
[♬] [■◂] [[🖹](http://www.fileformat.info/info/unicode/char/1f5b9/index.htm)]   |→[  ]←
Lets say you want a text document, in a folder that has a lot of other files.
Ideally one could click each icon, having already navigated to the correct folder, to make a selection.

Having made the selection, you could then press "make backup" to save just those files (?)

@AndyScherzinger It says "Setup a custom folder" in the first picture, and "setup" isn't a verb. It should be "Set up"

AndyScherzinger:
@mario agreed for 1.5.0 :) I think custom folders are quite powerful but aren't UX polished enough at the moment and can also lead to duplicate configurations at the moment, so expert mode seems to be the right choice for now. 

I'd like to further discuss how custom folders can be improved at the Hackweek in August and then see how to progress from there. :) 

@mario since I am afk can you fix the String mentioned by @comradekingu? Thanks :) 

As for the file selection that could also be an enhancement but is imho out of scope for this PR and also for the next release since this is basically three last remaining open PR that needs to be merged for the release except for bugfixing. 

@mario running the branch on my phone since Saturday. Works fine so far, switching networks, flight mode etc. :+1: Automatic restarts work fine :+1: I didn't see the menu on the upload screen but that might be due to my checked being to old by now. Great improvements to the upload jobs Mario! Looking forward to update to the latest code next week :)

mario:
@AndyScherzinger this has already been fixed. 

    <string name="autoupload_custom_folder">Set up a custom folder</string>

cc @comradekingu

mario:
@AndyScherzinger please rebase before doing a review.

@tobiasKaminsky please review.

AndyScherzinger:
@mario I will do, beginning of next week when I am back :)

gahujipo:
I'm not sure if it's appropriate to mention here a few things I noticed with the dev apk from 20170730. I took the link to that apk from another issue here. Is there a link for an always up to date apk of this branch? 

- My personal nextcloud is not available from WAN so I always receive a message "upload failed" when I access a WiFi which is not my home WiFi. Could it be an idea to add an option to limit the upload attempts to a specific SSID like for example DAVdroid it does? 
- Auto Upload is unidirectional, is that true? It does not sync files in both directions? When I actived auto upload on my notes folder it does not sync from the Nextcloud to the mobile phone. Is that behaviour intentioned? 
- Sometimes my phone loses the connection to the WiFi and an upload failed message appears. Could it be a good idea to give the user the option for a "retry now" button under the notification in the notification center? 
- The preferences window for auto upload shows a switch next to "Preferences for auto uploading". At the first sight it was not clear to me that that switch is uses to enable the auto upload feature for that folder pair. My idea was to grey out the options in the window unless the switch is set to true.
- The selection option for the remote folder shows "No files here - Nothing in here. You can add a folder". For me it was at first not clear that this windows is not intentioned to show remote content but is only a folder selection option. Is there a bettery way to let the user select the remote folder? 

For the rest: great work! Love to see that feature coming into 1.5.0 stable.

tobiasKaminsky:
> @tobiasKaminsky please review.

will do tomorrow (friday).
If you have the time can you rebase and check codacy?

mario:
@gahujipo thank you for testing, you're awesome. Here are my comments:

- unlikely. Adds too much complexity to the overal UX.
- for now, indeed. It also only syncs NEW files, not existing ones.
- if you enable expert mode, you have a retry in the Uploads view. Also, the system automatically retries failed uploads every now and then.
- this is actually a perfect idea.
- Will think about this.

mario:
@tobiasKaminsky done.

gahujipo:
> thank you for testing

No problem. 
> unlikely. Adds too much complexity to the overal UX.

I know it is out of the scope of this PR, but could that maybe be a general option for the app in the settings? 

> if you enable expert mode, you have a retry in the Uploads view. 

Oh, that's nice. Adding this option to the notification directly is too complex (maybe only in expert mode?), I guess?

mario:
@gahujipo maybe one day (for your last request) - can you please file a separate issue please?  :)

gahujipo:
@mario: 
- Restrict upload in specific SSIDs only is already contained in this issue: https://github.com/nextcloud/android/issues/270

- New retry button enhancement issue: https://github.com/nextcloud/android/issues/1261

mario:
@AndyScherzinger @tobiasKaminsky I disabled aapt2, so this should build now actually.

mario:
@AndyScherzinger @tobiasKaminsky Travis started using 3.0.0 so we'll see...

mario:
So many hours spent thinking, so many hours spent coding, so many hours spend testing, 137 comments, 172 commits, 148 changed files later and Drone is finally green.

Please review and let’s get this over with!

mario:

AndyScherzinger:
@mario @tobiasKaminsky something must have gone wrong with the latest rebases looking at the lint score being back at former, very hight score!

mario:
@AndyScherzinger lint is just being silly.

tobiasKaminsky:
Combining Ellipsize and Maxlines
../../../src/main/res/layout/activity_row.xml:53: Combining ellipsize=marquee and maxLines=1 can lead to crashes. Use singleLine=true instead.

But @AndyScherzinger changed this with following commit message:
replace deprecated android:singleLine with android:maxLines=1
https://github.com/nextcloud/android/commit/6c3af67bad5991d66ed4d6e928f10cdf5ef11de0

...ideas? ;-)

tobiasKaminsky:
13 errors: ../../../src/main/res/values-ar/strings.xml:265: "prefs_folder_sync_local_path_title" is translated here but not found in default locale

These should be automatically be removed from transifex once this PR is merged.

src/main/java/com/owncloud/android/jobs/AutoUploadJob.java
```diff
@@ -43,39 +51,72 @@
...
+                        requiresCharging,
+                        true
+                );
+
+                lock.release();
+                wakeLock.release();
+                return Result.SUCCESS;
+
+            } catch (FileNotFoundException e) {
+                Log_OC.d(TAG, "Something went wrong while trying to access file");
```
tobiasKaminsky: Exception should be logged as errors

src/main/java/com/owncloud/android/jobs/FilesSyncJob.java
```diff
@@ -0,0 +1,150 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.text.ParsePosition;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
+import java.util.TimeZone;
+
+public class FilesSyncJob extends Job {
```
tobiasKaminsky: Short description

src/main/java/com/owncloud/android/jobs/FilesSyncJob.java
```diff
@@ -0,0 +1,150 @@
...
+    public static final String TAG = "FilesSyncJob";
+
+    public static String SKIP_CUSTOM = "skipCustom";
```
tobiasKaminsky: final?

src/main/java/com/owncloud/android/jobs/FilesSyncJob.java
```diff
@@ -0,0 +1,150 @@
...
+
+        // Create all the providers we'll need
+        final FilesystemDataProvider filesystemDataProvider = new FilesystemDataProvider(contentResolver);
+        SyncedFolderProvider syncedFolderProvider = new SyncedFolderProvider(contentResolver);
+
+        for (SyncedFolder syncedFolder : syncedFolderProvider.getSyncedFolders()) {
+            if ((syncedFolder.isEnabled()) && (!skipCustom || MediaFolder.CUSTOM != syncedFolder.getType())) {
+                for (String path : filesystemDataProvider.getFilesForUpload(syncedFolder.getLocalPath(),
+                        Long.toString(syncedFolder.getId()))) {
+                    if (JobManager.instance().getAllJobRequests().size() < 80) {
```
tobiasKaminsky: This means that jobs are limited to 80 outstanding jobs?
What happens if there are more?

src/main/java/com/owncloud/android/jobs/NContentObserverJob.java
```diff
@@ -0,0 +1,66 @@
...
+import android.support.annotation.RequiresApi;
+
+import com.evernote.android.job.JobRequest;
+import com.evernote.android.job.util.support.PersistableBundleCompat;
+import com.owncloud.android.utils.FilesSyncHelper;
+
+import java.util.concurrent.TimeUnit;
+
+@RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+public class NContentObserverJob extends JobService {
```
tobiasKaminsky: description

src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -394,31 +416,85 @@ protected RemoteOperationResult run(OwnCloudClient client) {
...
             result = new RemoteOperationResult(e);
 
         } finally {
             mUploadStarted.set(false);
+
+            if (fileLock != null) {
+                try {
+                    fileLock.release();
+                } catch (IOException e) {
+                    Log_OC.d(TAG, "Failed to unlock file with path " + mOriginalStoragePath);
```
tobiasKaminsky: log_oc.e()

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
```diff
@@ -334,7 +372,37 @@ private SyncedFolderDisplayItem createSyncedFolderFromMediaFolder(@NonNull Media
...
+    }
+
+    private File[] getFileList(File localFolder) {
+        File[] files = localFolder.listFiles(new FileFilter() {
+            @Override
+            public boolean accept(File pathname) {
+                return !pathname.isDirectory();
+            }
+        });
+        Arrays.sort(files, new Comparator<File>(){
```
tobiasKaminsky: can files be null? Then there would be a NPE

src/main/java/com/owncloud/android/ui/activity/UploadListActivity.java
```diff
@@ -243,26 +265,22 @@ public boolean onOptionsItemSelected(MenuItem item) {
 
     @Override
     public boolean onCreateOptionsMenu(Menu menu) {
-        MenuInflater inflater = getMenuInflater();
-        inflater.inflate(R.menu.upload_list_menu, menu);
+        SharedPreferences appPrefs =
+                PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
+        if (appPrefs.getBoolean("expert_mode", false)) {
```
tobiasKaminsky: Better use a string constant here / extract to preferences.

src/main/java/com/owncloud/android/ui/adapter/SyncedFolderAdapter.java
```diff
@@ -1,22 +1,22 @@
...
- *   but WITHOUT ANY WARRANTY; without even the implied warranty of
- *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- *   GNU AFFERO GENERAL PUBLIC LICENSE for more details.
- *
- *   You should have received a copy of the GNU Affero General Public
- *   License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ * @author Andy Scherzinger
+ * Copyright (C) 2016 Andy Scherzinger
+ * Copyright (C) 2016 Nextcloud
+ * <p>
```
tobiasKaminsky: proper html tag with closing /> ?

src/main/java/com/owncloud/android/ui/dialog/parcel/SyncedFolderParcelable.java
```diff
@@ -38,6 +38,7 @@
     private Boolean mEnabled = false;
     private Boolean mSubfolderByDate = false;
     private Integer mUploadAction;
+    private int mType;
```
tobiasKaminsky: Type as enum?

src/main/java/com/owncloud/android/utils/FilesSyncHelper.java
```diff
@@ -0,0 +1,326 @@
...
+import org.lukhnos.nnio.file.Path;
+import org.lukhnos.nnio.file.Paths;
+import org.lukhnos.nnio.file.SimpleFileVisitor;
+import org.lukhnos.nnio.file.attribute.BasicFileAttributes;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.List;
+
+public class FilesSyncHelper {
```
tobiasKaminsky: Description?

src/main/java/com/owncloud/android/utils/FilesSyncHelper.java
```diff
@@ -0,0 +1,326 @@
...
+        final ContentResolver contentResolver = context.getContentResolver();
+        ArbitraryDataProvider arbitraryDataProvider = new ArbitraryDataProvider(contentResolver);
+
+        Long currentTime = System.currentTimeMillis();
+        double currentTimeInSeconds = currentTime / 1000.0;
+        String currentTimeString = Long.toString((long) currentTimeInSeconds);
+
+        String syncedFolderInitiatedKey = "syncedFolderIntitiated_" + syncedFolder.getId();
+        boolean dryRun = TextUtils.isEmpty(arbitraryDataProvider.getValue
+                ("global", syncedFolderInitiatedKey));
```
tobiasKaminsky: As final string?

src/main/java/com/owncloud/android/utils/FilesSyncHelper.java
```diff
@@ -0,0 +1,326 @@
...
+                        @Override
+                        public FileVisitResult visitFileFailed(Path file, IOException exc) {
+                            return FileVisitResult.CONTINUE;
+                        }
+                    });
+
+                }
+
+            } catch (IOException e) {
+                Log.d(TAG, "Something went wrong while indexing files for auto upload");
```
tobiasKaminsky: log_oc.e

src/main/java/com/owncloud/android/utils/FilesSyncHelper.java
```diff
@@ -0,0 +1,326 @@
...
+            for (SyncedFolder syncedFolder : syncedFolderProvider.getSyncedFolders()) {
+                if (MediaFolder.VIDEO == syncedFolder.getType()) {
+                    hasVideoFolders = true;
+                } else if (MediaFolder.IMAGE == syncedFolder.getType()) {
+                    hasImageFolders = true;
+                }
+            }
+        }
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
```
tobiasKaminsky: add requires api to this function and you can remove the check.

src/main/java/com/owncloud/android/utils/FilesSyncHelper.java
```diff
@@ -0,0 +1,326 @@
...
+            jobScheduler.cancel(ContentSyncJobId);
+        }
+    }
+
+    @RequiresApi(api = Build.VERSION_CODES.N)
+    private static void scheduleJobOnN(boolean hasImageFolders, boolean hasVideoFolders,
+                                       boolean force) {
+        JobScheduler jobScheduler = MainApp.getAppContext().getSystemService(JobScheduler.class);
+
+        if (((android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) && (hasImageFolders || hasVideoFolders)) &&
```
tobiasKaminsky: version check is not needed as requires api is specified

src/main/java/com/owncloud/android/utils/ReceiversHelper.java
```diff
@@ -0,0 +1,71 @@
...
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
+
+import com.evernote.android.job.JobRequest;
+import com.evernote.android.job.util.Device;
+import com.owncloud.android.MainApp;
+
+public class ReceiversHelper {
```
tobiasKaminsky: description

tobiasKaminsky:
Outstanding:
- Review of @AndyScherzinger 
- fixing codacy and lint

src/gplay/AndroidManifest.xml
```diff
@@ -59,6 +59,10 @@
             </intent-filter>
         </activity>
 
+        <activity
+            android:name=".ui.activity.FileDisplayActivity"
+            tools:node="remove"/>
```
mario: Magic. Used to have purpose. Not anymore. removing.

src/main/java/com/owncloud/android/datamodel/FileSystemDataSet.java
```diff
@@ -0,0 +1,103 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.owncloud.android.datamodel;
+
+
```
mario: Done.

src/main/java/com/owncloud/android/datamodel/FilesystemDataProvider.java
```diff
@@ -0,0 +1,189 @@
...
+import android.database.Cursor;
+import android.net.Uri;
+
+import com.owncloud.android.db.ProviderMeta;
+import com.owncloud.android.lib.common.utils.Log_OC;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class FilesystemDataProvider {
```
mario: Done.

src/main/java/com/owncloud/android/datamodel/SyncedFolder.java
```diff
@@ -76,9 +79,10 @@ public SyncedFolder(long id, String localPath, String remotePath, Boolean wifiOn
      * @param account         the account owning the synced folder
      * @param uploadAction    the action to be done after the upload
      * @param enabled         flag if synced folder config is active
+     * @param type            the type of the folder
      */
     public SyncedFolder(String localPath, String remotePath, Boolean wifiOnly, Boolean chargingOnly,
-                        Boolean subfolderByDate, String account, Integer uploadAction, Boolean enabled) {
+                        Boolean subfolderByDate, String account, Integer uploadAction, Boolean enabled, Integer type) {
```
mario: Feel free to change.

src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -418,6 +419,7 @@ private Bitmap doFileInBackground() {
     }
 
     public static class MediaThumbnailGenerationTask extends AsyncTask<Object, Void, Bitmap> {
+        private enum Type {IMAGE, VIDEO}
```
mario: Feel free to change.

src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -455,16 +407,29 @@ public void cancelPendingJob(String accountName, String remotePath){
         return getUploads(ProviderTableMeta.UPLOADS_STATUS + "==" + UploadStatus.UPLOAD_SUCCEEDED.value, null);
     }
 
+    public OCUpload[] getFailedButNotDelayedUploadsForCurrentAccount() {
+        Account account = AccountUtils.getCurrentOwnCloudAccount(mContext);
+
+        return getUploads(ProviderTableMeta.UPLOADS_STATUS + "==" + UploadStatus.UPLOAD_FAILED.value + AND +
+                        ProviderTableMeta.UPLOADS_LAST_RESULT + "<>" + UploadResult.DELAYED_FOR_WIFI.getValue() + AND +
+                        ProviderTableMeta.UPLOADS_LAST_RESULT + "<>" + UploadResult.LOCK_FAILED.getValue() + AND +
+                        ProviderTableMeta.UPLOADS_LAST_RESULT + "<>" + UploadResult.DELAYED_FOR_CHARGING.getValue() + AND +
```
mario: Done.

src/main/java/com/owncloud/android/jobs/AutoUploadJob.java
mario: Removed file

src/main/java/com/owncloud/android/jobs/FilesSyncJob.java
mario: Actually, we don't issue jobs here anymore. But nothing would happen even in the previous case - new files would just get scheduled on the next run, rather than right away. I removed this limitation since it's useless.

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
mario: Solving.

src/main/java/com/owncloud/android/ui/dialog/parcel/SyncedFolderParcelable.java
mario: Feel free to change.

src/main/java/com/owncloud/android/utils/FilesSyncHelper.java
mario: Already has it. Removing.

mario:
@tobiasKaminsky thanks for the review. Fixed all except for the ENUMs - if you feel they're an absolute MUST, please feel free to do it.

I guess @jasonbayton tested it quite a lot, I tested it a bit, and @AndyScherzinger did some as well I hope?

jasonbayton:
I have about 42 test build APKs on my phone. I've tested it well.

AndyScherzinger:
@tobiasKaminsky 

> But @AndyScherzinger changed this with following commit message:
replace deprecated android:singleLine with android:maxLines=1
6c3af67

fixed :)

mario:
@AndyScherzinger since you're working on this, can you also change those ints to Enums like @tobiasKaminsky requested?

AndyScherzinger:
> can you also change those ints to Enums like @tobiasKaminsky requested?

@mario I'll see what I can do if I find the time while I am not sure if it is worth it. Can Parcelables use Enums?

mario:
Complicates things but can be done. iHmo if tou agree leave as is

src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
AndyScherzinger: kept this enum since this is the thumbnail cache implementation and created a new one for the media folder types and would keep this separated!

AndyScherzinger:
@tobiasKaminsky all your review comment should now be solved except for one which I'd like to dismiss, see comment :)

tobiasKaminsky:
From lint:
`
Using wakeLock without timeout
../../../src/main/java/com/owncloud/android/jobs/FilesSyncJob.java:76: Provide a timeout when requesting a wakelock with PowerManager.Wakelock.acquire(long timeout). This will ensure the OS will cleanup any wakelocks that last longer than you intend, and will save your user's battery.
`

mario:
But we have no idea how long a job will run...

tobiasKaminsky:
While chatting with fdroid guys to verify if this build, they told me that gradle 4.1 is released: https://github.com/gradle/gradle/releases
so we can use the stable 4.1 instead of 4.1-rc2.

AndyScherzinger:
> While chatting with fdroid guys to verify if this build, they told me that gradle 4.1 is released: https://github.com/gradle/gradle/releases
so we can use the stable 4.1 instead of 4.1-rc2.

sweet, just give me a second :)

AndyScherzinger:
@tobiasKaminsky are you fine with my comment on the last remaining CR comment of yours? --> https://github.com/nextcloud/android/pull/797#discussion_r131752604

src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
tobiasKaminsky: Fine :+1:

AndyScherzinger:
👍  code fine by me, latest CR fixes done myself and also tested the update routine for the DB

* 191 comment
* 203 commit
* 309 files changed

@tobiasKaminsky awaiting your final approval :)

AndyScherzinger:
👍  code fine by me, latest CR fixes done myself and also tested the update routine for the DB

* 188 comment
* 206 commit
* 314 files changed

@tobiasKaminsky awaiting your final approval :)

[Approved with PullApprove](https://pullapprove.com/nextcloud/android/pull-request/797/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=nextcloud/android)

AndyScherzinger:
👍  code fine by me, latest CR fixes done myself and also tested the update routine for the DB

* 191 comment
* 203 commit
* 309 files changed

@mario awaiting your final approval :)

mario:
👍

AndyScherzinger:
@mario @tobiasKaminsky please wait until drone completes before hitting merge! Having this PR with all checks turned green would be awesome!!!

AndyScherzinger:
@mario @tobiasKaminsky nevermind :/ Tobias' lint fixes will trigger a bot push and thus another change rendering all approval obsolete... So whenever drone is done with the 2 rounds, let's merge... ;)

mario:
@AndyScherzinger so you will merge that other PR in here?

AndyScherzinger:
No, I'll merge now! 

Cc @tobiasKaminsky

tobiasKaminsky:
Hurray :tada: :tada: :tada: :tada: :tada: :tada:

oparoz:
Many thanks guys!

PR #827, opened 2017-04-12
Implement webview welcome pages

mario:


tobiasKaminsky:
Code wise 👍 
But for me the blue border on top and bottom is way too much?
It looks better if the border is just as big as the regular header in our app? 
Also "skip", "dots" and "right arrow" should be centered vertically 

screenshot_1492531430

mario:
a) This is the same as our regular welcome view
b) centered how? This is also the same as our regular welcome view. I have no intention of centering it to the web page itself since the page might change. (also, not sure how I'd do that - well, I am, but it is magic then)

mario:
@tobiasKaminsky if you agree with my comments, please let me know and I'll merge.

tobiasKaminsky:
Not centering the website itself, only the text elements.

An idea

 -->

mario:
@tobiasKaminsky will do it like that.

tobiasKaminsky:
I have already done it, so I can push the changes ;-)

mario:
Do it and I will check

mario:
Merging.

jancborchardt:
What about the top bar? Will there be text in it like the app name? Or should we remove it?

tobiasKaminsky:
@jancborchardt It is up to you :-D
Just open up a new issue with your wish and I am happy to do it.

PR #829, opened 2017-04-12
Ignore thumbnail files

mario:


AndyScherzinger:
:+1: looks good :)

tobiasKaminsky:
👍 
After the next addition we should extract it into a new function.

Issue #798, opened 2017-04-03
Navigation is broken on master for Activities screen


AndyScherzinger:
### Actual behaviour
- Navigation for favorites and shared screen is broken

### Expected behaviour
-Tell us what should happen
 
### Steps to reproduce
1. Navigate to activities screen
2. click on favorites (nothing happens)
3. click on shared (nothing happens) 

### Environment data
Android version: 7.1.1

Device model: OP3T

Stock or customized system:

Nextcloud app version: latest master

Nextcloud server version: 11.0.2

PR #845, opened 2017-04-19
bugfix for #798

tobiasKaminsky:
bugfix for #798
checking cast if remote request is long running
preventing MissingActivityException if fragment is detached from Activity


mario:
👍 

Might have bugs and UX could be improved, but it's better than what we have in there now. Thanks.

AndyScherzinger:
👍 I second that :)

PR #849, opened 2017-04-19
Fix drone

mario:


.drone.yml
```diff
@@ -6,7 +6,7 @@ pipeline:
       - emulator -avd test -no-window &
       - ./wait_for_emulator.sh
       # build app and assemble APK, in debug mode
```
tobiasKaminsky: Then the comment is wrong ;-)

mario:
Less relevant compared to drone failing :p

I can fix it later if you dont do it before :)

PR #858, opened 2017-04-20
prevent NPE if trying to store grid preference for favorites/shared

tobiasKaminsky:
If in favorites/shared and "switch to grid" it crashes as it tries to store preference for this mFile.
Instead now do nothing.

To think of, if we want to store the preference for these searches...?


AndyScherzinger:
👍 looks good, @mario feel free to hit the merge button if you also approve :)

PR #861, opened 2017-04-21
change all namings of beta/nightly to new "dev"

tobiasKaminsky:


mario:
So many name changes :( Hopefully this is the last time we do it.

tobiasKaminsky:
Beta was confusing with Play store Beta and nightly was confusing as this suggests a real daily build...
We just stay with this (hopefully) :-)

Issue #869, opened 2017-04-21
Shared search view: NPE due to null in result array


tobiasKaminsky:
`
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.owncloud.android.datamodel.OCFile.isFolder()' on a null object reference
at com.owncloud.android.utils.FileStorageUtils$6.compare(FileStorageUtils.java:422)
`

This happens due to have a null value in GetRemoteSharesOperation

Issue #857, opened 2017-04-20
Favorites/photos: clicking on images result in wrong image


tobiasKaminsky:
- open favorites
- click on one image
-> wrong image

Caused by PreviewImagePagerAdapter uses "parentFolder", which is the parent folder of the image to show and not the file list of favorites/images

According to that:
if FileDisplayActivity, startImagePreview(OCFile file) intents are set, but never used.
Instead this is handled by getFile() and setFile()

PR #872, opened 2017-04-21
always refresh ocShares prior listing them to have an up-to-date ocFile

tobiasKaminsky:
The expected result of this PR is to have a sharedView that does not crash any longer.
(clicking on an image might still crash or result in a wrong image due to #857)

Ref: #869

### DIFF
--- a/src/main/java/com/owncloud/android/ui/adapter/FileListListAdapter.java
+++ b/src/main/java/com/owncloud/android/ui/adapter/FileListListAdapter.java
@@ -49,7 +49,9 @@ import com.owncloud.android.datamodel.ThumbnailsCacheManager;
 import com.owncloud.android.db.PreferenceManager;
 import com.owncloud.android.files.services.FileDownloader.FileDownloaderBinder;
 import com.owncloud.android.files.services.FileUploader.FileUploaderBinder;
+import com.owncloud.android.lib.common.operations.RemoteOperationResult;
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/ui/adapter/FileListListAdapter.java
```diff
@@ -491,9 +493,19 @@ public void setData(ArrayList<Object> objects, ExtendedListFragment.SearchType s
 
                     shares.add(ocShare);
 
-                    OCFile ocFile = mStorageManager.getFileByPath(ocShare.getPath());
-                    if (!mFiles.contains(ocFile)) {
-                        mFiles.add(ocFile);
+                    // get ocFile from Server to have an up-to-date copy
+                    ReadRemoteFileOperation operation = new ReadRemoteFileOperation(ocShare.getPath());
```
mario: Please use search api instead to get all the properties for all the relevant files in 1 network call. You might need to extend the search magic in the library for this though, but that would be significantly better.

tobiasKaminsky: As of feature freeze on server the enhancement will not be available for NC12, so I opened up #879 and this shall be merged, though it is not perfect :/

mario:
👍 

@AndyScherzinger please review and merge.

AndyScherzinger:
👍 Looks good! :)

PR #886, opened 2017-04-25
Updated fr (French) translation

loi219:
Commit made via Stringlate by Sven Rouvinez


mario:
All translations are done via transifex, and this PR is strange.

PR #451, opened 2016-12-14
Fix file modification date on upload

AndyScherzinger:
work in progress PR to activate codacy checks

PR #887, opened 2017-04-25
propose instant image/video upload make use of file's last modified time

wowangus:
Propose instant image/video upload make use of file's last modified time rather than datetime as of upload trigger

Summary:
the changes is to respect and retain an image/video file's last modified time (i.e. mtime) of itself.

Description:
Uses of System.currentTimeMillis() as of upload triggered are not guaranteeing the truth of last modification datetime about users' uploaded data.

Uses of Media.DATE_MODIFIED are seemingly more appropriate for the usage.

==========
For example:
Consider that 
1. USER_A captured a screenshot at 2017-04-25 1800HKT, a photo at 2017-04-25 1900HKT, a video at 2017-04-25 2000HKT
2. USER_A then connected the android phone to a stable WiFi network at 2017-04-26 1200HKT, which enables Nextcloud android app to trigger those pending auto-upload/backup jobs

==========
Currently unexpected result:
USER_A will be told by Nextcloud that the uploaded files are all last modified on/at 2017-04-26 1200HKT

==========
Expected result:
Nextcloud should tell users an exact truth about last modification time of an uploaded file.

e.g.:
USER_A should be told by Nextcloud that the uploaded screenshot is last modified on/at 2017-04-25 1800HKT, the uploaded photo is last modified on/at 2017-04-25 1900HKT, the uploaded video is last modified on/at 2017-04-25 2000HKT.

==========
Other discussions:
1. comments are polluting the clean code but I temporarily write down comments to explain my changes for sack of facilitating others to review my changes. pls remove it after reviews if feeling uncomfortable.
2. File Upload Datetime Timestamp is also considerably important information for users to get an understanding about their uploaded file when users revisit their files. For file_upload_datetime_timestamp, uses of System.currentTimeMillis() (as of file upload trigger/completion at client/server) would be suitable to the usage. Should it be considered to be implemented in a planned future? Besides, considered that, file upload is async task, where a progress of file upload could be taking a lengthly upload duration, say probably an hour long for big video file over a slow network, take file upload completion timestamp at server as reference for file_upload_datetime_timestamp may be considerably more logical to usage? I think that could be discussed.

Thanks for your attentions!

### DIFF
--- a/src/main/java/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
+++ b/src/main/java/com/owncloud/android/files/InstantUploadBroadcastReceiver.java
@@ -87,7 +87,9 @@ public class InstantUploadBroadcastReceiver extends BroadcastReceiver {
         String file_path = null;
         String file_name = null;
         String mime_type = null;
-        long date_taken = 0;
+        // comments are polluting the clean code but I temporarily write down comments to explain my changes for sack of facilitating others to review my changes. pls remove it after reviews if feeling uncomfortable.
+        //long date_taken = 0; // deprecated by Angus on 20170425
+        long date_modified = 0; // created by Angus on 20170425 // date_modified as variable identifier is more appropriate
 
         Log_OC.i(TAG, "New photo received");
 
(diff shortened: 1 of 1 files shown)


mario:
First of all, thank you for your attempt at contributing!

InstantUpload is deprecated and not supported. In the future, the code for this will be completely removed so I see no point in approving this.

Maybe @tobiasKaminsky or @AndyScherzinger think otherwise.

wowangus:
it's fine. my changes was about to upload image/video to a subPath according to its modification datetime rather than the upload datetime, in case that option is enabled.

I have been keeping an eye on #451 and https://github.com/nextcloud/android-library/pull/41 as well

AndyScherzinger:
That's true.

@tobiasKaminsky @mario What do you think in general? I'd vote for the creation time in general simply since as a user I would expect a photo to be uploaded to the folder when I took the photo not the date when I did some image optimizations in a photo app on the phone.

tobiasKaminsky:
Creation time is the better approach, I guess.
@mario can we re-use this for auto upload? Or is it already working fine in auto upload?

mario:
This should work for auto upload already.

PR #901, opened 2017-04-26
Gplay edition

mario:
This PR does the following:

- refactors flavours so we now have three flavours:

- generic - F-droid one, no GPlay
- modified - branded one, GPlay for push & Analytics
- zstore - Play Store version, Gplay for push, Analytics disabled


tobiasKaminsky:
Currently we have
build.gradle with firebase
build.gradle.modified with gms

But what about fdroid (no gms and no firebase)?

tobiasKaminsky:
👍

Issue #882, opened 2017-04-24
ExternalLink Improvement  


aleister09:
### Actual behaviour
If the user open a external link without network the app shows a 404 WebPage 

### Expected behaviour
It would be nice that the app can load a custom error page instead of the 404 WebPage 

- The Error Page can be loaded from assets or raw 
- It can be the same to WebLogin and Welcome page 

### Steps to reproduce
1. Enable Airplane Mode
2. Open the app
3. Open an external link

### Environment data
Android version: 7.1.2

Device model: Nexus 5

Stock or customized system:

Nextcloud app version: Master

Nextcloud server version: 12

### Logs
![device-2017-04-24-113139](https://cloud.githubusercontent.com/assets/25756086/25349076/17a1151c-28e6-11e7-9931-a37f8b728d83.png)

tobiasKaminsky:
@aleister09 please test it with your custom error html page.

PR #904, opened 2017-04-27
ability to show custom error message for webview

tobiasKaminsky:
Ref: #882 

Show custom error web view if file "assets/customError.html" exists.
Currently a dummy file exists for custom build, for regular build there is no such file and therefore the normal error message is shown.


AndyScherzinger:
👍 code looks fine to to me, didn't test though since it is "modified only"

aleister09:
It could be the same for WebLogin, if you want I can change it

Also It would be better load the WebError from RAW to have a WebError by each language, what do you think?

mario:
Yes, we could/should support languages, and default to english if one isn't available.

Same for login.

Feel free to commit to PR and improve it.

mario:
👍  looks fine, but @AndyScherzinger can test it :P

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -336,6 +339,22 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
...
+                try {
+                    while (( line = buffreader.readLine()) != null) {
+                        text.append(line);
+                        text.append('\n');
+                    }
+                } catch (IOException e) {
+                    Log_OC.e(TAG,e.getMessage());
+                    return;
+                }
+                mLoginWebView.loadData(text.toString(),"text/html; charset=UTF-8", null);
```
AndyScherzinger: @aleister09 @tobiasKaminsky @mario the ```onReceiveError(...)``` in both classes is basically the same code except the ```...loadData(...``` call which I think could be ```view.loadData(...``` which is a method input parameter. So I woudl have guessed this code could be centralized.

AndyScherzinger:
Just one minor issue related to duplicated code :)

aleister09:
I changed the duplicated code to Utils

mario:
👍

AndyScherzinger:
:+1:

tobiasKaminsky:
Thank you for your help @aleister09 .
Just a minor enhancement, it is better to set the line wrap to 120 as this makes it much more easier to review changes/diffs here on github, otherwise you have to scroll horizontally.

Issue #885, opened 2017-04-24
Crash after enable backup contacts 


aleister09:
### Steps to reproduce
1. Open App
2. Enable Contacts Backup
3. Logout
4. Close App
5. Open App

### Environment data
Android version: 7.1.2

Device model: Nexus 5X

Stock or customized system:

Nextcloud app version:Master

Nextcloud server version:12

### Logs

```
04-24 15:15:57.020 25218-25218/com.nextcloud.client E/AndroidRuntime: FATAL EXCEPTION: main
                                                                      Process: com.nextcloud.client, PID: 25218
                                                                      java.lang.RuntimeException: Unable to start activity ComponentInfo{com.nextcloud.client/com.owncloud.android.ui.activity.FileDisplayActivity}: java.lang.NullPointerException: Attempt to read from field 'java.lang.String android.accounts.Account.name' on a null object reference
                                                                          at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2666)
                                                                          at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2727)
                                                                          at android.app.ActivityThread.-wrap12(ActivityThread.java)
                                                                          at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1478)
                                                                          at android.os.Handler.dispatchMessage(Handler.java:102)
                                                                          at android.os.Looper.loop(Looper.java:154)
                                                                          at android.app.ActivityThread.main(ActivityThread.java:6121)
                                                                          at java.lang.reflect.Method.invoke(Native Method)
                                                                          at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:889)
                                                                          at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:779)
                                                                       Caused by: java.lang.NullPointerException: Attempt to read from field 'java.lang.String android.accounts.Account.name' on a null object reference
                                                                          at com.owncloud.android.ui.activity.ContactsPreferenceActivity.startContactBackupJob(ContactsPreferenceActivity.java:306)
                                                                          at com.owncloud.android.ui.activity.FileActivity.checkContactsBackupJob(FileActivity.java:252)
                                                                          at com.owncloud.android.ui.activity.FileActivity.onCreate(FileActivity.java:167)
                                                                          at com.owncloud.android.ui.activity.FileDisplayActivity.onCreate(FileDisplayActivity.java:168)
[...]

PR #907, opened 2017-04-27
Contact backup if no account

tobiasKaminsky:
Ref: #885 
- NPE right after start
- start daily backup job only if account exists


mario:
👎 

This is a flawed fix. A real fix is to cancel the job once an account is removed.

AndyScherzinger:
@tobiasKaminsky can the branch be safely deleted then? :)

mario:
Yes, we fixed this already :)

PR #915, opened 2017-04-28
"And" Missing

privatereese:
I think the "and" is missing, found it while translating via transifex and wanted to quickly fix it if its an error.


mario:
👍 
Looks ok. Thanks!

tobiasKaminsky:
👍 
Thank you!

PR #874, opened 2017-04-21
Contacts backup improvements/bugfixes

mario:
This PR will fix the following things once completely ready:

- refactor contacts backup into two separate fragments
- fix inability to scroll in the backup fragment
- fix fragment changes
- fix rotation by remembering the state
- fix contacts selection (currently broken)
- fix several bugs that cause the application to crash
- various other issues

PR #916, opened 2017-04-28
Push improvements

mario:
This PR improves push in a way that allows unregistering from both the server and push proxy after account delete.

It depends on #874 being accepted.


mario:
Should be ok, but untested atm :)

tobiasKaminsky:
Is there some way that I can test this?

tobiasKaminsky:
Drone has errors and I guess that this needs a rebase as lots of Contacts* stuff is in it.

mario:
Easier for me to test - so I'd appreciate if you could do a code review...

But I have to see what's up with build...

src/gplay/java/com/owncloud/android/utils/PushUtils.java
```diff
@@ -124,8 +135,58 @@ public static int generateRsa2048KeyPair() {
...
+        try {
+            ocAccount = new OwnCloudAccount(account, context);
+            OwnCloudClient mClient = OwnCloudClientManagerFactory.getDefaultSingleton().
+                    getClientFor(ocAccount, context);
+
+            RemoteOperation unregisterAccountDeviceForNotificationsOperation = new
+                    UnregisterAccountDeviceForNotificationsOperation();
+
+            RemoteOperationResult remoteOperationResult = unregisterAccountDeviceForNotificationsOperation.
+                    execute(mClient);
```
tobiasKaminsky: https://github.com/nextcloud/android-library/blob/master/src/com/owncloud/android/lib/common/operations/RemoteOperation.java#L101 is suggesting to rather use execute(Account, Context)

With this you can remove L144 and 145
Especially L145 had caused some troubles in the past...

src/gplay/java/com/owncloud/android/utils/PushUtils.java
```diff
@@ -124,8 +135,58 @@ public static int generateRsa2048KeyPair() {
...
+                    Gson gson = new Gson();
+                    PushArbitraryData pushArbitraryData = gson.fromJson(arbitraryValue, PushArbitraryData.class);
+                    RemoteOperation unregisterAccountDeviceForProxyOperation =
+                            new UnregisterAccountDeviceForProxyOperation(context.getResources().
+                                    getString(R.string.push_server_url),
+                                    pushArbitraryData.getDeviceIdentifier(),
+                                    pushArbitraryData.getDeviceIdentifierSignature(),
+                                    pushArbitraryData.getUserPublicKey());
+
+                    remoteOperationResult = unregisterAccountDeviceForProxyOperation.execute(mClient);
```
tobiasKaminsky: same here

src/gplay/java/com/owncloud/android/utils/PushUtils.java
```diff
@@ -139,42 +200,56 @@ public static void pushRegistrationToServer() {
...
+                                OwnCloudClient mClient = OwnCloudClientManagerFactory.getDefaultSingleton().
+                                        getClientFor(ocAccount, context);
+
+                                RemoteOperation registerAccountDeviceForNotificationsOperation =
+                                        new RegisterAccountDeviceForNotificationsOperation(pushTokenHash,
+                                                publicKey,
+                                                context.getResources().getString(R.string.push_server_url));
+
+                                RemoteOperationResult remoteOperationResult = registerAccountDeviceForNotificationsOperation.
+                                        execute(mClient);
```
tobiasKaminsky: and here

src/main/java/com/owncloud/android/datamodel/PushArbitraryData.java
```diff
@@ -0,0 +1,81 @@
...
+    public String pushToken;
+    public String deviceIdentifier;
+    public String deviceIdentifierSignature;
+    public String userPublicKey;
+    public boolean shouldBeDeleted;
+
+    public PushArbitraryData() {
+    }
+
+    public PushArbitraryData(String pushToken, String deviceIdentifier, String deviceIdentifierSignature,
```
tobiasKaminsky: This is not arbitrary Data, although it is stored in a table named this way, but it is more like PushNotificationSettings or similar? The current name is somehow misleading.

src/main/java/com/owncloud/android/ui/activity/FileActivity.java
```diff
@@ -162,8 +162,6 @@ protected void onCreate(Bundle savedInstanceState) {
 
         setAccount(account, savedInstanceState != null);
 
-        checkContactsBackupJob();
```
tobiasKaminsky: Sure that this should be removed?

src/main/java/com/owncloud/android/db/PreferenceManager.java
```diff
@@ -47,33 +47,15 @@
...
-        saveLongPreference(context, PREF__PUSH_TOKEN_LAST_REGISTRATION_TIME, time);
-    }
-
-    public static long getPushTokenLastSentTime(Context context) {
-        return getDefaultSharedPreferences(context).getLong(PREF__PUSH_TOKEN_LAST_REGISTRATION_TIME, -1);
-    }
 
     public static void setPushToken(Context context, String pushToken) {
-        saveStringPreference(context, PREF__PUSH_TOKEN, pushToken);
+        saveStringPreferenceNow(context, PREF__PUSH_TOKEN, pushToken);
```
mario: Not atm. Maybe later, yes.

src/gplay/java/com/owncloud/android/utils/PushUtils.java
mario: This worked ok for me in all cases (I use it elsewhere without a problem as well). What problems? :)

src/main/java/com/owncloud/android/datamodel/PushArbitraryData.java
mario: How is it not arbitrary data? "Arbitrary Data" -> "Any data", and this seems to fit the definition? :)

src/main/java/com/owncloud/android/ui/activity/FileActivity.java
mario: Looks like it. It basically started/overwrote an already running period job? Why would we do that on every startup?

src/gplay/java/com/owncloud/android/utils/PushUtils.java
tobiasKaminsky: It crashed during client fetching on a separate thread.
But if it is working... ;-)

src/main/java/com/owncloud/android/ui/activity/FileActivity.java
tobiasKaminsky: Hm. What about phone restart?

mario: It should run anyway, since it is:

- scheduled periodically
- is persisted (from android-job 1.2.0, it will always be persisted) meaning it will run on boot

src/main/java/com/owncloud/android/datamodel/PushArbitraryData.java
mario: Ok, agreed. I will change this.

src/main/java/com/owncloud/android/ui/activity/FileActivity.java
tobiasKaminsky: Ok, then I am fine with removing it.
(I assume that you have also tested it?)

mario:
Updating according to review and merging.

tobiasKaminsky:
According to this #934, it would have been better to wait for another round of confirmation ;-)

Issue #911, opened 2017-04-27
PreviewFile is unavailable 


aleister09:
### Actual behaviour
The app can't show the preview if the file was uploaded by AutoUpload

### Steps to reproduce
1. Open App
2. Open AutoUpload
3. Enable a Folder for AutoUpload
4. Close App
5. Add a file in the folder and Wait until the upload finish
6. Open App
7. Find the file and open it 

### Environment data
Android version:7.1.2

Device model: Nexus 5X

Nextcloud app version:Master

Nextcloud server version:12

![device-2017-04-27-174623](https://cloud.githubusercontent.com/assets/25756086/25507404/7d3bd096-2b71-11e7-80f4-e61be81b4ea2.png)

tobiasKaminsky:
@aleister09 please confirm as the fix is merged to master.

aleister09:
It's working fine =)

tobiasKaminsky:
Thank you for your feedback.

PR #918, opened 2017-04-29
AutoUpload: default to "keep file in original folder"

tobiasKaminsky:
Ref: #911 

The default, if you enable a synced folder but do not choose any upload behaviour was "0", which then lead to a green "downloaded sign" but no file available.
The default now is "keep in folder" which is in sync with the UI.


AndyScherzinger:
👍 Looking good

mario:
👍

Issue #890, opened 2017-04-25
Activities and Notification Crash


aleister09:
### Actual behaviour
The App crashes in Activities and Notification Section when the user rotates the app when is loading 

### Steps to reproduce
1.Open App
2.Open Activities or Notifications
3.Swipe to refresh
4.Rotate the App when is loading

### Environment data
Android version:7.1.2

Device model: Nexus 5X

Stock or customized system:

Nextcloud app version:Master

Nextcloud server version:12

### Logs
```
04-25 16:55:53.066 18410-18410/com.nextcloud.client E/AndroidRuntime: FATAL EXCEPTION: main
                                                                      Process: com.nextcloud.client, PID: 18410
                                                                      java.lang.NullPointerException: Attempt to invoke virtual method 'void android.support.v4.widget.SwipeRefreshLayout.setVisibility(int)' on a null object reference
                                                                          at com.owncloud.android.ui.activity.ActivitiesListActivity$3$1.run(ActivitiesListActivity.java:200)
                                                                          at android.os.Handler.handleCallback(Handler.java:751)
                                                                          at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                          at android.os.Looper.loop(Looper.java:154)
                                                                          at android.app.ActivityThread.main(ActivityThread.java:6121)
                                                                          at java.lang.reflect.Method.invoke(Native Method)
                                                                          at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:889)
                                                                          at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:779)
```

mario:
Should be fixed.

PR #921, opened 2017-04-30
NPE during Activity orientation change

tobiasKaminsky:
Ref: #890 

Butterknife is causing a problem while rotation.
Currently the only way I found working is to not use it on swipeViews :/
If someone can resolve the real cause of the butterknifes problem I would be happy.

(if not and we want to do it this way, the same changes needs to be done for notification view)


mario:
Can you please do the same change for notifications in this PR? Thanks.

mario:
👍

PR #927, opened 2017-05-02
prevent crashing if fingerprint is not supported

tobiasKaminsky:
Settings crashed on < M

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/Preferences.java
+++ b/src/main/java/com/owncloud/android/ui/activity/Preferences.java
@@ -189,41 +189,49 @@ public class Preferences extends PreferenceActivity
         boolean fPrintEnabled = getResources().getBoolean(R.bool.fingerprint_enabled);
         fPrint = (CheckBoxPreference) findPreference(FingerprintActivity.PREFERENCE_USE_FINGERPRINT);
         if (fPrint != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -189,41 +189,49 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
...
-                                        .show();
-                                fPrint.setChecked(false);
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                if (FingerprintActivity.isFingerprintCapable(MainApp.getAppContext()) && fPrintEnabled) {
+                    fPrint.setOnPreferenceChangeListener(new OnPreferenceChangeListener() {
+                        @Override
+                        public boolean onPreferenceChange(Preference preference, Object newValue) {
+                            Boolean incoming = (Boolean) newValue;
+
+                            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
AndyScherzinger: Line 192 already checked for build version :)

AndyScherzinger:
👍 fine with me, just found a duplicate check

src/main/java/com/owncloud/android/ui/activity/Preferences.java
mario: Android studio would complain otherwise xD

mario:
👍

Issue #888, opened 2017-04-25
PreviewMedia is closed when rotate


aleister09:
### Actual behaviour
PreviewMediaFragment is replaced by FileDetailFragment when the user rotate the app
I happens because in the method ```FileDisplayActivity.chooseInitialSecondFragment()``` the property ```file.getLastSyncDateForProperties()``` is ```0``` and return FileDetailFragment
### Steps to reproduce
1. Open App
2. Open Favorites Section
3. Open a Video
4. Rotate the App

### Environment data
Android version: 7.1.2

Device model: Nexus 5X

Nextcloud app version: Master

Nextcloud server version:12

![ezgif com-video-to-gif 4](https://cloud.githubusercontent.com/assets/25756086/25397449/ecf5e6b2-29ad-11e7-9abb-3b4f66dd2096.gif)

mario:
Thanks for the report. Wanna fix? :)

mario:
@aleister09 also, do you think you could mail me? I need assistance with something since I see you have 7.1 :) (mario AT nextcloud DOT com)

aleister09:
Sure!

I can fix it but i don't understand why the method check getLastSyncDateForProperties also the line has the comment ```// temporal fix``` I think it isn't necessary now, what do you think?

mario:
Never touched that part .. @AndyScherzinger or @tobiasKaminsky , maybe you know what's this for? :)

AndyScherzinger:
Unfortunately no, never worked on that part

tobiasKaminsky:
This was introduced back in the old days with this commit message "Fixed: details about progress are shown when the notification of an upload is clicked"
https://github.com/nextcloud/android/commit/19bcc995386f6fbd0803d2a5bb506e8a40686b3c

PR #896, opened 2017-04-26
Search view wrong files

tobiasKaminsky:
Ref: #857 

The main clue is that the current system does not allow to have "any" ocfile, instead there must always be a parentId/parentFolder. Without that things like downloading files do not work.
As the favorites/photos can show any files (even if they have never been synced) we need to download the parentOCFile, maybe even recursive until "/".
Of course this is only done once for each folder.

Sidenote: I tried it with ~3500 and it took quite a long time to display the images. Fetching the files recursively was not the problem as on the second time all files were fetched already, but it still took ~20s.

PR #928, opened 2017-05-02
Fix bug #888

aleister09:
This PR fixes issue #888 and I found another issue in favorites and photos(Filter Sections) and I fixed it

### DIFF
--- a/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
+++ b/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
@@ -250,10 +250,6 @@ public class OCFileListFragment extends ExtendedListFragment implements OCFileLi
     public void onResume() {
         super.onResume();
 
-        if (remoteOperationAsyncTask != null) {
-            remoteOperationAsyncTask.cancel(true);
-        }
-
         if (getActivity() != null) {
             AnalyticsUtils.setCurrentScreenName(getActivity(), SCREEN_NAME, TAG);
         }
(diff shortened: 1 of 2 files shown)


src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -250,10 +250,6 @@ public void run() {
     public void onResume() {
         super.onResume();
 
-        if (remoteOperationAsyncTask != null) {
```
mario: This IHMO shouldn't be here. Why do we remove it? Async tasks should always be cancelled - thought I'd rather have them cancelled in onPause than onResume.

What does this fixes?

aleister09: This code cancel the Async Task after it is started and when the remote operation finish the async task asks if it was canceled and can't set the response
```
protected Object doInBackground(Object[] params) {
                if (getContext() != null && !isCancelled()) {
                    RemoteOperationResult remoteOperationResult = remoteOperation.execute(currentAccount, getContext());

                    if (remoteOperationResult.isSuccess() && remoteOperationResult.getData() != null
                            && !isCancelled() && searchFragment) {
                        mAdapter.setData(remoteOperationResult.getData(), currentSearchType);
                    }

                    return remoteOperationResult.isSuccess();
                } else {
                    return false;
                }
            }
```

mario:
@tobiasKaminsky well, it certainly solved the problem, but is it the right fix? You decide...

src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
tobiasKaminsky: This is somehow outdated as I have changed this again in https://github.com/nextcloud/android/pull/896
So we can ignore this here.
:+1: for the rest

mario:
@tobiasKaminsky so you want to approve as-is? No changes needed? We can just merge?

tobiasKaminsky:
I reverted this as it might get confusing.
So this PR is only about #888

tobiasKaminsky:
:+1:

mario:
👍

PR #936, opened 2017-05-03
Fix fab stuff

mario:
This method is sometimes called from background, so it wouldn't work. This fixes it.


tobiasKaminsky:
👍

AndyScherzinger:
👍

Issue #809, opened 2017-04-06
Preview Image is crashing 


aleister09:
### Actual behaviour
-The app is crashing in the Preview Image Fragment   with any image prior to Android API 23 

### Expected behaviour
-The app should show the preview image
 
### Steps to reproduce
1. Open the app 
2. Login
3. Choose an image 

### Environment data
Android version:5.0 and 4.4.2

Device model: Samsung Galaxy Tab SM-T530 and Moto X 1º Gen, 

Stock or customized system: 

Nextcloud app version: master

Nextcloud server version: 11

### Logs

```
`04-06 13:01:42.018 799-799/com.nextcloud.client E/AndroidRuntime: FATAL EXCEPTION: main
                                                                  Process: com.nextcloud.client, PID: 799
                                                                  java.lang.NoSuchMethodError: No virtual method setLayerHeight(II)V in class Landroid/graphics/drawable/LayerDrawable; or its super classes (declaration of 'android.graphics.drawable.LayerDrawable' appears in /system/framework/framework.jar:classes2.dex)
                                                                      at com.owncloud.android.ui.preview.PreviewImageFragment$LoadBitmapTask.showLoadedImage(PreviewImageFragment.java:532)
                                                                      at com.owncloud.android.ui.preview.PreviewImageFragment$LoadBitmapTask.onPostExecute(PreviewImageFragment.java:505)
                                                                      at com.owncloud.android.ui.preview.PreviewImageFragment$LoadBitmapTask.onPostExecute(PreviewImageFragment.java:400)
                                                                      at android.os.AsyncTask.finish(AsyncTask.java:632)
                                                                      at android.os.AsyncTask.access$600(AsyncTask.java:177)
                                                                      at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:645)
                                                                      at android.os.Handler.dispatchMessage(Handler.java:102)
                                                                      at android.os.Looper.loop(Looper.java:145)
                                                                      at android.app.ActivityThread.main(ActivityThread.java:6134)
                                                                      at java.lang.reflect.Method.invoke(Native Method)
                                                                      at java.lang.reflect.Method.invoke(Method.java:372)
                                                                      at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1399)
                                                                      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1194)
`
[...]

PR #941, opened 2017-05-03
remove suppressLint

tobiasKaminsky:
Checkerboard on PNG files is only shown in >= M and thus toggle Background is also used in >= M only

Ref #809


AndyScherzinger:
👍 code looks good to me

mario:
👍

Issue #807, opened 2017-04-06
External storage is in /data/ and therefore not scannable


tobiasKaminsky:
When changing storage path to external it is stored in Android/data/com.nextcloud.client/files and not scannable.
(reported by mail as a regression)

PR #947, opened 2017-05-03
Enhance external Sdcard usage

tobiasKaminsky:
- use externalMediaDir to allow downloaded files to be scanned again
- do not show "unknown" storage path, instead show complete path as fallback

#807

### DIFF
--- a/src/main/java/com/owncloud/android/datastorage/providers/SystemDefaultStoragePointProvider.java
+++ b/src/main/java/com/owncloud/android/datastorage/providers/SystemDefaultStoragePointProvider.java
@@ -21,13 +21,14 @@
 
 package com.owncloud.android.datastorage.providers;
 
-import android.os.Environment;
-
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
 import com.owncloud.android.datastorage.StoragePoint;
 
+import java.io.File;
 import java.util.Vector;
+
+import static android.os.Environment.getExternalStorageDirectory;
 
 /**
  * @author Bartosz Przybylski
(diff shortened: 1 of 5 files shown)


src/main/java/com/owncloud/android/datastorage/providers/SystemDefaultStoragePointProvider.java
```diff
@@ -44,8 +45,16 @@ public boolean canProvideStoragePoints() {
 
         final String defaultStringDesc =
                 MainApp.getAppContext().getString(R.string.storage_description_default);
-        final String path = Environment.getExternalStorageDirectory().getAbsolutePath();
-        result.add(new StoragePoint(defaultStringDesc, path));
+        File path;
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP) {
+            path = MainApp.getAppContext().getExternalMediaDirs()[0];
```
AndyScherzinger: can ```getExternalMediaDirs()``` be empty or null?

mario: Also if we do this, should we stop inserting files into media provider manually?

mario: Also yes, it could be null if say SD card is unavailable atm.

tobiasKaminsky: Yes, therefore I am testing if path != null in the next line.
We can also enhance this and see if it is writeable.

mario: Sure. Let me know when it's ready for testing & review.

tobiasKaminsky:
Added a check if writable

mario:
👍  Looks ok-ish.

AndyScherzinger:
👍 same here

remaining question though from @mario 

> Also if we do this, should we stop inserting files into media provider manually?

mario:
@tobiasKaminsky just answer the Q when you get the time ;)

Issue #814, opened 2017-04-07
Download fails if download folder is deleted


tobiasKaminsky:
When the storage path is on sdcard and the sdcard is changed/removed download will fail.
There should be shown a correct error message like "storage path does not exist any more" and then lead the user to the settings screen.

PR #948, opened 2017-05-03
if source folder does not exist, still allow to move to new folder

tobiasKaminsky:
This is useful if you have switched to use the external sdcard and then removed it.

filepj61fw

Ref: #814

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/StorageMigration.java
+++ b/src/main/java/com/owncloud/android/ui/activity/StorageMigration.java
@@ -221,10 +221,9 @@ public class StorageMigration {
         }
 
         private void askToStillMove() {
-
             new AlertDialog.Builder(mContext)
-                    .setMessage("Source directory not readable. Do you still want to change the storage path to "
-                            + mStorageTarget + "? Note: all data will have to be downloaded again.")
+                    .setTitle(R.string.file_migration_source_not_readable_title)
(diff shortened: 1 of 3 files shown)


src/main/java/com/owncloud/android/ui/activity/StorageMigration.java
```diff
@@ -205,15 +205,44 @@ protected void onPostExecute(Integer code) {
...
             if (mListener != null) {
                 mListener.onStorageMigrationFinished(succeed ? mStorageTarget : mStorageSource, succeed);
             }
         }
 
+        private void askToStillMove() {
+
+            new AlertDialog.Builder(mContext)
+                    .setMessage("Source directory not readable. Do you still want to change the storage path to "
+                            + mStorageTarget + "? Note: all data will have to be downloaded again.")
```
AndyScherzinger: Message needs to be extracted to string.xml :)

AndyScherzinger: Source directory not readable - could be used as a title :)

tobiasKaminsky: filejvnj84

mario:
👍

AndyScherzinger:
👍

PR #953, opened 2017-05-04
Fix issue with wifi

AndyScherzinger:
@mario is this branch/PR still valid and if so should it be part of 1.5.0?

cc: @tobiasKaminsky as a ping :)


mario:
@AndyScherzinger this is something we'll have to do, but differently. Please kill the branch.

PR #954, opened 2017-05-04
No gcm

AndyScherzinger:
@mario @tobiasKaminsky is this PR still needed/valid? If I understood correctly with the sheduler lib 1.0.19 and all the latest changes already available on master still PR is obsolete and the branch can be deleted, right?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,6 +49,7 @@ repositories {
     }
 }
 
+
 android {
     lintOptions {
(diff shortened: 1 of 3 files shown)


mario:
Google Play app will ship GCM (other reasons though, Push namely) - but FDroid was made to run without it. We can delete this PR and the branch. Please do it :)

Issue #959, opened 2017-05-05
Crash After Rotation


aleister09:
### Actual behaviour
The app is crashing after rotate many times when quota is enable

### Steps to reproduce
1. Open App
2. Open DrawerMenu
3. Rotate many times 

### Environment data
Android version:7.1.2

Device model: Nexus 5X

Nextcloud app version:Master

Nextcloud server version:12

### Logs

```
05-05 17:45:28.609 26313-26313/com.claro.sync E/AndroidRuntime: FATAL EXCEPTION: main
                                                                Process: com.claro.sync, PID: 26313
                                                                java.lang.IllegalArgumentException: You cannot start a load for a destroyed activity
                                                                    at com.bumptech.glide.manager.RequestManagerRetriever.assertNotDestroyed(RequestManagerRetriever.java:134)
                                                                    at com.bumptech.glide.manager.RequestManagerRetriever.get(RequestManagerRetriever.java:102)
                                                                    at com.bumptech.glide.manager.RequestManagerRetriever.get(RequestManagerRetriever.java:87)
                                                                    at com.bumptech.glide.Glide.with(Glide.java:629)
                                                                    at com.owncloud.android.utils.DisplayUtils.downloadSVGIcon(DisplayUtils.java:562)
                                                                    at com.owncloud.android.utils.DisplayUtils.downloadIcon(DisplayUtils.java:543)
                                                                    at com.owncloud.android.ui.activity.DrawerActivity.updateQuotaLink(DrawerActivity.java:877)
                                                                    at com.owncloud.android.ui.activity.DrawerActivity.setQuotaInformation(DrawerActivity.java:811)
                                                                    at com.owncloud.android.ui.activity.DrawerActivity.access$500(DrawerActivity.java:87)
                                                                    at com.owncloud.android.ui.activity.DrawerActivity$6$1.run(DrawerActivity.java:934)
                                                                    at android.os.Handler.handleCallback(Handler.java:751)
                                                                    at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                    at android.os.Looper.loop(Looper.java:154)
                                                                    at android.app.ActivityThread.main(ActivityThread.java:6121)
                                                                    at java.lang.reflect.Method.invoke(Native Method)
                                                                    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:889)
                                                                    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:779)
```

[...]

PR #960, opened 2017-05-06
prevent using destroyed activity

tobiasKaminsky:
Glide refuses to work on a destroyed activity, catching Exception.
No need to handle it better as this happens on the old activity while rotation and a new one is already in place where Glide correctly shows/loads the icons.

Ref: #959


mario:
👍

AndyScherzinger:
:+1:

PR #977, opened 2017-05-10
Add support for folder favorite action

### LINKED ISSUE #973: Setting folder as favourite not possible
* open android app
* select a folder
* expected: allow to set it as favourite 
* actual: no way to set as favourite (but for files this is possible)

cc @mario @tobiasKaminsky @AndyScherzinger

mario:

This fixes #973.

### DIFF
--- a/src/main/java/com/owncloud/android/files/FileMenuFilter.java
+++ b/src/main/java/com/owncloud/android/files/FileMenuFilter.java
@@ -243,14 +243,14 @@ public class FileMenuFilter {
         }
 
         // Favorite
-        if (!canBeFavorited() || synchronizing || allFavorites()) {
+        if (mFiles.isEmpty() || synchronizing || allFavorites()) {
             toHide.add(R.id.action_favorite);
         } else {
             toShow.add(R.id.action_favorite);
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/files/FileMenuFilter.java
```diff
@@ -243,14 +243,14 @@ private void filter(List<Integer> toShow, List<Integer> toHide) {
         }
 
         // Favorite
-        if (!allFiles() || synchronizing || allFavorites()) {
+        if (!canBeFavorited() || synchronizing || allFavorites()) {
```
tobiasKaminsky: To keep in sync with e.g. https://github.com/nextcloud/android/blob/d3ea3a8ccf404ea2499a57c54b0884e61f0ba1c1/src/main/java/com/owncloud/android/files/FileMenuFilter.java#L158
it is better to use "!mFiles.isEmpty" as the name "canBeFavorited" is misleading as you do not test is (in fact every existing file is favoritable)

tobiasKaminsky:
:+1:

Issue #984, opened 2017-05-11
Audio starts playing on device rotation


strugee:
STR:

1. Open an audio file
2. Wait for it to start playing
3. Pause
4. Rotate device
5. Audio starts playing, even though it shouldn't

Nexus 6P, CyanogenMod 13, dev version 20170430.

tobiasKaminsky:
Thank you for your bug report!
This also applies to the regular version, thus removing dev label.
It will be soon in the dev version.
If you have time, please try it and report if it working.

strugee:
Sure thing @tobiasKaminsky :)

By "soon" I'm assuming you mean "in the next release"? Happy to test this whenever a fix gets merged.

tobiasKaminsky:
"soon" depends on fdroid.
Here you can find it without delay: https://download.nextcloud.com/android/nightly/

squadramunter:
Hi I got a bug on my android phone. I have Cyanogenmod installed on it but when i open the app the video resolution is not correct. when rotating the screen it's not fitted full screen. may a feature for the next release to add support for aspect ratio. I have a samsung galaxy s5.

strugee:
I can still reproduce with the 20170511 build.

tobiasKaminsky:
Sorry, used the wrong label for the PR and thus it was not merged automatically.
This is now merged into "dev version" and you can download it this night.

strugee:
Whoops, sorry - totally forgot about this issue! I can confirm that this is fixed :)

Thanks for working on this <3

PR #985, opened 2017-05-12
Re-use second fragment if available

tobiasKaminsky:
Ref: #984 

Please test it very extensive as I am not 100% sure about any side-effects!

(also did some minor code formatting changes during debug. Always leave code in a better state ;-))


mario:
@AndyScherzinger will test this :P

tobiasKaminsky:
Better to have it tested by both of you and all the users that use dev version.

AndyScherzinger:
👍 code looks good and fixes the rotation bug, downside I found out testing the PR is that for some reason we seem to have lost the "audio layout" somewhere before this PR since an audio file is now also using the standard video preview while before we had a nicer version for audio with a styled progressbar :(

AndyScherzinger:
Screenshots

1.4.2 stable
device-2017-05-15-230043

1.5.0-master
device-2017-05-15-224721

MorrisJobke:
@aleister09 Could you test and review this as well? I think you also noticed this recently.

aleister09:
With rotation is working fine but if you turn off the screen and then turn on the music starts again

tobiasKaminsky:
> With rotation is working fine but if you turn off the screen and then turn on the music starts again

I could reproduce this on master but not on this branch.
What I did:
- start song
- pause song
- switch off display
- turn display on
--> still song paused

AndyScherzinger:
👍 Looks good!

mario:
👍

Issue #730, opened 2017-03-09
App forces logout after sharing a file with the same user twice


togglecode:
### Actual behaviour
-App forces logout after sharing a file (intern) with a user twice

### Expected behaviour
-The user should not be shown in the search list after you have added them in shared.
 
### Steps to reproduce
1. Open App
2. Long press on a file -> share 
3. Add user or Group
4. Search for a user (Let's say user "example")
5. Add the user "example" as a person which you want to share the file with
6. Search again for the user
7. Try to add him again
8. App logs out

### Environment data
Android version: 6.0.1

Device model: Moto G3

Stock or customized system: customized

Nextcloud app version: 1.4.2 RC4

Nextcloud server version: 11.0.1.2

AndyScherzinger:
Didn't have the change yet to look for the cause but tested the issue and can confirm the logout.

AndyScherzinger:
fixed for release 1.5.0

PR #986, opened 2017-05-12
Do nothing if sharee already exists

tobiasKaminsky:
This is merely a workaround as I do not figured out how to do it better:
The current approach shows all sharees, but if a sharee already exists a click on its name simply does nothing.

Better/right approach: Filter the sharee suggestion list from existing sharees.
Please feel free to replace this PR with the better approach.

But now it solves #730, which is an improvement ;-)


mario:
👍 

Unhappy, but ok ... :)

AndyScherzinger:
👍 also fine with this (temporary) fix :)

AndyScherzinger:
@tobiasKaminsky feel free to hit merge ;)

mario:
Isn't there a song ... when temporarily becomes permanent... :D

Issue #998, opened 2017-05-15
Transifex: 'quote' character should be escaped for Android String resource


AndyScherzinger:
```
Review, amend and write code, see 
&lt;a href="https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md">CONTRIBUTING.md&lt;/a> 
for details
```
can't be translated on transifex with the error:
```
'quote' character should be escaped for Android String resource
```

any idea @tobiasKaminsky , since you made the last config changes to transifex :)
Alternatively we could go for 
```
Review, amend and write code, see %1$s for details
```
and add the hyperlink via string formatting in code...

tobiasKaminsky:
I have enabled a check that verifies that exery UTF8-char from source is also translated into destination string.
The check is described as "Content of placeholder delimiters &...;  is preserved in the translation."

Your idea to extarct the hyperlink to code is cleaner as then the url in setup.xml can be used.

AndyScherzinger:
Right, didn't think about ```setup.xml``` but that is nicer and more flexible too 👍

tobiasKaminsky:
I read it completely wrong: it is not about utf8-chars but about " (quote sign) which is suggested to be quoted.
Anyway, a PR is here: #1001

AndyScherzinger:
@comradekingu the last remaining en_GB string will be translatable tomorrow. Thanks for the hint 👍

PR #1001, opened 2017-05-15
Minor improvements to participate view

tobiasKaminsky:
- Use contribute url from setup.xml
- Escape quote signs

Ref: #998


mario:
👍

tobiasKaminsky:
lint reports
`../../../src/main/res/values/strings.xml:567: Inconsistent number of arguments in formatting string participate_contribute_github_text; found both 0 and 1`

which is due to the fact that it is not translated correctly right now.

If you both agree, I will increase the error count to 75 and then we can merge it.

AndyScherzinger:
Agreed! 👍

AndyScherzinger:
👍 successfully tested 👍 

@mario your turn ;)

mario:
👍

PR #1002, opened 2017-05-15
use 1.0.17 again

tobiasKaminsky:


mario:
👍

tobiasKaminsky:
We need a new version as otherwise build will fail as the changes from logFolder branch are missing.

Issue #1005, opened 2017-05-15
AutoUpload tries to upload although mobile data is turned off


tobiasKaminsky:
- turn off mobile data
- take some photos
- wait > 30s
- auto upload tries to upload pictures and fails with "connection error"

There was a PR for the "old" auto upload.
This issue remains open to check this if it is working after rewite by mario.

mario:
Dear users,

thank you for your patience. After lots of work, we’re super excited to be able to give you a test version with new auto upload implementation. Keep in mind we will only accept bugs related to Auto Upload in this APK, as other things might be broken.

https://cloud.nextcloud.com/s/ftZraTruRQjTe0R

This is a dev APK so it will not overwrite your existing play store/f-droid APK unless you’re already running dev version. Also, you might need to remove your existing dev version before installing this one.

Explanation of the current functionality:
- added support for custom folders
- separated image and video folders
- checks for new files happens every 15-20 minutes
- automatic retry every 15-20 minutes or when you force close/open the app
- automatic pause once network is changed (untested!)
- significantly less battery usage

All and any testing is more than appreciated, especially with large and ever-going files like Videos. (like, try to record a 30 min video).

Let me know if you’ve got any questions.

Cheers,
Mario

mario:
This is most definitely fixed.

PR #1006, opened 2017-05-15
Use network constraint "connected" as default

tobiasKaminsky:
ANY: Network must not be connected.
CONNECTED: Network must be connected.
Ref: #1005


mario:
👎 

I'd have to look into this first, but this particular piece of code will be removed/replaced for 1.5 anyway ... :)

Issue #971, opened 2017-05-09
Allow to click the entries in Activity to show them in the app


MorrisJobke:
* refinement to the existing activity view
* maybe also show the preview of the files

cc @nextcloud/designers @oparoz @mario @aleister09 @tobiasKaminsky @AndyScherzinger

aleister09:
But what about when the activity entry has more than one file?
Like this
![more files](https://cloud.githubusercontent.com/assets/25756086/25909071/9ef1653e-3571-11e7-8b99-c8849bdb6794.png)

MorrisJobke:
> But what about when the activity entry has more than one file?

Of course the file name itself should be clickable.

Ultimate goal is to have something like this on Android as well:

<img width="502" alt="bildschirmfoto 2017-05-10 um 11 41 20" src="https://cloud.githubusercontent.com/assets/245432/25910343/a64ed682-3575-11e7-8beb-35f2cf3845db.png">

AndyScherzinger:
implemented and merged

PR #1012, opened 2017-05-16
Activities improvement

aleister09:
This PR fixes the issue #971 
Now the view looks like this

device-2017-05-16-101847


src/main/res/layout/activity_list_item.xml
```diff
@@ -61,6 +61,11 @@
             android:text="@string/placeholder_sentence"
             android:textColor="?android:attr/textColorSecondary"/>
 
+        <android.support.v7.widget.RecyclerView
```
AndyScherzinger: not sure about this approach, it basically nests recyclerviews inside the recycler view

@mario @aleister09 what do you think? I am not a recyclerview expert but would expect that this would need to be handled within this list item layout but not by adding another recycler view

src/main/java/com/owncloud/android/ui/activity/ActivitiesListActivity.java
```diff
@@ -136,6 +140,9 @@ public void onRefresh() {
 
             }
         });
+
+
```
AndyScherzinger: please remove any unneeded, added empty lines

src/main/java/com/owncloud/android/ui/activity/ActivitiesListActivity.java
```diff
@@ -156,18 +163,16 @@ public void showFiles(boolean onDeviceOnly) {
      * sets up the UI elements and loads all activity items.
      */
     private void setupContent() {
+
         emptyContentIcon.setImageResource(R.drawable.ic_activity_light_grey);
         setLoadingMessage();
 
-        adapter = new ActivityListAdapter(this);
+        adapter = new ActivityListAdapter(this,this);
```
AndyScherzinger: please add a space after the comma

src/main/java/com/owncloud/android/ui/activity/ActivitiesListActivity.java
```diff
@@ -308,4 +313,18 @@ protected void onResume() {
 
         AnalyticsUtils.setCurrentScreenName(this, SCREEN_NAME, TAG);
     }
+
+
+    @Override
+    public void onActivityClicked(RichObject richObject) {
+        Intent showDetailsIntent;
+        OCFile ocFile=new OCFile("/"+richObject.getPath());
+        if(PreviewImageFragment.canBePreviewed(ocFile))
```
AndyScherzinger: please use brackets, also for single command blocks in if/else statements

AndyScherzinger:
really nice work @aleister09 , just some minor code formatting comments and one issue for discussion with @mario and @tobiasKaminsky about the use of nested recycler views

src/main/res/layout/activity_list_item.xml
mario: This can and needs to be handled by dynamically creating views rather than using recycle view.

aleister09: I think there's not problem with that, It's like Play Store with a huge RecyclerView for categories and each item with a RecyclerView for Apps 
But if you think it's wrong I can change it to do it dynamically

AndyScherzinger: Hmm, true, so if we have vertical scroll for the overall list and horizontal scrolling for the file items this could also be a solution while I am not sure if Google is actually nesting recycler views. Do you have any reference for that layouting in Google Play?

aleister09: I saw it  here https://github.com/google/iosched/blob/master/android/src/main/java/com/google/samples/apps/iosched/explore/ExploreIOFragment.java

It has a RecyclerView and each item is a CardView and the CardView has a RecyclerView

mario: If you know how to do it dynamically, I'd appreciate it :)

mario:
Commented  re: recycler view, and we need to check capabilities to see if we actually support v2 etc etc.

mario:
Afaik rich objects are part of Activities v2 api which is related to the activities app. In case where we're still on v1 activities, it just won't have rich objects but I do agree we need to query capabilities to see what we have. 

As for OCS v2, it was (I think) available in NC9 so I think we're safe there.

AndyScherzinger:
@nickvergessen can you tell since when v2 of the API has been available on the server side, release-wise?

nickvergessen:
12.0

AndyScherzinger:
Also needs to upgrade to lib version 1.0.20 :)

cc @mario

src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java
```diff
@@ -134,7 +134,8 @@ public void onBindViewHolder(RecyclerView.ViewHolder holder, int position) {
 
             if (!TextUtils.isEmpty(activity.getSubject())) {
                 activityViewHolder.subject.setMovementMethod(LinkMovementMethod.getInstance());
-                activityViewHolder.subject.setText(addClickablePart(activity.getRichElement()), TextView.BufferType.SPANNABLE);
+                activityViewHolder.subject.setText(addClickablePart(activity.getRichSubjectElement()),
```
AndyScherzinger: This still seems to be needing some more logic, is is checking activity.getSubject() and if this isn't null it uses activity.getRichSubjectElement(). So the code is still lacking the no-RichObjects fallback ;)

mario: Sure. I also have to do other stuff in this PR.

AndyScherzinger: Ah, okay. I wasn't sure if this PR is complete code wise 👍

mario:
@AndyScherzinger feel free to test this PR :)

AndyScherzinger:
@mario I am actually. Do you have a pre Nc12 instance/account for me so I can test the fallback? (since I only have a Nc12 instance)

mario:
@AndyScherzinger sending via mail.

mario:
👍 

@AndyScherzinger should be fine, just tested. But let me know.

AndyScherzinger:
@mario just tested and Activities works fine while Notifications is broken, even with notifications visible on the server the client says No Notifications!

mario:
Merge this and I will fix the notifications.

AndyScherzinger:
@mario Erm, just tested... Activities works but Notifications is broken, says "No Notifications", but there are notifications...

AndyScherzinger:
👍 alright

mario:
But how is this related to notifications?

I dont think it touched notifications code?

AndyScherzinger:
I have no idea, maybe it is due to the rich objects?

tobiasKaminsky:
What is the status of this? 
Shall we open a new issue with "notifications broken on <nc12"?

mario:
No. This works perfectly.

PR #1021, opened 2017-05-18
Drawer branding

tobiasKaminsky:
if you right now try to change the drawer icon for e.g. "backup contacts" it will also change all user icons across the app.

-if an icon is only used in drawer, I have renamed it to be more consistent
-if an icon is also used somewhere else, I have copied the icon so that you can have in independent change of drawer only

(this is needed for custom branding)


AndyScherzinger:
👍 fine with me for a start while in the end for video and photos we should then also create the icons ins the correct size because right now the large icons for the empty list/result display are used which are larger than the Android standard 24dp icon sizes. But since it is done like that at the moment anyways we can postpone that part for later.

tobiasKaminsky:
Tell me more and I will do it ;-)

AndyScherzinger:
The photo and video images are used in other places too in the size which is actually present in the repo. For the drawer or any kind of menu they should be in 24dp sizing, you can basically check the size of the "other" drawer icons and create them in the same size per resolution folder :)

Like I said, I can do that at a later point in time, since it basically an optimization since no resizing during runtime needs to be done and the then have the native size so they won't be a bit blurry.

AndyScherzinger:
Icon sizes for action bar / menu / drawer etc. (your nav_* files) would be 24dp according to material design guidelines, meaning:

* mdpi: 24x24px
* hdpi: 36x36px
* xhdpi: 48x48px
* xxhdpi: 72x72px
* xxxhdpi: 96x96px

AndyScherzinger:
👍 looks good, icons in correct sizings now too :)

mario:
👍

PR #1024, opened 2017-05-18
Sort contact list

aleister09:
This PR sort the contact list
If the contact doesn't have Name or LastName is send it to the end of the list


tobiasKaminsky:
:+1: 
@aleister09 Thank you for your PR.
I just did some minor code formatting. 
As we use the default formatting style from android studio you just can use Code -> Reformat Code.

AndyScherzinger:
👍 looks fine to me too 🚀

PR #1025, opened 2017-05-19
[WIP] Use colors from setup.xml for folder icons

tobiasKaminsky:
Just a current WIP with using different colors from setup.xml
Useful for automatic branding.

Sometimes later this can also be enhanced to (hopefully) change colors dynamically if set on theming app on server, for all users.

Icons that are now changed:
- folder (shared, linked, regular)
- empty list icon
- auto upload cloud icon (only on checked state)
- checkbox (only on checked state)
- primary action buttons


src/main/java/com/owncloud/android/utils/MimeTypeUtil.java
```diff
@@ -94,14 +98,25 @@ public static int getFileTypeIconId(String mimetype, String filename) {
...
+            drawableId = R.drawable.shared_with_me_folder;
+        } else {
+            drawableId = R.drawable.ic_menu_archive;
         }
 
-        return R.drawable.ic_menu_archive;
+        Drawable drawable = ResourcesCompat.getDrawable(MainApp.getAppContext().getResources(), drawableId, null);
+        drawable = DrawableCompat.wrap(drawable);
+        DrawableCompat.setTint(drawable, MainApp.getAppContext().getResources().getColor(R.color.primary));
+        return drawable;
```
AndyScherzinger: This tinting code already exists in ExtendedListFragment.java so I vote for putting the tining logic into the DisplayUtils class since that is where we also do the tinting for other UI elements.

tobiasKaminsky: Right :+1:

src/main/java/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -677,4 +681,11 @@ public static String getData(InputStream inputStream){
         return text.toString();
     }
 
+    public static Drawable tintDrawable(@DrawableRes int id, @ColorRes int color) {
```
AndyScherzinger: Since we are always using the primary color, we could remove the color parameter for the time being and make this method private while offering ```public static Drawable tintDrawable(@DrawableRes int id)```

AndyScherzinger: Just a thought, nothing we really really need to implement

tobiasKaminsky: In the current state, yes.
But on master the "no files here" has a lighter folder icon than the list view folder icon.
If we want to support this also, then we need to stay this way.

Let's wait what else needs to be customizable and then we can change this if needed.

AndyScherzinger: sounds good 👍 - btw the light blue is imho "primary with 50% opacity" while I hardcoded the color in the PNGs without opacity...

tobiasKaminsky:
@AndyScherzinger what is missing?

src/main/res/layout/folder_sync_item_header.xml
```diff
@@ -52,6 +52,7 @@
             android:background="@color/transparent"
             android:clickable="true"
             android:padding="@dimen/standard_half_padding"
+            android:tint="@color/primary"
```
AndyScherzinger: The icon is afaik already tinted via code plus this might lead to the grey/deactivated icon to be colored to, which it shouldn't!

AndyScherzinger: pretty sure, since I have implemented it that way ;) - same for the user info view btw!

AndyScherzinger:
tl;dr @tobiasKaminsky you should be covered by your list. Besides that for a true full theming...

* folder (shared, linked, regular)
* empty list icon, empty uploads icon
* auto upload cloud icon --> should be done already in my code! Don't kill the grey/inactive version!
* check box (you didn't do that one, you just did it for the multi selection list items! It still won't work for any real check boxes except when the primary color is the one you want, due to the coloring defined by the theme)
* primary action button and border less button (always primary colored by style definition)
*EditText (same here)

I am stopping here since it'll work if primary is the right choice, else you pretty much have no chance for tinting UI elements on the fly since Android is build on style definitions!!!

src/main/res/layout/folder_sync_item_header.xml
tobiasKaminsky: Reverted and now using DrawableUtils.tintDrawable for this.

tobiasKaminsky: To what are you referring with "user info"?

AndyScherzinger: The user info screen (with the account data) also has icons which as far as I remember already use tinting (which might now need to be replaced by the centralized tinting method).

tobiasKaminsky:
Please track in #1030

PR #1030, opened 2017-05-22
[Branding] Use color for folders, ...

tobiasKaminsky:
Icons that are now changed:
- folder (shared, linked, regular)
- empty list icon
- auto upload cloud icon (only on checked state)
- checkbox (only on checked state)
- primary action buttons

PR #1031, opened 2017-05-22
[Theming] Dynamically use server color

tobiasKaminsky:
This overwrites the default server color and uses the color provided via theming app.

TODO:
- Common
  - [x] change font color based on background color: https://github.com/nextcloud/ios/issues/166#issuecomment-303175279
  - [x] compute lighter / darker color
  - [x] change color of headerbar
  - [x] test with disabled theming app
  - [x] test with older nc version (10,11)
  - [x] test with older android version
  - [x] test changing on the fly
  - [x] refactor: move all theming functions into ThemingUtil
  - [x] loading spinner
  - [x] all empty messages grey, except of folder (blue) and star icon (yellow)
- UserInfoActivity
  - [x] use background image for user info toolbar background
  - [x] handle background color for drawer (if image in theming app is disabled)
  - [x] change color of user info icons -> use the correct color of the account
  -  [ ] wrong aspect ratio
- ManageAccountActivity
  - [x] change color of font and active icon in accounts list
- FileDisplayActivity
  - [x] change color of FAB
  - [x] search hint and search input text is not black/white
  - [x] change color of folder icons
  - [x] show server name instead of Nextcloud when browsing root folder
  - [x] update slogan right after account changing
- Drawer
  - [x] change color of active entry
  - [x] change color quota bar
  - [x] switch account name and drop down to black/white in drawer header
  - [x] items are sometimes still colored: https://user-images.githubusercontent.com/1315170/27133252-d708b5ce-5111-11e7-988f-1b34d2d6e158.png
- [x] use background image for drawer
  - [x] handle background color for drawer (if image in theming app is disabled)
  - [x] preserve aspect ratio
  - [x] show image right after setup / account switching
  - [x] background image url is "/index.php/apps/theming/loginbackground?v=43": verify that new image is downloaded after changing image
 - image is scaled server-side to max 1920p @ 75%
- [x] PassCode
- [x] FingerPrintActivity

OPEN / UNRESOLVED:
- [x] change color of main FAB button -> currently not supported in library
- [x] toolbar: changed only after restart or entering and leaving multiselect
- [ ] UploadFilesActivity: "select all" in overflow menu: not colored and wrong icon(?!)
- [x] ShareDialog: theming header of DatePicker not possible, only with https://github.com/wdullaer/MaterialDateTimePicker
- [x] AutoUpload: uploadBehaviour: bullet points of dialog. Canntot use theme: Caused by: java.lang.IllegalStateException: This Activity already has an action bar supplied by the window decor.
- [x] SwitchPreference not possible


tobiasKaminsky:
In theming app on server we only define one basic color.
But in android we use some more:
- primary
- primary dark
- color accent

How can we get these colors? color accent is a bit lighter, the other one darker.
Shall this be auto-generated? If so, how? As I am not a designer something rgb transformation rule would be nice.

AndyScherzinger:
* use background image for drawer (do we need a scaled version for this?)
** definitely, that is why right now we ship a scaled version, since letting Android OS scale the original image at runtime is a real performance killer (drawer open/close starts to hang...!)
* handle background color for drawer (if image in theming app is disabled)
** this is either primary or primary dark at the moment

> How can we get these colors?

We are more or less free to choose how they get calculated since they do not really follow Material Design guidelines (simply since we do not use a truly different accent color but a slightly lighter one). Maybe @juliushaertl has an idea on color calculation here since he is the server side color calculation hero 😃

juliushaertl:
> We are more or less free to choose how they get calculated since they do not really follow Material Design guidelines (simply since we do not use a truly different accent color but a slightly lighter one). Maybe @juliushaertl has an idea on color calculation here since he is the server side color calculation hero ￼
It should work quite well if you convert the RGB value to the HSL color space and then either increase or decrease the lightness (L) component to get brighter/darker color tones. 

Unfortunately [android sdk doesn't seem to support HSL](https://developer.android.com/reference/android/graphics/ColorSpace.html) out of the box or at least I can't find it at the moment. But conversion is not that hard, see: https://tips4java.wordpress.com/2009/07/05/hsl-color/

AndyScherzinger:
primary_dark should be the fallback for drawer and user info screen background image in case no background image could be retrieved from server

tobiasKaminsky:
@juliushaertl thanks for pointing out with HSL.
There is a native way with ColorUtils; currently I am testing with +/- 0.2 for L.

AndyScherzinger:
@tobiasKaminsky just tested after rebase and also fitting the commit comment of https://github.com/nextcloud/android/pull/1031/commits/22556e0b84d08ffa31b08d82bf6b391b9ffc0e01 (use server slogan if present when browsing root folder), the root folder is showing the servers slogan instead of the name.

AndyScherzinger:
re-targeted to ```master``` since ```colorFolderNew``` has been merged to master already, see #1030

AndyScherzinger:
@tobiasKaminsky as for the solid color when background image is not available. This also needs to be done here: https://github.com/nextcloud/android/blob/userTheming/src/main/java/com/owncloud/android/ui/activity/UserInfoActivity.java#L275

AndyScherzinger:
@tobiasKaminsky just pushed you the https://github.com/nextcloud/ios/issues/166#issuecomment-303175279 based java method 
https://github.com/nextcloud/android/pull/1031/commits/608bf307640f73b4f0f02d7df53c5289effbb50a
to calculate black/white text color: 
https://github.com/nextcloud/android/pull/1031/commits/d8d8d0692ad81c416551ea25fccd112e3ff96aac 
https://github.com/nextcloud/android/pull/1031/commits/d8d8d0692ad81c416551ea25fccd112e3ff96aac#diff-fdc073b6a9b2f53fdd2f582f2c62ab95R738
If it is greater than 0.5 you can use black as a foreground color otherwise you can use white.

AndyScherzinger:
The quota colors shouldn't be themed, so I'd say, lock-in the blue color.
Else move over to a locked-in green color, e.g.: #66bb6a or #81c784 or #a5d6a7 (maybe too light)

AndyScherzinger:
@tobiasKaminsky While thinking about it a little longer, I'd for for a grey actually, minimal disturbance for the user!

AndyScherzinger:
@tobiasKaminsky something like this? https://stackoverflow.com/questions/31394265/navigation-drawer-item-icon-not-showing-original-colour

src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -297,11 +297,7 @@ private void setupQuotaElement() {
      * @param navigationView the drawers navigation view
      */
     protected void setupDrawerMenu(NavigationView navigationView) {
-        // on pre lollipop the light theme adds a black tint to icons with white coloring
-        // ruining the generic avatars, so tinting for icons is deactivated pre lollipop
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
-            navigationView.setItemIconTintList(null);
-        }
+        navigationView.setItemIconTintList(null);
```
AndyScherzinger: Now you removed removed the tinting!

AndyScherzinger: see https://stackoverflow.com/questions/15543186/how-do-i-create-colorstatelist-programmatically for programmatically create a ColorStateList, please :/

tobiasKaminsky: But I added it there: 
https://github.com/nextcloud/android/blob/bee1e9ddc51225ca4414b72a11cba6e3131ded07/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java#L918

AndyScherzinger:
@tobiasKaminsky NPE found when branch is run in emulator without an internet connection:
```
06-13 11:29:53.460 4700-4700/com.nextcloud.client E/AndroidRuntime: FATAL EXCEPTION: main
Process: com.nextcloud.client, PID: 4700
java.lang.RuntimeException: Unable to start activity ComponentInfo{com.nextcloud.client/com.owncloud.android.ui.activity.FileDisplayActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.isEmpty()' on a null object reference
    at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2665)
    at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2726)
    at android.app.ActivityThread.-wrap12(ActivityThread.java)
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1477)
    at android.app.ActivityThread.main(ActivityThread.java:6119)
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
 Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.isEmpty()' on a null object reference
    at com.owncloud.android.utils.DisplayUtils.primaryColor(DisplayUtils.java:782)
    at com.owncloud.android.utils.DisplayUtils.fontColor(DisplayUtils.java:752)
    at com.owncloud.android.ui.activity.ToolbarActivity.setupToolbar(ToolbarActivity.java:64)
    at com.owncloud.android.ui.activity.ToolbarActivity.setupToolbar(ToolbarActivity.java:73)
    at com.owncloud.android.ui.activity.FileDisplayActivity.onCreate(FileDisplayActivity.java:202)
    at android.app.Activity.performCreate(Activity.java:6679)
    at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1118)
    at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2618)
    at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2726) 
    at android.app.ActivityThread.-wrap12(ActivityThread.java) 
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1477) 
    at android.app.ActivityThread.main(ActivityThread.java:6119) 
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886) 
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776) 
```

tobiasKaminsky:
So this is ready for a first test.
Only the aspect ratio in UserInfoActivity is wrong.

AndyScherzinger:
....let's start....

Wipe image, start emulator, create account... (only initial start though!)
device-2017-06-14-145313

sort dialog...
device-2017-06-14-144952

upload file (within app)...
device-2017-06-14-144655

sharing dialog(s)...
device-2017-06-14-144840
device-2017-06-14-144907

simply click your way though the drawer menu...
device-2017-06-14-143809

empty upload screen...
device-2017-06-14-143852

filled upload screen...
device-2017-06-14-150320

auto upload preference...
device-2017-06-14-144245

create folder message...
device-2017-06-14-144307

general preferences + switches + dialogs...
device-2017-06-14-144455

selecting a file to upload...
device-2017-06-14-150115

no screenshot but progressbar/textcolor and icon background of notifications also stay Nc blue...

tobiasKaminsky:
> empty upload screen...
https://user-images.githubusercontent.com/1315170/27133264-de65372a-5111-11e7-8924-edb3e1b759e8.png

This is done by your PR, or?

AndyScherzinger:
> This is done by your PR, or? It should always be gray?

I have no idea actually... It has always been blue but according to @jancborchardt's comment I'd say it should also be gray. My PR hasn't taken care of this one so it'll still be colored. Easy to change to gray if we decide to go for that...

tobiasKaminsky:
Android sometimes does not allow to change the any arbitrary color.
In such cases, like the toggle switches in Settings, I would like to change them to a dark gray (thx @AndyScherzinger) otherwise it looks really strange.

Are you fine with this @jancborchardt ?

filerpba4i

AndyScherzinger:
We could create a "theming theme" that defines a gray color schema which can be dynamically set and would take care of the UI elements that can't be themed dynamically.

jancborchardt:
To confirm, empty upload screen icon should be grey. Folders are the only blue icons, and stars yellow. :)

@tobiasKaminsky how would that dark gray then look, can you show a screenshot?
Also, when the color is too bright we have to switch the subheading font color in settings to grey too. And the notifications bar is not colored here - is that possible?

tobiasKaminsky:
> @tobiasKaminsky how would that dark gray then look, can you show a screenshot?

More or less in this way:
fileioxg37

> Also, when the color is too bright we have to switch the subheading font color in settings to grey too

Based on the chosen color in theming app server-side, I create a darker (-0.2) and a lighter (+0.1) color, based on HSL (range 0...1). So we have three colors and they are used all across the app. Of course we can have a fallback to gray for the lighter color, but this means it will be changed wherever the lighter color is used.

> And the notifications bar is not colored here - is that possible?

Already done :-)

jancborchardt:
The toggle should be fine. Maybe even a bit darker to discern it more from the disabled one.

Regarding the subheading the issue of course is that here the color is used for text on white background. So then for the header we could say if it's so light that the main header bar text would be black (as in this example), the subheading color should use black as color too cause the main color would be too bright to be readable on the white background.

AndyScherzinger:
@tobiasKaminsky @jancborchardt here are some samples for grey colors:

disabled (fixed color theme!)
device-2017-06-19-155931

enabled (material grey600 - 757575)
device-2017-06-19-160038

enabled (material grey700 - 616161)
device-2017-06-19-160407

enabled (material grey800 - 424242)
device-2017-06-19-160556

@tobiasKaminsky beware regarding coloring of UI elements like buttons, checkboxes, etc. the color should actually be color_accent, not primary or pirmary_dark ;)

tobiasKaminsky:
> Regarding the subheading

or we always take the darker color for text on white background? For me this looks more consistent than switching to gray. Gray should in my opinion only be used if we cannot theme the elements due to limitation by android

AndyScherzinger:
I did mean using the Grey only as a fall back for cases where we cannot theme the element ;)

tobiasKaminsky:
--> 

@jancborchardt what about always using the darker color for text on white background?
I do not want to make it too complicated ;-)

tobiasKaminsky:
> Beware regarding coloring of UI elements like buttons, checkboxes, etc. the color should actually be color_accent, not primary or pirmary_dark ;)

@jancborchardt Is this correct? Of course I trust Andy, but this will probably require some changes and I want to be 100% that it is now the right way.

AndyScherzinger:
That I don't know. Just talking Android and Material. Just set the accent color to red and open the settings screen... 
Android by default uses the accent color. So not a matter of trust ;)

src/main/res/values/styles.xml
```diff
@@ -35,6 +35,12 @@
 		<item name="searchViewStyle">@style/ownCloud.SearchView</item>
 	</style>
 
+	<style name="FallbackThemingTheme" parent="Theme.AppCompat.Light.DarkActionBar">
```
AndyScherzinger: my guess is that the parent should be ```Theme.ownCloud```, but needs to be tested. Just saying since Theme.ownCloud defines some more infos that might be needed in case more complex activities might need the fallback style in the future.

src/main/java/com/owncloud/android/ui/activity/Preferences.java
```diff
@@ -135,6 +135,11 @@
     @SuppressWarnings("deprecation")
     @Override
     public void onCreate(Bundle savedInstanceState) {
+
+        if (DisplayUtils.themingEnabled()) {
```
AndyScherzinger: beware that ```onCreate``` in monitored by the OS. So if this called takes too much time the OS will display the wait/kill popup...

jancborchardt:
The color should be the same everywhere, as on the web interface. :) cc @juliushaertl @nextcloud/theming 

I’m confused myself a bit now. :D

juliushaertl:
> The color should be the same everywhere, as on the web interface. 

@jancborchardt I guess the problem is that android might use the accent color for text on white background which may not be the optimal one in case of contrast, see: https://github.com/nextcloud/android/pull/1031#issuecomment-309664160

src/main/java/com/owncloud/android/ui/activity/Preferences.java
tobiasKaminsky: This is "just" using the capability, so it is a quick database fetch.
The slower fetching of cap is done during account switching.

tobiasKaminsky:
> I guess the problem is that android might use the accent color for text 

I can change it to whatever you want ;-)
Also the computation of lighter/darker color based on the defined "base" color can be adjusted.

AndyScherzinger:
@juliushaertl @jancborchardt it is like @tobiasKaminsky said that he can apply any color we want. My comment was just the "material design" part meaning Android uses 3 "colors" called "primary" (Nc blue), "primary dark" (darker shade, system bar at the very very top) and "accent" (lighter shade, for checkboxes etc.) which leads to the Android typical material design so I would (strongly) advice that we also use an "accent" color to make it look material while this will still preserve the Nc look. :)

tobiasKaminsky:
ping @juliushaertl @jancborchardt

tobiasKaminsky:
ping @juliushaertl @jancborchardt

jancborchardt:
Sorry, was on vacation. :) The thing is that the subheadings need to stay readable, which is not the case wih a light theming color. And the darker primary color should not be overused as it's mainly for the system notification bar on top.

Hence for when the theming color is too light (when the text on it would switch from white to be black) I would also make the subheading text color black.

AndyScherzinger:
So ```primary color``` it is then 👍

jancborchardt:
Primary color, but if that is very bright, then use black. ;) Cause if you use a bright yellow as primary color (or the turquoise you see above), the subheadings will not be readable. We need a tiny bit of logic in there. :)

tobiasKaminsky:
So to clarify as apparently I used accentColor too often:
- accent color for UI elements, checkbox, toggleButtons, horizontal rows, ...
- primaryColor for subheading, folder color, ...
  - primaryColor as background: always use it  
  - primaryColor on white
     - if this color is too light (switching text to black): also use solid black
     - otherwise use primary color
- dark color only for status bar

jancborchardt:
As far as I know, accentColor should only be used for primary action buttons, if anything.

Everything else like header bar, checkboxes, toggles, folder icons etc should be primaryColor.

Right @juliushaertl @nextcloud/theming, could you confirm? :)

tobiasKaminsky:
https://material.io/guidelines/style/color.html#color-color-system

> **Primary Color**
A primary color is the color displayed most frequently across your app’s screens and components. It can also be used to accent elements, if you don’t have a secondary color.

> Secondary Color
A secondary color is used to accent select parts of your UI. It can be complementary or analogous to your primary color, but it should not simply be a light or dark variation of your primary color. 
...
Using a secondary color is optional. It’s not necessary if you use variations of your primary color to accent elements.
Secondary colors are best used for:
- Buttons, floating action buttons, and button text
- Text fields, cursors, and text selection
- Progress bars
- Selection controls, buttons, and sliders
- Links
- Headlines

AndyScherzinger:
see...

jancborchardt:
Oookay, thanks! So in our case, to keep in line with the web interface: the Android app secondary color should be the same as the primary color. :)

AndyScherzinger:
I'd say nope, since then the system bar will have the same color as the system bar...

jancborchardt:
Hm ok, then let's go for a just ever so slightly darker color than the primary.

(Sorry for the long discussion! :)

AndyScherzinger:
No worries @jancborchardt it is usually primary_dark = primary+"200" - giving it a slightly darker shade while I can't tell from the top of my head how much more grey needs to be added, but @tobiasKaminsky's implementation already calculates the color adequately.

For accent color, that is one where we can drop the theming support or do it like the standard Nc theme of the app does going for accent = primary - "200" - giving it a slightly lighter shade.

<3

tobiasKaminsky:
So...here it goes:
 --> 
 --> 
 --> 
 --> 
 --> 
 --> 
 --> 
 --> 
 --> 
 --> 

I'll hope you like it @jancborchardt

AndyScherzinger:
@tobiasKaminsky the divider line in the share dialog in your screenshots is still Nc blue

tobiasKaminsky:
> @tobiasKaminsky the divider line in the share dialog in your screenshots is still Nc blue

Done. Thanks for the the hint

AndyScherzinger:
@tobiasKaminsky some findings

* the backup contacts screen keeps crashing the app (not related to theming I think)
* progress wheel on activities screen isn't themed
* "link" (spannable strings) in the activity item texts aren't themed (accent or primary)
* multi-selection mode doesn't theme the progressbar anymore (!) - grey is missing
* checkbox in details screen for set available offline is still Nc blue

jancborchardt:
- color of the highlighted entry in the sidebar is too light for the turquoise theme, should be black instead like the header text

Otherwise seems good except Andy's comments. :)

Check it out @nextcloud/theming

AndyScherzinger:
> color of the highlighted entry in the sidebar is too light for the turquoise theme, should be black instead like the header text

@jancborchardt or we then use primary_dark (-> the systembar color)? :) That way we can keep the color and don't have to go 50 shades of grey here 😝  What do you think?

jancborchardt:
Let's try it out. :)

tobiasKaminsky:
1) 2) 
3) 4) 

I vote for 1) (top left)

juliushaertl:
> I vote for 1) (top left)

Yes, that is the best in case of readability since icon/text have the same color, it also is consistent to the rest of the sidebar.

tobiasKaminsky:
Merged in dev. 
Let's see how this works with all the custom colors the users have :-)

jancborchardt:
Yup, version 1 is best. :)

AndyScherzinger:
Just some minor issues (one already named in the issue description).

device-2017-07-10-164403
device-2017-07-10-164204

links should be colored (I thought they where at some point) and should be colored ```primary```

tobiasKaminsky:
@AndyScherzinger regarding colored links in participate view: which language do you use? The strings need to be changed on transifex, and I can only do this for en-us (the default) in github, so en-gb and german will not work currently.

tobiasKaminsky:
The bullet points are now grey as discussed as we cannot dynamically theme them.

AndyScherzinger:
> regarding colored links in participate view: which language do you use? The strings need to be changed on transifex, and I can only do this for en-us (the default) in github, so en-gb and german will not work currently.

@tobiasKaminsky than I guess that is the reason for it since I run English (Germany) on my device 😁

src/main/java/com/owncloud/android/ui/ThemeableSwitchPreference.java
```diff
@@ -0,0 +1,89 @@
...
+            View child = viewGroup.getChildAt(i);
+
+            if (child instanceof Switch) {
+                Switch switchView = (Switch) child;
+
+                int color = ThemeUtils.primaryAccentColor();
+                int trackColor = Color.argb(77, Color.red(color), Color.green(color), Color.blue(color));
+
+                // setting the thumb color
+                DrawableCompat.setTintList(switchView.getThumbDrawable(), new ColorStateList(
```
AndyScherzinger: needs an API check since we support 14+ while ```getThumbDrawable()``` is v16+

src/main/java/com/owncloud/android/ui/ThemeableSwitchPreference.java
```diff
@@ -0,0 +1,89 @@
...
+                        new int[][]{new int[]{android.R.attr.state_checked}, new int[]{}},
+                        new int[]{color, Color.WHITE}));
+
+                // setting the track color
+                DrawableCompat.setTintList(switchView.getTrackDrawable(), new ColorStateList(
```
AndyScherzinger: needs an API check since we support 14+ while ```getTrackDrawable()``` is v16+

AndyScherzinger:
Awesome fix @tobiasKaminsky - very minor thing which we can ignore: The settings switches are slightly lighter colored than the "regular ones, see screenshot mix:

device-2017-07-11-122915

tobiasKaminsky:
Feel free to try to set the color to a better value ;-)

AndyScherzinger:
I'll have a look!

AndyScherzinger:
👍 fine with all the changes and by now thoroughly tested on dev releases and also by me - @mario feel free to review/approve/merge :)

juliushaertl:
Great work everyone, works really nice in the dev release :clap:

AndyScherzinger:
@mario @tobiasKaminsky in case we get this merged today I' might be able to update the auto upload branch this evening if there aren't too many conflicts to resolve. :D

mario:

AndyScherzinger:
👍 Updated to latest master and ran a final test 👍

PR #1039, opened 2017-05-24
Auto upload light

tobiasKaminsky:
As wished by a customer this is auto upload light:
- set most settings to default
- only "on wifi" is toggable in settings

This is disabled by default, but can be enabled and adjusted in setup.xml if required.
fileiz7ogz

Configure folders, no options per folder, but will use global default values:
filejpqumj

TODO:
translate after merge in transifex:
- es: Carga automática
- pt-br: Upload automático


tobiasKaminsky:
@aleister09 please have a look into it :-)

AndyScherzinger:
@tobiasKaminsky what will the second setting do? open another settings screen? Will this the auto-hide the per folder setting in the auto upload recycler view items?

tobiasKaminsky:
@AndyScherzinger correct. I have uploaded an image to first post for better explanation.

AndyScherzinger:
👍 code looks fine, also run it in an emulator, so fine by me :)

AndyScherzinger:
@tobiasKaminsky beware that this is going to break (use case wise) as soon as #797 gets implemented since you need to choose local and remote folders (where you could then simply hide all other options) for custom folders and after creation you would (due to this pr) hide the options menu for custom folders too. So you could never ever change it again except on/off state... And also enforcing defaults makes them a hard coded setting which might be fine for photos/videos (as it is now) but doesn't seem to make sense for custom folder, so your other way of fixing it then would be to then allow the preferences menu for custom folders (where the check for the job creation would then also have to check for the folder type (to be custom) and then use the db entities values even if the customer/custom setup is configured...

So nothing that can be solved in different ways, but something you should be aware of...

MorrisJobke:
> @aleister09 please have a look into it :-)

Yep - @aleister09 please check this out :)

aleister09:
It's working fine 👍

AndyScherzinger:
:+1:

AndyScherzinger:
👍 re-approved, @mario please have a look and merge :)

mario:
👍

PR #1073, opened 2017-06-07
Removing uploads from account when logout

imjarp:
This PR will remove the uploads when the user logout from it´s account. Because when you login with another account you will see the previous uploaded file.


AndyScherzinger:
👍 @mario code looks good to me, haven't had the time to test it, but the PR includes a unit test. Kudos for that @imjarp 🎉

AndyScherzinger:
👍 re-approval due to commit fixing codacy issues 👍

mario:
This happens because you see uploads of all accounts, rather than just the one you're logged in with :P cc @tobiasKaminsky @AndyScherzinger 

But yes, looks very good - thanks.

mario:
👍

PR #1096, opened 2017-06-19
1089_sharing_with_password_rule: don't ask for authentification when …

AugustinVoiMa:
FileActivity.java: 
filter when operation is an instance of CreateShareViaLinkOperation before asking for creditentials.

ShareActivity.java:
adding this case to the condition that ask again for a correct password

strings.xml:
adding a string that describe the problem. By default (only value for now) "Password does not meet length requirements"

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FileActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileActivity.java
@@ -40,7 +40,6 @@ import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
 import com.owncloud.android.authentication.AccountUtils;
 import com.owncloud.android.authentication.AuthenticatorActivity;
-import com.owncloud.android.datamodel.ArbitraryDataProvider;
 import com.owncloud.android.datamodel.OCFile;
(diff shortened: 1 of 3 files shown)


AndyScherzinger:
@tobiasKaminsky can you take a look? I'd guess this is due to the fact that the PR is not a NC-branch.

tobiasKaminsky:
I suspect that the colon or the length could be a problem.
@AugustinVoiMa first thank you for this PR, I will soon have a look.
Can you just push a dummy change into this branch with a short commit message (and no special chars)?

AugustinVoiMa:
I'm not sure to understand... Do you want me to make an empty commit on this branch? I have never done that before can you confirm me you want `git commit --allow-empty -m "my short message"`
I do it when I'm back at home

tobiasKaminsky:
Looks good. It is just to test the build system.

AugustinVoiMa:
Hello, I have made a new commit as you said but It doesn't seems to resolve the build fail

tobiasKaminsky:
I copied your branch to our repo and there is no error. 
Also it worked with a dummy branch made by me #1111 (nice number...)

Issue #1097, opened 2017-06-20
Keep session cookies


LukasReschke:
### Actual behaviour
- Session cookie is not sent

### Expected behaviour
- Session cookie is sent
 
### Steps to reproduce
1. Connect to a Nextcloud
2. Sniff the traffic of the Nextcloud Android app
3. Cookies are not sent

### Environment data
Android version: 7.0

Device model: EVA-L09

Stock or customized system: No idea

Nextcloud app version: Newest from playstore

Nextcloud server version: 12.0.0

### Logs
```
PROPFIND /remote.php/webdav/XXXX HTTP/1.1
Depth: 0
Authorization: Basic XXXXXXXXX
User-Agent: Mozilla/5.0 (Android) ownCloud-android/1.4.3
Host: cloud.xxxxx.com
Content-Length: 336
Content-Type: text/xml; charset=UTF-8
Connection: close

<?xml version="1.0" encoding="UTF-8"?><D:propfind xmlns:D="DAV:"><D:prop><D:resourcetype/><D:getlastmodified/><size xmlns="http://owncloud.org/ns"/><D:creationdate/><id xmlns="http://owncloud.org/ns"/><D:displayname/><D:getcontentlength/><D:getetag/><permissions xmlns="http://owncloud.org/ns"/><D:getcontenttype/></D:prop></D:propfind>
```

### Upstream changes
https://github.com/owncloud/android/pull/1851 should fix this.

LukasReschke:
cc @tobiasKaminsky @mario The reason why brute-force protection is throttling all requests :)

LukasReschke:
Also this will massively affect LDAP servers and also the server load since it has to bcrypt every request 🙈

tobiasKaminsky:
I am on my way :-)

AndyScherzinger:
merged for 1.5.0 release

PR #1099, opened 2017-06-21
Bruteforce support: re-use cookie

tobiasKaminsky:
Only first start will take long, after that existing cookie will be re-used or refreshed if meanwhile a new cookie is used.

Library: https://github.com/nextcloud/android-library/pull/65
Ref: #1097


mario:
👍

mario:

AndyScherzinger:
👍 looks good!

PR #1101, opened 2017-06-21
codacy fixes: add break to switch statements

AndyScherzinger:
fixing missing breaks in switch statements.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/adapter/FileListListAdapter.java
+++ b/src/main/java/com/owncloud/android/ui/adapter/FileListListAdapter.java
@@ -249,7 +249,7 @@ public class FileListListAdapter extends BaseAdapter {
                     fileSizeV.setVisibility(View.VISIBLE);
                     fileSizeV.setText(DisplayUtils.bytesToHumanReadable(file.getFileLength()));
 
-
+                    break;
                 case GRID_ITEM:
                     // filename
                     fileName = (TextView) view.findViewById(R.id.Filename);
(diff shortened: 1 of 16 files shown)


tobiasKaminsky:
@mario please obey our own rules and wait for two approvals.

src/main/java/com/owncloud/android/ui/adapter/FileListListAdapter.java
```diff
@@ -249,7 +249,7 @@ public View getView(int position, View convertView, ViewGroup parent) {
                     fileSizeV.setVisibility(View.VISIBLE);
                     fileSizeV.setText(DisplayUtils.bytesToHumanReadable(file.getFileLength()));
 
-
+                    break;
```
tobiasKaminsky: This and the next "break" killed our list view completely...

tobiasKaminsky:
Two additional breaks in getView of FileListListAdapter killed our list view:


Please test it properly before merging!

src/main/java/com/owncloud/android/ui/adapter/FileListListAdapter.java
AndyScherzinger: I'll check this out again because it should work with breaks else it is just working around another bug...

mario: Nah, the thing is that grid item has all the things list item has + more :) That's why no break. Silly. but oh well :D

Issue #831, opened 2017-04-13
java.lang.RuntimeException: android.os.TransactionTooLargeException: data parcel size 1245708 bytes


tobiasKaminsky:
Happens when sending a very large (> 1Mb) log while.

Workaround: if file is > 1 mb truncate it so that it fits. 
AFAIK the newest entries are on top, therefore truncate from the end.

mario:
Perhaps we can serve the file via content provider and that way it'll work no matter the size?

tobiasKaminsky:
I am unsure if the file is readable as it is in /mnt/sdcard/nextcloud/log/currentLog.txt

tobiasKaminsky:
I do not get it...I tried just a plain 
Intent intent = new Intent(Intent.ACTION_SEND_MULTIPLE);
intent.setType(MAIL_ATTACHMENT_TYPE);
startActivity(intent);

and it still gives me 
```java.lang.RuntimeException: android.os.TransactionTooLargeException: data parcel size 2294796 bytes```

PR #1102, opened 2017-06-22
Transaction too large

tobiasKaminsky:
Ref: #831 
Saving log only if screen orientation changes and not if closing fragment


AndyScherzinger:
👍 looks good to me

Issue #627, opened 2017-02-06
Delete ALL relevant local files once account is removed


mario:
Currently when you remove an account all files stay on your phone.

This is, IMHO, a very big security problem and should be remedied for 1.5.0.

Todo:
Show a warning with the options OK and Cancel, but don't provide the option "Remove account but keep data locally".

tobiasKaminsky:
I think there was a discussion about it, either here or back then on ownCloud repo...
But I do not remember the conclusion of it...

When removing the account the user should get a warning "all files will be deleted from your phone and stay on server" to make sure the user knows what it is doing.
I think it should be a question and the user can choose to keep the files (as it is his/her phone and his/her data and he is responsible for it).

@AndyScherzinger @jancborchardt

mario:
Less options, more defaults - since the files stay on the server no harm done. 

But @jancborchardt is our UX guy, so let's just wait for him...

tobiasKaminsky:
(label 1.-4. are only for PRs, Issue should only have these labels "none", "approved", "pr exists").

AndyScherzinger:
I'd also say there should then be a warning that the files will be removed from the device but like Mario argues for the default also say that if the user chooses to remove the account the files will be deleted but never kept on the device. 

Else just don't remove the account. ;)

mario:
@AndyScherzinger some bugs in your answer maybe? :D

AndyScherzinger:
I edited the comment, tl;dr:

Show a warning with the options OK and Cancel, but don't provide the option "Remove account but keep data locally).

mario:
I can live with that, even though I'm always a fan of less warnings, especially with things that are easily revertable (like removing an account from an app is). :)

AndyScherzinger:
It is not about the part of re-adding the account but of possibly re-downloading all the files ;)

OlivierS1:
Should this apply to server password revocation as well. If security is a concern it would make sense content is made unavailable. Not sure how much can really be done unless device gets proper intrrnet access. Or maybe should this be left to antitheft apps instead.

jancborchardt:
Yep, just delete but with a warning is nice cause it's a very destructive action (hassle to add the account again) and very seldomly done. As @AndyScherzinger proposed.

AndyScherzinger:
merged to master for 1.5.0

PR #1109, opened 2017-06-23
Delete all local files on account removal

tobiasKaminsky:
Ref: #627 

To discuss:
- how to prevent that an user clicks on account removal and then immediately kills the app to prevent file deletion

### DIFF
--- a/src/main/java/com/owncloud/android/authentication/AccountUtils.java
+++ b/src/main/java/com/owncloud/android/authentication/AccountUtils.java
@@ -28,12 +28,14 @@ import android.net.Uri;
 import android.preference.PreferenceManager;
 
 import com.owncloud.android.MainApp;
+import com.owncloud.android.datamodel.ArbitraryDataProvider;
 import com.owncloud.android.datamodel.FileDataStorageManager;
 import com.owncloud.android.lib.common.accounts.AccountTypeUtils;
 import com.owncloud.android.lib.common.accounts.AccountUtils.Constants;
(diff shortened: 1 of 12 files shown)


AndyScherzinger:
you can't I#d say, the user could also copy all the files manually and then delete the account.
Else you would nbeed to do cyclic checks for account folders for which there is no account in the database...

tobiasKaminsky:
So you are fine with this?

mario:
Well you could run the remove process in a job ...

tobiasKaminsky:
Sure, but I wanted to have this done on the moment the user clicks on it and not scheduled for sometimes later.

mario:
Solution:

- strike out the account name (aka, you can't click it)
- switch to other acc if it exists, if not show login screen
- run a job (you can schedule it to run after 3 seconds, and in the future you'll be able to schedule it right away :P) with all the delete magic
- prevent adding the same account while there is pending delete for that account
- done :D

AndyScherzinger:
@tobiasKaminsky yes, that means I am fine with deleting all local files on account removal 👍

mario:
@tobiasKaminsky can you please do the things I've written here so we can merge this?

tobiasKaminsky:
Done :tada: 
Tested with
- initially 2 accounts, removing one -> manage account activity with striked through account, drawer only shows persisting account
- initially 1 account, removing one -> login screen

mario:

AndyScherzinger:
👍 tested successfully after latest rebase

PR #1110, opened 2017-06-23
1089 sharing with password rule

tobiasKaminsky:
Test if drone is also failing on nextcloud repo

Ref: #1096


AugustinVoiMa:
So it seems that the problem persists...
I don't know how I could help you with it.. When looking for where drone fails I see compiling errors for drone app... I don't understand what's wrong with my commits

tobiasKaminsky:
> found errors: 73
found warnings: 854
previous errors: 73
previous warnings: 853
FAIL: warning count increased

In this branch everything is fine with your commits. There is however an increased warning found by "android lint", but this is nothing to worry about.

AndyScherzinger:
@tobiasKaminsky fixed enough warnings to make it turn green :)

tobiasKaminsky:
While this PR solves the redirection to the login page, I think we need a better error message back from the server. Otherwise the user does not know what a valid password is.

AndyScherzinger:
@tobiasKaminsky is this PR depending on that server change or could that be added/enhanced afterwards?

AndyScherzinger:
meaningful error message from server: nextcloud/server#5545

AugustinVoiMa:
If I understand it would be an improvement.
This would be definitely usefull for users to know what is a good password. But the question is do the server reply already contain this information? 
Is there a documentation on server side app?

src/main/java/com/owncloud/android/ui/activity/ShareActivity.java
```diff
@@ -347,13 +347,14 @@ private void onCreateShareViaLinkOperationFinish(CreateShareViaLinkOperation ope
...
-            if (result.getCode() == RemoteOperationResult.ResultCode.SHARE_FORBIDDEN    &&
-                    (password == null || password.length() == 0)                        &&
-                    getCapabilities().getFilesSharingPublicEnabled().isUnknown()) {
+            if (((result.getCode() == RemoteOperationResult.ResultCode.SHARE_FORBIDDEN      &&
+                    (password == null || password.length() == 0)                            &&
+                    getCapabilities().getFilesSharingPublicEnabled().isUnknown())           ||
+                    result.getCode() == RemoteOperationResult.ResultCode.UNAUTHORIZED)   ) {
                     // Was tried without password, but not sure that it's optional.
-
+                Toast.makeText(this,"Please enter a valid password",Toast.LENGTH_SHORT).show();
```
tobiasKaminsky: This needs to be extracted to values/strings.xml

src/main/res/values/strings.xml
```diff
@@ -376,6 +376,7 @@
     <string name="update_link_file_error">An error occurred while trying to update the share</string>
     <string name="share_link_password_title">Enter a password</string>
     <string name="share_link_empty_password">You must enter a password</string>
+    <string name="share_link_invalid_password">Password does not meet length requirements</string>
```
tobiasKaminsky: Where is this used?

AugustinVoiMa: It is not used and I shouldn't have let it in my commit. I first thought to give a specific Toast in a new condition but then I saw that the issue could be avoided by completing a condition that already existed (in ShareActivity). So this value is not used anymore and shouldn't be there.

Do you want me to change these two things? I will have to make another pull request to this branch

tobiasKaminsky: I can do it, just wanted to know if it is necessary or if it work in progress. Removing stuff is easy ;-)

AugustinVoiMa:
Hi 
When we disable password enforcing for share links, we can set an optional password, and when it's length is smaller than security requirements, the toast say us the minimum length allowed by server.
However when we enable password enforcing the behavior changes. 
So are you sure this difference only come from server processing?
Unfortunately I'm away and I cannot have a look on this for some weeks, but I have just noticed this and want to share you.

mario:
Ergh so what is this?

AugustinVoiMa:
It is just that with password enforcing disabled, a toast say the minimum number of characters when we give a too short password. 
That is what we want with password enforcing enabled.

mario:
👍

tobiasKaminsky:
>When we disable password enforcing for share links ... the toast say us the minimum length allowed by server.

This is really strange. I will have a deeper look, so please not merging.

tobiasKaminsky:
I created an issue on server: https://github.com/nextcloud/server/issues/5598
They are sending the correct error message but only after an inital 401 response and this is the reason why we do not use it.
I am waiting for their response.

tobiasKaminsky:
I talked to Björn and he suggested to switch to ocsv2 for sharing:
Can you all test this please? https://github.com/nextcloud/android-library/pull/67

tobiasKaminsky:
To clarifiy, as the real fix was to use the new ocsV2 api, this PR is no longer needed.
@AugustinVoiMa thank you for this PR. It would be great if you still want to help us. I am sure there are many pending enhancements :-)

PR #1115, opened 2017-06-23
Widget

tobiasKaminsky:
@AndyScherzinger do you want to work on this branch?


AndyScherzinger:
I can give it a try, sure

tobiasKaminsky:
Currently it looks this way:


Do we want to keep it? Add other features? Close it?

AndyScherzinger:
No opinion really, I just rebased it, kept it up to date and fixed the logo and icons...

@jancborchardt @mario  what do you think?

jancborchardt:
I guess we can do a bit better, no? :) I’m not going to get into the visual design for now because we should look at the usecases first:

- Normal tap on the widget opens Nextcloud: is basically the same as the app icon
- Upload uploads a file – seems good, but we could have a single 1*1 widget for that
- Create new folder: will it create a folder in the root folder? Or where? Seems a bit arbitrary, and seldom used
- Refresh: Kinda makes it seem we are bad at refreshing

So, I think widgets which make sense for us are:
- 1*1 widget with folder icon for custom folder
- 1*1 widget for uploading a file

Resizable widget with file list and all that is more complicated and might be less useful.

What do you think @nextcloud/android?

AndyScherzinger:
@jancborchardt sounds good to me, as for custom folder, do you mean custom folder as in auto upload custom folder?

jancborchardt:
@AndyScherzinger I mean that you can say »this specific folder widget should open your »Paperwork« folder. And another one should open the Favorites view, and another should open some sub-sub-sub folder of a project called »Project results« or so. When setting up the widget, a filepicker would show where you can choose the folder you want. Is that possible?

AndyScherzinger:
@jancborchardt most definitely possible yes while of course we will need to extend today‘s implementation of course :)

tobiasKaminsky:
> 1 * 1 widget with folder icon for custom folder
1 * 1 widget for uploading a file

Both with a folder chooser on setup? I like the idea 👍 
But then I guess it is easier to start from scratch...

AndyScherzinger:
> But then I guess it is easier to start from scratch...

I agree while I'd say we keep the branch as a reference until somebody picks up the idea and starts an implementation.

jancborchardt:
@tobiasKaminsky the custom folder widget would have a folder chooser on setup. The upload widget could have a folder chooser on every run, defaulting to the last used (initially root).

Spartachetto:
What about a widget showing activities?

It could be of interest expecially if configurable.
For example if I could see all activities by other users, I could be informed about files shared with me and about my files downloaded / modified by other users

jancborchardt:
@Spartachetto let's start small with easier-to-do widgets. See if the widgets are useful, and then we can see about more. ;)

Any configuratibility needs to be tested so it's unlikely. Activities by other users are about to be implemented as filter in the Activity section.

tobiasKaminsky:
@AndyScherzinger do we wanna together try to enhance this?
If not, then we kill it. (it is always recoverable via git)

AndyScherzinger:
@tobiasKaminsky I'd say kill it since I am unsure if we even want to add widgets. I think launcher action would be more useful/used than widgets

AndyScherzinger:
@tobiasKaminsky can we delete the branch (and still restore it). Just asking for housekeeping reasons :)

tobiasKaminsky:
Yes, this is possible

AndyScherzinger:
cool! :+1:

PR #1117, opened 2017-06-23
Read access

tobiasKaminsky:
@mario keeps this branch?

### DIFF
--- a/src/main/java/com/owncloud/android/utils/PermissionUtil.java
+++ b/src/main/java/com/owncloud/android/utils/PermissionUtil.java
@@ -14,6 +14,7 @@ public final class PermissionUtil {
     public static final int PERMISSIONS_READ_CONTACTS_AUTOMATIC = 2;
     public static final int PERMISSIONS_READ_CONTACTS_MANUALLY = 3;
     public static final int PERMISSIONS_WRITE_CONTACTS = 4;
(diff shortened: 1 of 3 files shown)


src/main/java/com/owncloud/android/utils/PermissionUtil.java
```diff
@@ -52,4 +59,33 @@ public static void requestWriteExternalStoreagePermission(Activity activity) {
...
+     */
+    @RequiresApi(api = Build.VERSION_CODES.M)
+    public static void requestIgnoreBatteryOptimizationsPermission(Context context) {
+        Intent intent = new Intent();
+        String packageName = context.getPackageName();
+        PowerManager pm = (PowerManager) context.getSystemService(Context.POWER_SERVICE);
+        if (pm != null && pm.isIgnoringBatteryOptimizations(packageName)) {
+            intent.setAction(Settings.ACTION_IGNORE_BATTERY_OPTIMIZATION_SETTINGS);
+        } else {
+            intent.setAction(Settings.ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS);
```
AndyScherzinger: This gets reported by lint/AS as an Play Store violation so we should drop the "else"-part since we liekly don't want to get banned from the store ;)

tobiasKaminsky:
Seems not to be needed, thus closing.

PR #1118, opened 2017-06-23
333 update server changes on download files

tobiasKaminsky:
@AndyScherzinger I think you pulled this from upstream.
How to handle it?


AndyScherzinger:
I guess we need to investigate in case we want to add this feature and check how we can adapt this.

PR #1120, opened 2017-06-23
Give up if disconnected

tobiasKaminsky:
Might be a starting point for offline functionality

@LukeOwlclaw I know you contributed back in the oC days...Do you want to help us with Nextcloud? We would really appreciate it!


LukeOwncloud:
@tobiasKaminsky I'd love to help and code a little for nextcloud but life goes on and everything changes... My priorities can't be here right now. But I am using nextcloud! And one day I am sure I'll be able to contribute again! It is always nice to see when NC moves forward. And even nicer when my old code still can be used. Thanks a lot!

tobiasKaminsky:
@LukeOwncloud nice to hear that you still use nextcloud.
I will try to use this code and enhance it, when we will implement a proper offline functionality.
I wish you all the best!

tobiasKaminsky:
For having real offline support we would have to abstract every file operation, so that it can either work offline or directly online.

So we need a lot of prework until this can be done, thus closing this one.

PR #1126, opened 2017-06-25
Move via uploader

tobiasKaminsky:
This can be used to move files during sharing from another app.

TODO:
- test if file is writable/delete-able, then show move dialog
- otherwise: show "file cannot be moved, is readonly"


tobiasKaminsky:
As the "copy/move/delete" approach on upload from within our app has on some devices a bit trouble, I hesitate to implement via sharing.
Most of sharing today is via content://, so that the file in question cannot moved at all.
-> closing

PR #1135, opened 2017-06-26
Opens share by clicking icon in file listing

tobiasKaminsky:
Reviving old branch by c&p the changes:

a click on the link/shared icon in the file listing opens now directly the share dialog.
(wished by someone long long time ago)


AndyScherzinger:
that was @jancborchardt if I remember correctly

tobiasKaminsky:
Is this working if we have the three dots on every single line (which is also something jan wants)?

jancborchardt:
Aaaawesome, finally! :) and yeah, would be important that both the share icon and the future 3-dot icon work as being tappable. :)

mario:
👍

mario:

AndyScherzinger:
👍 works fine, justed tested it

PR #1148, opened 2017-06-28
New weblogin flow

tobiasKaminsky:
TODO
- [x] how to add two accounts to same server (tomorrow with @LukasReschke)
- [x] how to prevent adding the same account twice (only possible after solving issue above)
- [x] rotate
- [x] loading spinner, until website is loaded
- [x] test with url set in setup.xml

Server: https://github.com/nextcloud/server/issues/5693


tobiasKaminsky:
So this is ready to test.
- login on nc12 server with new flow
- login with same credentials as above
- login onto <nc12 server

AndyScherzinger:
👍 tested successfully on Nc12, with mario on vaca, let's merge 🚀

Issue #1151, opened 2017-06-28
Activity stream: clicking on image opens wrong one


tobiasKaminsky:
### Steps to reproduce
1. open activity stream
2. click on one image 
3. see a different one

Same for other links (file link)

OlivierS1:
I have noticed this behavior on Nextcloud Dev when clicking on a thumbnail when the actual picture has not been downloaded. The file name of the picture opened would be located on the Nextcloud app user folder.

tobiasKaminsky:
PR merged -> closing

mario:
Sorry, not on my Mac - what is the Pr?:)

tobiasKaminsky:
#1153 
(if you are not on your mac, you could also answer later when you have proper access. No need for such a hurry.)

PR #1153, opened 2017-06-29
Clicking on link in activity list shows correct file/folder

tobiasKaminsky:
Ref #1151 

One drawback is that we have to sync every file during activity fetching as it can happen that they have never been synced, so this will take a little bit longer than before.


AndyScherzinger:
@tobiasKaminsky  does it have to be done during activity fetching or could it be done "on demand" whenever someone actually clicks on an item?

mario:
I definitely vote for "on demand". Everything else is ... bad :)

tobiasKaminsky:
Currently neither FileDisplayActivity and all other PreviewActivity need to have a real, filled ocFile.
I tried retrieving the complete ocFile in onClick function (when you click on the activity info) but this is an UI thread, so I would have to add a new asyncTask with first shows a spinner while retrieving.
--> I'll try it.

(Retrieving the activity is only slow for the first time of every item. If ocFile is already downloaded there is no delay)

mario:
Make sure to take rotation into account ;)

tobiasKaminsky:
Now it is "on demand".
Screen rotation also works.

mario:
👍

mario:
@AndyScherzinger go for it.

AndyScherzinger:
:+1:  done

Issue #812, opened 2017-04-07
Preview non working image 


tobiasKaminsky:
1. create a test.jpg with 0 byte
2. open it in app

-> could we get the system bar back and make the bars blue? 
wished by @AndyScherzinger

tobiasKaminsky:
I tried it and I can only get the bar blue, but cannot get back the system bar :-/

AndyScherzinger:
That would be a start though and we then have to look how to get back toe systembar (I guess this has been done via setting a Window feature)

tobiasKaminsky:
@AndyScherzinger maybe you can have a look here?
Please note that this is now a bit trickier as we have resized images.

AndyScherzinger:
@tobiasKaminsky I can have a look but won't find the time before 2018, February respectively (or might be lucky to find some minutes this year, while unlikely).

tobiasKaminsky:
@AndyScherzinger 2018 has come… :rofl: 

First screenshot is when displaying a corrupt image with resized images, second after downloading the file:
![2018-09-20-081102](https://user-images.githubusercontent.com/5836855/45799133-bf056a00-bcac-11e8-892f-6c084693498f.png) ![2018-09-20-081106](https://user-images.githubusercontent.com/5836855/45799134-bf9e0080-bcac-11e8-9dce-457d4431e6ef.png)

This is the view of a working image:
![2018-09-20-081300](https://user-images.githubusercontent.com/5836855/45799222-055ac900-bcad-11e8-80e5-f55fc12be8d5.png)


--> To be honest, I would keep it this way…

PR #1154, opened 2017-06-29
Error image: color action bar again

tobiasKaminsky:
Ref: #812 

@AndyScherzinger can you help me here with the status bar, please :heart:

### DIFF
--- a/src/main/res/layout/preview_image_activity.xml
+++ b/src/main/res/layout/preview_image_activity.xml
@@ -18,29 +18,22 @@
   -->
 
 <android.support.v4.widget.DrawerLayout xmlns:android="http://schemas.android.com/apk/res/android"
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
I'll see what I can do 👍

AndyScherzinger:
Need to revert this PR I think. It messes up the Toolbar (turns it blue all the time also for images that work where it should be semi transparent...)

AndyScherzinger:
Not sure why though...

tobiasKaminsky:
Is this still valid anymore? I remember that there were changes recently...?

AndyScherzinger:
Action bar doesn't need to be colored anymore due to the Hacktoberfest PR. At some point we should bring back the system bar imho. So yes, can be closed.

Issue #435, opened 2016-12-08
scope of "available offline" in ... is incorrect


Ventusfahrer:
### Actual behaviour
I already described the behavior in #416, but it is a different problem than the one 416's title discusses.

After setting "available offline" for one file in folder, newly downloaded and modified files in the same folder are automatically uploaded to the server - without "available offline" defined. Files which where already in the folder before the other file was marked with "available offline" will not be automatically uploaded after modification.

This behavior continues for all additional downloaded and modified files until the android device is rebootet (or the app was restarted - suggestion, not tested). After Reboot, the behavior is as expected.

### Expected behaviour
regardless the result of #416, files created after "available offline" setting of a different file should be handled in the same way as files also not marked with "ao" before "ao" of this one file was changed. This is a bug in my oppinion.
 
### Steps to reproduce

All file create/modification happened on a Linux box with LibreOffice and on an android using AndrOpenOffice.

1. file 1.ods is created on Linux - ownClould client uploads it to the Nextcloud Server - some seconds later it becomes visible in nextcloud App
2. file 1.ods is clicked in ncApp, opened in AndrOpenOffice, modified and safed back to ncAPP directory. No upload is happening to ncServer
3. file 2.ods is created on Linux - ocClient, uploads to ncServer, some seconds later visible in ncApp on Android
4. file 2.ods is modified on Android - no Upload 
5. file 2.ods on ncApp is marked as "availabe offline" and get's immediately uploaded to the server (this was part of the discussion in #416)
6. file 3.ods was created on ocClient (Linux), syncs to ncServer, and is visible on ncApp (Android)
7. file 3.ods is modified in Android. It is immediately uploaded to the ncServer without being flagged "available offline".
8. you can repeat 6 and 7 for additional files, they behave the same.
9. modify 1.ods and save it - it will not been uploaded to ncServer
10. creating a folder in this directory and repeat 6. and 7. showed, that the problem is not being inherited to the subfolder.

### Environment data
Android version:
7.0
Device model: 
Nexus 9

Stock or customized system:
Stock

Nextcloud app version:
1.4.0

Nextcloud server version:
10.0.1

tobiasKaminsky:
Very very weird.
A folder is used to observe and if content of that folder changes it will be checked if the file is in "observedChildren".
For some reason I do not understand each new file that gets downloaded into an observed folder will be added to "observedChildren".

tobiasKaminsky:
[...]

PR #1155, opened 2017-06-30
Comment out adding all new downloaded files to folder observer

tobiasKaminsky:
Ref: #435 

Please @mario @AndyScherzinger test this *really* properly as I do not fully understand the background of the comment near the changed line.


mario:

mario:
@AndyScherzinger also this for 1.5?

AndyScherzinger:
@tobiasKaminsky what do you think? Reading your comment it seems we should postpone it for risk mitigation...

tobiasKaminsky:
Postponing is fine. 
It is not that much change (in terms of lines) so it should be easy to rebase.

mario:
After thinking about this:

Proper fix would be to check if file is marked as available offline and not already being observed, observe. Otherwise do nothing.

tobiasKaminsky:
FileObserverService does not longer exist, so closing.

Issue #1047, opened 2017-05-29
Nextcloud Dev: 20170528: Some folders do not appear in the list.


OlivierS1:
### Actual behaviour
-Tell us what happens

It would seem that some folders are not listed in the list. I can't seem to find any trace of a particular one.
In the log, I have the following details:
`
2017/05/28 18:54:56
Brand : lge

2017/05/28 18:54:56
Product : g3_bell_ca

2017/05/28 18:54:56
Device : g3

2017/05/28 18:54:56
Version-Codename : REL

2017/05/28 18:54:56
Version-Release : 6.0

2017/05/28 18:54:56
Debug : start logging

2017/05/28 18:54:56
SyncedFolderObserverService : Start service SyncedFolderObserverService

2017/05/28 18:54:56
FileObserverService : onCreate

2017/05/28 18:54:56
FileObserverService : Starting command null

2017/05/28 18:54:56
FileObserverService : Loading all kept-in-sync files from database to start watching them

2017/05/28 18:54:56
ThumbnailsCacheManager : create dir: /storage/emulated/0/Android/data/com.nextcloud.android.beta/cache/thumbnailCache

2017/05/28 18:54:57
FolderObserver : Started watching parent folder /storage/B4BA-91D0/Android/media/com.nextcloud.android.beta/nextcloud/XXXXX@XXXXXXXXXX%2Fowncloud/Recettes/
`
The folder named "Recettes" is the one missing from the list.

Folder

### Expected behaviour
-Tell us what should happen
All folder present under my account should be visible within the mobile app list.
 
### Steps to reproduce
1. 
2. 
3. 

### Environment data
Android version:
6.0 

Device model: 
LG 3 : LG-D852

Stock or customized system:
Stock

Nextcloud app version:
Dev 20170528
Nextcloud server version:
11.0.3
#### Web server error log
```
Insert your webserver log here
```

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```
**NOTE:** Be super sure to remove sensitive data like passwords, note that everybody can look here! You can use the Issue Template application to prefill some of the required information: https://apps.nextcloud.com/apps/issuetemplate

PR: #1165

OlivierS1:
Doing side by side comparison between Nextcloud 1.4.3 and Nextcloud Dev 20170528, I found I am missing about 25 top level folders.
Content from 1.4.3 matches what I have on the web interface.
The missing folders seem to be the one that are shared, but not all of them it seems.

[...]

PR #1165, opened 2017-07-05
Fix: no shared folders after "shared" view

tobiasKaminsky:
Ref: #1047


mario:

mario:

AndyScherzinger:
👍 tested successfully

Issue #1066, opened 2017-06-03
Uploading pictures with "Google Photos" via share doesn't work.


OlivierS1:
### Actual behaviour
-Tell us what happens
I just installed "Google Photos" app on my cellphone to debug issues my friend is having.
This new issue targets both Nextcloud 1.4.3 and Nextcloud Dev 20170602, but the Dev version doesn't work at all.

Issue with Nextcloud 1.4.3:

With a few pictures available in "Google Photos", I select a few of them, then click the "Share" logo.
I am given a choice of application. I choose "Nexctloud".
I select the format to be either "LArge size" or "Actual size".
A progress bar will appear indicating how many files are being downloaded.
If the files are not on the device, they will be downloaded (I believe the folder they are going to be stored at is the "Download" folder). When all the files have been downloaded, Nextloud app will appear asking for the folder to upload to. I use the pin to access Nextcloud.
I select my folder then press Upload. Upon pressing the button, "Google Photos" appears.
Background uploading of the pictures by Nextcloud starts.
The problem is that not all pictures are being uploaded.
At a minimum, it seems consitstent at uploading one pictures per share request.
However, when selecting more than one pictures, I could only get at most 3 pictures uploaded by Nextcloud.
My friend was able to upload a larger number of files, but was always missing a few.

### Expected behaviour
-Tell us what should happen
Nextcloud should be able to upload the entire selection provided by "Google Photos" and not skip a few files. 

### Steps to reproduce
1. Have "Google Photos" app installed.
2. Have a few pictures available in "Google Photos"
3. Select a few pictures (max of 200) 
4. Click the "Share" icon
5. Select "Nextcloud" within the list of apps.
6. Enter PIN number, select relevant Nextcloud account if more than one installed within app.
7. Select where to upload the files.
8. Press "Upload" and wait for transfer to complete.
9. Check files are properly uploaded on Nextcloud server. They will not.

### Environment data
Android version:
6.0

Device model: 
LG 3 D852
Samsung Note 5

Stock or customized system:
Stock on both phones

Nextcloud app version:
1.4.3

Nextcloud server version:
NC 12. (Just upgraded to NC12 a few days ago. Issue was first observed under NC 10 and NC 11)

#### Web server error log
```
Insert your webserver log here
```

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```
[...]

PR #1173, opened 2017-07-11
Catch a rare exception while setting activity list adapter

tobiasKaminsky:
Ref: https://github.com/nextcloud/android/issues/1066#issuecomment-312790704
(although the linked issue is not solved via this)


src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java
```diff
@@ -75,6 +76,7 @@
     public static final int ACTIVITY_TYPE = 101;
     private final ActivityListInterface activityListInterface;
     private final int px;
+    private static final String TAG = "ActivityListAdapter";
```
AndyScherzinger: should be 
```
private static final String TAG = ActivityListAdapter.class.getSimpleName();
```

AndyScherzinger:
👍 fine by me

mario:

Issue #1189, opened 2017-07-14
Local removal of folder is not working


tobiasKaminsky:
### Steps to reproduce
1. Select folder
2. delete folder -> local only

--> Removal failed
https://github.com/nextcloud/android/blob/master/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java#L521

AndyScherzinger:
merged to master for 1.5.0

PR #1190, opened 2017-07-14
fix removing of local only folder

tobiasKaminsky:
Ref: https://github.com/nextcloud/android/issues/1189

Previous change was introduced so that if ocFile == null --> return false, so added this explicitly.


AndyScherzinger:
👍 tested successfully

PR #1195, opened 2017-07-15
Media store auto upload: scrollable  toolbar

AndyScherzinger:
as discussed in #797 I implemented a scrollable toolbar so the custom folder creation action is part of the toolbar layout and scrolls away when you scroll down the recycler list items.

Look with theming on:
device-2017-07-15-162510

@tobiasKaminsky @mario Pleases have a look and let's discuss if this is the proper solution. Imho it solves the recycler view content while preserving the "scrolling away" of the action.


AndyScherzinger:
Assigned it to you Tobias for review and merge.

tobiasKaminsky:
:+1: nicer than my implementation

Issue #1100, opened 2017-06-21
AutoUpload view crashes app


despens:
### Actual behaviour
- Switching to instant view upload crashes the app. 
- No files are ever auto-uploaded

### Expected behaviour
- Switching to instant upload view should show instant upload view. 
- Files should be auto-uploaded as configured
 
### Steps to reproduce
1. Start nextcloud app
2. Select Instant Upload view
3. 💥 

### Environment data
Android version: 6.0.1

Device model: STV100-1 (BlackBerry Priv)

Stock or customized system: stock

Nextcloud app version: 1.4.3

Nextcloud server version: 11.0.3

PR: #1201

tobiasKaminsky:
Is this really "instantupload" and not "AutoUpload" as you are using Android 6?
Can you provide us with a logcat?

despens:
Yes it is indeed "AutoUpload", just the directory on my nextcloud is called "InstantUpload". I don't know what a logcat is, if you let me know I will try to provide.

tobiasKaminsky:
Unfortunately there is no easy way to get a proper error message/logcat: https://wiki.lineageos.org/logcat.html

despens:
Find my logcat here: https://nc.1x-upon.com/s/GA5u0VDJ155hLvT
I hope it is useful.

tobiasKaminsky:
Thanks for the logcat, but I cannot see anything related to nextcloud. Please try it again and open autoupload view / nextcloud 2-3 times and verify that in the logcat is something with "nextcloud" and "crash" or similar.

creideiki:
I'm seeing the same thing on my phone:

- Fairphone 2
- Fairphone Open 17.06.4 (based on Android 6.0.1)
- Nextcloud app 1.4.2 (from F-Droid)
- Nextcloud server 12.0.0

The newest file in my `InstantUpload` folder is from 2017-06-10. Shortly after, there were new releases of both the Nextcloud app (1.4.2 on 2017-06-13) and the OS (Fairphone Open 17.06.4 on 2017-06-15, which was an upgrade from Android 5.1 to 6.0).

Running `adb logcat` while clicking "Auto upload" in the app gives this part that seems relevant:
```
07-16 17:30:12.391  2270  2445 I ActivityManager: Displayed com.nextcloud.client/com.owncloud.android.ui.activity.FolderSyncActivity: +248ms
--------- beginning of crash
[...]

PR #1201, opened 2017-07-17
Auto upload crash if filepath is null

tobiasKaminsky:
Ref: https://github.com/nextcloud/android/issues/1100#issuecomment-315737685

Also changed the + to white, so it fits to the number.

### DIFF
--- a/src/main/res/layout/grid_sync_item.xml
+++ b/src/main/res/layout/grid_sync_item.xml
@@ -58,7 +58,7 @@
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 android:text="+"
-                android:textColor="#bcbcbc"
+                android:textColor="#ffffff"
                 android:textSize="22dp"
                 android:textStyle="bold"/>
 
--- a/src/main/java/com/owncloud/android/datamodel/MediaProvider.java
+++ b/src/main/java/com/owncloud/android/datamodel/MediaProvider.java
@@ -119,11 +119,17 @@ public class MediaProvider {
 
                 if (cursorImages != null) {
                     String filePath;
+                    int failedImages = 0;
                     while (cursorImages.moveToNext()) {
                         filePath = cursorImages.getString(cursorImages.getColumnIndexOrThrow(
                                 MediaStore.MediaColumns.DATA));
-                        mediaFolder.filePaths.add(filePath);
-                        mediaFolder.absolutePath = filePath.substring(0, filePath.lastIndexOf("/"));
+
+                        if (filePath != null) {
+                            mediaFolder.filePaths.add(filePath);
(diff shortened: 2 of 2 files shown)


src/main/res/layout/grid_sync_item.xml
```diff
@@ -58,7 +58,7 @@
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 android:text="+"
-                android:textColor="#bcbcbc"
+                android:textColor="#ffffff"
```
AndyScherzinger: why did you turn it into white? The slight grey was actually on purpose :D

AndyScherzinger:
👍 fine by me, as stated in the comment I explicitly made the + slightly grey-ish to focus on the number (like facebook does it)... But I am fine either way.

src/main/res/layout/grid_sync_item.xml
tobiasKaminsky: umm...ok...You are the ui guru ;-)
Lukas and me were just wondering and grey was a little bit hard to read.
Your decision :+1: or :-1: (like Ceasar)

AndyScherzinger: 👍 let's keep #ffffff then

AndyScherzinger:
do we now merge with a single +1 for the time being?

tobiasKaminsky:
This is a general "problem" if one of us is on vacation.
On short vacations (up to 1 week) we can just wait, I think. Mario will be available next week again, so I guess we can wait.
On longer vacations, at least I, will always have a short look into github to reply/review/postpone.

-> wait if it is not urgent, otherwise merge (should be an exception)

AndyScherzinger:
I have to say I completely disagree. "wait for a week" means "postpone a release by a week" since in my opinion a roadblock for a week means it'll take a week longer than planned. No integration of changes will take place so you will have to spend that effort "a week later" or descope stuff none of which I think should be the standard. 

The basic issue is that the 4-eyes principle doesn't work all the time for a 3 people core team but teams >=4 people...

Solution: let's find another dev who contributes regularly (to be found) and is willing to do reviews (to be found)... 😉

tobiasKaminsky:
> Solution: let's find another dev who contributes regularly (to be found) and is willing to do reviews (to be found)

Yes, this is the best solution.
But in the meantime, at lest in my opinion, it is quality over speed.

AndyScherzinger:
I'm fine with that and would say let's discuss this at the conference if we can't find another, yet better approach :)

mario:
Conflicts, otherwise ok.

AndyScherzinger:
@mario I resolved the conflicts (Lint), so merging... 🎉

PR #1204, opened 2017-07-18
Remove language kannada 

tobiasKaminsky:
Problem with BOM and not maintained


AndyScherzinger:
:+1: can imho be merged @mario

mario:
Do it if you think its fine

Issue #1171, opened 2017-07-10
[Share] enter password: toggle visibility of password


tobiasKaminsky:
- share a file as link
- enter password
- allow changing visibility of password 

Ref: https://medium.com/@moyinoluwa/password-visibility-toggle-android-support-library-revision-24-2-0-98b422087e5a

AndyScherzinger:
For any hints, please see the password edit text of the login activity :)

OlivierS1:
@tobiasKaminsky @AndyScherzinger I find entering the password for sharing a link a bit cumbersome and relatively unsecure. I have tried this feature in NC Dev 20170721 and when password doesn't meet requirement it has to be entered again.
Under NC 12, the server is able to generate a password automatically and the password is sent via email while the link to the share is sent on another email.
Can't we have this feature implemented based on server password requirements, best if password had same level of resiliency as app password.

tobiasKaminsky:
Please open a new issue for this. Mixing different things into one issue/discussion is quite confusing.

AndyScherzinger:
password visibility toggle has been merged for 1.5.0

PR #1205, opened 2017-07-18
[Share] Enter password: toggle visibility of password #1171

vnidens:
Pull request for #1171

### DIFF
--- a/src/main/res/layout-land/account_setup.xml
+++ b/src/main/res/layout-land/account_setup.xml
@@ -19,6 +19,7 @@
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+	xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
 	android:padding="@dimen/standard_padding"
@@ -228,7 +229,8 @@
 					android:id="@+id/input_layout_account_password"
 					android:layout_width="match_parent"
 					android:layout_height="wrap_content"
-					android:visibility="gone">
+					android:visibility="gone"
+					app:passwordToggleDrawable="@drawable/password_visibility_selector">
 
                     <EditText
                         android:id="@+id/account_password"
--- a/src/main/res/layout-land/account_setup.xml
+++ b/src/main/res/layout-land/account_setup.xml
@@ -19,7 +19,7 @@
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
-	xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
 	android:padding="@dimen/standard_padding"
@@ -231,7 +231,7 @@
 					android:layout_height="wrap_content"
 					android:visibility="gone"
 					app:passwordToggleDrawable="@drawable/password_visibility_selector"
-                    app:passwordToggleTint="@color/white">
+					app:passwordToggleTint="@color/white">
 
                     <EditText
                         android:id="@+id/account_password"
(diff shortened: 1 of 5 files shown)


tobiasKaminsky:
Thank you for your contribution! It is always nice to have new contributors. 

It is working fine, but I would assume that the icon will switch from "open eye" to "closed eye / crossed eye"?

vnidens:
I've just supposed that it should look the same as in the login activity (as you've mentioned in the issue comments). At this moment existing login activity has the same visibility indicator as your screenshots.

If you don't mind I would update both views with the "open eye" <-> "crossed eye" indicator.

AndyScherzinger:
@tobiasKaminsky unfortunately no.... Google changed that and also on the login screen you'll see that now Android "only" changes the opacity of the eye icon. So that is a platform matter :/

AndyScherzinger:
@vnidens sorry for the confusion, writing my post took too long...
@vnidens @tobiasKaminsky I am fine either way!

tobiasKaminsky:
https://material.io/guidelines/components/text-fields.html#text-fields-input-types
Search for "password" there it is still with "crossed eye".

But for me either ways are fine.

vnidens:
Updated to use crossed eye indicator for both password inputs

tobiasKaminsky:
Great work, but I think the icons are swapped?
At least, on KDE wifi manager it is the other way around:

you can read the password, click on "hide" to cover it....
screenshot_20170718_153454 screenshot_20170718_153514

src/main/res/layout-land/account_setup.xml
```diff
@@ -228,7 +229,8 @@
 					android:id="@+id/input_layout_account_password"
 					android:layout_width="match_parent"
 					android:layout_height="wrap_content"
-					android:visibility="gone">
+					android:visibility="gone"
+					app:passwordToggleDrawable="@drawable/password_visibility_selector">
```
tobiasKaminsky: remove white space after =

src/main/res/layout-land/account_setup.xml
```diff
@@ -19,6 +19,7 @@
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+	xmlns:app="http://schemas.android.com/apk/res-auto"
```
tobiasKaminsky: not needed whitespace

AndyScherzinger:
Nope, as you refered to https://material.io/guidelines/components/text-fields.html#text-fields-input-types it is right. It doesn't reflect the state but the action that will happen when you click.

Go home KDE, your drunk!

AndyScherzinger:
Not sure what happens since I couldn't test it yet, so asking: What will happen on Android 4.0? Since the ic_show is available as png also while the ic_hide icon is vector only and I am not sure if this will break 4.0 since I think proper vector support came in post 4.0, right?

tobiasKaminsky:
> Nope, as you referred to https://material.io/guidelines/components/text-fields.html#text-fields-input-types it is right.

Right, let's stick to material design :+1:

vnidens:
@AndyScherzinger The gradle plugin converts xml vector drawables to pngs if the minSdkVersion < 21 so it's fine.
I can't check with 4.0.3 (API 14) since the emulator is veeery slow on my machine but here is a screenshot from 4.1 (API 16) which is also < 21.
screenshot_1500386894

tobiasKaminsky:
I also tested it successfully with 4.0.4 but emulator of 4.0 does not work.

AndyScherzinger:
@tobiasKaminsky 4.0.4 is perfectly fine! As we don't support 4.0 but only 4.0.x and there it is basically 4.0.3. So 👍

AndyScherzinger:
@vnidens right, totally forgort about the build taking care of that... Thanks for the hint. I'll test drive it later today

tobiasKaminsky:
> @tobiasKaminsky 4.0.4 is perfectly fine! As we don't support 4.0 but only 4.0.x and there it is basically 4.0.3.

We have minSdk set to 14 in AndroidManifest, which is Android 4.0, 4.0.1, 4.0.2, acoording to
https://developer.android.com/guide/topics/manifest/uses-sdk-element.html

tobiasKaminsky:
> @tobiasKaminsky do you have a pre-12 instance so I can test the login?

I have them only local :-/

AndyScherzinger:
👍 fine code wise, also tested the link password part, codacy and lint are happy too ❤️ 

@tobiasKaminsky rebased to latest master so need another thumbs up :)

tobiasKaminsky:
:+1: 
Thank you very much for this contribution, @vnidens! :tada:

PR #1206, opened 2017-07-18
On receiving untrusted ssl cert, test if it was previously accepted user

tobiasKaminsky:
When trying to connect to an untrusted server we show a dialog with cert infos.
If the user accepts the untrusted cert it will get stored. 
New weblogin flow now tests if the cert has been accepted.

### DIFF
--- a/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -408,7 +408,7 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
                         handler.cancel();
                     }
                 } catch (Exception e) {
-                    e.printStackTrace();
+                    Log_OC.e(TAG, "Cert could not be verified");
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -391,6 +397,21 @@ public void onPageFinished(WebView view, String url) {
...
+                Certificate cert = getX509Certificate(error.getCertificate());
+
+                try {
+                    if (cert != null && NetworkUtils.isCertInKnownServersStore(cert, getApplicationContext())) {
+                        handler.proceed();
+                    } else {
+                        handler.cancel();
+                    }
+                } catch (Exception e) {
+                    e.printStackTrace();
```
AndyScherzinger: Please use the logger ;)

tobiasKaminsky: Changed, thank you!

mario:
1) Conflicts
2) Seems like we cancel if we don't find the cert. Can't we ask for user to accept it?

tobiasKaminsky:
The dialog to accept/dismiss the cert is shown when connecting to the server.
There should be no way to get to the part of webdav without seeing this dialog prior.

- enter url
- see dialog about untrusted cert
- accept -> webview -> also accepts cert
- dismiss -> go back to "enter url"

mario:
👍

mario:
Wait for @AndyScherzinger or not? :)

tobiasKaminsky:
I think yes, as we have it not that urgent as we have to wait for google releasing a stable toolchain.

AndyScherzinger:
:+1: code wise fine, can test it since I don't have such a setup.

Issue #983, opened 2017-05-11
On device: show only folders containing downloaded files


tobiasKaminsky:
Suggestion via email: Show only those folders that contain downloaded files.

~~This can be a little tricky if you have something like root -> folder a (no files) -> folder b (with files) then you would have to show folder a. 
Currently we would have to scan this manually as we do not store something like that in the database.~~

tobiasKaminsky:
#1143 

After thinking about it, this is rather easy: While iterating over the children of a folder:
- if the file/folder exists on local file system -> show
- otherwise: hide
This work as the folder "inbetween" (folder a in above example) are also created even if they only contain a subfolder.

jsrck:
I created pull request #1220 for this issue.

AndyScherzinger:
merged to master for 1.5.0

Issue #1143, opened 2017-06-27
Offline-Support


tobiasKaminsky:
Auto-detect offline/online
- show offline message beneath top bar
- show only offline files and only folders that contain offline files/folders
- make it much more responsive

https://github.com/nextcloud/android/issues/680
https://github.com/nextcloud/android/issues/1008

AndyScherzinger:
Hi @tobiasKaminsky what is meant by "make it much more responsive"? :)

tobiasKaminsky:
This is described in #1008: Try opening the image info options (very unresponsive)

I guess the problem is that all connections needs to timeout prior to do something else and therefore it all feels a bit laggy, but not confirmed it by myself.

jancborchardt:
Sorry, only saw this now:

> show only offline files and only folders that contain offline files/folders

Also here the same question as at https://github.com/nextcloud/android/issues/1270#issuecomment-322036378:

> can you elaborate here? I hope it doesn’t mean that when you are offline, the »All files« view will only show files which you already have on the device. Cause that would be confusing as hell. ;)

tobiasKaminsky:
This is what @mario idea was (and he said that he talked with you about it @jancborchardt).
I only wrote it down to not forget this.

jancborchardt:
Let's continue with the discussion in the other issue where we're already at it. :)

We probably discussed the other points, but definitely not the one I remarked. Because I always advocated for offline being not much separate from online. You should be able to delete, edit, rename etc just like you can when online. (With actions being queued til online.)

mario:
Absolutely. But this is very far off :)

Anywayy once I am on land I will find what is relevant and what is stored
online

PR #1220, opened 2017-07-23
On device: show only folders containing downloaded files #983

jsrck:
Folders that aren't available on the device are hidden when selecting "On device".
Pull request for #983.


tobiasKaminsky:
:+1: 

In the longer term we have https://github.com/nextcloud/android/issues/1143 to get rid of this menu entry.
But for now I vote for putting this in, as we will not get #1143 in next version and it is a real nice enhancement.

mario:
👍

PR #1224, opened 2017-07-24
Extract html code and make it non-translatable

tobiasKaminsky:
Some translator had problems with the html code in the translation strings and added a whitespace after "<", resulting in seeing html code instead of the real link.
So this extracts the html code and makes it non translatable.


AndyScherzinger:
👍 looks good to me

mario:
Please resolve the conflicts so I can approve.

AndyScherzinger:
resolved :)

AndyScherzinger:
👍 pinging @mario :)

mario:
👍

PR #1230, opened 2017-07-25
move tests to /src/test

tobiasKaminsky:
While writing an unit test I came across that it is not found by gradle/android as the folder is in /test and not in /src/test as it is expected.


AndyScherzinger:
👍 nice catch!

mario:
👍

Issue #1229, opened 2017-07-25
New android sharing view


tobiasKaminsky:
I came across this cool sharing view in ocReader

![2017-07-25-081312](https://user-images.githubusercontent.com/5836855/28558232-5556c820-7111-11e7-9d67-863aca63883e.png)

The first 4 pixelated items are whats app contacts.
I think this is way better than our current sharing (within android) view.

Please approve, if you like it :+1:

OlivierS1:
@tobiasKaminsky Can you elaborate, I am not sure how to interpret the idea.

tobiasKaminsky:
Currently we have this dialog, but I think the new one gives more options as it shows, eg. directly the most recent contacts of whatsapp

![2017-07-25-085715](https://user-images.githubusercontent.com/5836855/28559560-449952b8-7117-11e7-9e91-a0b289590f06.png)

AndyScherzinger:
I like the idea since it is a more modern way to display the available sharing option on the device.

OlivierS1:
@tobiasKaminsky Do you mean that sharing a file or folder from NC will go though this page first then, If I want to share with a NC member, then I would have to select "Nextcloud dev" icon then select the user?
Or is this going to be an option to chose between a NC user or using the external sharing apps?
One thing I don't like about the dialog icon is that they tend to shuffle around.

tobiasKaminsky:
This is about sending a file to an app on the device.
I know the working is difficult here, "share/send to another app" vs "share with a NC user".

jsrck:
Hi! I created pull request #1244 for this issue.

PR #1244, opened 2017-07-30
 New android sharing view #1229

jsrck:
This pull request enables the chooser dialog requested in #1229. In contrast to the previous version it doesn't exclude the Nextcloud-App from sharing.


mario:
Thanks for the contribution. Screenshots? :)

jsrck:
It looks like on first the screenshot tobiasKaminsky posted in #1229.

On my emulator (without WhatsApp/contacts) it looks like:
send

I tried it also on my smartphone with WhatsApp installed and it looks like on tobiasKaminskys screenshot. Is it possible to test the developed version from Android Studio without uninstalling version 1.4.3 on my phone? I couldn't find out how, yet.

tobiasKaminsky:
>  Is it possible to test the developed version from Android Studio without uninstalling version 1.4.3 on my phone

No as both apps use the same package id :-/

tobiasKaminsky:
The old implementation was filtering the nextcloud app itself. 
Is this also possible with the new view?

jsrck:
> The old implementation was filtering the nextcloud app itself. Is this also possible with the new view?

I tried that as described e.g. here: https://stackoverflow.com/questions/9730243/how-to-filter-specific-apps-for-action-send-intent-and-set-a-different-text-for

But the result
1. Contains duplicates of the ES File Explorer (somehow the different titles get lost...),
2. Doesn't include the contacts. I didn't find an other solution including the contacts...

So with excluding Nextcloud the result is:
excluded

With the version I commited:
with-nextcloud

tobiasKaminsky:
Then I vote for "not filtering nextcloud".
What do you think? @mario @AndyScherzinger

AndyScherzinger:
I also vote for "not filtering nextcloud" :)

mario:
@AndyScherzinger @tobiasKaminsky do you want to ship this with 1.5 if it passes review?

AndyScherzinger:
@mario @tobiasKaminsky I say it depends on how fast Google fixes the build tools. If it takes to long then I'd say we ship it.

mario:
👍 

Hope you guys do not mind, but I've fixed the PR by disabling NC inside of it. Please test.

Unfortunately, it does not include contacts, but all other things are shown properly. We could maybe look into this for 1.5.1, if it's all all possible without resorting to a custom view?

Thanks for the awesome contribution.

tobiasKaminsky:
> Unfortunately, it does not include contacts 

This was discussed in this issue and we agreed that it is better to have contacts (and nextcloud) in it than having no contacts, but filtered nextcloud...

mario:
@tobiasKaminsky I got the impression that this is because of the duplicate entries which I fixed.

Anyway, if @AndyScherzinger feels the same feel free to revert my commits and merge as-is.

tobiasKaminsky:
Right as said in https://github.com/nextcloud/android/pull/1244#issuecomment-319156043 

> Doesn't include the contacts. I didn't find an other solution including the contacts...

@AndyScherzinger revert to get the contacts back again?

AndyScherzinger:
@tobiasKaminsky  @mario I'd say revert if this brings back the contacts _and_ if sharing with contacts actually works :)

tobiasKaminsky:
:
Reverted and merged master.

mario:
Then merge

tobiasKaminsky:
Waiting for codacy ;-)

PR #1262, opened 2017-08-04
Revert description for clear successful uploads

tobiasKaminsky:


mario:
This has already been done in my branch. Do you want to merge it regardless?

tobiasKaminsky:
As translators from other languages may think that they have to change this (and later then would have to revert this), I think we should merge it.

If it is exactly the same on your branch it should not even result in a conflict ;-)

mario:
@tobiasKaminsky fine. Then merge right away. :)

Issue #1257, opened 2017-08-02
Check if LDAP works as expected


mario:
Check if search, favorites (and favoriting process) work as expected when LDAP is the backend.

If not, we probably need to get internal ID from somewhere instead of using the username.

mario:
Ok, so this is a serious issue and affect existing accounts too so we need a migration path.

Two use-cases:
1) open an app with accounts already in

Steps:
- poke /ocs/v2.php/cloud/user for each of the accounts
- replace username in account manager with "id" value found here:

https://usercontent.irccloud-cdn.com/file/8L6mFbyc/Bildschirmfoto%202017-08-02%20um%2009.02.00.png

Make sure if this procedure fails for any account, that it's repeated properly.

2) open an app with no accounts
- once you log-in, if success, poke /ocs/v2.php/cloud/user and store "id" as username
- only then proceed

cc @rullzer for confirmation
cc @tobiasKaminsky for potential implementation, since I have other issues to handle

mario:
Also cc @MorrisJobke since he likes to be involved in everything :P

MorrisJobke:
> Check if search, favorites (and favoriting process) work as expected when LDAP is the backend.

Why is this LDAP related?

You could also simulate this issue on android if you login via our email address. To do so you need to:

* setup an user account
* assign an email to this
* use the email address + password instead of username + password to log in into the Android app
* :boom:


Or am I wrong?

mario:
Sure. But the solution outlined above should solve this.

MorrisJobke:
> Sure. But the solution outlined above should solve this.

Correct. Just wanted to make it easier for you to test this, because then you don't need an LDAP setup.

mario:
You are awesome :)

rullzer:
Yes should work.

Or you do it the dav way and just fetch the current principal:

```xml
<d:propfind xmlns:d="DAV:">
  <d:prop>
     <d:current-user-principal />
  </d:prop>
</d:propfind>
```

MorrisJobke:
See also https://github.com/nextcloud/ios/issues/331#issuecomment-319905716 for a detailed explanation.

tobiasKaminsky:
Will do it on Friday

tobiasKaminsky:
With the new login flow the app gets the correct user id, even when login is done via email.

So this affects only 
- existing accounts
- new accounts on NC<12

AndyScherzinger:
merged for 1.5.0

PR #1265, opened 2017-08-04
Update user name with correct id found by user ocs call

tobiasKaminsky:
Ref: #1257 

There already was an existing function that updated accounts, so I extended this one.

So now there is no distinction between any NC version or old/new accounts, all are handled by the same:
- every existing/new account currently will have version 1
- an update is performed one time, correct id, set version to 2

We can discuss if and how we handle this once EOL of <NC12.


mario:
Cool. Will review later. Cookies PR coming too?:)

mario:
Looks decent, but please rebase.

tobiasKaminsky:
It is one click in github, you can also do this ;-)

mario:
👍

AndyScherzinger:
👍 fine by me but can we please please stop doing master merges to branches and use rebase :)

PR #1291, opened 2017-08-09
Remove unused ressources

tobiasKaminsky:


build.gradle
```diff
@@ -201,8 +201,8 @@ dependencies {
     implementation 'com.googlecode.ez-vcard:ez-vcard:0.10.2'
     implementation 'org.lukhnos:nnio:0.2'
     // uncomment for gplay, modified
-    // implementation "com.google.firebase:firebase-messaging:${googleLibraryVersion}"
-    // implementation "com.google.android.gms:play-services-base:${googleLibraryVersion}"
+ implementation "com.google.firebase:firebase-messaging:${googleLibraryVersion}"
+ implementation "com.google.android.gms:play-services-base:${googleLibraryVersion}"
```
AndyScherzinger: this is gplay, modified only

build.gradle
```diff
@@ -238,4 +238,4 @@ tasks.withType(Test) {
 }
 
 // uncomment for gplay, modified (must be at the bottom)
-//apply plugin: 'com.google.gms.google-services'
+apply plugin: 'com.google.gms.google-services'
```
AndyScherzinger: this is gplay, modified only

tobiasKaminsky: Nice catch :+1:

AndyScherzinger:
@tobiasKaminsky how should we proceed here? I removed a lot of completely unused strings from the main ```strings.xml``` which now leads to errors since we have translations without a master-item (since that is what I removed). So these errors will go away as soon as transifex will get the new master file and push an update for the translations :/

AndyScherzinger:
@tobiasKaminsky manually removed the translation strings so lint is happy which will be what transifex will do again after this PR has been merged to master :)

tobiasKaminsky:
Good one.
There is 
> disable 'MissingTranslation'

in build.gradle, but it works the other way around...

AndyScherzinger:
SO I'd say we should activate it after the next stable release.

tobiasKaminsky:
Then we get tons of warnings/errors if we add a new string to values/strings.xml as this check tests if the new key is translated in all languages, which will not be the case as this is done on a later point via transifex.

AndyScherzinger:
ah, okay. Then don't :)

AndyScherzinger:
👍 @mario

Issue #1307, opened 2017-08-11
Only show thumbnails on activities that have previews


nickvergessen:
![screenshot_2017-08-11-10-08-32](https://user-images.githubusercontent.com/213943/29205487-9209d676-7e7d-11e7-9aa2-e3bf63809e2e.png)

This looks very odd atm:
1. Persons don't have a thumbnail (maybe an avatar but that should then be inline)
2. Events/tasks don't have a thumbnail
3. Calendars/Task lists don't have a thumbnail

So don't show the grey box please ;)

PR: #1317

AndyScherzinger:
merged for release 2.0.0 🎉

PR #1317, opened 2017-08-11
No thumbnail for non-files

nickvergessen:
Quick fix for #1307 

Follow up should be to use `previews` from the API response instead.
Will try to do that too:
https://github.com/nextcloud/activity/blob/master/docs/endpoint-v2.md#activity-element


tobiasKaminsky:
:+1: welcome an team android :-D

AndyScherzinger:
👍 welcome 🚀

Issue #1322, opened 2017-08-11
Error deselecting hide file listing on share


jasonbayton:
### Actual behaviour
When enabling "hide file listing" then trying to disable it, the following error:

![image](https://user-images.githubusercontent.com/3178412/29227491-f01fd986-7ecd-11e7-98a5-0de9588034c8.png)

### Expected behaviour
Hide file listing should just disable
 
### Steps to reproduce
1. Share folder
2. Enable hide file listing
3. Try to disable it.

### Environment data
Android version: 7.1.1

Device model: Moto Z

Stock or customized system: Stock

Nextcloud app version: 2.0RC 

Nextcloud server version: 12.0.1

#### Web server error log
```
No server log generated from this
```

#### Nextcloud log (data/nextcloud.log)
```
No server log generated from this
```

tobiasKaminsky:
> Hide file listing should just disable

Why do you expect this? I rather except that one can use this function, so we need to find out why this error is thrown.

jasonbayton:
>Why do you expect this?

Because it's a switch - on/off. If I can switch it on, I should be able to turn it off.
Also, I can disable it on the server when choosing between file drop and another option

When the client stops crashing I'll recreate the issue.

tobiasKaminsky:
Sorry, I misunderstood it and thought that you wanted to have the option disabled and not that disabling should work ;-)
I'll have look into it.

jasonbayton:
:) fine, recreation just pending a working app.

tobiasKaminsky:
This is done in #1345 
Thank you very much for reporting this :+1:

PR #1345, opened 2017-08-15
Hide public upload: Permission needs to be 15 and not 31.

tobiasKaminsky:
Tested on NC10-master.
Ref: #1322

### DIFF
--- a/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
+++ b/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
@@ -479,8 +479,7 @@ public class FileOperationsHelper {
             OwnCloudVersion serverVersion = AccountUtils.getServerVersion(mFileActivity.getAccount());
 
             if (serverVersion != null && serverVersion.isNotReshareableFederatedSupported()) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -479,8 +479,7 @@ public void setHideFileListingPermissionsToShare(OCShare share, boolean hideFile
             OwnCloudVersion serverVersion = AccountUtils.getServerVersion(mFileActivity.getAccount());
 
             if (serverVersion != null && serverVersion.isNotReshareableFederatedSupported()) {
-                updateShareIntent.putExtra(OperationsService.EXTRA_SHARE_PERMISSIONS,
-                        OCShare.FEDERATED_PERMISSIONS_FOR_FOLDER_AFTER_OC9);
+                updateShareIntent.putExtra(OperationsService.EXTRA_SHARE_PERMISSIONS, 15);
```
AndyScherzinger: Don't have an env to test this, but can we make "15" a constant this way it is a magic number.

tobiasKaminsky:
Don't know where my head is today... :/
But this is completely useless as I changed the constant in lib: https://github.com/nextcloud/android-library/pull/74 so we can close this

Issue #1339, opened 2017-08-14
GPlay:   java.lang.StringIndexOutOfBoundsException - com.owncloud.android.ui.activity.UserInfoActivity.setHeaderImage [RC2]


AndyScherzinger:
java.lang.StringIndexOutOfBoundsException
com.owncloud.android.ui.activity.UserInfoActivity.setHeaderImage

```
java.lang.RuntimeException: 
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2927)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:2988)
  at android.app.ActivityThread.-wrap14 (ActivityThread.java)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1631)
  at android.os.Handler.dispatchMessage (Handler.java:102)
  at android.os.Looper.loop (Looper.java:154)
  at android.app.ActivityThread.main (ActivityThread.java:6682)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:1520)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1410)
Caused by: java.lang.StringIndexOutOfBoundsException: 
  at java.lang.String.charAt (Native Method)
  at android.graphics.Color.parseColor (Color.java:147)
  at com.owncloud.android.ui.activity.UserInfoActivity.setHeaderImage (UserInfoActivity.java:303)
  at com.owncloud.android.ui.activity.UserInfoActivity.onCreate (UserInfoActivity.java:217)
  at android.app.Activity.performCreate (Activity.java:6942)
  at android.app.Instrumentation.callActivityOnCreate (Instrumentation.java:1126)
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2880)
```

tobiasKaminsky:
This happens when colorString is empty, but not null, which can happen if capabilities are not correctly set.

AndyScherzinger:
Fix has been merged

PR #1352, opened 2017-08-16
Check if background string is empty

tobiasKaminsky:
Background string is by default empty, so if capability is not parsed correctly it leads to a crashing user info activity.
Fallback is the default image.

Ref: https://github.com/nextcloud/android/issues/1339


AndyScherzinger:
👍 fine by me

Issue #1336, opened 2017-08-14
GPlay:   java.lang.NullPointerException - com.owncloud.android.datamodel.MediaProvider.getImageFolders


AndyScherzinger:
java.lang.NullPointerException
com.owncloud.android.datamodel.MediaProvider.getImageFolders

* Reports this week 11
* _Android 6 only_
* _Fairphone FP2 only_

```
java.lang.RuntimeException: 
  at android.app.ActivityThread.handleBindApplication (ActivityThread.java:4710)
  at android.app.ActivityThread.access$1600 (ActivityThread.java:150)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1405)
  at android.os.Handler.dispatchMessage (Handler.java:102)
  at android.os.Looper.loop (Looper.java:148)
  at android.app.ActivityThread.main (ActivityThread.java:5417)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:726)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:616)
Caused by: java.lang.NullPointerException: 
  at com.owncloud.android.datamodel.MediaProvider.getImageFolders (MediaProvider.java:139)
  at com.owncloud.android.MainApp.splitOutAutoUploadEntries (MainApp.java:348)
 
  at com.owncloud.android.MainApp.onCreate (MainApp.java:132)
 
  at android.app.Instrumentation.callApplicationOnCreate (Instrumentation.java:1014)
 
  at android.app.ActivityThread.handleBindApplication (ActivityThread.java:4707)
```

tobiasKaminsky:
NPE in 
`
if (!mediaFolder.absolutePath.startsWith(dataPath))
`

AndyScherzinger:
yeah needs to be 

```
if (mediaFolder.absolutePath != null && !mediaFolder.absolutePath.startsWith(dataPath)) {
```

:/

tobiasKaminsky:
Already done ;-)

PR #1355, opened 2017-08-16
Prevent NPE on absolutePath

tobiasKaminsky:
Ref: #1336 

NPE in 
`
if (!mediaFolder.absolutePath.startsWith(dataPath))
`

maybe it is also in dataPath, but this would be way more critical:
`
String dataPath = MainApp.getStoragePath() + File.separator + MainApp.getDataFolder();
`
and then many other things would also crash.


AndyScherzinger:
👍 perfect :) As I stated in the issue...ufortunately ```mediaFolder.absolutePath``` can be ```null``` as seen in the locs above but it is not checked.

AndyScherzinger:
@tobiasKaminsky we should probably add the same logic for the videos... :/
Sorry I missed that part during the original implementation...

AndyScherzinger:
@tobiasKaminsky I hardened the video part accordingly https://github.com/nextcloud/android/pull/1355/commits/00d28a18a56c4d1840d7b1d34c601a9a1ea6e572

AndyScherzinger:
👍 I tested it on my phone to check if the video hardening works fine

tobiasKaminsky:
:+1: for your addition ;-)

AndyScherzinger:
so just awaiting @mario's review/approval

AndyScherzinger:
👍 rebased to update the lint score - cc @tobiasKaminsky

AndyScherzinger:
👍 rebased to update the lint score - cc @tobiasKaminsky

Issue #1340, opened 2017-08-14
GPlay:   android.util.AndroidRuntimeException - com.owncloud.android.authentication.AuthenticatorActivity.deleteCookies [RC2]


AndyScherzinger:
android.util.AndroidRuntimeException
com.owncloud.android.authentication.AuthenticatorActivity.deleteCookies

```
java.lang.RuntimeException: 
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2724)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:2789)
  at android.app.ActivityThread.-wrap12 (ActivityThread.java)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1527)
  at android.os.Handler.dispatchMessage (Handler.java:110)
  at android.os.Looper.loop (Looper.java:203)
  at android.app.ActivityThread.main (ActivityThread.java:6247)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:1063)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:924)
Caused by: android.util.AndroidRuntimeException: 
  at android.webkit.WebViewFactory.getProviderClass (WebViewFactory.java:371)
  at android.webkit.WebViewFactory.getProvider (WebViewFactory.java:194)
  at android.webkit.CookieManager.getInstance (CookieManager.java:39)
  at com.owncloud.android.authentication.AuthenticatorActivity.deleteCookies (AuthenticatorActivity.java:349)
  at com.owncloud.android.authentication.AuthenticatorActivity.onCreate (AuthenticatorActivity.java:263)
  at android.app.Activity.performCreate (Activity.java:6666)
  at android.app.Instrumentation.callActivityOnCreate (Instrumentation.java:1118)
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2677)
```

AndyScherzinger:
Fix merged

PR #1356, opened 2017-08-16
CookieManager.getInstance() can throw an Exception without saying so

tobiasKaminsky:
Ref: #1340 

If cookies cannot removed maybe the same credentials are used if an user tries to use the new weblogin flow for the same server.


AndyScherzinger:
👍 fine by me

Issue #1353, opened 2017-08-16
GPlay:  java.lang.ClassNotFoundException


tobiasKaminsky:
Version: 2.0.0RC2
Android: 7.x

```
java.lang.RuntimeException: 
  at android.app.ActivityThread.handleCreateService (ActivityThread.java:3177)
  at android.app.ActivityThread.-wrap5 (ActivityThread.java)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1568)
  at android.os.Handler.dispatchMessage (Handler.java:102)
  at android.os.Looper.loop (Looper.java:154)
  at android.app.ActivityThread.main (ActivityThread.java:6121)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:889)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:779)
Caused by: java.lang.ClassNotFoundException: 
  at dalvik.system.BaseDexClassLoader.findClass (BaseDexClassLoader.java:56)
  at java.lang.ClassLoader.loadClass (ClassLoader.java:380)
  at java.lang.ClassLoader.loadClass (ClassLoader.java:312)
  at android.app.ActivityThread.handleCreateService (ActivityThread.java:3174)
```

```
java.lang.RuntimeException: Unable to instantiate service com.evernote.android.job.gcm.PlatformGcmService: java.lang.ClassNotFoundException: Didn't find class "com.evernote.android.job.gcm.PlatformGcmService" on path: DexPathList[[zip file "/data/app/com.nextcloud.client-1/base.apk"],nativeLibraryDirectories=[/data/app/com.nextcloud.client-1/lib/arm64, /system/lib64, /vendor/lib64]]
	at android.app.ActivityThread.handleCreateService(ActivityThread.java:3177)
	at android.app.ActivityThread.-wrap5(ActivityThread.java)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1568)
	at android.os.Handler.dispatchMessage(Handler.java:102)
	at android.os.Looper.loop(Looper.java:154)
	at android.app.ActivityThread.main(ActivityThread.java:6121)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:889)
[...]

PR #1360, opened 2017-08-17
Add GCM to gplay/modified

tobiasKaminsky:
Ref: #1353 

Needs approval by @mario!


AndyScherzinger:
Will this then be re-added for the gplay build and is is necessary for the gplay build?

AndyScherzinger:
👍 code looks fine, open issue is the question above and the feedback of @mario

mario:
I would strongly recommend to not merge this branch. What needs to be done instead is find a separate dependency (previously, we used to include whole play services) that ships GCM network manager and put it in gradle for Modified and Gplay flavours. The reason is that we indeed want to rely on GCM when it's available, and since we use other play services functionality it's absolutely fine to use this one as well.

For F-droid build you don't need to ship this at all because we never shipped play services there.

tobiasKaminsky:
@mario I added gcm as an explicit dependency for gplay/modified.
After compiling I could access class "GcmTaskService", which was previously missing.
On "fdroid" build this class is not available, as expected.

mario:
Good. Feel free to merge.

tobiasKaminsky:
Merging regardless as we now get a new error which will be fixed via transifex.

Issue #1359, opened 2017-08-16
DEV/RC2: Sign on fails under some circumstances (maybe StartCom SSL issue)


DemoFreak:
### Actual behaviour
When trying to log in the applications screen remains blank forever just after tapping on right arrow icon next to the server address.

### Expected behaviour
Login screen appears for input of credentials.
 
### Steps to reproduce
1. install Dev/RC version
2. open, skip to the input of server address
3. input server address and tap right-arrow icon

### Comment
This behavior occurs with my Nextcloud server at the web address where the Startcom SSL certificate is issued to. The same server is accessible under another address (where the certificate isn't issued to). When trying to log into the server using this address a warning message appears asking me if I want to trust this certificate anyway. After affirmation all seems fine, the app works as expected (asking for grant access and login credentials).
When I go back at this point and try to use a the "right web address" instead a second time, this time it works also with this address.

### Environment data
Android version: 6.0.1
Device model: Galaxy Note4
Stock or customized system: stock
Nextcloud app version: 2.0 RC2
Nextcloud server version: 12.0.1

#### Web server access log
```
146.44.44.113 - - [17/Aug/2017:00:23:47 +0200] "GET /status.php HTTP/1.1" 200 5535 "-" "Mozilla/5.0 (Android) ownCloud-android/2.0.0RC2"
146.44.44.113 - - [17/Aug/2017:00:23:47 +0200] "HEAD /remote.php/webdav/ HTTP/1.1" 401 1206 "-" "Mozilla/5.0 (Android) ownCloud-android/2.0.0RC2"
```

#### Web server access log
empty

#### Nextcloud log (data/nextcloud.log)
First try:
```
{"reqId":"IsFEPqwRWq3EhGAF5FkN","level":0,"time":"2017-08-16T22:23:47+00:00","remoteAddr":"146.44.44.113","user":"--","app":"webdav","method":"HEAD","url":"\/remote.php\/webdav\/","message":"Exception: {\"Exception\":\"Sabre\\\\DAV\\\\Exception\\\\NotAuthenticated\",\"
Message\":\"No 'Authorization: Basic' header found. Either the client didn't send one, or the server is misconfigured, No 'Authorization: Bearer' header found. Either the client didn't send one, or the server is mis-configured\",\"Code\":0,\"Trace\":\"#0 [internal funct
[...]

PR #1361, opened 2017-08-17
Fix rare case: ssl cert is ok for checkServer, but not for webview

tobiasKaminsky:
If this is the case -> show untrusted dialog again.

Ref: #1359


AndyScherzinger:
👍 fine by me

PR #1362, opened 2017-08-17
Handle foreign broadcast

tobiasKaminsky:
Steps to reproduce
- install oc client
- install nc client
- start sync on oc client
- open nc client 
--> crash as nc client is listening to broadcast by oc client. Cause: oc client is not using internal broadcast / app specific broadcasts.

Related stacktraces:
`
java.lang.RuntimeException: Parcelable encountered IOException reading a Serializable object (name = com.owncloud.android.lib.common.operations.RemoteOperationResult)
`

`
java.lang.RuntimeException: Parcel android.os.Parcel@849613f: Unmarshalling unknown type code 3014765 at offset 4192
`


AndyScherzinger:
👍 fine by me

Issue #1365, opened 2017-08-17
2.0.0RC2 java.lang.StringIndexOutOfBoundsException on Auto Upload


taaem:
### Actual behaviour
When opening the auto upload page the app crashes.

### Expected behaviour
Open Auto Upload
 
### Steps to reproduce
1. Open App
2. Open Auto upload
3. Crash

### Environment data
Android version: 7.2.0

Device model: OnePlus X

Stock or customized system: customized

Nextcloud app version: 2.0.0RC2

Nextcloud server version: 12.0.2

#### Nextcloud log (logcat)
```
08-17 15:01:17.512 E/AndroidRuntime(8147): FATAL EXCEPTION: main
08-17 15:01:17.512 E/AndroidRuntime(8147): Process: com.nextcloud.client, PID: 8147
08-17 15:01:17.512 E/AndroidRuntime(8147): java.lang.RuntimeException: Unable to start activity ComponentInfo{com.nextcloud.client/com.owncloud.android.ui.activity.SyncedFoldersActivity}: java.lang.StringIndexOutOfBoundsException: length=5; regionStart=0; regionLength=-1
08-17 15:01:17.512 E/AndroidRuntime(8147): 	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2684)
08-17 15:01:17.512 E/AndroidRuntime(8147): 	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2745)
08-17 15:01:17.512 E/AndroidRuntime(8147): 	at android.app.ActivityThread.-wrap12(ActivityThread.java)
08-17 15:01:17.512 E/AndroidRuntime(8147): 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1492)
08-17 15:01:17.512 E/AndroidRuntime(8147): 	at android.os.Handler.dispatchMessage(Handler.java:102)
08-17 15:01:17.512 E/AndroidRuntime(8147): 	at android.os.Looper.loop(Looper.java:153)
08-17 15:01:17.512 E/AndroidRuntime(8147): 	at android.app.ActivityThread.main(ActivityThread.java:6208)
08-17 15:01:17.512 E/AndroidRuntime(8147): 	at java.lang.reflect.Method.invoke(Native Method)
08-17 15:01:17.512 E/AndroidRuntime(8147): 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:891)
[...]

PR #1367, opened 2017-08-18
Check if valid path

tobiasKaminsky:
Ref: #1365


AndyScherzinger:
:+1: right solution imho

PR #1377, opened 2017-08-19
Updated fr (French) translation

Credomo:
Commit made via Stringlate


AndyScherzinger:
Hi @Credomo, translations are purely handled via transifex: https://www.transifex.com/nextcloud/nextcloud/android/ so any direct changes within the repo would be overriden.

Issue #1386, opened 2017-08-21
GPlay: NPE com.owncloud.android.MainApp.getAuthority


AndyScherzinger:
```
java.lang.ExceptionInInitializerError: 
  at com.owncloud.android.services.observer.FileObserverService.startObservation (FileObserverService.java:197)
  at com.owncloud.android.services.observer.FileObserverService.onStartCommand (FileObserverService.java:167)
  at android.app.ActivityThread.handleServiceArgs (ActivityThread.java:3351)
  at android.app.ActivityThread.-wrap21 (ActivityThread.java)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1601)
  at android.os.Handler.dispatchMessage (Handler.java:102)
  at android.os.Looper.loop (Looper.java:154)
  at android.app.ActivityThread.main (ActivityThread.java:6186)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:889)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:779)
Caused by: java.lang.NullPointerException: 
  at com.owncloud.android.MainApp.getAuthority (MainApp.java:236)
  at com.owncloud.android.db.ProviderMeta$ProviderTableMeta.<clinit> (ProviderMeta.java:54)
```

happens since a while (e.g. 1.4.x).
causing line:
```
return getAppContext().getResources().getString(R.string.authority);
```

My guess is that ```getAppContext()``` returns null since the underlying variable hasn't been initialized yet. A quick fix would be to add a try catch and return the default value which will cause problems for custom/branded clients who run into the NPE fallback then...

Any ideas @tobiasKaminsky @mario ? This is the number 1 crash reported in the actual RC and for all app versions has been reported nearly 3000 times...

tobiasKaminsky:
> A quick fix would be to add a try catch and return the default value which will cause problems for custom/branded clients who run into the NPE fallback then...

There should be no problem as a branded client also has a change R.strnig.authority.
But how can we get the string if getAppContext() is null...

AndyScherzinger:
We don't... I'd have implemented something like this:

```
try {
            return getAppContext().getResources().getString(R.string.authority);
        } catch (Exception e) {
            return "org.nextcloud";
        }
```
[...]

PR #1390, opened 2017-08-21
if context is null, try to get set it

tobiasKaminsky:
I am not 100% sure if this works.
The problem seems that somehow the service is started prior the correct init of MainApp.

Ref: #1386


AndyScherzinger:
👍 hard to tell but the code won't hurt so it might fix the issue, so we should ship it with the next RC to see if the error still gets reported afterwards.

PR #1421, opened 2017-08-24
Start using the previews from the server

nickvergessen:
- [ ] API should return the mimetype as well, so the clients can use their local mimetype icons if available
- [ ] API should return path + id so apps can link to the file locally instead of redirecting to the browser
- [ ] Readd onClick handler (local when available otherwise browser (trashbin))
- [ ] Kill `setBitmap` after local mimetype icons are used

---

- [ ] Requires https://github.com/nextcloud/android-library/pull/73


tobiasKaminsky:
@nickvergessen do we want to finish this during hackweek (we/thu)?

nickvergessen:
I dont think I got time

tobiasKaminsky:
@nickvergessen once the activity API returns mentioned parts (mimetype, fileId, …) please re-open or create an issue, so that we can develop this.
Keeping this old PR just causes overhead while trying to keep it up to date.

PR #1424, opened 2017-08-24
Activities list activity: fix hsl computation, fix header color

### LINKED ISSUE #1422: Font problem on bright primary color
Use theming color: `#DFFFF4`

❌ Font is white instead of black
❌ Some activity items are blue instead of the color
🔧 When the theming color is too bright, we should use black as a font in activities

> bildschirmfoto vom 2017-08-24 15-01-55

tobiasKaminsky:
Computation of hsl was wrong, therefore always a white font was shown.

Fix #1422

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/ActivitiesListActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/ActivitiesListActivity.java
@@ -26,11 +26,12 @@ import android.accounts.AuthenticatorException;
 import android.accounts.OperationCanceledException;
 import android.content.Context;
 import android.content.Intent;
-import android.os.AsyncTask;
 import android.graphics.PorterDuff;
+import android.os.AsyncTask;
 import android.os.Bundle;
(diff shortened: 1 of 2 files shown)


AndyScherzinger:
👍 fine!

src/main/java/com/owncloud/android/ui/activity/ActivitiesListActivity.java
```diff
@@ -134,7 +135,8 @@ protected void onCreate(Bundle savedInstanceState) {
 
         // setup drawer
         setupDrawer(R.id.nav_activity);
-        getSupportActionBar().setTitle(getString(R.string.drawer_item_activities));
+        ActionBar actionBar = getSupportActionBar();
+        ThemeUtils.setColoredTitle(actionBar, getString(R.string.drawer_item_activities));
```
AndyScherzinger: Please add a null check for actionBar

AndyScherzinger:
👍 fine!

mario:
👍

PR #1465, opened 2017-08-31
Lint: only upload new master on decrease of warning/error

tobiasKaminsky:
This way we avoid having the "same" master lint file uploaded over and over again.


AndyScherzinger:
👍

mario:
👍

PR #1484, opened 2017-09-03
Folder-Based Sort Order

### LINKED ISSUE #1441: separate sort order for each folder
### Actual behaviour
- all folders have the same sort order

### Expected behaviour
- I tend to have list views sorted alphanumerically
- I tend to have tile views sorted new to old
- once I decided by which order a folder should sorted, it should stay this way (the same already happens when I switched from list view to tile view)

### Environment data
Android version: Android 7.1.1

Device model: OnePlus 5

Nextcloud app version: 2.0.0RC4

Nextcloud server version: 11.0.4

srkunze:
Closes #1441 .

### DIFF
--- a/src/main/java/com/owncloud/android/db/PreferenceManager.java
+++ b/src/main/java/com/owncloud/android/db/PreferenceManager.java
@@ -25,7 +25,7 @@ import android.content.SharedPreferences;
 import com.owncloud.android.datamodel.FileDataStorageManager;
 import com.owncloud.android.datamodel.OCFile;
 import com.owncloud.android.ui.activity.ComponentsGetter;
-import com.owncloud.android.utils.FileStorageUtils;
+import com.owncloud.android.utils.FileSortOrder;
 
 /**
  * Helper to simplify reading of Preferences all around the app
@@ -36,8 +36,7 @@ public abstract class PreferenceManager {
      * Value handled by the app without direct access in the UI.
      */
     private static final String AUTO_PREF__LAST_UPLOAD_PATH = "last_upload_path";
-    private static final String AUTO_PREF__SORT_ORDER = "sort_order";
-    private static final String AUTO_PREF__SORT_ASCENDING = "sort_ascending";
+    private static final String AUTO_PREF__SORT_ORDER_NAME = "sort_order_name";
     private static final String AUTO_PREF__UPLOAD_FILE_EXTENSION_MAP_URL = "prefs_upload_file_extension_map_url";
     private static final String AUTO_PREF__UPLOAD_FILE_EXTENSION_URL = "prefs_upload_file_extension_url";
     private static final String AUTO_PREF__UPLOADER_BEHAVIOR = "prefs_uploader_behaviour";
@@ -199,40 +198,21 @@ public abstract class PreferenceManager {
      * Gets the sort order which the user has set last.
      *
      * @param context Caller {@link Context}, used to access to shared preferences manager.
-     * @return sort order     the sort order, default is {@link FileStorageUtils#SORT_NAME} (sort by name)
+     * @return sort order     the sort order, default is {@link FileSortOrder#sort_a_to_z} (sort by name)
      */
-    public static int getSortOrder(Context context) {
-        return getDefaultSharedPreferences(context).getInt(AUTO_PREF__SORT_ORDER, FileStorageUtils.SORT_NAME);
+    public static FileSortOrder getSortOrder(Context context) {
+        String name = getDefaultSharedPreferences(context).getString(AUTO_PREF__SORT_ORDER_NAME, FileSortOrder.sort_a_to_z.mName);
+        return FileSortOrder.sortOrders.get(name);
     }
 
     /**
      * Save the sort order which the user has set last.
      *
      * @param context Caller {@link Context}, used to access to shared preferences manager.
-     * @param order   the sort order
+     * @param sortOrder   the sort order
      */
-    public static void setSortOrder(Context context, int order) {
-        saveIntPreference(context, AUTO_PREF__SORT_ORDER, order);
-    }
-
-    /**
-     * Gets the ascending order flag which the user has set last.
-     *
-     * @param context Caller {@link Context}, used to access to shared preferences manager.
-     * @return ascending order     the ascending order, default is true
-     */
-    public static boolean getSortAscending(Context context) {
-        return getDefaultSharedPreferences(context).getBoolean(AUTO_PREF__SORT_ASCENDING, true);
-    }
-
-    /**
-     * Saves the ascending order flag which the user has set last.
-     *
-     * @param context   Caller {@link Context}, used to access to shared preferences manager.
-     * @param ascending flag if sorting is ascending or descending
-     */
-    public static void setSortAscending(Context context, boolean ascending) {
-        saveBooleanPreference(context, AUTO_PREF__SORT_ASCENDING, ascending);
+    public static void setSortOrder(Context context, FileSortOrder sortOrder) {
+        saveStringPreference(context, AUTO_PREF__SORT_ORDER_NAME, sortOrder.mName);
     }
 
     /**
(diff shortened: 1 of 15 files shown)


AndyScherzinger:
@srkunze I do love the changes you brought in especially the overall solution of handling the data and the cleanup work, really makes it nice and tidy! 👍  One thing that you should add is a header to all new classes, since it is your work you have done 😃: 
```
/**
 * Nextcloud Android client application
 *
 * @author Sven R. Kunze
 * Copyright (C) 2017 Sven R. Kunze
 *
 * This program is free software; you can redistribute it and/or
 * modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
 * License as published by the Free Software Foundation; either
 * version 3 of the License, or any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU AFFERO GENERAL PUBLIC LICENSE for more details.
 *
 * You should have received a copy of the GNU Affero General Public
 * License along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
```

mario:
First of all, thanks for the contribution! Looking over the changes, I find the overall style fine but have one major gripe with the implementation. You seem to store the parent id and sort order in the preferences which is kind of strange, and I'm not sure will easily allow us to clean things up (say once that parent id doesn't exist anymore?!).

Can I suggest that you use the arbitrary storage provider we have implemented for this?

Of course, if there's something I've missed, let me know - I'm always happy to reconsider if I've overlooked something :)

srkunze:
@AndyScherzinger Done. :)

@mario You're welcome. As outlined in https://github.com/nextcloud/android/issues/1441#issuecomment-326631364, I based the changes on already existing code which does what you described.

What do you mean when you refer to 'arbitrary storage provider'. We need to store sort orders for non-cloudy files also. Could you point me to the right location? I also suggest, we handle this via a new issue/pull request. I find these changes here already big enough. :)

mario:
@srkunze I'd rather do it here since this won't be merged right away anyway. :-/ Arbitrary data provider code can be found here and is basically a key/value store we use to store additional data.

https://github.com/nextcloud/android/blob/bfe2b71210560f8e65eda801d102cc9571a21a2b/src/main/java/com/owncloud/android/datamodel/ArbitraryDataProvider.java

If you look at its usages, you'll see we use it to even store JSON serializations of objects inside the value part, and then deserialize when needed.

So I'd use it like this:

storeOrUpdateKeyValue(account.name, sort_path_to_whatever, value)

Notice how I pre-pend sort_ to the path so we can later do a LIKE search for a sort for a particular path.

The existing code used preferences, but was very generic and wasn't dir specific, so we never really had need for advanced manipulation like removing this preference if dir disappears. Here we will have such a need, so I want to make sure it's properly designed in the first place.

Hope this clarifies things? :)

srkunze:
> The existing code used preferences, but was very generic and wasn't dir specific, so we never really had need for advanced manipulation like removing this preference if dir disappears. Here we will have such a need, so I want to make sure it's properly designed in the first place.

That's not the case. As I said, I use existing code which uses dir-specific preferences. :)

mario:
Then my apologies, I will have to look into this in more detail. But the
question stands:

How do you clean up preference entries once say a folder dissappears?:)

srkunze:
> How do you clean up preference entries once say a folder dissappears?:)

Good question.

Well, first of all, I would evaluate if that's really an issue.
If it's really an issue, I then would ask the person who wrote isGridViewPreferred, what his evaluation was, and how he intended to solve it if at all. :)

If I had to solve it, I tend to use a database having foreign keys and constraints available. So, in case a folder gets removed, we only need to remove all sort-entries for all users of that particular folder.
Folder ids also have the benefit of being path-unrelated. So, less hassle for a user, if he moves some dirs.

My preliminary evaluation tends to "it's not an issue". I cannot imagine that those few bytes are a problem at all. Take this with a grain of salt, as I am new to the project, I might mis-estimate here.

srkunze:
Looking at storeOrUpdateKeyValue, I see that it has the benefits of storing sort orders and grid layout per user account.

Was that the primary reason for developing it? Other than that, SharedPreference looks also like a key-value store. So, changing that should be easy. :)

AndyScherzinger:
Afaik @tobiasKaminsky wrote the grid part back in the days :)

mario:
@srkunze talking practically, it might not be an issue. Talking good design, it is an issue.

As for Arbitrary Data ... yes, it was one of the ideas behind it. You'd basically have "sort_13245" (as an example, where 13245 is folder ID) as a key, and value as the actual sort preference.

I would not mind moving layout preferences to Arbitrary Data as well ... "layout_12345" and then "grid"/"list" as value. Wishful thinking? :)

srkunze:
@mario I suggest we take the discussion of this topic to a separate issue/pull request. Also because, when really needed, I would want to do it in a separate pull request, which is based on this one here.

AndyScherzinger:
I'd be perfectly fine with @srkunze's proposed approach on how to improve the sorting in an iterative way. What do you think @mario @tobiasKaminsky ? Imho this PR is (still) an improvement.

mario:
👍 
Thanks for the awesome contribution!

Approving, but I want to see my suggestions implemented later on. :)

srkunze:
Thanks for approving. Don't worry, @mario. I will look into it.

srkunze:
@mario I see that you sign off your commits. Do I need to do it as well? I remember it from other projects being mandatory to be merged.

mario:
@srkunze much appreciated, but not mandatory yet. So maybe in the future? :)

srkunze:
@mario I can rebase and commit amend. So no big deal. Can do it this evening.

Just to understand correctly, what's the purpose of signing off on the first place?

mario:
@srkunze the accepted answer here explains it better than I ever could :)

mario:
https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for

AndyScherzinger:
> Just to understand correctly, what's the purpose of signing off on the first place?

You thereby confirm that you did create the code yourself and have the right to put in under the license defined by this project.

When you rebase it tonight, can you also fix the codacy issue mentioned here please?: https://github.com/nextcloud/android/pull/1484#discussion_r136863368 that be great 😺

srkunze:
@AndyScherzinger Rebased.

Just to make sure I can fix the codacy issue correctly. The method does iterative recursion. So, I need to reassign a variable to itself. What's wrong with using the parameter variable for it?

AndyScherzinger:
> Just to make sure I can fix the codacy issue correctly. The method does iterative recursion. So, I need to reassign a variable to itself. What's wrong with using the parameter variable for it?

Well, codacy spits out a warning since Java (checks) consider it good behavior to not manipulate input parameters. Especially in this case since this is a call by reference so you could call setter methods on the folder (which you don't) and actually manipulate/change the original object of the caller. I'd be fine not leave it this way if it is too much of a hassle and can configure an ignore flag for this warning in codacy so the warning goes away.

srkunze:
@AndyScherzinger I let you decide as I am new to the project. :) It's an easy to add yet-another variable to pass the same object around. But it's still wouldn't prevent me from calling setters (if I would).

AndyScherzinger:
I know ;)
So I vote to leave it like it is for the reasons you mentioned. Hardcore fix possible though would be to make it immutable 😛 But not worth it.

What do you think @mario @tobiasKaminsky

mario:
Can we close this in favour of PR #1571 ?

AndyScherzinger:
We could still merge this one first and merge the other one on top after. Whenever ready to merge.

mario:
I would just merge the other one. Needs one tiny change though

AndyScherzinger:
Also fine. What do you think @srkunze?

srkunze:
Whatever you prefer. 🙂

PR #1485, opened 2017-09-03
Update SETUP.md

srkunze:
couldn't pull upstream:

Permission denied (publickey).
fatal: Could not read from remote repository.
Please make sure you have the correct access rights and the repository exists.

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -35,7 +35,7 @@ Next steps will assume you have a Github account and that you will get the code 
 * Clone your forked repository: ```git clone --recursive git@github.com:YOURGITHUBNAME/android.git```.
 * Move to the project folder with ```cd android```.
 * Pull any changes from your remote branch 'master': ```git pull origin master```
-* Make official Nextcloud repo known as upstream: ```git remote add upstream git@github.com:nextcloud/android.git```
+* Make official Nextcloud repo known as upstream: ```git remote add upstream https://github.com/nextcloud/android.git```
 * Make sure to get the latest changes from official android/master branch: ```git pull upstream master```
 
 At this point you can continue using different tools to build the project. Section 2 and 3 describe the existing alternatives.
--- a/SETUP.md
+++ b/SETUP.md
@@ -32,7 +32,7 @@ Next steps will assume you have a Github account and that you will get the code 
 
 * In a web browser, go to https://github.com/nextcloud/android, and click the 'Fork' button near the top right corner.
 * Open a terminal and go on with the next steps in it.
-* Clone your forked repository: ```git clone --recursive git@github.com:YOURGITHUBNAME/android.git```.
+* Clone your forked repository: ```git clone --recursive https://github.com:YOURGITHUBNAME/android.git```.
 * Move to the project folder with ```cd android```.
 * Pull any changes from your remote branch 'master': ```git pull origin master```
 * Make official Nextcloud repo known as upstream: ```git remote add upstream https://github.com/nextcloud/android.git```


AndyScherzinger:
@LukasReschke @jancborchardt can one of you add @srkunze to the organization please? :)

AndyScherzinger:
@srkunze can you pull upstream in case of using https: instead of git: ?

srkunze:
@AndyScherzinger Indeed. Don't know what's different to the git: protocol but using https: I could pull.

My .git/config
```
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
[remote "origin"]
	url = https://github.com/srkunze/android.git
	fetch = +refs/heads/*:refs/remotes/origin/*
[branch "master"]
	remote = origin
	merge = refs/heads/master
[remote "upstream"]
	url = https://github.com/nextcloud/android.git
	fetch = +refs/heads/*:refs/remotes/upstream/*
[branch "folder-based-sort-order"]
	remote = origin
	merge = refs/heads/folder-based-sort-order
```

AndyScherzinger:
Strange, but whatever solves the issue ;) I can't use git either but that is due to my employer's network...

strugee:
`git@` is SSH shorthand, which GitHub requires you to have set up in advance. (I.e. you need an SSH key [added to your account](https://github.com/settings/keys) to even read repositories). On the other hand `https://` is, well, HTTPS, which GitHub lets people use to read repos without any kind of login.

`git://`, on the gripping hand, is a custom Git protocol. IIRC it's basically SSH but without all the crypto. GitHub allows anonymous Git protocol access too but I wouldn't recommend it because the protocol itself(?) doesn't support writes and also can be trivially tampered with because it's basically plaintext :)

mario:
Um, pretty sure you can write via git protocol. Anyway, irrelevant ... The point is we should rather tell people to setup their keys properly for git@ than move to https. Or mention both and let people choose :P

AndyScherzinger:
I'd say mention both since for example people on corporate networks (like me throughout the day) can't use git but have to use https.
So having both described lowers the barrier... :)

srkunze:
@strugee As soon as mentioned it, I now remember git: being based on ssh:. So, it definitely needs some kind of setup on the users machine.

@AndyScherzinger I like the simplicity of only proposing only one way. https: just works for both push and pull. How would you phrase this passage with both protocols?

AndyScherzinger:
I am totally fine with just describing the https way to keep the barrier at a minimum and leave out git. While pros will know how to set it up using git and likely don't need this Infos at all.

jancborchardt:
Yep - simple instructions which just work are better than educating people. I think this change is good :+1:

AndyScherzinger:
Sweet! @srkunze would you mind changing the other git url to https also? That way everything should then be described for https:)

srkunze:
Done. :)

mario:
👍

AndyScherzinger:
:+1:

AndyScherzinger:
Merged 🎉

SETUP.md
```diff
@@ -32,7 +32,7 @@ Next steps will assume you have a Github account and that you will get the code
 
 * In a web browser, go to https://github.com/nextcloud/android, and click the 'Fork' button near the top right corner.
 * Open a terminal and go on with the next steps in it.
-* Clone your forked repository: ```git clone --recursive git@github.com:YOURGITHUBNAME/android.git```.
+* Clone your forked repository: ```git clone --recursive https://github.com:YOURGITHUBNAME/android.git```.
```
comradekingu: There shouldn't be a ":" after github.com when using HTTPS.

Should be `git clone --recursive https://github.com/YOURGITHUBNAME/android.git`

comradekingu:
See #1499

PR #1499, opened 2017-09-05
No : in git HTTPS link, GitHub, …

comradekingu:
Fixes #1485

Having the ":" in the URL means a port-number would have to follow it.


AndyScherzinger:
:+1: missed that in the first review. Nice catch!

tobiasKaminsky:
:+1: Thank you very much :-)

PR #1503, opened 2017-09-06
Lint: Add Plural strings

### LINKED ISSUE #1513: Footer text wrong in languages with multiple plurals
### Actual behaviour
OCFileListFragment.generateFooterText doesn't honor languages with multiple plurals. In those languages, translations are limited (and wrong). For example, "file_list__footer__files" is not plural resource and as such, will not make sense in those languages

### Expected behaviour
Use plural resource to show files and folders in footer. Ideally, logic of this method could be simplified greatly.
 
### Steps to reproduce
1. Change language of phone to language with multiple plurals (Serbian is example I know)
2. Notice that generated translations are nonsensical and weird

### Environment data
Android version:
Dev

AndyScherzinger:
converting Strings which need to be available in different quantities into plurals.

While it is a lot of changed files you only need to check the Java classes since anything else are the ```strings.xml``` files, maybe check the default one from ```values```.

resolves #1513


AndyScherzinger:
@tobiasKaminsky @mario While it is a lot of changed files you only need to check the Java classes since anything else are the ```strings.xml``` files, maybe check the default one from ```values```.

AndyScherzinger:
Would love to have this in 2.0.1 to get this to master :)

AndyScherzinger:
beware that we should not merge this PR before transifex hasn't fixed their plurals bug because else the fixed warnings will turn into errors as soon as they have been translated to Ukrainian...

@tobiasKaminsky do you have any news from them regarding this issue?

tobiasKaminsky:
> @tobiasKaminsky do you have any news from them regarding this issue?

I asked them once again via mail.

tobiasKaminsky:
Reply from transifex: no new news :/

AndyScherzinger:
pinging @mario @tobiasKaminsky for review :)

mario:
👍

tobiasKaminsky:
> beware that we should not merge this PR before transifex hasn't fixed their plurals bug because else the fixed warnings will turn into errors as soon as they have been translated to Ukrainian...

I think we should merge this now and then adjust the error count as otherwise this PR will stay there forever...

mario:
Do it

AndyScherzinger:
@tobiasKaminsky everything green, just awaiting your approval 😃

tobiasKaminsky:
:+1:

PR #1504, opened 2017-09-07
Use user id returned by server instead of entered username

tobiasKaminsky:
Scenario:
- Server <NC12 (or no weblogin flow)
- user enters email-adress as username and password (or others, like ldap propertiers)
- user id returned by server is now used, instead of username login field value


mario:
Have you tested favorites, login, search on regular and LDAP logins?

Because Afaik (at least how I figured it out) is that entered username is
for login, but ID is for all other APIs. Hopefully :)

tobiasKaminsky:
My only ldap server ist NC12, so this uses weblogin flow.
But I tested it on NC11 via email and there favorite is working, search is not available on NC11.

AndyScherzinger:
👍 I only tested it with my only instance Nc12 no LDAP: login, seach, favs, photos, etc. worked fine for me

mario:
This should have nothing to do with NC12 at all though :P

AndyScherzinger:
I know, just provided the feedback I can give due to the limitations of my server setup :)

AndyScherzinger:
@tobiasKaminsky @mario so is this then ready to merge or does it need further changes? Just asking since it is the last remaining open PR for RC6.

mario:
I am testing atm

mario:
👎 
This will not fix anything. Let me try to create a solution.

mario:
@tobiasKaminsky can we delete this?

PR #1505, opened 2017-09-07
Fix for authentication/favorite/etc

mario:
Keep in mind this means people who installed the beta will have to remove/re-add account, but this also fixes theming for 9/10/11.

Will also be tested further once I get access.


AndyScherzinger:
> Keep in mind this means people who installed the beta

Why? Is this a breaking change for beta users or for any upgrade as in 1.4.3 users too? Is there a why to migrate the accounts? Re-Adding is tricky since we then need to tell beta users (which we can't except we do a custom pop-up...)

mario:
This is a breaking change only for beta users. We have to do this because there is no way to revert to previously entered username (since we don't know it!).

AndyScherzinger:
So it is related to the web login then since that is the mayor change with 2.0.0, right?

mario:
Actually no. All things v12 should work as-is AFAIK. The issue exists with certain actions (in this case Favoriting) on LDAP configurations pre-12. And like you know, we didn't have this before 2.0.0 :)

AndyScherzinger:
So it only hits beta users with ldap or anyone? Which would be my follow up question trying to figure out how many users would actually be affected by this breaking change.

mario:
All beta users were upgraded to Account Manager level2 account. So this affects everyone who installed the beta.

AndyScherzinger:
Okay :/

AndyScherzinger:
@tobiasKaminsky can you test this PR? I don't have a fitting setup. Also you can ignore the lint warning, this is due to the Snapshot reference which would be gone before merging to master :)

tobiasKaminsky:
:-1: 

Tested on NC11:
with user/pass: tobi/tobi

```
tcloud.client E/GetRemoteUserInfoOperation: Exception while getting OC user information
                                                                                  org.json.JSONException: Value [] at data of type org.json.JSONArray cannot be converted to JSONObject
                                                                                      at org.json.JSON.typeMismatch(JSON.java:100)
                                                                                      at org.json.JSONObject.getJSONObject(JSONObject.java:613)
                                                                                      at com.owncloud.android.lib.resources.users.GetRemoteUserInfoOperation.run(GetRemoteUserInfoOperation.java:145)
                                                                                      at com.owncloud.android.lib.common.operations.RemoteOperation.execute(RemoteOperation.java:136)
                                                                                      at com.owncloud.android.authentication.AuthenticatorAsyncTask.doInBackground(AuthenticatorAsyncTask.java:84)
                                                                                      at com.owncloud.android.authentication.AuthenticatorAsyncTask.doInBackground(AuthenticatorAsyncTask.java:41)
                                                                                      at android.os.AsyncTask$2.call(AsyncTask.java:304)
                                                                                      at java.util.concurrent.FutureTask.run(FutureTask.java:237)
                                                                                      at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:243)
                                                                                      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
                                                                                      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
                                                                                      at java.lang.Thread.run(Thread.java:761)
```

mario:
I tried NC11 o.o can you send me server url pls

mario:
Just tried this on two separate instances - LDAP and non-LDAP 12 and 11. Works on all looks like...

tobiasKaminsky:
@AndyScherzinger can you please approve this? :-)

AndyScherzinger:
:+1: sure

Issue #1277, opened 2017-08-07
AutoUpload: not below certain battery percentage


tobiasKaminsky:
Idea via email:
Do not use automatic loading at battery level of 30% (10%, 20%, 30% ...)

tobiasKaminsky:
My opinion is that auto upload should obey the power saving settings, e.g. 15% or 5% if the user set this in the android settings, otherwise auto upload should always work.

-> no new setting in our app needed, stay consistent with regular system

mario:
We can simply disable detection jobs & don't do any uploads when in power saver mode.

mario:
(This is possible for 5+ at least)

mario:
This is quite easy, so 1.5.1.

sosed213:
I recently returned from a trip and photographed a lot on a smartphone. Such a setup would very much help me out.

tobiasKaminsky:
My idea was not to disable detection, because then we will miss uploads and this should never happen.
The intention is only to disable upload as this might be very battery consumpting if e.g. ~2gb are uploaded.

mario:
No we would not miss uploads? They would appear once you are not in power
save mode.

tobiasKaminsky:
Ah, right yes. I confused something :-/
Then: if "battery saver" is enabled (done via android preferences)
- disable upload (show some message per each job in uplad view, maybe "upload disabled due to battery saver")
- disable detection job

mario:
Two jobs actually. But yes.

mario:
@sosed213 not yet. We're still working on 1.5.0 :)

sosed213:
I currently have version 1.4.3

mario:
Feature rather than bugfix, so 2.1.0?

stalker314314:
I'll have a stab a this

AndyScherzinger:
Sounds great, looking forward to it :)
If you have any questions don't hesitate!

@mario is the job scheduling expert around but @tobiasKaminsky and I will also support you in any way possible :+1:

AndyScherzinger:
PR for Android 5+ using the power save mode detection #1527

AndyScherzinger:
implemented via #1527 (battery saver mode)

PR #1527, opened 2017-09-10
Delaying upload if device is in power save mode

stalker314314:
This change adds handling for uploads from device when it is operating in power save mode and is part of the fix for #1277. In those case, uploads are behaving similar as if upload requires Wi-Fi (or that device needs to be charging) and those conditions are not met - they are delayed until device exits power save mode. Same thing holds true for contacts backup and explicit uploads - they are not delayed. Difference with Wi-Fi (and charging) conditions is in that 1) this delay cannot be set per synced folder and 2) it is turned on by default. For devices that don't have this capability (API level < 21) - old behavior (backward-compatible) is used; they behave as this check is not present.

Also added is listener for power save mode, albeit it makes sense only if user explicitly turns off power save mode while it is in it, as putting device to be charged will restart all jobs already by design (another listener that already exists).

### DIFF
--- a/src/main/java/com/owncloud/android/jobs/FilesSyncJob.java
+++ b/src/main/java/com/owncloud/android/jobs/FilesSyncJob.java
@@ -47,6 +47,7 @@ import com.owncloud.android.ui.activity.Preferences;
 import com.owncloud.android.utils.FileStorageUtils;
 import com.owncloud.android.utils.FilesSyncHelper;
 import com.owncloud.android.utils.MimeTypeUtil;
+import com.owncloud.android.utils.UploadUtils;
 
 import java.io.File;
 import java.io.IOException;
@@ -66,6 +67,7 @@ public class FilesSyncJob extends Job {
     public static final String TAG = "FilesSyncJob";
 
     public static final String SKIP_CUSTOM = "skipCustom";
+    public static final String OVERRIDE_POWER_SAVING = "overridePowerSaving";
 
     @NonNull
     @Override
@@ -81,6 +83,12 @@ public class FilesSyncJob extends Job {
 
         PersistableBundleCompat bundle = params.getExtras();
         final boolean skipCustom = bundle.getBoolean(SKIP_CUSTOM, false);
+        final boolean overridePowerSaving = bundle.getBoolean(OVERRIDE_POWER_SAVING, false);
+
+        // If we are in power save mode, better to postpone upload
+        if (UploadUtils.isPowerSaveMode(context) && !overridePowerSaving) {
+            return Result.SUCCESS;
+        }
 
         Resources resources = MainApp.getAppContext().getResources();
         boolean lightVersion = resources.getBoolean(R.bool.syncedFolder_light);
--- a/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
+++ b/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
@@ -48,6 +48,7 @@ import com.owncloud.android.operations.common.SyncOperation;
 import com.owncloud.android.utils.FileStorageUtils;
 import com.owncloud.android.utils.MimeType;
 import com.owncloud.android.utils.MimeTypeUtil;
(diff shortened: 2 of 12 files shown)


mario:
In general nice, but you don't need an additional variable to see if you need to ignore Power Saving or not. Basically if created_by is UploadFileOperation.CREATED_BY_USER then IGNORE, otherwise don't upload during power saving.

stalker314314:
So, if I understand you correctly @mario, you are saying there is no need to pass "KEY_IGNORE_POWER_SAVE_MODE" in intent at all - everything can be deduced from "created_by"? Seems plausible. Seems not explicit enough for my taste, but I am OK with it. Also, if we ever want to move this as synced folder setting, it will mean digging this path again. If there is no reply, I will assume deducing from "created_by" - it will make this PR a lot slimmer:)

mario:
This is indeed correct :)

I don't believe we want more settings for Synced folders, and in 99,99% you don't want automatic uploads to happen if you're thin on battery. I'm fine with that :)

mario:
File UploadFileOperation.java and getCreatedBy() is where the magic is :D

mario:
This is beyond awesome! Thank you very much! :)

👍

mario:
@AndyScherzinger can you approve?

src/main/java/com/owncloud/android/jobs/FilesSyncJob.java
```diff
@@ -81,6 +83,13 @@ protected Result onRunJob(Params params) {
 
         PersistableBundleCompat bundle = params.getExtras();
         final boolean skipCustom = bundle.getBoolean(SKIP_CUSTOM, false);
+        final boolean overridePowerSaving = bundle.getBoolean(OVERRIDE_POWER_SAVING, false);
+
+        // If we are in power save mode, better to postpone upload
+        if (UploadUtils.isPowerSaveMode(context) && !overridePowerSaving)
+        {
```
mario: Please put this in line above, like:
) {

src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -347,6 +353,13 @@ protected RemoteOperationResult run(OwnCloudClient client) {
                 return new RemoteOperationResult(ResultCode.DELAYED_FOR_CHARGING);
             }
 
+            // Check that device is not in power save mode
+            if (!mIgnoringPowerSaveMode && UploadUtils.isPowerSaveMode(mContext))
+            {
```
mario: Also, put { in previous line.

mario:
@stalker314314 if you know how to rebase against master, I'd appreciate it! :) Otherwise awesome!

AndyScherzinger:
@mario @stalker314314 I'll be doing a rebase now (@stalker314314 being away) to reduce it to the actual changes of the PR itself.

AndyScherzinger:
@mario @stalker314314 @tobiasKaminsky I assigned it to release 2.1 since we shouldn't integrate more features into 2.0 but get that one done so we can go straight for the next release/rc/bugfix/etc. 🚀 
As also mentioned/discussed on IRC.

Will test drive it probably this or next week 🚗

AndyScherzinger:
@stalker314314 just in case you are not used to rebase (especially since I did it on your repo) you will likely have to clone/checkout your branch again since I rewrote your commit history... I'd be fine to take care of the rebasing for you in case you can't. Just let me now, from my understanding except code review and test this PR is code complete 😃

mario:
Needs just one more approval ;)

AndyScherzinger:
I know but I am sure @tobiasKaminsky would like to test drive it first 😉

AndyScherzinger:
Had to do another rebase and am willing to keep it up to date with the app's master. Hope that is fine with you @stalker314314, please let me know if not. Just trying to be of help :)

AndyScherzinger:
👍 tested successfully!

Great job @stalker314314 ! 🚀

PR #1542, opened 2017-09-13
Add proper error message to user info screen

AndyScherzinger:
resolves #1540 adding a proper error message

before:
screen shot 2017-09-05 at 10 04 15

after:
device-2017-09-13-183054

cc: @mario @tobiasKaminsky for review/merge - The detail message can't be changed that easily since thins is what we get from out lib-project and I'd vote for keeping it this way, at least until my error feedback PR has been merged. After that we can go for nicer, mapped http error messages (to some extend).

PR #1543, opened 2017-09-13
Use given empty list icon if it should not be tinted

AndyScherzinger:
Bugfix to bring back the non tinted icons which at the moment would not display but show the blue folder icon simply because that is set by default in the layout itself. This fixes empty lists that show non-tinted icons like the fav-star, search-icon etc. basically any icon that is not a folder icon (which is tinted and will override the default icon):

device-2017-09-13-192410

cc @mario @tobiasKaminsky


mario:
👍

tobiasKaminsky:
:+1:

tobiasKaminsky:
Shall we wait for approval by @jancborchardt, like in #1542

AndyScherzinger:
Imho no, since stated in #1542 this is 1:1 how it is done on the weUI 👍

jancborchardt:
Exactly, looks good :) Only folder icons are tinted to the theme color.

PR #1559, opened 2017-09-19
Support of downgrading database schema

tobiasKaminsky:
Downgrade from 25->24 is not really necessary.

### DIFF
--- a/src/main/java/com/owncloud/android/providers/FileContentProvider.java
+++ b/src/main/java/com/owncloud/android/providers/FileContentProvider.java
@@ -85,7 +85,6 @@ public class FileContentProvider extends ContentProvider {
     private static final String TEXT = " TEXT, ";
     private static final String ALTER_TABLE = "ALTER TABLE ";
     private static final String ADD_COLUMN = " ADD COLUMN ";
-    private static final String REMOVE_COLUMN = " REMOVE COLUMN ";
     private static final String UPGRADE_VERSION_MSG = "OUT of the ADD in onUpgrade; oldVersion == %d, newVersion == %d";
 
     @Override
@@ -1159,6 +1158,7 @@ public class FileContentProvider extends ContentProvider {
         public void onDowngrade(SQLiteDatabase db, int oldVersion, int newVersion) {
             if (oldVersion == 25 && newVersion == 24) {
                 // nothing needs to be done as the upgrade was adding columns only if they did not exist
+                Log_OC.i(TAG, "Downgrading v" + oldVersion + " to " + newVersion);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -1152,6 +1153,13 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
                 }
             }
         }
+
+        @Override
+        public void onDowngrade(SQLiteDatabase db, int oldVersion, int newVersion) {
+            if (oldVersion == 25 && newVersion == 24) {
+                // nothing needs to be done as the upgrade was adding columns only if they did not exist
```
mario: Well if we want to be precise, then we could remove the columns xD

AndyScherzinger:
@tobiasKaminsky looking at the code we basically "do nothing" with this code change except adding an empty method, right?

src/main/java/com/owncloud/android/providers/FileContentProvider.java
tobiasKaminsky: Schema 25 only re-added the columns if they were missing. This is mainly a fix for some, like Jason.
The columns are needed, e.g. in upgrade to 25 CAPABILITIES_SERVER_SLOGAN is added if it is missing, and this is also added on a fresh installation at https://github.com/nextcloud/android/blob/8272bfab161000b0f55cd8143436dd90dd5f7450/src/main/java/com/owncloud/android/providers/FileContentProvider.java#L1254-L1254

tobiasKaminsky:
@AndyScherzinger from "our" side this is true. But with calling this function the internal scheme db version is set to 24 after the function runs without an exception.
And if we later add new columns/tables we can then easily add them in the same way for downgrade.

AndyScherzinger:
:+1: ah, fine!
But shouldn't we then put in a log statement that we downgraded to 24 and remove the unused constant?

src/main/java/com/owncloud/android/providers/FileContentProvider.java
AndyScherzinger: master is on 24 so 25 is already used for e2e (or whatever might be merged before that...)

AndyScherzinger:
pinging @mario for review :)

AndyScherzinger:
rebased ❗️

tobiasKaminsky:
@mario please review, after that I can do a rebase and merge it.

AndyScherzinger:
pinging @mario for review :) This one important (spams the gplay console with crashes like crazy due to e2e users who moved to 2.0.0 stable....)

mario:
👍

Issue #1564, opened 2017-09-20
Support image url in photos in vCards


tobiasKaminsky:
From wikipedia:
```
BEGIN:VCARD
VERSION:2.1
N:Mustermann;Erika;;Dr.;
FN:Dr. Erika Mustermann
ORG:Wikimedia
ROLE:Kommunikation
TITLE:Redaktion & Gestaltung
PHOTO;JPEG:http://commons.wikimedia.org/wiki/File:Erika_Mustermann_2010.jpg
TEL;WORK;VOICE:(0221) 9999123
TEL;HOME;VOICE:(0221) 1234567
ADR;HOME:;;Heidestrasse 17;Koeln;;51147;Deutschland
EMAIL;PREF;INTERNET:erika@mustermann.de
REV:20140301T221110Z
END:VCARD
```

This is already correct parsed by vcard library: either data is filled or url is set.
So the url must be fetched and stored.

PR #1565, opened 2017-09-20
Fix bug if photo is URL

tobiasKaminsky:
Currently if photo is not a blob, but an URL a NPE results.
This prevents NPE, but shows currently "only" the default icon.
https://github.com/nextcloud/android/issues/1564 is an enhancement to also fetch & store these images.


mario:
👍

AndyScherzinger:
👍

Issue #1481, opened 2017-09-01
Integration with office apps


memen45:
### Actual behaviour
Office files are opened in read-only mode by the official Microsoft Office Android apps (Excel etc.)

### Expected behaviour
Office files are opened in read/write mode with the possibility to save changes directly to the server again
 
### Steps to reproduce
1. In any of the MS Office android apps choose 'Open' and then 'Other Cloud Storage'. 
2. From there select Nextcloud and then browse to any office file in your nextcloud account.
3. Now the file is being downloaded and opens in the Office app - in Read-Only mode.

### Environment data
Android version: 6.0

Device model: H60-L04

Stock or customized system: EMUI 4.0

Nextcloud app version: 1.4.3

Nextcloud server version:12.02

I am not sure whether this is a server issue or an issue in the nextcloud app. The latter could be because the app is first downloading the file instead of providing the office app with a 'cloud version' of the file (i.e. webdav link or something).

memen45:
After searching a little more for a possible cause I suspect it has to do with the webdav server. Since the webdav server used by nextcloud does not support locking of files, it will open in read-only mode. Can someone endorse this? Because in that case I should probably post this to the nextcloud/server git instead of this one. Thanks in advance

jasonbayton:
Can't state with authority, but I'd propose opening it on nc/server too just in case.

AndyScherzinger:
The issue with the Android client is that the app ships with a document provider which only implements ```read``` at the moment which is why it is opened as "read-only".
So the document provider would need to be extended to also support the write operation and then this should be possible.

jasonbayton:
@AndyScherzinger 2.1? 👍 😆

AndyScherzinger:
@jasonbayton rather not 😆 while this needs to handle some more infos... because we wouldn't get a signal when the file has been written, so no auto-magic uploads...

memen45:
You mean that modifications to the files would be directly uploaded to the server instead of through the nextcloud app?

AndyScherzinger:
no, I mean they wouldn't be uploaded since the app might not "know" of the changed file. It would simply be saved on the device by the other app but the Nextcloud app might not be aware of the file been written and thus wouldn't do anything.
This is jsut a "maybe" since we have no experience with writing files via the so called document provider.

memen45:
[...]

PR #1571, opened 2017-09-21
Folder-Based Sort Order

### LINKED ISSUE #1441: separate sort order for each folder

### LINKED ISSUE #1508: Using Arbitrary Data Provider for Layout and Sort Order
Based on #1484

########

> How do you clean up preference entries once say a folder dissappears?:)

Well, first of all, I would evaluate if that's really an issue.
If it's really an issue, I then would ask the person who wrote isGridViewPreferred, what his evaluation was, and how he intended to solve it if at all. :)

If I had to solve it, I tend to use a database having foreign keys and constraints available. So, in case a folder gets removed, we only need to remove all sort-entries for all users of that particular folder.
Folder ids also have the benefit of being path-unrelated. So, less hassle for a user, if he moves some dirs.

My preliminary evaluation tends to "it's not an issue". I cannot imagine that those few bytes are a problem at all. Take this with a grain of salt, as I am new to the project, I might mis-estimate here.

########

Talking practically, it might not be an issue. Talking good design, it is an issue.

As for Arbitrary Data ... yes, it was one of the ideas behind it. You'd basically have "sort_13245" (as an example, where 13245 is folder ID) as a key, and value as the actual sort preference.

I would not mind moving layout preferences to Arbitrary Data as well ... "layout_12345" and then "grid"/"list" as value. Wishful thinking? :)

@mario  @tobiasKaminsky

srkunze:
Resolves #1508
Resolves #1441

### DIFF
--- a/src/main/java/com/owncloud/android/db/PreferenceManager.java
+++ b/src/main/java/com/owncloud/android/db/PreferenceManager.java
@@ -25,7 +25,7 @@ import android.content.SharedPreferences;
 import com.owncloud.android.datamodel.FileDataStorageManager;
 import com.owncloud.android.datamodel.OCFile;
 import com.owncloud.android.ui.activity.ComponentsGetter;
-import com.owncloud.android.utils.FileStorageUtils;
+import com.owncloud.android.utils.FileSortOrder;
 
 /**
  * Helper to simplify reading of Preferences all around the app
@@ -36,8 +36,7 @@ public abstract class PreferenceManager {
      * Value handled by the app without direct access in the UI.
      */
     private static final String AUTO_PREF__LAST_UPLOAD_PATH = "last_upload_path";
-    private static final String AUTO_PREF__SORT_ORDER = "sort_order";
-    private static final String AUTO_PREF__SORT_ASCENDING = "sort_ascending";
+    private static final String AUTO_PREF__SORT_ORDER_NAME = "sort_order_name";
     private static final String AUTO_PREF__UPLOAD_FILE_EXTENSION_MAP_URL = "prefs_upload_file_extension_map_url";
     private static final String AUTO_PREF__UPLOAD_FILE_EXTENSION_URL = "prefs_upload_file_extension_url";
     private static final String AUTO_PREF__UPLOADER_BEHAVIOR = "prefs_uploader_behaviour";
@@ -199,40 +198,21 @@ public abstract class PreferenceManager {
      * Gets the sort order which the user has set last.
      *
      * @param context Caller {@link Context}, used to access to shared preferences manager.
-     * @return sort order     the sort order, default is {@link FileStorageUtils#SORT_NAME} (sort by name)
+     * @return sort order     the sort order, default is {@link FileSortOrder#sort_a_to_z} (sort by name)
      */
-    public static int getSortOrder(Context context) {
-        return getDefaultSharedPreferences(context).getInt(AUTO_PREF__SORT_ORDER, FileStorageUtils.SORT_NAME);
+    public static FileSortOrder getSortOrder(Context context) {
+        String name = getDefaultSharedPreferences(context).getString(AUTO_PREF__SORT_ORDER_NAME, FileSortOrder.sort_a_to_z.mName);
+        return FileSortOrder.sortOrders.get(name);
     }
 
     /**
      * Save the sort order which the user has set last.
      *
      * @param context Caller {@link Context}, used to access to shared preferences manager.
-     * @param order   the sort order
+     * @param sortOrder   the sort order
      */
-    public static void setSortOrder(Context context, int order) {
-        saveIntPreference(context, AUTO_PREF__SORT_ORDER, order);
-    }
-
-    /**
-     * Gets the ascending order flag which the user has set last.
-     *
-     * @param context Caller {@link Context}, used to access to shared preferences manager.
-     * @return ascending order     the ascending order, default is true
-     */
-    public static boolean getSortAscending(Context context) {
-        return getDefaultSharedPreferences(context).getBoolean(AUTO_PREF__SORT_ASCENDING, true);
-    }
-
-    /**
-     * Saves the ascending order flag which the user has set last.
-     *
-     * @param context   Caller {@link Context}, used to access to shared preferences manager.
-     * @param ascending flag if sorting is ascending or descending
-     */
-    public static void setSortAscending(Context context, boolean ascending) {
-        saveBooleanPreference(context, AUTO_PREF__SORT_ASCENDING, ascending);
+    public static void setSortOrder(Context context, FileSortOrder sortOrder) {
+        saveStringPreference(context, AUTO_PREF__SORT_ORDER_NAME, sortOrder.mName);
     }
 
     /**
--- a/src/main/java/com/owncloud/android/db/PreferenceManager.java
+++ b/src/main/java/com/owncloud/android/db/PreferenceManager.java
@@ -195,24 +195,23 @@ public abstract class PreferenceManager {
     }
 
     /**
-     * Gets the sort order which the user has set last.
+     * Get preferred folder sort order.
      *
      * @param context Caller {@link Context}, used to access to shared preferences manager.
      * @return sort order     the sort order, default is {@link FileSortOrder#sort_a_to_z} (sort by name)
      */
-    public static FileSortOrder getSortOrder(Context context) {
-        String name = getDefaultSharedPreferences(context).getString(AUTO_PREF__SORT_ORDER_NAME, FileSortOrder.sort_a_to_z.mName);
-        return FileSortOrder.sortOrders.get(name);
+    public static FileSortOrder getSortOrder(Context context, OCFile folder) {
+        return FileSortOrder.sortOrders.get(getFolderPreference(context, AUTO_PREF__SORT_ORDER_NAME, folder, FileSortOrder.sort_a_to_z.mName));
     }
 
     /**
-     * Save the sort order which the user has set last.
+     * Set preferred folder sort order.
      *
      * @param context Caller {@link Context}, used to access to shared preferences manager.
      * @param sortOrder   the sort order
      */
-    public static void setSortOrder(Context context, FileSortOrder sortOrder) {
-        saveStringPreference(context, AUTO_PREF__SORT_ORDER_NAME, sortOrder.mName);
+    public static void setSortOrder(Context context, OCFile folder, FileSortOrder sortOrder) {
+        setFolderPreference(context, AUTO_PREF__SORT_ORDER_NAME, folder, sortOrder.mName);
     }
 
     /**
--- a/src/main/java/com/owncloud/android/db/PreferenceManager.java
+++ b/src/main/java/com/owncloud/android/db/PreferenceManager.java
@@ -19,13 +19,18 @@
 
 package com.owncloud.android.db;
 
+import android.accounts.Account;
 import android.content.Context;
 import android.content.SharedPreferences;
 
+import com.owncloud.android.authentication.AccountUtils;
+import com.owncloud.android.datamodel.ArbitraryDataProvider;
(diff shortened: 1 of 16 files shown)


srkunze:
based on #1484

mario:
Does this requires merging your initial PR first, or not?

srkunze:
Have a look at the commits. Those are ones of #1484. So, even if you only merge this one, it will implicitly merge #1481.

src/main/java/com/owncloud/android/db/PreferenceManager.java
```diff
@@ -161,43 +178,86 @@ public static void setLastUploadPath(Context context, String path) {
...
-     * @param ascending flag if sorting is ascending or descending
+     * @param context Context object.
+     * @param preferenceName Name of the preference to lookup.
+     * @param folder Folder.
+     * @param defaultValue Fallback value in case no ancestor is set.
+     * @return Preference value
      */
-    public static void setSortAscending(Context context, boolean ascending) {
-        saveBooleanPreference(context, AUTO_PREF__SORT_ASCENDING, ascending);
+    public static String getFolderPreference(Context context, String preferenceName, OCFile folder, String defaultValue) {
```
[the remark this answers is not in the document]
AndyScherzinger: As discussed, can be ignored 👍

src/main/java/com/owncloud/android/db/PreferenceManager.java
```diff
@@ -161,43 +178,86 @@ public static void setLastUploadPath(Context context, String path) {
...
+     */
+    public static void setFolderPreference(Context context, String preferenceName, OCFile folder, String value) {
+        Account account = AccountUtils.getCurrentOwnCloudAccount(context);
+        ArbitraryDataProvider dataProvider = new ArbitraryDataProvider(context.getContentResolver());
+        dataProvider.storeOrUpdateKeyValue(account.name, getKeyFromFolder(preferenceName, folder), value);
+    }
+
+    private static String getKeyFromFolder(String preferenceName, OCFile folder) {
+        final String folderIdString = String.valueOf(folder != null ? folder.getFileId() : FileDataStorageManager.ROOT_PARENT_ID);
+        return preferenceName + "__" + folderIdString;
```
mario: My only question is why "__" instead of "_" like is used for AutoUpload and some other things? :)

srkunze: preferenceName could contain _. So, I remember using __ as a good convention to circumvent similar ambiguities since writing own mini-serializer would be overkill. Maybe, using __ is even overkill. :)

If you prefer _, I can change it.

Last thought, maybe a folder-aware arbitrary storage provider would be best, especially because you mentioned AutoUpload as a third use-case and to avoid future over-engineering.

srkunze:
Closes #1441.

src/main/java/com/owncloud/android/db/PreferenceManager.java
mario: Well AutoUpload already uses arbitrary storage provider. Let's get this merged and then we can get a separate folder-aware arbitrary storage provider.

Can you clarify when could preference name contain _ ? (maybe I'm missing the obvious)

srkunze: image

folder_sort_name and folder_layout

mario: Ah, got it. Well, since we don't split by first "_", but we always know where to split, I'd suggest simple "_" instead of "__". Then I approve this awesome PR! :)

srkunze: Simple _?

mario: Yes, sorry :) preferenceName + "_" + folderIdString

srkunze:
Finished!

mario:
👍

mario:
Thanks for the awesome contribution!

AndyScherzinger:
Hi @srkunze great PR! :)
I did some testing and found some issues during a short test:

1. standard (remote) file listing - works fine for me folder-based
2. local file list (e.g. when choosing files to upload in-app) -  I chose "newest first" and it no sorts all local folders "newest first"
3. share with Nextcloud (from another app) - completely unknown sort order of the remote root folder

So to me it seems that if I change the sorting within the local file list (upload manually via fab) changes the "standard" sorting for all listings except the one with a custom-set-sorting while I would expect the app to obey the "Standard" listing which to me would be alphabetically A-Z.

What do you think @srkunze @mario @jancborchardt

srkunze:
@AndyScherzinger 
1. sounds good! ;)
2. what's the folder id for them? I currently use 0 (aka I don't know)
3. err, did it work before?

AndyScherzinger:
@tobiasKaminsky can you help out with the questions? Regarding point 2. and 3. ?

tobiasKaminsky:
2. Local folder do not have an id, as far as I know, so we then need to use the path I guess.
3. I changed one folder to "newest first" and when sharing to this folder the same order was applied.

mario:
What are we gonna do here? I want this merged! :)

srkunze:
Me, too. :)

2) Will our mini-serialization language (and the reason why @mario wanted this - cleanup jobs, etc.) still work with paths?

3) Can somebody point me to the lines I changed which broke the behavior? I tried to be super-explicit about the changes, so I suspect calls to the  PreferenceManager.getSortOrder(..., null) where I passed null to the folder argument. This was because I couldn't find an equivalent folder object to pass.

So, what are the "locations" / paths / objects determining the sort order on those cases? :)

(null represents FileDataStorageManager.ROOT_PARENT_ID)

mario:
Maybe just dont store sort preferences in this case?

mario:
cc @srkunze  ^_^

srkunze:
@mario It seems to me that this wouldn't be possible. First step is writing the sort-order, second step is reading it in order to do the actual sort. So, no store, no correct sort. Or did I misunderstand your idea?

mario:
Exactly. If there's no sort stored, use whatever you decide is default :)

srkunze:
"Unfortunately", people CAN decide which sort order they want to use in Upload and similar non-nextcloud places. So, they will be irritated if they choose a non-default sort-order and nothing happens.

Does this makes sense? :)

AndyScherzinger:
@tobiasKaminsky do you have a suggestion on this, just asking since I am a bit lost on how to solve this best... :/

AndyScherzinger:
@srkunze I took the liberty to rebase your branch to latest Nc master since it was 539 commits behind nextcloud:master so now it is up-to-date with today's master 🚀

tobiasKaminsky:
I am a bit lost, but I hope I understood it correctly and this is a valid idea:
- store sort order on nc internal folders based on id: working
- handle all other views (upload, share file to nc app, ...) independent and store the last used sort order for them
This way it is not folder based for these (second) views, but it is at least "user-scenario" based.

Otherwise I doubt that this is feasible in terms of handling it internally and also from UX, as those screens are maybe not that often used and then having a changing sort order on maybe each folder is confusing.

AndyScherzinger:
👍 Sounds good to me @tobiasKaminsky, for local filesystem file listing storing the sort order per screen(as in per scenario like upload, share etc.) sounds right to me.

@srkunze do you have the time and would be willing to implement this idea?

srkunze:
@AndyScherzinger Will do.

@mario What serialization format do you propose for upload/share? And where do I get the "user-scenario" information from?

Just to get it right the first time. :)

tobiasKaminsky:
> And where do I get that "user-scenario" information from?

There are only two:
- upload a file from within the app, which then shows the local file system
- share a file from another app to nc app, which then shows the nc internal file structure
--> ❗️ then we could also use there the same sort order as in regular view, or?

srkunze:
@tobiasKaminsky Sure

srkunze:
>>> then we could also use there the same sort order as in regular view, or?
@tobiasKaminsky Sure, if you can point me to the variable/attribute where I can find the folder id there :) I didn't find it back then so I used "null" as a default representing the "UNKNOWN" folder.

tobiasKaminsky:
If you share file from e.g. gallery to our app ReceiveExternalFilesActivity is used.

On start mFile is here: https://github.com/nextcloud/android/blob/6e480449f4e1c1557be5fb4af652a330bedb2c95/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java#L164-L164
It can happen that there is no mFile

And while navigating throught subfolders:
https://github.com/nextcloud/android/blob/6e480449f4e1c1557be5fb4af652a330bedb2c95/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java#L740-L740

Both are OCFile so the should have an internal id.

mario:
@srkunze sorry to bother, but can I help with anything?

tobiasKaminsky:
> There are only two:
> - upload a file from within the app, which then shows the local file system
> - share a file from another app to nc app, which then shows the nc internal file structure
--> ❗️ then we could also use there the same sort order as in regular view, or?

I rebased the PR.
First one is already working and for second I added the code.

mario:
@tobiasKaminsky is this ready now?

srkunze:
@tobiasKaminsky Wow. Thanks for taking this over. Time was short lately. I saw you were able to replace the 'null' with the OCFile. Great. :)

While reading, I catched a typo: SORTING_ORDER_FRAGRMENT.

There's however something missing from what we specified above:

@AndyScherzinger wrote

> Sounds good to me @tobiasKaminsky, for local filesystem file listing storing the sort order per screen(as in per scenario like upload, share etc.) sounds right to me.

Right now, get/setSortOrder does not recognize the scenario.

@mario This was my question about how we intend to serialize the user-scenario information.

mario:
You can add "scenario" key in the value. Say "files", "uploads", "share"...

Would that work?

srkunze:
Isn't part of the key?

mario:
Ah you're right. Do it :D

tobiasKaminsky:
> While reading, I catched a typo: SORTING_ORDER_FRAGRMENT.

Yes, there was for a short time this type, but I fixed and rebased it.

Regarding the scenarios. Maybe I am missing something, but 
- regular cloud file list: use folder id --> working
- upload: show local file list: remember sorting order for view/scenario --> working
- share a file from e.g. gallery to app: use folder id --> working

What else is missing / not working? Please with a step by step how to as I am confused right now ;-)

tobiasKaminsky:
@srkunze @mario please answer my latest questions so I can finish this 👍

mario:
@srkunze sorry, can you confirm to @tobiasKaminsky ? Bit busy on Talk release this week :(

srkunze:
@tobiasKaminsky Alright. From what I see we had a misunderstanding here. Quoting you here:

image

We thought that "user-scenario" means something different. So, nevermind. Let's see how this works in practices. If it doesn't fit our needs we can choose to tweak it again. :+1: 

Nice work!

mario:
@tobiasKaminsky I guess this is ready then? 👍

tobiasKaminsky:
Yes, tested and works for me.

mario:
@srkunze thanks for the awesome awesome contribution and sorry this took so long. You are awesome in receiving feedback and acting based on it, and especially patient with us (mainly me :P) requesting change after change :)

tobiasKaminsky:

tobiasKaminsky:
@srkunze Thank you very much for this very great enhancement 👍 🎉

srkunze:
@tobiasKaminsky Thanks for filling the missing gaps! :+1:

PR #1578, opened 2017-09-22
Fix 4g retry

mario:
Will now retry sooner than 15 minutes in some cases if you go from no connection to 4g.


mario:
@AndyScherzinger @tobiasKaminsky would be cool to get this in RC7.

tobiasKaminsky:
I understand that this is only for restarting jobs. Each job will then verify if an unmetered connection is available and will only upload it then, so
:+1:

tobiasKaminsky:
Rebased, waiting for checks and then I'll merge it.
(silently assuming that @AndyScherzinger is fine with it)

AndyScherzinger:
👍 I am totally fine with it,just sick slightly sick so don't wait for my replies this next week

PR #1581, opened 2017-09-23
suppress annoying "warning: Ignoring InnerClasses"

tobiasKaminsky:
This is due to FreeMarker, and as we cannot change this lib, we can suppress the warning, I guess.

```
warning: Ignoring InnerClasses attribute for an anonymous inner class
(freemarker.ext.beans.MapModel$1) that doesn't come with an
associated EnclosingMethod attribute. This class was probably produced by a
compiler that did not target the modern .class file format. The recommended
solution is to recompile the class from source, using an up-to-date compiler
and without specifying any "-target" type options. The consequence of ignoring
this warning is that reflective operations on this class will incorrectly
indicate that it is *not* an inner class.
```


AndyScherzinger:
👍

proguard.cfg
```diff
@@ -34,3 +34,5 @@
 -keep class * implements android.os.Parcelable {
   public static final android.os.Parcelable$Creator *;
 }
+
+-keepattributes InnerClasses
```
mario: New line, and I will approve :P

tobiasKaminsky:
@mario new line is now in place

mario:
👍

Issue #23, opened 2016-06-14
GPLv2 vs Apache2 license


krt16s:
This app is GPLv2 licensed but uses Apache2 style dependencies like Google's Android support libraries.  This might be problematic, e.g. see https://www.apache.org/licenses/GPL-compatibility.html . So iirc switching to GPLv2+ ( adding the "version 2 and later"option) or GPLv3 solves this, but I am not a lawyer...

AndyScherzinger:
I am no lawyer either and not an OSS license specialist, so @fkarlitschek or @jancborchardt or @LukasReschke can you answer this?

LukasReschke:
Summoning @schiessle as well on this. Go GNU/Björn 😉

jancborchardt:
Any info on this @schiessle @karlitschek? If this is a problem, then it’s an issue with the original code too and the relicense to GPLv2 from the previous GPLv2+ might have been not proper.

karlitschek:
I think Apache dependencies are not a problem. But @schiessle is the expert here!

schiessle:
Yes, the Apache license and GPLv2 are not compatible because of the patent grant in the Apache license.

I'm not a Android expert, if the "Google's Android support libraries" are system libraries this shouldn't be a problem, because of the "system library exception" https://www.gnu.org/licenses/gpl-faq.html#SystemLibraryException

schiessle:
We can't change the license from GPLv2 only to GPLv2+ because we don't own the complete source code. But I would suggest to consider switching to GPLv2+ for every new code added to the android app. We do something similiar on the server (ownCloud code is AGPLv3 only but we add new code as AGPLv3+)

tobiasKaminsky:
We are doing it the @schiessle proposed: every new code done by us is GPL3+

PR #1589, opened 2017-09-25
Database upgrade #23 skipped entire remaining upgrades

tobiasKaminsky:
This is the real reason for the upgrade problems on some people and the reason we needed upgrade # 24.
Diff looks worser than it is: one brace was set wrong, the rest is indentation.

--> I guess RC8 is required


AndyScherzinger:
👍 nice catch!

mario:
o.O

👍

PR #1604, opened 2017-09-28
Fixes #1600

### LINKED ISSUE #1600: Translation: 2 different sentences needed
### Actual behaviour
Same key / label is used in 2 different contexts.

### Expected behaviour
To make the app more easy to use:
- need 1 translation in contextual menu => label for an action with a verb like other entries of this menu
- need 1 translation in file properties => label meaning a status (on/off)

See the 2 contexts in screen captures attached with currently the same label in french (confirmed: only one label available in transifex => "**Disponible hors-ligne**"

label for action
label for status

stalker314314:
Change "Set as available offline" text to "Available offline" for switch widget in file details dialog, to differnetiate between this and "Set as available offline" in menu

### DIFF
--- a/src/main/res/layout/file_details_fragment.xml
+++ b/src/main/res/layout/file_details_fragment.xml
@@ -166,7 +166,7 @@
 				android:layout_width="match_parent"
 				android:layout_height="wrap_content"
 				android:layout_gravity="start"
-				android:text="@string/favorite"
+				android:text="@string/favorite_switch"
 				android:textSize="16sp"
 				android:layout_alignParentTop="true"
 				android:layout_alignParentLeft="true"
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -270,6 +270,7 @@
     <string name="unfavorite">Unset as available offline</string>
     <string name="favorite_real">Set as favorite</string>
     <string name="unset_favorite_real">Unset favorite</string>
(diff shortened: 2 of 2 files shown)


mario:
👍 

Can go into 2.0.0 afaik.

tobiasKaminsky:
Thank you for this PR.
Please note that only the english values/strings.xml is used in github. Every other language is translated on transifex, so changes here on github will be overriden.
Can you remove the other languages and rebase? Then drone should also be satisfied.
If you do not have time for this, then I can handle this.

stalker314314:
I used only values/strings.xml, I think, other files came from merge later (so, lint was failing before merge with latest). Is this what you were thinking?

tobiasKaminsky:
👍

AndyScherzinger:
@tobiasKaminsky I'd say let's merge it. I checked the lint results and this PR adds 1 "plurals candidate" warning while lint simply found one more warning on one line which is exactly the same in both branches. So I think to get it back to green we should bump warnings to 511 for now. My PR for plurals will fix all plurals warnings once it gets merged so I think we can live with this temporary bump.

AndyScherzinger:
NOTE: I am fixing the warning count on master atm, will then rebase this PR and merge

AndyScherzinger:
Thaks for the PR @stalker314314 🥇 
Merged to master 🎉

Issue #579, opened 2017-01-24
File actions in the list


jancborchardt:
Currently it’s very cumbersome to do any action on a file. You have to go through long-press, then the multiselect menu, and then choose the action.

Instead we should do it obviously, like in the mobile web interface:
![capture du 2017-01-24 12-06-42](https://cloud.githubusercontent.com/assets/925062/22244940/d4a10924-e22d-11e6-958d-4be6e8d91043.png)
- A 3-dot-menu for every folder/file. Very obviously clickable
- The share icon should also be clickable, because people expect that. Playing around with making it appear un-clickable is just beating around the problem. ;) If there’s too little space then we could even only show the share icon when a file _has_ been shared (like we do now) but when tapped it should be interactive and open the share options.
- If there’s enough space, even have a column or some more space for the favorite icon on the left.

Already talked about this with @mario, what do you think @tobiasKaminsky @eppfel @AndyScherzinger?

AndyScherzinger:
Sounds good to me and is doable.
Only thing I am not sure if it is better to 
1. always show a share icon (in grey) and paint it black if the file is actually shared
2. only add the share icon if the file is actually shared, else don't

jancborchardt:
@AndyScherzinger cool! :) As said I’m also not entirely sure, but then maybe let’s go with 

> always show a share icon (in grey) and paint it black if the file is actually shared

This is consistent with the web interface, makes it easy to share, and doesn’t require us to change the menu.

> if we put the "share action" in there then I guess it should always be shown with the share action wired to it :)

Yep, that’s the idea! :)

mario:
Glad we all agree :P

tobiasKaminsky:
Just a "little" concern about space: have you checked it on low end devices, which have a low-res display? Your idea might lead to a nearly unreadable filename...
This is something we should check with a "mockup" prior developing it.

jancborchardt:
Yes, I mean you can quite easily check with the web interface and checking it in responsive design view of the browser or making it very narrow. The mobile web interface is the living mockup :)

If there are space constraints then I’d say we can do without the dedicated favorite and share action, but we definitely need the 3-dot-menu. In there we can then have the fav/share also. But it’s not ideal.

And I guess somehow we can check how big the screen is, and show the actions accordingly even?

AndyScherzinger:
As for the three-dot menu, should we then have the (non-Android-like) schaschlik 3-dot menu icon or the typical Dönerspieß 3-dot icon?
[...]

PR #1613, opened 2017-10-02
Add options for operations on single files

rhari991:
Related issue: #579 

TODO
- [x] Make proper icons
- [x] Make the layout responsive


tobiasKaminsky:
Looks & works great for a very first version.
Please try to stay in close contact with @jancborchardt. He is our designer and it is always best to show him the first versions, so he can give feedback as soon as possible.

jancborchardt:
Awesome @rhari991! :) Some feedback from the first look:

- The share and menu icons don't seem to be vertically aligned correctly.
- The icons should probably be aligned exactly with the menu and search icon in the top bar, right?
- As the menu is very long, let's remove the Select/Deselect all from here already and _only_ show it in the multiselect menu.
- Can we highlight the row of the file to which the menu corresponds? Or is there some Material Design standard? :)

tobiasKaminsky:
Here are the specs for lists: https://material.io/guidelines/components/lists.html#lists-specs
I remember that we discussed in another issue that the thumbnails are a little bit to low (not height enough).

AndyScherzinger:
The Thumbnails will be fixed by my PR, tbd soon :)
The list item which's menu has been clicked can't be highlighted, so you could call that "standard". There would be ways to implement that by hand but it is rather uncommon on Android (never saw any app with list item menu doing this)

AndyScherzinger:
@jancborchardt priview/filetype icon size PR is #1617

AndyScherzinger:
@rhari991 great work!

@jancborchardt @rhari991 can we please not have to fav icon/action in (again)? This is not material like but simply copying the webUI. So keeping it as an indicator on the thumbnail/filetype icon like before and having the action in the menu.

jancborchardt:
> can we please not have to fav icon/action in (again)? This is not material like but simply copying the webUI. So keeping it as an indicator on the thumbnail/filetype icon like before and having the action in the menu.

Absolutely agree – we actually are in the process of moving to what you describe in the web interface too: Only indicator in the top right of the filetype icon, and action in the menu.

rhari991:
New screenshot :
screens

rhari991:
@jancborchardt 

- The share and menu icons should be vertically centred now
- I'm not sure if it's always possible to align the action bar icons with these. The action bar icons are set by the Android framework while the per-item icons are set manually by us. I've put 16dp of space between the overflow menu and the right edge of the screen as per material design guidelines, but this does not seem the align them.
- Removed select and deselect all from the per-item menu

src/main/res/drawable/shared_via_users.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
rhari991: I'm not sure what image I should use for the "already shared" icon. I've added an svg for now, I suppose the right way would be to colour the existing grey icon to black ?

src/main/res/layout/list_item.xml
```diff
@@ -141,36 +138,48 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:layout_gravity="center_vertical"
+            android:paddingLeft="@dimen/standard_padding"
```
rhari991: I've added the dimensions as under "Avatar with text and icon" in "Two line list" from the material design specs

src/main/res/drawable/shared_via_users.xml
AndyScherzinger: That would probably the easiest way by just tinting it black. Else we could ship a grey and a black version. Not sure which one needs lees processing power. For a first iteration of this change and for this PR I'd say we are fine either way 👍

src/main/res/layout/list_item.xml
AndyScherzinger: Fine by me. You don't have to care too much about the left/icon side of the item since that part will be fixed by another PR #1617

AndyScherzinger:
> I'm not sure if it's always possible to align the action bar icons with these. The action bar icons are set by the Android framework while the per-item icons are set manually by us. I've put 16dp of space between the overflow menu and the right edge of the screen as per material design guidelines, but this does not seem the align them.

This is probably due to the white space of the three dot menu icon so I think we can/should play around with the padding/margin to get it aligned.

Great work @rhari991 🎉

src/main/res/drawable/shared_via_users.xml
rhari991: Do you mean I should keep this SVG or add a black version of the existing grey png ?

AndyScherzinger: keeping the SVG is fine by me since the build tools will generate PNGs anyways (due to the support of 4.0+). Only thing that is off is the size of it since the share icon(s) are slightly smaller than 18dp icons, see ```shared_via_users.png``` for example which is 24x24 pixels (hdpi) in comparison to a 18dp icon like ```ic_folder_star_18dp.png``` which is 27x27 pixels. So design wise the black share icon should have the same size as ```shared_via_users.png```. So since mdpi is the base reference the shared icon should be 16dp 😃

rhari991: 16 dp x 16 dp it is !

AndyScherzinger: Great 👍

rhari991:
s1
s2

AndyScherzinger:
@rhari991 looks great! 🎉 I just did a pixel deep dive and the icon needs a slight push to the left to be better aligned, see:

pixels

Sorry for nitpicking 😿

rhari991:
@AndyScherzinger haha, no worries :)
How about this ?

s3

src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -902,10 +934,10 @@ public void onItemClick(AdapterView<?> l, View v, int position, long id) {
      * Start the appropriate action(s) on the currently selected files given menu selected by the user.
      *
      * @param menuId Identifier of the action menu selected by the user
+     * @param checkedFiles List of files selected by the user on which the action should be performed
      * @return 'true' if the menu selection started any action, 'false' otherwise.
      */
-    public boolean onFileActionChosen(int menuId) {
-        final ArrayList<OCFile> checkedFiles = mAdapter.getCheckedItems(getListView());
+    public boolean onFileActionChosen(int menuId, ArrayList<OCFile> checkedFiles) {
```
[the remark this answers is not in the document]
rhari991: @AndyScherzinger should I fix these ?

AndyScherzinger:
@rhari991 PERFECT! 👍 👍 👍

rhari991:
Does not look bad on a 300px x 400px screen either... unless I configured the emulator wrong somehow

s4

AndyScherzinger:
Looks great @rhari991 !

One question, am I able to commit to your branch? Since I'd like to replace your share icon xml with the one from Nextcloud since it is slightly different (see the thickness of the lines in the icon).

rhari991:
I believe I ticked the box saying "Allow collaborators to modify the request" when creating the PR, so you should be able to :)

AndyScherzinger:
@rhari991 you did :) I just replaced the vector with the Nextcloud one and did the same thing for the shared by link icon :)

AndyScherzinger:
Looks good to me too, also tested on 320x480 (mdpi)

device-2017-10-05-161513

AndyScherzinger:
@rhari991 issue I found during testing is that the menu is also shown when you do a manual upload via the FAB where the menu shouldn't be shown:
device-2017-10-05-181853

AndyScherzinger:
@rhari991 fixed via https://github.com/nextcloud/android/pull/1613/commits/e9736be1387d72e5ab1ebad0821212cb0950b65b 🚀

AndyScherzinger:
👍 code review: Done/Approved - testing: Done/no open issues from my side / found

@jancborchardt since it is a visual change I flagged it with design, so it needs a @nextcloud/designers approval. To check: https://github.com/nextcloud/android/pull/1613#issuecomment-334478597 (low end device) and https://github.com/nextcloud/android/pull/1613#issuecomment-334441990 (high end device)

@tobiasKaminsky @mario please review 😃

mario:
👍  Looks decent, so that's two approvals. So only @nextcloud/designers missing :)

jancborchardt:
Nice! Looks good as far as I can see. :) Just one thing: Shouldn’t the select checkbox be overlayed on the filetype icon as with other apps? It’s strange on the right, also cause it moves the share icon out of place.

rhari991:
@jancborchardt Done !

s5

AndyScherzinger:
@jancborchardt now due to your latest demands all selection list views are screwed since you don't have a preview thumbnail anymore, e.g. The manual upload...just navigate to an image folder.... I'd like to not have this latest change done, else to not have it move flip the select box and the share icon for now and postpone the other idea until we can come up with a viable one. :/

AndyScherzinger:
@jancborchardt here's an idea,why don't we just hide the item menu in case multi selection is active? :)

comradekingu:
My idea was to start scrolling the underlying menu, so long as what is on top of it, intervenes with the length of the underlying menu.

1234   Menu over here. → not a problem

1 2 3 4 M 5 E 6 N 7 U  → problematic.

Now if it scrolled in the 1-4 section, then you could see both menus at once.

AndyScherzinger:
@comradekingu which underlying menu do you mean? There is one menu in the blue action bar and then there is a per item menu which depending on the items position will grow to the bottom or the top.

@jancborchardt @rhari991 since the share icon is shown for all items explicitely and opens the share dialog we should imho remove the share menu item from the menu since it is a duplicate.

rhari991:
Removing the share and menu icons when multi-select is taking place makes sense, since the user has indicated that he wants to perform an action on multiple files, and so would not be interested in the single-file options.

And yes, the share option in the per-item menu is probably redundant :)

jancborchardt:
Good point! So:

- In multiselect mode, replace the share- and 3-dot-icon with the checkbox
- Yes, absolutely do not duplicate the "Share" entry in the menu as it is already visible in the row.

(And for the record I meant the checkbox overlaying the icon, _not_ replacing of course. ;)

AndyScherzinger:
> And for the record I meant the checkbox overlaying the icon, not replacing of course.

Ah, my mistake so yes, let's go for the solution we all kind of described:

- In multiselect mode, replace the share- and 3-dot-icon with the checkbox
- Yes, absolutely do not duplicate the "Share" entry in the menu as it is already visible in the row.

This way it'll work just fine for all file list view local and remote 👍

rhari991:
@jancborchardt @AndyScherzinger 

s6

AndyScherzinger:
👍 fine by me, what do you think @jancborchardt

tobiasKaminsky:
Code wise it is fine for me 👍 
Once we want to merge it, it needs to be rebased and code formatting should be done (there are some missing whitespaces), but I can handle this then.

P.S: we will merge hacktoberfest PR within october, but as we are currently bug fixing 2.0 release, we will postpone this merge a little bit.

AndyScherzinger:
Woohoo @rhari991 yet another great contribution! Merged! 🥇

jancborchardt:
@rhari991 great contribution, thank you so much! :tada: :) Looking forward to more

PR #1618, opened 2017-10-02
Rename "remove" to "delete"

### LINKED ISSUE #655: Rename "remove" to "delete"
"Delete" is a standard word to describe this action and I know what it does already. "Remove" is something new that I have to think about what it does. 

I suspect this wording was used so that "remove from server" and "remove from device" can be options in the dialog, but I think "Delete from device" is just as clear. 

As an added benefit, "Delete" is a lot more visually recognizable in the menu compared to "Remove", which is easy to mix up with "Rename".

Tangentially related: #109

AndyScherzinger:
resolves #655

### DIFF
--- a/src/main/res/menu/file_actions_menu.xml
+++ b/src/main/res/menu/file_actions_menu.xml
@@ -71,12 +71,6 @@
         android:showAsAction="never"
         android:orderInCategory="1" />
     <item
-        android:id="@+id/action_remove_file"
-        android:title="@string/common_remove"
-        app:showAsAction="never"
-        android:showAsAction="never"
-        android:orderInCategory="1" />
-    <item
         android:id="@+id/action_select_all_action_menu"
         android:title="@string/select_all"
         app:showAsAction="never"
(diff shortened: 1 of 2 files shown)


jancborchardt:
Looks good! :) @smichel17 please also review as it resolves an issue you reported.

smichel17:
Looks good to me :)

smichel17:
Hmm. On second thought, I don't actually like "Delete" when used with accounts, since it does not delete the account on the server. So, I think I would use "delete" for files and "remove" for accounts.

comradekingu:
@smichel17 Sounds like a "deactivate" to me.

smichel17:
Android uses "Remove account" in the online accounts section of the settings.

comradekingu:
@smichel17 Makes sense if what you are removing is the attachment to the device. "Unlink account" would be an option, but far worse. "Remove" seems preferable if nothing is deleted.

AndyScherzinger:
That is why I preferred to use Remove nearly everywhere since remove is non-destructive while delete is destructive and in nearly all cases on the client it should be remove (except when you delete files on the server). But I stopped taking on these discussions with designer since it is frustrating so I carefully pick my fights that is UI/graphic/behavior-related only...

jancborchardt:
@AndyScherzinger it’s fine by me to use »Remove« for users, as that is more accurate. :) However, files should be Delete as that is the wording we use in the server menu too.

AndyScherzinger:
@jancborchardt @smichel17 @comradekingu fixed. Now remove is used for accounts and delete for files. 👌

tobiasKaminsky:
👍

AndyScherzinger:
pinging @mario for review

mario:
👍

PR #1629, opened 2017-10-05
Cleaned up cipher transformation arguments.

ardevd:
- Define commonly used transformation specs in constants.


AndyScherzinger:
👍 nice catch @ardevd 

@tobiasKaminsky this PR is for you ❤️

tobiasKaminsky:
👍 
Nice one, so everything is at the same place.

PR #1641, opened 2017-10-09
"Quota": If no quota is set, show the total usage/space of the server

### LINKED ISSUE #471: "Quota": If no quota is set, show the total usage/space of the server
If no quota is set the total usage/free space of the server should be shown as the user needs to know if the space is running low.

AndyScherzinger:
resolves #471 by showing the usage of the account

@tobiasKaminsky is this what you intended? Because it doesn't give you the free disc space... I shows used space by that user / total disc space

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -969,7 +969,7 @@ public abstract class DrawerActivity extends ToolbarActivity implements DisplayU
                         runOnUiThread(new Runnable() {
                             @Override
                             public void run() {
-                                if (quotaValue > 0
+                                if ((quotaValue > 0 || quotaValue <= GetRemoteUserInfoOperation.SPACE_UNLIMITED)
                                         || quotaValue == GetRemoteUserInfoOperation.QUOTA_LIMIT_INFO_NOT_AVAILABLE) {
                                     /**
                                      * show quota in case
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
> If no quota is set the total usage/free space of the server should be shown as the user needs to know if the space is running low.

Edge case (as in my server): won't work since it is a shared host so the disc space doesn't matter but the quota of my server user... ;)

tobiasKaminsky:
For me it shows used account size / free disk space.
It is what I had in mind, but now I am unsure...The problem is that the "free disk space" will change rather frequently and without any interaction of the user, e.g. when another user is filling up their space.
So it is a little bit un intuitive, but on the other hand it shows how much space is left, which is also nice to see.

@jancborchardt maybe?

tobiasKaminsky:
Regarding shared hoster: this is of course a problem. But I would have assumed that you can only see your quota and not the whole disc space...

jancborchardt:
On the server in the sidebar we simply say »3.5 GB used« and nothing after it if no upper limit is set. In the personal settings where there’s more space we say »You are using 8.6 MB of Unlimited«.

AndyScherzinger:
> Regarding shared hoster: this is of course a problem. But I would have assumed that you can only see your quota and not the whole disc space...

Need to dissapoint you, my Nc shows the disc space which on my shared hoster is HUUUUUGE! ;)

jancborchardt:
If you have quota, show the quota. If you have "unlimited", show nothing. ;)

tobiasKaminsky:
@jancborchardt the idea was to see how much space at all is left.
Imagine your server has only 100G and all is used and you try to upload a file. This will then result in an error.
With this you could then see "100G of 100G used".

jancborchardt:
Ok - we should align the server and iOS then though. cc @rullzer @marinofaggiana

AndyScherzinger:
> If you have quota, show the quota. If you have "unlimited", show nothing. ;)

This is how it is atm

>  the idea was to see how much space at all is left.

which only works on dedicated servers, shared hosting or VPS might lead to problems...

rullzer:
Showing the free disk space can indeed lead to all kinds of issues. I'd argue it is the job of tge admin to monitor their server for enough space if they set quota to unlimited.

Also what to show when object storage is in use etc.

So I vote to just show nothing if there is no quota set.

AndyScherzinger:
> So I vote to just show nothing if there is no quota set.

That is today's Android client implementation :)

marinofaggiana:
For iOS the quota (bar) derive from API User profile,

result.quota = userProfile.quota
result.quotaFree = userProfile.quotaFree
result.quotaRelative = userProfile.quotaRelative
result.quotaTotal = userProfile.quotaTotal
result.quotaUsed = userProfile.quotaUsed

I apply this values ... ever ...

tobiasKaminsky:
Ok, I am fine to close this and keep it the way we have it now.
@AndyScherzinger your opinion as it is your PR? ;-)

AndyScherzinger:
absolutely fine! :)

Issue #1638, opened 2017-10-09
In detail view (when viewing a picture), also make the background of the lower bar transparent


jancborchardt:
Just like the background of the header with filename etc is half-transparent, the bottom bar (with back, home, switcher) should be half-transparent too. Currently it’s black.

Then when you zoom into images there’s more perceived space for them. :)

PR #1643, opened 2017-10-10
add transparency to navigation bar for image preview

apicellaj:
Adds transparency to the navigation bar when viewing an image.

Fixes: #1638

### DIFF
--- a/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.java
+++ b/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.java
@@ -123,12 +123,6 @@ public class PreviewImageActivity extends FileActivity implements
                         }
                     }
                 });
-
-        // Add transparency to bottom navigation bar where supported
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
-            getWindow().setFlags(WindowManager.LayoutParams.FLAG_LAYOUT_NO_LIMITS,
-                    WindowManager.LayoutParams.FLAG_LAYOUT_NO_LIMITS);
-        }
          
         if (savedInstanceState != null) {
             mRequestWaitingForBinder = savedInstanceState.getBoolean(KEY_WAITING_FOR_BINDER);
--- a/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.java
+++ b/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.java
@@ -26,6 +26,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.ServiceConnection;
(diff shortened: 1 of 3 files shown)


AndyScherzinger:
Hi @apicellaj,

thanks for another Hacktoberfest session and PR 😃 

One small issue is that now the bar is transparent even when the image can be display (jsut use a text file and rename the file extension to .jpg an open it:

device-2017-10-11-004008

In that case the bar should not be transparent 😀 

cc: @jancborchardt for later approval from designers since you opened the issue itself.

apicellaj:
Hi @AndyScherzinger ,

I'm glad to be back! 😃 Thanks for pointing out this issue. I went ahead and pushed the transparency change to the theme which is probably the way I should have done it initially.

AndyScherzinger:
Hi @apicellaj,

happy to have you back ❤️ 
As for the corrupt version, could you paint it black please? :)
device-2017-10-11-115001

jancborchardt:
Thanks a lot @apicellaj! :) Also @AndyScherzinger’s remarks make sense. Good otherwise!

apicellaj:
@AndyScherzinger Is there a particular way you would recommend painting the navigation bar black? Making programmatic changes to the navigation bar seems like the way to go but the ViewPager makes it difficult to paint properly when swiping into the corrupt image. I'm also assuming that we want to keep the background white when an image fails to load.

Should the Action Bar be made black as well?

AndyScherzinger:
Hi @apicellaj,

I’d also say programmatically is the way to go. The action bar for a corrupt image should actually be colored in the primary color,the being retrieved via the theme utility class since it can be dynamically themed via the server. Background for corrupt images should stay white if possible.

jancborchardt:
Maybe the background for corrupt images should also be black (like when viewing and loading too). Otherwise the change of black to white to black is very jarring when swiping through your images.

AndyScherzinger:
Then let's go for a black background which basically solves the navigation bar coloring for corrupt images :)

AndyScherzinger:
And then we also don't need to change anything for the action bar coloring too :)

AndyScherzinger:
👍 fine and tested successfully! 🎉

AndyScherzinger:
pinging @mario for review/approval

mario:
👍 

[Approved with PullApprove](https://pullapprove.com/nextcloud/android/pull-request/1643/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=nextcloud/android)

mario:
@AndyScherzinger merge when Drone is done.

AndyScherzinger:
merged!
Thanks @apicellaj for yet another great hacktoberfest contribution! :)

jancborchardt:
Yep, thanks a lot for the fix @apicellaj! :)

PR #1649, opened 2017-10-10
Always show share icon in file details screen

### LINKED ISSUE #1637: Also display share icon in detail view, left next to 3-dot-icon
As a follow-up to File actions in the list #579, we should also have the share icon directly displayed in the detail view header (where currently only the 3-dot-icon is).

Also cc @rhari991 who is working on Add options for operations on single files #1613

rhari991:
Resolves #1637


AndyScherzinger:
Hi @rhari991 unfortunately the solution is not that easy ;) The menu.xml in use is utilized in several activies and thus the share icon now pops up everywhere, where it should not like, previews (images, video, test files) and multi select menu:
device-2017-10-11-113216
device-2017-10-11-113305
device-2017-10-11-113102
device-2017-10-11-113145

It should only show up here:
device-2017-10-11-113120

rhari991:
@AndyScherzinger whoops, I should have noticed that. 
I have fixed it, now it should work as intended :)

tobiasKaminsky:
:+1: tested and rebased

AndyScherzinger:
Thanks @rhari991 for yet another great contribution❤️ 

🚀 merged for release 2.1.0

PR #1651, opened 2017-10-10
Fix opening files

mario:
Some apps are stupid and won't allow us to open files in write mode. This fixes it.

I think this is important for 2.0, but you decide...

resolves #751

Signed-off-by: Mario Danic <mario@lovelyhq.com>

PR #1655, opened 2017-10-11
Fix syncing

mario:
Depends on https://github.com/nextcloud/android/pull/1651 (merged)

Guarantees that you'll get the newest version of the file every time you take a look at it.

### DIFF
--- a/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
+++ b/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
@@ -23,6 +23,7 @@ package com.owncloud.android.providers;
 import android.accounts.Account;
 import android.annotation.TargetApi;
 import android.content.ContentResolver;
+import android.content.Context;
 import android.content.Intent;
 import android.content.res.AssetFileDescriptor;
 import android.database.Cursor;
@@ -36,6 +37,11 @@ import com.owncloud.android.authentication.AccountUtils;
 import com.owncloud.android.datamodel.FileDataStorageManager;
 import com.owncloud.android.datamodel.OCFile;
 import com.owncloud.android.files.services.FileDownloader;
+import com.owncloud.android.lib.common.operations.RemoteOperationResult;
+import com.owncloud.android.operations.SynchronizeFileOperation;
+import com.owncloud.android.ui.activity.ConflictsResolveActivity;
+import com.owncloud.android.utils.FileStorageUtils;
+
 import org.nextcloud.providers.cursors.FileCursor;
 import org.nextcloud.providers.cursors.RootCursor;
 
@@ -103,12 +109,15 @@ public class DocumentsStorageProvider extends DocumentsProvider {
 
         OCFile file = mCurrentStorageManager.getFileById(docId);
 
+        Account account = mCurrentStorageManager.getAccount();
+        Context context = getContext();
+
         if (!file.isDown()) {
 
             Intent i = new Intent(getContext(), FileDownloader.class);
-            i.putExtra(FileDownloader.EXTRA_ACCOUNT, mCurrentStorageManager.getAccount());
+            i.putExtra(FileDownloader.EXTRA_ACCOUNT, account);
             i.putExtra(FileDownloader.EXTRA_FILE, file);
-            getContext().startService(i);
+            context.startService(i);
 
             do {
                 if (!waitOrGetCancelled(cancellationSignal)) {
@@ -117,6 +126,23 @@ public class DocumentsStorageProvider extends DocumentsProvider {
                 file = mCurrentStorageManager.getFileById(docId);
 
             } while (!file.isDown());
+        } else {
+            FileDataStorageManager storageManager =
+                    new FileDataStorageManager(account, context.getContentResolver());
+            SynchronizeFileOperation sfo =
+                    new SynchronizeFileOperation(file, null, account, true, context);
+            RemoteOperationResult result = sfo.execute(storageManager, context);
+            if (result.getCode() == RemoteOperationResult.ResultCode.SYNC_CONFLICT) {
+                // ISSUE 5: if the user is not running the app (this is a service!),
+                // this can be very intrusive; a notification should be preferred
+                Intent i = new Intent(context, ConflictsResolveActivity.class);
+                i.setFlags(i.getFlags() | Intent.FLAG_ACTIVITY_NEW_TASK);
+                i.putExtra(ConflictsResolveActivity.EXTRA_FILE, file);
+                i.putExtra(ConflictsResolveActivity.EXTRA_ACCOUNT, account);
+                context.startActivity(i);
+            } else {
+                FileStorageUtils.checkIfFileFinishedSaving(file);
+            }
         }
 
         return ParcelFileDescriptor.open(new File(file.getStoragePath()), ParcelFileDescriptor.parseMode(mode));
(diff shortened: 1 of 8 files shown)


mario:
Also, I tested ... it works :D

AndyScherzinger:
I rebased the PR since #1651 is now on master, so this can be tested and reviewed thus pinging @tobiasKaminsky :)

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
```diff
@@ -117,6 +126,23 @@ public ParcelFileDescriptor openDocument(String documentId, String mode, Cancell
...
             } while (!file.isDown());
+        } else {
+            FileDataStorageManager storageManager =
+                    new FileDataStorageManager(account, context.getContentResolver());
+            SynchronizeFileOperation sfo =
+                    new SynchronizeFileOperation(file, null, account, true, context);
+            RemoteOperationResult result = sfo.execute(storageManager, context);
+            if (result.getCode() == RemoteOperationResult.ResultCode.SYNC_CONFLICT) {
+                // ISSUE 5: if the user is not running the app (this is a service!),
+                // this can be very intrusive; a notification should be preferred
```
tobiasKaminsky: Then please show a notification (with what content?)

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1991,29 +2005,46 @@ public void startImagePreview(OCFile file, VirtualFolderType type) {
...
      */
-    public void startMediaPreview(OCFile file, int startPlaybackPosition, boolean autoplay) {
-        Fragment mediaFragment = PreviewMediaFragment.newInstance(file, getAccount(), startPlaybackPosition, autoplay);
-        setSecondFragment(mediaFragment);
-        updateFragmentsVisibility(true);
-        updateActionBarTitleAndHomeButton(file);
-        setFile(file);
+    public void startMediaPreview(OCFile file, int startPlaybackPosition, boolean autoplay, boolean showPreview) {
+        if (showPreview && file.isDown() && !file.isDownloading()) {
+            Fragment mediaFragment = PreviewMediaFragment.newInstance(file, getAccount(), startPlaybackPosition, autoplay);
```
tobiasKaminsky: line wrap

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -179,15 +211,48 @@ public void openFile(OCFile file) {
...
+            Intent finalOpenFileWithIntent = openFileWithIntent;
+            new Thread(new Runnable() {
+                @Override
+                public void run() {
+                    Account account = AccountUtils.getCurrentOwnCloudAccount(mFileActivity);
+                    FileDataStorageManager storageManager =
+                            new FileDataStorageManager(account, mFileActivity.getContentResolver());
+                    // a fresh object is needed; many things could have occurred to the file
+                    // since it was registered to observe again, assuming that local files
+                    // are linked to a remote file AT MOST, SOMETHING TO BE DONE;
```
tobiasKaminsky: So there is something needed to develop?

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -179,15 +211,48 @@ public void openFile(OCFile file) {
...
+                    SynchronizeFileOperation sfo =
+                            new SynchronizeFileOperation(file, null, account, true, mFileActivity);
+                    RemoteOperationResult result = sfo.execute(storageManager, mFileActivity);
+                    mFileActivity.dismissLoadingDialog();
+                    if (result.getCode() == RemoteOperationResult.ResultCode.SYNC_CONFLICT) {
+                        // ISSUE 5: if the user is not running the app (this is a service!),
```
tobiasKaminsky: Is this really a service here?

src/main/java/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -591,4 +592,20 @@ public static void deleteRecursively(File file, FileDataStorageManager storageMa
...
+        long lastModified = 0;
+        long lastSize = 0;
+        File realFile = new File(file.getStoragePath());
+        while ((realFile.lastModified() != lastModified) && (realFile.length() != lastSize)) {
+            lastModified = realFile.lastModified();
+            lastSize = realFile.length();
+            try {
+                Thread.sleep(1000);
+            } catch (InterruptedException e) {
+                Log.d(TAG, "Failed to sleep for a bit");
```
tobiasKaminsky: Log_OC.e as it is an error

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
mario: Where would you show the notification when our app doesn't even have to be running? Also, this is what sync adapter currently does too, and until we have a better way to show conflicts, I think this is fine.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
mario: Please fix.

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
mario: It's not, the explanation was c/p, but the issue stays: you could navigate away from the activity etc etc. It's a whole other can of worms.

src/main/java/com/owncloud/android/utils/FileStorageUtils.java
mario: Please fix.

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
tobiasKaminsky: I do not know, but it is stated in the comment:
> this can be very intrusive; a notification should be preferred

If this is not correct/possible then the comment should be adjusted.

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
tobiasKaminsky: This is still unclear for me...

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
tobiasKaminsky: Then please adapt the comment, maybe with a todo for later adjustement (if so then please open a new issue where we can discuss this then).
Currently I do not understand the problem 100% as the user can always navigate away from the current activity...

src/main/java/com/owncloud/android/utils/FileStorageUtils.java
tobiasKaminsky: As you have to adapt the other comments (I cannot do it as I do not know exactly what to write/the problem is, please do this also.)

tobiasKaminsky:
@mario I thought it would be only comment changing, which I would have done to get this out, but it is a little bit more.
I know that we both have too little time at the moment, but please try to find some time to do this, then I can review it.

> If mobile data is disabled and I open up e.g. a text file it shows shortly the dialog, then dismisses it, leading to the impression that we now have the updated file (which might be true).
So if the refresh fails we should show a snackbar(?) that says "we could not update the file. Last version available is shown" (or similar)

AndyScherzinger:
:+1:

tobiasKaminsky:
Damn. I approved the wrong PR yesterday.
In this the problem I mentioned in https://github.com/nextcloud/android/pull/1655#issuecomment-348149301 still remains.

tobiasKaminsky:
@mario I guess the easiest is to just create a new pull request with fix fix syncing ;-)

PR #1663, opened 2017-10-13
Use same algo for avatar placeholder

tobiasKaminsky:
From server: https://github.com/nextcloud/server/blob/master/core/js/placeholder.js#L126

Now we have exacty the same colors:
screenshot_20171013_081738


AndyScherzinger:
👍 code wise fine!

mario:
👍

PR #1672, opened 2017-10-15
Detect walled garden

mario:
Adds support for detecting walled gardens aka hotels and stuff. There are a few issues regarding this, but I can't find them atm.

### DIFF
--- a/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
+++ b/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
@@ -45,6 +45,7 @@ import com.owncloud.android.lib.resources.files.ReadRemoteFileOperation;
 import com.owncloud.android.lib.resources.files.RemoteFile;
 import com.owncloud.android.lib.resources.files.UploadRemoteFileOperation;
 import com.owncloud.android.operations.common.SyncOperation;
+import com.owncloud.android.utils.ConnectivityUtils;
 import com.owncloud.android.utils.FileStorageUtils;
 import com.owncloud.android.utils.MimeType;
 import com.owncloud.android.utils.MimeTypeUtil;
@@ -340,6 +341,11 @@ public class UploadFileOperation extends SyncOperation {
         }
 
         try {
+
+            if (Device.getNetworkType(mContext).equals(JobRequest.NetworkType.ANY) ||
+                    ConnectivityUtils.isInternetWalled(mContext)) {
+                return new RemoteOperationResult(ResultCode.NO_NETWORK_CONNECTION);
+            }
 
             /// Check that connectivity conditions are met and delays the upload otherwise
             if (mOnWifiOnly && !Device.getNetworkType(mContext).equals(JobRequest.NetworkType.UNMETERED)) {
--- a/src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
+++ b/src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
@@ -1,64 +1,62 @@
-/**
- * ownCloud Android client application
+/*
+ * Nextcloud Android client application
  *
- * @author David A. Velasco
- * Copyright (C) 2016 ownCloud Inc.
+ * @author Mario Danic
+ * Copyright (C) 2017 Mario Danic
+ * Copyright (C) 2017 Nextcloud GmbH.
  *
  * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License version 2,
- * as published by the Free Software Foundation.
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * at your option) any later version.
  *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
+ * GNU Affero General Public License for more details.
  *
- * You should have received a copy of the GNU General Public License
+ * You should have received a copy of the GNU Affero General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Inspired by: https://stackoverflow.com/questions/6493517/detect-if-android-device-has-internet-connection
  */
 
 package com.owncloud.android.utils;
 
 import android.content.Context;
-import android.content.Intent;
-import android.content.IntentFilter;
-import android.net.ConnectivityManager;
-import android.net.NetworkInfo;
-import android.os.BatteryManager;
-import android.support.v4.net.ConnectivityManagerCompat;
+import android.util.Log;
 
-import com.owncloud.android.lib.common.utils.Log_OC;
+import com.evernote.android.job.JobRequest;
+import com.evernote.android.job.util.Device;
+
+import org.apache.commons.httpclient.util.HttpURLConnection;
+
+import java.io.IOException;
+import java.net.URL;
 
 public class ConnectivityUtils {
 
     private final static String TAG = ConnectivityUtils.class.getName();
 
-    public static boolean isAppConnectedViaUnmeteredWiFi(Context context) {
-        ConnectivityManager cm = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
-        boolean result =
-                cm != null && cm.getActiveNetworkInfo() != null
-                        && cm.getActiveNetworkInfo().getType() == ConnectivityManager.TYPE_WIFI
-                        && cm.getActiveNetworkInfo().getState() == NetworkInfo.State.CONNECTED
-                        && !ConnectivityManagerCompat.isActiveNetworkMetered(cm);
-        Log_OC.d(TAG, "is AppConnectedViaWifi returns " + result);
-        return result;
-    }
-
-    public static boolean isAppConnected(Context context) {
-        ConnectivityManager cm = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
-        return cm.getActiveNetworkInfo() != null && cm.getActiveNetworkInfo().isConnected();
-    }
-
-    public static boolean isCharging(Context context) {
-        IntentFilter ifilter = new IntentFilter(Intent.ACTION_BATTERY_CHANGED);
-        Intent batteryStatus = context.registerReceiver(null, ifilter);
-
-        int status = 0;
-        if (batteryStatus != null) {
-            status = batteryStatus.getIntExtra(BatteryManager.EXTRA_STATUS, -1);
+    public static boolean isInternetWalled(Context context) {
+        if (!Device.getNetworkType(context).equals(JobRequest.NetworkType.ANY)) {
+            try {
+                HttpURLConnection urlc = (HttpURLConnection)
+                        (new URL("http://clients3.google.com/generate_204")
+                                .openConnection());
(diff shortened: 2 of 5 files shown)


src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -335,6 +336,11 @@ protected RemoteOperationResult run(OwnCloudClient client) {
 
         try {
 
+            if (Device.getNetworkType(mContext).equals(JobRequest.NetworkType.ANY) ||
+                    ConnectivityUtils.isInternetWalled(mContext)) {
```
AndyScherzinger: If ```JobRequest.NetworkType.ANY``` then ```isInternetWalled``` will always ```return true``` ;)

AndyScherzinger: Just out of curiosity, does ```JobRequest.NetworkType.ANY``` mean you don't have a network connection?

mario: True, but this way it would not even try to make a request :)

src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
```diff
@@ -1,64 +1,62 @@
...
-            status = batteryStatus.getIntExtra(BatteryManager.EXTRA_STATUS, -1);
+    public static boolean isInternetWalled(Context context) {
+        if (!Device.getNetworkType(context).equals(JobRequest.NetworkType.ANY)) {
+            try {
+                HttpURLConnection urlc = (HttpURLConnection)
+                        (new URL("http://clients3.google.com/generate_204")
+                                .openConnection());
+                urlc.setRequestProperty("User-Agent", "Android");
+                urlc.setRequestProperty("Connection", "close");
+                urlc.setConnectTimeout(1500);
```
AndyScherzinger: Also just a question, do we think 1500ms is enough? Might timeout on very slow networks maybe, so I'd think about going for 5000ms maybe.

src/main/java/com/owncloud/android/operations/UploadFileOperation.java
mario: Yes, as far as I know, it means there's no network connection.

src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
mario: 1500ms is 1.5 seconds. Think that's not enough for poking an empty page? :-/

src/main/java/com/owncloud/android/utils/FilesSyncHelper.java
```diff
@@ -270,15 +274,15 @@ public static void scheduleNJobs(boolean force, Context context) {
...
-                }
+        if (hasImageFolders || hasVideoFolders) {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                scheduleJobOnN(hasImageFolders, hasVideoFolders, force);
             }
+        } else {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                cancelJobOnN();
+            }
+        }
```
AndyScherzinger: This could imho be simplified to:
```
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
    if (hasImageFolders || hasVideoFolders) {
        scheduleJobOnN(hasImageFolders, hasVideoFolders, force);
    } else {
        cancelJobOnN();
    }
}
```

src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
AndyScherzinger: Don't know. In general I'd say yes but I don't know how it'll behave on very slow networks.

src/main/java/com/owncloud/android/utils/FilesSyncHelper.java
mario: Yes, but then you get Android Studio complaining :P

src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
mario: Set to 2.5 :)

mario: (Also, if you're on super slow networks, why would you want to upload?)

AndyScherzinger:
👍 no remarks, while I don't have an option to really test this :/

src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
```diff
@@ -1,64 +1,62 @@
...
+    private final static String TAG = ConnectivityUtils.class.getName();
 
-        int status = 0;
-        if (batteryStatus != null) {
-            status = batteryStatus.getIntExtra(BatteryManager.EXTRA_STATUS, -1);
+    public static boolean isInternetWalled(Context context) {
+        if (!Device.getNetworkType(context).equals(JobRequest.NetworkType.ANY)) {
+            try {
+                HttpURLConnection urlc = (HttpURLConnection)
+                        (new URL("http://clients3.google.com/generate_204")
```
tobiasKaminsky: Can we ping our server instead? I am not feeling comfortable with letting google know infos about me (from where, how often app is used/trying to connect)...
If the server does not have such a 204 page, we should ask the server guys to implement it and then use it.

AndyScherzinger: I agree, being able to check this via the server of the current account would be best 👍

tobiasKaminsky: They will at least know the ip and therefore the rough location, and could cross-check it with their other metadatas they have (which might be more theoretic). But we should rely on our own system and avoid exposing this to external...

mario: If you really want, try and get this in the server. But this would mean we don't detect walled gardens until 13.0. 

cc @schiessle or someone who could do this :P Basically an endpoint that just returns 204 :D

tobiasKaminsky: For <13 we could use nextcloud/status.php. It is not 204, but it is lightweight enough to not generate that much traffic.

tobiasKaminsky:
Regarding server I asked here: https://github.com/nextcloud/server/issues/6967

src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
mario: You'd have to parse the response to see if it's a JSON object at least if you get 200, but sure. Feel free to modify the PR.

j-ed:
If I understand the discussion correctly, you want to access a specific page of a Nextcloud server to identify if you're in a closed environment or not. As far as I remember many hotels are redirecting all kind of requests to a standard page so that you might not be able to judge if it is the correct reply or not. Some wifi environments are using fake certificates to pretend to be a specific page to intercept TLS communication and also send out the reply you request. Others are suppressing ping replies so that this way couldn't be gone either.
Wouldn't it be possible to request the certificate of the Nextcloud server, compare its subject with the server name which has been used to access it and additionally check the certificate chain and use this as flag if the server an can be reached or not? I assume that this function has already part of the code and could be used for the check to.

tobiasKaminsky:
PR for 13: https://github.com/nextcloud/server/pull/7041

tobiasKaminsky:
For NC13: https://github.com/nextcloud/server/pull/7041
For NC12 I would go and use status.php

src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
```diff
@@ -1,64 +1,62 @@
-/**
- * ownCloud Android client application
+/*
+ * Nextcloud Android client application
  *
- * @author David A. Velasco
- * Copyright (C) 2016 ownCloud Inc.
```
tobiasKaminsky: I guess we cannot simply remove copyright / author...

src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
```diff
@@ -1,64 +1,62 @@
...
+ * @author Mario Danic
+ * Copyright (C) 2017 Mario Danic
+ * Copyright (C) 2017 Nextcloud GmbH.
  *
  * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License version 2,
- * as published by the Free Software Foundation.
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * at your option) any later version.
```
tobiasKaminsky: Same here. The older code is GPL 2, we can add that newer code ist GPL 3+, but changing without asking is not allowed.

mario:
Please note that all old code has been removed from the file, so why would
I keep copyright for David?

tobiasKaminsky:
You are too fast, I already saw this and thus removed the comments ;-)

tobiasKaminsky:
This does not compile:
https://drone.nextcloud.com/nextcloud/android/2292/6:
>drone/src/github.com/nextcloud/android/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java:354: error: cannot find symbol
if (ConnectivityUtils.isAppConnected(MainApp.getAppContext())) {

Changing this leads to:
> Caused by: java.lang.ClassCastException: com.android.okhttp.internal.huc.HttpURLConnectionImpl cannot be cast to org.apache.commons.httpclient.util.HttpURLConnection

Please fix this and try it. After that I am happy to add the new url for NC13 and parsing status.php for <=N12.

mario:
@tobiasKaminsky this line doesn't even exist. Where did the drone pick that up?

tobiasKaminsky:
The latest commit here is  dc745e368f3c5442af2f3d4749dcb721a9126c43

And if you checkout/browse the code, you'll find it there.
https://github.com/nextcloud/android/blob/dc745e368f3c5442af2f3d4749dcb721a9126c43/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java#L354

mario:
Done all we need for 13.

tobiasKaminsky:
Is this working for you? 
With your latest commit I get:

> java.lang.RuntimeException: Unable to create application com.owncloud.android.MainApp: java.lang.ClassCastException: com.android.okhttp.internal.huc.HttpsURLConnectionImpl cannot be cast to org.apache.commons.httpclient.util.HttpURLConnection
at android.app.ActivityThread.handleBindApplication(ActivityThread.java:5364)
at android.app.ActivityThread.-wrap2(ActivityThread.java)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1528)
at android.app.ActivityThread.main(ActivityThread.java:6077)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:866)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:756)
Caused by: java.lang.ClassCastException: com.android.okhttp.internal.huc.HttpsURLConnectionImpl cannot be cast to org.apache.commons.httpclient.util.HttpURLConnection
 at com.owncloud.android.utils.ConnectivityUtils.isInternetWalled(ConnectivityUtils.java:54)

tobiasKaminsky:
Tested on 
- NC13 with http
- NC12 with http
- NC12 with https

only remaining is
- NC13 with https

tobiasKaminsky:
Please beware that we need a new lib release and change build.gradle before

AndyScherzinger:
👍

PR #1674, opened 2017-10-15
Fix 1521

mario:
Also excludes pretty much everything else, so you only restore the app itself I guess. 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.0-beta7'
+        classpath 'com.android.tools.build:gradle:3.0.0-rc1'
         classpath 'com.google.gms:google-services:3.0.0'
     }
 }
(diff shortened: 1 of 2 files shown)


mario:
Unsure we should do this (I'd rather prefer smooth migration) but *shrug*...

AndyScherzinger:
What do you think @tobiasKaminsky? 
Imho (if done right 😝) the what'S new "first run" could be used for cleaning up backup/restore data that should not be in place, like certain DB entries.

tobiasKaminsky:
We can use "firstRun" to detect, but DB cleanup or similar tasks should be done in a new function.

So this does what? Prevent apps like Titanium to backup data? If so, then there is no need to backup the app at all, because then it is at restore always a vanilla app?

mario:
Indeed. So maybe just disable backup? *shrug*

tobiasKaminsky:
For a backup we "only" need to have the account, auto upload and "keep available offline" restored. 
Or am I missing something?
Everything else (favorites, file cache, ...) can be retrieved during runtime.

While auto upload and "keep available offline" is harder to do, we maybe should focus on restoring account? And if we (somehow) can detect that it is a backup, we can show an hint that auto upload and keep available offline needs to be configured again.

mario:
Closing this.

PR #1688, opened 2017-10-19
Quotes within strings and "F-Droid"

comradekingu:
At least some of the ""s within strings are erroneous, so tentatively removed all such instances.
"F-Droid" as per https://f-droid.org/


AndyScherzinger:
rebased!

piniging @mario for review :)

mario:
👍

Issue #1698, opened 2017-10-19
All the favorites are displayed twice - refresh helps [2.0.0 RC9]


mr-bolle:
### Actual behaviour
- Every time if i open the favorites Page i see all items twice.

### Expected behaviour
- If i open the favorites page, i would only see every item once
 
### Steps to reproduce
1. open Nextcloud Add - select favorites > all items visible twice

### Environment data
Android version: 7.1.1

Device model: Sony Xperia Z5 dual

Stock or customized system: Stock ROM

Nextcloud app version: 2.0.0 RC9 & 2.0.0 (final 18.10.2017)

Nextcloud server version: 12.0.3

![favorites twice](https://user-images.githubusercontent.com/22414212/31784725-07396a54-b503-11e7-91ef-b43874047eff.png)

zweiflamme:
That's exactly what happens to me - the same behavior is also reproducible with the Gallery. After a refresh, all items ore only shown once, as expected. Android 7.1.2 on a BB Motion, Nextcloud server is 0.53.24, Nextcloud App is 3.0.3

zweiflamme:
Sorry, Nextcloud Server is 13.0.0.1 (I believe). Have installed NextCloudPi with the above v053, my apologies for the confusion.

AndyScherzinger:
PR is #2490

PR #1704, opened 2017-10-21
Prevent favorites from displaying twice

apicellaj:
Proposed fix for issue #1698 (also affects selecting "Photos" from Navigation Drawer)

### DIFF
--- a/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
+++ b/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
@@ -1598,9 +1598,7 @@ public class OCFileListFragment extends ExtendedListFragment implements OCFileLi
                     mAdapter.notifyDataSetChanged();
                 }
             }
-        };
-
-        remoteOperationAsyncTask.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR, true);
+        }.execute("");
     }
 
     private void setTitle(@StringRes final int title) {


AndyScherzinger:
@mario can you test this please? I would but for some reason favorites (only have 1 file) loads forever (only displaying "loading...")...

also cc @tobiasKaminsky

tobiasKaminsky:
I tested it successfully 👍, but I am unsure regarding the impact of directly executing instead of using the pool.
@mario what do you think?

mario:
It used to be this way. It didn't work. Then we changed it. Then it worked. Now it doesn't work again xD

Will check...

mario:
So both ways work for me. However, the real issue seems to be this:

 10-25 18:26:55.025 5964-6013/com.nextcloud.client D/OwnCloudClient #1: REQUEST SEARCH /server/remote.php/dav
10-25 18:26:56.788 5964-6013/com.nextcloud.client D/OwnCloudClient #1: REQUEST SEARCH /server/remote.php/dav

Search is issued twice no matter the executor...

mario:
cc @AndyScherzinger @apicellaj @tobiasKaminsky

AndyScherzinger:
Now also related to #1732 but I can't test it :( since Fav/Photos don't work for me and I don't have the time to investigate atm. 😢

alexanderdd:
Came here because of the photos showing up twice. Curious as to how this story continues :+)

tobiasKaminsky:
> 10-25 18:26:55.025 5964-6013/com.nextcloud.client D/OwnCloudClient #1: REQUEST SEARCH /server/remote.php/dav
10-25 18:26:56.788 5964-6013/com.nextcloud.client D/OwnCloudClient #1: REQUEST SEARCH /server/remote.php/dav
>
>Search is issued twice no matter the executor...

With latest app master and latest server master I cannot reproduce this: I only see one "request search" and only one search execuction:
https://github.com/nextcloud/android/blob/daf68791529a140a7a695cc7145e3bd6d39e636f/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java#L1551-L1551

What kind of environment did you used?

ChrisDeath:
Btw. for me its only shown twice the first time i enter Favorites. Refreshing then there (pulling down the list) removes the duplicates...maybe it helps for testing.

AndyScherzinger:
@tobiasKaminsky @mario I found the cause for the double search inside of OCFileListFragment...

```
@Override
    public void onActivityCreated(Bundle savedInstanceState) {
        ...
        if (searchEvent != null && savedInstanceState == null) {
            onMessageEvent(searchEvent);
        }
        ...
    }
```
and
```
@Override
    public void onViewStateRestored(@Nullable Bundle savedInstanceState) {
        ...
        if (searchEvent != null) {
            onMessageEvent(searchEvent);
        }
        ...
    }
```

So whenever you click on Favorites in the drawer the onActivityCreated gets called and right after onViewStateRestored gets called, both then triggering the search.

So we either:
* need to know in which of the two cases this should actually happen and remove the other event
* have a nasty quick fix by checking the list within the adapter...see OCFileListAdapter
```
private void parseVirtuals(ArrayList<Object> objects, ExtendedListFragment.SearchType searchType) {
        ...
        for (int i = 0; i < objects.size(); i++) {
            ...

                if (!onlyImages || MimeTypeUtil.isImage(ocFile)) {
                    mFiles.add(ocFile);
                }
               ...
```
where we replace it by:
```if (!mFiles.contains(ocFile) && !onlyImages || MimeTypeUtil.isImage(ocFile))) {```

I think i'd be nice if we fix the double calling since the other solution just throughs away the second result since all the elements will already be in the list the second time...

PR #1708, opened 2017-10-24
Prevent action bar title from being overwritten

### LINKED ISSUE #1693: Top bar should say name of current section when in root folder
Except when in "All Files" where it should say "Nextcloud" as now, the header bar text should reflect the current section. Like "Favorites", "On device" etc.

apicellaj:
Proposed fix for #1693 

Resolves #1693 

Setting the title within the AsyncTask to prevent the ToolbarActivity from overwriting the title with the default provider title. I did some testing and it seems to work fine since the setTitle function is being called to run on the UI thread.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -156,7 +156,7 @@ public class FileDisplayActivity extends HookActivity
         SendShareDialog.SendShareDialogDownloader {
 
     public static final String RESTART = "RESTART";
- 
+
     private SyncBroadcastReceiver mSyncBroadcastReceiver;
     private UploadFinishReceiver mUploadFinishReceiver;
     private DownloadFinishReceiver mDownloadFinishReceiver;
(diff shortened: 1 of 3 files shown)


AndyScherzinger:
Resolves #1693

AndyScherzinger:
Hi @apicellaj,

thanks for yet another PR 😃 
I tested it for "On Device" and the title is set to the Sever name (theming picked name from the server) while it should say "On Device".

apicellaj:
Hi @AndyScherzinger,

Thank you as well 😃 I completely forgot to add in the functionality for the "On device" title. I didn't see an existing method for setting "On device" so I went ahead and created one.

tobiasKaminsky:
When on slow internet connection and opening "favorites" it says "nextcloud" until the favourites are loaded. Same is for "photos".

On device is working only partly:
- when in file view and switching to "on device" -> works
- when in other view e.g. "notifications" and switching directly to "on device" -> "nextcloud"

apicellaj:
On device should be working now.

I'm not able to simulate slow speeds on my device or emulator that mimic what you're seeing with the delay in changing the action bar. Thoughts on setting the title in remoteOperationAsyncTask's onPreExecute as well?

tobiasKaminsky:
:+1: 

Rebased to latest master and now everything seems to work. I guess this is due to the way we are now handling our activities.

AndyScherzinger:
@tobiasKaminsky so merge?

tobiasKaminsky:
If you have tested and approved it, yes @AndyScherzinger

AndyScherzinger:
I couldn't test it yet due to time constraints on my side :(

mario:
👍

PR #1714, opened 2017-10-26
Upgrade gradle plugin & use docker image 27

tobiasKaminsky:
This speeds drone up again as all new build tools are included.
After this merge I will do another new docker image as we can only then include gradle 3.0 plugin

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,6 +1,6 @@
 pipeline:
   test:
-    image: nextcloudci/android:android-26
+    image: nextcloudci/android:android-27
     commands:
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
We have to go for gradle-3.0.0 here, otherwise: 

> ../../../build.gradle:17: A newer version of com.android.tools.build:gradle than 3.0.0-rc2 is available: 3.1.0-alpha01

AndyScherzinger:
👍 looks good to me 👍

tobiasKaminsky:
As gradle 3.0. is already in master, I have to recreate the docker image.

PR #1729, opened 2017-10-31
Update AndroidManifest.xml

anantshree:


tobiasKaminsky:
This is already in https://github.com/anantshree/android/blob/f76ee555fd44d7eb39c67edb30adfd6294221058/src/main/AndroidManifest.xml#L52

PR #1752, opened 2017-11-06
transifex -> fastlane converter

tobiasKaminsky:
Thanks to @grote (https://github.com/fastlane/fastlane/issues/9520#issuecomment-341915827)

This allows us to have full/short description on transifex in only one resource.
With @MorrisJobke I will intregrate this to our daily transifex job, so it will get updated each day.

- [ ] move translated strings from transifex android app store listing to android, only possible once this is merged


AndyScherzinger:
👍

tobiasKaminsky:
As https://github.com/nextcloud/docker-ci/pull/62 was merged, we need to merge this also, otherwise all transifex updates will fail.

lint.xml
```diff
@@ -4,4 +4,8 @@
         <ignore path="**/freemarker-2.3.23.jar"/>
         <ignore path="**/nnio-0.2.jar"/>
     </issue>
+    
+    <issue id="UnusedResources">
+        <ignore regexp="store_short_desc|store_full_desc" />
+    </issue>
 </lint>
```
mario: New line.

scripts/metadata/generate_metadata.py
```diff
@@ -0,0 +1,122 @@
...
+import os
+import shutil
+import codecs
+from xml.etree import ElementTree
+
+
+XML_PATH = '../../src/main/res'
+METADATA_PATH = '../../fastlane/metadata/android'
+DEFAULT_LANG = 'en-US'
+LANG_MAP = {
```
mario: Can we generate this dynamically?

grote: If you do, make sure you make a pull request to this file upstream ;)

scripts/metadata/generate_metadata.py
```diff
@@ -0,0 +1,122 @@
...
+    'values-no-rNO': 'no-NO',
+    'values-eu': 'eu-ES',
+    'values-lt-rLT': 'lt-LT',
+    'values-zh-rKN': 'zh-HK',
+    'values-zk': 'zk-CN',
+    'values-is': 'is-IS',
+    'values-id': 'id-ID',
+    'values-cs-rCZ': 'cs-CZ',
+    'values-sl': 'sl-SL',
+    'values-fa': 'fa-FA'
```
MorrisJobke: Just take the special mapping from `.tx/config` and then use normal mapping like for "'values-zh-rKN': 'zh-HK'," do 

* drop `values-`
* replace `-r` with `-`

Done ;)

scripts/metadata/generate_metadata.py
MorrisJobke: Makes sense. Just take the special mapping from `.tx/config` and then use normal mapping like for "'values-zh-rKN': 'zh-HK'," do 

* drop `values-`
* replace `-r` with `-`

Done ;)

tobiasKaminsky:
> ../../../src/main/res/values-ka-rGE/strings.xml:191: For locale "ka" (Georgian) the following quantities should also be defined: one

as figured out with @MorrisJobke georgian language does not have different quantities. 
Changing it from "other" to "one" works locally.
But I have no clue how to do this on transifex...

AndyScherzinger:
Seems to be a transifex issue since Georgian has the translations for one and other for the strings but then it seems that transifex doesn't export them?! I tried to replace it but as soon as I set the "one" value is is shown for other too and during export it will only be exported as other.

AndyScherzinger:
Seems to be a transifex issue since Georgian has the translations for one and other for the strings but then it seems that transifex doesn't export them?! I tried to replace it but as soon as I set the "one" value is is shown for other too and during export it will only be exported as other.

scripts/metadata/generate_metadata.py
tobiasKaminsky: I doubt that it is that easy.
For example in .txc.conf is id_ID: id-rID which is in our project only "values-id".
German is "values-de" and "values-de-rDE", and not translated to "rCH", "rAT".
So for switzerland (ch) values-de is used. Of course we can add "values-de-rCH", but this would double the work and I guess it is the same content...

tobiasKaminsky:
As Georgian language only has one case, it indeed seems fine/sane for me to use "other", as as this will be used if no other case will hold. So it is fine that transifex only needs and exports "other".
Strange is then, that lint/android studio wants to have also "one".
As a hotfix I set lint to ignore missing quantity errors on Georgian language.

scripts/metadata/generate_metadata.py
AndyScherzinger: Need to agree with @tobiasKaminsky - unfortunately the codes do different on the different plattforms :/

tobiasKaminsky: We can try to enhance this and discuss it on a new issue if you want.
But for now (and to get master building again) I would like to merge this as it is.

AndyScherzinger:
Build turned green now and new line is fixed @mario so you might take another look :)

mario:
👍

MorrisJobke:
And here it is: d4b7d264eff04feced22550fbb2e3915e95b7dde 😉

Issue #1308, opened 2017-08-11
Theming: misbehave on white background color


tobiasKaminsky:
As it is (hopefully) a rare corner case, I am fine with 2.0.1.

![2017-08-11-100235](https://user-images.githubusercontent.com/5836855/29205393-d7e5eb36-7e7c-11e7-8f9e-d35840d72ddf.png) ![2017-08-11-100825](https://user-images.githubusercontent.com/5836855/29205438-151f7db4-7e7d-11e7-8bd9-89ae46f68802.png)

Ref for further color handling questions: https://github.com/nextcloud/server/issues/5654

mario:
@nickvergessen gets black on other colors too :( Strange...

mario:
@nickvergessen since I'm away, can you please put screenshot here?

tobiasKaminsky:
Black was related to the font color and this is fixed via https://github.com/nextcloud/android/pull/1314 as we now use the same color threshold computing method.

AndyScherzinger:
@juliushaertl can you give us a hint how/where you calculate the icon color, especially in case a user chooses white or very light colors?

tobiasKaminsky:
Maybe the theming api can give us already the right color, so that we are exactly consistent and do not have do the math twice (same for font color).

juliushaertl:
We use a simple grey as "element color" for brightness levels greater than 0.8, see https://github.com/nextcloud/server/blob/master/apps/theming/lib/Util.php#L80-L87 

Makes sense to me to add this to the theming capabilities endpoint.

juliushaertl:
Here we go: https://github.com/nextcloud/server/pull/7094

PR #1762, opened 2017-11-08
Use text-color and element-color whenever possible (>=NC13)

tobiasKaminsky:
Fallback will be to computed values, and if this does not work fallback to default one.

Also computes element color on very light themes.

Fixes: #1308


AndyScherzinger:
👍 looks good code wise, don't have NC13 instance to test though

mario:
👍

PR #1764, opened 2017-11-08
Brander: support for custom weblogin urls

tobiasKaminsky:
Always use ocs-api header.
weblogin_url in setup.xml is generic on purpose so that any url/server can handle this.


AndyScherzinger:
👍 looks good

mario:
Extra header can't hurt, but we don't always need it :)

How does this helps with brander?

MorrisJobke:
> How does this helps with brander?

Currently we use weblogin_url for all newly branded apps and the login then does not work, because the header is not set.

mario:
Right. Guess this is because the custom url was initially not designed for this :)

Anyway...

👍

PR #1767, opened 2017-11-09
Tint background on drawer/user info as fallback

tobiasKaminsky:
In theming app one can set an image or a plain color.
On NC12 it was both on the same string via capabilities: either url (image) or hex code (color)
On NC13 this changed: it is always an image, but on color it is a transparent png.

This now always paints the background with the primary color and on top shows always the image.


AndyScherzinger:
👍 looks good to me

mario:
👍

Issue #1523, opened 2017-09-09
Dev version: no independant branch, but master


tobiasKaminsky:
As decided on conf we want to have the dev version to focus on the current master, to test all merged features as soon as possible. 

For this we need a new flavor which replaces the appid,provider names, ...
It is a good start to compare dev-base (https://github.com/nextcloud/android/tree/dev-base) with master to see which files are necessary to be changed.

There will still be a nightly cronjob to test if something has changed and then publish the new version both to Fdroid and our internal https://download.nextcloud.com/android/dev/

This change will be done as soon as all "ready for dev" branches are merged, so that ideally no one experience a difference ;-)

j-ed:
@tobiasKaminsky Might this be the reason why currently no nighty builds can be downloaded?
```
https://download.nextcloud.com/android/dev/latest.apk -> Access forbidden.
https://f-droid.org/packages/search/nextcloud -> Not found.
```

PR #1775, opened 2017-11-13
Dev version as flavor

tobiasKaminsky:
This introduces dev version as flavor and builds from master, as described in #1523.
Flavor is called "versionDev" to have "generic" still as first.


AndyScherzinger:
@tobiasKaminsky all build flavors fail atm :/

tobiasKaminsky:
Fixed, please review, so we then can do dev version based on master 👍 
@mario @AndyScherzinger

tobiasKaminsky:
https://drone.nextcloud.com/nextcloud/android/2346/13
> 1209 To run dex in process, the Gradle daemon needs a larger heap.
1210 It currently has 1024 MB.
1211 For faster builds, increase the maximum heap size for the Gradle daemon to at least 4608 MB (based on the dexOptions.javaMaxHeapSize = 4g).
1212 To do this set org.gradle.jvmargs=-Xmx4608M in the project gradle.properties.

Asked Morris if we can increase this on the drone workers.

tobiasKaminsky:
@mario @AndyScherzinger please review so we can get dev version back running 👍

AndyScherzinger:
👍 looks good only minor thing which is not that important is the usage of Toast instead of snackbar the new Android promoted way to show short information stuff.

tobiasKaminsky:
Nice catch 👍

AndyScherzinger:
😃👍

mario:
👍 


Rebase.

PR #1777, opened 2017-11-13
if privacy url is set to a pdf file, do not open it inline but external

tobiasKaminsky:
Title says it all, otherwise the pdf was not shown...
(needed for branding as some wants to link a pdf)


AndyScherzinger:
👍 please fix codacy though, other issue would imho be to maybe reverse it, no? Like if it is text/html mime type then webview, else call general view intent.

tobiasKaminsky:
Your idea will unfortunately not work with our privacy url: https://nextcloud.com/privacy 
Mimetype of "privacy" is unknown. It indeed would work back in 90s, where everything ended with "html" or "php" ;-)

AndyScherzinger:
You're right :)

mario:
👍

Issue #531, opened 2017-01-11
Integrate send and share in menu


jospoortvliet:
### Actual behaviour
If you want to share a file over WhatsApp the 'share' menu item is of little help. Confusingly you have to use 'send'. This makes sense from a technical pov but not from and end user perspective.

### Expected behaviour
Share should allow you to share - tech details be damned. I suggest to show a share via nextcloud action as part of the standard android share menu...

jospoortvliet:
@jancborchardt it probably needs a little thinking from you... what the logical behaviour is when sharing a file already on Nextcloud, sharing a file not on nextcloud etc...

tobiasKaminsky:
What about "share" -> "share via nextcloud" and "send" -> "share to other app"?

jospoortvliet:
Yeah, that could work, probably the easiest, quickest solution, though it might be too long for the menu...

The 'prettiest' solution, imho, would be to have, in the Android Share menu, one item (Nextcloud) which would act differently depending on the file you share:
* if already on Nextcloud, give the usual Nextcloud share dialog
* if not on Nextcloud, allow user to upload to Nextcloud

That way, you have only one share button/dialog.

mario:
@tobiasKaminsky suggestion, have a share button and edit the list of available shares to add Nextcloud as a first option. That way one menu, one way to do something.

Will see what I can talk @jancborchardt into drawing :P

tobiasKaminsky:
I do not understand your idea :/
Can you make me a mockup or find an app that is similar to the way you want to have it?

jospoortvliet:
I don't understand it either... Is what I proposed not possible?

jancborchardt:
I understand what @jospoortvliet means, I think. Basically, that there is not »Share« and »Send« but only »Share«. And tapping that opens the list of apps/options which are shown when you would tap »Send« – with Nextcloud / »Share via Nextcloud« being the first one. And tapping that would open what right now is visible under »Share«.

I agree this simplifies the flow because it combines two options which do the same into one.

Correct @jospoortvliet? And is this even possible @tobiasKaminsky @mario @AndyScherzinger?

If it’s not possible we could do what iOS does, calling it »Share« and renaming »Send« to »Open in …«

jospoortvliet:
@jancborchardt you're exactly right.

AndyScherzinger:
[...]

PR #1792, opened 2017-11-16
Combine send share

tobiasKaminsky:
WIP of #531 

Folder re-shareable:

Folder not re-shareable:

File re-shareable:

File not re-shareable:

Of course this is only a first version, so feedback regarding design is always welcome :-)


tobiasKaminsky:
Share icon and text needs to be clickable, currently only text is clickable.

tobiasKaminsky:
@jancborchardt @AndyScherzinger 
I have updated the UI, please have look :-)

jancborchardt:
Instead of "Resharing is not allowed" let's say "You don't have permissions to reshare this". And also say that below the greyed out icons in the share pane for files as feedback why they are greyed out. :)

AndyScherzinger:
Looks good to me, only issue would be that probably due to the proportions of the two icons the don't align horizontally looking at the labels.

In a future iteration of the feature it could also make use of a circular, primary color colored background but would have to check text color to tint the icon black or white. Nothing we need to do for this PR.

jancborchardt:
Right, looking at it again the link icon is actually bigger (or lower down) than the share icon. How come?

tobiasKaminsky:
It was a type 0 and 8 look too closely ;-)

AndyScherzinger:
Looks good to me 👍

tobiasKaminsky:
@AndyScherzinger do you have time to look into your idea with rounded image?
I would like to have this directly in the PR to have it looking nice on the very first time 🎉

AndyScherzinger:
> do you have time to look into your idea with rounded image?

Quite busy but I'll give it a try while I then might need some support for the theming part.

tobiasKaminsky:
I can do the theming part, but I tried & failed to get a round imageView with a solid background color and the image (chain/anchor) on it.

AndyScherzinger:
@tobiasKaminsky DONE, you not "just" need to theme the imageView's src and background (+remove the tint attribute I added to the two image views) 🎉

tobiasKaminsky:
I removed alpha so it is not that greyish any more.
@jancborchardt design feedback?

tobiasKaminsky:
Thank you @AndyScherzinger for this quick (and apparently small) change and help 🎉

AndyScherzinger:
@tobiasKaminsky you are very welcome. As for the icons one thing missing is the tinting. The circular background is correctly tinted with the primary color but the icon itself should also be tinted black/white like the text in the toolbar/actionbar 😁 (at least in my opinion). Else there will be a problem with very dark primary colors...

AndyScherzinger:
Plus the "Link" icon needs a slight bump to the right so maybe raise the left padding by 1 or 2 dp.

tobiasKaminsky:
....nooooooow? :-)

AndyScherzinger:
Nice, but then I'd say also use white for the icon so it matches the shared folder coloring :)

tobiasKaminsky:
Well, this white is hardcoded...

https://github.com/nextcloud/android/blob/f9983558c97bd0ef61158b9dc93ab89d20e60e2a/src/main/java/com/owncloud/android/utils/MimeTypeUtil.java#L142-L154

The black color on the new icons is the dynamic one from the server (font color) and the background color is "elementColor".
Using a very light yellow is showing the folder icon still white, but same does the server, which seems wrong to me...

 image

AndyScherzinger:
Hmmm, calling @juliushaertl to the rescue :)

tobiasKaminsky:
I did already ;-)
https://github.com/nextcloud/server/issues/7401

tobiasKaminsky:
Drone says: This tag and its children can be replaced by one <TextView/> and a compound drawable
for 
https://github.com/nextcloud/android/blob/3dc9e3852c8c62540d49df8d112050f2c3e188b5/src/main/res/layout/send_share_fragment.xml#L74-L74

But I do not see any chance to get a background in it.
So I fear that we have to live with the +2 warnings (or disable this check completely as it will create on many more occurences false positive warnings).

AndyScherzinger:
I'd say let's live with the two warnings for now :)

tobiasKaminsky:
This is ready for review.
As discussed this will add 2 warnings to lint: https://github.com/nextcloud/android/pull/1792#issuecomment-349653072

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -539,23 +543,16 @@ public boolean isSearchUserSupportedSupported() {
...
-            // Show dialog
-            final String sendTitle = mFileActivity.getString(R.string.activity_chooser_send_file_title);
-            mFileActivity.startActivity(Intent.createChooser(sendIntent, sendTitle));
-        } else {
-            Log_OC.e(TAG, "Trying to send a NULL OCFile");
-        }
+        SendShareDialog mSendShareDialog = SendShareDialog.newInstance(file);
+        mSendShareDialog.setFileOperationsHelper(this);
+        mSendShareDialog.setFileDisplayActivity(fileDisplayActivity);
+        mSendShareDialog.show(ft, "TAG_12");
```
AndyScherzinger: What is the meaning of ```"TAG_12"```?

src/main/res/layout/file_details_fragment.xml
```diff
@@ -187,13 +187,13 @@
...
-				android:layout_height="wrap_content"
-				android:layout_marginBottom="@dimen/standard_half_margin"
-				android:text="@string/action_share"/>
+                android:id="@+id/fdShareTitle"
+                style="@style/TextAppearance.AppCompat.Body2"
+                android:textColor="@color/color_accent"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_marginBottom="@dimen/standard_half_margin"
+                android:text="@string/action_send_share"/>
```
AndyScherzinger: identation is broken here

src/main/res/values-ar/strings.xml
```diff
@@ -95,7 +95,7 @@
     <string name="filedetails_select_file">اضغظ على الملف ليتم عرض خيارات أكثر</string>
     <string name="filedetails_download">تحميل</string>
     <string name="filedetails_sync_file">زامِن</string>
-    <string name="action_share">شارك</string>
+    <string name="action_send_share">شارك</string>
```
AndyScherzinger: renaming is fine for lint while after being merged to master due to the name-change transifex will remove all the translations since it considers it to be "a new string"

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
tobiasKaminsky: I will rename it. A tag is needed, but we do not use it.

src/main/res/layout/file_details_fragment.xml
tobiasKaminsky: I will reformat it with android studio, please expect a "mess", but then it is consistent.

src/main/res/values-ar/strings.xml
tobiasKaminsky: Correct, but this way we do not have lint crying ;-)
And our dear translator friends will handle this then on transifex.

src/main/res/layout/file_details_fragment.xml
AndyScherzinger: fine by me

src/main/res/values-ar/strings.xml
AndyScherzinger: True, was only a comment )as an FYI, no todo or else) :)

AndyScherzinger:
👍 fine by me, I tested the PR and also did a minor design optimization regarding size/whitespace https://github.com/nextcloud/android/pull/1792/commits/12a0a7aceea995848c9704b4bd48c578be3ef7af and https://github.com/nextcloud/android/pull/1792/commits/67fb57a433c144d9b7a554f328fff9ebda85d7cb

cc @mario for second review 😃

src/main/java/com/owncloud/android/ui/dialog/SendShareDialog.java
```diff
@@ -0,0 +1,232 @@
...
+        sharePeopleImageView.setOnClickListener(v -> shareFile(file));
+
+        // Share via link button
+        TextView shareLinkText = view.findViewById(R.id.share_link_button);
+        shareLinkText.setOnClickListener(v -> fileOperationsHelper.showShareFile(file));
+
+        ImageView shareLinkImageView = view.findViewById(R.id.share_link_icon);
+        shareLinkImageView.getBackground().setColorFilter(ThemeUtils.elementColor(), PorterDuff.Mode.SRC_IN);
+        shareLinkImageView.getDrawable().setColorFilter(ThemeUtils.fontColor(), PorterDuff.Mode.SRC_IN);
+        shareLinkImageView.setOnClickListener(v -> shareFile(file));
```
AndyScherzinger: @tobiasKaminsky just a guess but I'd have expected ```fileOperationsHelper.showShareFile(file)``` here ;) Probably a c&p issue.

tobiasKaminsky: This is exactly what I am doing ;-)
It is just a call to a helper function...

https://github.com/nextcloud/android/blob/f906d0f6b7128eaa86811d9938739120c58364ab/src/main/java/com/owncloud/android/ui/dialog/SendShareDialog.java#L193-L195

AndyScherzinger: My point is this:
```
sharePeopleText.setOnClickListener(v -> shareFile(file));
sharePeopleImageView.setOnClickListener(v -> shareFile(file));
shareLinkText.setOnClickListener(v -> fileOperationsHelper.showShareFile(file));
shareLinkImageView.setOnClickListener(v -> shareFile(file));
```

AndyScherzinger:
pining @mario for second review 😃 I also fixed some lint issues to drive down the lint score (just for fun)

tobiasKaminsky:
> I also fixed some lint issues to drive down the lint score (just for fun)

Well...the PR ideally should only contain changes that are related ;-)

AndyScherzinger:
> Well...the PR ideally should only contain changes that are related ;-)

I know but I wanted drone to turn green and after I did that I realized that we discussed that the PR would create two new warnings which we will ignore... I forgot about that...

mario:
@AndyScherzinger please ping me on Monday for this.

src/main/java/com/ortiz/touch/TouchImageView.java
```diff
@@ -1184,14 +1184,8 @@ public void run() {
...
-    		if (VERSION.SDK_INT < VERSION_CODES.GINGERBREAD) {
-    			isPreGingerbread = true;
-    			scroller = new Scroller(context);
-    			
-    		} else {
-    			isPreGingerbread = false;
-    			overScroller = new OverScroller(context);
-    		}
+    		isPreGingerbread = false;
+    		overScroller = new OverScroller(context);
```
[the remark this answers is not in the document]
AndyScherzinger: True, so let's keep it this way, just waiting for @mario's approval. Will have a look at the third-party code in 2018 fixing these warnings :)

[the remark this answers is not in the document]
tobiasKaminsky: 👍

AndyScherzinger:
rebased - pinging @mario for review :) - PR created 2 new warnings which as per discussion can be ignored for now. :)

AndyScherzinger:
rebased - pinging @mario for review :) - PR created 2 new warnings which as per discussion can be ignored for now. :)

mario:
👍

PR #1799, opened 2017-11-20
Set missing content description, RTL support & minor type errors

tobiasKaminsky:


AndyScherzinger:
👍 code looks good while list items should move to dynamic content descriptions since they should incorporate information about the item itself like "activity icon for XYZ" instead of "activity icon" but that can be postponed to a later point in time via another PR. 😃

tobiasKaminsky:
How would you do this? We only have the cryptic id, e.g. "@+id/sortByNameAscending".

AndyScherzinger:
> How would you do this? We only have the cryptic id, e.g. "@+id/sortByNameAscending".

Dynamic content descriptions for list/grid items would have to be set/updated programmatically in the List/Grid/Recycler/etc.-Adapter. But like I said, no worries for now. 👍

mario:
👍

PR #1801, opened 2017-11-20
Hide item options (share, overflow menu) & overflow items in move/copy dialog

tobiasKaminsky:
Before & after


src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
```diff
@@ -102,13 +102,17 @@ protected void onCreate(Bundle savedInstanceState) {
 
         // Action bar setup
         setupToolbar();
-        getSupportActionBar().setDisplayShowTitleEnabled(true);
+
+        if (getSupportActionBar() != null) {
+            getSupportActionBar().setDisplayShowTitleEnabled(true);
+            getSupportActionBar().setTitle(caption);
```
AndyScherzinger: The caption gets dynamically calculated in the lines below ;)

AndyScherzinger:
👍 Just a minor comment 👍

src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
tobiasKaminsky: Right. I changed the order 👍

AndyScherzinger:

mario:
👍

PR #1813, opened 2017-11-27
No snapshot on library

tobiasKaminsky:
We always use latest android library version and not a (random) snapshot.


tobiasKaminsky:
Did a new library release to get latest commits.

PR #1823, opened 2017-12-01
Remove unused translations from values/strings.xml

tobiasKaminsky:
Before we used lint we removed many code references without removing the strings, so this is only a cleanup.
When we merge this some time ahead of next release we can see in dev version if some translations are missing.

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -124,13 +124,10 @@
     <string name="common_yes">Yes</string>
     <string name="common_no">No</string>
     <string name="common_ok">OK</string>
-    <string name="common_remove_upload">Delete upload</string>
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
As I am too lazy to remove all already translated strings by hand, I would like to let transifex do the work.
The downside is, we have to merge this without lint is green.
But right after the merge to master I can trigger transifex bot and all translations should be removed.

AndyScherzinger:
:+1: fine by me, pinging @mario for 2nd review :)

AndyScherzinger:
pinging @mario for 2nd review :)

AndyScherzinger:
pinging @mario for 2nd review :)

mario:
Approved

PR #1826, opened 2017-12-02
Share via link use display name

tobiasKaminsky:
If you share a file via link it is said "tobi shared a file", but my displayname is "Tobias" and thus "Tobias shared a file" is better as other users / contacts do need not to know anything about an internal user id and maybe are even confused what it is.

 --> 

Thanks @nickvergessen for pointing this out


AndyScherzinger:
👍 looks good!

PR #1828, opened 2017-12-04
Change changelog for dev version to master commits as we do not have …

tobiasKaminsky:
Change changelog for dev version to master commits as we do not have any special changelog anylonger


AndyScherzinger:
Doesn't that mean we basically don't have a dev changelog anymore? Since the changelog is usually only up-to-date right before a release starting with RC1... :/

tobiasKaminsky:
Well, it links to the commit history of master, so it should be always up to date...
The real changelog.md is independent from that.

AndyScherzinger:
👍 right! ....Argh... misread the URL...

mario:
👍

PR #1830, opened 2017-12-04
Get rid of gallery API and use internal one

tobiasKaminsky:
So we do not have any dependency other to server and thus resized images will work for all those servers where image previews are enabled 🎉

### DIFF
--- a/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -68,6 +68,7 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.lang.ref.WeakReference;
+import java.net.URLEncoder;
 import java.util.ArrayList;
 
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
Is this supported for _all_ Nc server versions? Your change does suggest this... ;)

tobiasKaminsky:
According to @rullzer available in NC11/12 and upcoming 13.
As we additionally have a fallback if the response is no image, I think we are fine here.

AndyScherzinger:
So basically Nc11+ support. Else we fallback to what? (just out of curiosity)

rullzer:
Officially we only support current and current -1. So this means right now NC 11 and NC 12. Once 13 is out NC 12 and NC 13.

Just check if you get back a status code of 200. If not. then Fall back to the old way of getting the file via webdav I guess.

tobiasKaminsky:
Fallback is currently an error image and "No resized image available. Download full image?" as snackbar.
The idea is that no full sized images are downloaded without user interaction.

mario:
👍 

@AndyScherzinger  ^_^

AndyScherzinger:
👍

PR #1831, opened 2017-12-04
Better handling when switching to an account with disabled external links

tobiasKaminsky:
To test:
Account A with external links
Account B w/o external links

Switch  between them and see that A always has links, whereas B never has them.

Thanks to @rullzer for pointing this out 🎉

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -1335,53 +1335,50 @@ public abstract class DrawerActivity extends ToolbarActivity implements DisplayU
      */
     public void fetchExternalLinks(final boolean force) {
         if (getBaseContext().getResources().getBoolean(R.bool.show_external_links)) {
-            Thread t = new Thread(new Runnable() {
-                public void run() {
-                    // fetch capabilities as early as possible
-                    if ((getCapabilities() == null || getCapabilities().getAccountName().isEmpty())
-                            && getStorageManager() != null) {
-                        GetCapabilitiesOperarion getCapabilities = new GetCapabilitiesOperarion();
-                        getCapabilities.execute(getStorageManager(), getBaseContext());
-                    }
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -1335,53 +1335,50 @@ public void onAccountRemovedEvent(AccountRemovedEvent event) {
...
-                                runOnUiThread(() -> updateExternalLinksInDrawer());
+                            for (ExternalLink link : externalLinks) {
+                                externalLinksProvider.storeExternalLink(link);
                             }
-                        } else {
-                            sharedPreferences.edit().putInt(EXTERNAL_LINKS_COUNT, count + 1).apply();
                         }
                     } else {
-                        Log_OC.d("ExternalLinks", "links disabled");
+                        sharedPreferences.edit().putInt(EXTERNAL_LINKS_COUNT, count + 1).apply();
```
AndyScherzinger: I might be mistaken, but don't we have a "PreferencesDao" with static methods to do read/write for preferences?

tobiasKaminsky: We have a ArbitrayDataProvider now, but if I remember correctly this was introduced afterwards

AndyScherzinger: Ah, okay 👍

mario: This will at some point have to go into ArbitraryDataProvider.

mario:
👍 

@AndyScherzinger ^_^

AndyScherzinger:
👍 done :)

src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
AndyScherzinger: merged, issue created for this: #1850

PR #1834, opened 2017-12-05
File content provider: if selection and selectionArgs are null, let them pass

### LINKED ISSUE #1842: Unable to Upload file...
### Actual behaviour
-Tell us what happens
When try to Upload a file I give connection Error... 
Whit Google Store app version Upload file work correct... 
### Environment data
Android version:
7.1.2
Device model: 
Sm-t805

Nextcloud app version:
2.0.0 build 20000099, latest commit here 2 day ago
Nextcloud server version:
12.0.4

tobiasKaminsky:
Fix #1842

Otherwise we get an exception

### DIFF
--- a/src/main/java/com/owncloud/android/providers/FileContentProvider.java
+++ b/src/main/java/com/owncloud/android/providers/FileContentProvider.java
@@ -592,7 +592,8 @@ public class FileContentProvider extends ContentProvider {
             sqlQuery.setProjectionMap(projectionMap);
         }
 
-        if (selectionArgs == null) {
+        // if both are null, let them pass to query
+        if (selectionArgs == null && selection != null) {
             selectionArgs = new String[]{selection};
             selection = "(?)";
         }


src/main/java/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -592,7 +592,8 @@ private Cursor query(
             sqlQuery.setProjectionMap(projectionMap);
         }
 
-        if (selectionArgs == null) {
+        // if both are null, let them pass to query
+        if (selectionArgs == null && selection != null) {
```
AndyScherzinger: If ```selection``` is not ```null``` why would we set it to ```selection = "(?)";``` then? ;)

tobiasKaminsky: When doing it the way with (?) and passing the args to it we get automatically prevention of sql injection.

tobiasKaminsky:
@mario @AndyScherzinger we need this soon in, otherwise dev version will not upload anything.

src/main/java/com/owncloud/android/providers/FileContentProvider.java
AndyScherzinger: 👍

AndyScherzinger:
👍 fine by me

mario:
👍

PR #1836, opened 2017-12-05
Overhauled data storage management

### LINKED ISSUE #1707: Push doesn't work
Unfortunately for majority of users automatic storage location detection doesn't work. As a consequence generating the nc-keypair directory and key pair doesn't work either.

Manually selecting one of the available locations fixes this, but the location should be detected automatically or we'll never have some things working (like push here!).

Only a few of our users receive push notification, and this is bad :(

please try to look into it.

### LINKED ISSUE #1833: nc-keypair files on /sdcard
I notice that the Nextcloud app stores a keypair on the /sdcard partition under Nextcloud/nc-keypair/. The two files listed are push_key.priv and push_key.pub. 

Is there any good reason for a keypair to be stored on the world-readable/writable /sdcard partition? Why isnt these files at least stored in the app's private data folder? I havent looked through the source to find out what the keypair is used for but this doesnt look good. I can see they keyfiles are generated in PushUtils.java and are seemingly used to push some sort of registration token to the OwnCloud server.

I highly recommend storing these somewhere else. I'd be happy to provide a patch for this. Since the keys are seemingly only used for a single operation it would be trivial to simply store them in a more appropriate location and either keep them there or delete them after use since it doesnt seem like they are used more than once.

ardevd:
Storing user data and RSA keys on the root of the external storage is not a good idea. The data storage system has been overhauled here with the following changes.

- The `STORAGE_PATH` user preference is now set to the [private internal app storage directory](https://developer.android.com/guide/topics/data/data-storage.html#filesInternal) by default.
- The DataStorageProvider's getAvailableStoragePoints method now returns the [private internal data directory](https://developer.android.com/guide/topics/data/data-storage.html#filesInternal) and, if available, the [private external storage directory](https://developer.android.com/guide/topics/data/data-storage.html#filesExternal).
- `PushUtils` will now store and handle the RSA keypair in the [private internal storage directory](https://developer.android.com/guide/topics/data/data-storage.html#filesInternal). The keys need to be here since even the private external storage directory is accessible to everyone with the `READ_EXTERNAL_STORAGE` permission. Therefore we specify the internal storage directory for these instead of using the `STORAGE_PATH` preference. 

Please ignore 8ca99c6. I updated this as intended in d4670ad

This fixes #1833
fix #1707

### DIFF
--- a/src/gplay/java/com/owncloud/android/utils/PushUtils.java
+++ b/src/gplay/java/com/owncloud/android/utils/PushUtils.java
@@ -98,7 +98,7 @@ public class PushUtils {
     }
     
     private static int generateRsa2048KeyPair() {
-        String keyPath = MainApp.getStoragePath() + File.separator + MainApp.getDataFolder() + File.separator
+        String keyPath = MainApp.getAppContext().getFilesDir().getAbsolutePath() + File.separator + MainApp.getDataFolder() + File.separator
                 + KEYPAIR_FOLDER;
 
         String privateKeyPath = keyPath + File.separator + KEYPAIR_FILE_NAME + KEYPAIR_PRIV_EXTENSION;
@@ -270,7 +270,7 @@ public class PushUtils {
     }
 
     public static Key readKeyFromFile(boolean readPublicKey) {
-        String keyPath = MainApp.getStoragePath() + File.separator + MainApp.getDataFolder() + File.separator
+        String keyPath = MainApp.getAppContext().getFilesDir().getAbsolutePath() + File.separator + MainApp.getDataFolder() + File.separator
                 + KEYPAIR_FOLDER;
 
         String privateKeyPath = keyPath + File.separator + KEYPAIR_FILE_NAME + KEYPAIR_PRIV_EXTENSION;
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -111,8 +111,8 @@ public class MainApp extends MultiDexApplication {
 
         SharedPreferences appPrefs =
                 PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
-        MainApp.storagePath = appPrefs.getString(Preferences.PreferenceKeys.STORAGE_PATH, Environment.
-                getExternalStorageDirectory().getAbsolutePath());
+        MainApp.storagePath = appPrefs.getString(Preferences.PreferenceKeys.STORAGE_PATH,
+                Environment.getExternalStorageDirectory().getAbsolutePath());
 
         boolean isSamlAuth = AUTH_ON.equals(getString(R.string.auth_method_saml_web_sso));
 
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -112,7 +112,7 @@ public class MainApp extends MultiDexApplication {
         SharedPreferences appPrefs =
                 PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
         MainApp.storagePath = appPrefs.getString(Preferences.PreferenceKeys.STORAGE_PATH,
-                Environment.getExternalStorageDirectory().getAbsolutePath());
+                getApplicationContext().getFilesDir().getAbsolutePath());
 
         boolean isSamlAuth = AUTH_ON.equals(getString(R.string.auth_method_saml_web_sso));
 
--- a/src/main/java/com/owncloud/android/datamodel/OCFile.java
+++ b/src/main/java/com/owncloud/android/datamodel/OCFile.java
@@ -28,7 +28,9 @@ import android.content.Context;
 import android.net.Uri;
 import android.os.Parcel;
 import android.os.Parcelable;
+import android.support.v4.content.FileProvider;
 
+import com.owncloud.android.R;
 import com.owncloud.android.lib.common.network.WebdavUtils;
 import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.utils.MimeType;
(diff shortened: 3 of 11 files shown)


mario:
What about:

- existing keys
- existing storage issues

Will those be automagically migrated to proper values?

ardevd:
Remember, without this pull request the keys are stored in whatever directory the user has selected in settings (the STORAGE_PATH preference) which by default is set to the root of the external storage /sdcard. So when the user updated the setting the keys would be moved just like all the other files I think. It does seem like if you manually delete any of the keyfiles they wont be re-created automatically, but that's a different issue. Really, the keypair should be stored in the Android Keystore and not on the filesystem but at least with my pull request the key files are no longer world readable/writable.

What existing storage issues are you referring to?

mario:
Refering to an issue where the storage shown on the setting screen is NONE of the choices you actually have when you click on it.

Storing keys in the keystore is indeed good, but there are so many horor stories of stuff being deleted from keystore for random reasons ... 
https://doridori.github.io/android-security-the-forgetful-keystore/#sthash.vnLYU7pJ.sHQNLh4Z.dpbs

Thanks for the PR, we'll try to review it swiftly cc @AndyScherzinger @tobiasKaminsky

ardevd:
@mario I have completely redone the way the storage paths shown in the setting are generated. With this PR we now present two alternatives to the user. The private internal storage directory which the Android developer documentation confirms will always be available and the external storage directory which might be unavailable due to the nature of external storage. The storage provider now verifies that the external storage is in fact available before presenting it as an option to the user.

Since internal storage is more reliable and fewer ODM's now bother with support for removable storage (sdcards) the internal storage is now used by default with this PR which might by itself fix a lot of the storage related issues you are describing.

ardevd:
I've done basic testing myself and things seem to work well, but more extensive testing should obviously be conducted. Existing app installations have already set the STORAGE_PATH value so this PR would not overwrite that with the new default value. The upside is that the user wont be affected negatively by the change, the downside being that this means most users will continue to store all their user data in the root of the external storage directory. Which is bad. 

One potential issue is that the RSA keypair directory will be set to the internal storage directory and we dont explicitly move the existing key files.

ardevd:
Update: I realize now that if we store user files on the internal storage we need to use a [File Provider](https://developer.android.com/reference/android/support/v4/content/FileProvider.html) to share the files with other applications. This is not an issue we use the external storage option. I assumed the app already utilized a File Provider but this is not the case. I'll see if I can add support for it.

ardevd:
Updated the FileProvider implementation. Will now properly generate URI's for use when sharing files with other applications.

mario:
Have you seen the comment of the method in OCFile.java you've changed? :) Unfortunately not all apps understand the URIs we get through the File provider and thus you can't write to the file. Try with Microsoft Word as an example. (on both internal and external storages).

mario:
Do you think it's possible that we explicitly move the keys to the proper internal location?
(I will of course check and test the PR)

ardevd:
@mario I read it but I've never had issues with FileProvider URIs. However I just tested with MS Word and it opens the file just fine but considers the file ReadOnly. This is seemingly a bug with MS Office as there are several stackoverflow posts related to MS Office in particular.

It works fine with Google Docs however.

This begs the question whether we should sacrifice security for compatibility with buggy apps. 

Regarding the key files I simply don't know enough about what they are used for to give the a good recommendation. Can we simply regenerate the keys if they are not found?

mario:
@ardevd  we can regenerate the keys, but you need to unsubscribe from both nextcloud & proxy first for each of the accounts, and then subscribe again. 

Any progress on the Office front?

ardevd:
@mario There should be a system for regenerating keys anyway. As an experiment I deleted the keys from external storage (with the stock Nextcloud app) and it didn't seem like the app took any mitigating course of action. I haven't really looked through the relevant code and figured out how to best solve this but regardless of where key files are stored there should be some contingency plan should they keys be missing. 

I've done some testing with MS Word specifically and it seems like the app simple doesn't handle FileProvider URLs properly. I've tried reaching out to MS about it but no response so far. I'll try again though.

mario:
@ardevd permission to push to your PR?  I can add key regeneration, and you look into Word issue.

ardevd:
@mario of course! Please do.

ardevd:
Regarding the issue with Word the only sensible conclusion at this point is that Microsoft is handling the intent incorrectly. See [here](https://stackoverflow.com/questions/40157389/when-using-android-file-provider-files-dont-have-correct-permissions-despite-f) and [here](https://stackoverflow.com/questions/42252933/xamarin-forms-android-fileprovider-grantwriteuripermission-not-always-working).

Furthermore, I did some testing on my device. Downloading a .doc document locally to my phone and opening the file using Google's own [Files Go app](https://play.google.com/store/apps/details?id=com.google.android.apps.nbu.files) results in the exact same behavior with MS Word. The file is opened read-only. Actually, opening the file at all without any 3rd party app results in the same issue. I'm therefore not sure how much we can or even should do about this other than alerting Microsoft.

mario:
Can I suggest that we do a special handling for office files if MS office
is installed, and not use content provider for them for now?

Can you do that?

ardevd:
Unfortunately, I dont think this would work. As long as files are stored in the private data folder for the app on internal storage **you have to** use the FileProvider to share those files with other apps.

ardevd:
Quick update: I've been running the app with my PR for a while now and it seems to be working well. Opening videos especially is a lot faster and more reliable for some reason. 

MS Word will still open files in read-only mode though. Again this is the case even if you use Google's own filemanager so it's really up to MS to fix that bug.

AndyScherzinger:
> MS Word will still open files in read-only mode though. Again this is the case even if you use Google's own filemanager so it's really up to MS to fix that bug.

@mario but haven't you fixed that and so this PR would break it? Just asking :)

ardevd:
@AndyScherzinger You are correct. However, should we continue to store user files all over the world readable external storage just so we can work around an issue with MS Word or should we store files privately and only share files with third party applications explicitly. This PR also fixes the issues with storage location setting. 

It's up to you guys. I don't want to expose my user data to the whole world which is why I authored this PR. And I'm also fan of doing things correctly and follow Android's best practices rather than compromise on user privacy just to make things work with a buggy app.

AndyScherzinger:
@ardevd I totally support your PR and having the storage location management fixed which is what this PR does 🥇 So to explicitely state it: I am not disapproving this PR if that is the impression I might have given!

mario:
@ardevd still here?

mario:
I have fixed two bugs:

- uploads not working
- "All" Office files will now open in write mode if on external storage, otherwise read-only - I think this is a good compromise between security and usability.

Two issues remain:

- switching from internal to external (and vice versa) doesn't migrate files and fileobservers like it used to
- what happens when you have bunch of files on the old app and install the new version? Doesn't seem like any migration happens :(

mario:
@ardevd if you can look into the two issues I found tomorrow, we can merge this soon!

ardevd:
@mario I dont know the codebase well enough to comment on file migration. I was sure I tested toggling back and forth between internal and external storage and that all files were migrated but I'll double check. 

We certainly have to figure out how to handle app update behavior in this case. As it stands nothing will happen when a user upgrades by default I believe. The path for files is stored in a SharedPreference which is not modified unless the user updates the settings. 

I'll look into this tonight if I have time and see what I can figure out.

mario:
@ardevd happy to help, but with current state:

- the path gets updated once you updated the app from master to this branch, but files don't get migrated
- when you switch storage location on this branch files are not migrated :(

ardevd:
The build I'm running was from when I initially created the pull request. I just did the following:

- Set the user file storage path to /data/user/0/com.nextcloud.client/files in settings.
- Downloaded a couple of files
- Changed storage path to /storage/emulated/0/Android/data/com.nextcloud.client/files 
- The previously downloaded files have now been moved to the new path. 

So if that no longer works then it's either an issue that was introduced elsewhere or it's some other odd bug.

When users update from the old app I am pretty sure their old storage path is used until they decide to change it themselves? Could you reference where in the code we update the Preferences.PreferenceKeys.STORAGE_PATH from whatever the user previously specified?

mario:
@ardevd will try to investigate on Friday, if you can't do it sooner :-/

ardevd:
@mario Sounds good. Again, it works fine on my device.

mario:
@ardevd improved and rebased version is here - https://github.com/nextcloud/android/pull/1954. Thank you for the awesome contribution!

PR #1838, opened 2017-12-07
Add activity previews

tobiasKaminsky:
Integrate https://github.com/nextcloud/android-library/pull/73

### DIFF
--- a/src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java
+++ b/src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java
@@ -101,7 +101,8 @@ public class ActivityListAdapter extends RecyclerView.Adapter<RecyclerView.ViewH
         this.mClient = client;
         mValues.clear();
         String sTime = "";
-        for (Object o : activityItems) {
+        List<Activity> activityList = (List<Activity>) activityItems.get(0);
+        for (Object o : activityList) {
             Activity activity = (Activity) o;
             String time;
             if (activity.getDatetime() != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java
```diff
@@ -101,7 +101,8 @@ public void setActivityItems(List<Object> activityItems, OwnCloudClient client)
         this.mClient = client;
         mValues.clear();
         String sTime = "";
-        for (Object o : activityItems) {
+        List<Activity> activityList = (List<Activity>) activityItems.get(0);
+        for (Object o : activityList) {
```
AndyScherzinger: erm, wouldn't make sense to then use Activity right away since you casted to a List of Activities already? ;)

tobiasKaminsky: It was a really quick fix to get it working.
But currently I am unsure what the https://github.com/nextcloud/android-library/pull/73 pr is doing at all... ;-)

AndyScherzinger: We should also find the root cause of #1848 since that reported bug would still happen here and cause a crash of the app...

AndyScherzinger: Or not, I am not sure at this time of the day... too late at night...

tobiasKaminsky:
Postponed until clarification of library PR.

#1848 should be independent of this.

PR #1843, opened 2017-12-10
Only retry jobs if we cn do it

mario:
Should in part fix flashing notification. Untested so be kind :P

Keep in mind the method will still be invoked on network and power saving changes.

cc @jospoortvliet 


AndyScherzinger:
👍 code looks good to me while the PR needs a rebase to the latest master to resolve the droneCI build error due to the translation changes that are on master now. cc @tobiasKaminsky

tobiasKaminsky:
👍 rebased for merge (translation error should be fixed by now)

mario:
Awesome to have this merged so fast!

PR #1844, opened 2017-12-11
CRC32 magic

mario:
Further hardens file changes detection.

### DIFF
--- a/src/main/java/com/owncloud/android/datamodel/FileSystemDataSet.java
+++ b/src/main/java/com/owncloud/android/datamodel/FileSystemDataSet.java
@@ -20,6 +20,8 @@
  */
 package com.owncloud.android.datamodel;
 
+import android.support.annotation.Nullable;
+
 /*
     Model for filesystem data from the database
  */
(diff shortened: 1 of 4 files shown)


mario:
Dunno what's up with lint, pls check.

AndyScherzinger:
👍 code looks good to me while the PR needs a rebase to the latest master to resolve the droneCI build error due to the translation changes that are on master now. cc @tobiasKaminsky

mario:
@AndyScherzinger  can you rebase pls?

AndyScherzinger:
@mario rebased :)

tobiasKaminsky:
On E2E branch I included md5sum, having two different checksums is maybe too much?
Is there a benefit of crc over md5sum or vice vera? 

On your code you add crc to each file, but I cannot see where it is read/used?

mario:
@tobiasKaminsky it depends on the use case. I'm happy with MD5 for encryption branch, but CRC32 is faster and good enough(tm) for this use case, that's why I opted to do this.

As for how it works, check lines after:
 +                if (data.getCrc32() == null || (newCrc32 != -1 && !data.getCrc32().equals(Long.toString(newCrc32)))) {

tobiasKaminsky:
👍 
(I over read setting FILESYSTEM_FILE_SENT_FOR_UPLOAD)

Then let us keep both. md5sum is only used in unit tests.

mario:
Thanks for the quick review guys!

mario:
@jasonbayton  once this is in daily build, please test if remove/insert SD card still triggers reupload.

jasonbayton:
woop woop will do!

PR #1846, opened 2017-12-11
Fix walledGarden with resized images

tobiasKaminsky:
- moved check to thread, otherwise "not on main thread exception"
- show preview (resized image) if file is not downloaded
- check for file only if it is downloaded (prevents crash on clicking on an non-downloaded image)


AndyScherzinger:
👍 fine by me 😃

mario:
Approved

PR #1853, opened 2017-12-12
Dev version: use always latest library

tobiasKaminsky:
I expect lint to fail as warnings will increase, which I will then adjust

This was changed sometime back to 1.0.33 which is wrong.
For dev version we want to have the latest client commit alongside with the latest library, which is always "-SNAPSHOT"


AndyScherzinger:
Which means before doing a release we need to create the branch and change the lib version number?

tobiasKaminsky:
No, this is only valid if the build flavor is "versionDev" due to "versionDevImplementation".
These deps are only included if the flavor matches.

implementation 'com.github.nextcloud:android-library:1.0.33' -> on all flavors except versionDev
versionDevImplementation 'com.github.nextcloud:android-library:master-SNAPSHOT' -> only for versionDev

AndyScherzinger:
👍 fine by me

AndyScherzinger:
Pinging @mario for second review :)

mario:
👍

PR #1857, opened 2017-12-14
Suppress extra translation warning except on /values/strings.xml

tobiasKaminsky:
When removing a source string we always got an error at lint.
This suppresses the error as the next transifex will automatically remove the unneeded strings.

If we add a new language, we have to adapt the list.


AndyScherzinger:
👍 sweet! :)

mario:
👍

PR #1872, opened 2017-12-19
Moved spanish for 419/us to values-es-rUS

AndyScherzinger:
This is only a temporary hotfix as transifex will move/re-add the wrong folder tomorrow.
@tobiasKaminsky  already asked transifex how to proceed.


tobiasKaminsky:
Better solution: https://github.com/nextcloud/android/pull/1873

Issue #1874, opened 2017-12-19
Errorr: Invalid resource directory name


abhi472:
Project is not able to be build,
whenever i try to do so i get the error
the error is pointing towards this particular file
<Path to  android studio project folder>NextCloudApp/src/main/res/values-es_419

What i figured out here is that the issue is with resource resolution folder for latin american - spanish translation.

but according to the stackoverflow answer [here](https://stackoverflow.com/a/42282526/6166726)
we should rename the values folder to values-es-b+es+419.

i would like to create a PR for this issue if it's passed as valid.

tobiasKaminsky:
This was fixed just recently: https://github.com/nextcloud/android/pull/1873
Can you just do a pull and check if it is working?

abhi472:
This seems to be working, but as mentioned in the stackoverflow thread,
```to speed up resolution and improve maintainability, you should store resources in the most common parent dialect. For example, if you were storing Spanish resources in the values-es-rUS directory before, move them into the values-b+es+419 directory, which contains Latin American Spanish.```

I think using value-b+es+419 is a better idea

tobiasKaminsky:
There is android 7 and above explicitly mentioned. As we support back to api 15, I am unsure if this works for older devices.

tobiasKaminsky:
> /drone/src/github.com/nextcloud/android/src/main/res/values-es-b+es+419: Error: Invalid resource directory name
mergeGenericDebugResources FAILED

Android also seems to not like this :/

PR #1875, opened 2017-12-19
changing values-es-rUS to values-b+es+419

abhi472:
as mentioned above i am changing the values files
by the time of my first commit the values-es_419 file was causing build error
to resolve this i went through this stack overflow thread with this [answer](https://stackoverflow.com/a/42282526/6166726)
i have mentioned this in #1874


tobiasKaminsky:
Although this fix was not merged, if you want to contribute, then you are more than welcome. Just pick an interesting issue and talk with us about a possible solution and then solve it 👍

abhi472:
yeah, thanks for the concern, :)
although this build failed due wrong resource qualifier name(it should've been values-b+es+419 and not values-es-b+es+419),
i did that in hurry!
i creating a new pr for this issue

PR #1876, opened 2017-12-19
renaming values-es-rUS to values-b+es+419

abhi472:
i have tested the resource qualifier name for android version 16 device, although i do not have any sdk15 devices to test this on, but this change was introduced in api level 24 and it is working with version 16 so i think version 15 devices will also comply with it.
this pr is in reference to closed pr #1874


tobiasKaminsky:
We are using Transifex for translating the default language (values) to all other languages, where once a day all values- files are renewed if necessary.

For the mapping language -> values-folder we use 
https://github.com/nextcloud/android/blob/master/.tx/config#L9

So the permanent solution is to change "es_419: es-rUS" to "es_419: b+es+419".

> https://developer.android.com/guide/topics/resources/multilingual-support.html

suggest some other changes, like "fr-FR" -> "fr", where I am not sure about the impact.

abhi472:
So should i create a new pr with .tx and other locale changes(if necessary)too?

abhi472:
@tobiasKaminsky i have updated the pr with a commit renaming values-en-rGB -> values-b+en+001
in accordance to the same changes which led to the change of es_419 file.

mario:
@abhi472 would you be so kind to do a language-related PR according to https://developer.android.com/guide/topics/resources/multilingual-support.html for the following repository: https://github.com/nextcloud/talk-android

tobiasKaminsky:
> So should i create a new pr with .tx and other locale changes(if necessary)too?

Yes, it is sufficient to change it in .tx/config. 
Thank you very much for your help 🎉

abhi472:
>Yes, it is sufficient to change it in .tx/config.
Thank you very much for your help 🎉

>So the permanent solution is to change "es_419: es-rUS" to "es_419: b+es+419".


@tobiasKaminsky for the .tx/config change i would also need to know what change should i add for e-rGB file change

>@abhi472 would you be so kind to do a language-related PR according to https://developer.android.com/guide/topics/resources/multilingual-support.html for the following repository: https://github.com/nextcloud/talk-android

already working on it, will add a pr as soon as this pr is resolved, so that i would know exactly what changes would be required in that repo.

mario:
Thanks!

mario:
@abhi472 if the docs say values-b+en+001 is better, do that!

abhi472:
okay so i am gonna change en_GB: en-rGB -> en_GB: b+en+001 in .tx/config

abhi472:
guys, i have made all necessary changes and updated the pr,
once this is merged i will start the pr for talk-android 🎉

mario:
👍 

@tobiasKaminsky ^_^

abhi472:
this may come as silly but i just wanted to know by when this pr will be merged, if it takes too much time it may end up as a dead pr

AndyScherzinger:
👍 Hi @abhi472 there are no silly questions and I just approved and merged your PR 🎉 

I also can assure you PRs won't end up dead since then me, Tobias or Mario will take care of rebasing to keep it alive and up-to-date in scenarios where other PRs might need to get merged first.

AndyScherzinger:
Thanks for your contribution @abhi472 ❤️

PR #1887, opened 2017-12-21
Oreo support

### LINKED ISSUE #1174: Android O
Android O is introducing again some limits on background tasks:
https://developer.android.com/preview/behavior-changes.html#back-all

For 2.1 we need to evaluate this.

### LINKED ISSUE #1746: Adaptiv Icon
Hello,
is it possible to change the icon in an adaptive icon for new android os?
example: https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive.html

thanks a lot

### LINKED ISSUE #1886: Oreo: round app icon?
As you can see our icon is now very small within the circle.
I would like to make it bigger and fill the circle just like the phone app next to our app.

EDIT: as you can see on the right screenshot it seems to be possible to still have a regular icon, like Fdroid.
What do you prefer?

@nextcloud/designers?

tobiasKaminsky:
* [x]  bumbs sdk versions
* [x]  notification channels
* [x]  @AndyScherzinger will include round icon launcher #1886 

fixes #1746
fixes #1886 
fixes #1174


tobiasKaminsky:
Great @AndyScherzinger 
One small thing. I know this is according to the specs as far as I remember, but we discussed to ignore this and make the background/icon as big as possible, or?

image

AndyScherzinger:
I won't... The icons are created via Android Studio's helpers and I am not willing to put in enourmous efforts just to not make them Android guideline complaint, so if you, @jancborchardt or anyone else is willing to do so, please feel free.

Plus in my personal opinion we don't need to ignore this but need to stop fiddling around with overly visually-complex backgrounds...

mario:
Also voting here to completely remove the dots in the background and make it solid.

mario:
Few things regarding new lines in files, but like I said, cannot approve this background, sorry.

jancborchardt:
@AndyScherzinger ok, using the standard size is fine. But can we force the round shape wherever possible (yes, except where OEMs style it)?

AndyScherzinger:
@jancborchardt sure, we just need to reference the round legacy icon for the square shape too 😁

src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -231,6 +236,75 @@ public static void initAutoUpload() {
...
+    public static void notificationChannels() {
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.O && getAppContext() != null) {
+            Context context = getAppContext();
+            NotificationManager notificationManager = (NotificationManager)
+                    context.getSystemService(Context.NOTIFICATION_SERVICE);
+
+            if (notificationManager != null) {
+                if (notificationManager.getNotificationChannel(
+                        NotificationUtils.NOTIFICATION_CHANNEL_DOWNLOAD) == null) {
+                    CharSequence name = context.getString(R.string.notification_channel_download_name);
```
AndyScherzinger: @tobiasKaminsky all notification channel setups do the exact same thing except for the values, so this should be put into a separate method and just being called with the different setup values for each channel :)

AndyScherzinger:
> Few things regarding new lines in files, but like I said, cannot approve this background, sorry.

I fixed all line breaks that I found in the svg source files :) As for the background I'll leave it to a decision to be found in #1886 with the latest input from @jancborchardt being implemented on this branch already, not wanting to polute the PR thread to much with this matter but keeping it in the issue thread. Hope that is fine for everyone involved.

tobiasKaminsky:
@AndyScherzinger lint is complaining

> The launcher icon shape should use a distinct silhouette
../../../src/main/res/mipmap-hdpi/ic_launcher_background.png: ￼Launcher icons should not fill every pixel of their square region; see the design guide for details

> ../../../src/main/res/drawable/ic_launcher_foreground.xml:28: Attribute fillType is only used in API level 24 and higher (current min is 14)

>This check finds attributes set in XML files that were introduced in a version newer than the oldest version targeted by your application (with the minSdkVersion attribute).
This is not an error; the application will simply ignore the attribute. However, if the attribute is important to the appearance or functionality of your application, you should consider finding an alternative way to achieve the same result with only available attributes, and then you can optionally create a copy of the layout in a layout-vNN folder which will be used on API NN or higher where you can take advantage of the newer attribute.

According to the explanation above the solution would be to put it in -vNN (if this is possible)

src/main/java/com/owncloud/android/MainApp.java
tobiasKaminsky: fixed 👍

AndyScherzinger:
I'm going to have a look at it after x-mas :)

xXSTrikeXx:
I wish you good holidays and a well slide to the next year ;)

AndyScherzinger:
@tobiasKaminsky I rebased the PR and fixed anything I could find while the build stil broke due to lint since lint thinks the original score is 204 warnings, with the PR having 205 warning while the master actually has 207 warnings... :/ Let's see if the rebasing helps...

AndyScherzinger:
@tobiasKaminsky drone still insists on lint score to be 204... ?!

AndyScherzinger:
@jancborchardt @tobiasKaminsky fixed the icon to now always be "round" except for Android 8+ where it will be OEM specific (see screenshot from my OP3T where round is the default 😸)

------

**Legacy icon for anything below Android Oreo**
2017-12-27 12_36_33-oreo support by tobiaskaminsky pull request 1887 nextcloud_android

**Homescreen**
device-2017-12-27-123133

**Apps list**
device-2017-12-27-123157

jancborchardt:
Looking sharp! :+1: :)

AndyScherzinger:
cc @mario for second review. The launcher icon is now according to the specification of @jancborchardt with the enlarged, lighter dots in the background. For Android Oreo the adaptive icon definition for foreground and background uses xml vectors only.

cc @MorrisJobke for branding... With this PR we add the adaptive icon (app icon) which needs a foreground and a background "drawable". In this case it is always an Android xml drawable. So as a first step the brander should maybe remove the adaptive icon definitions in order to not ship the Nc icon on branded clients (since the brander doesn't override these files yet). If you want to support adaptive icons you'd basically would have to replace ```src/main/res/drawable-v26/ic_launcher_background.xml``` and ```src/main/res/drawable-v26/ic_launcher_foreground.xml```

mario:
Like I mentioned, anything before Oreo should use square, with exception of 25 which should use roundIcon. After Oreo, it's adaptive, so the manufacturer/whatever decides. Round icon pre-Oreo ALWAYS does not work, since for example Samsung on Android 7.0 puts a grey square around it and that looks very ugly.

As for the dots, I don't like them, but can't do much if you don't see the issue.

mario:
Can you please set roundIcon and then I'll approve? Thanks @tobiasKaminsky

tobiasKaminsky:
@mario done.
What about your report that the round icon is not working on certain devices?

mario:
Rebase and we can merge I guess?

MariusBluem:
@marinofaggiana and @Ivansss for the iOS Client icons :)

PR #1889, opened 2017-12-21
Tint correctly on very first login

### LINKED ISSUE #1856: After first login: no new files
On current master it happens for me that after a fresh installation and new login I do not see any files.
Instead I have to refresh/swipe manually.

tobiasKaminsky:
fix #1856 

Please note that this change result in a slightly more often tinting usage as refreshListOfFilesFragment() than onCreateView(), but I do not see another solution.


mario:
👍

PR #1894, opened 2017-12-27
Walled garden: 204 response: length is -1 (unknown) or 0 (null)

### LINKED ISSUE #1880: DEV App: cannot use the Android 'share' function to upload files
### Actual behaviour
I cannot use the Android 'share' function to upload files with the DEV version (from F-Droid). It starts to upload, reports upload error in status bar. File is in "to be uploaded"-section of the app, Re-upload doesn't change anything - error again. Works with PlayStore Beta App on the same server.

### Expected behaviour
Should upload as the beta/release versions did.
 
### Steps to reproduce
1. Open Gallery on your Android device
2. Select a photo (or screenshot in my case) and "share" it (Android-"Share" not Nextcloud's)
3. Select Nextcloud as "share"-Method
4. Confirm PIN/Fingerprint if configured
5. Switch account if necessary
6. Select target directory and press "Upload" button

### Environment data
Android version: 8.0
Device model: OnePlus 3T
Stock or customized system: Stock (OOS Beta)
Nextcloud app version: 20171212 (F-Droid)
Nextcloud server version: 13b3

#### Nextcloud log: No entry while trying to upload file.

tobiasKaminsky:
Fix #1880 
We only checked if length is -1, but it can also be 0.
As this is a bug fix, and prevents uploading file on NC13, it needs to get into next release.

### DIFF
--- a/src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
+++ b/src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
@@ -68,7 +68,8 @@ public class ConnectivityUtils {
                     int status = client.executeMethod(get);
                     
                     if (serverVersion.compareTo(OwnCloudVersion.nextcloud_13) > 0) {
-                        return !(status == 204 && get.getResponseContentLength() == -1);
+                        return !(status == 204 &&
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
:+1:

mario:
👍

PR #1905, opened 2018-01-02
mFiles should be never null

tobiasKaminsky:
getFiles() is for whatever reason null and therefore leads to a crash.
(more than 700x)

https://github.com/nextcloud/android/blob/dc2d7edc56fde6db6ee0e3e378900c8eff408d7a/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java#L660-L660

With this change mFile is never null, but may be empty, which should be fine.


mario:
👍

tobiasKaminsky:

PR #1906, opened 2018-01-02
Get notification manager if it does not exist

tobiasKaminsky:
Reported on google developer console.
Somehow notification manager is null, so trying to get a fresh one.


mario:
👍

tobiasKaminsky:

PR #1912, opened 2018-01-03
Notification activity: account can be null

tobiasKaminsky:
Found on google dev console.

Though I cannot reproduce it, but 
https://github.com/nextcloud/android/blob/6820bf1dba3c65d722e4c0642e10b15f0c8602e5/src/main/java/com/owncloud/android/authentication/AccountUtils.java#L58-L58
can return null, so it is safer to catch this.


mario:
👍

tobiasKaminsky:

PR #1913, opened 2018-01-03
Receive external files

tobiasKaminsky:
extraText can be null
getWindow can be null
action bar can be null
getType() or getAction on intent can be null

code enhancements


mario:
👍

tobiasKaminsky:

PR #1915, opened 2018-01-03
Auto upload without permission

tobiasKaminsky:
Steps to reproduce
- remove storage permission
- open up auto upload

I know it is weird to do that, but google dev console shows that it happened...
(also tested it with API21 to make sure it is working if permissions are only in manifest)


mario:
Who gave the users choice? xD Anyway, looks all good, but I hope users aren't that silly usually.

👍

tobiasKaminsky:
We ask on initial setup for the storage permission, but as with API21 you can revoke any permission on any time within the app settings...

mario:
I know, I was joking :P

PR #1922, opened 2018-01-04
Activity scrolling bug fix

tobiasKaminsky:
if nextPageURL is empty/null do not load any more items, otherwise yo…u get the first items again
if pageURL is null (when manually refreshing), then clear list

@nickvergessen


tobiasKaminsky:
@mario please review 👍

mario:
👍

tobiasKaminsky:

PR #1928, opened 2018-01-08
Full text search

tobiasKaminsky:
WIP to integrate FNS


tobiasKaminsky:
Out of date, if we want to fully support FNS, then we can still have a look into this for inspiration.

PR #1930, opened 2018-01-08
Fix crashing search while copy/move

### LINKED ISSUE #1800: Search in move/copy dialog crashes
- move / copy file
- search something
- app is crashing

The search result is tried to be shown in FileDisplayActivity as this is the normal way to show server side search results.

tobiasKaminsky:
fix #1800

Remaining: when searching in copy/move only folders should be shown, as files cannot be a target for copy/move
@mario as you implemented the search: can I limit the scope to folders only easily?

**NOTE: in order to test/compile it is currently linked against master branch of library. This must then be changed prior merge**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -227,8 +227,7 @@ dependencies {
     implementation 'com.github.bumptech.glide:glide:3.7.0'
     implementation 'com.caverock:androidsvg:1.2.1'
     implementation "com.android.support:support-annotations:${supportLibraryVersion}"
-
-    compile 'com.google.code.gson:gson:2.8.2'
+    implementation 'com.google.code.gson:gson:2.8.2'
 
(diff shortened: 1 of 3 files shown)


mario:
You'd need to add an argument in SearchOperation, and add (or don't add) in the where block.
<d:eq>
	<d:is-collection/>
</d:eq>

tobiasKaminsky:
It only works without <d:eq>, but now ready for review.
Please note that it needs a change to the lib and therefore I added this branch to build.gradle, which needs to be reverted prior merging.

AndyScherzinger:
👍 didn’t have the chance to test yet but look fine code wise

tobiasKaminsky:
Mario approved on IRC.

PR #1932, opened 2018-01-08
Contact backup removal

tobiasKaminsky:
Remove contacts backup job on account removal

If we do not do this this will result in a NPE...

### DIFF
--- a/src/main/java/com/owncloud/android/jobs/AccountRemovalJob.java
+++ b/src/main/java/com/owncloud/android/jobs/AccountRemovalJob.java
@@ -68,11 +68,9 @@ public class AccountRemovalJob extends Job implements AccountManagerCallback<Boo
         Account account = AccountUtils.getOwnCloudAccountByName(context, bundle.getString(ACCOUNT, ""));
         AccountManager am = (AccountManager) context.getSystemService(ACCOUNT_SERVICE);
 
-        if (account == null || am == null) {
-            return Result.FAILURE;
-        } else {
+        if (account != null && am != null) {
             // disable contact backup job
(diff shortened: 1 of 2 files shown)


src/main/java/com/owncloud/android/jobs/AccountRemovalJob.java
```diff
@@ -65,9 +66,14 @@ protected Result onRunJob(Params params) {
         Context context = MainApp.getAppContext();
         PersistableBundleCompat bundle = params.getExtras();
         Account account = AccountUtils.getOwnCloudAccountByName(context, bundle.getString(ACCOUNT, ""));
+        AccountManager am = (AccountManager) context.getSystemService(ACCOUNT_SERVICE);
+
+        if (account == null || am == null) {
+            return Result.FAILURE;
+        } else {
+            // disable contact backup job
+            ContactsPreferenceActivity.cancelPreviousContactBackupJobForAccount(context, account);
```
mario: I think you should use cancelContactBackupJobForAccount to cancel possibly running manual job too.

ContactsPreferenceActivity.cancelContactBackupJobForAccount

tobiasKaminsky: You are right, cancelContactBackupJobForAccount is not limited to "isPeriodic()", so it should be sufficient to use cancelContactBackupJobForAccount to get all jobs cancelled that belong to the account, right?

mario: In this case, yes :)

PR #1946, opened 2018-01-10
Add notifications on sync failure

mario:
In addition, I've hardened documents provider a bit as it's very fragile. Still doesn't always work for me, but it's something for a start at least...

### DIFF
--- a/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
+++ b/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
@@ -176,6 +176,9 @@ public class FileOperationsHelper {
                 } else {
                     if (file.isDown()) {
                         FileStorageUtils.checkIfFileFinishedSaving(file);
+                        if (!result.isSuccess()) {
+                            DisplayUtils.showSnackMessage(mFileActivity, R.string.file_not_synced);
+                        }
                         EventBus.getDefault().post(new SyncEventFinished(intent));
                     }
                 }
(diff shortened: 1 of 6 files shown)


tobiasKaminsky:
It works :-)
Though I have some questions/enhancement ideas:
- when opening an outdated text file a snackbar is shown, but relative short. Maybe set it permanent?
- when opening an image it is only shown "no network connection"
- when opening it via document provider "sync failed" is shown via notification, but it is not that obvious. On IRC you said we can use a toast. Maybe this is then too intrunsive, but at least more obvious to the user?

mario:
a) Well, it's set to LONG. Permanent - unsure.
b) Isn't that good enough, since after all, you know it's not newest if you have no connection? xD
c) That's because I don't want to interrupt the workflow. Unsure - let me know if you have a better idea.

tobiasKaminsky:
a) ok, I thought that it is still to short for such a long text
b) I like the other text more as it really explains the problem/behaviour. But this will introduce more complexity as it is still possible to have "no network" if you have not downloaded the image yet.
c) I just assumed that the regular user first sees the outdated file and is confused and do not pay attention to the notification. As the toast is only shown on "error" I still think it is a good way.

mario:
@tobiasKaminsky Gimme some time.

mario:
@tobiasKaminsky should all be fixed.

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -178,6 +179,11 @@ public void run() {
                         FileStorageUtils.checkIfFileFinishedSaving(file);
                         if (!result.isSuccess()) {
                             DisplayUtils.showSnackMessage(mFileActivity, R.string.file_not_synced);
+                            try {
+                                Thread.sleep(3000);
```
tobiasKaminsky: Isn't this "freezing" the whole app?
If snackbar duration is already set to "length_long" I guess I am fine with this, instead of adding such "hard" sleep mechanism.

tobiasKaminsky:
👍

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
tobiasKaminsky: You can decide if you want to revert the sleep or merge (after rebase).

mario: Well, we're showing the loading dialog anyway so you can't do anything. And length of snackbar would be irrelevant if I didn't do this since it'd open the file right away.

PR #1947, opened 2018-01-10
Fix external storage permission

mario:

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -52,9 +52,7 @@
 
     <!-- WRITE_EXTERNAL_STORAGE may be enabled or disabled by the user after installation in
         API >= 23; the app needs to handle this -->
-    <uses-permission
-        android:maxSdkVersion="22"
-        android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
 
     <!-- Next permissions are always approved in installation time, the apps needs to do nothing special in runtime -->
     <uses-permission android:name="android.permission.INTERNET" />
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -32,10 +32,8 @@
         android:name="android.permission.GET_ACCOUNTS" />
 
     <uses-permission
-        android:maxSdkVersion="22"
         android:name="android.permission.READ_CONTACTS" />
     <uses-permission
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
With this contacts backup is not working as the app does not have the permission to read/write contacts.
Although the permissions are asked during runtime on API >22 we still need them in AndroidManifest.

https://github.com/nextcloud/android/blob/61cbe88bd056ec1ec2f3933a8175cdc6a4252c6c/src/main/java/com/owncloud/android/ui/fragment/contactsbackup/ContactsBackupFragment.java#L372-L372

Doc says:
> Requests permissions to be granted to this application. These permissions must be requested in your manifest, ...

So we can keep maxSDK for USE_CREDENTIALS, MANAGE_ACCOUNTS, AUTHENTICATE_ACCOUNTS as these get removed on API 23.

But GET_ACCOUNTS, READ_CONTACTS, WRITE_CONTACTS needs to be in Manifest without SdkVersion as far as I understand.

mario:
I have removed it from READ and WRITE CONTACTS now. But are we sure we need to remove it from GET_ACCOUNTS too?

tobiasKaminsky:
https://developer.android.com/reference/android/Manifest.permission.html#GET_ACCOUNTS

our code says:
> GET_ACCOUNTS is needed for API <= 22.
        For API >= 23 results in the addition of CONTACTS group to the list of permissions that may be
        dynamically disabled or enabled by the user after installation; but it is not important,
        since GET_ACCOUNTS is an special case, the permission is not really needed to access accounts
        owned by the app, our use case.

Doc says:
> Allows access to the list of accounts in the Accounts Service.
>
> Note: Beginning with Android 6.0 (API level 23), if an app shares the signature of the authenticator that manages an account, it does not need "GET_ACCOUNTS" permission to read information about that account. On Android 5.1 and lower, all apps need "GET_ACCOUNTS" permission to read information about any account.

Not sure. It seems to work even with maxSDK.

mario:
Yup, because we put in READ and WRITE contacts manually. I'm fine with this now, since we do not read accounts of other apps.

tobiasKaminsky:
👍 
Then let us rebase and merge it this way :-)

tobiasKaminsky:
Please wait for Codacy and Drone.
It can happen that due to rebase some problems arise and thus we should wait for those checks.
(I know that waiting for Drone is sometimes pain ;-) )

mario:
I know, was too fast. I am waiting on other approved PRs :)

PR #1954, opened 2018-01-11
New, overhauled data storage management

mario:
Improved & rebased version of the awesome community PR found here:
https://github.com/nextcloud/android/pull/1836

We don't do files migration automatically yet, I'd suggest we implement that for 2.2 since it needs some UI as well, but push keys are properly migrated if they exist.


src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -236,6 +236,31 @@ public static void initAutoUpload() {
...
+        Context context = getAppContext();
+
+        if (!PreferenceManager.getKeysMigration(context)) {
+            String oldKeyPath = getStoragePath() + File.separator + MainApp.getDataFolder() + File.separator + "nc-keypair";
+            File oldPrivateKey = new File(oldKeyPath + File.separator + "push_key.priv");
+            File oldPublicKey = new File(oldKeyPath + File.separator + "push_key.pub");
+
+            String keyPath = getAppContext().getFilesDir().getAbsolutePath() +
+                    File.separator + MainApp.getDataFolder() + File.separator + "nc-keypair";
+            File privateKey = new File(keyPath + File.separator + "push_key.priv");
```
przybylski: Equivalent of: ``` File privateKey = new File(keyPath, "push_key.priv");```
same for the line up and down

src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -236,6 +236,31 @@ public static void initAutoUpload() {
...
+            File publicKey = new File(keyPath + File.separator + "push_key.pub");
+
+            if (oldPrivateKey.exists() && oldPublicKey.exists()) {
+                if (oldPrivateKey.renameTo(privateKey) && oldPublicKey.renameTo(publicKey)) {
```
przybylski: This doesn't handle the scenario when first rename succeeds and other fails.

src/main/java/com/owncloud/android/datastorage/providers/SystemDefaultStoragePointProvider.java
```diff
@@ -44,16 +41,9 @@ public boolean canProvideStoragePoints() {
...
-        } else {
-            path = getExternalStorageDirectory();
-        }
-
-        if (path != null && path.canWrite()) {
-            result.add(new StoragePoint(defaultStringDesc, path.getAbsolutePath()));
-        }
+        // Add private internal storage data directory.
+        result.add(new StoragePoint(defaultStringDesc,
+                MainApp.getAppContext().getFilesDir().getAbsolutePath()));
```
przybylski: I am not a fan of this storage point using context. IMO there should be a method in `MainApp` which directly returns the files dir and use that.

mario:
Feel free to improve directly in the branch?:)

src/main/java/com/owncloud/android/MainApp.java
mario: Would you be so kind to improve in the branch directly all things you think need improving? :)

src/main/java/com/owncloud/android/datastorage/providers/SystemDefaultStoragePointProvider.java
mario: I see we use this at a couple other places, so not worried about this much.

mario:
Fixed everything @przybylski except for the context thing which I'm unsure if we should fix.

src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -236,6 +236,37 @@ public static void initAutoUpload() {
...
+            String keyPath = getAppContext().getFilesDir().getAbsolutePath() +
+                    File.separator + MainApp.getDataFolder() + File.separator + "nc-keypair";
+            File privateKey = new File(keyPath, "push_key.priv");
+            File publicKey = new File(keyPath, "push_key.pub");
+
+            if ((privateKey.exists() && publicKey.exists()) || (!oldPrivateKey.exists() && !oldPublicKey.exists())) {
+                PreferenceManager.setKeysMigration(context, true);
+            } else {
+                if (oldPrivateKey.exists()) {
+                    oldPrivateKey.renameTo(privateKey);
```
tobiasKaminsky: Is renameTo working if used across filesystem boundary, e.g. from external sdcard to internal storage? As far as I remember this does not work, but we have to copy and delete instead.

src/main/java/com/owncloud/android/datastorage/DataStorageProvider.java
```diff
@@ -63,18 +62,14 @@ private DataStorageProvider() {}
...
-                    mCachedStoragePoints.addAll(p.getAvailableStoragePoint());
-                }
-            }
+        // Add internal storage directory
+        mCachedStoragePoints.add(new StoragePoint(MainApp.getAppContext().getFilesDir().getAbsolutePath(),
+                MainApp.getAppContext().getFilesDir().getAbsolutePath()));
+
+        // Add external storage directory if available.
+        if (isExternalStorageWritable()) {
+            mCachedStoragePoints.add(new StoragePoint(MainApp.getAppContext().getExternalFilesDir(null).getAbsolutePath(),
```
tobiasKaminsky: "getExternalFilesDir(null)" can be null: 
> May return null if shared storage is not currently available.

So this should be checked, otherwise we get a NPE

src/main/java/com/owncloud/android/datastorage/providers/SystemDefaultStoragePointProvider.java
tobiasKaminsky: Would indeed be better to centralize this, but I agree with @mario that this does not need to be done now.

src/main/java/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -624,8 +624,8 @@ private Cursor query(
 
         // if both are null, let them pass to query
         if (selectionArgs == null && selection != null) {
-            selectionArgs = new String[]{selection};
             selection = "(?)";
+            selectionArgs = new String[]{selection};
```
tobiasKaminsky: This was done to prevent sql injections: instead of querying the args directly those are inserted via "(?):
> select * where test = " + $foo; (foo can be dangerous)
becomes to
> "select * where test = (?)", $foo --> then we have use of built in sql prevention.

Changing this, would mean that selectionArgs is always "(?)" which is not what we want to have: if args is null, then we want to "force" selection to use "(?)":
selection:  "1=1"
--> select * from xy where "1=1" 
becomes
--> select * from xy where (?), "1" and this sql injection save

So this must be reverted

mario: But then the Uploads view crashes for me?

src/main/java/com/owncloud/android/MainApp.java
mario: Will investigate.

src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -236,6 +236,37 @@ public static void initAutoUpload() {
...
+                }
+
+                if (oldPublicKey.exists()) {
+                    oldPublicKey.renameTo(publicKey);
+                }
+
+                if (privateKey.exists() && publicKey.exists()) {
```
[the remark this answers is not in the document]
mario: Not always. Maybe just one file was created :)

src/main/java/com/owncloud/android/providers/FileContentProvider.java
tobiasKaminsky: When? With which uploads? Do you have a stacktrace?
This is for quite a time in dev version and I never experienced a crash on upload view (except of the TimSort exception).
Or it is a new thing with this branch...

tobiasKaminsky:
There is another severe bug:
- fresh installation
- open auto upload and enable any folder
- close auto upload view
- re-open: folder is not enabled and also upload does not work

(works on master)

mario:
I will definitely take a look, but what about the Uploads? That is crashing for me (sometimes) on master, and my change fixes it.

src/main/java/com/owncloud/android/providers/FileContentProvider.java
mario: No, was a problem with master. The Uploads view. But no stacktrace :(

tobiasKaminsky:
As said, give me a stacktrace/way to reproduce.
I think that the fix reverts sql injection protection...

mario:
@tobiasKaminsky ok, will revert the fix for now and we'll see in RC :)

mario:
@tobiasKaminsky just tried, works. so this PR is ready.

tobiasKaminsky:
Will test it with real device with external storage on monday.

src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -236,6 +243,53 @@ public static void initAutoUpload() {
         }
     }
 
+    public static void migratePushKeys() {
```
tobiasKaminsky: This is only use in PushUtils on gplay flavor, so this can be moved there, or?

mario: Maybe. :) Or just leave it here and don't worry :D

mario:
👍

src/main/java/com/owncloud/android/MainApp.java
AndyScherzinger: I guess it should be moved to not have the dependency in the f-droid build where this might be an issue then, isn't it?

AndyScherzinger:
👍 olny one minor comment about the f-droid build and the PushUtils dependency: https://github.com/nextcloud/android/pull/1954#discussion_r161483950

src/main/java/com/owncloud/android/MainApp.java
tobiasKaminsky: I moved it already :-)

tobiasKaminsky:
Welcome back @AndyScherzinger 🎉

src/gplay/java/com/owncloud/android/utils/PushUtils.java
```diff
@@ -335,4 +336,44 @@ private static int saveKeyToFile(Key key, String path) {
 
         return -1;
     }
+
+    public static void migratePushKeys() {
```
mario: this can be private void migratePushKeys() then :P

src/main/java/com/owncloud/android/MainApp.java
mario: Wouldn't be a problem because it's just moving files, no Google Play deps 📦

src/modified/java/com/owncloud/android/utils/PushUtils.java
```diff
@@ -99,8 +99,9 @@ public static String bytesToHex(byte[] bytes) {
     }
 
     private static int generateRsa2048KeyPair() {
-        String keyPath = MainApp.getStoragePath() + File.separator + MainApp.getDataFolder() + File.separator
-                + KEYPAIR_FOLDER;
+        MainApp.migratePushKeys();
```
mario: Also this won't work if you remove it from the MainApp :P

src/modified/java/com/owncloud/android/utils/PushUtils.java
```diff
@@ -330,4 +331,8 @@ private static int saveKeyToFile(Key key, String path) {
 
         return -1;
     }
+
+    private static void migratePushKeys() {
```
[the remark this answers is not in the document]
AndyScherzinger: This codacy issue seems true --> empty method.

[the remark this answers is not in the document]
tobiasKaminsky: Yeah, but modified will be removed #1966 as we do not need it anymore.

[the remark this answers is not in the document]
AndyScherzinger: 👍 alright! :)

PR #1962, opened 2018-01-15
Detect top of encryption folders

tobiasKaminsky:
As of subfolder detection I introduced a bug where the top of encrypted folders was not shown properly.


mario:
👍

PR #1978, opened 2018-01-17
harden notificationManager

tobiasKaminsky:
From Google dev console:
```
java.lang.NullPointerException: 
  at com.owncloud.android.files.services.FileDownloader.notifyDownloadResult (FileDownloader.java:578)
  at com.owncloud.android.files.services.FileDownloader.downloadFile (FileDownloader.java:470)
  at com.owncloud.android.files.services.FileDownloader.access$500 (FileDownloader.java:73)
  at com.owncloud.android.files.services.FileDownloader$ServiceHandler.handleMessage (FileDownloader.java:404)
  at android.os.Handler.dispatchMessage (Handler.java:105)
  at android.os.Looper.loop (Looper.java:164)
  at android.os.HandlerThread.run (HandlerThread.java:65)
```

This tries to get a notification manager if it is null.
It is not a real fix, but more a work around as I do not get why this can be null.

### DIFF
--- a/src/main/java/com/owncloud/android/files/services/FileDownloader.java
+++ b/src/main/java/com/owncloud/android/files/services/FileDownloader.java
@@ -566,6 +566,10 @@ public class FileDownloader extends Service
             String text = String.format(getString(R.string.downloader_download_in_progress_content), percent, fileName);
             mNotificationBuilder.setContentText(text);
 
+            if (mNotificationManager == null) {
+                mNotificationManager = (NotificationManager) getSystemService(NOTIFICATION_SERVICE);
+            }
+            
             if (mNotificationManager != null) {
                 mNotificationManager.notify(R.string.downloader_download_in_progress_ticker,
                         mNotificationBuilder.build());
@@ -623,9 +627,8 @@ public class FileDownloader extends Service
             } else {
                 // TODO put something smart in showDetailsIntent
                 Intent showDetailsIntent = new Intent();
-                mNotificationBuilder
-                        .setContentIntent(PendingIntent.getActivity(
-                                this, (int) System.currentTimeMillis(), showDetailsIntent, 0));
+                mNotificationBuilder.setContentIntent(PendingIntent.getActivity(this, (int) System.currentTimeMillis(),
+                        showDetailsIntent, 0));
             }
 
             mNotificationBuilder.setContentText(ErrorMessageAdapter.getErrorCauseMessage(downloadResult,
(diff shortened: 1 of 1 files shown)


mario:
I think this might potentially also fail since it's strange that it doesn't exist.

tobiasKaminsky:
Therefore I put a null check on every manager.
So first, it tries to get a new manager, if this fails at least no NPE -> crash should occur, but there will be no notification.
If you have a better way, I would appreciate it. As said, it is only to workaround the crash :/

src/main/java/com/owncloud/android/files/services/FileDownloader.java
```diff
@@ -561,7 +565,11 @@ public void onTransferProgress(long progressRate, long totalTransferredSoFar,
             String fileName = filePath.substring(filePath.lastIndexOf(FileUtils.PATH_SEPARATOR) + 1);
             String text = String.format(getString(R.string.downloader_download_in_progress_content), percent, fileName);
             mNotificationBuilder.setContentText(text);
-            mNotificationManager.notify(R.string.downloader_download_in_progress_ticker, mNotificationBuilder.build());
+
+            if (mNotificationManager != null) {
```
AndyScherzinger: shouldn't you then add the ```mNotificationManager == null``` check/instantiation here too?

src/main/java/com/owncloud/android/files/services/FileDownloader.java
```diff
@@ -617,21 +628,19 @@ private void notifyDownloadResult(DownloadFileOperation download,
...
-                // Sleep 2 seconds, so show the notification before remove it
-                NotificationUtils.cancelWithDelay(
-                        mNotificationManager,
-                        R.string.downloader_download_succeeded_ticker,
-                        2000);
-            }
+            mNotificationBuilder.setContentText(ErrorMessageAdapter.getErrorCauseMessage(downloadResult,
+                    download, getResources()));
+
+            if (mNotificationManager != null) {
```
AndyScherzinger: shouldn't you then add the ```mNotificationManager == null``` check/instantiation here too?

AndyScherzinger:
Just two cases where the check doesn't contain an assignment of the var with a manager.
I also can't tell if it helps (besides NPE prevention), but the errors get reported and so this might be worth a try (since we are in RC phase anyways).

src/main/java/com/owncloud/android/files/services/FileDownloader.java
tobiasKaminsky: Fixed.

src/main/java/com/owncloud/android/files/services/FileDownloader.java
tobiasKaminsky: On line 590 I am already trying to get a new one. 
I assume that it is enough to try it once in a method.

AndyScherzinger: 👍

AndyScherzinger:
👍 fine giving this a try in the RCs

mario:
👍

PR #1983, opened 2018-01-17
Resized images: snackbar fix

tobiasKaminsky:
Snackbar fix IAE: as the image generation is done in a thread it can happen that mMultiView for Snackbar no longer exists, and then it crashes.

Download only if activity exists, else show only snackbar without download option.
(I cannot reproduce a non-existing activity, but google dev console shows it happen some time)


AndyScherzinger:
👍 fine by me

mario:
👍

AndyScherzinger:
👍 reapproval thanks to pullapprove...

PR #1997, opened 2018-01-18
Internet walled: catch any exception

tobiasKaminsky:
This happens if getMethod returns null as response body.
I think this means an error and thus we can catch it and return true (=> internet walled).

```
java.lang.NullPointerException: 
  at org.json.JSONTokener.nextCleanInternal (JSONTokener.java:116)
  at org.json.JSONTokener.nextValue (JSONTokener.java:94)
  at org.json.JSONObject.<init> (JSONObject.java:155)
  at org.json.JSONObject.<init> (JSONObject.java:172)
  at com.owncloud.android.utils.ConnectivityUtils.isInternetWalled (ConnectivityUtils.java:80)
 ```


AndyScherzinger:
👍 fine by me

mario:
👍

Issue #1088, opened 2017-06-13
"Storage path" setting is too technical


jancborchardt:
Right now the storage path has values like "/data/emulated/0" which is suuper technical.

Instead the choice should be for "Data storage" and the choices could be "Internal storage", "SD card" and if there are multiple cards also "SD card 2" etc, each saying how much free space there is on them.

tobiasKaminsky:
:+1:

AndyScherzinger:
👍

tobiasKaminsky:
What do you think?

Ref: https://github.com/nextcloud/android/issues/802#issuecomment-309372150
> but we should in the end offer both /data and /media because now we are reverting back to pre 1.4 so users that wanted to block media indexing will get mad. So the best would be to offer both with the hint for (SD card - media indexed / SD card - not media indexed).

![fileqqc6ml](https://user-images.githubusercontent.com/5836855/27287508-360f6c98-5504-11e7-917c-edb2d535327d.png)

jancborchardt:
Even as slightly technical user, I don't understand what "media indexed" means. Does it mean "Don't show data in other apps, like Gallery"?

If so, we could put that as a toggle in the bottom of that popup (off by default). But right now this is equally as confusing.

AndyScherzinger:
>  I don't understand what "media indexed" means. Does it mean "Don't show data in other apps, like Gallery"?

Exactly and yes it is not easy to understand, a switch seems to be the nicer way to handle it 👍 
@tobiasKaminsky that might make this a bit more complicated since you need to have a custom dialog then due to the fact that is is not a simple array chooser anymore.

jasonbayton:
Not defending this behaviour but it's a pretty common thing to see. Few apps mask the underlying paths that I've seen.

jancborchardt:
That's where we do things differently to make technology simpler to understand and use. :)

AndyScherzinger:
@tobiasKaminsky do you already have an implementation for the screenshot you posted?

tobiasKaminsky:
It was something local, but I cannot find it, maybe I did not even saved it.. :-/

jancborchardt:
Could we just rename the current display from the technical path to "Internal storage" and "SD card" as mentioned in the original post? ;) The whole »media indexed« thing is a new feature rather than a bugfix and it blocks this small papercut issue.

AndyScherzinger:
[...]

PR #1998, opened 2018-01-18
Fix storage stuff

### LINKED ISSUE #1989: Missing Files in Android
### Actual behaviour
- in NC 3.0.0 RC1 the given save  path Changes, so the Client does a complete Resync and nie i have duplicates which does Not Show Up in Pictures for exampel
### Expected behaviour
-no Change in save path after Update
 
### Steps to reproduce
1. 
2. 
3. 

### Environment data
Android version: 7

Device model: Samsung Galaxy S7edge

Stock or customized system:stock

Nextcloud app version:3.0.0 RC1

Nextcloud server version:13.0.0 Beta 3

#### Web server error log
```
```

```
```

mario:
This should fix it, but please test along with keys migration.

Fix #1989

### DIFF
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -221,9 +221,7 @@ public class MainApp extends MultiDexApplication {
             String storagePath = appPrefs.getString(Preferences.PreferenceKeys.STORAGE_PATH, "");
             if (TextUtils.isEmpty(storagePath)) {
                 if (appPrefs.getInt(WhatsNewActivity.KEY_LAST_SEEN_VERSION_CODE, 0) != 0) {
-                /*
-                    We already used the app, but no storage is set - fix that! :)
-                 */
+                    // We already used the app, but no storage is set - fix that!
                     appPrefs.edit().putString(Preferences.PreferenceKeys.STORAGE_PATH,
                             Environment.getExternalStorageDirectory().getAbsolutePath()).commit();
                     appPrefs.edit().remove(PreferenceManager.PREF__KEYS_MIGRATION).commit();
@@ -271,7 +269,8 @@ public class MainApp extends MultiDexApplication {
         updateAutoUploadEntries();
 
         if (getAppContext() != null) {
-            if (PermissionUtil.checkSelfPermission(getAppContext(), Manifest.permission.WRITE_EXTERNAL_STORAGE)) {
+            if (PermissionUtil.checkSelfPermission(getAppContext(),
+                    Manifest.permission.WRITE_EXTERNAL_STORAGE)) {
                 splitOutAutoUploadEntries();
             } else {
                 PreferenceManager.setAutoUploadSplitEntries(getAppContext(), true);
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -52,8 +52,6 @@ import com.owncloud.android.datamodel.MediaProvider;
 import com.owncloud.android.datamodel.SyncedFolder;
 import com.owncloud.android.datamodel.SyncedFolderProvider;
 import com.owncloud.android.datamodel.ThumbnailsCacheManager;
-import com.owncloud.android.datastorage.DataStorageProvider;
-import com.owncloud.android.datastorage.StoragePoint;
 import com.owncloud.android.db.PreferenceManager;
 import com.owncloud.android.jobs.NCJobCreator;
 import com.owncloud.android.lib.common.OwnCloudClientManagerFactory;
@@ -220,15 +218,8 @@ public class MainApp extends MultiDexApplication {
                 appPrefs.getInt(WhatsNewActivity.KEY_LAST_SEEN_VERSION_CODE, 0) != 0) {
             String storagePath = appPrefs.getString(Preferences.PreferenceKeys.STORAGE_PATH, "");
             if (TextUtils.isEmpty(storagePath)) {
-                storagePath = Environment.getExternalStorageDirectory().getAbsolutePath();
-                StoragePoint[] storagePoints = DataStorageProvider.getInstance().getLegacyAvailableStoragePoints();
-                for(StoragePoint storagePoint : storagePoints) {
-                    if (storagePoint.getPath().startsWith(storagePath)) {
-                        storagePath = storagePoint.getPath();
-                        break;
-                    }
-                }
-                appPrefs.edit().putString(Preferences.PreferenceKeys.STORAGE_PATH, storagePath).commit();
+                appPrefs.edit().putString(Preferences.PreferenceKeys.STORAGE_PATH,
+                        Environment.getExternalStorageDirectory().getAbsolutePath()).commit();
                 appPrefs.edit().remove(PreferenceManager.PREF__KEYS_MIGRATION).commit();
             }
             PreferenceManager.setStoragePathFix(this, true);
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -103,8 +103,6 @@ public class MainApp extends MultiDexApplication {
 
     private static boolean mOnlyOnDevice = false;
 
-    private static final String KEY_LAST_SEEN_VERSION_CODE = "lastSeenVersionCode";
-
     private SharedPreferences appPrefs;
     @SuppressWarnings("unused")
     private boolean mBound;
@@ -219,11 +217,11 @@ public class MainApp extends MultiDexApplication {
 
     private void fixStoragePath() {
         if (!PreferenceManager.getStoragePathFix(this) &&
-                appPrefs.getInt(KEY_LAST_SEEN_VERSION_CODE, 0) != 0) {
+                appPrefs.getInt(WhatsNewActivity.KEY_LAST_SEEN_VERSION_CODE, 0) != 0) {
             String storagePath = appPrefs.getString(Preferences.PreferenceKeys.STORAGE_PATH, "");
             if (TextUtils.isEmpty(storagePath)) {
                 storagePath = Environment.getExternalStorageDirectory().getAbsolutePath();
-                StoragePoint[] storagePoints = DataStorageProvider.getInstance().getAvailableStoragePoints();
+                StoragePoint[] storagePoints = DataStorageProvider.getInstance().getLegacyAvailableStoragePoints();
                 for(StoragePoint storagePoint : storagePoints) {
                     if (storagePoint.getPath().startsWith(storagePath)) {
                         storagePath = storagePoint.getPath();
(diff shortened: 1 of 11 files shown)


src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -208,6 +217,28 @@ public static void initContactsBackup() {
 
     }
 
+    private void fixStoragePath() {
+        if (!PreferenceManager.getStoragePathFix(this) &&
+                appPrefs.getInt(KEY_LAST_SEEN_VERSION_CODE, 0) != 0) {
```
tobiasKaminsky: If this is the same Key as in WhatsNewActivity please use this. 
While it should technically work as it is the same string, I was really confused where it is set as IDE only shows one usage of it. 
I had to search for "KEY_LAST_SEEN_VERSION_CODE" to see where it set.

tobiasKaminsky:
It still changes the storage path if none was set
- install 2.0.1
- do not touch storage path (pref is empty / not set)
- install this 
- storage path gets changed

tobiasKaminsky:
2.0.1: /storage/emulated/0
3.0.0: /storage/emulated/0/Android/data/com.nextcloud.client/files

But it does not matter, or? The default path must not be changed as long as it is an update.
On a fresh installation it is quite fine to have a new default path.

mario:
AAA, you're right @tobiasKaminsky. It should change it, but not in this way. Will fix the PR.

tobiasKaminsky:
https://github.com/nextcloud/android/blob/9feeaa92f3b2cb446bf011f75c458f2130f8ddda/src/main/java/com/owncloud/android/MainApp.java#L226-L226

You use "startsWith" and thus  
 /storage/emulated/0 (2.0.1) still gets changed to /storage/emulated/0/Android/data/com.nextcloud.client/files.

Again, my expectation is that the storage path must *not* change on upgrade as otherwise all downloaded files are not found anymore. Is this not correct?

tobiasKaminsky:
/data/user/0/com.nextcloud.client/files is the path for a new installation.
The problem is that this path is not readable by other apps e.g. gallery.
So if you download an image you cannot see/use it in gallery, which is now working.

mario:
@tobiasKaminsky that is absolutely fine. We give users two choices:

- internal storage - secure, private
- external storage - more exposed to the system

Choosing the first by default is a better option.

tobiasKaminsky:
Tested on my wife's device there is indeed a choice, but both storage paths are not visible to other apps, but this may also come from #1954.

As this is already RC2 I would like to not have any "breaking" changes like this in it, without discussing and thinking thoroughly about it.
Related to this is also https://github.com/nextcloud/android/issues/1088.

Can we find a way that 3.0 is behaving exactly as 2.0 in regard of storage and for 3.1 discuss how we can improve this?

tobiasKaminsky:
Thinking a bit more about this, I would suggest as in #1088 
- external storage
- internal storage
(both with a toggle "secure, private" --> only app can access files)

My experience is that many people use our app to sync/download files, e.g. download mp3s to listen them with another app, download images to see them in QuickPic/Gallery.
Also as this is our current "default" I would suggest to have "visible for all / non-private" as default and if people want to have it privately they can migrate on storage path.

tobiasKaminsky:
> internal storage - secure, private
external storage - more exposed to the system

This is a bit oversimplified I guess as there are devices that do not have any external storage and also there we should have the possibility to have files "public available".

mario:
This indeed comes from #1954. External storage is more exposed since anyone with READ EXTERNAL STORAGE can access this, so I'm unsure why say gallery wouldn't be able to see it.

External storage is not always SD card btw.

Let me think about this.

Dunno, I like it the way it is now. Thoughts @AndyScherzinger ?

tobiasKaminsky:
whatsapp image 2018-01-22 at 07 49 57 whatsapp image 2018-01-22 at 07 49 48 

So both storage paths are not public available (the device has internal & external storage).

Changing this (silently to the user) is in my opinion not a good way. (Imagine you are used to the app and install freshly on a second device and now it behaves different without any warning? (the first device does not change as it is only an upgrade).
To have it clearer to the user, we can  add it to permanent What's new screen or add a "config screen" on very first startup which lets the user choose the

mario:
I will modify the PR in the following way for 3.1:

- keep whatever you have set OR if it's a new install  use proper indexed internal path
- allow 4 (!) storage options - 2 internal (indexed & non-indexed, but without explanation), 2 external (indexed & non-indexed, but without explanation)

For 3.2, we'll add a custom dialog.

Deal?

tobiasKaminsky:
But then all new installs on 3.0 would have "private" files? So this should probably be on 3.0 already
(I can help / do it if you have no time).

> indexed & non-indexed, but without explanation)
The whole #1088 issue is about missing explanations, so I guess we need something "private vs public" is maybe enough already.

But all in all your suggestions sounds fine 👍

mario:
Sorry I meant 3.0 and 3.1 :)

tobiasKaminsky:
Then I am totally fine with your idea.
And for 3.1 we can discuss this also with Jan (he might have some ideas, I fear :-D)

AndyScherzinger:
I'd also vote for 3.1 to not have the pressure on 3.0.0 or even block the release any further except for necessary RCs :)

mario:
This is now a signficant change. Please test Android 4 and one of the other Androids (5,6 or 7).

What is needed to test:
- new installs
- updates
- push keys migration

Thanks @tobiasKaminsky and @AndyScherzinger! :)

tobiasKaminsky:
Android 4.4 emulator:

No 
```
01-22 15:38:17.602 3299-3351/com.nextcloud.client E/RegisterAccountDeviceForProxyOperation: Exception while registering device for notifications
javax.net.ssl.SSLHandshakeException: javax.net.ssl.SSLProtocolException: SSL handshake aborted: ssl=0xb963b520: Failure in SSL library, usually a protocol error
error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure (external/openssl/ssl/s23_clnt.c:741 0x98946990:0x00000000)
at com.android.org.conscrypt.OpenSSLSocketImpl.startHandshake(OpenSSLSocketImpl.java:448)
at com.owncloud.android.lib.common.network.AdvancedSslSocketFactory.verifyPeerIdentity(AdvancedSslSocketFactory.java:266)
at com.owncloud.android.lib.common.network.AdvancedSslSocketFactory.createSocket(AdvancedSslSocketFactory.java:187)
at org.apache.commons.httpclient.HttpConnection.open(HttpConnection.java:707)
at org.apache.commons.httpclient.MultiThreadedHttpConnectionManager$HttpConnectionAdapter.open(MultiThreadedHttpConnectionManager.java:1361)
at org.apache.commons.httpclient.HttpMethodDirector.executeWithRetry(HttpMethodDirector.java:387)
at org.apache.commons.httpclient.HttpMethodDirector.executeMethod(HttpMethodDirector.java:171)
at org.apache.commons.httpclient.HttpClient.executeMethod(HttpClient.java:397)
at org.apache.commons.httpclient.HttpClient.executeMethod(HttpClient.java:323)
at com.owncloud.android.lib.common.OwnCloudClient.executeMethod(OwnCloudClient.java:232)
at com.owncloud.android.lib.resources.notifications.RegisterAccountDeviceForProxyOperation.run(RegisterAccountDeviceForProxyOperation.java:84)
at com.owncloud.android.lib.common.operations.RemoteOperation.execute(RemoteOperation.java:151)
at com.owncloud.android.lib.common.operations.RemoteOperation.execute(RemoteOperation.java:155)
at com.owncloud.android.utils.PushUtils.pushRegistrationToServer(PushUtils.java:244)
at com.owncloud.android.ui.activity.FileDisplayActivity.onMessageEvent(FileDisplayActivity.java:2203)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:515)
at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:485)
at org.greenrobot.eventbus.EventBus.invokeSubscriber(EventBus.java:479)
at org.greenrobot.eventbus.BackgroundPoster.run(BackgroundPoster.java:64)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
at java.lang.Thread.run(Thread.java:841)
Caused by: javax.net.ssl.SSLProtocolException: SSL handshake aborted: ssl=0xb963b520: Failure in SSL library, usually a protocol error
error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure (external/openssl/ssl/s23_clnt.c:741 0x98946990:0x00000000)
at com.android.org.conscrypt.NativeCrypto.SSL_do_handshake(Native Method)
at com.android.org.conscrypt.OpenSSLSocketImpl.startHandshake(OpenSSLSocketImpl.java:405)
... 22 more
```

tobiasKaminsky:
4.4 emulator:
on fresh 3.0 installation: Storage path is set to /data/data/com.nextcloud.client/files and thus images are not visible in gallery

tobiasKaminsky:
Tested android 4.4.2 with a real S3 neo device:
- no push notification, same error as on emulator
- migraton to /storage/extSdCard fails (both on 2.0.1 and 3.0)
- migration to /storage/emulated/0/Android/data/com.nexcloud.client/files ("private") is working 👍

tobiasKaminsky:
@mario rebased (needed to have correct db migration) and tested.
Summary is in: https://github.com/nextcloud/android/pull/1998#issuecomment-359455688

Really amazing so far :-)

mario:
@tobiasKaminsky so what do I have to fix?

javax.net.ssl.SSLHandshakeException: javax.net.ssl.SSLProtocolException: SSL handshake aborted: ssl=0xb963b520: Failure in SSL library, usually a protocol error

Is something I can't fix since this I guess Android version doesn't support the cipher used on the server.

To sum it up - is there something I need to fix in this PR further? xD

tobiasKaminsky:
As it "only" affects 4.x and push notification I am fine with merging 👍

tobiasKaminsky:
Please rebase and fix Drone :-)

mario:
@tobiasKaminsky rebased, but where do I see what the issue with lint is?

tobiasKaminsky:
I assumed you know it, as we have this rather long:
https://drone.nextcloud.com/nextcloud/android/3205/10
> found errors: 0
found warnings: 138
previous errors: 0
previous warnings: 130

go to https://nextcloud.kaminsky.me/index.php/s/tXwtChzyqMj6I8v and compare 3205.html with master.html
(maybe some are not (easy) fixable and then just mention them here.

mario:
@tobiasKaminsky please retest everything, including 4.

mario:
As for lint: errors are related to using commit instead of apply, and I do that on purpose.

tobiasKaminsky:
Regarding lint:
https://developer.android.com/reference/android/content/SharedPreferences.Editor.html#apply():
> As SharedPreferences instances are singletons within a process, it's safe to replace any instance of commit() with apply() if you were already ignoring the return value.

@mario if you "still" want to have commit(), I will add suppressLint to the function.

mario:
@tobiasKaminsky commit is blocking. apply isn't. So yes, I'd like to keep commits :)

tobiasKaminsky:
Tested and now everything works, except of push for 4.x
👍

AndyScherzinger:
:+1: fine by me

PR #2021, opened 2018-01-22
Prevent index out of bounds exception

tobiasKaminsky:
```
mStreamsToUpload.get(0) == null
```
can throw IndexOutOfBoundsException if we pass a new ArrayList() on it.


AndyScherzinger:
But shouldn't ```mStreamsToUpload.size() == 0``` then already be true?

tobiasKaminsky:
Yes, this is what I added ;-)
If mStreamsToUpload is empty, null, size = 0, first item is null, then we receive a text file, so we can directly go there.

AndyScherzinger:
👍 Whoops, completely misread the diff, Sorry! All fine by me!

mario:
👍

PR #2025, opened 2018-01-22
Fix conditions check

mario:

### DIFF
--- a/src/main/java/com/owncloud/android/jobs/NContentObserverJob.java
+++ b/src/main/java/com/owncloud/android/jobs/NContentObserverJob.java
@@ -49,7 +49,6 @@ public class NContentObserverJob extends JobService {
                 persistableBundleCompat.putBoolean(FilesSyncJob.SKIP_CUSTOM, true);
 
                 new JobRequest.Builder(FilesSyncJob.TAG)
-                        .addExtras(persistableBundleCompat)
                         .startNow()
                         .setExtras(persistableBundleCompat)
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
A bit more context please so I can at least test what you fixing :-)

mario:
Previously wifi only and charging only was ignored

AndyScherzinger:
Just tested wifi only and it still ignored it and simply uploads via cellular :(

AndyScherzinger:
Same at home :/

@tobiasKaminsky can you please give it a test drive?

mario:
@AndyScherzinger ah sorry. Can you re-pull and try?

AndyScherzinger:
No worries, pulled and re-tried, still ignored wifi-only :(

Will be off for today now, Sorry

mario:
Sorry, missing stuff all around :( Should be fine now.

On Mon, Jan 22, 2018 at 10:25 PM, Andy Scherzinger <notifications@github.com

AndyScherzinger:
👍 
Argh, had to test ;) Works again while a remaining bug is that (at least on Android 8) uploads queue due to missing wifi connection don't get (instantly) restarted when connected to a wifi network again, but that is another issue which I mentioned somewhere else earlier.

mario:
@AndyScherzinger in theory, they should restart in max 15-20 mins. You need to have at least part of the app active in order to use receivers on Android 8.

tobiasKaminsky:
👍

PR #2028, opened 2018-01-23
Fix crash on <4.4 when opening drawer

tobiasKaminsky:
Drawback is that, depending on primary color (via theme), it may look ugly ;-)


mario:
👍

AndyScherzinger:
👍

PR #2030, opened 2018-01-23
Show infos on untrusted domain if available

tobiasKaminsky:
Needs Server: https://github.com/nextcloud/server/pull/7991
If the server pr is not available, it will fall back to "unknown error" as before.

@mario @AndyScherzinger do we want to add this in 3.0 (I know it is feature freeze) or do we want to wait for 3.1? Both is fine for me.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -182,7 +182,7 @@ dependencies {
     // dependencies for app building
     implementation 'com.android.support:multidex:1.0.2'
 //    implementation project('nextcloud-android-library')
-    implementation 'com.github.nextcloud:android-library:1.0.35'
+    implementation 'com.github.nextcloud:android-library:1.0.36'
     versionDevImplementation 'com.github.nextcloud:android-library:master-SNAPSHOT' // use always latest master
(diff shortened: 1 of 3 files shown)


AndyScherzinger:
I am fine either way too (not valuing feature freezes that much to be honest) while the server schedule says Nc14. So it won't show anyways in the near future 😆 which would point to 3.1

tobiasKaminsky:
>Maybe good to backport to stable12 and stable13 to make it easier for apps to detect the real problem.

The first PR is always against master, and then, if decided, it will backported with two new PRs ;-)

AndyScherzinger:
👍 code change is fine by me, also 3.0 or 3.1 while I like to ship things that are done, so since this is a minor change i'd be fine with just putting it into RC2... :)

AndyScherzinger:
@tobiasKaminsky btw. doesn't compile you seem to have missed to push some changes :)

mario:
FIx the build and can go in 3.0.

tobiasKaminsky:
The build is failing as the library PR as there is no new library release.

mario:
👍

PR #2037, opened 2018-01-25
Do not ignore ignoreEtags

tobiasKaminsky:
Currently we have set ignoreEtags to true on refresh and on swap directories.
This leads to a full sync of folder and its direct content (fetching all metadata), even if the etag is not changed.

For browsing, refreshing I think it is safe to rely on changing Etags as this is the whole purpose of them.
@rullzer can you imagine any situation when you (on desktop) are doing a full resync of a folder?

"ignoreEtags" was introduced with this commit message, years ago:
> Add boolean parameter 'ignoreTag' to SyncronizeFolderOperation constructor and pass ignoreTag = true in calls to #startSyncFolderOperation at FileDisplayActivity and MoveActivity in #onRefresh()

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -276,7 +276,7 @@ public class FileDisplayActivity extends HookActivity
             }
         } else {
             createMinFragments(savedInstanceState);
-            refreshList(true);
+            refreshList(false);
         }
 
         setIndeterminate(mSyncInProgress);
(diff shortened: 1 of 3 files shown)


rullzer:
This is indeed the reason for etags. If the etag is the same there is no reason to refresh. We do the same on the client. Else full syncs would take ages.

pinging @icewind1991 to confirm

icewind1991:
Correct

mario:
👍

tobiasKaminsky:
@AndyScherzinger now your turn ;-)

AndyScherzinger:
:+1:

AndyScherzinger:
@tobiasKaminsky can you rebase and merge?

PR #2038, opened 2018-01-25
Speed up sync

tobiasKaminsky:
- change order as file.getFileID() is faster than file.getRemotePath()
- enhance finding of encrypted ancestor

This reduces a fresh sync of a folder with 5k images from ~40s to 12s.


AndyScherzinger:
👍 looks good to me :)

PR #2039, opened 2018-01-25
Replace list with recycler view

tobiasKaminsky:
This changes from ListView to RecyclerView.


AndyScherzinger:
@tobiasKaminsky this is the lib @mario suggested which looks good to me looking at the feature list and what we would have to (re-) implement acttually: https://github.com/mikepenz/FastAdapter

tobiasKaminsky:
I would love to have some people to test this.
Just use it and see if something is wrong. 
Meanwhile I will finish it.
@AndyScherzinger maybe? 🎉

AndyScherzinger:
> hiding toolbar, fix layout @AndyScherzinger maybe?

What is the issue here @tobiasKaminsky is the toolbar to be hidden or not to be hidden? :)

tobiasKaminsky:
- It gets hidden once you scroll down. I am unsure if we want to have this as we do not have it anywhere else (I do not know where it comes from).
- it is overlapping the first item

tobiasKaminsky:
Please test it and report any bugs, except those that are stated in the first comment ;-)
Parallel I will resolve codacy/lint.

AndyScherzinger:
@tobiasKaminsky I can have a look and yes the bar shouldn't overlap the list while I'd say either me (or @mario?) did this on purpose and I'd even say the bar should behave that way on all lists.

What do you think @mario @przybylski ?

AndyScherzinger:
other than that I'd also say the (sticky?) headers should only be there if ther eis at least one item for that category.

src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
```diff
@@ -186,7 +186,7 @@ public UploadListAdapter(FileActivity parentActivity) {
         mUploadsStorageManager = new UploadsStorageManager(mParentActivity.getContentResolver(), parentActivity.getApplicationContext());
         mUploadGroups = new UploadGroup[3];
 
-        shouldShowHeadersForEmptySections(true);
+        shouldShowHeadersForEmptySections(false);
```
AndyScherzinger: :heart:

tobiasKaminsky:
Fixed as far as I can.
Remaining codacy issues are not new as I just refactored those methods.

tobiasKaminsky:
@mario @AndyScherzinger this is now done.
Please test it. 
I am sorry that this is such a big PR, but changing ListView to RecyclerView was not possible to do step by step...

mario:
@tobiasKaminsky same was the case with auto upload ages ago, do you remember? :P 

But I'll review.

mario:
Quick code check doesn't reveal many bad things, however:

- why didn't we use libs to speed up dev & improve maintainability?
- can we please pass List around instead of array list?
- in addition, when initializing arraylist please do List<Object> whatever = new ArrayList

Once the above is fixed (except for the lib) i'll test.

tobiasKaminsky:
What would be the purpose of List? 

I tried it with the library, but this revealed a lot of more conflicts, also with handling thumbnails if I remember correctly.

AndyScherzinger:
Imho the purpose is the classic java pattern to work with the Interface not the implementation

mario:
@tobiasKaminsky when can we expect this to be ready for review & testing according to my comments? :)

tobiasKaminsky:
I will rebase now and replace ArrayList with List.

tobiasKaminsky:
Rebase done, now replacing ArrayList with List. Done in a different commit due to easily revert/change.

tobiasKaminsky:
> in addition, when initializing arraylist please do List whatever = new ArrayList

I changed this only for those occurrences that were introduced in this PR. There are also some "old" ones, but I did not want to enhance this PR any further ;-)

--> rebased & done

AndyScherzinger:
* File list (general) seems fine
* :exclamation: "copy to..." and "move to..." list is multi selection enabled (while it shouldn't, since you are choosing a location)
* What about also having the dividers with "72dp margin left" like CityZen does now (I implemented it there) or Amaze does: 

* thumbnail fetching should maybe change to make use of glide (maybe postpone this to a new, future PR since the old implementation didn't do that either). While using glide when using the recycler view would be nice since then it'll always focus on creating the thumbnails for the list items currently displayed.

tobiasKaminsky:
Issue regarding copy/move is fixed.

Changing dividers is fine for me, but please in another PR, based on this one.
Same for glide, there is already an issue regarding this. And this should be tackled after merge this or based on this.
--> all in all I am trying to keep PRs as small as possible. With this is obviously did not worked...

AndyScherzinger:
> Issue regarding copy/move is fixed.

Did you push this already? Haven't seen it

> Changing dividers is fine for me, but please in another PR, based on this one.
Same for glide, there is already an issue regarding this. And this should be tackled after merge this or based on this.
--> all in all I am trying to keep PRs as small as possible. With this is obviously did not worked...

All fine by me and yes, I wouldn't put it into this PR either :+1:

tobiasKaminsky:
Fine with this, especially all these problems are not new, I mereley did a code move on them and thus Codacy is complaining.
It is however a good idea to tackle all Codacy stuff at some point.

mario:
Folders have a "Sync" option that downloads all files inside. Yet the file has a "Download" option instead which turns into a "Sync" option once its downloaded. Why is that?

mario:
Other than that complaint (which I'm not sure is related to this PR), this looks decent. 👍

tobiasKaminsky:
> Folders have a "Sync" option that downloads all files inside. Yet the file has a "Download" option instead which turns into a "Sync" option once its downloaded. Why is that?

AFAIK this also the current behaviour as I did not touched the overflow menu content, at least not intentionally ;-)
Sync does more than only download, if files are already present in a folder those file gets synced via SynchronizeFileOperation. 
Syncing on an empty local folder is in fact a complete download.
If you have an idea how to improve this, let us discuss this then in a separate issue.

tobiasKaminsky:
@mario pullapprove needs a :+1: as first character.
Thank you for the review.

mario:
👍

AndyScherzinger:
:+1: all fine by me, also tested the grid view as discussed.
@tobiasKaminsky can you rebase and do we merge it for 3.1? (I'd be fine with that btw)

AndyScherzinger:
:+1: restest for grid looks good! From my point of view it'S ready to go

tobiasKaminsky:
This is the latest change @mario.
I improved the way images are shown in grid view.

 
 

If you are fine with this, we can merge :+1:

PR #2048, opened 2018-01-26
Fix renaming folder

tobiasKaminsky:
Filename must only be changed on encrypted files (a folder name itself is never encrypted)

Steps to test
- rename folder
- see if folder was renamed

This must be in 3.0 as it is a regression.
I'll see if google dev console has anything else critical, otherwise we need to have a RC3 for this :/


AndyScherzinger:
👍 fine.by.me

mario:
Merge

PR #2058, opened 2018-01-28
Fix crashing issue on bootup

mario:

This should fix an issue with one of the most reported crash on the play store.


tobiasKaminsky:
Which issue?
Is there another way to solve it? As now you have a permanent notification:

 

Of course the user can disable tho notification channel, but this is then required by each single user...

mario:
When you call context.startForegroundedService (this is something you added to bootup receiver for 8.0), you also need to call startForeground which we didn't do before this PR.

This service will hopefully go away anyway eventually.

tobiasKaminsky:
👍

AndyScherzinger:
👍

PR #2067, opened 2018-01-30
getExternalFilesDir(null) can be null --> preventing NPE

tobiasKaminsky:
Reported via google dev console, only 1x, but still should be in 3.0 release.


AndyScherzinger:
👍

mario:
👍

PR #2068, opened 2018-01-30
show download snackbar only if activity exists

tobiasKaminsky:
Found via google dev console.

As the resized image fetch is done async, it can happen that the activity does not exist anylonger and so we have to check this.
It should be safe to just do not show the "download" snackbar as it user is already at another view/image.


AndyScherzinger:
👍

mario:
👍

AndyScherzinger:
@tobiasKaminsky can you rebase and rerun drone? It seems to have had a hick up last night :/

Issue #1884, opened 2017-12-20
Release 3.0 - Roadmap


tobiasKaminsky:
Todo
- [x] https://github.com/nextcloud/android/issues/1837 Show message on refreshing if there is no connection
- [x] #1856 After first login: No new files
- [x] #1174 After first login: No new files
- [x] https://github.com/nextcloud/android/issues/1242 Support for Android 8.0
	- [x] #1887 Oreo support in app
	- [x] https://github.com/nextcloud/android/issues/1886 Oreo: Round app icon?
	- [x] https://github.com/nextcloud/android-library/pull/102
- [ ] #1081 Screenshots
- [x] #1881 Screenshots
- [x] #1869 New background for NC13
- [x] E2E
	- [x] https://github.com/nextcloud/android/pull/1944
	- [x] https://github.com/nextcloud/android-library/pull/92
	- Test upgrade from
    	- [x] 2.0.0
	    - [x] All e2e versions
- [x] Android library
	- [x] Merge https://github.com/nextcloud/android-library/milestone/34
	- [x] New Android lib release
- [x] https://github.com/nextcloud/ios/issues/452
- Bug fixing

Roadmap (to discuss)
- Feature freeze (only this time, next time we have a new release cycle): 22.12.
- Only E2E (#1598) & bugfixes should be merged afterwards
- RC1 at beginning of January
- Release end of January

- [x] Bump version name and version code in **all (3)** ```AndroidManifest.xml``` files
- [x] Update ```CHANGELOG.md``` adding new features

**Test app with all supported Android version**
- Access all activities
- Auto upload
- Upload / download files/folders
- Use external SD card
- What's new slides correctly shown on first start

All tests on
  - [ ] Android 4
  - [ ] Android 5
  - [ ] Android 6
  - [ ] Android 7
  - [ ] Android 8

**RC1**
- [x] Verify that the new version number is shown in settings
- [ ] Verify that logging is enabled
- [x] Tag new version as ```rc-3.0.0-01```
- [x] Build and release to Play Store as beta release and :shipit: 
- [ ] Publish it on F-droid as optional release: Edit metadata and open new merge request on GitLab
- [ ] Verify that it is shown on F-droid (within 24h)
- [x] Publish APK (generic) to https://download.nextcloud.com/android/

**RC2**
- [x] Verify that the new version number is shown in settings
- [x] Verify that logging is enabled
- [x] Tag new version as ```rc-3.0.0-02```
[...]

Issue #150, opened 2016-07-23
Release x.y.z - Roadmap


tobiasKaminsky:
- [ ] Bump version name and version code in AndroidManifest.xml
- [ ] Update CHANGELOG.md adding new features
- [ ] RC1: tag new version as rc-x.y.z-01 (YYYY-MM-DD)
 - [ ] Verify that the new version number is shown in settings
 - [ ] Build and release to Play Store as beta release and :shipit: 
 - [ ] Publish it on f-droid as optional release: edit metadata and open new merge request on gitlab
- [ ] After two weeks time (max) release as stable and replace screenshots (see updates ones here: https://github.com/nextcloud/screenshots/tree/master/android/phone)
- [ ] Tag new version as stable-x.y.z (triggers an f-droid release within 24h)
- [ ] Build and release to Play Store as stable release and :shipit:

tobiasKaminsky:
Dummy roadmap used for all further releases.

strugee:
@tobiasKaminsky a better way to manage the changelog would be to just require that all major PRs include an entry under an "Unreleased" header. Then "update CHANGELOG.md above" becomes just "change 'Unreleased' to have a version number and date."

And, as a bonus, then people have a preview of what's already been implemented for the next release.

/cc @AndyScherzinger @przybylski

AndyScherzinger:
@strugee makes sense process wise. We also discussed this vs. the all-in-one approach. Both have their up and downsides. Pro: people have a preview of what has already been implemented for the next release (and got merged) Con: All PRs will then have a merge conflict in the changelog when updating to latest master (can be prevented by doing the changelog part right before the merge). Already implemented features for the next release can be seen anyways by checking the milestone (open/closed PRs).

What do you think @tobiasKaminsky ?

tobiasKaminsky:
For me it does not matter. 
The idea of @strugee has the advantage that we make sure that we do not forget something in the changelog.

AndyScherzinger:
If everything is properly assigned to a milestone we do have the list over there. That is what I used up to now to create the changelog entries. 
But it doesn't matter to me neither.

PR #2072, opened 2018-01-30
Auto upload: Prioritize Screenshots Folder

### LINKED ISSUE #1493: Auto upload: Screenshots folder should have priority sorting
Currently there’s two folders important for me to auto-upload: Camera and Screenshots.

Camera is correctly sorted up top. Screenshots however is all the way on the bottom between folders of album covers and such on the SD card. It should be given priority sorting and put to the top, even when not synced.

What do you think @tobiasKaminsky @AndyScherzinger?

tilosp:
Closes #1493

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
@@ -87,7 +87,7 @@ public class SyncedFoldersActivity extends FileActivity implements SyncedFolderA
         SyncedFolderPreferencesDialogFragment.OnSyncedFolderPreferenceListener {
 
     private static final String SYNCED_FOLDER_PREFERENCES_DIALOG_TAG = "SYNCED_FOLDER_PREFERENCES_DIALOG";
-    public static final String PRIORITIZED_FOLDER = "Camera";
+    public static final String[] PRIORITIZED_FOLDERS = new String[] { "Camera", "Screenshots" };
     public static final String EXTRA_SHOW_SIDEBAR = "SHOW_SIDEBAR";
 
     private static final String SCREEN_NAME = "Auto upload";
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@tilosp thanks for the PR, but it is unfortunately not that is while still easy ;) Basically the sorting needs to do a compare in case both folder names are _in_ the priority list because then you need to do a 
```
f1.getFolderName().toLowerCase(Locale.getDefault()).compareTo(
-                            f2.getFolderName().toLowerCase(Locale.getDefault()));
```
simply because it should just prioritize the special folders over the other while still internally sort all prioritized folders (since it is now a list of possible values) because also the prioritized folders should be sorted alphanumerically :)

Sorry for the complication :(

tobiasKaminsky:
As a related question: we are sure that on every language "Camera" and "Screenshot" is used for the folders?

tilosp:
@AndyScherzinger sure i will change it.

I thought it makes more sense to use the order of the array, to let the developer define the order. This would be useful if we later prioritize a folder starting with a or b and we want camera to stay at the top.

AndyScherzinger:
> I thought it makes more sense to use the order of the array, to let the developer define the order. This would be useful if we later prioritize a folder starting with a or b and we want camera to stay at the top.

Right! Haven't thought about that. So sorry again for the confusion but yes, it makes sense to leave it as it just in case a different sorting than alphanumerically is needed at some point! 👍 

> As a related question: we are sure that on every language "Camera" and "Screenshot" is used for the folders?
Don't know but I'd guess so, since your phone is German, what are the folder names on your phone? But yes, strings isn't bulletproof since any oem would have the power to change this, but then it just won't be a non-prioritized folder anymore, no harm in that.

jancborchardt:
Awesome stuff @tilosp, thank you for picking this up! :)

mario:
@tobiasKaminsky we are not and I don't think they are.

jancborchardt:
@tobiasKaminsky @mario even if, it's a good starting point and we can add/fix others as we go along, right?

mario:
Sure @jancborchardt. 👍 
@AndyScherzinger ? :)

AndyScherzinger:
Sure!
Haven't tested it yet so can't approve right away since it else might get merged by accident and @tobiasKaminsky wanted to do a RC3 release tomorrow.
Other than that, code-wise fine as already stated 👍

tobiasKaminsky:
👍 Working fine for me.
I created https://github.com/nextcloud/android/tree/stable-3.0.x which is our new stable-3.0.x branch as discussed.
So we can now go on with master and merge stuff.
If we want to do bug fixing (hopefully not that much), we have to backport this.

AndyScherzinger:
:+1: all fine by me.
For the 3.0 final we then still need to backport the translations

tobiasKaminsky:
> For the 3.0 final we then still need to backport the translations

Right, added/changed this to #1884 and #150, so we do not forget.

PR #2074, opened 2018-01-30
Fixes #2070: Move the responsibility of FAB's theming to theris owner

### LINKED ISSUE #2070: FAB doesn´t use theme color
### Actual behaviour
- After switching from files view to shared view and back the FAB doesn´t use the theme colour anymore

### Expected behaviour
- The FAB should be green
 
### Steps to reproduce
1. Switch from files view to shared view
2. Switch back to files view

### Environment data
Android version: 7

Device model: Nexus 6

Stock or customized system: Custom ROM

Nextcloud app version: Dev Version from F-Droid

Nextcloud server version: 13

theming

przybylski:
Since ExtendedListFragment is owning all FAB's where is no need for any other activity to alter the theming of those buttons. ThemeUtils are wide accessible, and at the time of construction, so apply the theming there.


AndyScherzinger:
👍 looks good cc @mario and @tobiasKaminsky for second review, since it is a bug fix for a reported issue I'd say let's put it into the next RC 😃

mario:
👍

PR #2078, opened 2018-01-31
Store info that file is encrypted

tobiasKaminsky:
Previously this info was fetched only during refresh operation.


AndyScherzinger:
:+1:

mario:
👍

PR #2079, opened 2018-01-31
Fix upgrade from 1.4.3 to 3.0.0

tobiasKaminsky:
As checkIfColumnExists is not working reliable when the same column is altered within the same onUpgrade(), we need to check the old version.
1.4.3 has versionCode 20, so if you upgrade from this, it will get all needed columns on the respective upgrade step and therefore we can ignore it in the sanity step 30.
For all other version the sanity step 30 is needed as it can happen due to beta/alpha testing, that some columns are missing.

--> next time we never ever are allowed to change upgrade once it is published

### DIFF
--- a/src/main/java/com/owncloud/android/providers/FileContentProvider.java
+++ b/src/main/java/com/owncloud/android/providers/FileContentProvider.java
@@ -1628,28 +1628,30 @@ public class FileContentProvider extends ContentProvider {
                         db.execSQL(ALTER_TABLE + ProviderTableMeta.FILE_TABLE_NAME +
                                 ADD_COLUMN + ProviderTableMeta.FILE_ENCRYPTED_NAME + " TEXT ");
                     }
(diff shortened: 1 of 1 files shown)


mario:
👍

przybylski:
Is there any reason not to protect the entire block with `if (oldVersion  > 20)` instead of adding additional check to the sequence of if's?

tobiasKaminsky:
No, I was testing it incrementally and added the check where an error occurred, but this is cleaner 👍

tobiasKaminsky:
@przybylski do you want to approve it? :-)

przybylski:
if I can than I will

👍 

[Approved with PullApprove](https://pullapprove.com/nextcloud/android/pull-request/2079/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=nextcloud/android)

PR #2084, opened 2018-02-01
Apply new theming when folder refresh finishes

przybylski:
When synchronization of the folder finishes it is possible that theming can also change. At this moment it is not possible to tell. So what we do is to re-setup parts of the UI to stay consistent with the server theming.

===

This is not even close to being elegant solution, but at the moment so many things would required to be hanged that the small pain of re-setup views doesn't sound as a bad trade-off.

The purest of the solution would require the following:
* Storing capabilities should be moved from ContentResolver to other database API, preferably Room, since it supports LiveData
* Activity should subscribe to the database by using LiveData for any theming changes, then apply them to its subchildren. Fragments could do the same
* Ideally there should be a way to generate Theme object on the fly and apply it to the children, doable, but have several drawbacks

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -40,6 +40,7 @@ import android.content.pm.PackageManager;
 import android.content.res.Resources.NotFoundException;
 import android.os.Build;
 import android.os.Bundle;
+import android.os.Handler;
 import android.os.IBinder;
 import android.os.Parcelable;
 import android.support.design.widget.BottomNavigationView;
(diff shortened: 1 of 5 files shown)


tobiasKaminsky:
Thank you for your help 🎉 
I'll have a look into it on monday

przybylski:
A bit more progress on that I did a while ago but haven't pushed it. I will continue work on that soon

tobiasKaminsky:
> A bit more progress on that I did a while ago but haven't pushed it. I will continue work on that soon

So we wait until your PR is completely finished? :-)

tobiasKaminsky:
As there were no activity over the last months, I'll close this PR, but keep the branch.

PR #2106, opened 2018-02-06
getServerVersion() can be null, if so, then chunked = false

tobiasKaminsky:
Found via google dev console


AndyScherzinger:
👍

AndyScherzinger:

AndyScherzinger:

mario:
What versions don't support chunking? Afaik all versions that are not EOL do...

tobiasKaminsky:
Indeed, we check for 4.5.

--> I will enhance this PR to remove all unneeded checks for version <11 as those are end of life.

mario:
Let me know when done.

tobiasKaminsky:
This is done via #2178, therefore this PR can be closed.

PR #2109, opened 2018-02-06
Don't beautify empty inputs

przybylski:
Empty string as twitter handle as well as empty website is invalid. So don't beautify it.

### DIFF
--- a/src/main/java/com/owncloud/android/utils/DisplayUtils.java
+++ b/src/main/java/com/owncloud/android/utils/DisplayUtils.java
@@ -196,14 +196,14 @@ public class DisplayUtils {
         }
 
         if (url.length() >= 7 && url.substring(0, 7).equalsIgnoreCase(HTTP_PROTOCOLL)) {
-            return url.substring(HTTP_PROTOCOLL.length());
+            return url.substring(HTTP_PROTOCOLL.length()).trim();
         }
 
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Using multiple whitespaces on twitter are still shown, so we need to trim the input.
As (maybe) other use the library, maybe this is better in library? We could make sure that each string is either "" or has a correct value.

AndyScherzinger:
@tobiasKaminsky so how to progress? I'd be fine with this PR while that doesn't render the discussion about doing a library enhancement absolete.

tobiasKaminsky:
> Using multiple whitespaces on twitter are still shown, so we need to trim the input.

This is still a "bug".

Regarding lib/no lib: It was just an idea. I have no strong feelings regarding this.

AndyScherzinger:
👍 

> This is still a "bug".

@tobiasKaminsky @przybylski I added a trimming to the other cases for beautifying URL/twitter-handle. So I am fine with the PR +1 :)

mario:
👍

tobiasKaminsky:
Have you tested this?
Using "  " (two whitespaces) on twitter, results in



The twitter handle needs first to get trimmed, before checking for prefix.

AndyScherzinger:
@tobiasKaminsky you are right, created an issue for that #2159

PR #2118, opened 2018-02-08
Bump to 3.0

tobiasKaminsky:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -78,8 +78,8 @@ android {
 
         multiDexEnabled true
 
-        versionCode = 30000003
-        versionName = "3.0.0 RC3"
+        versionCode = 30000099
+        versionName = "3.0.0"
 
         // adapt structure from Eclipse to Gradle/Android Studio expectations;
         // see http://tools.android.com/tech-docs/new-build-system/user-guide#TOC-Configuring-the-Structure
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -78,8 +78,8 @@ android {
 
         multiDexEnabled true
 
-        versionCode = 30000003
-        versionName = "3.0.0 RC3"
+        versionCode = 30000099
+        versionName = "3.0.0"
```
rullzer: This  is master. So 3.1 right?

On the server when we fork off a new release that relase gets the final version.

Here you'd then get:

```
versionCode = 30010000
versionName = 3.1.0 Alpha
```

or something. Since your master will become 3.1.0. And the versioncode then just always increases whenever you do changes that require this.

Does this also make sense on android?

tobiasKaminsky: We can do that.
But I am not sure if this is worth the effort. 
Our next release based off master will be 30010001, which is 3.1.0 RC1.
And on some point in the future then, maybe at RC2, we will again branch off...

rullzer: okidoki

tobiasKaminsky: Hmm. After thinking about it ;-) I think it is indeed weird, that master is called 3.0, where it can happen that we already have a 3.0.1 based on 3.0.x branch.

AndyScherzinger:

AndyScherzinger:
Pinging @mario for second review :)

mario:
👍

PR #2146, opened 2018-02-12
Fix time display in activities list

### LINKED ISSUE #2141: Wrong time display in activity time
In the Activities tab of the app, it looks like there was a confusion between minutes and monthes (e.g. 18:02 is displayed instead of 18:53 because we are in february).

AndyScherzinger:
Fix #2141

Error has been the use of ```HH:MM``` instead of ```HH:mm``` since ```MM``` is month not minute.

cc @mario @tobiasKaminsky - Do we want to backport this for 3.0.1?


tobiasKaminsky:
:+1: 
Regarding backport: I think we should only backport really important stuff, like data loss, no login, etc.
This is a "minor" UI bug, which can wait.

AndyScherzinger:
fine by me :)

mario:
👍

PR #2148, opened 2018-02-12
Add nicer variable based versioning calculation

AndyScherzinger:
Based on Jake Wharton's post https://plus.google.com/+JakeWharton/posts/6f5TcVPRZij I rewrote the version code and name calculation. This way it is easier to maintain the version code and the name :)

cc @tobiasKaminsky @mario

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,7 +55,7 @@ def versionMajor = 3
 def versionMinor = 1
 def versionPatch = 0
 def versionBuild = 1 // alphas, betas, release candidates (<99=non-stable, 99=stable)
-def versionNameSuffix = " Alpha" // empty for stable releases
+def versionNameSuffix = "Alpha" // ignored for stable releases
 
 android {
     lintOptions {
@@ -86,7 +86,12 @@ android {
         multiDexEnabled true
 
         versionCode versionMajor * 10000000 + versionMinor * 100000 + versionPatch * 100 + versionBuild
-        versionName "${versionMajor}.${versionMinor}.${versionPatch}${versionNameSuffix}"
+
+        if (versionBuild > 98) {
+            versionName "${versionMajor}.${versionMinor}.${versionPatch}"
+        } else {
+            versionName "${versionMajor}.${versionMinor}.${versionPatch} ${versionNameSuffix}"
+        }
 
         // adapt structure from Eclipse to Gradle/Android Studio expectations;
         // see http://tools.android.com/tech-docs/new-build-system/user-guide#TOC-Configuring-the-Structure
--- a/build.gradle
+++ b/build.gradle
@@ -54,8 +54,7 @@ repositories {
 def versionMajor = 3
 def versionMinor = 1
 def versionPatch = 0
-def versionBuild = 1 // alphas, betas, release candidates (<99=non-stable, 99=stable)
-def versionNameSuffix = "Alpha" // ignored for stable releases
+def versionBuild = 0 // 0-49=Alpha / 50-98=RC / 99=stable
 
 android {
     lintOptions {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -50,6 +50,13 @@ repositories {
     }
 }
 
+// semantic versioning for version code
+def versionMajor = 3
+def versionMinor = 1
+def versionPatch = 0
+def versionBuild = 1 // alphas, betas, release candidates (<99=non-stable, 99=stable)
```
tobiasKaminsky: Alpha would be 0, or?
RC1 is 1, RC2 is 2...

build.gradle
```diff
@@ -50,6 +50,13 @@ repositories {
...
+def versionNameSuffix = " Alpha" // empty for stable releases
```
tobiasKaminsky: maybe hardcode the first whitespace so we cannot forget this?

build.gradle
AndyScherzinger: Tricky... basically it is "unknown" as to be able to upgrade this simply needs to increment its value with each published apk so it is 0-98 for any of the ones mentioned.

build.gradle
AndyScherzinger: DONE 👍  https://github.com/nextcloud/android/pull/2148/commits/a00e64dd0ea2eb9551dad27b4ccbf9b74b31c027

build.gradle
tobiasKaminsky: But if we release RC1 it would have the same number as current master, so there is no upgrade.
As we have 1 for rc1 reservered, I suggest to not re-use this / not use this prior it is a really rc1.

AndyScherzinger: So how would you release more than one alpha?

AndyScherzinger: If you want to map the version build to a release by convention than 10RC1, 2=RC2 and 0=Alpha/Beta/Whatever, while then there can only be excactly 1 publish for pre-RC (!)

AndyScherzinger: > tl;dr: to distinguish current development from RC1 I would like to have "0" on stable.

you mean on ```master```, right? Because the stable release would be ```99``.
If we need alphas in the future (with beta being RC or like mentioned above by you) then I'd say let's go for the numbers you defined.
* 0-19: alpha (GPlay Alpha)
* 20-49: beta (still GPlay Alpha)
* 50-98: RC (GPlay Beta)
* 99: stable

This way we can still use the alpha channel to release alpha/beta versions and use the beta channel for the RCs. We could even go further and allow more values for beta and less for RC (since I think we have a different Problem in case we need ~50RCs) as in:
* 0-39: alpha (GPlay Alpha)
* 40-87: beta (still GPlay Alpha)
* 88-98: RC (GPlay Beta)
* 99: stable

tobiasKaminsky: To not overcomplicate things, I would just go for
0-49 alpha (Gplay Alpha)
50 -98 RC (Gplay Beta)

In our current workflow we do not have a real beta, we call it directly RC.

AndyScherzinger: All right, will do! :)

AndyScherzinger: DONE https://github.com/nextcloud/android/pull/2148/commits/8dcfa887f34498bd8eace5f756a7a2ce9b48570c

AndyScherzinger:
@tobiasKaminsky all your comments have been implemented, the version name is now completely generated based on the version code numbers.

@mario please review :)

tobiasKaminsky:
:+1:

AndyScherzinger:
New week, new ping :grin: @mario please :heart:

mario:
Good things come to those that wait :p

AndyScherzinger:
Not merging to the dark side it leads you :frog: - Yoda @ OSS conference :stuck_out_tongue_closed_eyes:

mario:
👍

PR #2154, opened 2018-02-12
Encryption: Get IV directly from the cipher object when encrypting the private key

### LINKED ISSUE #2134: E2E: Unsupported class: class javax.crypto.spec.IvParameterSpec when encrypting the private key.
Using the latest version from Google Play, whenever I try to enable an encrypted folder, the app shows me a new recovery key phrase, and when proceeding I get an error saying "Keys could not be stored, please try again".

This is on a Pixel 2 XL running Android 8.1.0.

ardevd:
In Android O and newer (at least) javax.crypto.spec.IvParameterSpec is unsupported. Get the IV directly from the cipher object instead. This has been tested on API level 27 and found to work well. E2E encryption keys can now be stored correctly since the private key can be encrypted successfully.

This fixes #2134 and all the other issues describing the same issue.

### DIFF
--- a/src/main/java/com/owncloud/android/utils/EncryptionUtils.java
+++ b/src/main/java/com/owncloud/android/utils/EncryptionUtils.java
@@ -514,7 +514,7 @@ public class EncryptionUtils {
         byte[] bytes = encodeStringToBase64Bytes(privateKey);
         byte[] encrypted = cipher.doFinal(bytes);
 
-        byte[] iv = cipher.getParameters().getParameterSpec(IvParameterSpec.class).getIV();
+        byte[] iv = cipher.getIV();
         String encodedIV = encodeBytesToBase64String(iv);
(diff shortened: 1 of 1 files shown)


ardevd:
Should probably be tested with lower API levels to make sure it works for all of them. Worst case we will have to use a if statement to generate the IV differently newer/older API levels.

ardevd:
Tested on API level 23. Works fine there too.

AndyScherzinger:
Calling @tobiasKaminsky for e2e, should imho then be backported for 3.0.1

tobiasKaminsky:
:+1: 
Thank you for this great finding! :tada:

AndyScherzinger:
❤️ Thanks a lot for your contribution @ardevd ❤️

ardevd:
You're most welcome! Thanks for keeping things open source ;)

PR #2158, opened 2018-02-13
Delete public key if storage of private key did not worked

### LINKED ISSUE #2153: If app fails to store the E2E encryption keys the public key should be deleted from the server
Currently, the public key is posted to the E2E server side app and stored there for the user before the Android app tries to store the associated keys locally. If this fails, the user is prompted to create a new key the next time they try to go through the encryption process. This will then always fail because the server will return a 409 error as a public key already exists. I think the correct behavior would be to send a DELETE request to the E2E API to remove the uploaded public key.

Furthermore, the private key needs to be removed as well.

tobiasKaminsky:
Fix #2153 

First we generate a public key, if this fails, nothing is stored on server, so everything is fine.
If public key is stored successfully, we generate a private key and store this. 
If private key is stored correctly, we store both locally.

This PR now removes the public key if storage of the private key fails.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -182,7 +182,7 @@ dependencies {
     // dependencies for app building
     implementation 'com.android.support:multidex:1.0.2'
 //    implementation project('nextcloud-android-library')
-    implementation 'com.github.nextcloud:android-library:1.0.37'
+    implementation 'com.github.nextcloud:android-library:1.0.38'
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
@mario please approve, this is also good for backporting as otherwise people with one failed e2e setup cannot retry it.

mario:
👍

Issue #2034, opened 2018-01-24
Show message in auto upload configuration dialog when manditory fields haven't been filled and the users hit save.


j-ed:
### Actual behaviour
After I’ve installed the latest app version (3.0 RC2) I wanted to set a custom folder for the automatic upload function but the Save-button doesn't work at all.

### Expected behaviour
It should be possible to save custom settings in the automatic upload dialog.
 
### Steps to reproduce
1. Selected the automatic upload menu.
2. Select `create your own folder`.
3. A configuration dialog opens which allows to set different parameters and also the upload folder itself.
Everything seem to work, except that the new configuration cannot be saved because the Save-button is not working. The Abord-button works without any problem.

### Environment data
Android version: 7.0
Device model: Samsung Galaxy S7
Stock or customized system: Stock ROM
Nextcloud app version: 3.0 RC2
Nextcloud server version: 12.0.4

#### Web server error log
```
No error messages were logged.
```

#### Nextcloud log (data/nextcloud.log)
```
No error messages were logged.
```

AndyScherzinger:
@j-ed just tested it and it (at least) works for me. Did you set the local and the remote folder? Else the save button is blocked.

j-ed:
@AndyScherzinger You're absolutely right, I missed setting the local folder. Due to the fact that no error message was displayed I thought all other settings would have been pre-populated by the app. It might be worse to
a) set the local folder parameter to a default value.
b) through-out an error message if it hasn't been set or couldn't be accessed.

AndyScherzinger:
Happy to hear it work and thanks for the suggestions.
Since you are basically setting up the upload for a local folder I would not set it to a default (might be rather an option for the remote folder).
So showing a message for missing fields might be the easier option to implement :)

What do you think @tobiasKaminsky @mario ?

mario:
Yup yup.

tobiasKaminsky:
![2018-02-13-120853](https://user-images.githubusercontent.com/5836855/36146959-aa23217e-10b6-11e8-9fee-1f03a3aaf419.png)

Save button needs to be grayed out / disabled so that user can see that something is missing.
Showing a message on top of this dialog will look weird, I think. 

@jancborchardt do you have an idea how to make it more visible to the user that both local and remote folder needs to be chosen?

AndyScherzinger:
> Showing a message on top of this dialog will look weird, I think.

true, so yes the best would be to either
* grey out the button or
[...]

PR #2164, opened 2018-02-13
Grey-out auto upload options until enabled

### LINKED ISSUE #1260: Grey-out auto upload options until enabled
As suggested by @gahujipo here: https://github.com/nextcloud/android/pull/797#issuecomment-319973580

> The preferences window for auto upload shows a switch next to "Preferences for auto uploading". At the first sight it was not clear to me that that switch is uses to enable the auto upload feature for that folder pair. My idea was to grey out the options in the window unless the switch is set to true.

mario:

Fixes #1260

### DIFF
--- a/src/main/java/com/owncloud/android/ui/dialog/SyncedFolderPreferencesDialogFragment.java
+++ b/src/main/java/com/owncloud/android/ui/dialog/SyncedFolderPreferencesDialogFragment.java
@@ -60,15 +60,13 @@ import static com.owncloud.android.datamodel.SyncedFolderDisplayItem.UNPERSISTED
  */
 public class SyncedFolderPreferencesDialogFragment extends DialogFragment {
 
-    private final static String TAG = SyncedFolderPreferencesDialogFragment.class.getSimpleName();
     public static final String SYNCED_FOLDER_PARCELABLE = "SyncedFolderParcelable";
-    private static final String BEHAVIOUR_DIALOG_STATE = "BEHAVIOUR_DIALOG_STATE";
     public static final int REQUEST_CODE__SELECT_REMOTE_FOLDER = 0;
     public static final int REQUEST_CODE__SELECT_LOCAL_FOLDER = 1;
-
-    private CharSequence[] mUploadBehaviorItemStrings;
-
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
👍 looking good code wise

mario:
@tobiasKaminsky  :P

tobiasKaminsky:
Isn't there a giant discussion how to change this at #2034 ?
Will this fit into your idea @AndyScherzinger?

AndyScherzinger:
@tobiasKaminsky this still fits with my idea on the overall change since this is related to the on/off status while my other ideas are related to media/custom type folder pairs. Thus this is still valid :+1:

tobiasKaminsky:
The visual difference between enabled and disabled is too low in my opinion.
It is also confusing that there is an checked, but disabled checkbox (only upload on wifi).
Moreover you can click on save when the switch is disabled (left image), which is also confusing. What is saved?

I still like the idea, but somehow it is now even more confusing for me.

As we have sane defaults for these synced folders, we could enable the folder upon on save and then omit this switch in preferences dialog.

@jancborchardt

tobiasKaminsky:
Please feedback by @nextcloud/designers, for me the questions in my last post are still open.

mario:
@nextcloud/designers whenever you get a chance :P

jancborchardt:
Absolutely agree with Julius. Unless the first setting is enabled, everything below is irrelevant and does not need to be shown.

mario:
Can we get this merged for 3.1, and then we'll improve for 3.2? @AndyScherzinger @tobiasKaminsky

AndyScherzinger:
I already approve and would also be fine with merging and improving upon this foundation with 3.2 cc @tobiasKaminsky

AndyScherzinger:
@tobiasKaminsky what do you think? Imho this PR could be merged being a better solution than the actual state and improve upon it afterwards?

tobiasKaminsky:
:+1: 
I will open up a following issue to enhance this.

AndyScherzinger:
@tobiasKaminsky the corresponding issue is assigned to milestone 3.1 -> so should this PR then be backported?

AndyScherzinger:
Nevermind, you already wrote that in the roadmap issue for 3.1... sorry...

PR #2178, opened 2018-02-15
Remove check for older version (<11)

tobiasKaminsky:
Started with removing old version in OwncloudVersion and then changing the checks to true, simplifying whereever possible.
Also changed webdav path from "remote.php/webdav" to "remote.php/dav".

Note: after merge of library PR, build.gradle needs to be adjusted.


AndyScherzinger:
How incompatible to oC will we then be? Just asking since we already get 1-star ratings since 3.0.0 didn't work with oC X. I am not criticizing this I just like it to be an explicit decision :)

mario:
@AndyScherzinger entirely incompatible :P 

@tobiasKaminsky did you test everything with /dav? :) (uploads, etc)

tobiasKaminsky:
I reverted the webdav stuff as this indeed would need more changes and this PR should be kept "simple".

Regarding oC compatibility, I would not say that it is "entirely incompatible", if you see, what version checks are removed, they are even on oC end of life:
https://github.com/nextcloud/android-library/pull/122/files#diff-5df905fe275038070a60858bf0d74786

I did a test with oC and fixed one problem during login: they do not have quota in UserInfo, so we need to ignore this. Then I can login, browse & download files, ...
I think this is enough as NC != oC.
(Any negative rating regarding oC can in my opinion be closed as it is out of scope of this app).

tobiasKaminsky:
Lint is showhing +1 error because of the referenced library branch.
Once we have a new library release, this warning will be gone.

AndyScherzinger:
> Any negative rating regarding oC can in my opinion be closed as it is out of scope of this app

totally agree on this while the ratings (potential crashes) don't show the server version / info so we don't really know ;)

tobiasKaminsky:
As with latest E2E user agents changes oC server are already not working anylonger, it is also safe to rely completly on NC with this PR.

tobiasKaminsky:
This needs a new library release.

AndyScherzinger:
@tobiasKaminsky So will this now completely drop oC compat?

tobiasKaminsky:
As said [here](https://github.com/nextcloud/android/pull/2178#issuecomment-372209598) is is already not working.
With this PR we assume that we have at least all functionality that NC10 has, e.g. support for forbidden chars, etc. 
I think for most of the stuff it would still work with oC (if not the user agent would block it).

AndyScherzinger:
Needs a rebase with massive amount of conflicts, other than that I'd be fine with supporting Nc11+

src/main/java/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -633,7 +632,7 @@ public int onStartCommand(Intent intent, int flags, int startId) {
                             account,
                             files[i],
                             ocUpload,
-                            chunked,
+                            true,
```
AndyScherzinger: If is is always true -> shouldn't we not just drop the parameter completely?

src/main/java/com/owncloud/android/ui/fragment/ShareFileFragment.java
```diff
@@ -227,7 +226,7 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
         addUserGroupButton.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View view) {
-                boolean shareWithUsersEnable = AccountUtils.hasSearchUsersSupport(mAccount);
+                boolean shareWithUsersEnable = true;
```
AndyScherzinger: same here, if always true, just drop the `else` part and the variable :)

src/main/java/com/owncloud/android/utils/glide/HttpStreamFetcher.java
```diff
@@ -59,23 +59,19 @@ public InputStream loadData(Priority priority) throws Exception {
...
+                int status = mClient.executeMethod(get);
+                if (status == HttpStatus.SC_OK) {
+                    return get.getResponseBodyAsStream();
+                } else {
+                    mClient.exhaustResponse(get.getResponseBodyAsStream());
                 }
-            } else {
-                Log_OC.d(TAG, "Server too old");
+            } catch (Exception e) {
+                Log_OC.d(TAG, e.getMessage(), e);
```
AndyScherzinger: should then be `Log_OC.e`

AndyScherzinger:
haven't had the chance to test this yet, code-wise just some minor comments

mario:
Unsure if this is what we want. Let me look into the consequences in more detail. Remind me if I forget.

mario:
My general opinion regarding this is that regardless if something is EOL or NOT, the client should continue to support it for some time further, especially if it's not a big burden. 

We also should strive to be compatible with oC where possible, even when that means supporting oC only features (if such exist).

Unfortunately, since I don't have the time these days to work on Files much (but rather spend it on Talk) all I can do is voice my opinion rather than offer working hours, and I guess neither can @AndyScherzinger since he has a full-time job.

Ultimately it'll boil down to you or maybe @ardevd to decide on this, but I felt it's important I say what I think here.

AndyScherzinger:
I agree with @mario. For me it is also not about support oC-only features but to support features that are present on both sides (the same way) and also kept backward compatibility with Nc/oC server versions that maybe eol like 10 or now 11 since some users may not be able to upgrade that quickly, like with freeNas distros. so to me 11 is (kind of) a _must_ and 10 is to be discussed. But looking at the speed of mayor releases in the server I don't feel comfortable dropping server support that early. To me something like at least 2 years or maybe even 3 year (which would then even include Nc9!) should be supported by the clients since clients shouldn't be bound to server eol cycles.

What do you think @jancborchardt @MorrisJobke (from a _project_not_company_ pov)?

MorrisJobke:
> 10 or now 11 since some users may not be able to upgrade that quickly

Just to be sure: 10 has at least 3-4 security vulnerabilities. Also it is EoL since 10 months. It's not that there is much we can do here. It is end of life and it should be upgraded as soon as possible.



> To me something like at least 2 years or maybe even 3 year (which would then even include Nc9!)

That also means you ran a web service with known security vulnerabilities. Sure you could run that, but then you should also be able to feel the burden of that: doing your own Android provisioning by using old Android versions.

We try to make updates as easy as possible. And also end of life means end of life ... and not supporting it from a client side view 2 more years. Then you would also get issues with "feature A is broken", because we never fixed it in the server part.

MorrisJobke:
Couldn't you add a banner with "Not supported version"? Just a splash screen with "The server is outdated and thus not supported anymore - contact your admin" and maybe a link to a "end of life page"? Then you would properly communicate to the user WHY the app is not working anymore. That a given distributions ships a totally outdated version doesn't mean we need to support it super long. There should be an incentive to move to a newer version.

tobiasKaminsky:
To clarify, this PR is not removing any functionality on NC10/11.
It is only removing certain checks, e.g. there is support for capabilities built in oC 8.1 and thus it is in every NC version.
We simply assume, if we have no server info, we can use every NC10 feature.

tobiasKaminsky:
> Couldn't you add a banner with "Not supported version"? Just a splash screen with "The server is outdated and thus not supported anymore - contact your admin" and maybe a link to a "end of life page"? Then you would properly communicate to the user WHY the app is not working anymore. That a given distributions ships a totally outdated version doesn't mean we need to support it super long. There should be an incentive to move to a newer version.

This is a complete different issue, so I'll open up a new one.

mario:
Rebase.

mario:
Code wise fine, so if this is what you want to do, I'll approve after rebase.

mario:
(But please feel comments from @AndyScherzinger )

tobiasKaminsky:
Rebased.
I tested it against NC10:
- login is only possible with legacy login 
- upload a file
- showed resized image
- download a file
- sharing a file
…

Please test it also with NC10 :-)

Again, this PR is not removing anything or limiting support for outdated sever.
It just assumes that we can rely on any feature that is built in NC10, e.g.:
- sharing API: 5.0.27
- searching user: 8.2
- remote thumbnails: 7.8
- federated sharing: 9.1 (oc version, not nc9)

--> really nothing changed, except that there is no a lot less checks.

AndyScherzinger:
All fine by me then, my point is rather that I do not think cutting 3 lines of code and knowingly break the client for legacy, unsupported server is a good idea.

Also we I understand the argument of creating some kind of "pressure" to update the server while I think (like mentioned by Morris with the banner, popup, whatever) it should be done gracefully and also the issue here is that you are else hurting the users who in many cases can't do anything about the server and can't put pressure on the admins thus you are just hurting users non the server-providing people...

And also I wouldn't want to see this going into a direction where the client breaks for unsupported servers except when you pay for support... I know none of us is doing so at the moment, I just want to make sure we don't start with changes that would more or less actively support/push towards such a scheme.

:+1: All good for now and yes we should have #2240 since this is how a client should behave anyways :)

MorrisJobke:
> Also we I understand the argument of creating some kind of "pressure" to update the server while I think (like mentioned by Morris with the banner, popup, whatever) it should be done gracefully and also the issue here is that you are else hurting the users who in many cases can't do anything about the server and can't put pressure on the admins thus you are just hurting users non the server-providing people...
> 
> And also I wouldn't want to see this going into a direction where the client breaks for unsupported servers except when you pay for support... I know none of us is doing so at the moment, I just want to make sure we don't start with changes that would more or less actively support/push towards such a scheme.

This was also not meant like this. It's just a general warning, that users know why something doesn't work. It should just give the possibility to detect it without digging into logs and debugging things. Basically like something we do with the setup checks on the server. Giving some indicators, that something could break and why it does so. Otherwise people are frustrated that it just™ doesn't work.

AndyScherzinger:
>  Giving some indicators, that something could break and why it does so. Otherwise people are frustrated that it just™ doesn't work.

Excatly why I like the suggestion of having some kind of warning, that the server is EOL/legacy/whatever and some things might crash (also lowering the amount of issues raised due to such situations)

tobiasKaminsky:
Rebased.
Are you all fine with merging (after library merging & changing lib reference)?

AndyScherzinger:
:+1: fine by me then, yes

AndyScherzinger:
:tada: needs to be rebased again :(

AndyScherzinger:
Sidenote: #2728 with the server version warning has been merged already

tobiasKaminsky:
Rebased

mario:
👍

AndyScherzinger:
needs yet another rebase due to today's PR merge marathon while I think it can wait until nextcloud/android-library#124 has been merged ;)

PR #2180, opened 2018-02-15
Correctly load saved folder on resume

### LINKED ISSUE #2161: App gets stuck in synching state when put into background and then killed by the  system
### steps to reproduce:

1. Be logged in and have the folder list fully synced
2. Enable the "Don't keep activities" setting in androids developer options to force this bug to happen instantly
3. Put nextcloud in background.
4. Bring it into foreground again

The app now displays only empty folders or the "synching" state this apparently goes on indefinitely.

Explicitly closing nextcloud (swiping it away from recently used apps) and opening it again instantly brings back all folders and sync state.

### notes

This has been happening definitely since v2.0 and now with 3.0 as well.

Cc @tobiasKaminsky as we talked about this yesterday.

tobiasKaminsky:
Fix #2161 


AndyScherzinger:
👍

mario:
👍

PR #2183, opened 2018-02-15
If no thumbnail exists, check if resized version exist and use this

tobiasKaminsky:
This is useful if you have a very large image folder, scroll to a decent point and then immediately click on an image. While swiping the resized version is fetched.
If you then browse the file listing, the app would download the thumbnails, but now uses the resized images as source.

### DIFF
--- a/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -536,10 +536,8 @@ public class ThumbnailsCacheManager {
                     if (resizedImage != null) {
                         thumbnail = ThumbnailUtils.extractThumbnail(resizedImage, pxW, pxH);
                     } else {
-                        //Download thumbnail from server
-                    OwnCloudVersion serverOCVersion = AccountUtils.getServerVersion(mAccount);
-                    if (mClient !=  null) {
-                        if (serverOCVersion.supportsRemoteThumbnails()) {
+                        // Download thumbnail from server
+                        if (mClient != null && AccountUtils.getServerVersion(mAccount).supportsRemoteThumbnails()) {
                             getMethod = null;
                             try {
                                 // thumbnail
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
@AndyScherzinger ping :-)

AndyScherzinger:
:+1: Nice solution/idea :)

src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -528,55 +528,62 @@ private Bitmap doThumbnailFromOCFileInBackground() {
...
-                                Log_OC.d(TAG, e.getMessage(), e);
-                            } finally {
-                                if (getMethod != null) {
-                                    getMethod.releaseConnection();
+                    if (resizedImage != null) {
+                        thumbnail = ThumbnailUtils.extractThumbnail(resizedImage, pxW, pxH);
+                    } else {
+                        // Download thumbnail from server
+                        OwnCloudVersion serverOCVersion = AccountUtils.getServerVersion(mAccount);
+                        if (mClient != null && serverOCVersion != null) {
```
mario: serverOCVersion cannot be null :P

src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -528,55 +528,62 @@ private Bitmap doThumbnailFromOCFileInBackground() {
...
+                            if (serverOCVersion.supportsRemoteThumbnails()) {
```
mario: please use AccountUtils.getServerVersion(mAccount).supportsRemoteThumbnails() where possible.

Also put if (mClient != null && AccountUtils.getServerVersion(mAccount).supportsRemoteThumbnails()) in the else above.

src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
tobiasKaminsky: #2211 is not yet merged and there it is already changed

src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
tobiasKaminsky: The first one is already done in #2211.
I'll wait until #2211 is merged, rebase this and the adjust the second one.

tobiasKaminsky:
@mario rebased & implemented your idea :

mario:
👍

AndyScherzinger:
@tobiasKaminsky all green except this _continuous-integration/drone/push_ check. This started within the last week or so I think. Any idea where this comes from?

MorrisJobke:
> @tobiasKaminsky all green except this continuous-integration/drone/push check. This started within the last week or so I think. Any idea where this comes from?

Yes - somebody configured it to be required (and thus GitHub automatically adds it (you can see this by the missing details link - this link gets added once the service actually says, that it received the check). As push checks are not executed by default by drone it is not filled and therefore missing ;)

bildschirmfoto 2018-03-20 um 07 56 19

tobiasKaminsky:
Thanks @MorrisJobke for your help, I changed the settings back to the previous one
- only PR
- no github pull request review is requried

PR #2185, opened 2018-02-16
If image size is zero, try to get size from media store

tobiasKaminsky:
This is a fix for certain OnePlus devices, where the image size taken by the default camera is zero.
I tested it with many images, but still some are uploaded corrupted. This then due to a wrong size reported by media store (instead of 0, it is something between 20-100kb).

So it is better than nothing, but not 100% reliable. As this is only used if file.length() is not working, it should have no impact on other devices.

Requires: https://github.com/nextcloud/android-library/pull/125

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -212,9 +212,9 @@ dependencies {
     // dependencies for app building
     implementation 'com.android.support:multidex:1.0.3'
 //    implementation project('nextcloud-android-library')
-    genericImplementation "com.github.nextcloud:android-library:master-SNAPSHOT"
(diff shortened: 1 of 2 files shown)


mario:
Could you instead try a different approach? If phone is oneplus, only schedule things for upload if 30 seconds has passed since they've been written to.

tobiasKaminsky:
I tested auto-upload with "only on wifi" and turned the wifi on 5mins later after taking an image and still had the problem (on master).
--> The file size is always zero, and does not matter how long I wait.
(But it might help with 20-100kb problem)

tobiasKaminsky:
closing as it is a fix for minor amount of users.

PR #2186, opened 2018-02-16
E2E: do not modifiy userid

### LINKED ISSUE #2134: E2E: Unsupported class: class javax.crypto.spec.IvParameterSpec when encrypting the private key.

### LINKED ISSUE #2173: key storage error continues e2e
I have updated to nextcloud 3.0.1 I have read of the resolution of the problem of the key storage on oreo, but for those who have nougat still not good. i have s8 samsung and nexus 6 with android 7.x but error also persists with 3.0.1
screenshot_20180214-145614

tobiasKaminsky:
Fix #2173 
Fix #2134 

As we now have a call to get the uid, we shall not split the string, as usernames such as are valid: tobi@nc

### DIFF
--- a/src/main/java/com/owncloud/android/utils/CsrHelper.java
+++ b/src/main/java/com/owncloud/android/utils/CsrHelper.java
@@ -58,7 +58,7 @@ public class CsrHelper {
      */
     private static PKCS10CertificationRequest generateCSR(KeyPair keyPair, String userId) throws IOException,
     OperatorCreationException {
-        String principal = "CN=" + userId.split("@")[0];
+        String principal = "CN=" + userId;
         AsymmetricKeyParameter privateKey = PrivateKeyFactory.createKey(keyPair.getPrivate().getEncoded());
         AlgorithmIdentifier signatureAlgorithm = new DefaultSignatureAlgorithmIdentifierFinder().find("SHA1WITHRSA");
         AlgorithmIdentifier digestAlgorithm = new DefaultDigestAlgorithmIdentifierFinder().find("SHA-1");


AndyScherzinger:
but wouldn't you now be using tobi@nc@yourawesomeserver.com?

tobiasKaminsky:
No, we fetch the userid, which never has server included, e.g:
userid: tobi@nc
name: tobias kaminsky
email: tobias@nc.server.com

Now with userid fetching, it does not matter with what you authenticate (userid, name, email), the CSR will always use userid (tobi@nc).

Previously, we took simply the "id" with which we authenticated (on account setup).
So if you back then used "tobi@nc" it was fine, as loginid matches the userid.
But if you used your email address, it would not have worked.

AndyScherzinger:
👍 alright! I did think the userId is actually username@server which is why we usually do the @-split.

mario:
Sure, but how do we migrate existing users?

tobiasKaminsky:
We do not have to: CSR is only signed by server if common name in CSR matches the userid.
Therefore we get always the userid before generating a CSR:
https://github.com/nextcloud/android/blob/04de5cab62a77c18d51532aafe3eafdc0f36265d/src/main/java/com/owncloud/android/ui/dialog/SetupEncryptionDialogFragment.java#L348-L363

(the above is true for all stable app releases, on alpha testing there was no server side checking of userid)

mario:
👍

PR #2201, opened 2018-02-19
use correct size for png

### LINKED ISSUE #2190: Full size previews do not scale to fullscreen
### Actual behaviour
When opening a fresh image, the full preview is shown after short time. As intended. But it seems that the temporal preview in full size is not scaled well to full-screen. They appear to be too small. This does not apply to downloaded images.

### Expected behaviour
The image is previewed. After short time the full size image is shown. The temporal full size preview is scaled well to full screen, just as the downloaded one.
 
### Steps to reproduce
1. Upload image to server which has less pixels than the screen of your mobile device.
2. Browse to this image with android client.
3. Open (non downloaded) image with android client. Wait for  full size image being delivered automatically.
4. Check, if image is scaled to full screen.

### Environment data
Android version: 7.1.2

Device model: BQ X5 Plus

Stock or customized system: LOS 14.1

Nextcloud app version: 3.0.1

Nextcloud server version: 13.0

tobiasKaminsky:
Fix #2190 

For PNG we create a canvas. Previously it was fullscreen, which results in a small image if the image is smaller than the display.
--> use size of image instead and scale it up, if necessary.

 

### DIFF
--- a/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -304,8 +304,8 @@ public class ThumbnailsCacheManager {
                                 }
 
                                 // Handle PNG
-                                if (file.getMimetype().equalsIgnoreCase(PNG_MIMETYPE)) {
-                                    thumbnail = handlePNG(thumbnail, pxW, pxH);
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
:+1:

mario:
👍

PR #2205, opened 2018-02-20
Fix crash on 4.x devices

### LINKED ISSUE #2184: App crashes on back arrow and menu.
On new version 3 nextcloud crashes when I'm using the back arrow on the upper left of the app or the menu on the same place. Works fine when I use the back button of my device or slide from left to right to perform the same actions. It was ok on previous versions. 
Lenovo a850 - android 4.2.2

tobiasKaminsky:
Fix #2184 

Android 4.x does not allow tinting the action bar, so this is checked within the method and not on calling code.

I suggest a backport


AndyScherzinger:
:+1: for code and backport

mario:
👍

PR #2207, opened 2018-02-21
Close cursor

tobiasKaminsky:
Found via google play console: android.database.CursorWindowAllocationException

This means that there are too many open cursors.

The main reason is that cursor was only closed if != null, which is sane, but also only if c.moveToFirst() was true. But it can happen that an empty cursor exists (cursor with no result) and this then would never be closed.


AndyScherzinger:
:+1: Nice catch!

mario:
👍

PR #2210, opened 2018-02-21
if searchView does not exist, search is not opened

tobiasKaminsky:
Reported by google play console.
I could not reproduce it.


AndyScherzinger:
:+1:

mario:
👍

PR #2221, opened 2018-02-22
Copy/move starts within current folder

### LINKED ISSUE #400: When moving/copying, show the folder the user started in
Often when I move or copy files in the app I'm organizing them within a particular hierarchy in my Nextcloud. Currently the folder picker starts at the root, but it'd be nice if it started in the same folder the user was already looking at. (This is particularly nice because even if it's more convenient for the user to start at the root, it's much easier for them to press back a couple times than it is for a user in the current situation to navigate through a bunch of folders.)

Alternately, we could start in the last used folder.

tobiasKaminsky:
Fix #400 

When copy/move a file the folder picker activity will start with current folder.


AndyScherzinger:
:+1: Nice, this improves UX :rocket:

AndyScherzinger:

mario:
👍

PR #2222, opened 2018-02-23
Move lib implementation to flavors

tobiasKaminsky:
On the old implementation the stable lib was used for versionDev.
So stable lib has to be set for each flavor.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,6 +32,7 @@ configurations.all {
 ext {
     supportLibraryVersion = '26.1.0'
     googleLibraryVersion = '11.2.2'
+    androidLibraryVersion = '1.0.39'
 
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
:+1:

mario:
Why are the versions different though?

tobiasKaminsky:
Nice catch, some sort of weird c&p error ;-)

AndyScherzinger:
> Nice catch, some sort of weird c&p error ;-)

To not run into this in the future shouldn't we put it into a variable like we do with the google lib version?

mario:
Yes please :P

tobiasKaminsky:
Very nice idea, Andy :+1:

mario:
👍

PR #2228, opened 2018-02-26
TimSort: Add correct null check

tobiasKaminsky:
Spend quite a long time on this to understand the problem.

Imagine you compare o1 and o2 and both are null.
Previous, depending on your compare order, it could happen:
compare(o1, o2) -> -1, which leads to o1 < o2
compare(o2, o1) -> -1, which leads to o2 < o1
and this is not allowed.

So if both are null, there is no sorting order, so it returns 0.

This, however does not solve other sorting problems, e.g. where OCFiles are sorted via the compareTo function in OCFile as there would be a NPE if some file/remote path is null.


AndyScherzinger:
:+1:

mario:
👍

PR #2233, opened 2018-02-26
Show files in FolderPicker

### LINKED ISSUE #2220: Only folders shown on copy/move
Left: source folder, file 1.txt is moved
Right: same folder as "target", but only folders are shown.

If you enter a folder with only files, it would show "no files here", which is wrong.
This activity is used widely across the app, from copy/move to folder picker for synced folders.

Can you imagine any scenario where it would be sane to hide/filter out all files?

@jancborchardt

tobiasKaminsky:
Fix #2220 

This is used in 
- copy
- move
- select remote folder on synced folder

Clicking on a file just does nothing.


AndyScherzinger:
:+1:

AndyScherzinger:

mario:
👍

PR #2252, opened 2018-02-28
App context is maybe null, so use getContentResolver() directly

tobiasKaminsky:
Of course this should never be null, but according to google dev console this is happening :S


AndyScherzinger:
:+1:

AndyScherzinger:
cc @mario for review

mario:
👍

PR #2262, opened 2018-03-01
Add natural sorting

tobiasKaminsky:
This hopefully helps to fix TimSort errors.
With previous implementation transient comparison was not well defined:
a < b
b < c
--> a < c

This should now be the case.
I also added natural sorting:
1, 01, 001, 2, 002, 11, 12, 100

- first use value without leading zero
- if then value is equal, sort by leading zeros

### DIFF
--- a/src/main/java/com/owncloud/android/datamodel/OCFile.java
+++ b/src/main/java/com/owncloud/android/datamodel/OCFile.java
@@ -269,6 +269,18 @@ public class OCFile implements Parcelable, Comparable<OCFile> {
         return mMimeType != null && mMimeType.equals(MimeType.DIRECTORY);
     }
 
+
+    /**
+     * Sets mimetype to folder and returns this file
+     * Only for testing
+     *
+     * @return OCFile this file
+     */
+    public OCFile setFolder() {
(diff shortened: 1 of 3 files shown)


src/main/java/com/owncloud/android/datamodel/OCFile.java
```diff
@@ -618,10 +618,8 @@ public int describeContents() {
     }
 
     @Override
-    public int compareTo(OCFile another) {
-        if (isFolder() && another.isFolder()) {
-            return getRemotePath().toLowerCase(Locale.ROOT).compareTo(another.getRemotePath().toLowerCase(Locale.ROOT));
-        } else if (isFolder()) {
+    public int compareTo(@NonNull OCFile another) {
+        if (isFolder()) {
```
AndyScherzinger: This will now fail when both are folders! because a-b = -1 and b-a = -1 (!)

tobiasKaminsky: Right. I changed this and added more testcases.

AndyScherzinger:
:+1:

mario:
👍

PR #2286, opened 2018-03-05
Fixed layout when old login method enforced

alexcustos:
The main issue with mAuthStatusText which is null when showAuthStatus looking for an empty string. It also not clear why serverStatus should be hidden on old login method screen, so I removed that check.

### DIFF
--- a/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -212,7 +212,7 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
     private TextView mServerStatusView;
 
     private TextWatcher mHostUrlInputWatcher;
-    private String mServerStatusText;
+    private String mServerStatusText = "";
     private int mServerStatusIcon = 0;
 
     private boolean mServerIsChecked = false;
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
:+1: 

cc @tobiasKaminsky for login part

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1977,7 +1977,7 @@ public void onRegisterClick(View view) {
      * to the last check on the ownCloud server.
      */
     private void showServerStatus() {
-        if (mServerStatusIcon == 0 && "".equals(mServerStatusText) || forceOldLoginMethod) {
+        if (mServerStatusIcon == 0 && "".equals(mServerStatusText)) {
```
tobiasKaminsky: Why do you remove "forceOldLognMethod"?
This is needed if you use the new weblogin method, but it hangs and you want to use the old method as a fallback (there is a snackbar showing up after 60s offering this).

On the other hand it is not needed to assign an empty string to mAuthStatusText as "".equals(null) will work and result in false.

alexcustos: I found it weird that serverStatus didn't appear on the old login method screen. At first, I thought that there is a bug with inverted logic. In other words, it should be hidden just on WebLogin screen, but then I decided that it's not a problem at all if serverStatus will be shown on both login screens.

AndyScherzinger: @tobiasKaminsky @mario how do wa want to progress here?

tobiasKaminsky:
:+1: 
I tested it and indeed, server auth status was missing, so I was wrong on my first comment.

PR #2289, opened 2018-03-05
Bump to 3.0.3

tobiasKaminsky:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 3.0.3 (Mar, 05, 2018)
+- Fix creating folders in auto upload
+
 ## 3.0.2 (Feb, 27, 2018)
 - Fix crash on old android versions
 - Fix E2E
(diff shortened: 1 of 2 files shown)


mario:
👍

xXSTrikeXx:
@tobiasKaminsky Why is the changelog only updated only in the branch stable-3.0.x and not in master? I think this confuse people when they want to see the changelog and in master there is only the "big" changelog 3.0.0

tobiasKaminsky:
Indeed this is missing.
@mario @AndyScherzinger do we want to have changelog still in github or can we put it into wiki? So then we could avoid porting from 3.x branch to master.
Pointing changelog to wiki instead of github file should make no big difference.

AndyScherzinger:
I am fine either way while I think @jancborchardt urged us in the beginning to not make use of the GH wiki since we then depend on GH.

tobiasKaminsky:
Ok, good point. So it will be a "backport" ;-)

tobiasKaminsky:
https://github.com/nextcloud/android/pull/2338

Issue #2266, opened 2018-03-02
nextcloud android does not install apk from unknown sources Oreo Samsung s8


tigernero79:
With introduction of oreo each application must be specified if it has permission to install apk from unknown sources, this with nougat had to be done by enabling installation from unknown sources, in oreo must be done for each application, but when I try to click on an apk file that is in my nextcloud does not start installing and asks me if I want to authorize installation of apk from unknown sources, with nougat that happened without problems

I also click the voice open with after downloading it but nothing happens

tigernero79:
Practically with oreo Nextcloud is not detected among programs that can install unknown applications

ardevd:
@tigernero79 Why did you open this issue and then #2269?

tigernero79:
to indicate that it is clicking on the file and then going into the open menu with the same error.

surely it is due to the same bug. that you solved in pr

AndyScherzinger:
Fixed via #2311 for 3.1

tigernero79:
@AndyScherzinger
how far is expected to exit nextcloud 3.1 app?

AndyScherzinger:
We plan to release a release candidate 1 to Google Play beta program in ~1+ week with a RC phase of ~2 weeks. So the stable release 3.1.0 should be somewhere around 9th of April.

AndyScherzinger:
Other than that tomorrows nightly dev build will ship this fix

PR #2311, opened 2018-03-07
Added the REQUEST_INSTALL_PACKAGES permission to the Android manifest.

### LINKED ISSUE #2269: the open function with , does not work in Oreo for apk files
@tobiasKaminsky  
the open function with does not work in Oreo for apk files

ardevd:
With API 26 (Android O) any app that wants to install unknown apps must declare the REQUEST_INSTALL_PACKAGES permission. The user can then control what specific apps should be allowed to install unknown apps unlike the previous system where it was a system wide binary option for all apps. Without this declared permission the Nextcloud app will not be able to install apk's that the user have in their Nextcloud storage. With this permission added the user will be prompted to grant the app permission to install apps from unknown sources should they attempt to do so .

This fixes #2269 #2266

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -62,6 +62,7 @@
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.USE_FINGERPRINT" />
+    <uses-permission android:name="android.permission.REQUEST_INSTALL_PACKAGES" />
 
 
     <application


AndyScherzinger:
:+1: cc @tobiasKaminsky @mario

mario:
Uh sure. Though I don't like this permission. 👍

AndyScherzinger:
@tobiasKaminsky what do you think?

ardevd:
@mario I understand your concern about adding more permissions but this is literally the only way to enable apk installation from the NextCloud app. You're free to define that operation as unsupported of course.

More info [here](https://android-developers.googleblog.com/2017/08/making-it-safer-to-get-apps-on-android-o.html)

mario:
@ardevd :  "Remember that Play policies still apply to apps distributed on Google Play if those apps can install and update other apps. In the majority of cases, such behavior is inappropriate; you should instead provide a deep link to the app's listing on the Play Store. "

AndyScherzinger:
plus @ardevd You can also query the state of this permission using the PackageManager canRequestPackageInstalls() API --> So it can be taken away any time so it would need to be checked imho.

ardevd:
@mario Except when Im running Nextcloud where I have a bunch of apk's stored it doesnt really make sense to direct the user to Google Play.

@AndyScherzinger The permission is obviously disabled by default. Only after the user goes into the System settings and explicitly grant the Nextcloud app the permission to install apps from unknown sources is this possible. Also, Android handles the runtime checking for this for you automatically. If you try to install an APK, Android will prompt you with a dialog saying you have to grant the Nextcloud app permission to install unknown apps and even directs you to the relevant settings page if you so choose.

AndyScherzinger:
>  If you try to install an APK, Android will prompt you with a dialog saying you have to grant the Nextcloud app permission to install unknown apps and even directs you to the relevant settings page if you so choose.

Nice :+1:

AndyScherzinger:
@tobiasKaminsky ping ;-) What do you think? Code-wise imho fine but this is rather about should we support APK installations (we did in the past, since it was allowed if the user allowed it for any "unknown source")

ardevd:
I'll leave the decision up to you guys obviously. While it's a very good idea to think things through before slapping a new permission in the Manifest I struggle to see why you wouldn't do it in this case. There is a clear use case for installing apks from a self hosted cloud storage solution. Google drive provides this functionality as well. Furthermore, it's an opt-in permission which is completely left to the users discrection. The Nextcloud app won't nag the user about it and unless you actually want to install an apk you'll be none the wiser. By not adding this permission you'll end up with a situation where a piece of functionality (installing apks) is possible on some versions of Android but not others.

Just my two cents ;)

AndyScherzinger:
@ardevd I totally agree with your arguments and also vote for integrating this PR while it is something Android prevents by default with 8.x and that needed to be re-implemented so it is important to me that we explicitly decide on doing or not doing it, simply for the sake of making this a very explicit decision with as many people as possible having a say.

again cc @tobiasKaminsky for a comment/approval

tobiasKaminsky:
As I understand it, it is an opt-in where the user explicitly needs to enable installing apks from this app, same like it was previously done system-wide, but now app-wise. This is already the same at Fdroid.
If the above is true, I am fine with this as it helps people that want to install an apk, does nothing bad on people that do not want to use this...

:+1:

AndyScherzinger:
:+1:

AndyScherzinger:
Thanks a lot for the fix @ardevd :heart:

Issue #2293, opened 2018-03-05
Non-uniform functionality of back button for different activities


arundhati24:
### Actual behaviour

Currently, if we press back button while on a particular activity, the back functionality for different set of activities is different.

### Expected behaviour

The behaviour of back button should be uniform across all activities.
 
### Steps to reproduce
1. Open the app or log into the app
2. Open Activities from the navigation drawer. Then press back button. This takes us to the All files activity.
3. Open Favourites/Photos/Shared/On Device activity. Then press back button. This causes exit from the app.
4. Open Notifications/Auto Upload/Uploads/Settings/Participate activity. Then press back button. This takes us to the On Device activity.

### Environment data
Android version: 7.0

Device model: Xiaomi Redmi Note 4

arundhati24:
@mario @AndyScherzinger @tobiasKaminsky Please have a look at this issue. Do we need to work on it? :)

xXSTrikeXx:
I will appreciate it, exspecially for external sites. ;)

tobiasKaminsky:
Currently "All files / any view with server side files" is our "top" view. If you click back on this, it will close the app. 
All other activities (settings, notifications, ...) will return back to the "top" view.

I like the idea. :+1: 
In addition the settings activity should also have a slider and behave the same like the other activities.

Let's wait for another feedback and then it can get approved.

xXSTrikeXx:
Added design label since it is for the UX. Ping @nextcloud/designers

arundhati24:
sure @tobiasKaminsky thanks :)

jancborchardt:
Agree to this from a design side, good point! :)

arundhati24:
Thanks @jancborchardt :)

zc-devs:
`Nextcloud Android Files 2026-02-19`

Main view: _All files_, _Favorites_, _Media_.

- _Favorites_ and Media go back to _All files._
- _Personal files_ back click always closes the app.
- _Shared_ goes back to _All files_, does not respect previously selected main view.
- _Activities_ goes back to a main view, respects previously selected main view.
- _Uploads_ goes back to a main view, respects previously selected main view.
- _On device_ goes back to _All files_, if _Favorites_ or _Media_ were selected. If _All files_ was selected back click closes the app. Therefore, does not respect previously selected main view.
- _Deleted files_ goes back to a main view, respects previously selected main view.
- _Settings_ files goes back to a main view, respects previously selected main view.

[...]

PR #2313, opened 2018-03-07
FileDisplayActivity: onBackPressed() 

greenorca:
disabled call to base class onBackPressed since it caused closing of entire activity (at least on Motorola G5 Plus, Android7)
works way better now on my phone, unfortunately I don't have any others to test.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -53,6 +53,7 @@ import android.support.v4.view.MenuItemCompat;
 import android.support.v7.app.AlertDialog;
 import android.support.v7.widget.SearchView;
 import android.text.TextUtils;
+import android.util.Log;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1036,7 +1036,9 @@ public void onBackPressed() {
          */
 
         if (mBackToExitPressedOnce) {
-            super.onBackPressed();
+            //super.onBackPressed(); //removed, led to closing of activity
```
xXSTrikeXx: Please consider: https://github.com/nextcloud/android/issues/2293

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -53,6 +53,7 @@
 import android.support.v7.app.AlertDialog;
 import android.support.v7.widget.SearchView;
 import android.text.TextUtils;
+import android.util.Log;
```
xXSTrikeXx: The bot says it :D

greenorca:
Sorry lads, I've tested it again and found it not working that well. Closing the pull request for now, will provide a better solution later.

xXSTrikeXx:
@greenorca Feel anyway free to contribute 😸

tobiasKaminsky:
Pressing back twice to exit is already reverted, so you can exist with a single click, just like on 3.0.

PR #2330, opened 2018-03-12
Revert "Do not ignore ignoreEtags"

tobiasKaminsky:
If we rely on eTag changes we do not get notified if a file is favorited/unfavorited.
It does not affect our server side serch, but our favorite icon would be outdated as long until the parent folder is changed and thus gets a new eTag.

A change on server side is discussed [here](https://github.com/nextcloud/server/issues/8477), but will not be ready for our next version.

This needs to be in 3.1 as otherwise the behaviour is broken.


AndyScherzinger:
:+1:

PR #2348, opened 2018-03-14
Use modification date from server

tobiasKaminsky:
- store mod timestamp from server in file
- check if timestamp and length differ
- only then wait for finish saving

I hope my idea is correct?

### DIFF
--- a/src/main/java/com/owncloud/android/operations/DownloadFileOperation.java
+++ b/src/main/java/com/owncloud/android/operations/DownloadFileOperation.java
@@ -212,6 +212,7 @@ public class DownloadFileOperation extends RemoteOperation {
                 }
             }
             moved = tmpFile.renameTo(newFile);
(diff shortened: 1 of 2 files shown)


mario:
Unsure if you can rely on remote file being equal size to one of the local filesystem since it depends on a few things (namely, FS, how it's formatted, etc).

tobiasKaminsky:
If it is false positive (local size is different, but file is not downloaded), then we would unnecessary wait 1s.
The other problem around would be a problem: having the same size local and remote, but the file is still downloading. But somehow I doubt that in this very exact moment exactly the same byte length is computed...

mario:
👍

tobiasKaminsky:
@AndyScherzinger please review, so we can get it into tomorrow RC1 (slightly delayed)

AndyScherzinger:
:+1: fine by me!

PR #2350, opened 2018-03-14
Fix: Middle ellipsize in folder path in Auto Upload

### LINKED ISSUE #2280: Auto upload folder path isn't visible past 2 lines
### Actual behaviour
When doing a custom folder setup in Auto Upload, after selecting the path of the local folder, the displayed path is visible only as far as the two lines flow on the screen, the rest is clipped. For me, it usually clipped it at the end of a folder name, so it made the impression that it's always a parent folder selected, and never the exact folder I selected in the explorer. Of course, this wasn't the actual case, but it remained frustrating until I went with it and tested it to see that it actually checks the right folder.

### Expected behaviour
 The full path should be visible, broken into as many lines as needed.

### Steps to reproduce
Go to Auto Uploads and Set up a custom folder with a really long local path.

### Environment data
Android version: 5

Nextcloud app version: v3.0.2

Nextcloud server version: v12

harshika-arya:
Fixes #2280 

changes: Added single line constraint to fix Middle ellipsize in folder path of Auto Upload

Screenshot: 
screenshot_2018-03-14-18-00-47-019_com nextcloud client

@AndyScherzinger @tobiasKaminsky Please review.


AndyScherzinger:
:+1: In general fine by me, what do you think @tobiasKaminsky ? Before we already had middle-ellipsizing but maxLines=2 now it it singleLine=true not sure if that is better or if we should keep the 2 lines since in any case the path will be quite long and we definitely loose information with the single line. As suggested by the referring issue we could also show a Toast on long click or even better expand the path on long click.

@harshika-arya sorry for the confusion, I didn't know that the path has already been middle-ellipsized. :/

tobiasKaminsky:
With limiting to one line, we will loose one line of context.
I like the proposed idea in the issue: click on it, then it will expand to full lengtgh.

AndyScherzinger:
@harshika-arya would you be willing to change the PR as proposed? :heart: 

* revert to 2 lines with ellipsized-middle
* add a click listener upon which the path gets fully expanded

harshika-arya:
Sure @AndyScherzinger @tobiasKaminsky :-)

AndyScherzinger:
Thanks a lot @harshika-arya :hugs:

harshika-arya:
> add a click listener upon which the path gets fully expanded

@AndyScherzinger  can you please elaborate?
Do I have to show a dialog with the full path?

AndyScherzinger:
> Do I have to show a dialog with the full path?

The idea is that initially via the layout xml the path will be ellipsized while the TextView should then have a click listener which upon the click event on that TextView will change the maxLines attribute of the TextView which then hopefully leads to an expansion of the TextView, might need a redraw though.

AndyScherzinger:
I'll have a look when I am back at a PC.

Cc @mario @tobiasKaminsky for review

tobiasKaminsky:
:+1: nice enhancement

If we use this on other trimmed folder names, we should move the function to an util class, to reduce code duplication.

AndyScherzinger:
@tobiasKaminsky like with the other approved PRs, feel free to merge in case you create a 3.1.x branch :)

tobiasKaminsky:
@AndyScherzinger ellipsizing is also not done on master for me.
So we can merge this and should try to fix it in another PR.

tobiasKaminsky:
Issue for this is in #2506

Issue #2214, opened 2018-02-21
Already downloaded files take long to open (especially when offline)


jancborchardt:
1. Download a file
2. Go offline, or simply bad connection (or even normal connection)
3. Open the file again

You get one toast and one blocking loading modal:
![](https://i.imgur.com/B6cJJAH.png)

Depending on your connection they stay pretty long (especially when offline) and are not cancellable. This makes the app seem slow and even a bit broken.
Funnily enough, files you did not download yet show instantly, as their previews are shown.

**Instead, the file should immediately be shown with no modal or wait at all.** The updated file could be fetched in the background, and if there is one indeed, only _then_ could a toast be shown saying "Newer version available" with action "Show".

mario:
This was actually a client requirement.

jancborchardt:
Would my proposed solution interfere with that? Because this absolutely breaks the UX.

AndyScherzinger:
> Would my proposed solution interfere with that? Because this absolutely breaks the UX.

It can't be solved this way actually :/ This could be done for Text files or Images/Videos because they get opened in-app while any other file type will be send to another, external app and we can't ping that one about an updated file thus if we need to enforce an up-to-date file then we need to check before opening :/

What I don't know is how the check is actually done. ETag? Because that should be fast.

tobiasKaminsky:
This is done via SynchronizeFileOperation:
https://github.com/nextcloud/android/blob/a77c453941e640f4153193f4eb42a2e27c846a0a/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java#L249-L269

So, let us think a bit about it:
On folder entry we do check for a changed etag, which then indicates that any of the folder content has changed. But we do *not* do anything with this information.

_This would be first enhancement:_
store the information that a file is out of sync and indicate this to the user. 
On file click we then know that the file is out of sync and can sync it (either up- or download it).
@jancborchardt this is already used if a sync conflict occurs:
![image](https://user-images.githubusercontent.com/5836855/36534076-c8267926-17c5-11e8-8e95-de4812ef4cab.png)


As we check on folder entry for changes, I would assume that for certain timespan nothing has changed. Maybe 30s just like the server? So within this timespan we simply open up the file.
After this we can use the blocking "check for changes" as it is now.

Offline support must be handled better across the whole app and is out of scope to keep this rather small.

rullzer:
[...]

PR #2352, opened 2018-03-14
check walled garden only on wifi

### LINKED ISSUE #2349: Check if internet is walled only on wifi
Am I right that we can assume that any mobile data is not walled?

So then we could change the check of this to only wlan connections:
https://github.com/nextcloud/android/blob/b71c302db83dbce593195f43d7d196b9571ee8b8/src/main/java/com/owncloud/android/utils/ConnectivityUtils.java#L49-L51

tobiasKaminsky:
Fix #2349 as it is was quite quick and I wanted to test it in conjunction with #2214 ;-)
Hopefully my assumption is correct ;-)

### DIFF
--- a/src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
+++ b/src/main/java/com/owncloud/android/utils/ConnectivityUtils.java
@@ -30,6 +30,8 @@ import android.content.Context;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 
+import com.evernote.android.job.JobRequest;
+import com.evernote.android.job.util.Device;
 import com.owncloud.android.authentication.AccountUtils;
 import com.owncloud.android.lib.common.OwnCloudAccount;
 import com.owncloud.android.lib.common.OwnCloudClient;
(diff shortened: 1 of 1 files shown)


mario:
All good, except this won't work on VPN. So let's make sure you iterate all networks and check if we're connected to wifi.

tobiasKaminsky:
Can you explain this a bit more? 
The active network is VPN, but you want to check if another connected network is wifi?

mario:
Yes, because if VPN is active, it'll be active network (as you can only have one on Android really), but you're connected to VPN via wifi. So you have to check if ANY of the networks you're on is Wifi and if you're connected.

tobiasKaminsky:
I guess I understood it and hopefully implemented it correctly

mario:
OK. You don't need isOffline method. 

Instead use: Device.getNetworkType(context).equals(JobRequest.NetworkType.ANY)

If this is true, you're offline.

mario:
👍

tobiasKaminsky:
@AndyScherzinger ping :-)

AndyScherzinger:
:+1: here! ;)

AndyScherzinger:
@tobiasKaminsky all green just needs a rebase&merge :)

PR #2357, opened 2018-03-16
Wrap text properly in the grid view of All files activity (Fixed #2321)

### LINKED ISSUE #2321: Improper text wrapping in the grid view of All files activity
### Actual behaviour

Currently, the height of the text contents in the text view is greater than the height of the view itself. Due to this a fraction of the height of the text strings becomes invisible.

### Expected behaviour

The text needs to be wrapped properly within the text view.
 
### Steps to reproduce

1. Open the app and log into your account.
2.  Open the All files activity 
3. Choose Grid view from the Options menu in the Action Bar
4. You'll see the following, as in the attached screenshot.

### Environment data

Android version: 7.0

Device model: Xiaomi Redmi Note 4

### Screenshot

nextcloud_text_wrap

arundhati24:
Fixed #2321 

Changes : Earlier the image icon size was quite large due to which a fraction
of text height was invisible even though the layout_height for the
text view was set to wrap_content. Now, the image icon dimensions have
been changed from 72dp to 64dp so that the text is wrapped properly
within the text view.

**Screenshot for the change**

nextcloud_text_height

### DIFF
--- a/src/main/res/values/dims.xml
+++ b/src/main/res/values/dims.xml
@@ -42,7 +42,7 @@
     <dimen name="standard_quarter_margin">4dp</dimen>
     <dimen name="standard_eighth_margin">2dp</dimen>
     <dimen name="min_list_item_size">56dp</dimen>
-    <dimen name="standard_list_item_size">72dp</dimen>
+    <dimen name="standard_list_item_size">64dp</dimen>
     <dimen name="two_line_primary_text_size">16sp</dimen>
     <dimen name="two_line_secondary_text_size">14sp</dimen>
     <dimen name="list_item_avatar_icon_margin">20dp</dimen>


tobiasKaminsky:
:+1: 
Wow, never realized that this is happening.

arundhati24:
@tobiasKaminsky Sometimes we do not realise but we miss on some minute problems while working on a bigger project. ;) :slightly_smiling_face:

AndyScherzinger:
:+1: Eagle eyes! :+1:

AndyScherzinger:
Well, one question remains, if the standard list item size is changed what happens to the List views? BEcause the list views should have an item size of 72dp by material design definition!

arundhati24:
That's right @AndyScherzinger 
What we can do is, we can define different sizes for grid item and list item in the dimens.xml and reference them accordingly. Would that be fine?

tobiasKaminsky:
What happens/should happen to the text size when changing the grid size, e.g. from 3 columns to 2 (which is min) or 5 (max).

arundhati24:
Is that in context of our Grid view (do we allow our user to choose the number of columns? I couldn't find that customization) or is that a general question @tobiasKaminsky ? :)

tobiasKaminsky:
Ahh, sorry @arundhati24 :tada: 
I forgot that you are not that familiar (yet) with our app: via pinch zoom you can change column count from 2-5.

arundhati24:
@tobiasKaminsky yes.. I am new here. Need little time :)
But I just checked it out. Here are a few things I saw : 
screenshot_2018-03-19-18-32-03-365_com nextcloud client

screenshot_2018-03-19-18-31-33-451_com nextcloud client

screenshot_2018-03-19-18-31-45-529_com nextcloud client

screenshot_2018-03-19-18-31-11-720_com nextcloud client

No text when 5 columns.

arundhati24:
@tobiasKaminsky yes.. I am new here. Need little time :)
But I just checked it out. Here are a few things I saw :

arundhati24:
screenshot_2018-03-19-18-31-11-720_com nextcloud client
 
No text when 5 columns.

arundhati24:
screenshot_2018-03-19-18-31-33-451_com nextcloud client

Strange behavipir of the first  item. Sometimes text disappears. Sort of flickers while we zoom it.

arundhati24:
screenshot_2018-03-19-18-32-03-365_com nextcloud client

For two columns.
That flickering of text for first grid item is happening in almost all the cases.

arundhati24:
Also for two columns icons are very small. Most of the screen is empty. It doesn't look that good imho.

tobiasKaminsky:
Indeed, this looks not that good if you have different file types. It is mainly for large image folders, like your camera folder.
I think we can & should discuss how to improve this in a different issue, also with designers, if needed.

I pointed this out only in this issue so that we verify if the changed text size is also working with 2-5 columns.
It is out of scope of this PR to enhance the grid view.

mario:
@AndyScherzinger @tobiasKaminsky is this ready for review or?

AndyScherzinger:
@mario In general it is ready for review while I disapproved since `standard_list_item_size` is used for the standard material list item height so it should not be changed just to get grid items right but like mentioned by @arundhati24 then use a different, to be introduced value for the grid items. So in the current state the PR should not be merged.

arundhati24:
Yes.. @AndyScherzinger  Should I introduce that change?
Then, after this as mentioned by @tobiasKaminsky we can work on enhancing the grid view for different number of columns in a separate issue.

AndyScherzinger:
@arundhati24 yes perfectly fine by me, beware though that @tobiasKaminsky is actually complete rewriting the lists and grids to recyclerview. @tobiasKaminsky will this PR then be rendered obsolete?

arundhati24:
@AndyScherzinger I think this PR won't become obsolete because even though recycler view is an enhancement over list view, in both the cases we would have to define our list row layout. This icon size is a part of that. What do you suggest @tobiasKaminsky ?

AndyScherzinger:
@arundhati24 you are absolutely right! :+1:  this PR should fix the issue via the layouts not the list view implementaiton itself...sorry for not having seen that in the first place!

arundhati24:
So, should I go ahead and fix it? @AndyScherzinger @tobiasKaminsky :)

AndyScherzinger:
In my opinion, sure :+1:

arundhati24:
@AndyScherzinger @tobiasKaminsky Actually, there is no need to change the standard_list_item_size. That is the size of the list row item i.e. the height of the list row layout. What we need to change is the file icon size. In the list_item.xml the dimensions of the file icon are defined by file_icon_size while in the grid_item.xml the dimensions have been defined by standard_list_item_size. Thus, according to me we simply need to rename file_icon_size as file_icon_size_list_item, and introduce a new dimension called file_icon_size_grid_item which would define the size of the grid item. Everything else remains the same. :)

AndyScherzinger:
Sounds good and right to me @arundhati24 :+1:

tobiasKaminsky:
> introduce a new dimension called file_icon_size_grid_item which would define the size of the grid item

Will this then still have auto-scale the image on grid view, if you use only 2 or 3 columns?

arundhati24:
@tobiasKaminsky Everything would remain the same as it is now. What would change is just the size of the grid view file icon. I tried to implement and check it, that problem still persists when we change the number of columns. As you had mentioned, we will have to look into the auto-scaling of icons as we change the number of columns in another PR. Shall I open an issue for that? We can then figure out how to implement that. Or do you have an idea how to implement the auto scale feature as we change the number of columns? :)

tobiasKaminsky:
Yes, please open up a new issue, where we can discuss this.
Currently I have no idea, but hopefully we will find together a way to solve it.

tobiasKaminsky:
@AndyScherzinger you are the UI expert. What do you think of this PR?

AndyScherzinger:
Hard to say, we do need auto scaling for this to work well, but I didn't find the time to do so, so I guess we need to start over to get this right even though I can't say how to achieve this right away.

tobiasKaminsky:
Ok, then let us close this and take it as inspiration when we want to resume it.
(therefore not deleting the branch)

PR #2360, opened 2018-03-16
Hiding the action bar in case of fullscreen video playback

### LINKED ISSUE #1971: 3.0RC1: Videoplayback in full-screen isn't really full-screen
I was checking 3.0RC1 towards my earlier reported bug #1761 that has been greatly resolved - many thanks to @AndyScherzinger 

Another finding that I recognized after I reported #1761 and still existing in 3.0RC1 is that Videoplayback in full-screen isn't really full-screen. There is always a thick bar labeled "Nextcloud" that appears unnecessary to me:

skizze

AndyScherzinger:
Fixes #1971 

cc @mario @tobiasKaminsky

### DIFF
--- a/src/main/java/com/owncloud/android/ui/preview/PreviewVideoActivity.java
+++ b/src/main/java/com/owncloud/android/ui/preview/PreviewVideoActivity.java
@@ -104,6 +104,10 @@ public class PreviewVideoActivity extends FileActivity implements OnCompletionLi
           
         // keep the screen on while the playback is performed (prevents screen off by battery save)
         mVideoPlayer.setKeepScreenOn(true);
+
+        if(getSupportActionBar() != null) {
+            getSupportActionBar().hide();
+        }
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
I tested it on my OP3T Android 8

AndyScherzinger:
Would also be nice to get this into 3.1 since it is a bug fix for something it seems we introduced with 3.0 and is bad for video previews UX wise.

tobiasKaminsky:
:+1: 
Working also on android 6.x on Samsung S5 Neo.

src/main/java/com/owncloud/android/ui/preview/PreviewVideoActivity.java
```diff
@@ -105,7 +105,7 @@ public void onCreate(Bundle savedInstanceState) {
         // keep the screen on while the playback is performed (prevents screen off by battery save)
         mVideoPlayer.setKeepScreenOn(true);
 
-        if(getSupportActionBar() != null) {
+        if (getSupportActionBar() != null) {
```
AndyScherzinger: :heart:

AndyScherzinger:
@mario ping ;-)

mario:
👍

PR #2367, opened 2018-03-19
Use long instead of integer

tobiasKaminsky:
I discovered this with dev version, so yeah for dev having this ;-)

Integer is too small. 
Long is also not unlimited, but if number is too big, then it falls back to string comparison, which might not be exact, but at least does not lead to an Exception.

Please in 3.1 as it is a bugfix.

### DIFF
--- a/src/main/java/third_parties/daveKoeller/AlphanumComparator.java
+++ b/src/main/java/third_parties/daveKoeller/AlphanumComparator.java
@@ -123,35 +123,41 @@ public class AlphanumComparator<T> implements Comparator<T>, Serializable {
                 // extract digits
                 int thisChunkZeroCount = 0;
                 boolean zero = true;
-                int c = 0;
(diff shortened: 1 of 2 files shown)


AndyScherzinger:
:+1: fine by me

mario:
👍

tobiasKaminsky:
@mario @AndyScherzinger has anyone changed something with drone web hooks? We have now two:
continuous-integration/drone/pr ---> working
continuous-integration/drone/push --> always waiting

mario:
@tobiasKaminsky I temporarily had to unprotect the master, and then protected it back. Didn't touch the hooks.

tobiasKaminsky:
Ok, the master branch checks were a bit messed.

MorrisJobke:
> @mario @AndyScherzinger has anyone changed something with drone web hooks? We have now two:

https://github.com/nextcloud/android/pull/2183#issuecomment-374495167 ;)

PR #2371, opened 2018-03-19
Remove push for account

### LINKED ISSUE #2167: Push notification is still received if account is removed
- add account
- test push notification -> working
- remove account (no account available now)
- test push notification -> still receiving

On AccountRemovalJob we should test if this is the last account and then remove the push key.
Or do we just want to suppress showing the notification?

tobiasKaminsky:
Should fix #2167 

https://github.com/nextcloud/android/blob/87802615c2c5054d6fe6137c3796edfe9a93672c/src/gplay/java/com/owncloud/android/utils/PushUtils.java#L158-L158

is false, so that the unregister operation is not called.
@mario as you developed this, do you have an idea?

### DIFF
--- a/src/generic/java/com/owncloud/android/utils/PushUtils.java
+++ b/src/generic/java/com/owncloud/android/utils/PushUtils.java
@@ -52,7 +52,4 @@ public final class PushUtils {
     }
 
 
-    public static void deleteRegistrationForAccount() {
-        // do nothing
-    }
 }
--- a/src/generic/java/com/owncloud/android/utils/PushUtils.java
+++ b/src/generic/java/com/owncloud/android/utils/PushUtils.java
@@ -51,4 +51,8 @@ public final class PushUtils {
         return null;
     }
 
+
+    public static void deleteRegistrationForAccount() {
+        // do nothing
+    }
 }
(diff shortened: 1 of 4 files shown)


mario:
Well this is because the account should be deleted after this was ran?

mario:
You don't need to call deleteRegistrationForAccount since it's already called via pushRegistrationToServer when setShouldBeDeleted is true, and it is for accounts which are pending deletion.

mario:
Due to changes in https://github.com/nextcloud/android/pull/2372, push won't be shown for accounts which are not active. But I'll investigate why they're not unregistered for 3.2.

tobiasKaminsky:
@mario had you already a chance to find out why the devices are not unregistered?

mario:
Rebased and slightly updated the PR. Might work, even though it's not perfect so please test.

If the unregistration fails, we still remove the account so next time we try to unregister it'll fail. But oh well.

tobiasKaminsky:
Where and how is push unregistration now started?

tobiasKaminsky:
I can't reproduce this on latest master, even without this PR.
So close this, and the related issue.

PR #2373, opened 2018-03-19
Check if we’re on unmetered wifi

mario:

Make sure we only sync offline files via job when on unmetered.

3.1 please.


tobiasKaminsky:
:+1: those many restrictions will drive me crazy at some point ;-)

AndyScherzinger:
:+1: makes total sense! :+1:

PR #1657, opened 2017-10-11
Add support for device credentials

rhari991:
Fixes #1286 

ToDo:
- [x] fix #1924 
- [x] verify if https://github.com/nextcloud/android/issues/1879 happens also here
- [x] dynamically tint list preference

[Reference](https://github.com/googlesamples/android-ConfirmCredential)

PR #2381, opened 2018-03-21
Show mnemonic based on device credentials

### LINKED ISSUE #2128: How can I display the 12 word key E2E on android apps?
I have two android phones with nextcloud installed on the first access to encrypted folder and not having marked the key on the second mobile with the same account asks me the password to access the same folder?

I do not remember how I can do it? on app ios of a friend of mine there is voice display key would serve a function like that to remember it, or there is a way to get it back or reset it?

tobiasKaminsky:
Fix #2128 

* [x] This is based on #1657, but not showing the diff correctly as #1657 is another repo. So once #1657 is merged, this will have a cleaner diff.
* [x] rebase to latest `master`

- show preference entry only if e2e mnemonic is saved
- showing mnemonic is only possible if device credentials are setup in android

- if enabled, show credentials request
	resized_screenshot_1521625105
- then show mnemonic

### DIFF
--- a/src/main/java/com/owncloud/android/authentication/PassCodeManager.java
+++ b/src/main/java/com/owncloud/android/authentication/PassCodeManager.java
@@ -95,14 +95,6 @@ public class PassCodeManager {
             activity.startActivityForResult(i, PASSCODE_ACTIVITY);
         }
 
-        if (!sExemptOfPasscodeActivites.contains(activity.getClass()) && Build.VERSION.SDK_INT >=
-                Build.VERSION_CODES.M && deviceCredentialsShouldBeRequested() &&
-                !DeviceCredentialUtils.tryEncrypt()) {
-            Intent i = new Intent(MainApp.getAppContext(), RequestCredentialsActivity.class);
-            i.setFlags(Intent.FLAG_ACTIVITY_REORDER_TO_FRONT);
-            activity.startActivity(i);
-        }
-
         visibleActivitiesCounter++;    // keep it AFTER passCodeShouldBeRequested was checked
     }
 
--- a/src/main/java/com/owncloud/android/authentication/PassCodeManager.java
+++ b/src/main/java/com/owncloud/android/authentication/PassCodeManager.java
@@ -95,6 +95,14 @@ public class PassCodeManager {
             activity.startActivityForResult(i, PASSCODE_ACTIVITY);
         }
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/owncloud/android/authentication/PassCodeManager.java
```diff
@@ -65,7 +68,7 @@ public static PassCodeManager getPassCodeManager() {
     protected PassCodeManager() {}
 
     public void onActivityCreated(Activity activity) {
-        if (passCodeIsEnabled() || fingerprintIsEnabled()) {
+        if (passCodeIsEnabled() || deviceCredentialsAreEnabled(activity)) {
```
mario: So there's passcode AND device credentials which also have passcode?

mario:
Code wise good, but I do have one question so please answer 👍

src/main/java/com/owncloud/android/authentication/PassCodeManager.java
tobiasKaminsky: Correct, passcode is our old system, which can be enabled independent from device credentials and also works on older devices.

rullzer:
Unrelated to this. But maybe store the mnemonic not as a space less string. So the output here is still readble ;)

AndyScherzinger:
:+1: fine by me

tobiasKaminsky:
> Unrelated to this. But maybe store the mnemonic not as a space less string. So the output here is still readble ;)

Not possible. We are storing what we get from the user...


Before we merge this, we should merge #1657

AndyScherzinger:
@tobiasKaminsky are you able to rebase this one? (tricky one...)

tobiasKaminsky:
Security wise: mnemonic is stored as plain text in database, but this is only accessible when root. So should be fine?
@rullzer what are you doing on client side?

rullzer:
>Security wise: mnemonic is stored as plain text in database, but this is only accessible when root. So should be fine?
@rullzer what are you doing on client side?

Client stores it in the keychain. Which is encrypted. However of course if you have full control of the system there are always ways to get it.

AndyScherzinger:
@mario rebased :)

jospoortvliet:
no way to show the spaces, btw?

tobiasKaminsky:
@jospoortvliet I created https://github.com/nextcloud/android/pull/3013 for you ;-)

PR #2383, opened 2018-03-21
context might be null, so fallback

tobiasKaminsky:
Reported via google play console.

We are using this in all static theming calls.
It seems to happen, maybe due to race conditions, that context is null (MainApp not entirely set up yet), but there is already a call to theming.

This prevents from having a NPE within AccountUtils.getCurrentOwnCloudAccount(context) when context is null.
Instead we simply return a default OCCapability.

This function is only used from within ThemeUtils, so the "worst" that could happen is that we fallback to the default theming.

### DIFF
--- a/src/main/java/com/owncloud/android/utils/ThemeUtils.java
+++ b/src/main/java/com/owncloud/android/utils/ThemeUtils.java
@@ -380,16 +380,19 @@ public class ThemeUtils {
 
     private static OCCapability getCapability(Account acc) {
         Account account;
+        Context context = MainApp.getAppContext();
+
+        if (context == null) {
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
:+1:

mario:
I hate this MainApp.getContext :( 👍

tobiasKaminsky:
> I hate this MainApp.getContext :( ￼

Yeah, we should really try to get rid of this and replace those static calls one by one with dynamic ones...

tobiasKaminsky:
A start to get rid of getAppcontext is in #2394, but still ~250 occurrences to go...

PR #2385, opened 2018-03-21
Drawer: check if account is null

tobiasKaminsky:
- exit if account is null
- use better context, to prevent nullable context


AndyScherzinger:
:+1: fine by me

mario:
👍

PR #2386, opened 2018-03-21
Re-assign accountmanager

tobiasKaminsky:
This is kind of strange:
Here it works without any problems: https://github.com/nextcloud/android/blob/88cf975861657b3c952e03e282c6616120c2a8c4/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java#L193-L197

But here 
https://github.com/nextcloud/android/blob/88cf975861657b3c952e03e282c6616120c2a8c4/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java#L1012-L1015
it crashes.
Therefore I re-assign AccountManager...

If you have any better idea...
Reported via google play console:
all 7.x
all Galaxy devices. Samsung...oh...Samsung... :fearful: 

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
@@ -171,6 +171,7 @@ public class ReceiveExternalFilesActivity extends FileActivity
             
             mFile = savedInstanceState.getParcelable(KEY_FILE);
         }
+        mAccountManager = (AccountManager) getSystemService(Context.ACCOUNT_SERVICE);
 
         super.onCreate(savedInstanceState);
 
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
:+1: oh, well.... :/

mario:
Can you instead just create mAccountManager in onCreate once, and use it elsewhere? Thanks!

mario:
Poke?

AndyScherzinger:
pinging @tobiasKaminsky :)

tobiasKaminsky:
Well, as said the assignment is already done in setAccount(), but for $reason it is null in isHaveMultipleAccount(). So I am not sure if it would help to move the assignment into onCreate(), but we can try it.

tobiasKaminsky:
It has to be prior super.onCreate() as in this method setAccount is called.

mario:
rebase @tobiasKaminsky

AndyScherzinger:
@mario  rebased :)

mario:
👍

PR #2387, opened 2018-03-21
OnServiceConnnected: File not found

tobiasKaminsky:
Reported via google console

- file is updated, but somehow not anylonger in storagemanager
- check to prevent NPE on `!mWaitingToPreview.isDown()`
- add @Nullable to warn via IDE
(as somewhere discussed it is better to handle this via a real FileNotFoundException, but for now...)


AndyScherzinger:
:+1: alrighty

mario:
*shrug* 👍

PR #2394, opened 2018-03-22
Any theming call has now context, instead of app context

tobiasKaminsky:
With this any theming call uses now the context from the calling instance.
Only in both service (FileDownloader, FileUploader) getApplicationContext is used.

We are using theming on that many calls that the PR is such big :/


AndyScherzinger:
:+1:

mario:
Build fails?

AndyScherzinger:
:+1: fine by me, rebased and fixed all api changed and lint warnings

mario:
👍

PR #2404, opened 2018-03-27
Resume activity crash due to grid view check

tobiasKaminsky:
If the app is suspended and re-activated, it crashed as it tries to use adapter when it was not set, therefore:
- move scroll/switchGrid/List to onActivityCreated
- check if adapter is already set
- fix npe on storage manager 

To check, this can be enabled in developer options:


AndyScherzinger:
👍 

Strange thing are the 3 new lint warnings?!

tobiasKaminsky:
> Strange thing are the 3 new lint warnings?!

1x gradle warning due to newer version
2x replace "..." with "…"

I fixed this in #2405

mario:
👍

PR #2406, opened 2018-03-27
New media detection

### LINKED ISSUE #1447: Feature request: "Would you like to auto upload this new folder"
A feature of Google Photos I like, because I'm lazy mostly, is the notification that alerts me Photos has found a new folder on disk, and if I'd like to start backing it up.

This would be a _great_ addition in promoting continued use of the NC app whenever new media is found.

mario:
Fixes  #1447

Detects new media folder and notifies users about it.

### DIFF
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -44,6 +44,7 @@ import android.text.TextUtils;
 import android.view.WindowManager;
 
 import com.evernote.android.job.JobManager;
+import com.evernote.android.job.JobRequest;
 import com.owncloud.android.authentication.AccountUtils;
 import com.owncloud.android.authentication.PassCodeManager;
 import com.owncloud.android.datamodel.ArbitraryDataProvider;
@@ -56,6 +57,7 @@ import com.owncloud.android.datamodel.ThumbnailsCacheManager;
 import com.owncloud.android.datastorage.DataStorageProvider;
 import com.owncloud.android.datastorage.StoragePoint;
 import com.owncloud.android.db.PreferenceManager;
+import com.owncloud.android.jobs.MediaFoldersDetectionJob;
 import com.owncloud.android.jobs.NCJobCreator;
 import com.owncloud.android.lib.common.OwnCloudClientManagerFactory;
 import com.owncloud.android.lib.common.OwnCloudClientManagerFactory.Policy;
@@ -77,6 +79,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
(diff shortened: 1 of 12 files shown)


mario:
This is now ready for review folks! @AndyScherzinger @tobiasKaminsky

mario:
Ummm so I checked lint report, but don't see what new warnings are?

tobiasKaminsky:
You have to compare it with master.html which is in the same directory.
There is a lint tool to use current master as baseline, which then should only show the difference, but this does not work, so we have to compare it manually :/

mario:
Looks ok now @tobiasKaminsky. Please test and review.

AndyScherzinger:
I tested it installing OpenCamera and that didn't completely work for me.

* Works: Detection of new Folder :+1: 
* Failed: Directly configuring the new folder via the notification :( 

The configuration is active while the pictures taken after don't get detected / uploaded.

mario:
@AndyScherzinger could you look into NCContentObserverJob and see what syncedFolderProvider.countEnabledSyncedFolders() returns there?

mario:
@AndyScherzinger @tobiasKaminsky fixed for real this time. Please rebase and test. Thanks!!!

AndyScherzinger:
Can't test right now (time-wise) but did the rebase :)

mario:
Anyone tested this?

AndyScherzinger:
works for me tested with Simply Camera and Open Camera with small issues:

* back arrow is shown in dark color instead of white
* hitting that back button brings up the settings screen
* hitting the back button there restarts the app
* Auto Upload UI list doesn't get refreshed right away (activating the new folder for auto upload still shows it as inactive in the auto upload list, I need to open the auto upload screen again (after the back arrow hitting restart).
device-2018-04-22-222411

mario:
@AndyScherzinger as I mentioned in the IRC unsure if this is related to the PR itself since I didn't touch those parts xD

AndyScherzinger:
@mario I know but that might just be because the activity hasn't been used in the way this PR does so the issues that I found don't happen in the existing cases ;)

tobiasKaminsky:
If if happens now due to this PR, I think it should also be fixed here, even if the bug/bad UX was not introduced by this PR, but only revealed by it.

AndyScherzinger:
As discussed with @mario on IRC I'll take it from here and plan to fix the issues raised within the reviews.

AndyScherzinger:
Imho now ready to review @tobiasKaminsky 

I fixed:
* Directly configuring the new folder via the notification (now calculation the index in the list)
* Fixing the navigation issues described in https://github.com/nextcloud/android/pull/2406#issuecomment-383409879
* Fixing the codacy issues that now came up due to the more restrictive checks due to the activation of pmd

AndyScherzinger:
:+1: approved from my side while I of course added / rewrote some code to fix the issues so pinging @tobiasKaminsky and @ardevd for second review :)

mario:
👍 

cc @AndyScherzinger @tobiasKaminsky

tobiasKaminsky:
:+1:

jasonbayton:
Oh this is exciting! Thanks folks!

PR #2412, opened 2018-03-29
Bump to 3.1.0 RC2

tobiasKaminsky:
RC2 contains only two bug fixes.


tobiasKaminsky:
Drone seems to be already trying to find easter eggs...

PR #2426, opened 2018-04-03
Prevent IllegalArgumentException on generating preview

tobiasKaminsky:
- enable "don't keep activities" on developer settings, otherwise you would have to wait long
- open detail view on image
- minimize app
- restore app
-> crash on generating preview due to nullable storage manager

FileDetailFragment is called twice
- once via onStart, which is too early
- once via regular cycle, where everything is setup

(reported via google play console, but not a regression, so only for next version, not requiring a new RC)


AndyScherzinger:
:+1:

AndyScherzinger:
@mario please review ;-)

mario:
👍

PR #2427, opened 2018-04-03
FileDisplayActivity: Replaced duplicate subexpression in if statement with instantVideoUploadEnabled preference.

ardevd:
It makes no sense to have identical expressions on both side of the binary operator, but I guess this is pretty self explanatory.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -288,7 +288,8 @@ public class FileDisplayActivity extends HookActivity
      */
     private void upgradeNotificationForInstantUpload() {
         // check for Android 6+ if legacy instant upload is activated --> disable + show info
-        if (PreferenceManager.instantPictureUploadEnabled(this)) {
+        if (PreferenceManager.instantPictureUploadEnabled(this) ||
+                PreferenceManager.instantVideoUploadEnabled(this)) {
 
             // remove legacy shared preferences
             SharedPreferences.Editor editor = PreferenceManager.getDefaultSharedPreferences(this).edit();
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@ardevd @tobiasKaminsky not sure about this change since I'd say instead of removing the duplicate check it should check for `instantVideoUploadEnabled` since it checks for legacy instant upload.

ardevd:
But the check is still there?

AndyScherzinger:
@ardevd doesn't the check only contain the instant _image_ upload preference check instead of image _and_ video preference check? cc @tobiasKaminsky

mario:
The second one should be instantVideoUploadEnabled @ardevd. Dunno how this bug persisted here for so long...

tobiasKaminsky:
this should indeed be:
```
if (PreferenceManager.instantPictureUploadEnabled(this) ||
                PreferenceManager.instantVideoUploadEnabled(this)) {
...
}
```

AndyScherzinger:
Thanks @ardevd :)

@jancborchardt can you add @ardevd to the oraginaztion please? :heart:

ardevd:
Guess my code cleanup hunting revealed a sneaky bug then. :) It should be good now!

AndyScherzinger:
:+1: looks good to me :)

> Guess my code cleanup hunt revealed a sneaky bug then. :)

Looks like, yes :+1:

tobiasKaminsky:
It is not a real bug, as this does not harm anything, but the app is not working as expected, so all in all it is very good that you found it :tada:

AndyScherzinger:
@tobiasKaminsky so is your comment an approval? :grinning:

mario:
👍 

@AndyScherzinger merge.

PR #2429, opened 2018-04-03
Update to Android Studio 3.1 tooling including build fixes

AndyScherzinger:
like the title said, PR updates config to Android Studio 3.1 which also enforces several gradle / build tools upgrades which then lead to further necessary changes.


AndyScherzinger:
cc @mario @tobiasKaminsky for review. Beware that this means you should be running Android Studio 3.1 :rocket:

AndyScherzinger:
@Bubu do you know if this would case issues with the f-droid builds since Android Studio 3.1 is available since one week thus the new build tools, gradle version etc. are also not that wide spread with OSS Android app repos?

Bubu:
@AndyScherzinger should be good. We have platform 27, build-tools 27.0.3 and gradle 4.4. The gradle android plugin is downloaded anyway each build.

AndyScherzinger:
Thanks for the Feedback @Bubu - sounds good! :smiley:

AndyScherzinger:
When merged we should also upgrade the docker config for the new toolset while I never got lucky doing so... 

Could you take care of that @tobiasKaminsky please? :heart:

tobiasKaminsky:
What happens if we merge this and someone with <AS 3.1 wants to use our project?
We should make sure that we do not break others dev environment, it it is not needed.

AndyScherzinger:
I agree with @mario, keeping a legacy setup doesn't make sense so devs should just upgrade their ide. Same if you upgraded already then you have to do the changes within this PR everytime over and over again which makes even less sense ;)

AndyScherzinger:
As discussed with @tobiasKaminsky on IRC also approved by him. Rebased, just waiting for the build & checks to turn green for merge.

PR #2433, opened 2018-04-04
Removal of change password menu entry

### LINKED ISSUE #2229: Remove "change password"
Change password is currently changing only the password with which the app authenticates to the server.
If this password is not valid anylonger, e.g. when the auth token is revoked, the user gets the Authenticator/Login screen presented. So this is all fine.

Currently when clicking on "change password" this just opens up AuthenticatorActivity / Weblogin flow, which can change the internal stored password, but as seen above, this is useless.
A real password changing is only possible via webUI and should only be done there, especially if other user backends are handled there.

Therefore I suggest to remove the "change password" option.

AndyScherzinger:
Resolves #2229 by removing the change password option in the user info screen

cc @mario @tobiasKaminsky

### DIFF
--- a/src/main/res/menu/user_info_menu.xml
+++ b/src/main/res/menu/user_info_menu.xml
@@ -1,10 +1,24 @@
 <?xml version="1.0" encoding="utf-8"?>
-<menu xmlns:android="http://schemas.android.com/apk/res/android">
+<!--
+  Nextcloud Android client application
 
-    <item
-        android:id="@+id/change_password"
-        android:title="@string/change_password">
-    </item>
+  Copyright (C) 2017 Andy Scherzinger
+  Copyright (C) 2017 Nextcloud
+
+  This program is free software; you can redistribute it and/or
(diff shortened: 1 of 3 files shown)


tobiasKaminsky:
:+1:

AndyScherzinger:
@mario ping for review ;-)

src/main/res/menu/user_info_menu.xml
```diff
@@ -1,10 +1,24 @@
 <?xml version="1.0" encoding="utf-8"?>
-<menu xmlns:android="http://schemas.android.com/apk/res/android">
+<!--
+  Nextcloud Android client application
 
-    <item
-        android:id="@+id/change_password"
-        android:title="@string/change_password">
-    </item>
+  Copyright (C) 2017 Andy Scherzinger
```
mario: 2018 now? xD

mario:
One small comment, then merge.

PR #2434, opened 2018-04-05
Disable TypographyEllipsis lint check

tobiasKaminsky:
Currently we have lots of lint warnings for this:

> You can replace the string "..." with a dedicated ellipsis character, ellipsis character (…, &#8230;). This can help make the text more readable.

Instead of having this as an warning, I would like to have this as an info in Transifex, but not letting lint fail.

### DIFF
--- a/lint.xml
+++ b/lint.xml
@@ -78,5 +78,63 @@
         <ignore path="**/values-cs-rCZ/strings.xml"/>
         <ignore path="**/values-en-rGB/strings.xml"/>
     </issue>
+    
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
👍 in general fine by me even though this means that it'll also not fail if this is done wrong on the master string.xml, but we should have that covered via the code reviews 👍

tobiasKaminsky:
I can change it, so that only the translated files are skipped.
I also asked some of our transifex gurus how to do this properly on transifex.

mario:
👍

PR #2442, opened 2018-04-09
FileContentProvider: limit calling package check

tobiasKaminsky:
As calling package is only the package name, we can use equals instead of contains, so we can prevent other apps like com.nextcloud.client2


AndyScherzinger:
:+1: nice catch!

mario:
👍

PR #2448, opened 2018-04-09
Added a splash screen

### LINKED ISSUE #2444: Add splash screen for Nextcloud
When you do not have Nextcloud app running already on your Android phone and you turn it on, you will receive a blank screen initially while the application starts up. My recommendation is to add a splash screen which can then be customized by anyone who wishes to compile their version. This splash screen would show up while the application initializes.

ardevd:
Resolves #2444

Added a splash screen. This is done by defining a launch screen theme with an associated XML resource that configures the appearance. Then we associate the launch screen theme with the launcher activity (`FileDisplayActivity`) in the manifest. Finally, we set the default theme in `onCreate()` before super in `FileDisplayActivity` so that it's applied when the activity is ready to the view is set. 

This approach is better than implementing the launch screen as a separate activity since we dont get the negative performance impact of adding an additional activity transition every time the app is started.

I used the default logo drawable which I think looks fine but I'm sure the designers should chime in and give their input. Screenshot is attached.

screenshot_20180409-220616

This fixes #2444

### DIFF
--- a/src/main/res/drawable/launch_screen.xml
+++ b/src/main/res/drawable/launch_screen.xml
@@ -1,4 +1,22 @@
 <?xml version="1.0" encoding="utf-8"?>
+<!--
+  Nextcloud Android client application
+
+  Copyright (C) 2018 Edvard Holst
+
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
+  License as published by the Free Software Foundation; either
+  version 3 of the License, or any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+
+  You should have received a copy of the GNU Affero General Public
+  License along with this program. If not, see <http://www.gnu.org/licenses/>.
+-->
 <layer-list xmlns:android="http://schemas.android.com/apk/res/android">
 
     <item
--- a/src/main/res/drawable/launch_screen.xml
+++ b/src/main/res/drawable/launch_screen.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <item
+        android:drawable="@color/primary"/>
+
+    <item>
+        <bitmap
+            android:gravity="center"
+            android:src="@drawable/logo" />
+    </item>
+
+</layer-list>
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -80,7 +80,7 @@
             android:name=".ui.activity.FileDisplayActivity"
             android:label="@string/app_name"
             android:configChanges="orientation|screenSize"
-            android:theme="@style/Theme.ownCloud.Toolbar.Drawer">
+            android:theme="@style/Theme.ownCloud.Launcher">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
 
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -182,7 +182,8 @@ public class FileDisplayActivity extends HookActivity
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         Log_OC.v(TAG, "onCreate() start");
-
+        // Set the default theme to replace the launch screen theme.
+        setTheme(R.style.Theme_ownCloud_Toolbar_Drawer);
         super.onCreate(savedInstanceState); // this calls onAccountChanged() when ownCloud Account
         // is valid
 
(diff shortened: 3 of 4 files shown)


src/main/res/drawable/launch_screen.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
```
AndyScherzinger: Please add a license / copyright header with your name like:
```
<!--
  Nextcloud Android client application

  Copyright (C) 2018 YOUR NAME

  This program is free software; you can redistribute it and/or
  modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
  License as published by the Free Software Foundation; either
  version 3 of the License, or any later version.

  This program is distributed in the hope that it will be useful,
  but WITHOUT ANY WARRANTY; without even the implied warranty of
  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
  GNU AFFERO GENERAL PUBLIC LICENSE for more details.

  You should have received a copy of the GNU Affero General Public
  License along with this program. If not, see <http://www.gnu.org/licenses/>.
-->
```

AndyScherzinger:
:+1: looks good code-wise pinging @mario @tobiasKaminsky for code review and @nextcloud/designers @jancborchardt for design review. Since it is the splash screen I am unsure if blue-background/white-logo is better or rather white-background/blue-logo but I'll leave that to the designers. I am fine either way.

src/main/res/drawable/launch_screen.xml
ardevd: What name should I be putting? Just one line with my own name or two lines with my own name and Nextcloud 2018?

AndyScherzinger: Just or own since you are the author so by definition you own the copyright and nobody else (of course only until somebody else modifies the file than the copyright is shared/split) :)

ardevd: Cool. Done. :)

c0fe:
@AndyScherzinger I think the Blue background and white logo is better, it fits with the default install and color scheme of Nextcloud.

jancborchardt:
Looks nice! :) Can you only move the logo a little bit up? Vertical centering tends to visually look a bit like the icon is too far down. :)

And yes, it should be white on blue like this. It’s how the app icon you tapped looks like, and how the header looks like too.

ardevd:
@jancborchardt Sure, I can push the logo up a bit. I'm glad you agree that white on blue is the way to go. Looks better by far in my opinion :)

ardevd:
@jancborchardt Actually, I'm not entire sure how I'd go about pushing the view up from the center. I'm using `android:gravity="center"` on the bitmap item. I'm unsure how I would go about pushing the view up.

AndyScherzinger:
@ardevd is is possible to add a padding or margin Bottom to the bitmap definition in the xml, if so that could push it up a little, maybe 16dp or something like that.

simonspa:
Just another thought: the rest of the app is taking the theming color of the server instance it is connected to - would this also be possible for the splash screen?

MariusBluem:
@simonspa Makes only limited sense, since the splash screen is shown in situations in which the color from the theming-app is not fetched. If you have multiple instances accounts connected with the Nextcloud Android app, which color should be used? It’s much simpler to just use the nextcloud blue here :)

simonspa:
@MariusBluem fair point. Just thought the color would probably be cached and not fetched from the Nextcloud instance on every start. Also, most users probably have one account set up - otherwise the color could be the one of the currently "active" account (i.e. the one displayed after startup).

Not very important though. :)

ardevd:
@simonspa It's a cool suggestion but it adds a lot of complexity to an otherwise really simple dynamic. At the end of the day I think it makes more sense to consider the splash screen part of the Nexctloud branding just like the app icon. 

@AndyScherzinger are we happy with this PR? I could add padding but it would probably involve adding a new drawable just for that purpose. I think the current implementation looks fine but it's up to you guys :)

jancborchardt:
@simonspa good suggestion. It should show the color of the currently active account.

However, that's a separate pull request since as @ardevd said it makes it more complex. :)

@ardevd if it's too much work pushing the logo up, it's also fine to leave it for now. Step by step :)

ardevd:
Then I would suggest merging this. If someone wants to work on the icon then they are free to submit a PR for that. :)

AndyScherzinger:
@ardevd I already approved the PR, but it needs a second developer approval by either @mario or @tobiasKaminsky and since I labeled it "Design" it needs an approval by one of the @nextcloud/designers while we could argue that @jancborchardt's last comment can be seen as such.

AndyScherzinger:
:+1: re-approving since for some reason pullapprove lost the approval...

AndyScherzinger:
So pinging @tobiasKaminsky @mario for second dev review :-)

c0fe:
I have a suggestion @ardevd perhaps adding "Nextcloud" at the bottom would be good?

tobiasKaminsky:
:+1: 
Great enhancement :tada: 
As @ardevd said, dynamically branding it to the first account will be a lot more work.
This can (if at all) done in a new PR. I would also be fine with keeping the default splash screen unbranded, after all it is still the NC app, also the launcher remains NC blue...

(@ardevd for upcoming PRs, please feel free to open up the PRs directly in our repository. It makes handling a bit easier)

PR #2455, opened 2018-04-10
Fix upload list header design

AndyScherzinger:
@mario @tobiasKaminsky please review.

This is a bug fix for the recycler view header design. The deleted file within this PR is the file that was in use for the ListView we had before and the changed file is the one used by the RecylerView now enhanced with the original theming/design parts that where completely missing and also had a negative impact on the material design 16dp spacing which is hereby fixed.

I'd also even vote to include this in the 3.1.0 release since this restores the design we had in 3.0.x and this way we don't create a UI regression bug... But I leave that up to you guys.


mario:
👍 

I also vote for 3.1 cc @tobiasKaminsky

PR #2482, opened 2018-04-18
Use unicode elipsis instead of html one -> desired behaviour on transifex

tobiasKaminsky:
I talked with @rakekniven and the desired behaviour is to use the unicode character for ellipsis instead of html code.
He will soon make an announcement for this.

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -61,7 +61,7 @@
     <string name="recommend_urls">%1$s or %2$s</string>
 
     <string name="auth_check_server">Check server</string>
-    <string name="auth_host_url">Server address https:// …</string>
+    <string name="auth_host_url">Server address https://…</string>
     <string name="auth_username">Username</string>
     <string name="auth_password">Password</string>
     <string name="auth_register">No Nextcloud yet?\nClick here to get one from a provider</string>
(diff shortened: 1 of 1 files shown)


rakekniven:
@tobiasKaminsky We do not use a space between a word and ellipsis for nearly all strings I have seen the last month.

tobiasKaminsky:
"nearly all" or "really all"? ;-)
I am just asking how to decide, when a space should be used and when not.

According to https://practicaltypography.com/ellipses.html it is said
> Should you put word spaces around an el­lip­sis? As with the em dash (see hy­phens and dashes), that’s up to you. Typ­i­cally you’ll want spaces be­fore and af­ter, but if that looks odd, you can take them out.

So I set a whitespace in front everywhere.
But if you think we should not use a whitespace, I am happy to change it.
Just confirm it please :-)
(then we should also put it on wiki page as a common project rule)

AndyScherzinger:
I vote against this since Android lint wants the code not the Unicode character so I think Transifex is just wrong... Just my opinion though...

AndyScherzinger:
:+1: fine by me then!

tobiasKaminsky:
I removed the leading whitespace and I must say it looks a lot nicer in my opinion.

@rakekniven would be cool if you can do the announcement to use unicode character (https://github.com/nextcloud/server/wiki/Translation-knowledge-(valid-for-whole-nextcloud-project)#common-rules).

comradekingu:
@rakekniven My view of space before ellipsis is not NC policy, and therefore errors and inconsistencies run rampant. Norwegian Bokmål works, because i saw too it, I have checked to see that the current situation doesn't work. It doesn't matter what arguments are presented, so I can't hope to win support for my view.

However, if "…" can be used, a lot of errors are avoided, because translators, myself included, don't have every code memorized. To the degree they are memorized, one is also likely to bring up the wrong connotation, remembering one thing as something else, and using it to that effect.
Worst case that changes the content significantly.

rakekniven:
@tobiasKaminsky Announcement is done. See https://www.transifex.com/nextcloud/nextcloud/announcements/8627/

PR #2503, opened 2018-04-23
adapter: Pass valid context to setImageDrawable()-call.

### LINKED ISSUE #2502: Null context passed in ActivitiesListAdapter. Results in NPE crash.
I'm assuming this is due to 5b432b3ed0a608d50388768f0503e47b1504afe8 but the null value passed as the context argument at L257 in the ActivityListAdapter now results in a NullPointerException crash. 

https://github.com/nextcloud/android/blob/3aad742e0c1f9d4f44c96bd2b9ca5863c9efb1f6/src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java#L257

Context.getResources() is being called on the passed null value which leads to the crash.

The issue seems to be fixed when a valid context object is passed to the `setImageDrawable` call.

PR coming shorty.

ardevd:
Pass a valid `Context` to the call to setImageDrawable(). With the recently refactored theming syntax introduced in 5b432b3ed0a608d50388768f0503e47b1504afe8 the null `Context` passed here seemingly results in an NPE crash.

Fixes #2502.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java
+++ b/src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java
@@ -254,7 +254,7 @@ public class ActivityListAdapter extends RecyclerView.Adapter<RecyclerView.ViewH
                         .error(placeholder).into(fileIcon); // using custom fetcher
 
             } else {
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
Nice catch @ardevd 

cc @tobiasKaminsky @mario for second review :)

AndyScherzinger:
:+1:

tobiasKaminsky:
:+1:

PR #2509, opened 2018-04-24
fix more  lint issues

AndyScherzinger:
see subject fixing some more lint warnings:

* lint: fix set android:baselineAligned="false" on this element for better performance
* lint: fix the image ic_favorite.png varies significantly in its density-independent (dip) size across the various density versions (ic_ needs to comply with standard icon sizes)
* lint: fix the image ic_account_circle.png varies significantly in its density-independent (dip) size across the various density versions
* lint: fix This tag and its children can be replaced by one <TextView/> and a compound drawable _and_ unnecessary parent element
* lint: fix attribute networkSecurityConfig is only used in API level 24 and higher (current min is 14)
* lint: fix attribute splitTrack is only used in API level 21 and higher (current min is 14)
* lint: Fix the resource R.string.upload_list_loading appears to be unused
* lint: fix the resource R.layout.list_fragment_expandable appears to be unused
* lint: fix possible overdraw: Root element paints background @color/background_color with a theme that also paints a background (inferred theme is @style/Theme.ownCloud.Toolbar)
* lint: fix implicitly using the default locale is a common source of bugs: Use toLowerCase(Locale) instead. For strings meant to be internal use Locale.ROOT, otherwise Locale.getDefault().


AndyScherzinger:
Build fails atm due to the translation error in Italian, which sould be fixes as soon as the transifex bot runs again _and_ I rebased the PR :)

tobiasKaminsky:
:+1: 
We are getting close to 100 :tada:

mario:
👍

Issue #2510, opened 2018-04-24
Can not connect to server: Unknown error occurred!


ostcar:
### Actual behaviour
The problem is also described [here](https://help.nextcloud.com/t/can-not-connect-to-server-unknown-error-occurred/30741).

When I type in the uri of the NextCloud-Server, I get the response `Unknown error occurred!`.

### Expected behaviour
I expect to get to the next "page" of the app, to insert the username and password.
 

### Environment data
Android version: 8.1.0 

Device model: Nexus 6P

Stock or customized system: Plain Andoid

Nextcloud app version: 3.0.3 and Dev-Version 20180419

Nextcloud server version: 12.0.6.1

#### Web server error log
```
X.X.X.X - - [23/Apr/2018:08:27:56 +0200] “cloud.oshahn.de1” “GET /status.php HTTP/1.1” 200 157 “-” “Mozilla/5.0 (Android) ownCloud-android/3.0.3”
```

respectively

```
X.X.X.X - - [23/Apr/2018:08:28:09 +0200] “cloud.oshahn.de1” “GET /status.php HTTP/1.1” 200 157 “-” “Mozilla/5.0 (Android) ownCloud-android/20180419”
```

I am sure, that there is only this one request.

#### Nextcloud log (data/nextcloud.log)
Nothing is logged when trying to connect the app.

ostcar:
I upgraded the Server to nextcloud 13.0.1.1 and there is still the same behavior.

AndyScherzinger:
Hmm, I get an exception from the http client implementation:

tobiasKaminsky:
What kind of web server do you use?
Apache? Nginx?

AndyScherzinger:
I created #2511 to at least show some more information in case of exceptions / unknown errors due to exceptions.

tobiasKaminsky:
Unfortunately in this case the error message is not that helpful, but better than a generic error.

ostcar:
```
$ nginx -v
nginx version: nginx/1.10.3 (Ubuntu)
```

tobiasKaminsky:
@nextcloud/server-triage I know that there was a similar issue back a while...

Here the idea was to have some basic nginx checking, as it seems that nginx is harder to configure / easier to break than apache: https://github.com/nextcloud/server/issues/8422

MorrisJobke:
> @nextcloud/server-triage I know that there was a similar issue back a while...
> 
[...]

PR #2511, opened 2018-04-24
added more information on initial connection error

AndyScherzinger:
for issues like #2510 provide the exception message to make reporting/analyzing of connection issue during login easier for server-admins and the client devs

*before*
device-2018-04-24-122940

*after*
device-2018-04-24-123111

I wouldn't do any string manipulation on exception message since in this case I guess the chunk size is "" and we don't know what message we might get back from the exceptions in general.

### DIFF
--- a/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -1562,8 +1562,8 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
                 mServerStatusText = getResources().getString(R.string.auth_unknown_error_http_title);
                 break;
             case UNKNOWN_ERROR:
-                if (result.getException() != null && result.getException().getMessage() != null
-                        && !"".equals(result.getException().getMessage())) {
+                if (result.getException() != null &&
+                        !TextUtils.isEmpty(result.getException().getMessage())) {
                     mServerStatusText = getResources().getString(
                             R.string.auth_unknown_error_exception_title,
                             result.getException().getMessage()
--- a/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -1562,7 +1562,7 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
                 mServerStatusText = getResources().getString(R.string.auth_unknown_error_http_title);
                 break;
             case UNKNOWN_ERROR:
-                if(result.getException() != null) {
+                if (result.getException() != null) {
                     mServerStatusText = getResources().getString(
                             R.string.auth_unknown_error_exception_title,
                             result.getException().getMessage()
--- a/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -1562,7 +1562,8 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
                 mServerStatusText = getResources().getString(R.string.auth_unknown_error_http_title);
                 break;
             case UNKNOWN_ERROR:
-                if (result.getException() != null) {
+                if (result.getException() != null && result.getException().getMessage() != null
+                        && !"".equals(result.getException().getMessage())) {
                     mServerStatusText = getResources().getString(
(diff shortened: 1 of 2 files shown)


AndyScherzinger:
I assigned it to 3.2.0 prematurely since I'd "vote" to include this in 3.2.0 since it'll help us and the server admins so it might leads to less connectivity issue reports which are caused by the server rather then the client.

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1559,8 +1559,17 @@ private void updateServerStatusIconAndText(RemoteOperationResult result) {
...
                 break;
             case UNHANDLED_HTTP_CODE:
+                mServerStatusText = getResources().getString(R.string.auth_unknown_error_http_title);
+                break;
             case UNKNOWN_ERROR:
-                mServerStatusText = getResources().getString(R.string.auth_unknown_error_title);
+                if(result.getException() != null) {
+                    mServerStatusText = getResources().getString(
+                            R.string.auth_unknown_error_exception_title,
+                            result.getException().getMessage()
```
tobiasKaminsky: I would check also if getMessage() has text, otherwise it would look a bit strange.

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1559,8 +1559,17 @@ private void updateServerStatusIconAndText(RemoteOperationResult result) {
                 mServerStatusText = getResources().getString(R.string.auth_oauth_error_access_denied);
                 break;
             case UNHANDLED_HTTP_CODE:
+                mServerStatusText = getResources().getString(R.string.auth_unknown_error_http_title);
+                break;
             case UNKNOWN_ERROR:
-                mServerStatusText = getResources().getString(R.string.auth_unknown_error_title);
+                if(result.getException() != null) {
```
tobiasKaminsky: can you add a whitespace after if?

src/main/res/values/strings.xml
```diff
@@ -239,7 +239,9 @@
 	<string name="auth_not_configured_title">Malformed server configuration</string>
 	<string name="auth_account_not_new">An account for the same user and server already exists on the device</string>
 	<string name="auth_account_not_the_same">The entered user does not match the user of this account</string>
-	<string name="auth_unknown_error_title">Unknown error occurred!</string>
+    <string name="auth_unknown_error_http_title">Unknown HTTP error occurred!</string>
+    <string name="auth_unknown_error_title">Unknown error occurred!</string>
+    <string name="auth_unknown_error_exception_title">Unknown error: %1$s</string>
```
tobiasKaminsky: somehow indention is wrong

tobiasKaminsky:
:+1: 
But please do not assign any Milestones, as we discussed, we assign it afterwards. Chances are high that we get this within the next 4 weeks merged ;-) ;-)

If there are some PRs that we want to get into next release, we just should keep track in it in release issue, or (even better) just try to get them in as soon as possible.

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
AndyScherzinger: DONE

AndyScherzinger:
@tobiasKaminsky fixed all your comments :)

tobiasKaminsky:
:+1:

AndyScherzinger:
@mario for review/approval :)

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1559,8 +1559,18 @@ private void updateServerStatusIconAndText(RemoteOperationResult result) {
                 mServerStatusText = getResources().getString(R.string.auth_oauth_error_access_denied);
                 break;
             case UNHANDLED_HTTP_CODE:
+                mServerStatusText = getResources().getString(R.string.auth_unknown_error_http_title);
+                break;
             case UNKNOWN_ERROR:
-                mServerStatusText = getResources().getString(R.string.auth_unknown_error_title);
+                if (result.getException() != null && result.getException().getMessage() != null
+                        && !"".equals(result.getException().getMessage())) {
```
mario: Please replace "result.getException().getMessage() != null && !"".equals(result.getException().getMessage())" with !TextUtils.isEmpty(result.getException().getMessage())

mario:
👍

PR #2512, opened 2018-04-24
Preferences: Don't start FileDisplayActivity when pressing the back button

### LINKED ISSUE #2507: Splash screen shown when returning from preferences
### Actual behaviour
- Splash screen is shown

### Expected behaviour
- Splash screen should only be shown on app start
 
### Steps to reproduce
1. start app (see splash screen)
2. go to settings
3. click "back arrow" -> see splash screen

@ardevd since you implemented the splash screen, can you have look into this?

ardevd:
Since the `FileDisplayActivity` has its launch mode set to '`multiple`' (the default) and `FLAG_ACTIVITY_SINGLE_TOP` is not being used, using `FLAG_ACTIVITY_CLEAR_TOP` will effectively [finish any existing activity and then re-create it](https://developer.android.com/reference/android/content/Intent.html#FLAG_ACTIVITY_CLEAR_TOP). This will cause the launch screen to be showed again. This is probably not the behavior we want.

We could consider declaring the FileDisplayActivity with `android:launchMode="singleTask"` if you want to prevent duplicate entries in the back stack.

This fixes #2507

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/Preferences.java
+++ b/src/main/java/com/owncloud/android/ui/activity/Preferences.java
@@ -839,24 +839,6 @@ public class Preferences extends PreferenceActivity
     }
 
     @Override
-    public boolean onMenuItemSelected(int featureId, MenuItem item) {
-        super.onMenuItemSelected(featureId, item);
-        Intent intent;
-
-        switch (item.getItemId()) {
-            case android.R.id.home:
-                intent = new Intent(getBaseContext(), FileDisplayActivity.class);
-                intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
-                startActivity(intent);
-                break;
-            default:
-                Log_OC.w(TAG, "Unknown menu item triggered");
-                return false;
-        }
-        return true;
-    }
-
-    @Override
     protected void onActivityResult(int requestCode, int resultCode, Intent data) {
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
I found one problem:

- start app
- go to settings
- back to main view, no splash screen (:+1:)
- click back
- see still main view -> second time back needed
-> it seems that a new FDA instance is created and thus 2xFDA is on stack.

Adding "singleTask" to FDA gives following mishbehaviour:
- open app
- open an image (preview is shown)
- minimize app
- open app again
-> should show preview, but shows file list (FDA)

For me it works, if we remove the whole block:
https://github.com/nextcloud/android/blob/20e55b89a5e58a6eab36890bc37710f2d454f522/src/main/java/com/owncloud/android/ui/activity/Preferences.java#L841-L855

ardevd:
@tobiasKaminsky good find!

By removing the entire block we simply finish the Preference activity, popping it off the stack. This will return us back to the previous activity unlike the current behavior where we are always returned to the FileDisplayActivity. This makes more sense in my opinion anyway so lets do that.

tobiasKaminsky:
:+1:

tobiasKaminsky:
Rebased and squashed so that the first (outdated) commit is gone.

AndyScherzinger:
:+1:

PR #2312, opened 2018-03-07
Add support for integration tests

tobiasKaminsky:
This adds support for integration tests.

First there are only two basic test "suites", create folder and upload file.
Both have two test cases, which try to create folder / upload file in one existing folder and in an non-existing folder.
(with the later one [uploading a file into a non-existing folder] we could have avoid 3.0.3).

It spins up a server with latest master, which as a user test, where we can do testing.

CodeCov takes reports from unit and integration tests and shows a summary.

After all this is only a very basic/tiny step to introduce testing:
- write unit tests
- write integration tests
- write UI tests

Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

PR #2514, opened 2018-04-25
bump testing/espresso lib versions 1.0.1->1.0.2 ; 3.0.1->3.0.2

AndyScherzinger:
see subject, bump test lib versions (bugfix updates)

cc @mario @tobiasKaminsky


tobiasKaminsky:
This should be directly in #2312, so that we see that it is working at least for those tests that are already existing.

AndyScherzinger:
Okay :)

PR #2520, opened 2018-04-26
Do not sort on getFolderContent

tobiasKaminsky:
~~Currently this is the only left occurrence I can find for our TimSort problem.~~
I went through all calling methods and either sorting order is not important, or it is done afterwards anyway (like in OCFileListFragment).

At the only occurrence where it is import, I added our AlphaNumComparator.


AndyScherzinger:
:+1:

mario:
👍 

@AndyScherzinger merge.

Issue #379, opened 2016-11-11
[Feature request] Enable video streaming


xylo:
### Actual behaviour
When I click on a video that has not been downloaded so far the app starts downloading the video file instead of playing it directly.

### Expected behaviour
When I click on a video that has not been downloaded so far the video should played as video stream, i.e. playing should start immediately and the video should not be downloaded to the internal memory.

There is already a pull request for this feature: https://github.com/nextcloud/android/pull/183

Moreover the feature is already included in the Beta app and works for me (at least when playing the video in an external player).  For some reason the internal player has some issues with several or all video formats.

### Steps to reproduce
1. Navigate to a folder with videos that have not been downloaded.
2. Click on a video.

tobiasKaminsky:
Thanks for creating the issue and remaining me of the stale PR.
I schedule it for the release after the next.

tobiasKaminsky:
Regarding your second issue: as this seems to be independent of streaming please open a new issue and maybe you can provide some sample videos and specify on which device the problem occurs.

xylo:
Thanks a lot, Tobias!  Regarding the second issue, I will file another ticket next week.

lftsy:
Good evening, I am testing the app 3.0.0RC2 from PlayStore and streaming feature seems not present yet so I keep my old owncloud Android app to be able to stream movies
If you need some testers for the Android video streaming, I can test and report problem If any
=> I think you can update roadmap to next version since 3.0.0 seems not having it
Thanks and Best Regards

supportreq:
@tobiasKaminsky is this feature available??

PR #2524, opened 2018-04-27
Media streaming with NC14 endpoint

tobiasKaminsky:
Needs: 
- [x] https://github.com/nextcloud/android-library/pull/142
- [x] https://github.com/nextcloud/server/pull/9178

Fixes: #379 

This enables video/audio streaming without exposing any credentials.

Test cases:
NC14
- non downloaded file
    - stream video internally
		- toggle to fullscreen
    - stream audio internally
    - stream video with external app
    - stream audio with external app
- downloaded file
    - play video/audio directly

NC13
- no option to stream
- non downloaded file
    - download media, then play
- downloaded file
    - play video/audio directly

### DIFF
--- a/src/main/java/com/owncloud/android/files/StreamMediaFileOperation.java
+++ b/src/main/java/com/owncloud/android/files/StreamMediaFileOperation.java
@@ -1,0 +1,90 @@
+/*
+ * Nextcloud Android client application
+ *
+ * @author Tobias Kaminsky
+ * Copyright (C) 2018 Tobias Kaminsky
+ * Copyright (C) 2018 Nextcloud GmbH.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.owncloud.android.files;
+
+import com.owncloud.android.lib.common.OwnCloudClient;
+import com.owncloud.android.lib.common.operations.RemoteOperation;
+import com.owncloud.android.lib.common.operations.RemoteOperationResult;
+import com.owncloud.android.lib.common.utils.Log_OC;
+
+import org.apache.commons.httpclient.HttpStatus;
+import org.apache.commons.httpclient.methods.PostMethod;
+import org.json.JSONObject;
+
+import java.util.ArrayList;
+
+public class StreamMediaFileOperation extends RemoteOperation {
+    private static final String TAG = StreamMediaFileOperation.class.getSimpleName();
+    private static final int SYNC_READ_TIMEOUT = 40000;
+    private static final int SYNC_CONNECTION_TIMEOUT = 5000;
+    private static final String STREAM_MEDIA_URL = "/ocs/v2.php/apps/dav/api/v1/direct";
+    
+    private String fileID;
+
+    // JSON node names
+    private static final String NODE_OCS = "ocs";
+    private static final String NODE_DATA = "data";
+    private static final String NODE_URL = "url";
+    private static final String JSON_FORMAT = "?format=json";
+
+    public StreamMediaFileOperation(String fileID) {
+        this.fileID = fileID;
+    }
+
+    protected RemoteOperationResult run(OwnCloudClient client) {
+        RemoteOperationResult result;
+        PostMethod postMethod = null;
+
+        try {
+            postMethod = new PostMethod(client.getBaseUri() + STREAM_MEDIA_URL + JSON_FORMAT);
+            postMethod.setParameter("fileId", fileID);
+
+            // remote request
+            postMethod.addRequestHeader(OCS_API_HEADER, OCS_API_HEADER_VALUE);
+
+            int status = client.executeMethod(postMethod, SYNC_READ_TIMEOUT, SYNC_CONNECTION_TIMEOUT);
+
+            if (status == HttpStatus.SC_OK) {
+                String response = postMethod.getResponseBodyAsString();
+
+                // Parse the response
+                JSONObject respJSON = new JSONObject(response);
+                String url = (String) respJSON.getJSONObject(NODE_OCS).getJSONObject(NODE_DATA).get(NODE_URL);
+
+                result = new RemoteOperationResult(true, postMethod);
+                ArrayList<Object> urlArray = new ArrayList<>();
+                urlArray.add(url);
+                result.setData(urlArray);
+            } else {
+                result = new RemoteOperationResult(false, postMethod);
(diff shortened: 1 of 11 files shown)


tobiasKaminsky:
For such PR that need a library PR, I guess the best is to reference the branch directly, or?
Problem will be, when we merge the corresponding PR, then there is no PR and the build fails. Then we would have to use master until we release a new library version, otherwise any other client PR will fail.

--> I am not sure how to handle this best…

src/main/java/com/owncloud/android/files/StreamMediaFileOperation.java
```diff
@@ -0,0 +1,90 @@
...
+
+            int status = client.executeMethod(postMethod, SYNC_READ_TIMEOUT, SYNC_CONNECTION_TIMEOUT);
+
+            if (status == HttpStatus.SC_OK) {
+                String response = postMethod.getResponseBodyAsString();
+
+                // Parse the response
+                JSONObject respJSON = new JSONObject(response);
+                String url = (String) respJSON.getJSONObject(NODE_OCS).getJSONObject(NODE_DATA).get(NODE_URL);
+
```
AndyScherzinger: couldn't you just use getString instead of get()+cast ?

tobiasKaminsky: Indeed :+1:

mario:
Shouldn't stream just be the default instead of having a dedicated stream menu button? What does "stream" means to say, my grandma?

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -342,6 +343,34 @@ public void run() {
...
+
+                Intent openFileWithIntent = new Intent(Intent.ACTION_VIEW);
+                Uri uri = Uri.parse((String) result.getData().get(0));
+
+                openFileWithIntent.setDataAndType(uri, file.getMimetype());
+
+                mFileActivity.startActivity(Intent.createChooser(openFileWithIntent,
+                        mFileActivity.getString(R.string.stream)));
+            }
+        }).start();
```
AndyScherzinger: optional: What about using a lambda and putting the whole run() code into a separate method?

src/main/java/com/owncloud/android/ui/preview/PreviewMediaFragment.java
```diff
@@ -495,16 +499,68 @@ private void prepareVideo() {
...
+            mVideoPreview.setVideoURI(getFile().getStorageUri());
+        } else {
+            try {
+                OwnCloudAccount ocAccount = new OwnCloudAccount(mAccount, getContext());
+                OwnCloudClient client = OwnCloudClientManagerFactory.getDefaultSingleton().
+                        getClientFor(ocAccount, getContext());
+
+                new LoadStreamUrl(this, client).execute(getFile().getLocalId());
+            } catch (Exception e) {
+                Log_OC.e(TAG, "Loading stream url not possible: " + e.getMessage());
```
AndyScherzinger: shouldn't we also add the exception itself to the log statement?

src/main/java/com/owncloud/android/ui/preview/PreviewMediaFragment.java
```diff
@@ -495,16 +499,68 @@ private void prepareVideo() {
...
+                    previewMediaFragment.mVideoUri = uri;
+                    previewMediaFragment.mVideoPreview.setVideoURI(uri);
+                } else {
+                    previewMediaFragment.mMultiView.setVisibility(View.VISIBLE);
+                    previewMediaFragment.setMessageForMultiList(
+                            previewMediaFragment.getString(R.string.stream_not_possible_headline),
+                            R.string.stream_not_possible_message, R.drawable.file_movie);
+                }
+            } else {
+                Log_OC.e(TAG, "Error streaming file: no previewMediaFragment!");
```
AndyScherzinger: Same here: shouldn't we also add the exception itself to the log statement?

AndyScherzinger:
Just some minor code comments

> Shouldn't stream just be the default instead of having a dedicated stream menu button?

I agree it should be the default cation when clicking on an item (that can be streamed in general) and I remember that this is also what Tobias told me :)

tobiasKaminsky:
It is: if you click on audio/video it will try to stream the file with our internal player.
You can also open the overflow menu and select "stream with…" to directly open up the file in e.g. vlc.

If our internal player fails, it says "Please download media instead or use external app.".

src/main/java/com/owncloud/android/media/MediaService.java
```diff
@@ -720,4 +726,48 @@ protected MediaControlView getMediaController() {
         return mMediaController;
     }
 
+    private static class LoadStreamUrl extends AsyncTask<String, Void, String> {
```
mario: Is there no other way but an AsyncTask? I thought we got rid of this ages ago :P

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
mario: Yup please.

mario:
To further clarify my comment on AsyncTasks:

<pygi> TobiasK: as for media streaming: I'd appreciate if we'd move away from asynctasks, especially for new code
<pygi> even a service-based download + eventbus/interface download notification would work better
<pygi> but I guess you could go for a Job + eventbus notification too, which would make it cleeeeeaner :)

tobiasKaminsky:
For up/downloading I agree to use services or jobs/eventbus.
But for this we need to have this executed as soon as possible as it fetches the video streaming url. 

So I would go for this:
- if it "urgent"/ as soon as possible, use asyncTask (afaik it is still the recommended way of android to show results of small background options on the UI).
- all other, especially things that are really asynchronous, like upload/download, keep-up-to-date should be done via jobs, so that android doze and other battery optimizations can take care of it

rullzer:
Would be great to get this in. NC14 will go into freeze soon. So having a client that uses it will allow us to properly test it.

src/main/java/com/owncloud/android/ui/preview/PreviewMediaFragment.java
tobiasKaminsky: Done.

src/main/java/com/owncloud/android/ui/preview/PreviewMediaFragment.java
tobiasKaminsky: Here is no exception.

tobiasKaminsky:
Rebased & ready for final review

AndyScherzinger:
:+1: fine by me :)

AndyScherzinger:
@tobiasKaminsky what about @mario's latest comment: https://github.com/nextcloud/android/pull/2524#issuecomment-390923068 ?

tobiasKaminsky:
Both loading url task are done when playing media, so I think it is fine.
Android Dev guide also recommends async task for short running calls.

Beside that, there is somewhere an issue, where I started a discussion how to do long  running background tasks, that maybe even survive activity rotate/open/close.

--> I would keep it this way and later replace it with a potential better solution.

AndyScherzinger:
> --> I would keep it this way and later replace it with a potential better solution.

Fine with me

tobiasKaminsky:
E.g. UserInfoActivity is the same:
open an account -> fetches data
rotate screen -> fetches data again

Of course, doing the same is not pushing the code to a better state, but I think we should find a better and generic approach first and then change it.

Issue #2485, opened 2018-04-18
Roadmap: file detail & new sharing


tobiasKaminsky:
During hackweek @jancborchardt draw a new sharing system, which is nearly the same as web UI and which should then be also in sync with iOS.

* [x] Iteration 1 for File detail & new sharing #2533 (revamped UI approach)
* [x] Iteration 2 for File detail & new sharing #2573 (sharing in general)
* [x] Iteration 3 for File detail & new sharing #2671 (e-mail sharing permissions, details)
* [x] Iteration 4 for File detail & new sharing #2700 (create e-mail shares)
* [x] Iteration 5 for File detail & new sharing #2690 (comments)
* [x] Display copy icon and adding send action to the link menu #2730
* [x] Share link creation via bottom sheet for files in OCFileList #2747
* [x] implement #2742
* [ ] Add label/tag support - #11515
* [ ] Remove legacy sharing screen completely - TBD

Enhanced file detail view:
![detail](https://user-images.githubusercontent.com/5836855/38919147-102edd3a-42f0-11e8-8fcb-3b4403b27f6b.jpg)
This is an enhanced view for images/videos which have a preview:
- no overflow menu on top, but next to filename
- clicking on star toggles favorite
- tags opens up editing
- activity and sharing as "tabs" which can be toggled

Activity tab:
- combined activity, comments, versions (endpoint on server side is needed for this, https://github.com/nextcloud/server/issues/658)
- on top: text input box for commenting

Share tab:
![sharing](https://user-images.githubusercontent.com/5836855/38919558-32a34026-42f1-11e8-8c11-0b33347f3528.jpg)
- sharees are sorted by permissions (from max permissions to read only)
- on each sharee there is an overflow menu with its specific sharing permissions

Sharing to a new user/group/link:
- click on text input on share tab hides activity/sharing toggle to have more space for actually searching
- share link as entry beneath search field
	- with "can edit" 
	- click on it opens up existing app chooser to receive link
- during input we show the list of results
	- username
	- group
	- federated sharing
	- share by mail

![2018-04-18-102204](https://user-images.githubusercontent.com/5836855/38919944-5ed57262-42f2-11e8-8d79-1e14863440d3.png)
- share (left) -> opens up share dialog of file detail
- link (right) -> directly creates read-only link (if needed) and open list with apps to send to. "copy link" should be first entry as otherwise users might not find it (@schiessle)

cc: @rullzer @marinofaggiana for other clients

[...]

PR #2533, opened 2018-04-30
Iteration 1 for File detail & new sharing 

AndyScherzinger:
WiP for #2485 and collab with @tobiasKaminsky 

* [x] refactored the details screen to have 2 tabs for activities and sharing including swipe left/right for activities/sharing
* [x] add activities loading on a per file basis (possibly needs a server API change)
* [x] clarify "star use"
* [x] implement star / keep available offline (after discussion)
* [x] move overflow menu from top to Filename "row"
* [x] remove sharing icon from top menu
* [x] make header transparent, only back arrow visible (for preview-able files), non-preview-able files do not have any icon/preview, so that the first "row" is file name (and then we can show again a solid header).
* [x] add previews
* [x] add preview-able/non-preview-able toolbar handling
* [x] fix pull-to-refresh on activities tab
* [x] fix theming 

Remarks:
* add commenting and sharing option via separate PRs _after_ this one is merged

What do you think @tobiasKaminsky @jancborchardt ?

Actual WiP UI-wise
device-2018-04-30-184640
device-2018-04-30-184658


AndyScherzinger:
Added initial loading of activities (copied the code from the activities activity, so not file specific yet):
device-2018-04-30-191745

tobiasKaminsky:
Filter by file is working

tobiasKaminsky:
> star use

I think this should toggle favorite state.

AndyScherzinger:
> I think this should toggle favorite state.

Okay, so would we then 
* drop the "available offline" part in this screen? (which leads to a download of the file) or 
* have that part purely via the menu item?

nickvergessen:
@AndyScherzinger 
> any hints on how to just load the activities of a given file? https://github.com/nextcloud/activity/blob/master/docs/endpoint-v2.md did give me any hints and the web UI is capable of doing so while I of course don't know if the web UI uses an API which allows for that and is also exposed for other clients.

using `/filter` and then providing the object types (`files`) and the object id (file id)

AndyScherzinger:
@nickvergessen 

> using /filter and then providing the object types (files) and the object id (file id)

While I wouldn't want to get only files but all types of activities but just for a single file (file id). From what I understood from Tobias this can be done while I read the API docs as-in you have to provide type and id while it seems you can only provide the id. (don't want a activities type filter being used here :)

tobiasKaminsky:
>Okay, so would we then
>- drop the "available offline" part in this screen? (which leads to a download of the file) or
>- have that part purely via the menu item?

The behaviour should stay the same, but I do not think that we need an overflow menu item
-> only star icon, but with current functionality

tobiasKaminsky:
To not have this too big, I would say
- implement star / keep available offline (after discussion)
- move overflow menu from top to Filename "row"
- remove sharing icon from top menu
- make header transparent, only back arrow visible
- comment tags (as we do not have them yet; I would do this in a following PR)

This way we have a relative small PR, but with nice UI changes.

AndyScherzinger:
> make header transparent, only back arrow visible

this makes sense for preview-able files while it doesn't for any other kind and another issue here is, how should this be treated for the table use case where the details would just be the right pane?

AndyScherzinger:
@tobiasKaminsky overflow menu has been put to the filename, tabs are now left aligned:
device-2018-05-03-151504
device-2018-05-03-151516

tobiasKaminsky:
>> make header transparent, only back arrow visible

>this makes sense for preview-able files while it doesn't for any other kind and another issue here is, how should this be treated for the table use case where the details would just be the right pane?

Jan's idea is that non previewable files do not have any icon, so that the first "row" is file name (and then we can show again a solid header).

AndyScherzinger:
@tobiasKaminsky I added all remaining/done tasks to the issue description for iteration 1 :)

AndyScherzinger:
@tobiasKaminsky @jancborchardt 
Status quo of the PR:
device-2018-05-04-154202
device-2018-05-04-154144

like mentioned in the original issue we can also remove the icon display here, or do that in a later PR. Both options are fine with me :)

AndyScherzinger:
@mario @tobiasKaminsky @jancborchardt @nextcloud/designers iteration one of the rewrite of Details&Sharing is ready for review. Beware that since this is just the first iteration (any further changes to be done in the upcoming PR) there is some code duplication (sharing implementation via file list and sharing info in the details screen tab), no further use of butterknife (to be added later) and desgin-wise e.g. the file type icon is still shown for non-preview-able files.

@tobiasKaminsky :exclamation: beware that due to all tasks for iteration 1 have been completed - I did a rebase

tobiasKaminsky:
:+1: 
I am fine with this, great stuff. :tada:

AndyScherzinger:
@mario @jancborchardt can you please comment if this is fine for a first step? :)

AndyScherzinger:
Cool, then I'll revert it back to centered. I only did it since on the web UI it is also left aligned. Will fix within this PR :)
Thanks for the feedback @jancborchardt ❤️

jancborchardt:
@AndyScherzinger yeah, that is cause in the web UI we still didn't reduce to 2 tabs :)

mario:
👍

AndyScherzinger:
I'll soon start with Iteration 2 for follow up on this PR.

Things to be done
* adding further functionality described in #2485
* technical cleanup
  * listview->recyclerview
  * butterknife
  * MVP-ing
  * avatar fetching

Happy for any collaboration @tobiasKaminsky @mario @ardevd

PR #2541, opened 2018-05-02
authentication: Added @Override annotations to applicable methods.

ardevd:
Using the @Override annotation is useful for two reasons :

It elicits a warning from the compiler if the annotated method doesn't actually override anything, as in the case of a misspelling.
It improves the readability of the source code by making it obvious that methods are overridden.


AndyScherzinger:
:+1: looks good to me, while I have no idea what the parameter name "icicle" is... but that isn't part of this PR I guess ;)

mario:
👍

AndyScherzinger:
Thanks @ardevd :heart: you are on fire :1st_place_medal:

MeiRos:
What about icicle on the lines 52 and 53? Should those be changed to savedInstanceState too?

AndyScherzinger:
Good point @MeiRos, yes they should be changed too :+1:

PR #2544, opened 2018-05-03
Brander with white primary color

tobiasKaminsky:
If in setup.xml white is chosen as primary color, this will now show login screen.

To test use in setup.xml:
< color name="primary">#ffffff</color>
< color name="login_text_color">@color/black</color>

TODO
- [x] do the same for old login method
- [x] revert button on snackbar not correct tinted

### DIFF
--- a/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -433,6 +433,7 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
             @Override
             public void run() {
                 Snackbar.make(mLoginWebView, R.string.fallback_weblogin_text, Snackbar.LENGTH_INDEFINITE)
+                        .setActionTextColor(getResources().getColor(R.color.primary_dark))
                         .setAction(R.string.fallback_weblogin_back, new View.OnClickListener() {
                             @Override
                             public void onClick(View v) {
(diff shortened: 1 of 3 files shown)


tobiasKaminsky:
Now also for legacy login
(we should start thinking to get rid of this old login)

AndyScherzinger:
> we should start thinking to get rid of this old login

why? It is used as a fallback at the moment plus it might then completely break compatibility with oC thus break clients for some of the app's current users. 
So if the decision is to do this we _must_ have a release first which informs each user using a server pre-12 or oC that the next release will potentially break their client so they need to deactivate the auto-update in GPlay / can't hit update in f-droid b/c else their client is broken.

Not doing so would be really bad (UX) in my opinion plus it'll result in another mass of 1-star ratings wich due to the initial login issues since the webLogin change already pushed down the ratings and this hurts since I've put a lot of time into the app and now have to watch the ratings going down... So please let's not make this even worse...

AndyScherzinger:
@tobiasKaminsky according to issue description "do the same for old login method" is still to be done ;)

AndyScherzinger:
:+1: fine by me then

mario:
Where is the logo though? xD

tobiasKaminsky:
Well, it is a brander, you can use whatever logo you want. 
As I wanted to show on the screenshots only the problematic font colors, I did not changed the logo.

mario:
👍 
Feel free to merge.

AndyScherzinger:
> Where is the logo though? xD

The logo is white ;) Well, if somebody changes background to white and uses a white logo, so be it ;)
We can't tint the logo since it would ruin any logo that is multi-colored or low-contrast on purpose. So fine by me

PR #2545, opened 2018-05-04
ortiz.touch: End switch case with an unconditional break.

ardevd:
No idea who maintains the TouchImageView class, but it seems like a bug to me that the switch case is missing a break.


tobiasKaminsky:
:+1: 
Indeed!
Current behaviour means that CENTER_INSIDE is the same like FIT_CENTER…

A quick search shows me, that we never change the default value (FIT_CENTER), so this will not affect anything, but still good to have this in a sane state.

AndyScherzinger:
:+1:

AndyScherzinger:
aaaaaand merged :+1: 

Thanks @ardevd :)

PR #2546, opened 2018-05-04
ui: Corrected '&' to '&&' in boolean if statement.

ardevd:
The use of non-short-circuit logic in a boolean context is likely a mistake - one that could cause serious program errors as conditions are evaluated under the wrong circumstances.


tobiasKaminsky:
:+1: 
Thank you very much for your contribution :tada:

AndyScherzinger:
:+1: looks good, thanks @ardevd :heart:

PR #2547, opened 2018-05-04
Anonymous inner classes containing only one method should become lambdas

ardevd:
Before Java 8, the only way to partially support closures in Java was by using anonymous inner classes. But the syntax of anonymous classes may seem unwieldy and unclear.

With Java 8, most uses of anonymous inner classes should be replaced by lambdas to highly increase the readability of the source code.

### DIFF
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -511,20 +511,14 @@ public class MainApp extends MultiDexApplication {
                     new AlertDialog.Builder(context, R.style.Theme_ownCloud_Dialog)
                             .setTitle(R.string.drawer_synced_folders)
                             .setMessage(R.string.synced_folders_new_info)
-                            .setPositiveButton(R.string.drawer_open, new DialogInterface.OnClickListener() {
-                                public void onClick(DialogInterface dialog, int which) {
-                                    // show Auto Upload
-                                    Intent folderSyncIntent = new Intent(context,
-                                            SyncedFoldersActivity.class);
-                                    dialog.dismiss();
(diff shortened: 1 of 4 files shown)


src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -511,20 +511,14 @@ private static void updateToAutoUpload() {
...
-                                }
-                            })
-                            .setNegativeButton(R.string.drawer_close, new DialogInterface.OnClickListener() {
-                                public void onClick(DialogInterface dialog, int which) {
-                                    dialog.dismiss();
-                                }
+                            .setPositiveButton(R.string.drawer_open, (dialog, which) -> {
+                                // show Auto Upload
+                                Intent folderSyncIntent = new Intent(context,
+                                        SyncedFoldersActivity.class);
```
tobiasKaminsky: We use line break at 120 chars, so this can be in one line, which then also enhances readability.
(though we have no common rules or strict code formatting, my rule of thumb is to have as much as possible in one line, and if it does not fit, then have a line break on the latest ",").

tobiasKaminsky:
:+1:

src/main/java/com/owncloud/android/MainApp.java
AndyScherzinger: Fixed via https://github.com/nextcloud/android/pull/2547/commits/cd9bfb12b8b4a9c8bf5016e8ae06944ed3db203d

AndyScherzinger:
:exclamation: CR fixed + rebased, just waiting for the checks to turn green for merge

AndyScherzinger:
:+1: approved and merged :) Thanks @ardevd :rocket:

PR #2548, opened 2018-05-04
datamodel: String function use should be optimized for single characters

ardevd:
An lastIndexOf call with a single letter String can be made more performant by switching to a call with a char argument.


tobiasKaminsky:
:+1:

ardevd:
@tobiasKaminsky I have a self hosted SonarQube instance that helps me out ;)

tobiasKaminsky:
Seems that SonarQube is finding a bit more than Codacy…

ardevd:
SonarQube is a really awesome tool! Never used Codacy though so not sure how it compares.

AndyScherzinger:
:+1: 

codacy is basically "just running" pmd and findbugs

Issue #2539, opened 2018-05-02
Upload button / design issue with white color in theming app


BornToBeRoot:
You cannot read the text in the button...

![image](https://user-images.githubusercontent.com/16019165/39525959-c124593e-4e1d-11e8-8bbc-498685e351d9.png)

App Version: 3.1.0 (PlayStore)

NC13.0.2 with theming app and color white (`#FFFFFF`) 

![image](https://user-images.githubusercontent.com/16019165/39525989-d3b35208-4e1d-11e8-9801-3ac67487f1bf.png)

AndyScherzinger:
Same goes for the icon in the middle of the screen (folder icon for empty content) which is tinted white while here it should be grey.

tobiasKaminsky:
We just should forbid this ;-)
Similar to #2544, this needs indeed to be improved.

PR #2550, opened 2018-05-04
White theming

tobiasKaminsky:
Fix https://github.com/nextcloud/android/issues/2539

 

There is still a glitch with the overflow menu.

### DIFF
--- a/src/main/java/com/owncloud/android/files/services/FileDownloader.java
+++ b/src/main/java/com/owncloud/android/files/services/FileDownloader.java
@@ -132,7 +132,7 @@ public class FileDownloader extends Service
                 .setContentText(getApplicationContext().getResources().getString(R.string.foreground_service_download))
                 .setSmallIcon(R.drawable.notification_icon)
                 .setLargeIcon(BitmapFactory.decodeResource(getResources(), R.drawable.notification_icon))
-                .setColor(ThemeUtils.primaryColor(getApplicationContext()));
+                .setColor(ThemeUtils.primaryColor(getApplicationContext(), true));
 
         if ((android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.O)) {
             builder.setChannelId(NotificationUtils.NOTIFICATION_CHANNEL_DOWNLOAD);
--- a/src/main/java/com/owncloud/android/files/services/FileUploader.java
+++ b/src/main/java/com/owncloud/android/files/services/FileUploader.java
@@ -468,7 +468,7 @@ public class FileUploader extends Service
                 .setContentText(getApplicationContext().getResources().getString(R.string.foreground_service_upload))
                 .setSmallIcon(R.drawable.notification_icon)
                 .setLargeIcon(BitmapFactory.decodeResource(getResources(), R.drawable.notification_icon))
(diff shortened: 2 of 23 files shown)


tobiasKaminsky:
Or do we want to try to have it still complete white, but change text to black/grey?

MorrisJobke:
cc @nextcloud/theming

MorrisJobke:
> Or do we want to try to have it still complete white, but change text to black/grey?

cc @nextcloud/designers

MariusBluem:
> Or do we want to try to have it still complete white, but change text to black/grey?

I'd vote for that 👍

tobiasKaminsky:
Damn Marius… :rofl: I feared that someone wanted this ;-)

tobiasKaminsky:
On the right, it confuse me a bit that the primary button is less visible than the secondary one…

MariusBluem:
Make the primary button black :)

google-drive-app-device-backups-840x744

AndyScherzinger:
> On the right, it confuse me a bit that the primary button is less visible than the secondary one…

Primary button should be black, whith the white action bar all menu icons + text + back arrow should be black (or at least instead of black it should be a dark grey tone like the folder icons)

tobiasKaminsky:
The primary button is now grey, as otherwise we could not use font color (which is black), but we would have to create an additional logic for font on primary colors, when white.

Please just play a bit with the app, I hope I found every white / non readable font/background.

Some samples:

jancborchardt:
Yeah, agree with Marius that proper white is the way to go, just like it looks on the server. :) And the newest screenshots look good! :)

AndyScherzinger:
@tobiasKaminsky  manual upload has a white title ;)
device-2018-05-07-135320

tobiasKaminsky:
Nice finding @AndyScherzinger :+1:

AndyScherzinger:
@tobiasKaminsky found two other minor issues: 
* The spinner in the action bar of your screenshot is white with a white action bar background (the triangle), see your last screenshots
* the "report an issue on Github" on the participate page isn't colored like the primary action button (e.g. like now for the manual upload)

tobiasKaminsky:
I could not fix the spinner, all I found on the internet only changed the background, but not the triangle. Maybe you have an idea?

Issue 2 is fixed

AndyScherzinger:
> Maybe you have an idea?

My guess is that this is then again related to Androids theming. In contrast to the action bar theme the spinner has a `dark` theming since also the opened spinner has a dark background thus the triangle is white. only thing I can think of would be to apply a different, fallback Android theme to the spinner in case of a light action bar background.

mario:
What's the status here?

tobiasKaminsky:
Triangle of spinner is still white.
I tried to solve it programatically, but this did not worked out, yet.

@AndyScherzinger idea to use a fallback theme would be an option.
@mario/@ardevd maybe you have another idea how to solve this programatically?

tobiasKaminsky:
I finally found a way to tint the spinner :tada: 

 

So this is finally complete and ready to review again.

mario:
👍 
Rebase.

AndyScherzinger:
:+1:

PR #2564, opened 2018-05-09
MainApp: Fixed auto upload state check.

ardevd:
This is almost certainly a mistake. Using the same value on either side of a binary operator makes no sense. It should check whether photo uploads or video uploads are enabled.

### DIFF
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -488,7 +488,8 @@ public class MainApp extends MultiDexApplication {
 
     private static void updateToAutoUpload() {
             Context context = getAppContext();
-            if (PreferenceManager.instantPictureUploadEnabled(context)) {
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
I was sure that we fixed this at some point...
Correct is: 
`
if (PreferenceManager.instantPictureUploadEnabled(context) ||
                            PreferenceManager.instantVideoUploadEnabled(context)) {
`
as we previously had instant upload for picture and video.
And if one of both is set, it should be updated to auto upload (our current system).

ardevd:
@tobiasKaminsky I was also certain we fixed this previously. However, you are probably referring [to this](https://github.com/nextcloud/android/pull/2427). Same issue, different file. Will update this PR with correct fix.

AndyScherzinger:
:+1: looks good to me :1st_place_medal:

mario:
👍

tobiasKaminsky:
@ardevd thanks for pointing me to the other PR :+1:

PR #2565, opened 2018-05-09
authentication: Fixed HTTPS prefixed URL normalization.

ardevd:
I'm assuming here that the if statement is supposed to be checking whether the normalized URL starts with HTTP or HTTPS. The current version seemingly has a typo which this commit fixes.


AndyScherzinger:
:+1: Nice catch @ardevd 

... merged :medal_sports:

PR #2570, opened 2018-05-11
Verify context before getting file type icon.

### LINKED ISSUE #2569: App crashes when opening Auto Upload screen.
### Actual behaviour
- When opening the Auto Upload screen, the app crashes.

### Expected behaviour
- No crash should occur. 

### Steps to reproduce
1. Start app 
2. Pick "Auto Upload" from the nav drawer.
3. App crashes

### Environment data
Android version: 8.1.0

Device model: Emulator

Nextcloud app version: Master.

### Logs
```
05-11 09:39:41.905  8980  8980 D AndroidRuntime: Shutting down VM
05-11 09:39:41.906  8980  8980 E AndroidRuntime: FATAL EXCEPTION: main
05-11 09:39:41.906  8980  8980 E AndroidRuntime: Process: com.nextcloud.client, PID: 8980
05-11 09:39:41.906  8980  8980 E AndroidRuntime: java.lang.NullPointerException: Attempt to invoke virtual method 'android.content.res.Resources android.content.Context.getResources()' on a null object reference
05-11 09:39:41.906  8980  8980 E AndroidRuntime: 	at com.owncloud.android.utils.MimeTypeUtil.getFileTypeIcon(MimeTypeUtil.java:98)
05-11 09:39:41.906  8980  8980 E AndroidRuntime: 	at com.owncloud.android.utils.MimeTypeUtil.getFileTypeIcon(MimeTypeUtil.java:85)
05-11 09:39:41.906  8980  8980 E AndroidRuntime: 	at com.owncloud.android.datamodel.ThumbnailsCacheManager$MediaThumbnailGenerationTask.onPostExecute(ThumbnailsCacheManager.java:704)
05-11 09:39:41.906  8980  8980 E AndroidRuntime: 	at com.owncloud.android.datamodel.ThumbnailsCacheManager$MediaThumbnailGenerationTask.onPostExecute(ThumbnailsCacheManager.java:642)
05-11 09:39:41.906  8980  8980 E AndroidRuntime: 	at android.os.AsyncTask.finish(AsyncTask.java:695)
05-11 09:39:41.906  8980  8980 E AndroidRuntime: 	at android.os.AsyncTask.-wrap1(Unknown Source:0)
05-11 09:39:41.906  8980  8980 E AndroidRuntime: 	at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:712)
05-11 09:39:41.906  8980  8980 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:106)
05-11 09:39:41.906  8980  8980 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:164)
05-11 09:39:41.906  8980  8980 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:6494)
05-11 09:39:41.906  8980  8980 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
05-11 09:39:41.906  8980  8980 E AndroidRuntime: 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
05-11 09:39:41.906  8980  8980 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
05-11 09:39:41.920  1774  2359 W ActivityManager:   Force finishing activity com.nextcloud.client/com.owncloud.android.ui.activity.SyncedFoldersActivity
05-11 09:39:41.922  1774  2359 W ActivityManager:   Force finishing activity com.nextcloud.client/com.owncloud.android.ui.activity.FileDisplayActivity

```

ardevd:
The context object passed to `getFileTypeIcon` can be null and we should therefore verify that it's not before calling `context.getResources()`. 

Furthermore, `context.getResources()` was deprecated in Marshmellow and we should be using the new theme aware equivalent.

This fixes #2569

### DIFF
--- a/src/main/java/com/owncloud/android/utils/MimeTypeUtil.java
+++ b/src/main/java/com/owncloud/android/utils/MimeTypeUtil.java
@@ -38,6 +38,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
+import javax.annotation.Nullable;
+
 /**
  * <p>Helper class for detecting the right icon for a file or folder,
  * based on its mime type and file extension.</p>
@@ -94,6 +96,7 @@ public class MimeTypeUtil {
      * @param account account which color should be used
      * @return Drawable of an image resource.
      */
+    @Nullable
     public static Drawable getFileTypeIcon(String mimetype, String filename, Account account, Context context) {
         if (context != null) {
             int iconId = MimeTypeUtil.getFileTypeIconId(mimetype, filename);
--- a/src/main/java/com/owncloud/android/utils/MimeTypeUtil.java
+++ b/src/main/java/com/owncloud/android/utils/MimeTypeUtil.java
@@ -22,6 +22,7 @@ import android.accounts.Account;
 import android.content.Context;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
+import android.support.v4.content.ContextCompat;
 import android.webkit.MimeTypeMap;
 
 import com.owncloud.android.R;
@@ -96,7 +97,7 @@ public class MimeTypeUtil {
     public static Drawable getFileTypeIcon(String mimetype, String filename, Account account, Context context) {
         if (context != null) {
             int iconId = MimeTypeUtil.getFileTypeIconId(mimetype, filename);
-            Drawable icon = context.getResources().getDrawable(iconId);
+            Drawable icon = ContextCompat.getDrawable(context, iconId);
 
             if (R.drawable.file_zip == iconId) {
                 ThemeUtils.tintDrawable(icon, ThemeUtils.primaryColor(account, context));
--- a/src/main/java/com/owncloud/android/utils/MimeTypeUtil.java
+++ b/src/main/java/com/owncloud/android/utils/MimeTypeUtil.java
@@ -94,14 +94,18 @@ public class MimeTypeUtil {
      * @return Drawable of an image resource.
      */
     public static Drawable getFileTypeIcon(String mimetype, String filename, Account account, Context context) {
-        int iconId = MimeTypeUtil.getFileTypeIconId(mimetype, filename);
-        Drawable icon = context.getResources().getDrawable(iconId);
+        if (context != null) {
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
:+1: looking good :)

tobiasKaminsky:
If, for some reason, context is null, we should at least show some dummy file icon:
R.drawable.file

ardevd:
@tobiasKaminsky I do agree, but if the `context` passed to `getFileTypeIcon()` is null then we cant really reference any drawables (afaik), right?

AndyScherzinger:
@tobiasKaminsky as stated by @ardevd https://github.com/nextcloud/android/pull/2570/files#diff-fbb14dfcb5f7ae95f086f403b1aeb672R100 will always need a context and thus lead to an NPE so the calling code would then need to take care of the situation if we simply can't load any drawable.

tobiasKaminsky:
Ah right, sorry, I was too fast.
I just wanted to avoid returning null. 

What about adding @NonNull to Context and if context is still, somehow null, use MainApp.getAppContext() as fallback? (though we ideally want to get rid of it…?)

AndyScherzinger:
> What about adding @nonnull to Context and if context is still, somehow null, use MainApp.getAppContext() as fallback? (though we ideally want to get rid of it…?)

This will fix explicit null values like the one you fixed in the other PR while it won't fix runtime issues where a getContext/getApplicationContext would return null (and we have seen this happening in the gplay console for whatever reason) then it'll still crash at runtime.

ardevd:
While I might not have a definite answer, in my experience you can never assume that the context will never be null. My Google Play developer console also shows the odd `NullPointerExceptions` where context is null.

AndyScherzinger:
So my vote would be to go with @ardevd's approach and annotate the return value with Nullable.

What do you think @ardevd @tobiasKaminsky?

ardevd:
@AndyScherzinger I agree with your approach. PR updated.

tobiasKaminsky:
:+1:

AndyScherzinger:
merged :) Thanks @ardevd :tada:

Issue #2532, opened 2018-04-30
Storage Continuously Increaseing


msonst:
### Actual behaviour
- Apps used storage increases significantly over time even though no files are kept local (7GB) (edit: setting do not delete locally in nextcloud. Manually deleting pictures once the content on phone once it appears in my backups)
- Only way to  (maybe) get rid of this is delete stored data via Google clean files, but then you'll get warned that you'd also remove settings, db, etc.

### Expected behaviour
- Almost constant storage usage when nothing kept local
- Way to safely clean app data, without being required to setup credentials again
 
### Steps to reproduce
1. Autosync dcim
2. Manual remove of pictures from device
3. Use for approx 7mo

### Environment data
Android version: 7

Device model: SM-G930A
Kernel 3.8.31-12564870

Stock or customized system: Stock

Nextcloud app version: 3.1.0

current unavailable, will add once I get back to the office

Nextcloud server version: 

#### Web server error log
```
Insert your webserver log here
```

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```
**NOTE:** Be super sure to remove sensitive data like passwords, note that everybody can look here! You can use the Issue Template application to prefill some of the required information: https://apps.nextcloud.com/apps/issuetemplate

tobiasKaminsky:
Can you use an app to see which folder is that big, e.g. DiskUsage?

grospolina:
I have the same issue on my device. The uploaded images remain in the temporary folder on my sd-card:
`SD-Card/Android/media/com.nextcloud.client/nextcloud/tmp/MyUsername@MyServer/MySyncedFolder`

### Environment data
Android version: 6.0.1
Device model: Sony Xperia Z3 Compact
Stock or customized system: Stock
Nextcloud app version: 3.1.0
Nextcloud server version: 13.0.2

msonst:
Hi,

Fthe folder looks similar in my side.

Sd/media/android/data/com.nextcloud.client/files/nextcloud/tmp/<user@domain>/<server folder structure>

Thanks Michael

tobiasKaminsky:
I created a fix for this, which will be soon in dev version. 
It would be cool if you can test it there :+1:

adyanth:
Is this still not fixed? I'm on 3.15.1 and currently nextcloud app is using 11+GB on my phone!

Mister-Muffin:
Same Problem here!
Version: 3.21.0 RC1

[...]

PR #2577, opened 2018-05-14
Better handling of temp file during upload

### LINKED ISSUE #2505: tmp folder is filled by auto upload
### Actual behaviour
When activating auto upload all files are copied into the folder /internal/nextcloud/tmp and left there after the upload. After some months of taking fotos and videos with my phone this is wasting more than 6GB of data. Deleting the files in the tmp folder causes a complete re-upload of all fotos and videos. My fotos and videos are stored by default on an SD card.

### Expected behaviour
I expected that the tmp folder will be cleared after the upload.
 
### Steps to reproduce
1. Choose a folder on an SD card for auto upload

### Environment data
Android version: 7.0 Security Patch 01.02.2018

Device model: Samsung J5 2017

Stock or customized system: Stock system

Nextcloud app version: 3.1.0

Nextcloud server version: 13.0.0

tobiasKaminsky:
Fix https://github.com/nextcloud/android/issues/2532
Fix #2505 

- use internal temp directory as it seems that external sd storage might not get deleted
- always try to delete temp file


mario:
What if the internal temp "drive" doesn't have enough space, which will often be the case in such strange setups? :)

tobiasKaminsky:
Good question…
We can first check this, or return a correct error message, if copying is not working.
What do you think?

mario:
@tobiasKaminsky both is good. First we check and if something happens during the copying process, we return the correct error message.

tobiasKaminsky:
> both is good. First we check and if something happens during the copying process, we return the correct error message.

This is now both implemented.

AndyScherzinger:
👍 looks good but needs a rebase 😃

PR #2579, opened 2018-05-14
Housekeeping of Activities Presenter and associated repository classes.

ardevd:
Minor cleanup of the Activities screen.

- Formatting fixes (indentation, newlines, etc)
- Adjusted access declarations
- Removed a hard coded error string and use the associated string resource entry instead.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activities/ActivitiesPresenter.java
+++ b/src/main/java/com/owncloud/android/ui/activities/ActivitiesPresenter.java
@@ -1,18 +1,18 @@
 /**
  * Nextcloud Android client application
- * <p>
+ *
  * Copyright (C) 2018 Edvard Holst
- * <p>
+ *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
  * License as published by the Free Software Foundation; either
  * version 3 of the License, or any later version.
- * <p>
+ *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU AFFERO GENERAL PUBLIC LICENSE for more details.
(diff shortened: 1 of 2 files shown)


src/main/java/com/owncloud/android/ui/activities/ActivitiesPresenter.java
```diff
@@ -37,9 +37,9 @@
     private final FilesRepository filesRepository;
 
 
-    public ActivitiesPresenter(@NonNull ActivitiesRepository activitiesRepository,
-                               @NonNull FilesRepository filesRepository,
-                               @NonNull ActivitiesContract.View activitiesView) {
+    ActivitiesPresenter(@NonNull ActivitiesRepository activitiesRepository,
```
mario: Hm but why remove public? Isn't it better to have an explicit scoping?

ardevd: @mario I'm simply following the principle of least privilege. Package-private, which is the default, is visible only within its own package which is all that is required in this case. Am I missing something?

mario: Aa you're right. What I meant is, it should say "private" in front :)

ardevd: package-private != private. Very different things. In this case, making it private would not be possible as the ActivitiesPresenter is accessed from the View as well. However, since the View resides in the same package, package-private access declaration is sufficient.

mario: Aaaah you're right. Sorry.

AndyScherzinger: I am fine with package private for classes especially since codacy is fine with that and only creates warnings if package-private is used for member variables and methods.

AndyScherzinger:
:+1: looks good to me

tobiasKaminsky:
:+1:

PR #2580, opened 2018-05-15
Cleanup of the TouchViewImage class

ardevd:
Removed redundant modifiers, useless statements and variable declarations.

### DIFF
--- a/src/main/java/com/ortiz/touch/TouchImageView.java
+++ b/src/main/java/com/ortiz/touch/TouchImageView.java
@@ -63,7 +63,7 @@ public class TouchImageView extends AppCompatImageView {
     //
 	private Matrix matrix, prevMatrix;
 
-    private enum State { NONE, DRAG, ZOOM, FLING, ANIMATE_ZOOM };
+    private enum State { NONE, DRAG, ZOOM, FLING, ANIMATE_ZOOM }
     private State state;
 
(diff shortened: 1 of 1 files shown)


mario:
Bonus points if you completely remove the usage of this class :P

ardevd:
@mario First I'd need to figure out why it's there in the first place :P

AndyScherzinger:
@ardevd that is easy to explain:
* double tap / pinch - to zoom on the (preview) images

What I would suggest is that we move to https://github.com/chrisbanes/PhotoView which would also support animated gifs :)

AndyScherzinger:
Plus would laos be that the refered lib is still maintained while the TouchView is dead

ardevd:
@AndyScherzinger Nice! I'll look into it asap.

ardevd:
@AndyScherzinger Sounds good to me. Added #2583 for reference.

mario:
Sure

tobiasKaminsky:
:+1: 

> support animated gifs

@AndyScherzinger this already works out of the box with android.
But replacing it with a newer, maintained version is a good idea.

AndyScherzinger:
:+1: 
> @AndyScherzinger this already works out of the box with android.

Right! Didn't know that :+1:

PR #2573, opened 2018-05-11
Iteration 2 for file detail & new sharing 

AndyScherzinger:
Iteration 1 can be found here: #2533 :ballot_box_with_check: 

WiP for #2485 and collab with @tobiasKaminsky

* [x] use butterknife for view binding
* [x] remove file type icon for non-preview-able files
* [x] support editing/deletion of existing sharees
* [x] support for expiration date
* [x] support adding new/further sharees
* [x] support sharing via link
* [x] add fetching of sharee avatars
* [x] fallback theming for checkable menu items
* [x] FIX #2610 :exclamation: 
* [x] replace `ListView`with `RecyclerView`

**Open for iteration 3/4**
* [ ] implement actual email share type menu options (behaves as is, so atm like a user share type)
* [ ] remove old/legacy ShareActivity and launch FileDetailFragment with sharing tab active
* [ ] MVP the sharing part

PR #2588, opened 2018-05-17
Support for trashbin

### LINKED ISSUE #2557: Trashbin view
@nextcloud/designers I am currently implementing trashbin view into android.
Web UI has only "delete" on overflow menu (three dots), which is a bit uncommon for me, if there is only one menu item. Is it okay to have the delete button directly next to restore?

To save some space I want to put the "deletion time" beneath the file name:
.Contacts-Backup
deleted 25 days ago

(file size will be removed).

What do you think?
@AndyScherzinger also for you :-)

tobiasKaminsky:
Fixes #2557

- initial support for trashbin
- introduces a common interface for OCFile and TrashbinFile
- no storage, all "on the fly" as it son meaningful to show possible outdated trashbin
- used MVP, but no repository pattern (I am unsure, if we really need this here. For files, activities, etc., we can try it so that we have two repositores (online & offline), but adding this here is "useless" overhead for me).

Discuss and maybe TODO in later PR: 
- multiselect items and delete them

Please have a look into it, especially @ardevd as you added MVP :-)


mario:
Build failed + conflicts.

AndyScherzinger:
@tobiasKaminsky the reason for that is that the server is returning a 404 :) My Server is 13.0.2

mario:
@AndyScherzinger you need 14. But I'd guess trashbin just shouldn't be there if it's not 14.

mario:
Also failure to load trashbin should be reflected via image in the middle + text, not snackbar.

AndyScherzinger:
Putting @mario's hints and comments into requirements :smiley: 

* thrashbin menu should not be shown on pre-14 servers
* in case of an error fetching the trashbin infos: 
1. show an aproppriate error icon (e.g. https://materialdesignicons.com/icon/alert-octagon)
2. show error headline
3. show error text line

AndyScherzinger:

AndyScherzinger:
@tobiasKaminsky for the error view have a look at https://github.com/nextcloud/android/pull/2573/commits/ec2459352434ea3748e021fbb8591743ab11da74 which just implemented that for the activities loading on the file details activities fragment :)

src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -352,7 +353,9 @@ public void run() {
         if (getAccount() != null) {
             FileDataStorageManager storageManager = new FileDataStorageManager(getAccount(), getContentResolver());
             OCCapability capability = storageManager.getCapability(getAccount().name);
-            if (capability.getFilesUndelete().isFalse() || capability.getFilesUndelete().isUnknown()) {
+            
+            if (ownCloudVersion.compareTo(OwnCloudVersion.nextcloud_14) < 0 &&
```
AndyScherzinger: wrong... My server is 13.0.2 so the items should be hidden, which it isn't and when opening it I then get the error screen you showed while will always get this pre-14 so it should def. just be hidden and imho `&&`needs to be replaced with `||`

AndyScherzinger:
@tobiasKaminsky the Nc14 check isn't properly working, see my comment :)

tobiasKaminsky:
@AndyScherzinger thanks for testing.
Strangly it worked for my nc13 instance, but you are right that having "||" is the right way.

AndyScherzinger:
:+1: Test with Nc13, all hidden->:+1:

tobiasKaminsky:
I changed the endpoint for thumbnail generation on trashbin files, this needs then https://github.com/nextcloud/android-library/pull/146.

tobiasKaminsky:
I changed build.gradle to use the library branch.
Prior merging this needs to be changed back to master.

AndyScherzinger:
@mario your review comments have been resolved, so ready for approval :)

:exclamation: Prior merging the lib rev has to be reset to master snapshot

AndyScherzinger:
@mario rebased to resolved latest merge conflicts due to today's merged PRs

AndyScherzinger:
@tobiasKaminsky works fine for me one thing I realized though is that the navigation state for the drawer menu is completely broken in case of (from anywhere) clicking "all files" because the state of the menu stays on the actual one and doesn't change to "all files" but you are then in all files and if you hit the all files menu item again the state changes and all files is checked in the menu...

I can't say if this issue is already present in the master branch or is introduced in this PR :/

mario:
@AndyScherzinger should probably be fixed here though.

Anyway, one small thing...can we make the error icon red? Thanks!

AndyScherzinger:
> @AndyScherzinger should probably be fixed here though.

it is also happending on master so I'd say we should rather open a new issue, mandatory for 3.3.0 (or even 3.2.1). What do you think @mario @tobiasKaminsky ?

> Anyway, one small thing...can we make the error icon red? Thanks!

We could do that, while we only use colored icons for folder icon and favorites. What do you think @tobiasKaminsky @jancborchardt @nextcloud/designers ?

jancborchardt:
Instead of

> Error icon
> Error
> Loading trashbin failed!

Could we show:

> Trash icon
> Problem loading deleted files
> [ Retry ]

Retry is a button, and never ever say »trashbin« – it’s called »Deleted files«. :)

Also, the snackbars for »could not be deleted« on individual files should all have a »Retry deletion« (or if too long, just »Retry«) action on the right.

And no, the icon should not be red, it’s way too alarming for a case which might just be a network error. If the »Deleted files« app is not enabled, the navigation entry should simply not show.

AndyScherzinger:
Right now we do not have any logic for retry buttons anywhere so the first step would probably be to have

> Trash icon
> Error
> Problem loading deleted files

and in a new PR we could start adding retry buttons in all places where we use this kind of "empty/error view". Would that make sense @jancborchardt? Else we would have a retry button for the Deletes files while we don't for things like the activity stream etc. Other than that the empty view usually has a pull-to-refresh to do exactly that so we should then probably not introduced several, different ways to "retry/refresh" or should we?

jancborchardt:
@AndyScherzinger good call to do it in a next step. Regarding pull-to refresh: Yes, it’s there, but still not soo discoverable to regular users as we think. A retry button directly there is very friendly and puts the next step directly in your view.

AndyScherzinger:
> A retry button directly there is very friendly and puts the next step directly in your view.

The retry button makes total sense to me, my question is if we should then have both (button and pull-to-refresh) or drop the pull then? Sorry for the confusion.

Personally I'd be fine with having both in place.

jancborchardt:
@AndyScherzinger sorry – to clarify: Both should be there. :)

mario:
@ardevd do you have time to review this?

Also, @tobiasKaminsky  - rebase.

PR #2590, opened 2018-05-18
Lib refers always to master

tobiasKaminsky:
- lib refers always to master
- only releases, in our release branch, target a specific library version

This way we can merge any PR to library without having to release inbetween version.


tobiasKaminsky:
Lint is complaining due to "unused quanity", which is fixed in #2591.

ardevd:
+1

Issue #2596, opened 2018-05-20
Snackbar messages overlap the navigation bar when viewing images


ardevd:
### Actual behaviour
- When viewing a image in the app, the view overlaps the translucent navigation bar at the bottom meaning view elements aligned to the bottom overlap with the navigation bar. One example being the snackbar messages that can appear when viewing images. 

### Expected behaviour
- View elements should not overlap with the navigation bar.
 
### Steps to reproduce
1. Preview a .gif or .svg
2. Notice the snackbar message appearing at the bottom behind the navigation bar.

### Environment data
Android version: 8.1

Device model: Emulator

Stock or customized system: Stock

Nextcloud app version: 3.1.0

Nextcloud server version: 13.0.2

![device-2018-05-20-192651](https://user-images.githubusercontent.com/366855/40281565-c833ad7a-5c63-11e8-89a0-34443366becc.png)

AndyScherzinger:
Fixed via #2599

PR #2599, opened 2018-05-21
res:styles: Don't use a translucent navigation bar in the Preview activity

ardevd:
This causes view elements at the bottom to overlap with the navigation bar which makes it impossible to interact with said elements. 

With a normal navigation bar the snackbar messages in the Preview fragment will appear correctly above the navigation bar and tapping the image will still hide the navigation bar just like before.

Unless anyone feels very strongly about the navigation bar having to be translucent I think this is a good solution to a pretty nasty bug.

### DIFF
--- a/src/main/res/values-v19/styles.xml
+++ b/src/main/res/values-v19/styles.xml
@@ -6,7 +6,6 @@
         <item name="android:windowActionBarOverlay">true</item>
         <item name="android:windowFullscreen">true</item>
         <item name="android:windowTranslucentStatus">true</item>
-        <item name="android:windowTranslucentNavigation">true</item>
         <!-- Support library compatibility -->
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
Totally fine by me, what do you think @mario @tobiasKaminsky ?

mario:
👍

tobiasKaminsky:
I built this branch but I cannot see any difference.
Can you show us/me two pictures to compare the changes?

ardevd:
@tobiasKaminsky See #2596 for a screenshot of the issue. With this PR, at least on my devices and emulator, the snackbar appears properly above the navigation bar.

tobiasKaminsky:
:+1: 
Now I got it, thanks :-)

PR #2602, opened 2018-05-21
ui:preview Simplified generateCheckerboardLayeredDrawable()

ardevd:
Simplified `generateCheckerboardLayeredDrawable() `by replacing the separate calls to `setLayerWidth` and `setLayerHeight` with `setLayerSize()`. Also added a check to verify that `getActivity() `is not null before generating the `layerDrawable`.


mario:
Almost good. Now we just need to prevent getting height and width of images twice :)

Fix this and I approve.

ardevd:
@mario Good catch! Code simplified :)

mario:
👍

PR #2603, opened 2018-05-22
Disable analytics via manifest

mario:
Please make a point release with this.

### DIFF
--- a/src/gplay/AndroidManifest.xml
+++ b/src/gplay/AndroidManifest.xml
@@ -28,6 +28,9 @@
         android:theme="@style/Theme.ownCloud.Toolbar"
         android:manageSpaceActivity="com.owncloud.android.ui.activity.ManageSpaceActivity">
 
+        <meta-data android:name="firebase_analytics_collection_deactivated" android:value="true" />
+        <meta-data android:name="google_analytics_adid_collection_enabled" android:value="false" />
+
         <meta-data android:name="com.google.android.gms.version"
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
RC1 is planned to be released on 28th.
Can you also give some background please? We do not have analytics enabled at built time, or?

mario:
We also don't have it disabled. Since the GPlay version ships with Analytics, and even if we don't really track anything, Firebase still sends some random data which might not be compliant with GDPR, so we need to do it now.

This disables it.

mario:
Ah, @tobiasKaminsky , can wait until 28th - I was smart enough to disable it via code earlier.

ardevd:
:+1:

mario:
@AndyScherzinger afaik all analytics code has been removed by @tobiasKaminsky earlier anyway, as far as I know :)

mario:
So there's not much to brand :P

AndyScherzinger:
> So there's not much to brand :P

I know but how could this feature (which has been built for paying customers) be activated by a customer after this PR has been merged?

mario:
@AndyScherzinger it can't, for a good reason. Unless we implement a consent/consent withdrawal, this feature is useless.

AndyScherzinger:
merged then :+1:

mario:
Cool. As soon as we get a proper consent workflow (both on start-up and in settings), we can remove this.

PR #2613, opened 2018-05-24
setting the menu item order according to the webUI

### LINKED ISSUE #1749: Single file: overflow menu: same order as web ui
Web UI:
image

Android:
image

As far as I remember we said that those entries that are android only should be at bottom, @jancborchardt?

AndyScherzinger:
Fixes #1749

simply a reordering of the menu items.

cc @tobiasKaminsky @mario @ardevd for review

### DIFF
--- a/src/main/java/third_parties/daveKoeller/AlphanumComparator.java
+++ b/src/main/java/third_parties/daveKoeller/AlphanumComparator.java
@@ -105,6 +105,11 @@ public class AlphanumComparator<T> implements Comparator<T>, Serializable {
     }
 
     public int compare(String s1, String s2) {
+        if ((s1 == null) || (s2 == null))
+        {
+            return 0;
+        }
+        
         int thisMarker = 0;
         int thatMarker = 0;
         int s1Length = s1.length();
--- a/src/main/java/third_parties/daveKoeller/AlphanumComparator.java
+++ b/src/main/java/third_parties/daveKoeller/AlphanumComparator.java
@@ -105,11 +105,6 @@ public class AlphanumComparator<T> implements Comparator<T>, Serializable {
     }
 
     public int compare(String s1, String s2) {
-        if ((s1 == null) || (s2 == null))
-        {
-            return 0;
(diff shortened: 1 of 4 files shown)


mario:
👍 
orderInCategory 1 on everything *shrug*

AndyScherzinger:
> orderInCategory 1 on everything shrug

@mario Just removed it via https://github.com/nextcloud/android/pull/2613/commits/cd3953b32fa963e63b7cb444505fcb34bda125b5 :+1:

mario:
Thanks.

src/main/java/third_parties/daveKoeller/AlphanumComparator.java
```diff
@@ -105,6 +105,11 @@ public int compare(T t1, T t2) {
     }
 
     public int compare(String s1, String s2) {
+        if ((s1 == null) || (s2 == null))
+        {
+            return 0;
```
tobiasKaminsky: Why is this in this PR?
I am also not sure if we want to return "equal" (which is 0) when one of the two strings is null.
Please remove this and if needed, do it in a different PR. For sorting we have also test cases which then should be extended with "null" values. Currently we still have from time to time the TimSort problem, so any change should be decently tested.

tobiasKaminsky:
:-1: 
Please remove the sorting stuff

src/main/java/third_parties/daveKoeller/AlphanumComparator.java
AndyScherzinger: Stupid but done... have fun with the possible NPEs... not opening a PR for search + TimSort has oh so nothing to do with this since it'll crash anyways if one of the values is null... :roll_eyes:

AndyScherzinger: > Why is this in this PR?
Because the actual implementation is obviously wrong? Dough....

tobiasKaminsky:
:+1:

PR #2616, opened 2018-05-24
Remove analytics

tobiasKaminsky:
As discussed this will completely remove analytics.
If we ever need this again, it will be in a different build flavor, so that regular gplay flavor will not have it included.


AndyScherzinger:
:+1:  Nice :smiley:

mario:
👍 

With a remark that I'm definitely not fan of the approach taken, but some battles cannot be won.

PR #2618, opened 2018-05-25
Obsolete Gradle Dependency

### LINKED ISSUE #2614: Obsolete Gradle Dependency
### Actual behavior
Warnings are shown by Android Lint as some Gradle dependencies are outdated.

### Expected behavior
Gradle dependencies should be updated to the latest version.

### Environment data
Warnings are shown in all build variants.

Stock or customized system: Stock, dev

Nextcloud app version: 3.1.0

screenshot from 2018-05-24 16-08-47

Would you like to work on this issue?
Yes, I want to work on this issue.

gupta1anubhav:
Fixes #2614 
I have updated the outdated gradle dependencies to the latest version.
Build is successful on all build variants.
App is tested on Redmi Note 4.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -32,6 +32,7 @@ pipeline:
     commands:
       # needs gplay
       - sed -i '/.*com.google.*/s/^.*\\/\\///g' build.gradle
+      - sed -i '/.*mobilesdk.*/s/""/"1:111111111111:android:aaaaaaaaaaaaaaaa"/g' src/*/google-services.json
       - export BRANCH=$(scripts/lint/getBranchName.sh $GIT_USERNAME $GIT_TOKEN $DRONE_PULL_REQUEST)
       - scripts/lint/lint-up-wrapper.sh $GIT_USERNAME $GIT_TOKEN $BRANCH $LOG_USERNAME $LOG_PASSWORD $DRONE_BUILD_NUMBER
     secrets: [ GIT_USERNAME, GIT_TOKEN, LOG_USERNAME, LOG_PASSWORD ]
--- a/.drone.yml
+++ b/.drone.yml
@@ -4,6 +4,7 @@ pipeline:
     commands:
       # uncomment gplay for Gplay
       - sh -c "if [ '${FLAVOR}' = 'Gplay' ]; then sed -i '/.*com.google.*/s/^.*\\/\\///g' build.gradle; fi"
+      - sed -i '/.*mobilesdk.*/s/""/"1:111111111111:android:aaaaaaaaaaaaaaaa"/g' src/*/google-services.json
       # build app and assemble APK
       - sh -c "if [ '${FLAVOR}' != 'Lint' ]; then ./gradlew assemble${FLAVOR}; fi"
     when:
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ buildscript {
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:3.1.2'
-        classpath 'com.google.gms:google-services:3.0.0'
+        classpath 'com.google.gms:google-services:3.2.0'
         classpath 'org.codehaus.groovy:groovy-all:2.4.12' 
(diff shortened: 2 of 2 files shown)


AndyScherzinger:
Hi @gupta1anubhav Thanks for the PR!
Unfortunately the builds are failing hard, see https://drone.nextcloud.com/nextcloud/android/4672/8
The reason fort this is:
```
* What went wrong:
Execution failed for task ':processGplayDebugGoogleServices'.
> Missing Google App Id. Please follow instructions on https://firebase.google.com/ to get a valid config file that contains a Google App Id
```

@mario @tobiasKaminsky any idea how to fix this?

tobiasKaminsky:
It seems that they are now checking the app id during compilation.
We have an app id for releasing the app to google play, but we keep this config private.
Now it just uses a dummy one.

mario:
@tobiasKaminsky there's no real need for it. See Talk on how I configure Firebase.

tobiasKaminsky:
@mario can you give us an hint? I searched for firebase|google-services in talk, but did not found any config.

mario:
@tobiasKaminsky absolutely. Look at the push notifications config: https://github.com/nextcloud/talk-android/blob/master/app/src/main/res/values/setup.xml

Values are extracted from the configuration file.

tobiasKaminsky:
Are those the real ids? I remember, that you said that the google-services.json should not be checked in into git?

mario:
@tobiasKaminsky they are. Indeed, I did. But lately I've looked into it further and there's absolutely no harm in doing so (even Google says so). Sorry for the late response, somehow I missed this ticket.

tobiasKaminsky:
> @tobiasKaminsky they are. Indeed, I did. But lately I've looked into it further and there's absolutely no harm in doing so (even Google says so). Sorry for the late response, somehow I missed this ticket.

So I will put the complete google-service.json in src/gplay/ (those that I got via email long long time ago), @mario?
(there is also api_key and some references to iOS in it).

mario:
Please don't do that and instead extract the values like I did. That allows us to use Push on F-Droid apps that have MicroG installed.

mario:
(Clearly later, once firebase SDK is open sourced, like it should be - but doing it this way removes the need for a 

    // uncomment for gplay (must be at the bottom)
// apply plugin: 'com.google.gms.google-services'

tobiasKaminsky:
Can you explain a bit further how to do this?
I searched the talk app code, but e.g. firebase_database_url is never used, neither in gplayDebug (here I also used gplay.gradle) nor in fdroidDebug.

In addition how will microG know these values?

mario:
I clarified myself further in the second comment. It's the Firebase SDK which is responsible for using the correct string values which is the reason why the names of those strings are as they are here: 
https://github.com/nextcloud/talk-android/blob/master/app/src/main/res/values/setup.xml

The reason why my approach will eventually work on F-Droid is because we drop the proprietary dependency of com.google.gms.google-services, and the Firebase SDK will be open source soon enough.

In order to get the values you need open the google-services.json and find the relevant app (in this case Nextcloud Files) and get values from there.

mario:
Let me know and if needed, I can do this.

tobiasKaminsky:
Is Firebase finding the string via R.class automatically? 

If we then still need to manually copy the google-service.json, then we can also go with my approach and use a dummy value.
(Until firebase sdk is open source and really working with microG).

mario:
Yes it is - it finds everything automatically.

Anyway if you insist on using the google-services.json file (which I don't see why would you, but hey) then I'll provide a new one which holds only Nextcloud Files values.

tobiasKaminsky:
No no, I understood https://github.com/nextcloud/android/pull/2618#issuecomment-397554458 that we still need the google services file.

I guess it is easier, that you just change this PR in the way it is needed and then I try to understand this gcm/firebase/push mess ;-)

tobiasKaminsky:
Successor is in #2825

PR #2619, opened 2018-05-25
ui:activity Removed unused override method in ManageAccountsActivity

ardevd:
Overriding a method just to call the same method from the super class without performing any other actions is useless and misleading.


AndyScherzinger:
:+1:

tobiasKaminsky:
:+1:
It seems that I overlooked a few things on #2616, but luckily I have you guys :tada:

PR #2621, opened 2018-05-25
ui:activity: Removed @NonNull annotation for method where null can be returned

ardevd:
The return value is marked @NonNull and is therefore assumed to only return non-null values. Therefore setting one of these values to null, or failing to set such a class field in a constructor, could cause NullPointerExceptions at runtime. In this case, view might be null.


AndyScherzinger:
:+1:

tobiasKaminsky:
:+1:

PR #2622, opened 2018-05-25
ui:preview Verify that bitmap is not null before setting imageDrawable.

ardevd:
A reference to null should never be dereferenced/accessed. Doing so will cause a NullPointerException to be thrown


AndyScherzinger:
:+1:

tobiasKaminsky:
:+1:

PR #2664, opened 2018-06-01
Fix for #2629

### LINKED ISSUE #2629: Syntax Errors in SQL
### Actual behaviour
- There are various syntax errors in FileContentProvider.java  which are affecting the functioning of SQL database.

### Expected behaviour
- These errors should be removed .

### Environment data

Nextcloud app version: latest build version

screenshot from 2018-05-28 11-33-15

### Would you like to work on this issue?
Yes , I want to work on this issue.

gupta1anubhav:
Fixes #2629 
Lot of Android Lint errors in the syntax of SQL are fixed.
The database is working as expected.

### Tested On: Redmi Note 4 (API 24)

### DIFF
--- a/src/main/java/com/owncloud/android/providers/FileContentProvider.java
+++ b/src/main/java/com/owncloud/android/providers/FileContentProvider.java
@@ -736,8 +736,8 @@ public class FileContentProvider extends ContentProvider {
                 + ProviderTableMeta.FILE_ETAG + TEXT
                 + ProviderTableMeta.FILE_SHARED_VIA_LINK + INTEGER
                 + ProviderTableMeta.FILE_PUBLIC_LINK + TEXT
-                + ProviderTableMeta.FILE_PERMISSIONS + " TEXT null,"
-                + ProviderTableMeta.FILE_REMOTE_ID + " TEXT null,"
+                + ProviderTableMeta.FILE_PERMISSIONS + " TEXT,"
+                + ProviderTableMeta.FILE_REMOTE_ID + " TEXT ,"
                 + ProviderTableMeta.FILE_UPDATE_THUMBNAIL + INTEGER //boolean
                 + ProviderTableMeta.FILE_IS_DOWNLOADING + INTEGER //boolean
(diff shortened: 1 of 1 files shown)


gupta1anubhav:
Sorry, @AndyScherzinger I was so focused on the errors are shown by Android that I unknowingly changed column name. This will not happen again.

AndyScherzinger:
No worries, all good @gupta1anubhav, like I said I don't want it to sound harsh but need to make sure the PR doesn't get merged in its current state :)

AndyScherzinger:
Hi @gupta1anubhav, are you planing on updating the PR? Or should we close it?

gupta1anubhav:
Can you please tell @AndyScherzinger what changes should I do?

AndyScherzinger:
@mario would you mind taking a look. I am not certain if we can just change the data type of a column (even for string/text).

@gupta1anubhav what definitely needs to be fixed is https://github.com/nextcloud/android/pull/2664/files#diff-f9acf3921f3be4a83ad881a769f5231eR871 like mentioned before via https://github.com/nextcloud/android/pull/2664#issuecomment-395050110

mario:
@AndyScherzinger ergh this only does it for new tables if I'm reading it right. Where's the upgrade procedure etc etc? xD

mario:
Two things to note though!

1) "The subtle thing to note here is that SQLite does not enforce the data type of values you put into columns. That means that you can put text into a numeric field, and so on."

2) Difference between String and Text:

"If a string looks like a numeric value, STRING will convert it into a numeric value, while TEXT will not perform any conversion."

mario:
(In general I don't think anything BAD would happen given our usage of this, but as it is subtle and each vendor ships their own sqlite implementation (we don't yet bundle sqlite binary with the app) this might backfire in the ways we don't understand)

mario:
@AndyScherzinger @tobiasKaminsky ?

tobiasKaminsky:
I vote to do no changes to the database as it is working right now (even with the warnings), so closing this PR.

AndyScherzinger:
I agree with Tobias‘ assassment

PR #2665, opened 2018-06-01
fix transifex mappings: he->iw and id->in

AndyScherzinger:
fix 2 lint warnings:
* `../../../src/main/res/values-he`: The locale folder `"he"` should be called `"iw"` instead; see the `java.util.Locale` documentation
* `../../../src/main/res/values-id`: The locale folder `"id"` should be called `"in"` instead; see the `java.util.Locale` documentation

cc @MorrisJobke since you know the transifex bot way better than me :smile:


tobiasKaminsky:
Ha, funny. 
I already created a PR this morning in https://github.com/nextcloud/docker-ci/pull/95.
That does the same but on a file level basis and for all android apps.
Maybe that one is then a bit more generic and better instead of changing it on every android app?

AndyScherzinger:
Nice!
Yeah, fixing it in the docker image would be better, while I'd say we just do both, since this way we don't have to wait for a new translation on one of these locales or a removal of a string since else these files won't get an update and will stay this way. 

Having both will fix the situation quicker :) Just my pov though. I am fine either way

tobiasKaminsky:
:+1: 
@MorrisJobke dislike the central way, so we do it in every app.

tobiasKaminsky:
We then should do this also a PR on the other repos.

MorrisJobke:
> Yeah, fixing it in the docker image would be better, while I'd say we just do both, since this way we don't have to wait for a new translation on one of these locales or a removal of a string since else these files won't get an update and will stay this way.

I just don't like it, because there is an official way of doing it and then we add another hack on top of it and when debugging this it is just confusing. Also we don't have a lot of android repos. 😉

MorrisJobke:
> We then should do this also a PR on the other repos.

passman, news and talk basically ;)


And https://github.com/nerzhul/ownCloud-SMS-App if you like to (those are the ones handled on our transifex sync setup)

AndyScherzinger:
cc @mario for approval :)

AndyScherzinger:
Any chance we get this merged before 2020? :stuck_out_tongue:

MorrisJobke:
Looks good :+1:


[Approved with PullApprove](https://pullapprove.com/nextcloud/android/pull-request/2665/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=nextcloud/android)

mario:
👍 

There.

mario:
@AndyScherzinger please make the same PR against Talk :)

AndyScherzinger:
> @AndyScherzinger please make the same PR against Talk :)

@mario will do!

AndyScherzinger:
@mario here is the talk PR: https://github.com/nextcloud/talk-android/pull/183

MorrisJobke:
> @MorrisJobke So only news I guess... Passman wants to drop all repos since MS bought GH

🙈

PR #2666, opened 2018-06-01
lint: ignore setJavaScriptEnabled

AndyScherzinger:
we enable this for all our web views which is fine and lint just adds a warning so this is carefully reviewed. Thus the only way to then get rid of the warning is to suppress it.


tobiasKaminsky:
:+1:

AndyScherzinger:
cc @mario for approval

mario:
👍

PR #2680, opened 2018-06-04
Use context instead of static getAppContext

tobiasKaminsky:
Found via google console that at some point getAccountType had a NPE, which comes from getAppContext, I think.
As we somewhere discussed to get rid of these static calls, this is a first step.


AndyScherzinger:
:+1: fine by me

mario:
👍

mario:
Build failed?

AndyScherzinger:
> Build failed?
@mario I'll have a look, it failed because the lint warnings count is higher than on master

PR #2682, opened 2018-06-04
ui:fragment: Use isEmpty() to check if list of activities is empty.

ardevd:
Using Collection.size() to test for emptiness works, but using Collection.isEmpty() makes the code more readable and can be more performant

### DIFF
--- a/src/main/java/com/owncloud/android/ui/fragment/FileDetailActivitiesFragment.java
+++ b/src/main/java/com/owncloud/android/ui/fragment/FileDetailActivitiesFragment.java
@@ -243,7 +243,7 @@ public class FileDetailActivitiesFragment extends Fragment implements ActivityLi
 
                     activity.runOnUiThread(() -> {
                         populateList(activities, ownCloudClient, pageUrl == null);
-                        if (activities.size() > 0) {
+                        if (!activities.isEmpty()) {
                             empty.setVisibility(View.GONE);
                             list.setVisibility(View.VISIBLE);
                         } else {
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/ui/fragment/FileDetailActivitiesFragment.java
```diff
@@ -243,7 +243,7 @@ private void fetchAndSetData(String pageUrl) {
 
                     activity.runOnUiThread(() -> {
                         populateList(activities, ownCloudClient, pageUrl == null);
-                        if (activities.size() > 0) {
+                        if (!activities.isEmpty()) {
```
AndyScherzinger: wouldn't it be even nicer for readability if the statement isn't inverted as in `if (activities.isEmpty())` so you'd flip the two code blocks but I think that be easier to read :)

tobiasKaminsky: There are recommendations out there that the positive if branch is always the first.
But having such "inverted" condition, it is tricky to say, if we want to have the positive if condition as first ("we have no activities" or having the positive result (which then would be "we have some activities").

Anyway, I have no clear opinion on that, just do what you prefer ;-)

tobiasKaminsky:
Commented, but otherwise :+1:

AndyScherzinger:
@ardevd your choice ;) Do you want to flip it or leave it as-is?

mario:
@ardevd we need you!

ardevd:
@mario I'll take care of it ASAP I promise. Damn vacation gets in the way of my coding efforts. I'll fix this PR tonight or tomorrow for sure

mario:
@ardevd enjoy your vacation! :)

ardevd:
@tobiasKaminsky @AndyScherzinger I agree with you guys and have inverted the if statement for increased readability.

ardevd:
@tobiasKaminsky wow, I really appreciate the gesture. That would be awesome. I'll email you :)

AndyScherzinger:
Unfortunately I can't make it this year but it is a great way to meet in person and have a good time :) Been to the conference the last 2 years :+1:

PR #2683, opened 2018-06-04
util: Added @Override annotation where applicable.

ardevd:
Using the @Override annotation is useful for two reasons :

It elicits a warning from the compiler if the annotated method doesn't actually override anything, as in the case of a misspelling.
It improves the readability of the source code by making it obvious that methods are overridden.


AndyScherzinger:
:+1: Nice catch!

mario:
👍

PR #2685, opened 2018-06-05
Add support for file versions

tobiasKaminsky:
- adds support for versions
- only visible for NC14 user
- they are shown in activity stream
	- you can click on "revert" icon
- currently this is not MVP, but I will open up a new PR after the other file detail PRs by @AndyScherzinger are merged

TODO
- [x] going back from detail view, reload of file list needs to be done, @AndyScherzinger do you have an idea for this?
- [x] after successful restore: current downloaded file needs to be removed
- [x] @rullzer what about old thumbnails? E.g. you have an image you change?


AndyScherzinger:
> going back from detail view, reload of file list needs to be done, @AndyScherzinger do you have an idea for this?

There is no going back ;) On a tablet it would be a split pane between file list and file details :) I also did this for the file details already by introducing a listener interface (implemented by the activity) so you could just extend that one and ping the activity that it needs to refresh the other fragment/list :)

rullzer:
@tobiasKaminsky how do you currently evaluate if the preview is still valid? Because you could use the same logic here. Either check the etag of the preview against the endpoint. Or check the etag of the actual file that should also change.

tobiasKaminsky:
Thanks @rullzer, currently we are checking during sync if the modification date has changed and if so, we simply fetch a new thumbnail.
As we currently do not check eTags for thumbnails, I will create this in a different branch, so that this one will not blown up.

tobiasKaminsky:
As comment PR is based on this, I did not rebase, but merge.
As soon as this is then ready, prior merging this PR we need to rebase, to get rid of the merge commit.

tobiasKaminsky:
This is now ready for code review & testing.

src/main/java/com/owncloud/android/ui/fragment/FileDetailFragment.java
```diff
@@ -183,6 +183,15 @@ public FileDetailSharingFragment getFileDetailSharingFragment() {
         return ((FileDetailTabAdapter)viewPager.getAdapter()).getFileDetailSharingFragment();
     }
 
+    /**
+     * return the reference to the file detail activity fragment to communicate with it.
+     *
+     * @return reference to the {@link FileDetailSharingFragment}
+     */
```
AndyScherzinger: --> FileDetailActivitiesFragment ;)

tobiasKaminsky: Copy&paste error ;-)

AndyScherzinger:
@tobiasKaminsky I only get an error screen (with the icon and so on) when opening a file that has versions on your testing Nc14 server instance :/

tobiasKaminsky:
@AndyScherzinger can you write me on irc or per mail the exact file alongside with the credentials? Then I can have a look at exact the same file.

tobiasKaminsky:
@AndyScherzinger this was because I cloned server manually and forgot activity app.
Please re-test it.

! I rebased it!

src/main/res/drawable/ic_history.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
```
AndyScherzinger: can/should we add a license header here, donb't know under which license the Nextcloud icons are published though

AndyScherzinger:
:+1: works like a charm add @tobiasKaminsky - I also fixed any (really tiny, minor) issues I found :+1: 

Only one small question remained regarding license header of the restore icon.

src/main/res/drawable/ic_history.xml
tobiasKaminsky: @nextcloud/designers maybe? :-)

tobiasKaminsky:
@AndyScherzinger thanks for your feedback, I added everything :+1:

src/main/res/drawable/ic_history.xml
jancborchardt: AGPLv3, and if that doesn’t work CC-BY-SA :)

mario:
👍

AndyScherzinger:
@tobiasKaminsky shouldn't you have merged https://github.com/nextcloud/android-library/pull/150 first and moved back to referencing master-SNAPSHOT???

tobiasKaminsky:
Ups. 
As I did not wrote this in the first comment, I just forgot about it :/

@mario can you quickly have a look at https://github.com/nextcloud/android-library/pull/150? :-D

AndyScherzinger:
@tobiasKaminsky nextcloud/android-library#150 has been merged and master has been build via jitpack

PR #2696, opened 2018-06-08
Disable lint check for gradle dependencies

tobiasKaminsky:
Once a new version comes up, this prevents all our PRs from merging as lint is complaining.
From time to time we simply should look at our deps and update them.
This should happen then in a separate PR, with extensive testing.


AndyScherzinger:
I am unsure if I should +1 or -1 this PR since it completely disagree with the implications this will have...

> Once a new version comes up, this prevents all our PRs from merging as lint is complaining.

Which to me the motivation to ignore this kind of lint warnings, so +1

> From time to time we simply should look at our deps and update them.

Imho no, we should always instantly update any lib update which is a bugfix or minor version update, so it gets shipped to the dev builds and the alpha testers so to speak. Else we simply slow down, raising the risk because you have to update multiple libs at once

> This should happen then in a separate PR, with extensive testing.

The point above, I am fine with a separate PR but I am totally against the "extensive testing"-pseudo argument, because it isn't one. If minor lib updates ship to the dev builds fast and thus update all PRs/ branches then there is no need to do so since they are implicitely tested. Also in case a lib update fixes a critical issue it is the same thing as with the authentication fix (context issue): If it takes several days for a fix to hit master and being included in dev builds and on other PRs code bases then QA as in 2 peer review doesn't add value it blockes other QA measures like the dev builds.

TL;DR The time it takes at the moment to get the two "mandatory" peer reviews is way to long so I don't understand why we want to make it even worse. Imho (out of scope for this PRs discussion) we need to find a way back to faster adding value which is 
a) shipping things and 
b) getting stuff merged to master

Writing code is preparing a value add and reviews/tests are a QA measure but the only value add when a change is made available to the users... just 2 two cents

mario:
I would -1 this.

tobiasKaminsky:
Ok, what about the other way around? 
We update the libs as soon as possible, so that they very fast in latest dev version, so we have plenty of time before the next release, to see if everything works?

Those PRs then can be merged without review?

AndyScherzinger:
> We update the libs as soon as possible, so that they very fast in latest dev version, so we have plenty of time before the next release, to see if everything works?

That is the idea, yes

> Those PRs then can be merged without review?

Imho that is how we should do it, given the prerequisite that the build turns green (so no compile errors).

tobiasKaminsky:
Ok, then let us do it this way.

AndyScherzinger:
@tobiasKaminsky the server team uses a bot to auto update their dependencies, don't know if that could also be used for the Android app (dependabot)

tobiasKaminsky:
At a first glance there is only alpha support for java, but nothing for gradle.
Also i am unsure if this will work with our commented gplay stuff, but it would be indeed nice

PR #2698, opened 2018-06-09
fragment:util: Added private constructor to hide the inherit public one

ardevd:
Utility classes, which are collections of static members, are not meant to be instantiated.

Java adds an implicit public constructor to every class which does not define at least one explicitly. Hence, at least one non-public constructor should be defined.


AndyScherzinger:
:+1: Nice catch @ardevd !

mario:
👍

PR #2733, opened 2018-06-20
Update release cycle description

tobiasKaminsky:
- 2 months developement cycle
- library also gets dot releases, if needed

We discussed this sometime ago, but it should be also added to description.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -182,7 +182,7 @@ For dev the version name is in format YYYYMMDD. It is mainly as a reference for 
 * after feature freeze a public release candidate on play store and f-droid is released
 * ~2 weeks testing, bug fixing
 * release final version on f-droid and play store
-* Bugfix releases (dot releases, e.g. 3.2.1) are released on demand from the branch created with first stable release (stable-3.2.x). If changes to the library are required, we do the same: create a branch from the version used in stable release (e.g. 1.0.43) and then release a dot release (1.0.43.1).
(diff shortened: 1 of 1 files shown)


mario:
👍

CONTRIBUTING.md
```diff
@@ -178,11 +178,11 @@ beware, that beta releases for an upcoming version will always use the minor and
 For dev the version name is in format YYYYMMDD. It is mainly as a reference for reporting bugs and is not related to stable/release candidates as it is an independent app.
 
 ## Release cycle
-* for each release we choose several PRs that will be included in the next release. Currently there are many open PRs from ownCloud, but after merging them, the intention is to choose the PRs that are ready (reviewed, tested) to get them merged very soon.
-* these will be merged into master, tested heavily, maybe automatic testing
+* Releases are planned every ~2 months, with 6 weeks of developing and 2 weeks of stabilising
 * after feature freeze a public release candidate on play store and f-droid is released
 * ~2 weeks testing, bug fixing
 * release final version on f-droid and play store
+* Bugfix releases (dot releases, e.g. 3.2.1) are released on demand from the branch created with first stable release (stable-3.2.x). If changes to the library are required, we do the same: create a branch from the version used in stable release (e.g. 1.0.43) and then release a dot release (1.0.43.1).
```
AndyScherzinger: We should as for the dot-releases for the lib also use the correct semantic versioning which would then be 1.1.x for dot releases which would lead to now using the 1.x.0 for feature releases of the lib and I think we should go for that versioning schema (like we do for the app)

tobiasKaminsky: Fine from my side.
Then let us start with next release (3.3.0) with lib 1.1.0, and dot release would then be 1.1.1?

AndyScherzinger: Sounds right to me :+1:

AndyScherzinger:
:+1:

PR #2749, opened 2018-06-22
Intent may be null

tobiasKaminsky:
Found via google play console.

> The Intent supplied to Context.startService, as given. This may be null if the service is being restarted after its process has gone away, and it had previously returned anything except START_STICKY_COMPATIBILITY.


mario:
👍

PR #2750, opened 2018-06-22
Account may be null

tobiasKaminsky:
Found via google play console.

I have no steps to reproduce and I do not understand how we can have no valid account, when opening manageAccountActivity, but this should help to prevent a crash.


AndyScherzinger:
:+1:

mario:
👍

PR #2760, opened 2018-06-26
Added missing @Override annotations above applicable method signatures.

ardevd:
Using the @Override annotation is useful for two reasons:

- It elicits a warning from the compiler if the annotated method doesn't actually override anything, as in the case of a misspelling.
- It improves the readability of the source code by making it obvious that methods are overridden.


tobiasKaminsky:
:+1:

mario:
👍

AndyScherzinger:
:+1:

AndyScherzinger:
Any idea why the lint warnings increased by 28?!

@ardevd is the branch up to date with latest master?

AndyScherzinger:
Just checked the master and the master is now also at 28 new warnings. Merging then

Issue #2287, opened 2018-03-05
Release x.y.Z - Roadmap


tobiasKaminsky:
**Planned RC1: every second month in last week so that release is in first week of month / replace with date**
**Planned Release date:**

- [ ] Bump version code in ```build.gradle``` file
- [ ] release new library if needed
- [ ] release new SSO if needed
- [ ] Verify that the new version number is shown in settings
- [ ] Verify that logging is enabled
- [ ] add $version.txt to /fastlane/metadata/android/en-US/changelogs/ to support GPlay version info
- [ ] release with fastlane: creates GH release, uploads to Gplay, creates Gitlab MR for F-Droid
- [ ] Verify that it is shown on [F-Droid](https://f-droid.org/de/packages/com.nextcloud.client/) (within 24h)
- [ ] inform marketing about changelog and expected release date

**Final Release, after one weeks time (max) of latest RC release**
- [ ] Bump version code in ```build.gradle``` file
- [ ] release new library if needed
- [ ] Verify that the new version number is shown in settings
- [ ] Verify that logging is disabled
- [ ] Test that what's new is shown correctly on upgrade
- [ ] add $version.txt to /fastlane/metadata/android/en-US/changelogs/ to support GPlay version info
- [ ] release with fastlane: creates GH release, uploads to Gplay, creates Gitlab MR for F-Droid
- [ ] Verify that it is shown on [F-Droid](https://f-droid.org/de/packages/com.nextcloud.client/) (within 24h)

Backports:
- [App](https://github.com/nextcloud/android/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+label%3Abackport-request)
- [Library](https://github.com/nextcloud/android-library/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+label%3Abackport-request)

joshtrichards:
Closing since this has no work tracking utility as-is and can still be used as a template while closed if desired.

tobiasKaminsky:
Re-open as this is a template for your .1 releases

PR #2767, opened 2018-06-28
Delete Branch stable-3.1.x

AndyScherzinger:
@tobiasKaminsky since we released 3.2.0 I think we can delete this branch, right (housekeeping FTW) :smiley:


tobiasKaminsky:
Let us add this as a point to the release roadmap and then we can also directly delete the branch at https://github.com/nextcloud/android/branches/all?utf8=%E2%9C%93&query=stable

AndyScherzinger:
I updated the release template issue :+1:

tobiasKaminsky:
I already did this in https://github.com/nextcloud/android/issues/150, which is the place/time to do it, I think.
At minor releases (https://github.com/nextcloud/android/issues/2287) it is not needed (we do not have a bugfix branch).

AndyScherzinger:
Right :+1:

PR #2773, opened 2018-06-29
Correct menu highlighting

tobiasKaminsky:
Now every combination (as far as I tested) should work, e.g.
- on device
- settings
- on device

Please note that this does not work
- favorites
- settings
- back
-> it shows "all files" as also the "search" vanished.
So the menu highlighting is correct, but the behaviour is not ;-)
This is true for all other "search"-related menus.


AndyScherzinger:
:+1:

mario:
👍

PR #2774, opened 2018-06-29
Escape $summary correctly

tobiasKaminsky:


tobiasKaminsky:
Merging as is it only internal change for findbgus.

PR #2775, opened 2018-06-29
Avatar generation does not need storageManager

tobiasKaminsky:
Found via google play console:
- have many files downloaded (so account removal takes a while)
- remove account
- see app sometimes crashing as storageManager is null

--> we do not need it at all, therefore removing it


AndyScherzinger:
:+1:

mario:
👍

PR #2777, opened 2018-06-29
Add clickable link to notification

tobiasKaminsky:
Needed by @nickvergessen.
Subject is clickable (indicated with ↗, when link is available).
@AndyScherzinger do you have an idea, how we can make the clickable area a bit bigger without having the line span bigger?

@nextcloud/designers please have a look.


tobiasKaminsky:
> Why not just make the list items clickable and open the link in case there is one, else do nothing?

In a following PR there will be buttons, if available, e.g. for accept/decline remote sharing.
I want to do them beneath message and above timestamp, but this is also not fixed and I am open to any ideas :-)

AndyScherzinger:
Ah, okay.... Well then the only thing I can think of would be what you mentioned like spanning the TextView wider, maybe make use of padding.

nickvergessen:
Can't you make it fully clickable and just have a different behaviour when you click the "to be added" button

tobiasKaminsky:
> Can't you make it fully clickable and just have a different behaviour when you click the "to be added" button

This would be possible, but if you then have a link and a button and click a bit wrong on the button, then the link will open. Which will be surprising for the user ;-)

AndyScherzinger:
:+1: fine by me in general :)

jancborchardt:
Not sure if it's possible on Android, but on the web you can increase clickability without whitespace by using `padding: 10px; margin: -10px;` for example.

Also, shouldn't the first line / subject be bold? That might also help make clear clickability.

AndyScherzinger:
@jancborchardt  @tobiasKaminsky I enlarged the clickable area and also made the subject bold :)
device-2018-07-03-163220

nickvergessen:
Shouldn't all subjects be bold, not only the clickable ones?
That's how it's handled on the server

AndyScherzinger:
> Shouldn't all subjects be bold, not only the clickable ones?
That's how it's handled on the server

Kinda.... on the server the subject lines use a grey text color hightlighting parts of the subjects using black text color, I am not sure if we get the subject as a string or a rich text in order to adapt to the server's coloring logic.

tobiasKaminsky:
We have subjectRich, so we can do specific theming:
`You received {share} as a remote share from {user}`

As this takes (maybe) a bit more back and forth, I would like to do this in a separate PR.

AndyScherzinger:
I completely agree @tobiasKaminsky :+1:

tobiasKaminsky:
Rebased & ready for review (design wise except of subject formatting, please see #2794)

src/main/java/com/owncloud/android/ui/adapter/NotificationListAdapter.java
```diff
@@ -120,10 +136,10 @@ public int getItemCount() {
 
         private NotificationViewHolder(View itemView) {
             super(itemView);
-            activityIcon = (ImageView) itemView.findViewById(R.id.activity_icon);
-            subject = (TextView) itemView.findViewById(R.id.activity_subject);
-            message = (TextView) itemView.findViewById(R.id.activity_message);
-            dateTime = (TextView) itemView.findViewById(R.id.activity_datetime);
+            activityIcon = itemView.findViewById(R.id.activity_icon);
```
mario: butterknife.

AndyScherzinger: DONE: https://github.com/nextcloud/android/pull/2777/commits/df5838dd416704b1cefbcf096ccd3864fe8da205

mario:
👍

PR #2782, opened 2018-07-02
Remove m prefix for OCupload

tobiasKaminsky:
This is just removing m prefix.
I did this as a separate PR as this is no changing any code and makes the other one a lot easier to review.


AndyScherzinger:
:+1: Nice, I just commented in the other PR that we should do this but in a separate PR and here it is already :tada:

mario:
👍

PR #2789, opened 2018-07-03
mlistener might be null?

tobiasKaminsky:
Found via google play console

- all other mListener calls have a null check, except the last one, which then crashed
- I cannot reproduce a null mListener (which is the calling activity), even after suspend/resume, the calling activity is there
- used lambda
- use requireActivity, so that we do not get a NPE, but an IllegalArgumentException


AndyScherzinger:
:+1: fine by me, the drone builds fails since it doesn't find the lint wrapper script!

tobiasKaminsky:
> drone builds fails since it doesn't find the lint wrapper script!

Whatever this was, I rebased and it is now working.

mario:
👍

PR #2791, opened 2018-07-03
Image defined in density-independent drawable folder

tobiasKaminsky:
Lint:
>The res/drawable folder is intended for density-independent graphics such as shapes defined in XML. For bitmaps, move it to drawable-mdpi and consider providing higher and lower resolution versions in drawable-ldpi, drawable-hdpi and drawable-xhdpi. If the icon really is density independent (for example a solid color) you can place it in drawable-nodpi.
More info: http://developer.android.com/guide/practices/screens_support.html

--> I checked both usages on a high-end device and it looks good to me:
spinner is shown somewhat small, so density is good enough
checker: is repeated and thus also not blurry


AndyScherzinger:
:+1:

mario:
👍

PR #2794, opened 2018-07-04
Notifications correct formatting

tobiasKaminsky:
image

image

To Change:
- on subject rich objects: only "special" is bold
- on regular subjects: nothing is bold
- text is regular, with 0.57 opacity

@nextcloud/designers @nickvergessen is this correct?


AndyScherzinger:
👍 code wise Fine with me while I Wolldecke prefer to not work with opacitity but User the corresponding grey color. That way we will have a lot lese work in we eher start working on supporting a dark theme :)

nickvergessen:
The dark theme is exactly why the server uses opacity. You just replace the black font with white and the opacity will make it a grey tone again

The time is opactiy 0.5, the message is 0.57 I just checked our source.

tobiasKaminsky:
> The time is opactiy 0.5, the message is 0.57 I just checked our source.

Ha, I was lazy and thought both are the same ;-)
I'll adjust this.

tobiasKaminsky:
@AndyScherzinger what do we want to do with opacity?
DateTime is using textColorSecondary, which is a hardcoded value.

Do you have any plans for supporting dark theme? Or do you mean a dark color provided by theming app?

AndyScherzinger:
@nickvergessen is right, completely missed that part...
Let's keep the opacity solution :+1:

jancborchardt:
Good stuff! :) And yes, opacity is always better than grey value, because it also works on other backgrounds and better for dark theme. Right now we don’t support that, but it will be much easier to do that then. ;)

Just a question: What’s that long timestamp in the notification 2–4 there? It’s just test text, right? Cause we shouldn’t show a timestamp there as it duplicates the relative date display.

nickvergessen:
That should be the message of the notification, i guess he used the time to make the message longer for testing

tobiasKaminsky:
Correct, this is just the message. 
Previously I had multiple "test" messages, but then I got confused, if and when I created the notification, so I switched to creating test notifications with timestamp.

AndyScherzinger:
Anythin except moving the values to xml to get this PR merged (+ one missing review of course)

mario:
Butterknife in NotificationViewHolder. Thanks.

AndyScherzinger:
> Butterknife in NotificationViewHolder. Thanks.

@mario Rebased so codacy is happy and Butterknife is in place :)

AndyScherzinger:
pinging @mario for final review :) :heart:

mario:
👍

PR #2795, opened 2018-07-04
ThemeUtils: do not crash if context is null

tobiasKaminsky:
If context is null, then https://github.com/nextcloud/android/blob/b4b1790ef675db20cf436d510fa24e61c81d5eec/src/main/java/com/owncloud/android/authentication/AccountUtils.java#L99
will throw an IllegalArgumentException.

Context may be null due to starting/ending app on first login, though this is not consistently reproducible.


AndyScherzinger:
:+1:

mario:
👍

PR #2797, opened 2018-07-04
3.2.1 RC1

tobiasKaminsky:
As discussed with @AndyScherzinger, as there are many backports, first a small RC1 for beta testing on gplay.


AndyScherzinger:
:+1: Yeah! Go for it :) cc @mario for second approval

PR #2801, opened 2018-07-05
Harden extra text subject

tobiasKaminsky:
Found via Google play console:
extra text / extra subject can be null.


AndyScherzinger:
:+1:

mario:
👍

PR #2802, opened 2018-07-05
intent may be null on service restart

tobiasKaminsky:
> intent - The Intent supplied to Context.startService, as given. This may be null if the service is being restarted after its process has gone away, and it had previously returned anything except START_STICKY_COMPATIBILITY.

--> I am unsure about the START_STICKY_COMPATIBILITY thing.


AndyScherzinger:
:+1:

mario:
👍

PR #2804, opened 2018-07-05
Show maintenance info

### LINKED ISSUE #989: Enhance maintenance recognition
With #502 we have a recognition of maintenance mode for
- create, rename, move, delete of files/folders
- upload new files

But you can still browse the file list and refresh the folder listing without any knowledge that the server is currently in maintenance mode.

@jancborchardt @eppfel please give us an hint how to show it to the user.

tobiasKaminsky:
Fix #989

We have on nearly all occasions a snackbar, when server is in maintenance mode (e.g. renaming, delete files).
But if you browse into a subfolder or go back, you do not get an info, but outdated/none folder content.

Please note that there is no automatic/in background check if server is available again.

Steps to test:
- open up app w/o server in maintenance mode
- browse
- set server in maintenance mode
- browse, see info on top
- try other operations (CRUD on files, share, remove share)

The first part (extension of Toolbar) was done by @AndyScherzinger :tada:.


AndyScherzinger:
I fixed the info box layout so that it will not break the message as early as seen in the first screenshot - https://github.com/nextcloud/android/pull/2804/commits/810570cf49737fbac7ac5894f2f04ef95d25bf81

I also reworked the info box API a bit :)

AndyScherzinger:
:+1: looks good to me :)

mario:
👍

PR #2805, opened 2018-07-05
Show offline info

tobiasKaminsky:
When on remote operation offline/no connection is detected, this will be shown on top, until the next successful refresh is done.

Needs https://github.com/nextcloud/android-library/pull/155

### DIFF
--- a/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
+++ b/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
@@ -289,13 +289,6 @@ public class RefreshFolderOperation extends RemoteOperation {
             if (result.getCode() == ResultCode.FILE_NOT_FOUND) {
                 removeLocalFolder();
             }
-            if (result.getCode() == ResultCode.MAINTENANCE_MODE) {
-                return result;
-            }
-            if (result.getCode() == ResultCode.HOST_NOT_AVAILABLE ||
-                    result.getCode() == ResultCode.NO_NETWORK_CONNECTION) {
-                return result;
-            }
             if (result.isException()) {
                 Log_OC.e(TAG, "Checked " + mAccount.name + remotePath + " : " +
                         result.getLogMessage(), result.getException());
--- a/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
+++ b/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
@@ -292,6 +292,10 @@ public class RefreshFolderOperation extends RemoteOperation {
             if (result.getCode() == ResultCode.MAINTENANCE_MODE) {
                 return result;
             }
+            if (result.getCode() == ResultCode.HOST_NOT_AVAILABLE ||
+                    result.getCode() == ResultCode.NO_NETWORK_CONNECTION) {
+                return result;
+            }
             if (result.isException()) {
                 Log_OC.e(TAG, "Checked " + mAccount.name + remotePath + " : " +
                         result.getLogMessage(), result.getException());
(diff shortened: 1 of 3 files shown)


src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -292,6 +292,10 @@ private RemoteOperationResult checkForChanges(OwnCloudClient client) {
             if (result.getCode() == ResultCode.MAINTENANCE_MODE) {
                 return result;
             }
+            if (result.getCode() == ResultCode.HOST_NOT_AVAILABLE ||
+                    result.getCode() == ResultCode.NO_NETWORK_CONNECTION) {
+                return result;
+            }
```
AndyScherzinger: Looking at the result code checked and the code blocks end I'd vote to remove the result code checks completely since in the end if
* if any result code check is true -> we return the result object (unchanged) 
* if all result code checks are false -> we return the result object (unchanged) 

:grin:

AndyScherzinger:
@tobiasKaminsky :exclamation: rebased to adapt to the more generic API for the info box (see base branch PR) and also one minor comment (arguing for removal of the result code checks)

AndyScherzinger:
:+1: fine with me except for the commented piece of code :)

src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
AndyScherzinger: Reminder for @tobiasKaminsky ;)

AndyScherzinger:
I also vote to simply merge this PR (to the given branch) and include it in the PR #2804 directly :+1:

src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
tobiasKaminsky: True, I misread the next part and thought that the 
`if (result.isException())… `
was catching the return, but it only logs it is an error.

But as we are handling this "error" correctly, it should not be logged, so I would still leave it in?

mario:
👍

src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
AndyScherzinger: @tobiasKaminsky keeping the logging part seems fine, while anything else seems to be something that can just be deleted since it does absolutely nothing ;) Please check the whole code fragment...

```
             if (result.getCode() == ResultCode.MAINTENANCE_MODE) {
                 return result;
             }
            if (result.getCode() == ResultCode.HOST_NOT_AVAILABLE ||
                    result.getCode() == ResultCode.NO_NETWORK_CONNECTION) {
                return result;
            }
            if (result.isException()) {
                 Log_OC.e(TAG, "Checked " + mAccount.name + remotePath + " : " +
                         result.getLogMessage(), result.getException());
             } else {
                 Log_OC.e(TAG, "Checked " + mAccount.name + remotePath + " : " +
                         result.getLogMessage());
             }
         }
         return result;
```
This could then simply be shortened to (at least):
```
             if (result.getCode() == ResultCode.MAINTENANCE_MODE ||
                    result.getCode() == ResultCode.HOST_NOT_AVAILABLE ||
                    result.getCode() == ResultCode.NO_NETWORK_CONNECTION) {
                return result;
            }
...
```

while I still think it could just be:
```
             if (result.isException()) {
                 Log_OC.e(TAG, "Checked " + mAccount.name + remotePath + " : " +
                         result.getLogMessage(), result.getException());
             } 
         }
         return result;
```

tobiasKaminsky: Okay, reverted.

tobiasKaminsky:
@AndyScherzinger please have a final look and then we can merge it

AndyScherzinger:
:+1:

Issue #1794, opened 2017-11-17
Hide FAB while showing snackbar


tobiasKaminsky:
- delete a file
- you will see a snackbar (deleted)
- it overlaps FAB for the the snackbar is shown

![screenshot_20171117_130434](https://user-images.githubusercontent.com/5836855/32946454-e5ba3b8c-cb97-11e7-93de-c050a31ed220.png)

przybylski:
Generally this should not be dissapearing, it should be moving up and down when snackbar. This should be automatically when using CoordinatorLayout and Snackbar. Maybe instead of hiding FAB Layout should be transformed to CoordinatorLayouts?

AndyScherzinger:
@przybylski is right,I also checked Gmail's behavior and it is as described, moving up and down with the snackbar.

tobiasKaminsky:
Right, this is indeed the better idea 👍

przybylski:
I had a look and library which the app is using is not supporting moving FAB when snackbar appears.
There is a pull request but authors doesn't seems to be interested in it: https://github.com/futuresimple/android-floating-action-button/pull/317 
i guess it might be useful to clone the library and apply the patch to avoid this bug.

AndyScherzinger:
@przybylski sounds good especially since Tobias already did that for some theming issues which weren't possible with the original source. Other possibility which might be even more time consuming though would be to switch the library for one that is still getting active support.

tobiasKaminsky:
So I updated the lib to 1.10.3 which includes the proposed PR.
@przybylski you may test it if you want, but #2039 will change this anyway, so maybe you can do a PR against #2039?

przybylski:
https://github.com/nextcloud/android/pull/2084

tobiasKaminsky:
#2084 is for #2075 and not for this one, or?

przybylski:
@tobiasKaminsky Sorry I commented on wrong thread :D

PR #2807, opened 2018-07-05
enhance folder already exists

tobiasKaminsky:
- [x] Needs: https://github.com/nextcloud/android-library/pull/156

Steps to test:
- create a folder whose name is already taken
- previously "method not allowed" as snackbar
- now: "Folder already exists"

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -208,9 +208,9 @@ dependencies {
     // dependencies for app building
     implementation 'com.android.support:multidex:1.0.3'
 //    implementation project('nextcloud-android-library')
-    genericImplementation "com.github.nextcloud:android-library:master-SNAPSHOT"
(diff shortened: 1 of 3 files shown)


AndyScherzinger:
:+1:

mario:
👎  Sorry.

Snackbar should either push the FAB up, or ideally show above FAB (then it would not be in full width)

tobiasKaminsky:
> Snackbar should either push the FAB up, or ideally show above FAB (then it would not be in full width)

This is a known problem (#1794) , which also happens on current master:
image

I can have a look into it, but this should not prevent merging this PR.

tobiasKaminsky:
Mario said on IRC, I can merge, so merging ;-)
I will have a look into this overlapping FAB problem now.

PR #2817, opened 2018-07-11
Add 3.2.1 changelog

tobiasKaminsky:
This also creates a folder structure needed by fastlane.
Fastlane then can automatically pick the changelog and uploads it to play store.
(background: this allows to tag, create signed generic/gplay version and upload gplay to play store completely automatically).

I would like to keep changelog.md and additionally have these version text files. Although it is a bit of c&p I think it is better to also have it in one central place.

This is targeting master branch, so that it will be in upcoming stable-3.3.x branch.
(I do not expect to have any 3.2.x)


AndyScherzinger:
:+1: fine by me, while the release date would now be the 11th ;)

mario:
👍

PR #2818, opened 2018-07-13
hide info_box on default

tobiasKaminsky:
Hide it initially and only show it, when needed.
Otherwise we will see a grey bar with no text on every start.

(and I have no idea, why android studio is changing the intent)


AndyScherzinger:
👍

AndyScherzinger:
pinging @mario for second review :heart:

mario:
👍

PR #2819, opened 2018-07-13
activity: Remove useless curly braces.

ardevd:
There are two ways to write lambdas that contain single statement, but one is definitely more compact and readable than the other.


tobiasKaminsky:
Fine for me for oneliner.

AndyScherzinger:
:+1: looks good

PR #2820, opened 2018-07-13
jobs: Use previously defined constants instead of hard coded values.

ardevd:
Use ACCOUNT_NAME_GLOBAL and KEY_MEDIA_FOLDERS constants where applicable.

Duplicated string literals make the process of refactoring error-prone.


AndyScherzinger:
:+1: sweet!

PR #2821, opened 2018-07-13
operations: Replace use of Vector with a unsynchronized equivalent.

ardevd:
Early classes of the Java API, such as Vector, Hashtable and StringBuffer, were synchronized to make them thread-safe. Unfortunately, synchronization has a big negative impact on performance, even when using these collections from a single thread.

It is better to use their new unsynchronized replacements. ArrayList is pretty much a drop-in replacement for Vectors and I cant see any reason to use Vector unless there is some sort of special requirement for it. ArrayList will generally give you better performance and only half the data growth when items are added to the list.


tobiasKaminsky:
:+1: Totally fine for me, as we iterate over folderAndFiles, which is also an ArrayList.

AndyScherzinger:
:+1: also fine by me

PR #2823, opened 2018-07-13
WIP: Overlapping fab with snackbar

tobiasKaminsky:
For FAB behaving correctly, we need to use a reference to CoordinatorLayout 
https://github.com/nextcloud/android/blob/3e6a782af8cc26f434bc287d7230aa15b4fe7507/src/main/res/layout/list_fragment.xml#L26-L29.

We used to have this for any Snackbar calling an activity/used in FDA:
```
DisplayUtils.showSnackMessage(this,
                            ErrorMessageAdapter.getErrorCauseMessage(result, operation, getResources()));
```

Now we would have to check if OCFileListFragment is available and if so, use the CoordinatorLayout, else use the previous way.

Or do you have a nicer approach @mario, @AndyScherzinger?

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -2109,9 +2109,13 @@ public class FileDisplayActivity extends HookActivity
             refreshListOfFilesFragment(false);
         } else {
             try {
-                DisplayUtils.showSnackMessage(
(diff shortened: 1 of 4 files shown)


AndyScherzinger:
Not knowing of a nicer, quick approach. We'd have to rewrite layouts in order to noch have to go this way which isn't worth the effort at this moment I think.

tobiasKaminsky:
Closing this for now, as is it too much effort right now, with little benefit…

PR #2825, opened 2018-07-13
Add push values to setup.xml

### LINKED ISSUE #2824: please downgrade googleLibraryVersion to 9.6.* to fix MicroG support
https://github.com/nextcloud/android/blob/dd6f7c824e7c41ca78ff14e6b475220f9a9cb04e/build.gradle#L39

See https://github.com/nextcloud/talk-android/commit/e724cd045a09d21893a6fcb74818fb148b32578b

tobiasKaminsky:
Fix #2824 

As discussed in https://github.com/nextcloud/android/pull/2618, this changes the way we provide our push values:
- they are now in gplay/setup.xml
- gplay.build is used if task contains "Gplay"

Big thanks to @mario as I did c&p from talk app.


mario:
We will have to check MicroG in more detail as changing to 9.6.1 has other issues for people using app from google play.

mario:
Looks like it was my mistake. 9.6.1 works, and we should use that.
Late again :(

tobiasKaminsky:
As discussed with @mario, the 9.6.x series has its own problems, which produce multiple notifications, so switching back to the newest version.

AndyScherzinger:
:+1:

PR #2852, opened 2018-07-27
further codacy improvements

AndyScherzinger:
* codacy: Avoid using redundant field initializer
* codacy: Useless parentheses
* codacy: Avoid using implementation types; use the interface instead
* codacy: Avoid unused private fields + organize imports

Fixed after the codacy changes
* findbugs: use log instead of print stack trace
* findbugs: rename IDE generated parameter names

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,7 +14,7 @@ buildscript {
         google()
     }
     dependencies {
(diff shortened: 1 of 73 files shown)


mario:
👍

tobiasKaminsky:
:+1:

PR #2860, opened 2018-08-01
Fix findbugs

AndyScherzinger:
* replace int based for-loops
* literal string comparisons passing the literal as an argument
* fix trims a String temporarily
* use isEmpty() to check for empty collection instead of size <= 0
* remove unused parameters
* remove locally saving values right before returning them

### DIFF
--- a/src/main/java/com/owncloud/android/utils/ReceiversHelper.java
+++ b/src/main/java/com/owncloud/android/utils/ReceiversHelper.java
@@ -63,7 +63,7 @@ public class ReceiversHelper {
         BroadcastReceiver broadcastReceiver = new BroadcastReceiver() {
             @Override
             public void onReceive(Context context, Intent intent) {
-                if (intent.getAction() != null && intent.getAction().equals(Intent.ACTION_POWER_CONNECTED)) {
+                if (intent.getAction() != null && Intent.ACTION_POWER_CONNECTED.equals(intent.getAction())) {
                     FilesSyncHelper.restartJobsIfNeeded();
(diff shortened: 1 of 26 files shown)


AndyScherzinger:
pinging @ardevd @tobiasKaminsky @mario for a code review :) :heart:

mario:
👍

src/main/java/com/owncloud/android/utils/ReceiversHelper.java
```diff
@@ -63,7 +63,7 @@ public static void registerPowerChangeReceiver() {
         BroadcastReceiver broadcastReceiver = new BroadcastReceiver() {
             @Override
             public void onReceive(Context context, Intent intent) {
-                if (intent.getAction() != null && intent.getAction().equals(Intent.ACTION_POWER_CONNECTED)) {
+                if (intent.getAction() != null && Intent.ACTION_POWER_CONNECTED.equals(intent.getAction())) {
```
tobiasKaminsky: first check can be ommitted, or?

tobiasKaminsky:
Very nice :tada:

src/main/java/com/owncloud/android/utils/ReceiversHelper.java
AndyScherzinger: right, fixed via https://github.com/nextcloud/android/pull/2860/commits/d736519c9862f30ceb7f4f938f3ed598de638d2f waiting for the checks to turn green :)

PR #2872, opened 2018-08-11
Fix npe during files sync

tobiasKaminsky:
While debugging S9+ I found that the FilesSyncJob sometimes fails, if the date cannot be extracted from Exif (file not found, unparseable, …).
This then leads to killing the whole job, so only those files get uploaded until the crash is reached.


AndyScherzinger:
:+1: codacy to ignore imho

tobiasKaminsky:
Right, while we should try to reduce the methode length, this is out of scope for this PR.

PR #2873, opened 2018-08-11
ADD: [instantupload] setting to also upload existing files

koying:
Please be kind, first time I put my nose into the code.

This adds a setting to the instant uploads to also upload files already present on the device at the time instant upload is setup.
Ref: https://help.nextcloud.com/t/why-doesnt-android-client-upload-existing-pictures/10365

95% of the PR is for handling of the setting itself.
It might be worth wondering whether this deserves a setting at all, really, ie whether *not* uploading existing files is worthwhile, as it's a bit counter-intuitive, imo

### DIFF
--- a/src/main/java/com/owncloud/android/datamodel/SyncedFolder.java
+++ b/src/main/java/com/owncloud/android/datamodel/SyncedFolder.java
@@ -42,6 +42,7 @@ public class SyncedFolder implements Serializable, Cloneable {
     private String remotePath;
     private Boolean wifiOnly;
     private Boolean chargingOnly;
+    private Boolean existing;
     private Boolean subfolderByDate;
     private String account;
     private Integer uploadAction;
@@ -62,12 +63,13 @@ public class SyncedFolder implements Serializable, Cloneable {
      * @param type            the type of the folder
      */
     public SyncedFolder(String localPath, String remotePath, Boolean wifiOnly, Boolean chargingOnly,
-                        Boolean subfolderByDate, String account, Integer uploadAction, Boolean enabled,
-                        MediaFolderType type) {
+                        Boolean existing, Boolean subfolderByDate, String account, Integer uploadAction,
+                        Boolean enabled,MediaFolderType type) {
         this.localPath = localPath;
         this.remotePath = remotePath;
         this.wifiOnly = wifiOnly;
         this.chargingOnly = chargingOnly;
+        this.existing = existing;
         this.subfolderByDate = subfolderByDate;
         this.account = account;
         this.uploadAction = uploadAction;
--- a/src/main/java/com/owncloud/android/datamodel/SyncedFolderDisplayItem.java
+++ b/src/main/java/com/owncloud/android/datamodel/SyncedFolderDisplayItem.java
@@ -45,6 +45,7 @@ public class SyncedFolderDisplayItem extends SyncedFolder {
      * @param remotePath      remote path
      * @param wifiOnly        upload on wifi only flag
      * @param chargingOnly    upload on charging only
+     * @param existing        also upload existing
      * @param subfolderByDate create sub-folders by date (month)
      * @param account         the account owning the synced folder
      * @param uploadAction    the action to be done after the upload
@@ -55,19 +56,19 @@ public class SyncedFolderDisplayItem extends SyncedFolder {
      * @param type            the type of the folder
      */
     public SyncedFolderDisplayItem(long id, String localPath, String remotePath, Boolean wifiOnly, Boolean chargingOnly,
-                                   Boolean subfolderByDate, String account, Integer uploadAction, Boolean enabled,
-                                   List<String> filePaths, String folderName, long numberOfFiles, MediaFolderType type)
+                                   Boolean existing, Boolean subfolderByDate, String account, Integer uploadAction,
+                                   Boolean enabled, List<String> filePaths, String folderName, long numberOfFiles, MediaFolderType type)
     {
-        super(id, localPath, remotePath, wifiOnly, chargingOnly, subfolderByDate, account, uploadAction, enabled, type);
+        super(id, localPath, remotePath, wifiOnly, chargingOnly, existing, subfolderByDate, account, uploadAction, enabled, type);
         this.filePaths = filePaths;
         this.folderName = folderName;
         this.numberOfFiles = numberOfFiles;
     }
 
     public SyncedFolderDisplayItem(long id, String localPath, String remotePath, Boolean wifiOnly, Boolean chargingOnly,
-                                   Boolean subfolderByDate, String account, Integer uploadAction, Boolean enabled,
-                                   String folderName, MediaFolderType type) {
-        super(id, localPath, remotePath, wifiOnly, chargingOnly, subfolderByDate, account, uploadAction, enabled, type);
+                                   Boolean existing, Boolean subfolderByDate, String account, Integer uploadAction,
+                                   Boolean enabled, String folderName, MediaFolderType type) {
+        super(id, localPath, remotePath, wifiOnly, chargingOnly, existing, subfolderByDate, account, uploadAction, enabled, type);
         this.folderName = folderName;
     }
 }
--- a/src/main/java/com/owncloud/android/datamodel/SyncedFolderProvider.java
+++ b/src/main/java/com/owncloud/android/datamodel/SyncedFolderProvider.java
@@ -339,6 +339,8 @@ public class SyncedFolderProvider extends Observable {
                     ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_WIFI_ONLY)) == 1;
             Boolean chargingOnly = cursor.getInt(cursor.getColumnIndex(
                     ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_CHARGING_ONLY)) == 1;
+            Boolean existing = cursor.getInt(cursor.getColumnIndex(
+                    ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_EXISTING)) == 1;
             Boolean subfolderByDate = cursor.getInt(cursor.getColumnIndex(
                     ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_SUBFOLDER_BY_DATE)) == 1;
             String accountName = cursor.getString(cursor.getColumnIndex(
@@ -350,8 +352,8 @@ public class SyncedFolderProvider extends Observable {
             MediaFolderType type = MediaFolderType.getById(cursor.getInt(cursor.getColumnIndex(
                     ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_TYPE)));
 
-            syncedFolder = new SyncedFolder(id, localPath, remotePath, wifiOnly, chargingOnly, subfolderByDate,
-                    accountName, uploadAction, enabled, type);
+            syncedFolder = new SyncedFolder(id, localPath, remotePath, wifiOnly, chargingOnly, existing,
+                subfolderByDate,accountName, uploadAction, enabled, type);
         }
         return syncedFolder;
     }
@@ -369,6 +371,7 @@ public class SyncedFolderProvider extends Observable {
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_REMOTE_PATH, syncedFolder.getRemotePath());
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_WIFI_ONLY, syncedFolder.getWifiOnly());
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_CHARGING_ONLY, syncedFolder.getChargingOnly());
+        cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_EXISTING, syncedFolder.getExisting());
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_ENABLED, syncedFolder.isEnabled());
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_SUBFOLDER_BY_DATE, syncedFolder.getSubfolderByDate());
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_ACCOUNT, syncedFolder.getAccount());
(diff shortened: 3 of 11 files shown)


koying:
I'm not totally sure server-side duplicate checking falls into the scope of already existing files on a single device.
The duplicate scenario might also happen in case of multiple devices syncing to the same server...

AndyScherzinger:
> The duplicate scenario might also happen in case of multiple devices syncing to the same server...

That's true!

jancborchardt:
> From my point of view it should be a it-should-just-work meaning the user shouldn't have to expplicitely state that existing images should be uploaded. This is a bit more complex though since the question remains rather about when / in which cases this upload should be triggered. If you activate it for the very first time I'd say it should just upload the existing ones without asking while the question is what should happen when the auto upload has been activated/deactivated in the past and gets activated again because in this case it definitely should check against the server as-in a one-way sync.

@AndyScherzinger totally agree with that UX-wise! :)

And @koying welcome, and thank you for your contribution! :)

javaisbetterthanpython:
> From my point of view it should be a it-should-just-work meaning the user shouldn't have to expplicitely state that existing images should be uploaded.

I agree that it should be automatic, but still think the user should be able to opt-out. Just having the setting enabled by default on addition of a new directory to autoUpload could be fine, right?

> Since we don't have file hashes provided by the server we could only check for files with the same name and append a (1) at the end or maybe check for some other meta information present like timestamps :/

@AndyScherzinger I think the approach you suggested could work alright. Check for the same filename, and then compare timestamps. Although, I think instead of appending "(1)" to the filename on upload it might be better to append the timestamp of the upload, so you end up with something like "originalfilename-YYYYMMDD-HHMMSS.ext". This would account for multiple duplicates being merged into one directory and easily distinguish which version is which, rather than (1), (2), (3), etc.

How does the app currently handle uploads when a filename conflict occurs?

@koying Nice job man! This has been a much requested feature!

mario:
Just to clarify on something. @AndyScherzinger said:

" we could only check for files with the same name and append a (1)"

We don't need to do this since there's already a way to do it. When creating an upload job, it should set overwrite to no and (1) and such would get appended automatically.

mario:
I will go over the code today.

AndyScherzinger:
> I will go over the code today.

Thanks a lot for finding the time to look into this matter.

> @koying Nice job man! This has been a much requested feature!

Also a big thank you and kudos to your work on this feature, this highly appreciated, great job! :1st_place_medal: 

> How does the app currently handle uploads when a filename conflict occurs?

Like @mario said if you set override to "no" it'll take care of the renaming automatically.

mario:
First of all, thank you for the contribution.

Ok, so last parameter to uploadFileWithOverwrite could be changed to “false” if it’s an initial sync just to make sure we don’t step on anybodys’ toes in this “simple” scenario, even though in general auto upload would overwrite everything which is why I’ve always hesitated going with this approach. (This is in FilesSyncJob)

The issue there is you don’t really know if it’s an initial upload while at that stage.

Anyway…

In order to accomplish an initial sync (on FIRST enable state of a particular folder) you don’t need an “existing” column. 

You would need to look into FilesSyncHelper, remove the “dryRun” condition, and instead do what the usual syncing would do (the “else”) part. Keep in mind if the syncedFolderInitiatedKey for this folder is null, you’d still need to initialize it.

If a folder is disabled, you’d remove the syncedFolderInitiatedKey so the next time user runs you could once again ask if they want to upload all files.

Hope I've helped at least a little bit!

koying:
@mario Thanks for reviewing. Synthetically:

1) You suggest to remove the parameter and replace it wit a dialog box
2) When first enabled, a folder sync will push existing files, no matter what, with uploadFileWithOverwrite  = false
3) On subsequent disabling/enabling of the folder, the dialog would be shown

Did I get you right?

The problem I see is how to determine it is an initial sync.
As the syncedFolderInitiatedKey is removed on disable, we cannot use that

OTOH, to follow the same logic of syncing "everything", I don't think removing syncedFolderInitiatedKey  is desirable.
Keeping it across disable/enable both tells us if it's an initial sync and, if not, what new files we have to sync, so we wouldn't even need to ask the user.

mario:
Well, you could ask the user "Do you want to push existing files" once first enabled if you want as well.

As for the syncedFolderInitiatedKey - you could indeed opt NOT to remove it.

mario:
You just need to make sure you update it at the right time. :)

tobiasKaminsky:
Which NC15 there will be (hopefully) checksums on server side.
So starting with NC15 we then can make sure to upload only files that are not on server yet.
Do we want to postpone this until then?

bk1285:
> So starting with NC15 we then can make sure to upload only files that are not on server yet.
Do we want to postpone this until then? @tobiasKaminsky 

Some user feeback on this: Using nextcloud together with your awesome android app, this PR would resolve fix the most crucial pain point within the app for me! Please make it work and merge asap.... :D

geirgp:
Some additional user feedback

For the last 1.5 day's I've been troubleshooting my fresh nextcloud setup, trying figure out whye the android auto uploader was not working - only to find this thread and discover it is a _feature_ not to upload existing files. Pleeeease prioritize this one.

tmolteno:
This is a wonderful pull request. I too have spent hours debugging this. Is there any timeframe for a merge? Perhaps a forked app is required (I am baffled by anyone thinking that this is a 'feature')

AndyScherzinger:
> This is a wonderful pull request. I too have spent hours debugging this. Is there any timeframe for a merge?
No, see https://github.com/nextcloud/android/pull/2873#issuecomment-428574710 - to be sure to have some kind of "sync" implementation to not create duplicates on the server or override differing files with the same file name we need to have checksums (server doesn't support this yet).

> I am baffled by anyone thinking that this is a 'feature'

Uploading existing files to the server is something that hasn't been implemented yet so of course it is a feature. A functionality not present and not mandatory to be compliant with any kind of specification/standard is of course a feature and not a bug. Just becaouse you expect this to be happening doesn't make it a bug...

While I get your frustration, this is something that needs to be more or less bullet proof since such a change in behavior will hit >250K installations. So possibly creating duplicates on the server (in case you activate/deactivate the auto upload for a single folder several times) will lead to the exact opposite of what you @tmolteno are trying to achieve. It won't make them happy - it'll make them angry and frustrated.

Just my 2 cents.

javaisbetterthanpython:
In the meantime, couldn't we implement the feature to only function on a target directory that is empty? That would probably be sufficient for most users in the interim while we wait for checksums

AndyScherzinger:
> In the meantime, couldn't we implement the feature to only function on a target directory that is empty? That would probably be sufficient for most users in the interim while we wait for checksums

@javaisbetterthanpython that would imho work fine and wouldn't lead to any trouble for the app users.

what do you think @koying @tobiasKaminsky @mario @ardevd ?

koying:
You mean only upload all the very first time instantupload is used?
Seems like a corner case to me...

The typical use case (at least the one I created this PR for) is to add a local folder to the list to be synced, and naturally wanting the existing local files to be uploaded as well. 

As I said above, the conflict/overwrite issue already exists when syncing multiple devices, and must be resolved at FilesSyncJob level, globally, not at the level of choosing which local files to sync. 

Implementing this one (w/o the gui option, if so preferred) adds a case, but:
1) no data is lost due to file versioning
2) when checksum and conflict resolution is implemented at FilesSyncJob level, it will fix this case as well

koying:
Maybe an option would be to make "forceOverwrite" a tri-state, and not allow overwrites nor renames on initial sync?
Not perfect, but I guess will fit the 80/20 rule ;)

AndyScherzinger:
@koying w/o having tested it what would happen in case of a "regular" media folder like the camera if I active/deactivate/activate? Would this simply create duplicates for all files from the camera folder?

koying:
As it is now, It would re-upload everything, with overwrite.

ASDFGamer:
I also think that this is a wonderful pull request, because it improves a part of the app that is often used directly after the app was first installed.
Nextcloud 15 is now released so to my understanding the problem with the hashes on the server could be solved.

jancborchardt:
@koying could you rebase this pull request onto master, and also change it to not be a setting, but a section which shows up top of the Instant uploads view, asking if you want to upload existing images?

That’s my feedback UX-wise. But the technical implementation feedback I leave to the @nextcloud/android experts. :)

flokli:
@tobiasKaminsky can we retrieve checksums of remote files via the [`{http://owncloud.org/ns}checksums` property](https://docs.nextcloud.com/server/14/developer_manual/client_apis/WebDAV/index.html)?
Or is https://github.com/nextcloud/server/issues/11138#issuecomment-429727128 about sometimes not knowing these checksums if the files are stored on external storages, and were not uploaded via NextCloud itself?

I guess a simple collision prevention could be:

- A) if file at the destination path already exists
   - AA) check remote hash, if it's the same as the local one, consider it uploaded (and delete locally, if configured like that)
   - AB) If the remote hash is different, skip uploading (and don't delete locally)
 - B) if it doesn't already exist, upload (and delete locally, if configured like that)

Even if we don't have server side checksums, we could already add upload support for non-overwriting instant uploads (B), which would already work on [empty InstantUpload directories](https://github.com/nextcloud/android/pull/2873#issuecomment-437957653).

droide02:
> APK file: https://www.kaminsky.me/nc-dev/android-artifacts/10831.apk
> 
> qrcode
> 
> To test this change/fix you can simply download above APK file and install and test it in parallel to your existing Nextcloud app.

With this version, App crash already when I activate an Automated transfer...

OnePlus 6 on Android 9 and security patch August 2019

Excel1:
https://www.kaminsky.me/nc-dev/android-artifacts/10872.apk - tested --> ERROR
Nextcloud QA
Samsung Galaxy Note 8
Android 9

Steps to reproduce:
1. start app
2. automatic uploads
3. open 3-Points Menu for "Camera" to choose autoupload
4. App restarts instantly, no menu shown

App didnt created error log (or i didnt know where)

ArisuOngaku:
I added a PR to this as the dev seems to be inactive on this PR

https://github.com/koying/android/pull/1

This fixes the immediate issues that prevented the feature from working on my tests.

AndyScherzinger:
@ArisuOngaku thanks for the fixes, I cherry picked them to @koying's fork and to make things easier in the process pushed the branch to this repo (hope you are fine with this @koying), so I'll be closing this in favor of a PR from a branch within the repo which makes it easier to work on it. 

Successor PR is #4788 - so closing this one any further discussion/work should happen in #4788

koying:
@AndyScherzinger Sure, no prob.
I'm a bit at loss regarding the status of this PR, tbh.
I had the feelings from initial comments that it was basically rejected, waiting for some server feature, but it seems to be alive ;)

AndyScherzinger:
> @AndyScherzinger Sure, no prob.

❤️ 

> I'm a bit at loss regarding the status of this PR, tbh.

I took care of keeping it up-to-date with the app's master branch and @ArisuOngaku fixed some issues.

> I had the feelings from initial comments that it was basically rejected, waiting for some server feature, but it seems to be alive ;)

Yes and no. In general a solution for uploading existing files is to be achieved (one way or the other). As discussed here and in other places - yes, it is unclear how this should be achieved. I agreee that for a "proper" solution we need checksums from the server (which would be the 100% bulletproof way) but as also stated from @jospoortvliet to have "some"/80%-solution would be a good first step so I personally think thins can be the basis for such a change while it might need some more love.

From my point of view we still do need an agreement within the project how a 80% sultion should look like (which is the issue that block any further progress on this at the moment) - while this is just my personal view on the matter.

Still I try to keep this PR and your efforts alive since I don't want to let it go to waste!

dgtlmoon:
Hi all! thanks for the great contribution

```
For the last 1.5 day's I've been troubleshooting my fresh nextcloud setup, trying figure out whye the android auto uploader was not working - only to find this thread and discover it is a feature not to upload existing files. Pleeeease prioritize this one.```

I've also spent a lot of time scratching my head trying to figure out what's going on, yes please! please merge this is in, this is one of the general functionalities a 'cloud' type sync app should perform that everyone assumes (it's what all major connectors do already), I believe the advantage of having this in the main branch far outway any issues that fall out from this.


Quoting @AndyScherzinger in #issuecomment-412458129

```
From my point of view it should be a it-should-just-work meaning the user shouldn't have to expplicitely state that existing images should be uploaded.```

yes absolutely - I just had to enable this option for about 26 various folders on my device, but not a show stopper, It certainly is what I believe to be "expected behaviour" of this nextcloud application on the whole.

I tested the most recent APK in #issuecomment-549124650 and it does indeed work, some minor things that can get resolved after this merge

- Would be nice to know how many files are in the upload queue/how many have been uploaded, I have several hundred (maybe thousands) of images, perhaps "Uploading.... $x remaining"

image

- What's the logic for triggering a sync? mine seems to stop for a sometime and then start again, but overall seems to work.

dgtlmoon:
By the way, can anyone clarify what happens when a file cannot be uploaded in the sync, for whatever reason? I saw one said 'upload failed', is it retried? does it go to the bottom the queue to be tried again?

ArisuOngaku:
Hey @dgtlmoon this PR was moved to a fresh new one in order for the discussion to have a better chance of succeeding. Don't worry, it's not abandoned. See #4788

Excel1:
> APK file: https://www.kaminsky.me/nc-dev/android-artifacts/11516.apk
> 
> qrcode
> 
> To test this change/fix you can simply download above APK file and install and test it in parallel to your existing Nextcloud app.

This APK didnt work for me. App directly restart after clicking the cloud icon (autoupload). Note 8 - Android 9

NFDWADM:
@dgtlmoon 

Could you please link to that apk? I just spent two days installing owncloud only to find that this isn't possible atm and that you can't push folders but only single files (which would have been an acceptable one time workaround).

dekiesel:
@dgtlmoon 

Could you please link to that apk? Or is that the last compiled apk in this thread, i.e.:
https://www.kaminsky.me/nc-dev/android-artifacts/11516.apk 

I just spent two days installing nextcloud only to find that this isn't possible atm and that you can't push folders but only single files (which would have been an acceptable one time workaround).

AndyScherzinger:
@dekiesel best is to just install 3.10.0 RC1 via the GPlay beta programm

dekiesel:
@AndyScherzinger

Thanks for the tip! 

To give a little feedback:

On Android 9 on a custom rom: uploading smaller folders 35 files, 9mb, everything worked fine.

T app freezes on bigger folders (51 files, 50mb).

I sent a feedback via mail too that should include a logfile.

AndyScherzinger:
@dekiesel glad to hear that 🎉

So for future issues regarding auto upload, best thing is to open an issue with the steps plus log and pinging @tobiasKaminsky and @ArisuOngaku

PR #2881, opened 2018-08-13
Simple signup

tobiasKaminsky:
- splits up in What's new screen (for upgrades) and first run
- allows to create account on provider


src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -446,7 +446,9 @@ private void selectNavigationItem(final MenuItem menuItem) {
                 UserInfoActivity.openAccountRemovalConfirmationDialog(getAccount(), getFragmentManager(), true);
                 break;
             case R.id.drawer_menu_account_add:
-                createAccount(false);
+                Intent firstRunIntent = new Intent(getApplicationContext(), FirstRunActivity.class);
+                firstRunIntent.putExtra(FirstRunActivity.EXTRA_ALLOW_CLOSE, true);
+                startActivity(firstRunIntent);
```
AndyScherzinger: Why do you want to run the "FirstRunActivity" every time a user wants to add another account? I am fine with having this as a very first run whenever the app is installed (not updated) but even during testing this annoyed my since I always have to click login to get to the real screen since all of my accounts are on the same server. And even if not, why would I want to see the wizard again and not just type in the URL?

AndyScherzinger:
:+1: @tobiasKaminsky all good and working except for one comment, see https://github.com/nextcloud/android/pull/2881#discussion_r209927079 - I strongly vote against showing this screen every time a user adds an account. :pray:

src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
tobiasKaminsky: So you propose to show this only when creating the very first account?
But what would be if you want to create a second account on a different provider?
That is why I chose to show it every time, but I am up for better ideas :-)

cc @jancborchardt

AndyScherzinger: Also interested in @jancborchardt's opinion UX wise since it leads to one additional click every time someone wants to add another account :)

AndyScherzinger:
@tobiasKaminsky I did a bit of code cleanup: https://github.com/nextcloud/android/pull/2881/commits/c6b25745b7cf1d540a380a84658db1f0111dda3c :smiley: Hope you are fine with that

src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
jancborchardt: Good point @AndyScherzinger, but I would say we should indeed always show this first run activity. There are lots of cases which lead you to adding another account:
- One of them is having another account already set up
- Another is wanting to create a completely new one
- Another is wanting to create one and then being made aware of the ability to set up your own server.
- Another is just to see what happens when you tap it, to find out about what the feature means.

Especially as adding another account will not happen that often, the additional click in one of  the cases is not that bad. Does that sound ok?

AndyScherzinger:
@jancborchardt all fine for me :+1:

src/main/java/com/owncloud/android/ui/fragment/FeatureFragment.java
```diff
@@ -0,0 +1,119 @@
...
+
+        if (!shouldContentCentered) {
+            textView.setGravity(Gravity.START);
+        } else {
+            textView.setGravity(Gravity.CENTER_HORIZONTAL);
+        }
+
+        return textView;
+    }
+}
```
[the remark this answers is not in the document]
AndyScherzinger: fixed via https://github.com/nextcloud/android/pull/2881/commits/0c5b0202aa63c9a5d1540dd0b1354598e9469f08

src/main/java/com/owncloud/android/ui/activity/FirstRunActivity.java
```diff
@@ -0,0 +1,221 @@
...
+
+/**
+ * Activity displaying general feature after a fresh install.
+ */
+public class FirstRunActivity extends BaseActivity implements ViewPager.OnPageChangeListener {
+
+    public static final String EXTRA_ALLOW_CLOSE = "ALLOW_CLOSE";
+    public static final int FIRST_RUN_RESULT_CODE = 199;
+
+    private ProgressIndicator mProgress;
```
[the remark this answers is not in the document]
AndyScherzinger: fixed via https://github.com/nextcloud/android/pull/2881/commits/0c5b0202aa63c9a5d1540dd0b1354598e9469f08

AndyScherzinger:
also cc @ardevd for feedback if this is fine behavior-wise

AndyScherzinger:
To get 3.3.0 out the door I am fine with merging this as-is while I'd still hope we can find a better solution post-3.3.0 like showing the login screen directly (after the initial account has been created) and having the provider selection as an option at the bottom of the login screen.

What do you think @mario? I am not really happy with the actual state but consider it a trade-off to get 3.3.0 ready for an RC even though there are still some more PRs I'd like to see in 3.3.0 :/

AndyScherzinger:
also fine as-is hoping for us to find a better solution later

PR #2882, opened 2018-08-13
Backport of #2881

tobiasKaminsky:


AndyScherzinger:
So we can't start with RCs before the conference but work on a bug fix release? 0_o

PR #2884, opened 2018-08-13
Opens file in new task

### LINKED ISSUE #2867: Open files in new task
### Actual behaviour
Resuming Nextcloud opens into the last app used if a file was opened from Nextcloud.

### Expected behaviour
When opening a file, in my opinion the app used to open the file should be opened in its own task so that when resuming Nextcloud later on, you actually launch into the Nextcloud app itself instead of opening the last used app, which can be pretty confusing.
 
### Steps to reproduce
1. Open Nextcloud from the app drawer
2. Open a keepass .kdbx file using Keepass and unlock the database
3. Lock the database and return to the home screen (not using the back button to navigate back into the Nextcloud app)
4. Open Nextcloud from the app drawer to be taken to the Keepass login dialog

### Environment data
Android version: 8.1

Stock or customized system:

Nextcloud app version: 3.2.1

Nextcloud server version: 13.0.5

tobiasKaminsky:
Fix #2867 

### DIFF
--- a/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
+++ b/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
@@ -283,6 +283,8 @@ public class FileOperationsHelper {
                                     }
                                 }
 
+                                openFileWithIntent.setFlags(openFileWithIntent.getFlags() |
+                                        Intent.FLAG_ACTIVITY_NEW_TASK);
                                 mFileActivity.startActivity(
                                         Intent.createChooser(
                                                 openFileWithIntent,


AndyScherzinger:
@tobiasKaminsky @linucksrox @nextcloud/designers I am not sure this is a nice behavior. 

While it fixes #2867 it also leads to other issues as a side effect, for example:
* open a pdf file
* chosen a pdf view and launched, hit the back button
* you get back to the Nc app :+1: 
* check the running apps tab: pdf view is still there
* click on the pdf viewer within the list of opened apps: pdf viewer closes and you're back on the home screen :-1: 

So while I understand that this works as intended for keepass files I am not sure it is what users might expect opining several, different types of files ending up with all the used apps within the running apps list. It is not too annoying so I am fine with whatever decision is made.

AndyScherzinger:
> And yes, if you open up an app from recent apps and close it, the default is to go to home screen (and not the last used app).

> So I am even more fine with this approach :-)

My point is like mentioned in the scenario, the app list says "running" (for the pdf viewer) but clicking on the pdf viewer in the app list, open/closes it immidiately.

tobiasKaminsky:
Hm. I guess this is depending on the pdf app:
Drive PDF Viewer behaves a bit weird (opens up app, but without content) where "MuPDF viewer" is always showing the file, even when NC is closed and you go back to the app with "recent apps".

It seems not to be consistent: "files" of android is behaving different on both pdf apps:
"drive pdf viewer" is not opened in a new task
"muPDF viewer" is opened in a new task (-> have it as a separate app in "recent apps"), but crashing if trying to access it again

Amaze is opening it in a separate task.

AndyScherzinger:
Adobe Acrobat is opening it in a separate task, navigation back brings you back to the Nc app, choosing the Adobe Acrobat app from the running apps list opens Acrobat and closes it right away within milliseconds.

That is why I argued that the UX is kind of bad in this scenario.

tobiasKaminsky:
> Adobe Acrobat is opening it in a separate task, navigation back brings you back to the Nc app, choosing the Adobe Acrobat app from the running apps list opens Acrobat and closes it right away within milliseconds.

Can you try opening a pdf with adobe acrobat with amaze/file explorer/any other file browsing app?

AndyScherzinger:
> Can you try opening a pdf with adobe acrobat with amaze/file explorer/any other file browsing app?

Amaze treats it like this PR does, well even worse since the Amaze app doesn't come up on back navigation... So fine by me to do it like this PR changes it behavior-wise

AndyScherzinger:
:+1:

juliushaertl:
:+1: I didn't test, but it makes sense from a UX point of view.

AndyScherzinger:
@mario isn't available for reviews atm, thus merging with only one approval

PR #2887, opened 2018-08-14
show disk usage if no quota is set

### LINKED ISSUE #2876: Database Usage Status is not same in ios and android version.
simulator screen shot - iphone 6 - 2018-08-12 at 01 49 51
screenshot_20180812-014811

We can see 2 screenshots about database usage status.
I hope android version is same as ios version.

tobiasKaminsky:
Fix #2876 

If you have unlimited space / no quota set, you will not see an info.

This is what I developed now, when quota is not set:


AndyScherzinger:
Look fine to me while the server still shows a progress value, I guess in comparison to the total disk space so we might want to show that kind of progress since it adds the valuable information if you are running out of disk space.

What do you think @tobiasKaminsky @jancborchardt ?

tobiasKaminsky:
Good point, I think we should stay as close as possible to the server and that is just using their values.



image

AndyScherzinger:
Sweet :)

As a side note totally up for discussion should we just also use the color coding we use for the limited quota version to also visually (progress + color) show the criticality of this info? Meaning "primary"/yellow/red depending on how close you are to the total usage value?

AndyScherzinger:
That way the code would be even simpler by just using actual/total and actual/quota for the progress bar and anything else within the logic is exactly the same? (while not shown the "...of xxGB used" for unlimited)

AndyScherzinger:
> I assume that we are doing this, as this is based on "relative", which is an info we get from server.

Yes we do: https://github.com/nextcloud/android/pull/2887/files#diff-82c7c73602977065d97dd6984ad49d94R813

> Shall I flood my file system until I reach 98%? ;-)

You absolutely should, yes :grin:  

As for
> For coloring:
> < 90% : grey or NC blue if no theming

I'd vote to just change this since just looking at the server (Nc13 at least), the progress uses the primary/theme color while we use grey or Nc blue-ish and should just do it like the server does.

@juliushaertl can you tell us which color is used for the progress bar showing the disk/quota usage?

AndyScherzinger:
> I'd vote to just change this since just looking at the server (Nc13 at least), the progress uses the primary/theme color while we use grey or Nc blue-ish and should just do it like the server does.

@tobiasKaminsky I'd be fine with this being done in a separate PR since this PR is already an improvement over the existing feature :)

AndyScherzinger:
:+1:

juliushaertl:
> @juliushaertl can you tell us which color is used for the progress bar showing the disk/quota usage?

We should use the primary theming color with a fallback to `#555555` (which is the default color fallback for to bright colors)

tobiasKaminsky:
@juliushaertl is this for any progress value the same color? Or is this changing to e.g. red of > 95%?

juliushaertl:
@tobiasKaminsky For >80% we use `#e9322d`. The empty space of the quota bar has a light-grey background of `#ededed`

tobiasKaminsky:
Thanks @juliushaertl 

@AndyScherzinger, maybe you have an idea how to tint the background to grey? Seems that this is done automatically…

AndyScherzinger:
> @AndyScherzinger, maybe you have an idea how to tint the background to grey? Seems that this is done automatically…

Sure ;) grey is the default color of the progress background in Android. So in order to change something like this, we would need to replace the drawable, see for example `.setProgressDrawable` (https://stackoverflow.com/questions/2020882/how-to-change-progress-bars-progress-color-in-android). I kind of did this with the indeterminate drawable in our toolbar already :)

So you'd need to do something like this:
```
<?xml version="1.0" encoding="utf-8"?>
<layer-list xmlns:android="http://schemas.android.com/apk/res/android" >

    <item android:id="@android:id/background">
        <shape>
            <solid android:color="@color/yourLovelyGreyColor" />

            <corners android:radius="0dp" />
        </shape>
    </item>
    <item android:id="@android:id/progress">

        <scale
            android:drawable="@drawable/OurLovelyPrimaryColor"
            android:scaleWidth="100%" />
    </item>
</layer-list>
```
and 
```
<?xml version="1.0" encoding="utf-8"?>
<shape xmlns:android="http://schemas.android.com/apk/res/android" >

    <corners
        android:radius="0dp"
        android:topLeftRadius="0dp"
        android:topRightRadius="0dp" />

    <solid android:color="@color/colorPrimary" />

</shape>
```

AndyScherzinger:
> @AndyScherzinger, maybe you have an idea how to tint the background to grey? Seems that this is done automatically…

@tobiasKaminsky Sorry I misread that, I will have a look :)

AndyScherzinger:
background of the progress bar is now grey (Android default) for Android 5+, below we still rely on tinting which tints the whole thing.

AndyScherzinger:
It also uses the default Android color (no code added) which is grey200 and a bit darker since the webUI grey tone is too light for a mobile device (not enough contrast)

AndyScherzinger:
Mario isn't available for reviews atm, thus merging with only one approval

PR #2892, opened 2018-08-15
Use capability to check if activity app is enabled

tobiasKaminsky:
This uses capability to check if activity app is enabled and if not remove it from drawer.

Only problem is, that right after adding an new account, we do not have capability in place and thus we do not know if activity app is enabled.
So what to do? Show the menu entry and maybe later hide it?
(on Trashbin this is the same problem, there I chose to hide it first, and then show it if it is supported).

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -346,7 +346,8 @@ public abstract class DrawerActivity extends ToolbarActivity implements DisplayU
     private void filterDrawerMenu(Menu menu, Account account) {
         DrawerMenuUtil.filterForBottomToolbarMenuItems(menu, getResources());
         DrawerMenuUtil.filterSearchMenuItems(menu, account, getResources());
-        DrawerMenuUtil.filterTrashbinMenuItems(menu, account, getContentResolver());
+        DrawerMenuUtil.filterTrashbinMenuItem(menu, account, getContentResolver());
+        DrawerMenuUtil.filterActivityMenuItem(menu, account, getContentResolver());
 
         DrawerMenuUtil.setupHomeMenuItem(menu, getResources());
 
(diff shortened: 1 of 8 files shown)


AndyScherzinger:
Hiding it first and displaying it when we can detect if it is available sounds fine to me:
* consistent with trashbin behavior
* better hiding it first then to have to tell the user it is not available after the user clicked on the menu item.

src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -346,7 +346,8 @@ public void run() {
     private void filterDrawerMenu(Menu menu, Account account) {
         DrawerMenuUtil.filterForBottomToolbarMenuItems(menu, getResources());
         DrawerMenuUtil.filterSearchMenuItems(menu, account, getResources());
-        DrawerMenuUtil.filterTrashbinMenuItems(menu, account, getContentResolver());
+        DrawerMenuUtil.filterTrashbinMenuItem(menu, account, getContentResolver());
+        DrawerMenuUtil.filterActivityMenuItem(menu, account, getContentResolver());
```
AndyScherzinger:
```
if (account != null) {
            FileDataStorageManager storageManager = new FileDataStorageManager(account, contentResolver);
            OCCapability capability = storageManager.getCapability(account.name);
```

is present is both filter methods so the most efficient way would be to have the account check here and provide the filter methods with the capability or capability+account respectively. That way we would save on creating the storage manager twice and retrieving the capabilities twice. I am fine tough with keeping it this way since the actual code is easier to read which also has its benefits and I don't know much about the costs for creating the `FileDataStorageManager` and retrieving the `OCCapability `.

So I'll leave this up to you :)

tobiasKaminsky: It was mainly not refactored, because I thought we need / find another way.

tobiasKaminsky:
As activity app is pre-installed, I would rather keep it enabled and then have it disabled by capabilities than the other way (that new user do not see it directly).

In the end it does not harm if you click on "activity" if the app is disabled, it just gives an error as toast message.

AndyScherzinger:
> As activity app is pre-installed, I would rather keep it enabled and then have it disabled by capabilities than the other way (that new user do not see it directly).

> In the end it does not harm if you click on "activity" if the app is disabled, it just gives an error as toast message.

:+1: fine with me too, sure!

mario:
👍

PR #2901, opened 2018-08-20
ui:activity: Check if OCFileListFragment is null before calling onItemClicked().

ardevd:
A reference to null should never be dereferenced/accessed


tobiasKaminsky:
:+1: 
If file list fragment is, for any reason, null, we cannot do anything. And I cannot imagine any meaningful Snackbar, so it is ok to fail silently here.

tobiasKaminsky:
Rebased onto latest master to get rid of two warnings (wrongly translated "…" in latvia)

tobiasKaminsky:
@ardevd we do not use github reviews, but instead use pull approve, so it is not needed to select users for review ;-)

AndyScherzinger:
:+1:

PR #2915, opened 2018-08-23
Fix the select all button in the file picker

tilosp:
Before the select all button also selected folders.


tobiasKaminsky:
:+1: Thanks for the finding and the really quick fix :tada:

PR #2918, opened 2018-08-23
Fix the removal of "select all" from the folder picker

tilosp:
Before the wrong id for the menu entry was used


AndyScherzinger:
:+1:

mario:
👍

PR #2928, opened 2018-08-27
Check library hook

tobiasKaminsky:
image

--> will fail if library is not set to master


tobiasKaminsky:
image

--> will fail if library is not set to master

AndyScherzinger:
:+1:

AndyScherzinger:
Due to Mario not being available for reviews at the moment, merging with just one approval

PR #2945, opened 2018-08-30
Use generic nextcloud user agent on simple signup

tobiasKaminsky:
Previously we used the weblogin user agent, which e.g. is "Nexus 4". Th brand/device user agent should only be used for weblogin flow, so that user can distinguish the auth tokens.


AndyScherzinger:
:+1:

PR #2954, opened 2018-09-03
Use hashed/salted token

tobiasKaminsky:
@David-Development I created this as a different PR, based on #1116, so that this is easier to review.

As discussed, this generates a new token on each AuthToken call, returns the token in cleartext and stores it hashed in its own database for more security.
On each call, the 3rd party app has to send the cleartext token and files app will hash and compare it.


tobiasKaminsky:
@David-Development are you fine with merging this into the base branch?
And then we can merge base branch (sso) into master?

PR #2969, opened 2018-09-05
Do not block on check for dev version

tobiasKaminsky:
Previously we/I used a blocking call, so that UI hung until network call was finished.

This affects only manually installed dev version, as fdroid does not have this switch enabled.


AndyScherzinger:
:+1: looks good to me :)

PR #2971, opened 2018-09-05
Library upgrades

AndyScherzinger:
pushing each lib update individually to see the CI builds


AndyScherzinger:
@tobiasKaminsky please see the changed findbugs results which is due to the upgraded findbugs libs (shifting warnings between categories and finding some new things). Ready to review :)

tobiasKaminsky:
:+1: 

I am fine with findbugs, though our long term goal is to get rid of every warning (or suppress it).

Once drone is green, we should merge it, so that we have a very long time on dev version to see if everything is working.

AndyScherzinger:
> Once drone is green, we should merge it, so that we have a very long time on dev version to see if everything is working.

Done, merged, Drone is green :tada:

Issue #2526, opened 2018-04-28
Re-add large picture preview when uploading


BlackDex:
### Actual behaviour
- When uploading new files (pictures in this case) there is no option to have big tiles without a filename.

### Expected behaviour
- Have an option to show larger image thumb/preview when uploading. I know this was available before and i really liked that. Now i can only show small previews with the filename and date etc.. besides it.
 
### Steps to reproduce
1. Choose a folder
2. Click the "+" sign at the bottom right
3. Select "Upload Files"
4. Browse to your camera folder (In my case DCIM)
5. Enable grid view using the 3dot menu top-right.

There is no option to view large images two side-by-side for instance. This was available before and very nice to preview the images. Now they are very small and sometimes not good to see when there are small details in the pictures.

### Environment data
Android version: 8.1.0 and 7.0

Device model: Nexus 5X and Galaxy S6 (SM-G920F)

Stock or customized system: Stock

Nextcloud app version: 20180425

Nextcloud server version: Owncloud 9.1.6 (I know i need to switch & upgrade ;) )

#### Web server error log
```
n/a
```

#### Nextcloud log (data/nextcloud.log)
```
n/a
```
**NOTE:** Be super sure to remove sensitive data like passwords, note that everybody can look here! You can use the Issue Template application to prefill some of the required information: https://apps.nextcloud.com/apps/issuetemplate

AndyScherzinger:
Cc @tobiasKaminsky I can reproduce this. While you can pinch to zoom you'll basically end up with a 2 column list style grid for manual upload file picker instead of the typical grid layout used in the main remote files view.

AndyScherzinger:
solved via #3008 to be shipped with 3.4.0

PR #2983, opened 2018-09-07
File picker icon size

NilsGriebner:
Hey guys,

this solves #2526.

I added a "Large Preview" section to upload options menu.

image

On click you end up with a two column grid with larger thumbnails

image

Let me know what you think.


AndyScherzinger:
@NilsGriebner Hi Nils, thanks for working on this! I am not sure if this is the right approach. You can already turn on 2 columns by simply doing a pinch-to-zoom in the grid layout. The issue I still see there is that the grid layout isn't properly using the grid_item layouts when you pinch-to-zoom. So the "right" fix would be to make sure when doing a pinch-to-zoom that the layout-type list/grid is preserved and then you should be done.

NilsGriebner:
Hi @AndyScherzinger , thanks for the quick response and feedback. Haven't thought this way, yet. I will close this and open a new PR regarding your approach.

AndyScherzinger:
Thanks again for taking on this issue! I'll be offline for most of the weekend, same for @tobiasKaminsky while he'll also be back on Monday.

PR #3008, opened 2018-09-10
Use grid_item view for gridview in upload filepicker

NilsGriebner:
This is the next attempt to solve  #2526. Solution is based on @AndyScherzinger 's advice in  #2983. Let me know what you think :)

### DIFF
--- a/src/main/java/com/owncloud/android/ui/adapter/LocalFileListAdapter.java
+++ b/src/main/java/com/owncloud/android/ui/adapter/LocalFileListAdapter.java
@@ -519,6 +519,10 @@ public class LocalFileListAdapter extends RecyclerView.Adapter<RecyclerView.View
         return output;
     }
 
+    public void setGridView(boolean gridView) {
+        this.gridView = gridView;
+    }
+
     static class LocalFileListItemViewHolder extends LocalFileListGridViewHolder {
         private final TextView fileSize;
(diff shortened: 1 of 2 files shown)


AndyScherzinger:
cc @tobiasKaminsky @ardevd @mario since I likely won't find the time to review/test today, not sure about tomorrow

tobiasKaminsky:
The footer item seems to be also a GridView, but should be a LocalFileListFooterViewHolder.

@AndyScherzinger do we want to remember the state of this? I am not sure about it. I would also be fine that list view is always default and one has to manually enable grid view.

AndyScherzinger:
> do we want to remember the state of this? I am not sure about it. I would also be fine that list view is always default and one has to manually enable grid view.

I'd be fine with not remembering the state grid/list for now and see what we get in regards to user feedback and in case people want it to be remembered add that part in a later PR.

NilsGriebner:
@AndyScherzinger I can work on footer items. IMHO a separate PR is cleaner for this.

tobiasKaminsky:
@NilsGriebner a separate PR is always a good idea. 
Then we can merge this one.

Thank you very much for your contribution! :tada:

AndyScherzinger:
Thanks for the contribution @NilsGriebner :heart:

PR #3013, opened 2018-09-11
Store mnemonic as typed in

tobiasKaminsky:
Ref: https://github.com/nextcloud/android/pull/2381#issuecomment-399385582

We have to distinguish two cases:
- first setup e2e on user, then also currently the mnemonic is stored with withespaces
- setup client with existing e2e: this now stores the mnemonic as the user types it.

We cannot automagically split the words, but I guess it is ok if the user gets back what he/she typed in before.


AndyScherzinger:
:+1:

PR #3018, opened 2018-09-11
Fix Avatar fetching for Nc14 server

### LINKED ISSUE #3015: [3.3.0 RC2] missing avatar fetching
I mentioned it to @tobiasKaminsky already but haven't had the time to create an issue for it. Testing 3.3.0 on my device I realized that the avatars aren't showing up anymore at all (restarts don't change that) and afaik @tobiasKaminsky could also reproduce this with a fresh install on an emulator. So for some reason avatars don't get fetched from the server anymore.

AndyScherzinger:
Fixes #3015 - server sends a HTTP-201 (not 200) for a newly created avatar on Nc14

HTTP-201 (CREATED) is actually the correct HTTP code since the resource has just been created and is part of the body/reply (in case no other client has requested the avatar with the requested solution)

### DIFF
--- a/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -870,7 +870,8 @@ public class ThumbnailsCacheManager {
                     // we are using eTag to download a new avatar only if it changed
                     switch (status) {
                         case HttpStatus.SC_OK:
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@tobiasKaminsky I vote for a backport request since this is a really bad UI/UX issue since without this fix you wont see anything on the client in case it is a fresh and clean installation.

tobiasKaminsky:
:+1:

PR #3024, opened 2018-09-12
Backport #3018: Server sends a HTTP-201 (not 200) for a newly created avatar on Nc14

tobiasKaminsky:
HTTP-201 (CREATED) is actually the correct HTTP code since the resource has just been created and is part of the body/reply


AndyScherzinger:
:+1:

PR #3028, opened 2018-09-12
Delete proguard

tobiasKaminsky:
As discussed in https://github.com/nextcloud/android-library/issues/160 proguard.cfg is not correctly working.
As we do not use it, this should be removed.


AndyScherzinger:
:+1:

PR #3031, opened 2018-09-12
Fix upload filepicker footer view

NilsGriebner:
@tobiasKaminsky This fixes upload filepicker footer in gridview as discussed in #3008. Maybe there is as a smarter solution, for example integrate it into "ExtendedListFragment" to avoid dublicated code ?! Let me know what you think.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/fragment/LocalFileListFragment.java
+++ b/src/main/java/com/owncloud/android/ui/fragment/LocalFileListFragment.java
@@ -24,6 +24,8 @@ import android.app.Activity;
 import android.os.Bundle;
 import android.os.Environment;
 import android.support.annotation.NonNull;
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
:+1: 
@tobiasKaminsky what do you think? @NilsGriebner is correct that this is now duplicated code, see https://github.com/nextcloud/android/blob/master/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java#L1157 while we can't move it to `ExtendedListFragment` since the span logic does only make sense if the list/grid has a footer and having a footer depends on the used adapter implementation for the recycler view so you can't put it in the parent class since a child might use an adapter that doesn't use a footer item.

So I'd be fine with the duplicated code at the moment.

tobiasKaminsky:
:+1: 
Thank you for your contribution!

PR #3033, opened 2018-09-13
Quickfix for unknown SharingTypes

tobiasKaminsky:
If there is an unknown shareType, e.g. rooms, circles, … the app crashed before.
Now it simply does not show the share.

I am working on a real fix to add those missing ShareTypes, but this is something for next version.


AndyScherzinger:
:+1:

PR #3044, opened 2018-09-14
Correct handle write protected folders

tobiasKaminsky:
Test:
- share a folder with another user
- disable "can write"

I chose to disable the button instead of hiding it, so that it is more visible to the user that it is on purpose.

 


AndyScherzinger:
:+1: looks good to me

PR #3050, opened 2018-09-18
Fix CanvasTooLargeException

### LINKED ISSUE #3049: [3.3.0RC3] crashing when viewing a picture
### Actual behaviour
Crash when I try to view the image (I wanted to share it but never get a chance)

### Expected behaviour
Please don't crash ;-)
 
### Steps to reproduce
1. took picture
2. go to the folder view and try to view it
3. booooom

### Environment data
Android version: 8.0.0
plus 'Samsung Experience 9.0'

Device model: Samsung Galaxy S7
Stock, unrooted.

Nextcloud app version: 3.3.0RC3

Nextcloud server version: 14.0.0

#### Web server error log

-> note: also crashes when network is OFF. Thumbnail preview shows in grid and list view, just opening the image - BOOM.

Image in question:

https://cloud.nextcloud.com/s/4B7JJTP8ciwFC5T

tobiasKaminsky:
Fix #3049 

The sample factor is copied from [Android developer's guide](https://developer.android.com/topic/performance/graphics/load-bitmap), so I am unsure if we want to change this.
But for very large landscape images this is maybe necessary.

For "regular" images this should make no difference, e.g. 1024x768 other 4:3 image formats.

### DIFF
--- a/src/main/java/com/owncloud/android/utils/BitmapUtils.java
+++ b/src/main/java/com/owncloud/android/utils/BitmapUtils.java
@@ -104,8 +104,7 @@ public final class BitmapUtils {
 
             // calculates the largest inSampleSize value (for smallest sample) that is a power of 2 and keeps both
             // height and width **larger** than the requested height and width.
-            while ((halfHeight / inSampleSize) > reqHeight
-                    && (halfWidth / inSampleSize) > reqWidth) {
+            while ((halfHeight / inSampleSize) > reqHeight || (halfWidth / inSampleSize) > reqWidth) {
                 inSampleSize *= 2;
             }
         }


AndyScherzinger:
@tobiasKaminsky the original code matched the Developer guide's code except for `>`vs. `>=` so why did we need to change `&&`to `||`?

tobiasKaminsky:
The image in question was ~ 8000x1000, so it matched the desired height and therefore no scaling was needed. But this then resulted in a too large canvas.
So this change verifies that width and height are within the desired height / width.

AndyScherzinger:
I do understand the code, I am just not sure if it needs to differ from the Android dev guide as in not sure if we need the OR or if we just need the >= instead of >

tobiasKaminsky:
The image in question has 16512x2064.
The screen size of my emulator has 1920x1280.
SampleSize is 1.
halfHeight = 1032
halfWidth = 8256

so if you would use "&&" then this check would be ok, leading to sampleSize=1 and thus no changes in the scaled image size. And this then leads to a crash.
This approach in my opinion does only work with images that have regular aspect ratios (4:3, 16:3, 1:1, …), but not on panorama landscapes like the one above.

With the "||" approach we then would have a sampleSize=4, with height=1032 and 129.

Im also not sure if this is the best approach. But somehow we have to make sure that the scaled bitmap is not too large, so that it fits on the canvas.

AndyScherzinger:
:+1: agreed and fine with the code

AndyScherzinger:
@tobiasKaminsky does this have to be back-ported for 3.3.0? or 3.3.x?

AndyScherzinger:
Sounds good to me

AndyScherzinger:
Backport to be done: 3.3.1

PR #3051, opened 2018-09-18
Add room share type

tobiasKaminsky:
Adds room share type: display & allow for search/add rooms to a file / folder.

### DIFF
--- a/src/main/java/com/owncloud/android/operations/CreateShareWithShareeOperation.java
+++ b/src/main/java/com/owncloud/android/operations/CreateShareWithShareeOperation.java
@@ -55,7 +55,8 @@ public class CreateShareWithShareeOperation extends SyncOperation {
      */
     public CreateShareWithShareeOperation(String path, String shareeName, ShareType shareType, int permissions) {
         if (!ShareType.USER.equals(shareType) && !ShareType.GROUP.equals(shareType)
-                && !ShareType.FEDERATED.equals(shareType) && !ShareType.EMAIL.equals(shareType)) {
+                && !ShareType.FEDERATED.equals(shareType) && !ShareType.EMAIL.equals(shareType)
+                && !ShareType.ROOM.equals(shareType)) {
             throw new IllegalArgumentException("Illegal share type " + shareType);
         }
         mPath = path;
--- a/src/main/java/com/owncloud/android/operations/CreateShareWithShareeOperation.java
+++ b/src/main/java/com/owncloud/android/operations/CreateShareWithShareeOperation.java
@@ -1,4 +1,4 @@
-/**
+/*
  *   ownCloud Android client application
  *
  *   @author masensio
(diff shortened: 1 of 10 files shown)


src/main/java/com/owncloud/android/operations/CreateShareWithShareeOperation.java
```diff
@@ -55,7 +55,8 @@
      */
     public CreateShareWithShareeOperation(String path, String shareeName, ShareType shareType, int permissions) {
         if (!ShareType.USER.equals(shareType) && !ShareType.GROUP.equals(shareType)
-                && !ShareType.FEDERATED.equals(shareType) && !ShareType.EMAIL.equals(shareType)) {
+                && !ShareType.FEDERATED.equals(shareType) && !ShareType.EMAIL.equals(shareType)
+                && !ShareType.ROOM.equals(shareType)) {
```
AndyScherzinger: We could start to think about rather having a map with the supported share types and use `if !contains`

src/main/java/com/owncloud/android/ui/adapter/UserListAdapter.java
```diff
@@ -99,25 +100,25 @@ public void onBindViewHolder(@NonNull UserViewHolder holder, int position) {
...
+                    name = context.getString(R.string.share_group_clarification, name);
+                    setImage(holder, name, R.drawable.ic_group);
+                    break;
+                case EMAIL:
+                    name = context.getString(R.string.share_email_clarification, name);
+                    setImage(holder, name, R.drawable.ic_email);
+                    break;
+                case ROOM:
+                    name = context.getString(R.string.share_room_clarification, name);
+                    setImage(holder, name, R.drawable.ic_group);
```
AndyScherzinger: should we use a different icon for "room", like the talk icon maybe?

AndyScherzinger: Or a speech bubble?

tobiasKaminsky: We can, but this is a fallback and is replaced by circular avatar with first character in it.

image

or talk icon?

AndyScherzinger: I'd then say let's go for `char_bubble` since all other fallback icons are also filled icons

tobiasKaminsky: Done

src/main/java/com/owncloud/android/operations/CreateShareWithShareeOperation.java
AndyScherzinger: @tobiasKaminsky what do you think?

tobiasKaminsky: Done

AndyScherzinger:
:+1: thanks for the latest CR change @tobiasKaminsky jsut awaiting the build to be finished

PR #3052, opened 2018-09-18
Codacy fixes

AndyScherzinger:
* codacy: field has the same name as a method
* codacy: avoid using Literals in Conditional Statements
* codacy: avoid the use of value in annotations when its the only element
* codacy: consider simply returning the value vs storing it in local variable
* codacy: avoid unused imports
* codacy: all methods are static. Consider using a utility class instead. 
* codacy: String.indexOf(char) is faster than String.indexOf(String).
* codacy: useless qualified this usage in the same class.
* codacy: avoid using redundant field initializer
* codacy: avoid instantiating new objects inside loops
* codacy: useless parentheses
* remove unused return value from method signature


tobiasKaminsky:
:+1:

PR #3069, opened 2018-09-20
Show local size when switching to "on device" view

tobiasKaminsky:
When switching to "on device" view, you see how much space a directory consumes, making it simplier to delete big folders/files.

 

(in this case only 12mb of folder size are locally available)


AndyScherzinger:
:+1: sweet! makes total sense while I never thought about the different scenario remote/local folder :+1:

PR #3070, opened 2018-09-20
Codacy fixes

AndyScherzinger:
needs https://github.com/nextcloud/android-library/pull/180

fixing several codacy warning categories, while for the for-loop I am respecting https://developer.android.com/training/articles/perf-tips :
> With an ArrayList, a hand-written counted loop is about 3x faster (with or without JIT), but for other collections the enhanced for loop syntax will be exactly equivalent to explicit iterator usage.

* codacy: String.indexOf(char) is faster than String.indexOf(String).
* codacy: Do not add empty strings
* codacy: Avoid the use of value in annotations when its the only element
* codacy: Avoid unnecessary constructors - the compiler will generate t…
* codacy: No need to check for null before an instanceof
* codacy: String.startsWith can be rewritten using String.charAt(0)
* codacy: A getX() method which returns a boolean should be named isX()
* codacy: No need to call String.valueOf to append to a string.
* codacy: Avoid using redundant field initializer
* codacy: This for loop can be replaced by a foreach loop

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -2020,7 +2020,7 @@ public class FileDisplayActivity extends HookActivity
             snackbar.show();
         }
 
-        if (fileDetailFragment != null && fileDetailFragment instanceof FileDetailFragment) {
+        if (fileDetailFragment instanceof FileDetailFragment) {
             ((FileDetailFragment) fileDetailFragment).getFileDetailSharingFragment()
                     .onUpdateShareInformation(result, getFile());
         }
--- a/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
+++ b/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
@@ -455,7 +455,7 @@ public class FileDataStorageManager {
         cv.put(ProviderTableMeta.FILE_PUBLIC_LINK, folder.getPublicLink());
         cv.put(ProviderTableMeta.FILE_PERMISSIONS, folder.getPermissions());
         cv.put(ProviderTableMeta.FILE_REMOTE_ID, folder.getRemoteId());
-        cv.put(ProviderTableMeta.FILE_FAVORITE, folder.getIsFavorite());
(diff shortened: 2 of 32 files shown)


AndyScherzinger:
@tobiasKaminsky this PR contains the change needed for the merged lib change (isFavorite): https://github.com/nextcloud/android/pull/3070/commits/6b59ade7bde9233a840a57856e3747237b81c001 - This is a breaking change in the library

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -2020,7 +2020,7 @@ private void onUpdateShareInformation(RemoteOperationResult result, @StringRes i
             snackbar.show();
         }
 
-        if (fileDetailFragment != null && fileDetailFragment instanceof FileDetailFragment) {
+        if (fileDetailFragment instanceof FileDetailFragment) {
```
tobiasKaminsky: getSecondFragment() could return null and thus null check is needed.
We should add @nullable as return

tobiasKaminsky:
One null check should remain, otherwise :+1: 

Then let us merge it as soon as drone is green

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
AndyScherzinger: @tobiasKaminsky why would we need the nullCheck? `null instanceOf Whatever`should always be false, shouldn't it?

AndyScherzinger:
@tobiasKaminsky see https://docs.oracle.com/javase/specs/jls/se10/html/jls-15.html#jls-15.20.2

> At run time, the result of the instanceof operator is true if the value of the RelationalExpression is not null and the reference could be cast to the ReferenceType without raising a ClassCastException. Otherwise the result is false.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
AndyScherzinger: see https://github.com/nextcloud/android/pull/3070#issuecomment-424310161 for `null`it is `false`by language design :)

tobiasKaminsky:
:+1:

PR #3086, opened 2018-09-25
Remember last path which was used on local file selection

tobiasKaminsky:
- click on "+" to upload a file
- choose any folder
- upload the file
- click on "+" again
- see that you are in the same folder again
(unless it does not exist, then you are navigated up until any folder exists)


AndyScherzinger:
:+1: Nice UX improvement!

PR #3087, opened 2018-09-25
Use assemble explicitly

tobiasKaminsky:
Try to fix findbugs "no class found" error.


tobiasKaminsky:
@AndyScherzinger location of intermediate classes was changed by gradle.

I also added a check for missing lint / findbugs, so that drone will fail, which then prevents from merging.

AndyScherzinger:
@tobiasKaminsky looks good, for some reason `FindBugs (master)` is missing :(

PR #3094, opened 2018-09-26
Add editor config to configure Android Studio 

AndyScherzinger:
...according to the project guidelines automatically

IntelliJ/Android Studio support `.editorcong` files out of the box. This way AS will be configured according to out guidelines :)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,33 @@
+# .editorconfig
+
+# see http://EditorConfig.org
+
+# This is the file in the root of the project.
+# For sub folders you can have other files that override only some settings. 
+# For these, this settings should be false.
+root=true
+
+[*]
+max_line_length=120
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
Nice, shall we add linux line break (LF) as default for all files?

AndyScherzinger:
> Nice, shall we add linux line break (LF) as default for all files?

Not sure... I run Windows (for a reason) and configured Git to do auto linebreak-conversions so I'll always checkout `crlf` and push `lf`. We could add it, but I am unsure about the effects on non-posix dev environments

tobiasKaminsky:
:+1: 

I am also unsure :-)
I just tried to avoid that we have crlf in sourecode.
So let us keep it this way for now.

Maybe we find some nice things we want to add later

PR #3113, opened 2018-10-04
Changed "Copied to clipboard" text to "Link copied" 

### LINKED ISSUE #3110: Link share "Copied to clipboard" bar should say "Link copied"
Simpler wording and shorter. People don't really know what a "clipboard" is. ;) 

This shows up as a bottom toast/snackbar when you share via link.

Tijs-B:
Resolves #3110.


AndyScherzinger:
:+1:

tobiasKaminsky:
:+1: 

Thanks for your contribution :tada: 
As we use Transifex to translate our strings, it is only needed to change the wording in values/strings.xml, as the other one will automatically be overwritten by Transifex.

PR #3120, opened 2018-10-09
Show unread comments count

tobiasKaminsky:
Shows unread comments as "talk bubble".

To test
- share a file/folder with your user from user B
- as user B: comment on file/folder
- see "bubble" after refresh in app
- click on comment -> see comment
- back: bubble is removed


src/main/java/com/owncloud/android/db/ProviderMeta.java
```diff
@@ -32,7 +32,7 @@
 public class ProviderMeta {
 
     public static final String DB_NAME = "filelist";
-    public static final int DB_VERSION = 35;
+    public static final int DB_VERSION = 37;
```
AndyScherzinger: Send Note seems to be missing the 36er dB upgrade, if all upgrades are done via this PR it should put all of them into one new dB version instead of two.

src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -287,6 +309,30 @@ public void onBindViewHolder(@NonNull RecyclerView.ViewHolder holder, int positi
...
                 OCFileListItemViewHolder itemViewHolder = (OCFileListItemViewHolder) holder;
 
+                if (!TextUtils.isEmpty(file.getOwnerId()) && !mAccount.name.split("@")[0].equals(file.getOwnerId())) {
+                    try {
+                        itemViewHolder.sharedAvatar.setImageDrawable(
+                                TextDrawable.createAvatarWithUserId(file.getOwnerDisplayName(),
+                                        mContext.getResources().getDimension(R.dimen.list_item_avatar_icon_radius)));
+                        itemViewHolder.sharedAvatar.setVisibility(View.VISIBLE);
+                    } catch (NoSuchAlgorithmException e) {
+                        e.printStackTrace();
```
AndyScherzinger: Log statement

AndyScherzinger:
@nextcloud/designers should be show unread comments as "talk buble", see https://github.com/nextcloud/android/pull/3120#issue-221313657 or rather go for a default chat/note ison like http://materialdesignicons.com/icon/file-document-outline or http://materialdesignicons.com/icon/message-text for example?

src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -180,6 +180,24 @@ public void setFavoriteAttributeForItemID(String fileId, boolean favorite) {
...
 
+    public void refreshCommentsCount(String fileId) {
+        for (OCFile file : mFiles) {
+            if (file.getRemoteId().equals(fileId)) {
+                file.setUnreadCommentsCount(0);
+                break;
+            }
+        }
+
+        for (OCFile file : mFilesAll) {
```
AndyScherzinger: wouldn't `mAllFiles` contain `mFiles`so the objects would be updated?

src/main/res/drawable/ic_comment.xml
```diff
@@ -0,0 +1,9 @@
+<vector android:height="24dp"
```
AndyScherzinger: if we keep this icon should we add the corresponding license header? (not sure if the icon has been created by Nc project orr commes from a icon library.

src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
tobiasKaminsky: mFilesAll and mFiles are different vectors containing the files.
So we have to do it twice.

AndyScherzinger:
👍 fine with me

jancborchardt:
Yup, comments icon from server is good. :)

For server (at least Grid view) it's planned to put the comment icon in place of the Share icon when there's comments. Cause Comments can only come when it's shared so it's directly related, and doesn't take up more space.

tobiasKaminsky:
> comment icon in place of the Share icon

Hm. But clicking on comments should open comment/activity/version view and clicking on share should open share view. So we would loose these shortcurts…

tobiasKaminsky:
@jancborchardt has this comment icon any licence we can add to?

tobiasKaminsky:
> > comment icon in place of the Share icon
> 
> Hm. But clicking on comments should open comment/activity/version view and clicking on share should open share view. So we would loose these shortcurts…

Misread this, as you refer to grid view only.
I'll do this on our grid view also.

tobiasKaminsky:
@nextcloud/designers what do you think? Can we have something like a white border around the comments? Would this help?

src/main/java/com/owncloud/android/ui/fragment/FileDetailSharingFragment.java
```diff
@@ -217,6 +231,10 @@ public void setShareWithUserInfo() {
         }
     }
 
+    private void setShareWithYou() {
```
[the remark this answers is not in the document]
tobiasKaminsky: Hm. This is not in recent version. 
For me it also shows "outdated" on path to file: src/main/java/com/owncloud/android/ui/fragment/FileDetailSharingFragment.java Outdated

Maybe codacy just forgot to delete this old comment.

AndyScherzinger:
@tobiasKaminsky one issue with the comment is is that is it `black+50% opacity` -> `#7F000000`, while imho it should be `grey600` as in `#757575`

tobiasKaminsky:
> The comments icon looks very strange / big in the grid view.

True… Is this better, or even smaller? (changed from 24dp to 22dp):
 


> Also, why no share icon if we have comments?

https://github.com/nextcloud/android/pull/3120#issuecomment-445735679
--> to save space

tobiasKaminsky:
> @tobiasKaminsky one issue with the comment is is that is it `black+50% opacity` -> `#7F000000`, while imho it should be `grey600` as in `#757575`

Hm. 
I use black for the icon and then alpha 30%: 
https://github.com/nextcloud/android/blob/cad6cd6a19c5e5c15259b35d14ad3b7592e2a08d/src/main/res/drawable/ic_comment.xml#L7
https://github.com/nextcloud/android/blob/9e115d91d032a7346f181b558b6d02e79c286531/src/main/res/layout/grid_image.xml#L73

If I change it in xml, I get this:

Does this look yellow?!

AndyScherzinger:
@tobiasKaminsky for the outline simply add the following to the patch xml element:
```
android:fillColor="#FF000000"
android:strokeColor="#FFFFFFFF"
android:strokeWidth="0.2"
android:strokeMiterLimit="4"
```

as for the alpha thing, sure, that work too as long as the icon is not hovering over the mime type icon ;)

AndyScherzinger:
@tobiasKaminsky pushed the xml, can you try again. I can't launch this branch (app crahses right away, liekly because I already installede the syncDownload branch)

AndyScherzinger:
both ways would be fine with me

tobiasKaminsky:
In code is not working as these requires min 21.
But having two xml is also not that bad.

AndyScherzinger:
👍 works for me

PR #3129, opened 2018-10-10
Add autofill hints

tobiasKaminsky:
Can be tested with https://github.com/googlesamples/android-AutofillFramework

Username and password fields I specified for auto fill.
Other things, like folder name, I set to "no" as I cannot imagine any value by autofill.


AndyScherzinger:
:+1: Nice Addition!

PR #3130, opened 2018-10-11
Activity: use image url instead of generating previews by oneself

tobiasKaminsky:
Needs https://github.com/nextcloud/android-library/pull/73 and https://github.com/nextcloud/activity/pull/300

Previously we tried to find a referenced file in our database and then generated a thumbnail out of it.
Now we rely on the server to get a preview url. This way we can have thumbnails for files we never had in our database before.

TODO
- [x] add backwards compatibility

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activities/ActivitiesActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activities/ActivitiesActivity.java
@@ -171,7 +171,7 @@ public class ActivitiesActivity extends FileActivity implements ActivityListInte
                 PorterDuff.Mode.SRC_IN);
 
         FileDataStorageManager storageManager = new FileDataStorageManager(getAccount(), getContentResolver());
-        adapter = new ActivityListAdapter(this, this, storageManager, false);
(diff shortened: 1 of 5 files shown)


tobiasKaminsky:
Damn, I totally forgot the backward compatibility :-/

AndyScherzinger:
@tobiasKaminsky the lib change say compatible with Nc14. Is that what you meant with backwards compatible, because there is still also Nc13 maybe 12?

tobiasKaminsky:
I am using old thumbnail system on <15 and new on >=15:
https://github.com/nextcloud/android/pull/3130/files#diff-521d2e61330253bb0889bddc78c08d0dR204

It depends on activity app version, but I would like to bind it to a NC version, so that we can remove it on NC19 :-)

AndyScherzinger:
👍

mario:
👍

AndyScherzinger:
@tobiasKaminsky fancy taking a look? I am on the train right now so I can't :/

tobiasKaminsky:
What shall I do? I merged it ~45min before your comment :-)

AndyScherzinger:
Nothing all. Good it seems.

PR #3132, opened 2018-10-11
App crashes when uploading files to Nextcloud

Blackhat96:
- In Android Pie(API 28) if user tries to upload a files to nextcloud it will crash with below errors.

**Logcat error:**

Process: com.nextcloud.client, PID: 30549 java.lang.RuntimeException: Unable to start service com.owncloud.android.files.services.FileUploader@6fd73c0 with Intent { cmp=com.nextcloud.client/com.owncloud.android.files.services.FileUploader (has extras) }: java.lang.SecurityException: **Permission Denial: startForeground from pid=30549, uid=10278 requires android.permission.FOREGROUND_SERVICE**

Reason:
Apps that target Android 9 (API level 28) or higher and use foreground services must request the FOREGROUND_SERVICE permission.
https://developer.android.com/guide/components/services#Foreground

- Added **android.permission.FOREGROUND_SERVICE** permission in android manifest.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -65,6 +65,8 @@
     <uses-permission android:name="android.permission.USE_FINGERPRINT" />
     <uses-permission android:name="android.permission.REQUEST_INSTALL_PACKAGES" />
 
+    <!-- Apps that target Android 9 (API level 28) or higher and use foreground services must request the FOREGROUND_SERVICE permission -->
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
 
 
     <application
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
Nice catch @Blackhat96 👍

@tobiasKaminsky can you test this PR since I won't have a chance to do so before mid/end of next week, which seems a bit late :/

tobiasKaminsky:
:+1: 
Thank you very much for this fix :tada: 
(I only wrapped the line)

AndyScherzinger:
:+1: Thanks again for the contribution @Blackhat96 ❤️🙏🎉

Blackhat96:
@AndyScherzinger @tobiasKaminsky you're welcome🙂

PR #3133, opened 2018-10-11
fix exception handling

David-Development:
Exceptions such as NetworkOnMainThread exceptions were not passed on to the client as some error occurred while reading the exception message. (`println needs a message`)

### DIFF
--- a/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
+++ b/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
@@ -239,6 +239,10 @@ public class InputStreamBinder extends IInputStreamService.Stub {
     }
 
     private boolean validateToken(String hash, String token) {
+        if(hash.isEmpty() || !hash.contains("$")) {
+            throw new IllegalStateException(EXCEPTION_INVALID_TOKEN);
+        }
+
         String salt = hash.split("\\$")[1]; // TODO extract "$"
 
         String newHash = EncryptionUtils.generateSHA512(token, salt);
--- a/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
+++ b/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
@@ -111,7 +111,7 @@ public class InputStreamBinder extends IInputStreamService.Stub {
             NextcloudRequest request = deserializeObjectAndCloseStream(is);
             httpStream = processRequest(request);
         } catch (Exception e) {
-            Log_OC.e(TAG, e.getMessage());
+            Log_OC.e(TAG, e.toString());
             exception = e;
         }
 
--- a/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
+++ b/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
@@ -111,7 +111,7 @@ public class InputStreamBinder extends IInputStreamService.Stub {
             NextcloudRequest request = deserializeObjectAndCloseStream(is);
             httpStream = processRequest(request);
         } catch (Exception e) {
-            Log_OC.e(TAG, e.toString());
+            Log_OC.e(TAG, "Error during Nextcloud request", e);
             exception = e;
         }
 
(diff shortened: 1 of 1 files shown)


David-Development:
Bug found by @stefan-niedermann 

References:
- https://github.com/nextcloud/Android-SingleSignOn/issues/35 
- https://github.com/stefan-niedermann/nextcloud-deck/pull/1

stefan-niedermann:
cc @desperateCoder

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
```diff
@@ -111,7 +111,7 @@ public int read() {
             NextcloudRequest request = deserializeObjectAndCloseStream(is);
             httpStream = processRequest(request);
         } catch (Exception e) {
-            Log_OC.e(TAG, e.getMessage());
+            Log_OC.e(TAG, e.toString());
```
AndyScherzinger: Shouldn't it rather be `Log_OC.e(TAG, "Error during Nextcloud request", e);`?

That way you won't loose the stack trace, or will that be logged later on? If logging of the stack trace is done at a later point, please dismiss my comment. :)

David-Development: @AndyScherzinger Good point, thank you! No, it isn't logged later on. I'll fix it :)

AndyScherzinger: Thanks :) Glad my reviews are useful to you 😊

David-Development: They really are, I wasn't aware, that this will preserve the stack trace. Learned something new 👍

David-Development: I hope so too! Next year we'll meet in person at the conference :)

David-Development:
@stefan-niedermann @desperateCoder Any more issues that you guys encountered during your development? Otherwise I would recommend merging this soon.

Any objections? 

@tobiasKaminsky there is still one small issue left on codacy but I don't know how else to transfer the exception message properly as the interface requires a throwable. I mean for sure we could introduce our own throwable..

stefan-niedermann:
No further issues yet from my point of view :)

tobiasKaminsky:
I think we can ignore the codacy warning.
Is this otherwise ready to test?

David-Development:
> I think we can ignore the codacy warning.

@tobiasKaminsky Okay, sounds good to me!

Ready to test then :)

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
```diff
@@ -213,7 +215,19 @@ private InputStream processRequest(final NextcloudRequest request) throws Unsupp
...
         } else {
-            throw new IllegalStateException(EXCEPTION_HTTP_REQUEST_FAILED, new IllegalStateException(String.valueOf(status)));
+            StringBuilder total = new StringBuilder();
+            InputStream inputStream = method.getResponseBodyAsStream();
+            // If response body is available
+            if(inputStream != null) {
+                BufferedReader r = new BufferedReader(new InputStreamReader(inputStream));
+                String line = r.readLine();
+                while(line != null) {
+                    total.append(line).append('\n');
```
tobiasKaminsky: Do we want to have line breaks here? Is this working as expected?

David-Development: This code is supposed to read the response-body if a request returned a code other than 2XX (E.g. 404 / or 5XX). Sometimes the response-body contains some additional information about the problem that occurred. Therefore we're basically reading whatever the response-body contains and put it in an exception. The line-breaks are basically just passed on to the exception. Otherwise we might be breaking the formatting of the error messages.

David-Development: @tobiasKaminsky We had an issue once were the request failed and the response body contained some helpful additional information. Example: https://github.com/stefan-niedermann/nextcloud-deck/pull/1#issuecomment-429102940

tobiasKaminsky:
:+1:

AndyScherzinger:
👍

PR #3143, opened 2018-10-14
Fix IllegalArgumentException on files with dots in name, add test case

slohse:
This PR fixes a problem were dots in file names can cause the app to crash with an `IllegalArgumentException: Comparison method violates its general contract!`.

I added a test case. Since I don't want to disclose information on the person who found this bug (i.e. file names), and I wasn't able to cause the error with a nice and simple 5 item list, I generated a test case with 200 random names in the format `Year.Month.RandomNumber`.


AndyScherzinger:
:+1: Nice catch @slohse :tada: 
:1st_place_medal: for adding a testcase :heart: 

@tobiasKaminsky please have a look since this is a central piece of code.

tobiasKaminsky:
:+1: 

All tests should have sorted list, which then get shuffled on testing and then should return the same order.
I adjusted this for the new test (and an older one).

For me this is fine now.

PR #3152, opened 2018-10-16
Fixing further codacy issues

AndyScherzinger:
just some minor fixes:

* codacy: Avoid unused private fields such as 'account'.
* extract size to variable
* codacy: This for loop can be replaced by a foreach loop
* codacy: Avoid using redundant field initializer for 'onlyOnDevice'


tobiasKaminsky:
:+1:

PR #3157, opened 2018-10-17
Pom: increase httpclient version

tobiasKaminsky:
Though we do not use automationTest or pom right now, it is better to at least refer to a version without know security problems.

(was reported by Github)


AndyScherzinger:
:+1:

PR #3173, opened 2018-10-23
Fix AutoUpload notification

### LINKED ISSUE #3163: Auto Upload Notification recreation
### Actual behaviour
- Picture is taken
- Nextcloud detects the new picture
- Uploads it
- Deletes the local picture (like it should)
- DCMI folder is empty
- Nextcloud sends a notification that a new upload folder has been found (actual bug)

### Expected behaviour
- There should be no notification about the new folder becuase it is not a new folder.
 
### Steps to reproduce
- Like above in "Actual behaviour".

### Comments from me
It is like a nag screen. The good thing is though that in the app no settings are altered. It seems from the outside just like a bug with the notifications.

### Environment data
Android version: 

Device model: Huawei P9 Plus (VIE-L09) & Samsung Galaxy S8 (SM-G950)

Stock or customized system: Huawei with Smart Launcher (5.1 build 048K), Samsung 100% Stock

Nextcloud app version: 3.3.0 on both

Nextcloud server version: 14.0.1 & 14.0.3

#### Web server error log
```
Should not be relevant and the timespan is too huge to include everything.
```

```
The log is too long that I can assure that there is no sensetive data in it. From my point of view there are even no current log entrys. I will now empty the log and trigger the bug and look into the new one.

UPDATE: There is no log of any events. Not the successful upload of the picture nor the re-detection notification of the Camera folder. On both phones. The log is blank, not a single entry.
```

tobiasKaminsky:
fix #3163 

- fix not working detection of already defined synced folders
- delete not used (and wrong) function

Problem was that passing everything into selection did not worked.
Also this is error prone to sql injection (yes, very theoratical with folder names), but changing it to use correct parameter solves both.

Steps to test:
- have images in camera folder
- make sure that download folder is not in media scanner (delete images from download in gallery app)
- clean installation of app
- set up camera folder
- take image -> no notification
- delete all files in camera folder
- take image -> no notification
- download image e.g. with Firefox --> see notification


AndyScherzinger:
:+1:

AndyScherzinger:
> Problem was that passing everything into selection did not worked.

Can you give me some hints here since looking at the code it kind of looks the same to me as before except for the query change using parameters. I am pretty sure I am just overlooking something looking at the diff...

tobiasKaminsky:
I trimmed it a bit for better readability.
 
Old: 
selection: SYNCED_FOLDER_LOCAL_PATH == localPath AND SYNCED_FOLDER_ACCOUNT == account.name
arguments: /

New:
selection: SYNCED_FOLDER_LOCAL_PATH =? AND SYNCED_FOLDER_ACCOUNT =?
arguments: new String[]{localPath, account.name}

With new system we let Android construct the query by inserting the arguments to the correct position. This prevents from SQL injection and also does correct escaping.

AndyScherzinger:
Thanks, now I can see it. :pray:

PR #3180, opened 2018-10-24
Change repositories order

tobiasKaminsky:
This resolves failing drone jobs, as first google is used and only then jcenter.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,18 +7,18 @@
 
 buildscript {
     repositories {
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Just merging as this fixes all failing drone builds, but does not change any code.

AndyScherzinger:
:+1: Nice catch @tobiasKaminsky :heart:

PR #3183, opened 2018-10-24
In both if branches was the same

tobiasKaminsky:
Found via FindBug.
Previously we did this
- switched to grid on PHOTO_SEARCH
- post event on NO_SEARCH) or REGULAR_FILTER
- OR post event on everything else

now last two branches are combined


AndyScherzinger:
:+1: Nice catch!

PR #3187, opened 2018-10-25
Wrapper to start urls/intents if no app is available

tobiasKaminsky:
This checks if any app is available to handle intent, if not shows an error message.

(and changing order in "participate", so that both Fdroid are aligned together)

Found via gPlay dev console. I wonder how many people have no browser app installed…But having our app crashing is also not good…


AndyScherzinger:
:+1: Erm, okay. Also no idea what device wouldn't have any web browser installed... really strange...

PR #3089, opened 2018-09-25
Switch to AndroidX

tobiasKaminsky:
This was done automatically by AndroidStudio and "should" only be a replace of imports and names.

Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

PR #3188, opened 2018-10-25
Biometric

tobiasKaminsky:
This uses new Biometric API to replace DeviceCredentials as the current implementation has problems on some devices.

The new support library is currently alpha-02, so we should wait until stable release.
(based on #3089)


tobiasKaminsky:
Let us close this for now as the current fingerprint system is working well enough.
Once Biometric is stable and there is a need for this, we can still retry this.

PR #3189, opened 2018-10-25
Fix talk icon on first run

tobiasKaminsky:
Use another svg as source for talk icon. previous got corrupted on 6.0

Steps to test:
- run & install app on 6.0
- run & install app on 9.0


tobiasKaminsky:
Thanks for the hint & fix @jancborchardt :tada:

AndyScherzinger:
:+1:

jancborchardt:
Also cc @mario because I think the same issue exists in the notifications of the Android Talk app. Could you check again, or was that fixed already?

mario:
Can you clarify what does "got corrupted" means?

tobiasKaminsky:
This is how it looked previously on 6.0:

mario:
Ah. I guess you could have adapte useCompatVectorIfNeeded to do its changes on 23 as well instead of a new SVG. But a solution is a solution :)

@jancborchardt I think that was fixed a while ago? Please correct me if not.

tobiasKaminsky:
This function is used, but still the issue happened.
This happens even if I do not use this function. 
According to the comment in this function, this is fixing another issue: [Why isn't my vector drawable scaling as expected?](https://stackoverflow.com/questions/34936590/why-isnt-my-vector-drawable-scaling-as-expected)

>  @mario because I think the same issue exists in the notifications of the Android Talk app. Could you check again, or was that fixed already?

> @jancborchardt I think that was fixed a while ago? Please correct me if not.


any of you should check it ;-)

mario:
@tobiasKaminsky yes it still happens because it's currently made to only work on max. API 22, and 6.0 is 23 afaik :)

It's actually solving the exact this issue.

AndyScherzinger:
👍

PR #3199, opened 2018-10-29
Changelog for 3.3.1

tobiasKaminsky:
Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

PR #3200, opened 2018-10-29
List bug fixes

MorrisJobke:
See #3199

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## 3.3.1 (October 29, 2018)
-- Bug fixes
+- Bug fixes:
+  - [android#3066](https://github.com/nextcloud/android/pull/3066) Quickfix for unknown SharingTypes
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
Sweet! @tobiasKaminsky usually we only list the "text" not any references to actual PRs. We could use this and for the store releases strip away that part since we only have 500 characters for store changelogs

tobiasKaminsky:
Those individual PR description are in my opinion not needed in any release info.
I am more into linking the milestone within changelog, so that interested people can have a look.
(And yes, for next time we  (and most myself) should create a more meaningful changelog.
For 3.3.1 it could have been
Bugfix 
- fix crash on shared folder/file via Talk
- fix crash on Notification activity

MorrisJobke:
> For 3.3.1 it could have been

Yes - then also do it, because there is nothing more useless than a release with "Bug fixes" as release notes. Only "we changed something" would be worse.

tobiasKaminsky:
Are you fine with this? 
https://github.com/nextcloud/android/pull/3201

PR #3203, opened 2018-10-30
Switch font from openSans to Nunito

tobiasKaminsky:
This is the first implementation.
It is also working on 4.1.

We should compare screen by screen to check if the new font looks good.


AndyScherzinger:
1 extra lint warning: 
```
../../../src/main/res/font/nunito.xml:2: The resource R.font.nunito appears to be unused
```

AndyScherzinger:
@tobiasKaminsky Does this properly fallback for 4.0.x?

tobiasKaminsky:
Even better: on my HTC Desire Z with 4.0.4 it also shown correctly…

image

tobiasKaminsky:
Do we use somewhere bold fonts? Then we should check if this is also correctly used.

AndyScherzinger:
Yes we do, just search for `android:textStyle="bold"` examples would be file details, what's new, drawer header, user info screen

AndyScherzinger:
👍 Bold works fine for me, changes looks good. Only issue seems to be certain elements (checkbox, toggle, spinner), see https://issuetracker.google.com/issues/63250768 which has been fixed and will ship in a future lib release. Thus I think this is good to go and we ignore the font on checkboxes for know. See
device-2018-10-31-131157
device-2018-10-31-131134

Else we would have to set the font programmatically either by extending the checkbox implementation or directly in each implementation by setting it on the checkbox. I think it is not worth the effort and needs to be taken out with a later lib release.

PR can imho be merged.

tobiasKaminsky:
Thanks for testing @AndyScherzinger 
Then let us merge it.
Do you know if we can prevent the lint error? Or do we have to update warning count?

AndyScherzinger:
I'd say let's raise the lint score for now... no idea at the moment how to remove the warning...

github-tomster:
this is a code of conduct violation!
- it is not considerate: the user is **dictated** a font he **was able to change before**
- it is not respectful: the user is not given an option to change the forced setting.
instead the so called 'contributors' laughing at the users stating: 
'we won’t add a setting for something as basic as the font.'
https://github.com/nextcloud/android/issues/3348#issuecomment-448764246
- it is not collaborative because only personal favors are considered user feedback is not
https://github.com/nextcloud/android/issues/3348
- it is not pragmatic because the user was tied down by hard-coding the font it into the application
and there is nothing like an more important result.

tangible result is to provide settings to choose:
- use system font
- use built in font
- use own font -> choose from filesystem

revert it until, a proper solution is available

PR #3205, opened 2018-10-30
Add specific sort order for upload/local/trashbin

### LINKED ISSUE #2919: Sort Deleted files by "Deletion date" by default
### Actual behaviour
- Deleted files are sorted alphabetically by default, which is quite confusing.

### Expected behaviour
- Because you are likely to want to restore files which you recently deleted, they should be sorted like that.
 
_Not_ by "Newest first" but by "Recently deleted". :) Also, this should happen regardless of what you set by default on your main file view, as "Recently deleted" makes most sense for deleted files. Of course unless people change that sort option – then it should be remembered (for Deleted files only).

tobiasKaminsky:
Fix #2919 

Trashbin: new->old by default

We previously had one setting for all three above. There is no good transition in my opinion, so it starts for upload/local a->z and trashbin: new->old


AndyScherzinger:
👍

PR #3206, opened 2018-10-30
improved translation

RobinJespersen:
"Hochladen fehlgeschlagen, erneut anmeldung" is no german.


AndyScherzinger:
Hi @RobinJespersen thanks for the fix, we handle translations via transifex: https://www.transifex.com/nextcloud/nextcloud/translate/#de_DE/android/91098483

I took the liberty to fix it directly on transifex.

tobiasKaminsky:
@RobinJespersen thank you for your bug report and @AndyScherzinger thanks for fixing it :tada:

PR #3207, opened 2018-10-31
Fix fingerprint bug

tobiasKaminsky:
On certain devices no fingerprint, but only pin worked.

Fix: #3166 
Fix: #2968 
Fix: #2879 
Fix: #2763 
Fix: #3115

Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

PR #3208, opened 2018-10-31
Backport #3207: Fix fingerprint bug

tobiasKaminsky:

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -242,7 +242,9 @@
         <service android:name=".media.MediaService" />
 
         <activity android:name=".ui.activity.PassCodeActivity" />
-        <activity android:name=".ui.activity.RequestCredentialsActivity" />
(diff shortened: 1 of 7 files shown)


AndyScherzinger:
👍 looks good, lint warnings need to be raised to 97, since master is now at 97 not 94 anymore which is what breaks the builde @tobiasKaminsky

tobiasKaminsky:
Hm. This is against stable-3.3.x, so master lint should not be relevant.
I guess there is a bug in the lint script…

AndyScherzinger:
@tobiasKaminsky please also see the master targeting PR, there are two newly introduced lint warnings (unused resources) that should be fixed

tobiasKaminsky:
To be honest, I do not really care if a backport is increasing lint.

So let us merge it, if you are fine with it.

AndyScherzinger:
Also my opinion. Merged!

PR #3225, opened 2018-11-05
As async, it can happen that list already changed

tobiasKaminsky:
As action handling on notification is done async, it can happen that when clicking multiple actions at once, the list changed and thus it crashed with OutOfBounds.

When using this one by one or with a fast enough connection, this does not happen as the list internally gets refreshed.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/adapter/NotificationListAdapter.java
+++ b/src/main/java/com/owncloud/android/ui/adapter/NotificationListAdapter.java
@@ -234,8 +234,11 @@ public class NotificationListAdapter extends RecyclerView.Adapter<NotificationLi
         protected void onPostExecute(Boolean success) {
             if (success) {
                 int position = holder.getAdapterPosition();
-                notificationsList.remove(position);
-                notifyItemRemoved(position);
+
+                if (position > 0 && position < notificationsList.size()) {
+                    notificationsList.remove(position);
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/ui/adapter/NotificationListAdapter.java
```diff
@@ -234,8 +234,11 @@ protected Boolean doInBackground(Action... actions) {
         protected void onPostExecute(Boolean success) {
             if (success) {
                 int position = holder.getAdapterPosition();
-                notificationsList.remove(position);
-                notifyItemRemoved(position);
+
+                if (position > 0 && position < notificationsList.size()) {
```
AndyScherzinger: Shouldn't it be `position >= 0`since 0 is a valid index value for a list?
https://docs.oracle.com/javase/8/docs/api/java/util/List.html#get-int-
>     java.lang.IndexOutOfBoundsException - if index < 0 || index >= size()

AndyScherzinger:
Just a minor comment about the if-block

src/main/java/com/owncloud/android/ui/adapter/NotificationListAdapter.java
tobiasKaminsky: Indeed.

AndyScherzinger:
👍

PR #3226, opened 2018-11-05
UI can already be non-existent

tobiasKaminsky:
Found via gplay console.

If async takes a bit and user already closed trashbin view, showing an error could lead to a NPE, as view is already destroyed.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
+++ b/src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
@@ -279,7 +279,7 @@ public class TrashbinActivity extends FileActivity implements TrashbinActivityIn
             swipeListRefreshLayout.setRefreshing(false);
         }
 
-        if (emptyContentMessage != null) {
+        if (emptyContentMessage != null && emptyContentHeadline != null && emptyContentIcon != null) {
             emptyContentHeadline.setText(R.string.common_error);
             emptyContentIcon.setImageDrawable(getResources().getDrawable(R.drawable.ic_list_empty_error));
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
👍 NPE check is fine with me, one question though: Wouldn't the same thing apply to the `emptyContent*` layout elements right after the refresh layout part?

tobiasKaminsky:
> extra lint warning (ellipsizing on translation) has come in via master while this shouldn't increase the lint warnings but be ignored. @tobiasKaminsky it seems this lint ignore fix you introduced a while ago doesn't work anymore?

--> #3230

AndyScherzinger:
@tobiasKaminsky what is your opinion on https://github.com/nextcloud/android/pull/3226#issuecomment-435785533 ?

tobiasKaminsky:
Hm. My comment must have been went away…
You are right. I confused emptyContentMessage with a check of the container.
I'll fix it tomorrow.

AndyScherzinger:
@tobiasKaminsky friendly ping ❤️

tobiasKaminsky:
Not today :-/
I am half way on packing my luggage, as we leave tomorrow very early.
I'll take care of it after vacation (and also resolve all conflicts, promised!)

AndyScherzinger:
> Maybe we can do this better, like when "onDestroy" is called, set a boolean to false and check for this…

might be worth a try a then monitor gplay after the release of that fix. But would look a bit nicer, sure.

tobiasKaminsky:
Let us first keep it this way, as this works ;-)
I think I remember that on my MVPP branch I did it a bit different, so we can do enhance this later.

PR #3236, opened 2018-11-05
Use lombok to generate POJO boilerplate code

### LINKED ISSUE #3148: Use Lombok
Please see the video on https://projectlombok.org

It is a compile-time only library, but it helps to avoid lots of boilerplate code, e.g. you do not have to write any getter/setter. Instead you briefly annotate it with "@getter).
There are lots of other features.

AndyScherzinger:
Resolves #3148 adding lombok support for the app

### DIFF
--- a/src/androidTest/java/com/owncloud/android/datamodel/OCFileUnitTest.java
+++ b/src/androidTest/java/com/owncloud/android/datamodel/OCFileUnitTest.java
@@ -1,4 +1,4 @@
-/**
+/*
  * ownCloud Android client application
  *
  * @author David A. Velasco
@@ -77,7 +77,7 @@ public class OCFileUnitTest {
         mFile.setFileId(ID);
         mFile.setParentId(PARENT_ID);
         mFile.setStoragePath(STORAGE_PATH);
-        mFile.setMimetype(MIME_TYPE);
+        mFile.setMimeType(MIME_TYPE);
         mFile.setFileLength(FILE_LENGTH);
         mFile.setCreationTimestamp(CREATION_TIMESTAMP);
         mFile.setModificationTimestamp(MODIFICATION_TIMESTAMP);
@@ -86,8 +86,8 @@ public class OCFileUnitTest {
         mFile.setLastSyncDateForData(LAST_SYNC_DATE_FOR_DATA);
         mFile.setAvailableOffline(true);
         mFile.setEtag(ETAG);
-        mFile.setShareViaLink(true);
-        mFile.setShareWithSharee(true);
+        mFile.setSharedViaLink(true);
+        mFile.setSharedWithSharee(true);
         mFile.setPublicLink(PUBLIC_LINK);
         mFile.setPermissions(PERMISSIONS);
         mFile.setRemoteId(REMOTE_ID);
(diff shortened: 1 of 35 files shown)


AndyScherzinger:
beware that desugaring for D8 needs to be disabled due to a bug in Android gradle, to be fixed with 3.3.0, see https://github.com/rzwitserloot/lombok/issues/1882#issuecomment-429976073

tobiasKaminsky:
Shouldn't this show all getter/setter in IDE on e.g. SyncedFolderProvider, where SyncedFolder is created/used?

image

AndyScherzinger:
> Shouldn't this show all getter/setter in IDE on e.g. SyncedFolderProvider, where SyncedFolder is created/used?

Yes and it absolutely does so (at least for me). Most important point is that yo need to build the app first since lombok is doing the work during compile time. There is also a IntelliJ Plugin. Not sure if that one would show you the accessor methods even before they have been generated just by looking for annotations.

AndyScherzinger:
@tobiasKaminsky but yeah, I see an issue here, using lombok for constructors saves some code but also makes it a bit tricky since the order of the variables in the class defines the order within the constructor (!) so this might potentially break existing contructor signatures without compile errors (e.g. swapping the order of two string paramters...).

So we need to be super careful when using constructor code generation...

tobiasKaminsky:
I enabled the plugin and now it works.

> So we need to be super careful when using constructor code generation...

But the order is then consistent? Then we can still do this, I think.

AndyScherzinger:
> But the order is then consistent? Then we can still do this, I think.

Yes we can, I am just saying that we need to be careful, the following scenario might happen, e.g.
```
String a;
String b;
String c;

MyClass(String a, String, c, String b)...
```

when making use of lombok constructor generation you end up with
```
MyClass(String a, String, b, String c)...
```

So no compile errors, all good but you flipped the order of the paramters. So when removing an explicitely implemented constructor you first need to check if the order will be the same.

AndyScherzinger:
Codacy warnings to be ignored since they are basically code movements and are not newly introduced by this PR and have nothing to do with the introduction of lombok.

@tobiasKaminsky from my point of view this PR is ready for a review. I wouldn't put more code refactorings into this PR since it'll get to big then but rather do further lombok changes via separate, easy to review PRs.

src/androidTest/java/com/owncloud/android/datamodel/OCFileUnitTest.java
```diff
@@ -126,9 +126,8 @@ public void writeThenReadAsParcelable() {
         assertThat(fileReadFromParcel.getPublicLink(), is(PUBLIC_LINK));
         assertThat(fileReadFromParcel.getPermissions(), is(PERMISSIONS));
         assertThat(fileReadFromParcel.getRemoteId(), is(REMOTE_ID));
-        assertThat(fileReadFromParcel.needsUpdateThumbnail(), is(true));
+        assertThat(fileReadFromParcel.isNeedsUpdateThumbnail(), is(true));
```
tobiasKaminsky: This sounds a bit weird, but I guess it is due to auto generation.
Maybe we can change this to isThumbnailUpdateNeeded()?

tobiasKaminsky:
:+1: just one minor question/idea

Great to have less boilerplate code!

src/androidTest/java/com/owncloud/android/datamodel/OCFileUnitTest.java
AndyScherzinger: > Maybe we can change this to isThumbnailUpdateNeeded()?
True, I'll change that 👍

AndyScherzinger: DONE: https://github.com/nextcloud/android/pull/3236/commits/e46818f3fa95877afda6f26038d92bb84f810205

tobiasKaminsky:
I have problems with AS running lombok, I get the RuntimeException as described in linked issue.
This does not happen when running gradlew, so build/CI is not affected.

A local quickfix for me it so change gradle-plugin to 3.3.0-beta02.
(this then also needs gradle-4.10.x)

AndyScherzinger:
so the disabled desugering in the gradle config of the project doesn't work for you?

tobiasKaminsky:
No, same like another one in the issue said.
Hopefully there is soon 3.3.0 stable and then we can also revert the property

Issue #2188, opened 2018-02-16
dark theme


b-harper:
The issue template seem overkill for a feature request of a dark theme.  I do not see a GitHub issue for this request, but there was [one in the forums][0].

[0]: https://help.nextcloud.com/t/desktop-mobile-apps/129/77

AndyScherzinger:
@b-harper thanks for the report and yes this would be an enhancement. However I can't find a reference to it in the link you posted.

b-harper:
I linked to the wrong post within that thread.  You replied to nullnvoid's request in this [post][0]:


> > nullnvoid : Also is there any possibility of a dark theme?
> 
> That would be a new feature and yes that should be technically possible even though I am not sure how much work this would be at this time since the code base isn’t perfect for multi theme support. We might also rather look into the new feature of Android day/night theme :smiley: But again this would then also need some input from @jan and the design team on how the dark theme should look like.

[0]: https://help.nextcloud.com/t/desktop-mobile-apps/129/78

AndyScherzinger:
@mario @tobiasKaminsky @nextcloud/designers What do you think?

As mentioned yes this would be possible for newer Android versions since I'd then vote to just support a dark theme via the day/night theming capabilities of actual Android versions to not pollute the code base with switches.

tobiasKaminsky:
Do you know how this behaves with custom theming? E.g. if I have light blue or a very dark blue, how will both behave on a night/dark mode?
As for now I think the priorities are somewhere else, but if someone wants to pick this up, first discuss it with @nextcloud/designers and then feel free to implement it.

> to not pollute the code base with switches.

I fear that at least one switch is needed, which enables/disables theming on day/night.

AndyScherzinger:
> I fear that at least one switch is needed, which enables/disables theming on day/night.

No, not if we use Androids day/night system. :grin: 

> Do you know how this behaves with custom theming? E.g. if I have light blue or a very dark blue, how will both behave on a night/dark mode?

The theme won't chance much, it will basically chance the background color via the Android themes (to be written) and the Nc theming will stay the same. There are certain edge cases where this will be an issue like choosing white or black as Nc's main color where we already have checks to then jump to another color. @juliushaertl what do you think?

juliushaertl:
>  There are certain edge cases where this will be an issue like choosing white or black as Nc's main color where we already have checks to then jump to another color. 

[...]

PR #3237, opened 2018-11-05
Material design revamp

AndyScherzinger:
Move to new Android X implementation and thus upgrade to latest material design theming.

**DONE**
* [x] Use sentence case for button labels
* [x] Use primary color for status bar too (instead of primary dark color), like refresh Google apps do too
* [x] replace accent color with primary color
* [x] use new outlined style for secondary buttons and for textfields, use pill style buttons. 

**Screenshot on Android 9**
device-2018-12-28-152236

device-2018-12-28-152357

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
@@ -73,7 +73,7 @@ public class FolderPickerActivity extends FileActivity implements FileFragment.C
     public static final String EXTRA_FILES = FolderPickerActivity.class.getCanonicalName() + ".EXTRA_FILES";
     public static final String EXTRA_ACTION = FolderPickerActivity.class.getCanonicalName() + ".EXTRA_ACTION";
     public static final String EXTRA_CURRENT_FOLDER = FolderPickerActivity.class.getCanonicalName() +
-            ".EXTRA_CURRENT_FOLDER";
+        ".EXTRA_CURRENT_FOLDER";
     public static final String MOVE = "MOVE";
     public static final String COPY = "COPY";
 
@@ -158,7 +158,7 @@ public class FolderPickerActivity extends FileActivity implements FileFragment.C
     }
 
     /**
-     *  Called when the ownCloud {@link Account} associated to the Activity was just updated.
+     * Called when the ownCloud {@link Account} associated to the Activity was just updated.
      */
     @Override
     protected void onAccountSet(boolean stateWasRecovered) {
@@ -211,10 +211,10 @@ public class FolderPickerActivity extends FileActivity implements FileFragment.C
         if (listFragment != null) {
             if (!mSyncInProgress) {
                 listFragment.setMessageForEmptyList(
-                        R.string.file_list_empty_headline,
-                        R.string.file_list_empty_moving,
-                        R.drawable.ic_list_empty_create_folder,
-                        true
+                    R.string.file_list_empty_headline,
+                    R.string.file_list_empty_moving,
+                    R.drawable.ic_list_empty_create_folder,
+                    true
                 );
             } else {
                 listFragment.setEmptyListLoadingMessage();
@@ -227,7 +227,7 @@ public class FolderPickerActivity extends FileActivity implements FileFragment.C
     protected OCFileListFragment getListOfFilesFragment() {
         Fragment listOfFiles = getSupportFragmentManager().findFragmentByTag(FolderPickerActivity.TAG_LIST_OF_FOLDERS);
         if (listOfFiles != null) {
-            return (OCFileListFragment)listOfFiles;
+            return (OCFileListFragment) listOfFiles;
         }
         Log_OC.e(TAG, "Access to non existing list of files fragment!!");
         return null;
@@ -235,7 +235,7 @@ public class FolderPickerActivity extends FileActivity implements FileFragment.C
 
     /**
      * {@inheritDoc}
-     *
+     * <p>
      * Updates action bar and second fragment, if in dual pane mode.
      */
     @Override
@@ -258,7 +258,7 @@ public class FolderPickerActivity extends FileActivity implements FileFragment.C
 
         // perform folder synchronization
         RemoteOperation refreshFolderOperation = new RefreshFolderOperation(folder, currentSyncTime, false,
-                ignoreETag, getStorageManager(), getAccount(), getApplicationContext());
+            ignoreETag, getStorageManager(), getAccount(), getApplicationContext());
 
         refreshFolderOperation.execute(getAccount(), this, null, null);
         setIndeterminate(true);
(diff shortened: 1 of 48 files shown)


tobiasKaminsky:
Where is this crashing? Then i can try to fix it.
A first smoke test did not lead to major crashes.

AndyScherzinger:
@tobiasKaminsky I did an additional commit which introduced the usage of the bridge theme which prevents the crashes and allow for mixed UI elements ;)

AndyScherzinger:
pinging @nextcloud/designers for feedback on the pill-style button design.
Is this what we should be aiming for (color, fontStyle, etc.) for buttons?
device-2018-12-18-130831

jancborchardt:
@AndyScherzinger yeah, this looks awesome! :) 
- The primary button color is `#0073b3` in the screenshot, but Nextcloud blue is `#0082c9` (lighter)
- I assume that all uppercase letters on buttons are Android standard?
- Same for the slight drop shadow on the primary button, guess it’s standard?

AndyScherzinger:
>@AndyScherzinger yeah, this looks awesome! :) 
>- The primary button color is `#0073b3` in the screenshot, but Nextcloud blue is `#0082c9` (lighter)

I'll check, this is possibly either a wrong value in our config or the accent color which gets computed and set

> - I assume that all uppercase letters on buttons are Android standard?

Yes, while looking at the latest changes in Google's app, uppercase has been dropped so I'll check if we can get rid of uppercase which should look much nicer

> - Same for the slight drop shadow on the primary button, guess it’s standard?

It has been before but I'll check if it can be dropped since the latest material design guide also doesen't ship the drop shadow anymore. So we should be able to get rid of that. Will have to investigate.

tobiasKaminsky:
@AndyScherzinger many many thanks for taking care of UI :tada: :tada: :tada: tada:

AndyScherzinger:
@jancborchardt @tobiasKaminsky see https://material.io/design/components/buttons.html#contained-button -> Text labels can be written in sentence case, as long as the button is clearly distinguishable from elements around it.

So not sure if we should keep them capitalized or not. To me it seems capitalzed is easier to process for the user (personal impression)

Upper Case:
device-2018-12-20-124814

Sentence Case:
device-2018-12-20-124337

Beware that I didn't fiddle around with the font size but let Android use its standard font size for the buttons so in this case or maybe due to the font we use the button text seems a bit small when sentence cased.

AndyScherzinger:
@jancborchardt at the moment all primary buttons use the accent color, so should be make all primary buttons being colored using the primary color?

@jancborchardt @tobiasKaminsky here is a themed version from Tobi's server:
device-2018-12-20-141101

AndyScherzinger:
@nextcloud/designers @tobiasKaminsky besides the latest changes in the screenshots above regarding button styling, please also take a look (and provide some feedback) on the new styling of input text fields which would now be styled with material design's outlined box (grey unfocused, accent  color when focused)

see screenshots below from a themed server instance:
device-2018-12-20-170546
device-2018-12-20-170513

jancborchardt:
Great great stuff! :slightly_smiling_face: 

> Text labels can be written in sentence case, as long as the button is clearly distinguishable from elements around it.
> So not sure if we should keep them capitalized or not. To me it seems capitalzed is easier to process for the user (personal impression)

YEAH! Finally they changed it. Yes, please let’s use sentence case, as this is much more readable. Sentence case (regular casing of words) is easier to read because we don’t read words letter by letter, but scan them by their overall form. Writing them all uppercase basically makes them all rectangles, which is horrible for readability.


> at the moment all primary buttons use the accent color, so should be make all primary buttons being colored using the primary color?

Yes. The header of the app, the folder icon, and the buttons should all have the _same_ color. Which should be the primary color. :)


> please also take a look (and provide some feedback) on the new styling of input text fields which would now be styled with material design's outlined box (grey unfocused, accent color when focused)

Love it! :+1: The only thing that looks weird is that when it’s empty, the placeholder is shown vertically centered. Would it be possible to always show it in the first line, or is that non-standard?

AndyScherzinger:
> The only thing that looks weird is that when it’s empty, the placeholder is shown vertically centered. Would it be possible to always show it in the first line, or is that non-standard?

Will have to check that, it is basically the default behavior of a multi-line text field.

AndyScherzinger:
@rullzer @tobiasKaminsky why and when did you enable the sign off enforcer bot? Not sure if we really want to activate this for the app (will likely leads to failed status for any community contribution atm).

jancborchardt:
Looks great to me! Awesome work! :)

AndyScherzinger:
@tobiasKaminsky can you rebase and review please? :heart:

tobiasKaminsky:
Did a smoke test and it looks great so far!

Only thing I found it on adding a new account, this looks a bit strange, or?

src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
```diff
@@ -418,6 +418,10 @@ private void initControls(){
             mChooseBtn.getBackground().setColorFilter(ThemeUtils.primaryColor(this, true), PorterDuff.Mode.SRC_ATOP);
             mChooseBtn.setOnClickListener(this);
         }
+
+        if (mCancelBtn != null) {
```
tobiasKaminsky: Is this NPE check not too late? This will explode on L414 if button is null

tobiasKaminsky:
:+1: 
Code changes are also very good :+1: :tada:

tobiasKaminsky:
Rebased!

src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
AndyScherzinger: Right!

AndyScherzinger:
@jancborchardt please see the comment by @tobiasKaminsky regarding the buttons on the login screen. :heart:

AndyScherzinger:
Hi @github-tomster thanks for the feedback. Dark theme is out of scope for this PR but is in discussion in another issue. The short version: it's on our radar and will come in the future, the long version: it'll take a while to be done since it is a bit more complicated than with the mentioned app. This client (files app) already implemented dynamic theming based on the server side theme so we do calculate some colors already based on a base/primary color from the server and for a dark theme these calculations will have to work a bit different plus with the server probably having proper dark theme support in the future we will have to support that too in a proper way.

So again: it'll come but might take a while due to limited time. If you or anybody else is willing to step up and implement it, please, feel free and be sure we will support you wherever we can to get it done.

If nobody steps up I'll probably implement it but won't have the time to start working on it before February or March... Busy day job...

jancborchardt:
@AndyScherzinger @tobiasKaminsky right, the primary button needs to be a bit adjusted to work on the blue (same-colored) background. Can we use a border? Quick mockup based on the web interface log in screen:
log in and sign up with provider

@github-tomster the dark theme is tracked at #2188 and being worked on by awesome community member @dan0xii. :)

AndyScherzinger:
Sure thing... y need to have to separate style implemented here since the second button would be an outlined one then (like the new secondary action buttons) 👍

tobiasKaminsky:
:tada: cool to have this in

PR #3241, opened 2018-11-06
Further lombok usage

### LINKED ISSUE #3148: Use Lombok

AndyScherzinger:
follow-up PR of #3236

Resolves #3148


tobiasKaminsky:
:+1:

PR #3247, opened 2018-11-07
Remove not needed class

tobiasKaminsky:
This class is not used anywhere, so we can remove it.


AndyScherzinger:
:+1:

PR #3248, opened 2018-11-07
disable lint warning on not secure RNG on < 4.3

tobiasKaminsky:
> Key generation, signing, encryption, and random number generation may not receive cryptographically strong values due to improper initialization of the underlying PRNG on Android 4.3 and below.

The RNG in question is used to generate a random number for inter-app communication and thus it is not needed to that strong, but only needs to be random.


AndyScherzinger:
:+1:

PR #3249, opened 2018-11-07
use suggested timeout of 10min

tobiasKaminsky:
> Wakelocks have two acquire methods: one with a timeout, and one without. You should generally always use the one with a timeout. A typical timeout is 10 minutes. If the task takes longer than it is critical that it happens (i.e. can't use JobScheduler) then maybe they should consider a foreground service instead (which is a stronger run guarantee and lets the user know something long/important is happening).

So using 10min (10 min * 60s /min * 1000msec/s)


AndyScherzinger:
:+1: looks good but needs a rebase before merging to get the lint warning count right :+1:

PR #3250, opened 2018-11-07
SyncedFolderAdapter: add locale for number

tobiasKaminsky:
@AndyScherzinger why is this
https://github.com/nextcloud/android/blob/0b21837bf75701a08ff9f393f269e4390e1906c2/src/main/java/com/owncloud/android/ui/adapter/SyncedFolderAdapter.java#L229-L231

@Nullable? 
This results in many warnings.
If it is indeed nullable, we need to check every usage.


tobiasKaminsky:
I split the view holder into main and header.
Now Butterknife does not complain and we also have no nullable.

PR #3260, opened 2018-11-12
move MINIMUM_SUPPORTED_SERVER_VERSION to MainApp

tobiasKaminsky:
So we have both in place and  can change it directly.


AndyScherzinger:
👍

PR #3276, opened 2018-11-19
Simplify grep + awk command to just awk in readme file

hlovdal:
Awk is just as capable as grep to filter lines, so grep is not needed.


AndyScherzinger:
Hi @hlovdal thanks a lot for your contribution ❤️ 

@tobiasKaminsky will have a look when he is back, so we can then get this merged (depending on his feedback of course).

tobiasKaminsky:
:+1: tested and working.
Thanks for your contribution :tada:

AndyScherzinger:
:+1:

PR #3299, opened 2018-11-29
use 90 as first stable

tobiasKaminsky:
With next stable release, build version should be starting with 90 for stable.
This is not needed in a normal release cycle, but could be needed for an emergency release, like when a changelog is missing, or if a branded version needs to be rebuild and thus needs to have a new version code.


AndyScherzinger:
👍

PR #3303, opened 2018-12-01
Sets the cursor colour on the passcode screen to grey.

dan0xii:
Before this change, if the user has a custom colour set on their
nextcloud server, the cursor would be blue regardless of the custom
theme colour.

Before the change:
screenshot_20181201_123203

After:
screenshot_20181201_123241


tobiasKaminsky:
:+1: Thank you very much for your contribution! :tada:

PR #3318, opened 2018-12-04
Analysis: change to support 3.4.x

tobiasKaminsky:


AndyScherzinger:
@tobiasKaminsky broken?!

tobiasKaminsky:
Yes, the stable fingbugs info is only generated when a merge commit to the branch is done.
So we can see this only after merging this one.

AndyScherzinger:
👍 I see, well, then let's merge it I guess? :)

PR #3320, opened 2018-12-05
Show shared user

tobiasKaminsky:
Shows shared users on file list.

Needs feedback from https://github.com/nextcloud/server/issues/12874


src/main/java/com/owncloud/android/datamodel/OCFile.java
```diff
@@ -96,6 +96,12 @@
     @Getter @Setter private boolean favorite;
     @Getter @Setter private boolean encrypted;
     @Getter @Setter private WebdavEntry.MountType mountType;
+    @Getter
+    @Setter
+    private String ownerId;
+    @Getter
+    @Setter
+    private String ownerDisplayName;
```
AndyScherzinger: should be formatted as single lines

src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -857,6 +858,16 @@ Drawable doAvatarInBackground() {
 
             int px = getAvatarDimension();
 
+            boolean notOnSameServer = false;
+
+            if (notOnSameServer) {
+                try {
+                    return TextDrawable.createAvatar(mAccount.name, mAvatarRadius);
+                } catch (NoSuchAlgorithmException e) {
+                    e.printStackTrace();
```
AndyScherzinger: should be a proper log statement

src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -277,6 +279,21 @@ public void onBindViewHolder(@NonNull RecyclerView.ViewHolder holder, int positi
             if (holder instanceof OCFileListItemViewHolder) {
                 OCFileListItemViewHolder itemViewHolder = (OCFileListItemViewHolder) holder;
 
+                if (!TextUtils.isEmpty(file.getOwnerId()) && !mAccount.name.split("@")[0].equals(file.getOwnerId())) {
```
AndyScherzinger: we should probably move the split method based on the `"@"` into a utils class (not sure if we already have one for account/name things since I believe we do splitting based on the @ sign already in some other places so we should centralize that part.

src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -489,7 +506,12 @@ private void showShareIcon(OCFileListGridImageViewHolder gridViewHolder, OCFile
             sharedIconView.setImageResource(R.drawable.ic_unshared);
             sharedIconView.setContentDescription(mContext.getString(R.string.shared_icon_share));
         }
-        sharedIconView.setOnClickListener(view -> ocFileListFragmentInterface.onShareIconClick(file));
+
+        if (!TextUtils.isEmpty(file.getOwnerId()) && !mAccount.name.split("@")[0].equals(file.getOwnerId())) { // TODO refactor
```
AndyScherzinger: see comment abpve regarding handdling of @

src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -694,6 +716,17 @@ public Filter getFilter() {
...
     }
 
+    @Override
+    public void avatarGenerated(Drawable avatarDrawable, Object callContext) {
+        ((ImageView) callContext).setImageDrawable(avatarDrawable);
+    }
+
+    @Override
+    public boolean shouldCallGeneratedCallback(String tag, Object callContext) {
+//        return ((ImageView) callContext).getTag().equals(tag);
```
AndyScherzinger: remove commented out code in case we alsways return `false`

src/main/java/com/owncloud/android/datamodel/OCFile.java
tobiasKaminsky: Damn Android Studio is always changing this.
I had to do this on a different branch without AndroidStudio… :/

tobiasKaminsky: Please see #3325 for this

src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
tobiasKaminsky: :+1:

tobiasKaminsky:
2 findbugs increased:
- onBindView is longer/complexer and therefore generates a warning
- AccountUtils.accountOwnsFile:
`
STT_STRING_PARSING_A_FIELD: This method parses a String that is a field
This method calls a parsing method (indexOf, lastIndexOf, startsWith, endsWith, substring, indexOf) on a String that is a field, or comes from a collection that is a field. This implies that the String in question is holding multiple parts of information inside the string, which would be more maintainable and type safe if that value was a true collection or a first class object with fields, rather than a String.
`
This is referring to account.name, which we split to get the userId. 
We could use:
```
AccountManager accountManager = AccountManager.get(context);
        String userId = accountManager.getUserData(account,
                                                com.owncloud.android.lib.common.accounts.AccountUtils.Constants.KEY_USER_ID);
```

But this is not that lightweight like a string operation.

tobiasKaminsky:
@AndyScherzinger ready for review.
Although you did this some time ago, please have a look again, as there were quite some changes.

AndyScherzinger:
> Although you did this some time ago, please have a look again, as there were quite some changes.

I tested it, is the idea here to just show avatars from the user who shared it with me?

tobiasKaminsky:
> I tested it, is the idea here to just show avatars from the user who shared it with me?
 

- show avatar in list, hide share button as this behaves the same way
- show avatar and info in sharing view

AndyScherzinger:
both cases work just fine 👍

PR #3323, opened 2018-12-06
Show old lint warning from stable branch

tobiasKaminsky:
Now it retrieves value from stable branch and shows old value, corresponding to lint.html.

Previously it used the current branch, so if a lint decrease from 90 to 89 happened, lint-result would have been updated and thus it would have shown 89, but lint.html would still show 90 (as this is only updated on merge to $stable branch).


AndyScherzinger:
👍 sweet

AndyScherzinger:
merged, now running rebased #3322 PR

PR #3325, opened 2018-12-07
Export code style

tobiasKaminsky:
With this AndroidStudio should use the same code formatting for everyone.
Ref: https://github.com/nextcloud/android/pull/3320#discussion_r239712844

@AndyScherzinger can you test if this does the following:
```
package com.owncloud.android;
import lombok.Getter;
import lombok.Setter;
public class CodeFormatting {
    @Getter @Setter private String test; // this has to be done manually
    private void loop() {
        for(int i=0;i<10;i++)System.currentTimeMillis();

		boolean b;
        if(System.currentTimeMillis()>10)b=true;

        while(System.currentTimeMillis()>10)System.currentTimeMillis();
    }
}
```

should change after Alt+Shift+L (Reformat Code) to 
```
package com.owncloud.android;

import lombok.Getter;
import lombok.Setter;

public class CodeFormatting {
    @Getter @Setter private String test; // this has to be done manually, but it keeps the way

    private void loop() {
        for (int i = 0; i < 10; i++) {
            System.currentTimeMillis();
        }

		boolean b;
        if (System.currentTimeMillis() > 10) {
            b = true;
        }

        while (System.currentTimeMillis() > 10) {
            System.currentTimeMillis();
        }
    }
}
```
- correct new lines
- braces around one liner if/for/while

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,207 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
+    <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
+    <option name="PACKAGES_TO_USE_IMPORT_ON_DEMAND">
+      <value />
+    </option>
+    <option name="IMPORT_LAYOUT_TABLE">
+      <value>
+        <package name="android" withSubpackages="true" static="false" />
+        <emptyLine />
+        <package name="com" withSubpackages="true" static="false" />
+        <emptyLine />
+        <package name="junit" withSubpackages="true" static="false" />
+        <emptyLine />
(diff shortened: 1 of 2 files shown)


AndyScherzinger:
I gave it a try while during import AS is simply removing all the formatting config changes introduced in this PR.

mario:
Should you really be importing though @AndyScherzinger ? Or just opening?

AndyScherzinger:
Hmmm, well I just opened the project since due to the AS files AS is detecting it as a Android Project right away. So yes, I am just opening it.

tobiasKaminsky:
This also does not work for me :/
There is a https://www.jetbrains.com/help/idea/settings-tools-settings-repository.html, but this also does not work for me, when I am trying it locally…

tobiasKaminsky:
So I gave it a new try
- complete new clone from this repo
- close all projects in AS
- in this screen I used "open an existing Android Studio project"
image
- c&p code sample from first post
- Code -> Reformat Code
- seems to work

AndyScherzinger:
@tobiasKaminsky let's merge it then :)

PR #3339, opened 2018-12-11
ignore warning about reading file

tobiasKaminsky:
https://find-sec-bugs.github.io/bugs.htm#PATH_TRAVERSAL_IN

we have this warning in nearly every case, e.g:
https://github.com/nextcloud/android/blob/59cf6b9ba4965f38f2749f0624013daf0b83c946/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java#L534-L536

This is not a user input field, but OCFile, which cannot changed by the user.

So let us disable this.

### DIFF
--- a/findbugs-filter.xml
+++ b/findbugs-filter.xml
@@ -8,4 +8,5 @@
     <Match>
         <Class name="~.*\.R\$.*"/>
     </Match>
+    <Bug pattern="PATH_TRAVERSAL_IN" />
 </FindBugsFilter>


tobiasKaminsky:
@AndyScherzinger at some point we should think of adding findbugs the same like lint, so that drone errors if findbugs count increases…

AndyScherzinger:
👍 agreed yes

AndyScherzinger:
@tobiasKaminsky at some point could actually be "now" since we always want the scores to go down and never to go up.

Issue #3327, opened 2018-12-07
[3.4.0RC1] NPE on OCFileListBottomSheetDialog.onCreate


AndyScherzinger:
via GPlay Android 8+
```
java.lang.NullPointerException: 
  at com.owncloud.android.ui.fragment.OCFileListBottomSheetDialog.onCreate (OCFileListBottomSheetDialog.java:92)
  at android.app.Dialog.dispatchOnCreate (Dialog.java:494)
  at android.app.Dialog.show (Dialog.java:342)
  at com.owncloud.android.ui.fragment.OCFileListFragment.lambda$registerFabListener$0$OCFileListFragment (OCFileListFragment.java:394)
  at com.owncloud.android.ui.fragment.OCFileListFragment$$Lambda$0.onClick (Unknown Source:4)
  at android.view.View.performClick (View.java:6897)
  at android.view.View$PerformClick.run (View.java:26101)
  at android.os.Handler.handleCallback (Handler.java:789)
  at android.os.Handler.dispatchMessage (Handler.java:98)
  at android.os.Looper.loop (Looper.java:164)
  at android.app.ActivityThread.main (ActivityThread.java:6944)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.Zygote$MethodAndArgsCaller.run (Zygote.java:327)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1374)
```

tobiasKaminsky:
So capability is null. 

https://github.com/nextcloud/android/blob/eae43cf464f0b682879c02952b06c9e29284cb1d/src/main/java/com/owncloud/android/ui/activity/BaseActivity.java#L165-L173

It is stated that null only occurs if the account is not set, but when opening the FAB, there must be an account :thinking:

PR #3360, opened 2018-12-13
Capability was read too early

tobiasKaminsky:
Fix https://github.com/nextcloud/android/issues/3327
So we are now referencing the activity and once everything is up, we can use capability.

Test:
- enable "do not keep activities" in dev mode on android
- open app
- suspend app
- open app again
- click on "+"

Previously: crash
Now: no crash


AndyScherzinger:
👍

Issue #3035, opened 2018-09-13
[3.3.0RC2] NPE in TrashbinActivity.showTrashbinFolder


AndyScherzinger:
From GPlay
```
ava.lang.NullPointerException: 
  at com.owncloud.android.ui.trashbin.TrashbinActivity.showTrashbinFolder (TrashbinActivity.java:257)
  at com.owncloud.android.ui.trashbin.TrashbinPresenter$1.onSuccess (TrashbinPresenter.java:72)
  at com.owncloud.android.ui.trashbin.RemoteTrashbinRepository$ReadRemoteTrashbinFolderTask.onPostExecute (RemoteTrashbinRepository.java:211)
  at com.owncloud.android.ui.trashbin.RemoteTrashbinRepository$ReadRemoteTrashbinFolderTask.onPostExecute (RemoteTrashbinRepository.java:175)
  at android.os.AsyncTask.finish (AsyncTask.java:695)
  at android.os.AsyncTask.-wrap1 (Unknown Source)
  at android.os.AsyncTask$InternalHandler.handleMessage (AsyncTask.java:712)
  at android.os.Handler.dispatchMessage (Handler.java:105)
  at android.os.Looper.loop (Looper.java:164)
  at android.app.ActivityThread.main (ActivityThread.java:6798)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.Zygote$MethodAndArgsCaller.run (Zygote.java:240)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:767)
```

AndyScherzinger:
TrashbinActivity.showTrashbinFolder (TrashbinActivity.java:257):
```
swipeListRefreshLayout.setRefreshing(false);
```

tobiasKaminsky:
I expect this to be some kind of race condition, as a regular "show folder list" is working. Maybe the remote operation takes too long and then the view is already destroyed…

PR #3365, opened 2018-12-13
Trashbin: do UI changes of async only if activity is still up and running

tobiasKaminsky:
Fix https://github.com/nextcloud/android/issues/3035


AndyScherzinger:
:+1:

mario:
I'm sure we could use lifecycle observers here, but ... if it makes you happy :)

👍

tobiasKaminsky:
> use lifecycle observers

Good idea, I'll have a look into this, but for a quick bugfix it is fine, I hope.

Issue #2923, opened 2018-08-24
Drop Android 4.0.x support for app release 3.4.0


AndyScherzinger:
I'd like to slowly but steadily remove support for (very) old Android versions and would like to start with dropping 4.0.x making 4.1 the minimum version (baby steps). Looking at the numbers the app lost 16% of the total 4.0.x active installations dropping down to 161 devices, so I think we are safe to remove the support after 3.3.x which would then be the last feature release which is the first one to support Nc14 features.

Here is the actual 4.0.x active installations graph for the last month:
![android_4_0_x](https://user-images.githubusercontent.com/1315170/44570865-7c20a700-a77f-11e8-885d-e69137f4a683.png)

What do you think @mario @ardevd @tobiasKaminsky

ardevd:
I absolutely support this! I would probably make 5.0 the minimum version but as you say - baby steps!

mario:
5.0 pls.

AndyScherzinger:
Well, dropping 4.x and making 5.x the minimum version would drop support for 3,4% of the active users where 4.4 has the same amount of users as 5.0 (with both having more than 5000 users each!). That is why I thought of baby steps since I don't see a good reason to drop support for >5000 people.

Easier maintenance isn't an argument from my point of view if nobody finds the time to take care of improving the auto upload which is where > Android 5.0 might be helpful. So´in my opinion just dropping 4.x and leaving the actual features as-is, is just dropping support for a lot of people for no reason.

AndyScherzinger:
Which is why I opened the issue for **dropping_4.0.x_only**

rullzer:
4.0 is fine and totally defendable IMO. And indeed baby steps.

AndyScherzinger:
@oparoz can you clarify on this, especially talking about 4.0.x (since it has been release Oct. 2011, overall market share <0,3%, for Nc it's <0,1% meaning 150 devices, declining by about 20% each month).

AndyScherzinger:
@oparoz @tobiasKaminsky any news and feedback on dropping Android 4.0.x making Android 4.1 the minimum requirement?

tobiasKaminsky:
From my side, last info is that 4.x is required by a customer and thus we have to support it.
@oparoz, please clarify

rullzer:
Also note that google will most likely drop support for chrome in anything below 4.4 soon.

tobiasKaminsky:
@oparoz what is our least supported version?
[...]

PR #3383, opened 2018-12-17
Add support for QR codes & deep links

tobiasKaminsky:
- bump min sdk from 14 to 15. Needed by lib, but according to google we do not have any with <4.0.3
image
- in login screen, you can now scan a QR code (requires Camera access)
- you can also click on any nc://login… link and it will also log you in.
- if new library version up released, we can switch to use this


AndyScherzinger:
Can I be annoyed by this? Simply because I vote/fight/argue for a version bum 14->15 (see #2923) and get blocked but when we need a QR code scanner we happily do so?

tobiasKaminsky:
Hm. You are right, sorry :heart:!
The last time we talked about this, there were active 4.0.x devices, now there are none.
So we still should keep min 14, until we have a go in #2923.

I can show the button for >=15 only, but I have to figure out how I can do this in build.gradle.

AndyScherzinger:
> The last time we talked about this, there were active 4.0.x devices, now there are none.

Well, 13 to be true ;) And for customers it could be a case if they run a branded client ;)

> So we still should keep min 14, until we have a go in #2923.

No, we should drop 14. If _whoever_ can't provide any feedback regarding this, even within the last **3 months** then they shouldn't have a say in this or have to deal with the situation afterwards. You are either involved in this or you are not and if you are not (which seems obvious) than you don't get to vote on this. Just my 2 cents.

> I can show the button for >=15 only, but I have to figure out how I can do this in build.gradle.

Please let's move on to v15 🙏

tobiasKaminsky:
> Well, 13 to be true ;)

API14: 4.0, 4.0.1, 4.0.2 
API15: 4.0.3, 4.0.4

So, apart from any customers, fdroid, manual installations, we do not have 14 in use.

AndyScherzinger:
> So, apart from any customers, fdroid, manual installations, we do not have 14 in use.

correct 👍

AndyScherzinger:
@tobiasKaminsky I revamped the qr code icon :)
device-2018-12-19-090801

AndyScherzinger:
@tobiasKaminsky tested the scanner with non-Nc-login-link qr codes. The scan then returns to the login screen (like in the screenshot above) but I don't get any feedback as a user that the qr code is not a Nc-Login-code so I think it would be good to show a Toast or Snackbar stating exactly that. :)

AndyScherzinger:
@tobiasKaminsky some issues I found:

* use non fitting qr code -> return back to login screen (no feedback)
* use nc:// url with wrong credentials (spinner and return back to login overview screen where you choose between login and provider!) -> should end up on login screen again and also provide a feedback about the failed login
* test suite still uses `nextcloud:` as the protocoll not `nc://` should be updated within this PR

What works:
* login via proper qr code with valid credentials 👍

tobiasKaminsky:
@AndyScherzinger can you re-test it again?

AndyScherzinger:
@tobiasKaminsky erm, my commit is gone?! with my visual changes?!

AndyScherzinger:
@tobiasKaminsky rebased and fixed the scan icon. non-fitting QR codes still return back to login screen without any feedback. Just try this one:
frame

tobiasKaminsky:
Due to force push, where your qr image gone away, a permission for vibrate was also missing…

AndyScherzinger:
👍 Nice, all good. Just signing the code is missing...

marinofaggiana:
@tobiasKaminsky where is the image that you have used for QRCode button ?

AndyScherzinger:
@marinofaggiana it from the material design icons list: http://materialdesignicons.com/icon/qrcode-scan

PR #3391, opened 2018-12-18
Add check for template support

tobiasKaminsky:


AndyScherzinger:
@tobiasKaminsky bot failed to create the backport :/

tobiasKaminsky:
Yes, but this is not a big deal, as we cannot backport database changes without getting in a lot of mess.
So for 3.4.1 there will no check for template support, but we will silently assume that there is template support.

PR #3392, opened 2018-12-18
341rc1

tobiasKaminsky:
Possible fix for crash on Android 9.
Will be released from this branch to GPlay only for testing.


tobiasKaminsky:
So this does not seem to help, therefore I am closing this one.

PR #3399, opened 2018-12-20
Detail view: show complete date on click

### LINKED ISSUE #3389: File detail: show exact modification timestamp
In the tab 'All Files' the full time stamps (of the form hh:mm(:ss) dd-mm-yyyy) are shortened to 'x days ago' or the date 'Nov 18' for example. This is all fine and makes the overview much more readable, however, this same shortened version is also shown in the file details activity. That makes it impossible to check the full date and time of a file or folder, which is very inconvenient. 
screenshot_20181218-133132__01

tobiasKaminsky:
Resolves #3389 

### DIFF
--- a/src/main/java/com/owncloud/android/db/PreferenceManager.java
+++ b/src/main/java/com/owncloud/android/db/PreferenceManager.java
@@ -47,6 +47,7 @@ public final class PreferenceManager {
     private static final String AUTO_PREF__UPLOAD_FILE_EXTENSION_URL = "prefs_upload_file_extension_url";
     private static final String AUTO_PREF__UPLOADER_BEHAVIOR = "prefs_uploader_behaviour";
     private static final String AUTO_PREF__GRID_COLUMNS = "grid_columns";
+    private static final String AUTO_PREF__SHOW_DETAILED_TIMESTAMP = "detailed_timestamp";
     public static final String AUTO_PREF__LAST_SEEN_VERSION_CODE = "lastSeenVersionCode";
     private static final String PREF__INSTANT_UPLOADING = "instant_uploading";
     private static final String PREF__INSTANT_VIDEO_UPLOADING = "instant_video_uploading";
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
@AndyScherzinger 

Is there a reason why all UI elements are nullable?
```
 @Nullable @BindView(R.id.modified)
     TextView fileModifiedTimestamp;
```

Even if we open up an non-existing file, it should find the UI elements, or?

AndyScherzinger:
> Is there a reason why all UI elements are nullable?

can't really tell without looking at the code. Nullable is only needed in case where the UI element code be misse (e.g. when an activity or fragment loads different layout files).

AndyScherzinger:
👍 In general I am fine with the PR since it solves the problem, while I am also fine with discussing possibly better, yet unknown solutions to this problem, but first of all we just have some kind of fix for it 👍

tobiasKaminsky:
This needs to be changed so that it toggles complete/relative date.

AndyScherzinger:
> This needs to be changed so that it toggles complete/relative date.

@tobiasKaminsky DONE :) Please review :)

tobiasKaminsky:
:+1:

tobiasKaminsky:
Thanks for taking care of this!

tobiasKaminsky:
> @tobiasKaminsky regarding your question: [#3399 (comment)](https://github.com/nextcloud/android/pull/3399#issuecomment-448943887)
> yes the elements can technically be null, in case the other layout gets loaded, see [/src/main/java/com/owncloud/android/ui/fragment/FileDetailFragment.java@`master`#L173](https://github.com/nextcloud/android/blob/master/src/main/java/com/owncloud/android/ui/fragment/FileDetailFragment.java#L173)

If I pass a null file into FileDetailFragment, I get a loading spinner and nothing more happens, as of:
https://github.com/nextcloud/android/blob/94621b15ab99a4a23f7c5706866c086e59612802/src/main/java/com/owncloud/android/ui/fragment/FileDetailFragment.java#L247-L256

So we should be safe to change Butterknife, if you agree? (we can do this and the discussion on a separate PR)

AndyScherzinger:
> Thanks for taking care of this!

Always a pleasure

> So we should be safe to change Butterknife, if you agree? (we can do this and the discussion on a separate PR) 

Another PR would be good, yes. While I would expect it to crash then...

AndyScherzinger:
PR can be merged I think :)

PR #3428, opened 2019-01-06
adding propfind method to sso

PhieF:
Hi

Nextcloud app is missing the propfind method I need to implement webdav
Could it be added ? Thanks

ref: https://github.com/nextcloud/Android-SingleSignOn/issues/41

### DIFF
--- a/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
+++ b/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
@@ -26,6 +26,7 @@ import android.accounts.AuthenticatorException;
 import android.accounts.OperationCanceledException;
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.net.Uri;
 import android.os.Binder;
 import android.os.ParcelFileDescriptor;
 import android.text.TextUtils;
@@ -42,13 +43,21 @@ import com.owncloud.android.lib.common.OwnCloudClientManagerFactory;
 import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.utils.EncryptionUtils;
 
+import org.apache.commons.httpclient.HttpConnection;
 import org.apache.commons.httpclient.HttpMethodBase;
+import org.apache.commons.httpclient.HttpState;
 import org.apache.commons.httpclient.NameValuePair;
 import org.apache.commons.httpclient.methods.DeleteMethod;
 import org.apache.commons.httpclient.methods.GetMethod;
+import org.apache.commons.httpclient.methods.InputStreamRequestEntity;
 import org.apache.commons.httpclient.methods.PostMethod;
 import org.apache.commons.httpclient.methods.PutMethod;
+import org.apache.commons.httpclient.methods.RequestEntity;
 import org.apache.commons.httpclient.methods.StringRequestEntity;
+import org.apache.jackrabbit.webdav.DavConstants;
+import org.apache.jackrabbit.webdav.client.methods.MkColMethod;
+import org.apache.jackrabbit.webdav.client.methods.PropFindMethod;
+import org.apache.jackrabbit.webdav.property.DavPropertyNameSet;
 
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
@@ -100,9 +109,16 @@ public class InputStreamBinder extends IInputStreamService.Stub {
     }
 
     public ParcelFileDescriptor performNextcloudRequest(ParcelFileDescriptor input) {
+        return performNextcloudRequestAndBodyStream(input, null);
+    }
+
+    public ParcelFileDescriptor performNextcloudRequestAndBodyStream(ParcelFileDescriptor input,
+                                                                     ParcelFileDescriptor requestBodyParcelFileDescriptor) {
         // read the input
         final InputStream is = new ParcelFileDescriptor.AutoCloseInputStream(input);
 
+        final InputStream requestBodyInputStream = requestBodyParcelFileDescriptor != null ?
+            new ParcelFileDescriptor.AutoCloseInputStream(requestBodyParcelFileDescriptor) : null;
         Exception exception = null;
         InputStream httpStream = new InputStream() {
             @Override
@@ -110,10 +126,11 @@ public class InputStreamBinder extends IInputStreamService.Stub {
                 return 0;
             }
         };
+
         try {
             // Start request and catch exceptions
             NextcloudRequest request = deserializeObjectAndCloseStream(is);
-            httpStream = processRequest(request);
+            httpStream = processRequest(request, requestBodyInputStream);
         } catch (Exception e) {
             Log_OC.e(TAG, "Error during Nextcloud request", e);
             exception = e;
@@ -122,7 +139,12 @@ public class InputStreamBinder extends IInputStreamService.Stub {
         try {
             // Write exception to the stream followed by the actual network stream
             InputStream exceptionStream = serializeObjectToInputStream(exception);
-            InputStream resultStream = new java.io.SequenceInputStream(exceptionStream, httpStream);
+            InputStream resultStream;
+            if (httpStream != null) {
+                resultStream = new java.io.SequenceInputStream(exceptionStream, httpStream);
+            } else {
+                resultStream = exceptionStream;
+            }
             return ParcelFileDescriptorUtil.pipeFrom(resultStream, thread -> Log.d(TAG, "Done sending result"));
         } catch (IOException e) {
             Log_OC.e(TAG, "Error while sending response back to client app", e);
@@ -139,7 +161,8 @@ public class InputStreamBinder extends IInputStreamService.Stub {
         return new ByteArrayInputStream(baos.toByteArray());
     }
 
-    private <T extends Serializable> T deserializeObjectAndCloseStream(InputStream is) throws IOException, ClassNotFoundException {
+    private <T extends Serializable> T deserializeObjectAndCloseStream(InputStream is) throws IOException,
+        ClassNotFoundException {
         ObjectInputStream ois = new ObjectInputStream(is);
         T result = (T) ois.readObject();
         is.close();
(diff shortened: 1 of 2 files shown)


src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
```diff
@@ -147,6 +147,22 @@ public int read() {
         return result;
     }
 
+    public class PropfindMethod extends HttpMethodBase {
+        public PropfindMethod() {
+            setFollowRedirects(true);
+        }
+        public PropfindMethod(String uri) {
+            super(uri);
+            setFollowRedirects(true);
```
David-Development: Are redirects always required for Propfind calls? If so, this solution could have serious side effects as this line: https://github.com/nextcloud/android/blob/0d9b1c4acc3776663cc37af68b97f9845ae26c91/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java#L238 overrides it later on again.

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
```diff
@@ -176,6 +192,10 @@ private InputStream processRequest(final NextcloudRequest request) throws Unsupp
                 method = new GetMethod(requestUrl);
                 break;
 
+            case "PROPFIND":
+                method = new PropfindMethod(requestUrl);
```
David-Development: Suggestion: 

```java
method = new PropfindMethod(requestUrl);
request.setFollowRedirects(true);
```

or we could add a check and refuse calls that don't specify the setFollowRedirect..?

David-Development:
I guess the proper solution would be to use the `PropFindMethod` from the `android-library` which uses jackrabbit for all webdav methods. Example: https://github.com/nextcloud/android-library/blob/15c0f0dfb8845758f45646f4985c7d9ad199b1c6/src/com/owncloud/android/lib/resources/files/ReadFileRemoteOperation.java#L80 


@tobiasKaminsky Maybe we can support more webdav methods this way as well? (If I recall correctly there are quite a few). Which ones are relevant for nextcloud?

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
PhieF: I've just followed the same model as GetMethod (setfoolowredirects is always true in getmethod too)

David-Development: Oh okay, I see. It'll be overridden later on anyways so it won't make a difference I guess.

tobiasKaminsky:
Yes, we should use JackRabbit's method for this.
Setting "follow redirect" should not be needed.



> Which ones are relevant for nextcloud?

We are using delete, copy, mkcol, move options, search, (lock/unlock for E2E).
However I hesitate to put all of them in without proper testing…

David-Development:
> We are using delete, copy, mkcol, move options, search, (lock/unlock for E2E).

that's quite a few methods..

> However I hesitate to put all of them in without proper testing… 

agreed! 👍 

Since I'll be pretty busy this week I won't have much time to look into it right now. I guess one would have to write some tests to verify that the attributes from the `NextcloudRequest` will be converted into the actual webdav request. Furthermore I'm not sure if we need additional fields in the `NextcloudRequest`?

PhieF:
Would it be possible to put a high priority on propfind because I will need it, I won't have time to test the other properly but the propfind from my pull request works well at least for webdav

what do you think about that ?

PhieF:
@David-Development done ! thanks

David-Development:
@tobiasKaminsky is it possible that we need to call `exhaustResponse` before reading the response? Code looks fine to me so far. 

@PhieF can you post the code you used for making the request? (where you build the `NextcloudRequest`).

PhieF:
here it is

```
NextcloudAPI nextcloudAPI = mNextCloudWrapper.getNextcloudApi();
NextcloudRequest nextcloudRequest = new NextcloudRequest.Builder()
          .setMethod("PROPFIND")
          .setHeader(head)
          .setUrl("/remote.php/webdav/"+path)
          .build();
```

tobiasKaminsky:
> @tobiasKaminsky is it possible that we need to call `exhaustResponse` before reading the response? Code looks fine to me so far.

exhaust is normally done if we do not need the result, e.g.:
https://github.com/nextcloud/android/blob/master/src/main/java/com/owncloud/android/operations/UpdateOCVersionOperation.java#L75-L80

David-Development:
@PhieF just tested the provided code. I cannot verify the given error. I'm running the following request:

```java
final NextcloudRequest nextcloudRequest = new NextcloudRequest.Builder()
    .setMethod("PROPFIND")
    //.setHeader(head)
    .setUrl("/remote.php/webdav/")
    .build();

InputStream is = nextcloudAPI.performNetworkRequest(nextcloudRequest);
StringBuffer total = new StringBuffer();
// If response body is available
if (is != null) {
    BufferedReader reader = new BufferedReader(new InputStreamReader(is));
    String line = reader.readLine();
    while (line != null) {
        total.append(line).append('\n');
        line = reader.readLine();
    }
}
Log.e(TAG, total.toString());
```

what's does your header look like? What does your path look like?

My response is empty though.. But the requests returns a http status code: 207.

David-Development:
Setting break points in the `nextcloud-android` app might be helpful to figure out in which line the `NullPointerException` is thrown. (So basically you run both apps in debug mode and step through them).

PhieF:
Hi, 

I've removed the header and it still bugs
but removing

```
multiStatus = MultiStatus.createFromXml(getRootElement())
```
from
DavMethodBase

makes it work...
If you don't have any bugs on your side, could you please send me your apk ? It can come from my build for an unknown reason

David-Development:
@PhieF I'm not sure if mine is working properly either. I don't get an exception but if I run the following code, the result is empty. Reading through the docs, this request should return a list of folders, right? [Ref to docs](https://docs.nextcloud.com/server/12/developer_manual/client_apis/WebDAV/index.html#webdav-basics)

```java
final NextcloudRequest nextcloudRequest = new NextcloudRequest.Builder()
	.setMethod("PROPFIND")
	.setUrl("/remote.php/dav/files/david/")
	.build();
```

@tobiasKaminsky any ideas what's wrong here?

PhieF:
Yep, I might have found something : the inputstream can be read only once, and when they read it in MultiStatus.createFromXml(getRootElement()) it might result to an empty inputstream...

I guess I have to deal with getResponseBodyAsMultiStatus instead

David-Development:
@PhieF Oh, I see! Great, that makes perfect sense then! Reading an `InputStream` twice won't work. Let me know once you figured it out!

PhieF:
Sorry, I've edited my comment and I guess you haven't seen it

"Or perhaps I should make a custom PropFindMethod extending PropFindMethod of jackrabbit to avoid parsing the doc, because I guess it shouldn't be done in the nextcloud app"

what do you think of that ? Because nextcloud should send back the real inputstream, not modifying it

David-Development:
@PhieF Indeed, I didn't see your edit. Thanks!

 > Because nextcloud should send back the real inputstream, not modifying it

That is a serious design choice?


Do we have arguments in favor or against that "behaviour"? I see two options here:

1. Parse it in the nextcloud files app, serialize it, and return it as a xml document
  - less work for most devs as they don't have to parse the xml themself
2. Don't parse it and return the raw input stream
  - more work for devs as everyone has to decode the xml themself / handle errors etc?
  - Do we need jackrabbit in the "client-app" to parse it? Or does any xml parser does it's job here?


@tobiasKaminsky what are your thoughts on this? What would be the "cleaner" solution?

PhieF:
On my side I use "javax.xml.parsers" and it seems to work very well ! :)

The client app will need jackrabbit to handle multistatus response, I guess, as they are jackrabbit classes

PhieF:
By the way, 
Post Method and Put Method are stuck with JSON String as request body, I would need to add inputstream to send files (InputStreamRequestEntity)

I suggest, to keep it compatible, to add a requestInputStream in NextCloudRequest.java

David-Development:
@PhieF thanks for the update! 

I have a couple of questions: 
- In your last commit, you fixed "crashing when httpStream was null"
   - when does this happen? Why is the httpStream null?

> Post Method and Put Method are stuck with JSON String as request body, I would need to add inputstream to send files (InputStreamRequestEntity)

Yes, that is definitely true. Adding the input as an inputstream is the way to go I guess. Need to think about how to actually implement this as this is a major change to the library. Sending a file-path won't work in most cases due to permissions and visibility of files for different apps in android. So you would have to write your own file-provider or store the files in a public directory. The former is a lot of work and the later one is a security issue.

I'm quite busy right now with my thesis so I'm not sure if I'll be able to work on it this week.

For the server response, I use the `SequenceInputStream` to combine the serialized exception and the http response stream. I guess for sending files with the request (request-body) we would have to send them this way as well (first send the serialized `NextcloudRequest` and then the actual body) (Here: https://github.com/nextcloud/Android-SingleSignOn/blob/master/src/main/java/com/nextcloud/android/sso/api/NextcloudAPI.java#L295).

Can't think of another proper solution yet.. Or maybe you have some ideas as well?

PhieF:
"when does this happen? Why is the httpStream null?"

When I use PutMethod for example, the file is put and on server but the response input stream is null

about file path, I guess apps can always store their files in (former) /sdcard/Android/data/com.gnagna.app/ and can authorize other apps to access it, but that's right that's not the best way to do. 

I was looking at the SSO lib, and I've changed some stuffs to be allowed to send another inputstream to the nextcloud app

this:
```
public InputStream performNetworkRequest(NextcloudRequest request) throws Exception {

```
becomes:

```
public InputStream performNetworkRequest(NextcloudRequest request, InputStream requestBodyInputStream) throws Exception {

```

And of course I keep compatibility with the old version of the lib


On nextcloud app side I've also added an adequate parcefiledescriptor and handled case where it is null or not there to avoid breaking apps

I'm testing it right now

David-Development:
@PhieF 

> When I use PutMethod for example, the file is put and on server but the response input stream is null

Oh okay, I see. Makes sense!


> On nextcloud app side I've also added an adequate parcefiledescriptor and handled case where it is null or not there to avoid breaking apps

Awesome! Thank you so much for helping to improve this single-sign-on library! 

> it work pretty well for now ! :)

Looking forward to see the changes! 🎉

PhieF:
Well, that's normal, I'm using it :P And I really like the nextcloud "suite" 
I will push as soon as I'm sure I don't have to add anything :P
I will also add the "mkcol" method

PhieF:
Hi, 
I've pushed changes with the inputstream as we talked about ! :)

PhieF:
OK, now it should be alright ! Sorry for the mistakes

PhieF:
Hi, it should be okay now, do you need something else ?

Thanks !

David-Development:
@PhieF yes, I reviewed it yesterday and it looks great! Awesome! Thank you! I'll mention you as a contributor in the Android Single-Sign-On Library if that's okay with you?

@tobiasKaminsky I'm okay with merging this pr! 👍 I didn't actually test the webdav features as @PhieF already did that. I tested backwards compatibility with client apps using an older version of the sso library.

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
```diff
@@ -202,10 +222,50 @@ private InputStream processRequest(final NextcloudRequest request) throws Unsupp
                 method = new DeleteMethod(requestUrl);
                 break;
 
+            case "PROPFIND":
+                method = new NCPropFindMethod(requestUrl, DavConstants.PROPFIND_ALL_PROP, DavConstants.DEPTH_1);
```
tobiasKaminsky: Shouldn't we try to pass properties and depth?
Especially on query files it it useful to limit properties to only needed one.
This will reduce amount of transferred data.

tobiasKaminsky:
Just one idea, after that I'll would like to change code syntax a bit,
- line length 120 char
- spaces between if and (), …

Just ping me once it is ready :-)

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
PhieF: Well it can be set in the header by the client

tobiasKaminsky: Ah, ok. And then it will be used? So ALL_PROP and DEPTH_1 is only the default?

PhieF: It should be
on my client I use

```
 List<String>depth = new ArrayList<>();
        depth.add("0");
        header.put("Depth", depth);
        NextcloudRequest nextcloudRequest = new NextcloudRequest.Builder()
                .setMethod("PROPFIND")
                .setHeader(header)
                .setUrl("/remote.php/webdav/"+remotePath)
                .build();
```

It should work because the original header is replaced, and this is for depth

Properties are in the resquestbody

```
 if (request.getRequestBody() != null) {
                    //text/xml; charset=UTF-8 is taken from XmlRequestEntity... Should be application/xml
                    StringRequestEntity requestEntity = new StringRequestEntity(
                        request.getRequestBody(),
                        "text/xml; charset=UTF-8",
                        CHARSET_UTF8);
                    ((PropFindMethod) method).setRequestEntity(requestEntity);
                }
```

This is exactly what the PropFind constructor does

tobiasKaminsky:
:+1: fine by me

PhieF:
Hi ! 
Just to be sure: you don't need me to do anything before merging ? The DCO seems to fail again

PR #3439, opened 2019-01-10
Document provider: new file

tobiasKaminsky:
This needs edit file.

To create a new file, we first have to upload an empty file (webdav limitation), and then the 3rd party apps can write to it:

- open text editor
- create new file
- save via storage manager


AndyScherzinger:
@tobiasKaminsky could you rebase, target master & resolve conflicts please ❤️

AndyScherzinger:
@tobiasKaminsky seems to work while I have the impression to get to notifications as in the file being uploaded twice. Also likely to be that way since the activity screen will show two actions for the new file "created" and "changed" right afterwards.

PR #3446, opened 2019-01-12
Use default/device font

### LINKED ISSUE #3348: 3.4.0RC1 system font not used
I'm not sure if it's supposed to be a feature or if it's a bug, but on current 3.4.0RC1, the font used is not the same as the Android system font. I'm on Pixel 2, with Android 9,, with the regular system font and no special configuration. I think that for integration purpose, the ND app should stick with system font, like other apps.

AndyScherzinger:
Resolves #3348 use default/device font

follow up for #3237 so should also make it into 3.5.0


tobiasKaminsky:
:+1:

PR #3447, opened 2019-01-12
Removing expert mode checks

### LINKED ISSUE #3398: About "Expert mode"
Currently we have a switch in the settings called "Expert mode". Enabling it has no instant effect and there’s no description.

Do we have a list somewhere which lists what it does, and why we have it?

I was browsing the Signal repository the other day and came across their great [development ideology](https://github.com/signalapp/Signal-Android/blob/master/CONTRIBUTING.md#development-ideology) which is pretty similar to ours. A relevant excerpt:

> **3. There are no power users.** The idea that some users "understand" concepts better than others has proven to be, for the most part, false. If anything, "power users" are more dangerous than the rest, and we should avoid exposing dangerous functionality to them.

At Nextcloud we always aim to make everything properly usable, and ideally automatic. Most people do not switch settings. Anything we add to "expert mode" we would need to test and ensure it works, which is a not really justified amount of work for the small percentage of people who use these settings.

So let’s analyze what we have in expert mode, what we need it for, and how we could solve these things better. Ideally getting rid of "expert mode".

cc @nextcloud/android. Also recommended reading: [Choosing our preferences](https://ometer.com/preferences.html)

AndyScherzinger:
Resolves #3398 removing expert mode checks


tobiasKaminsky:
Good so far :+1: 
I think we agreed / discussed to have in upload list view:
- swipe to refresh for force re-scan
- trash bin icon on uploaded / failed row
- retry button failed (or should this be done via swipe to refresh? )

AndyScherzinger:
@tobiasKaminsky yes, these are the things we might want to change as described while I thought we simply bring backk the functionality first and then see how to adapt or maybe even completely remove (as mentioned by @janborchardt).

So I'd say we do that in a different PR and we can do it afterwards

AndyScherzinger:
@tobiasKaminsky yes, these are the things we might want to change as described while I thought we simply bring backk the functionality first and then see how to adapt or maybe even completely remove (as mentioned by @janborchardt).

So I'd say we do that in a different PR and we can do it afterwards

tobiasKaminsky:
Ok, but we should try to get it into next release, as otherwise this is a bit of half finish, at least to me.

tobiasKaminsky:
:+1:

tobiasKaminsky:
> @tobiasKaminsky wouldn't the discussed on top changes just be "UI optimizations" but wouldnt change the functionality?

Partly right. Expert mode was introduced by mario to hide these functions. If we now revert this, but not enhance it, we are on the same point as before.

I'll see what I can do.
Nevertheless let us merge this now.

jancborchardt:
@tobiasKaminsky for the record, agree with your points of what needs to be fixed. As for "retry button failed (or should this be done via swipe to refresh? )" – I would say both should be done. In the uploads view it might not be immediately obvious you can also retry via pull to refresh, and we should be as helpful as possible if there’s upload failures. So in the header "Upload failed" on the right there should be a way to "Retry all".

tobiasKaminsky:
Can you have a look at the current state over there https://github.com/nextcloud/android/pull/3472#issuecomment-455516257 and then comment in that PR?

PR #3448, opened 2019-01-12
Adding notification option to not show notifications for new media folders

### LINKED ISSUE #3080: Setting to disable new folder found notification
I use an app called Media Merger to automatically move pictures from various app folders into a central folder, and Nextcloud gives me a notification whenever it rediscovers these initial app folders when they briefly have a photo appear in them.

AndyScherzinger:
Resolve #3080 

| Notification | Settings |
| ------------- | --------- |
||device-2019-01-12-234906|


src/main/java/com/owncloud/android/jobs/MediaFoldersDetectionJob.java
```diff
@@ -60,6 +65,9 @@
 
     private static final String ACCOUNT_NAME_GLOBAL = "global";
     private static final String KEY_MEDIA_FOLDERS = "media_folders";
+    public static final String NOTIFICATION_ID = "NOTIFICATION_ID";
+
+    private static String DISABLE_DETECTION_CLICK = "DISABLE_DETECTION_CLICK";
```
tobiasKaminsky: also final?

tobiasKaminsky:
Code review: :heart: 

Do you have an easy way for testing?
I tried to download an image do sdcard/downloads, but no notification is popping up (I re-installed the app previously to not have downloads in any list)

src/main/java/com/owncloud/android/jobs/MediaFoldersDetectionJob.java
AndyScherzinger: yeas, should be final...

AndyScherzinger:
> Do you have an easy way for testing?

Yes incredibly easy ;)

Use a file manager, go to downloads and have an image in the download folder, then:
1. create a new folder "ABC" within downloads, copy the image into it and wait -> notification comes up -> click Disable
2. create a new folder "BCD" within downloads, copy the image into it and wait -> no notification coming up
3. go into the app's settings and toggle the notification hidding setting, then back in the file manager create a new folder "CDE" within downloads, copy the image into it and wait -> notification comes up

tobiasKaminsky:
Works as expected, great stuff!


device-2019-01-12-234906
Only thing I stumbled about is, that it is a bit strange for me that the new setting has a toggle button, but if you click on it / enable it, it vanishes right away.

I would rather keep it all the time (so that you can also disable it without having a notification before) 
or
change wording (and toggle button) to something like "click here to enable notification again". Then it would be fine with instantly vanishing and only showing it once notifications are disabled.

AndyScherzinger:
Reason there is a very, very small group of people who want to disable it and they can, so always showing it in the settings is way to prominent ;)

src/main/java/com/owncloud/android/jobs/MediaFoldersDetectionJob.java
AndyScherzinger: fixed with latest commit

AndyScherzinger:
@tobiasKaminsky changed the behavior as described in my latest comments and also fixed the code review issue you found, so afaik should be fine now and also has the behavior I would prefer

tobiasKaminsky:
:+1: 
Fine with the change, great work :tada:

tobiasKaminsky:
Two new FindBug warnings, one in "BadPractise" and one in "Security".
Can you have a look at them?
(at some point we should make this failing, like with lint)

AndyScherzinger:
> Can you have a look at them?

I'll have a look 👍

AndyScherzinger:
> at some point we should make this failing, like with lint
I created an issue for it: #3465

dbielz:
Hey, how about turning off adding new folders automatically, not only the notification? Where's the logic behind that? I could accept if new found subfolders in a synced folder would be added automatically, but adding EVERY new media folder, just because I want to sync ONE media folder? WTF?

dbielz:
Sorry, but this one annoys me already for quite a while, and now I found this discussion about turning off the notifications only.

AndyScherzinger:
The action should still be there, didn't get removed. However the action as any notification action only shows up when the notification has been expanded by the user

PR #3451, opened 2019-01-14
Show correct share error message

tobiasKaminsky:
- share file to user A
- share file to group B which contains user A
- previous: sharee failed
- now: path already shared with user

This also applies for other failing update share operations, eg. password to weak.


AndyScherzinger:
:+1:

PR #3457, opened 2019-01-14
FileDetails: no nullable in butterknife

tobiasKaminsky:
As discussed https://github.com/nextcloud/android/pull/3399#issuecomment-453925465

To test I simply set file to null in https://github.com/nextcloud/android/blob/58a2fd904a2196e28ad360989a47cb1e4a03eab6/src/main/java/com/owncloud/android/ui/fragment/FileDetailFragment.java#L170-L178


AndyScherzinger:
@tobiasKaminsky did you reformat the layout file? The diff is quite large on that one :/

tobiasKaminsky:
> @tobiasKaminsky did you reformat the layout file? The diff is quite large on that one :/

"no", I added the empty layout, but this had to be done in a LinearLayout, as ScrollView only allows one child, thus everything moved one layer down :/

PR #3461, opened 2019-01-15
Upload: only allow upload behaviour if writable folders

tobiasKaminsky:
Only allow upload behaviour if writable folders, otherwise they cannot be move/deleted.

Left: sdcard is writable, thus upload behaviour is enabled (and choosable), right / is read-only.
 

Same works for manual uploads.

 

@AndyScherzinger what do we want to do with existing synced folders?
Having some kind of migration step would require to show it somehow to the user…

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/UploadFilesActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/UploadFilesActivity.java
@@ -216,6 +216,8 @@ public class UploadFilesActivity extends FileActivity implements
             mCurrentDialog = null;
         }
 
+        checkWritableFolder(mCurrentDir);
+
         Log_OC.d(TAG, "onCreate() end");
     }
 
@@ -351,6 +353,7 @@ public class UploadFilesActivity extends FileActivity implements
             popDirname();
             mFileListFragment.onNavigateUp();
             mCurrentDir = mFileListFragment.getCurrentDirectory();
+            checkWritableFolder(mCurrentDir);
 
             if (mCurrentDir.getParentFile() == null) {
                 ActionBar actionBar = getSupportActionBar();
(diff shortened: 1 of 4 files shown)


AndyScherzinger:
> what do we want to do with existing synced folders?
Having some kind of migration step would require to show it somehow to the user…

would we have to show it or could we just do it in the background?

As for the manual upload, is the action then "copy"? Like greyed out below the only action from my pov is "keep in source folder"

AndyScherzinger:
> keep file in source folder

should then imho be the "default action" (wether read-only or not) 😃 would you choose a different one for display here?

tobiasKaminsky:
> > keep file in source folder
> 
> should then imho be the "default action" (wether read-only or not)  would you choose a different one for display here?

Once we detect read-only, we change the behaviour to "keep", and change back to stored behaviour once it is write-able again.

What exactly do you mean?

AndyScherzinger:
@tobiasKaminsky 
 

The image clearly says "copy", if you can't write, you can't copy. If for an upload we need to do an internal copy hidden from the user it is nothing that should be stated here. "keep in source folder" means "NO_COPY"

tobiasKaminsky:
> The image clearly says "copy", if you can't write, you can't copy. If for an upload we need to do an internal copy hidden from the user it is nothing that should be stated here. "keep in source folder" means "NO_COPY"

Sorry, I do not get it:
the source folder, in the screen shot above "/" is read only, thus we cannot move/delete the selected file.
So we can only copy/upload it to NC.

Is it about the wording?
Then maybe "source folder is read-only; file will only be uploaded"?

AndyScherzinger:
> Is it about the wording?
Then maybe "source folder is read-only; file will only be uploaded"?

That is exactly what I meant, yes 👍

tobiasKaminsky:
Done :+1:

AndyScherzinger:
:+1: Nice

PR #1357, opened 2017-08-16
Codacy: fix further issues

AndyScherzinger:
* fixing further issues reported by codacy: https://www.codacy.com/app/Nextcloud/android/issues
* move helper functions to a separate utils class and clean up dependencies between the classes
* cleanup upload list adapter

local tests on my phone where successful 😺

PR #3464, opened 2019-01-16
Chunked upload

### LINKED ISSUE #348: Dynamically increase upload chunk size for videos/fast wifi
### Actual behaviour

Direct upload of videos is incredible slow compared to uploading of photos
### Expected behaviour

I would expect that the upload will run with about 1-3mb/s which my phone usually transmits in the private wifi network e.g. ftp or samba. Instead upload speed is about 200kb/s which results in long upload times for videos with e.g. 300-400mb.
While a normal picture with size 1-2MB uploads within about or a less than a second, the videos don't have this speed.
### Steps to reproduce
1. activate direct upload for picture and video
2. take picture or video
3. watch them uploading.
### Environment data

Android version: 6.0 custom rom. 

Device model: 
Nexus 5

Stock or customized system:
Custom

Nextcloud app version:
1.3.1 - the current one

Nextcloud server version:
10.0.1 - the current one
#### Web server error log

n /a

n/a

```

I'd like to support investigation with answering questions, taking tests, etc. as requested and needed.. :)
```

### LINKED ISSUE #1026: Switch to new dav / enhance chunked uploading
Currently we are storing successful uploaded chunks client side.
This can now be enhanced to reduce client code:

- use new dav:  /dav/uploads/userx/<YOURRANDOMID>/
- list shows already uploaded chunks
  - iterate over them and check file size, if chunk is completely uploaded
- upload (remaining) chunks
- move from /dav/uploads/userx/<YOURRANDOMID>/.file to /dav/files/userx/<PATH>

### LINKED ISSUE #1238: Use different chunk sizes depending on network
Use bigger chunks for wifi, and smaller for 4g

tobiasKaminsky:
Needs: https://github.com/nextcloud/android-library/pull/223
Fix #348
Fix #1026
Fix #1238

- chunks based on connection type
	- wifi: 10mb
	- mobile: 1mb

To test:
- upload many files, check if it is all working
- upload file, cancel in between, switch connection, restart upload

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -210,9 +210,9 @@ dependencies {
     // dependencies for app building
     implementation 'androidx.multidex:multidex:2.0.1'
 //    implementation project('nextcloud-android-library')
-    genericImplementation 'com.github.nextcloud:android-library:master-SNAPSHOT'
-    gplayImplementation 'com.github.nextcloud:android-library:master-SNAPSHOT'
-    versionDevImplementation 'com.github.nextcloud:android-library:master-SNAPSHOT'
+    genericImplementation 'com.github.nextcloud:android-library:tests-SNAPSHOT'
+    gplayImplementation 'com.github.nextcloud:android-library:tests-SNAPSHOT'
+    versionDevImplementation 'com.github.nextcloud:android-library:tests-SNAPSHOT'
     implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
     implementation 'androidx.legacy:legacy-support-v4:1.0.0'
     implementation 'com.google.android.material:material:1.0.0'
--- a/build.gradle
+++ b/build.gradle
@@ -210,9 +210,9 @@ dependencies {
     // dependencies for app building
     implementation 'androidx.multidex:multidex:2.0.1'
 //    implementation project('nextcloud-android-library')
-    genericImplementation 'com.github.nextcloud:android-library:tests-SNAPSHOT'
-    gplayImplementation 'com.github.nextcloud:android-library:tests-SNAPSHOT'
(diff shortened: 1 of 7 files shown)


AndyScherzinger:
👍 code wise it looks fine to me. I don't have the time or infrastructure to test this right now. Also I couldn't find the part where 1mb/10mb is calculated.

tobiasKaminsky:
> Also I couldn't find the part where 1mb/10mb is calculated.

https://github.com/nextcloud/android/pull/3464/files#diff-be9f58efccd643d4cdacff9ff8572f24R585

If on mobile connection, we use 1Mb, else 10Mb.
Splitting chunks is then done in library.

AndyScherzinger:
@tobiasKaminsky thanks for the feedback, so should we merge this and wait for feedback via nightly dev and RCs?

tobiasKaminsky:
We rely now on having a userId, which is set by all new accounts
There was once a migration step, to add this for all existing accounts.

Do we hope that everyone has used the versions with this migration step? 
Nothing bad will happen, if no userId is present, but upload will simply not work.

AndyScherzinger:
> Do we hope that everyone has used the versions with this migration step?
Nothing bad will happen, if no userId is present, but upload will simply not work.

@tobiasKaminsky Do you know _when_ this change with the migration step got introduced?

tobiasKaminsky:
> @tobiasKaminsky Do you know _when_ this change with the migration step got introduced?

Introduced: https://github.com/nextcloud/android/pull/1265 at 07.08.2017 with version 2.0.0
Removed: https://github.com/nextcloud/android/pull/1357 at 08.11.2017 with version 3.0.0

Hm. Maybe a bit too fast…?

AndyScherzinger:
As discussed on IRC I vote to include it in 3.5.0 RC1 🚀

PR #3467, opened 2019-01-17
Direct camera upload

tobiasKaminsky:
Allows to directly upload an image taken by the camera.

@AndyScherzinger started the work and I finished it.

As there are also changes to computing available space and upload, please also test this.
(I did tested it, but who knows ;-))

Signed-off-by: Tobias Kaminsky <tobias@kaminsky.me>


AndyScherzinger:
> E.g. where does it store the image taken?

It stores it in the folder where you actually opened the bottom menu, it is basically take-picture+trigger-manual-upload -> thus doesn't interfere with Auto Upload

AndyScherzinger:
As for auto upload, well see https://github.com/nextcloud/android/pull/3467/files#diff-486af2fc402fdd9e8cb77c5cc88cdfc8 If somebody configured this folder for auto upload / cusotm folder then yes it'll also trigger auto upload

tobiasKaminsky:
Oh boy, I should not have added anything to ConfirmationDialog…
I'll fix it now.

tobiasKaminsky:
Rebased :exclamation:

tobiasKaminsky:
>ANDROID_EXTERNAL_FILE_ACCESS: External file access (Android)
The application write data to external storage (potentially SD card). There are multiple security implication to this action. First, file store on SD card will be accessible to the application having the READ_EXTERNAL_STORAGE permission. Also, if the data persisted contains confidential information about the user, encryption would be needed.

Disabling this FindBugs bug as this is known and in our case intended.

AndyScherzinger:
👍 all fine with me (obviously 😉))

PR #3471, opened 2019-01-18
Receive files activity: reflect write status

tobiasKaminsky:
Left: no write permission
Right: write permission
 

To test:
- share a file to NC
- have a shared folder which is read only


AndyScherzinger:
:+1: Fine with me 👍

PR #3481, opened 2019-01-21
PullApprove: Only one review

tobiasKaminsky:
Although we disussed to have two reviews, this seems not to scale (for now).
So I revert this to at least one review. Otherwise we would have to use admin rights to merge a PR all the time.
This does not mean that we should not try to have 2 reviews, if possible.


AndyScherzinger:
👍 fully agree on your statement

PR #3490, opened 2019-01-23
AndroidX for AlertDialogs

AndyScherzinger:
basically housekeeping, making use of AndroidX implementation for remaining AlertDialogs where we missed this before. Most AlertDialogs have already been done via AndroidX.


tobiasKaminsky:
:+1:

PR #3500, opened 2019-01-27
Added blacklisting of files for uploads

### LINKED ISSUE #3418: Filter Uploads to exclude files
Filter Uploads to exclude files, quiet like the desktop app. That would enable the user to exclude files like the .thumbdataXX files android creates for dcim-folders.

newhinton:
Closes #3418

### DIFF
--- a/src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
+++ b/src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
@@ -35,11 +35,16 @@ import com.owncloud.android.ui.activity.PassCodeActivity;
 import com.owncloud.android.ui.activity.SettingsActivity;
 import com.owncloud.android.utils.FileSortOrder;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
 
 import androidx.annotation.Nullable;
 
+import static com.owncloud.android.ui.activity.SettingsActivity.PreferenceKeys.EXCLUDED_AUTOUPLOAD_PATTEN_DEFAULT_VALUES;
 import static com.owncloud.android.ui.fragment.OCFileListFragment.FOLDER_LAYOUT_LIST;
 
 /**
@@ -87,11 +92,20 @@ public final class AppPreferencesImpl implements AppPreferences {
     private static final String PREF__MIGRATED_USER_ID = "migrated_user_id";
     private static final String PREF__PHOTO_SEARCH_TIMESTAMP = "photo_search_timestamp";
     private static final String PREF__POWER_CHECK_DISABLED = "power_check_disabled";
+    public static final String EXCLUDED_AUTOUPLOAD_PATTEN_KEY = "EXCLUDED_AUTOUPLOAD_PATTEN_KEY";
+    public static final Set<String> EXCLUDED_AUTOUPLOAD_PATTEN_DEFAULT_VALUES = new HashSet<>();
 
     private final Context context;
     private final SharedPreferences preferences;
     private final CurrentAccountProvider currentAccountProvider;
     private final ListenerRegistry listeners;
+
+    static {
+        String[] SET_VALUES = new String[] {
+            ".thumbdata*"
+        };
+        EXCLUDED_AUTOUPLOAD_PATTEN_DEFAULT_VALUES.addAll(Arrays.asList(SET_VALUES));
+    }
 
     /**
      * Adapter delegating raw {@link SharedPreferences.OnSharedPreferenceChangeListener} calls
@@ -563,6 +577,16 @@ public final class AppPreferencesImpl implements AppPreferences {
     @Override
     public long getPhotoSearchTimestamp() {
         return preferences.getLong(PREF__PHOTO_SEARCH_TIMESTAMP, 0);
+    }
+
+    @Override
+    public void setAutoUploadPatternBlackList(List<String> patternBlackList) {
+        preferences.edit().putStringSet(EXCLUDED_AUTOUPLOAD_PATTEN_KEY, new HashSet<>(patternBlackList)).apply();
+    }
+
+    @Override
+    public List<String> getAutoUploadPatternBlackList() {
+        return new ArrayList<>(preferences.getStringSet(EXCLUDED_AUTOUPLOAD_PATTEN_KEY, EXCLUDED_AUTOUPLOAD_PATTEN_DEFAULT_VALUES));
     }
 
     /**
--- a/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
+++ b/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
@@ -84,8 +84,10 @@ import java.io.RandomAccessFile;
 import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
 import java.nio.channels.OverlappingFileLockException;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -150,6 +152,9 @@ public class UploadFileOperation extends SyncOperation {
     final private PowerManagementService powerManagementService;
 
     private boolean encryptedAncestor;
+
+    private List<String> excludedFilesyncPatterns=new ArrayList<String>();
+
 
     public static OCFile obtainNewOCFileToUpload(String remotePath, String localPath, String mimeType) {
 
@@ -228,6 +233,8 @@ public class UploadFileOperation extends SyncOperation {
         // Ignore power save mode only if user explicitly created this upload
         mIgnoringPowerSaveMode = mCreatedBy == CREATED_BY_USER;
         mFolderUnlockToken = upload.getFolderUnlockToken();
+
+        excludedFilesyncPatterns.add(".thumbdata*");
     }
 
     public boolean isWifiRequired() {
@@ -739,7 +746,45 @@ public class UploadFileOperation extends SyncOperation {
             remoteOperationResult =  new RemoteOperationResult(ResultCode.LOCAL_FILE_NOT_FOUND);
         }
 
+        if(isBlacklistedFile(originalFile.getName())){
+            Log_OC.d(TAG, "name: "+originalFile.getName()+" is blacklisted. Not uploading.");
+            //Todo: Proper returncode to ignore this result. Currently the app displays "Wrong username or password" which is not correct.
+            remoteOperationResult =  new RemoteOperationResult(ResultCode.UNAUTHORIZED);
+        }
+
         return remoteOperationResult;
+    }
+
+    private boolean isBlacklistedFile(String fileName){
+        boolean isBlacklisted=false;
+        for (String pattern : excludedFilesyncPatterns) {
+            if (fileName.matches(createRegexFromGlob(pattern))) {
+                Log_OC.d(TAG, "Blacklisted file: '"+fileName+"' matched with pattern :'"+pattern+"'");
+                isBlacklisted = true;
+            }
+        }
+
+        return isBlacklisted;
+    }
+
+    //https://stackoverflow.com/questions/1247772/is-there-an-equivalent-of-java-util-regex-for-glob-type-patterns
+    private static String createRegexFromGlob(String glob)
+    {
+        String out = "^";
+        for(int i = 0; i < glob.length(); ++i)
+        {
+            final char c = glob.charAt(i);
+            switch(c)
+            {
+                case '*': out += ".*"; break;
+                case '?': out += '.'; break;
+                case '.': out += "\\."; break;
+                case '\\': out += "\\\\"; break;
+                default: out += c;
+            }
+        }
+        out += '$';
+        return out;
     }
 
     private RemoteOperationResult normalUpload(OwnCloudClient client) {
--- a/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
+++ b/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
@@ -24,6 +24,7 @@ package com.owncloud.android.operations;
 import android.accounts.Account;
 import android.annotation.SuppressLint;
 import android.content.Context;
+import android.content.SharedPreferences;
 import android.net.Uri;
 import android.os.Build;
 import android.text.TextUtils;
@@ -42,6 +43,7 @@ import com.owncloud.android.datamodel.OCFile;
 import com.owncloud.android.datamodel.ThumbnailsCacheManager;
 import com.owncloud.android.datamodel.UploadsStorageManager;
 import com.owncloud.android.db.OCUpload;
+import com.owncloud.android.db.PreferenceManager;
 import com.owncloud.android.files.services.FileUploader;
 import com.owncloud.android.lib.common.OwnCloudClient;
 import com.owncloud.android.lib.common.network.OnDatatransferProgressListener;
@@ -62,6 +64,8 @@ import com.owncloud.android.lib.resources.files.ReadFileRemoteOperation;
 import com.owncloud.android.lib.resources.files.UploadFileRemoteOperation;
 import com.owncloud.android.lib.resources.files.model.RemoteFile;
 import com.owncloud.android.operations.common.SyncOperation;
+import com.owncloud.android.ui.activity.Preferences;
+import com.owncloud.android.utils.ConnectivityUtils;
 import com.owncloud.android.utils.EncryptionUtils;
 import com.owncloud.android.utils.FileStorageUtils;
 import com.owncloud.android.utils.MimeType;
@@ -756,8 +760,12 @@ public class UploadFileOperation extends SyncOperation {
     }
 
     private boolean isBlacklistedFile(String fileName){
+
+        SharedPreferences sharedpreferences = PreferenceManager.getDefaultSharedPreferences(this.getContext());
+        List<String> list = new ArrayList<String>(sharedpreferences.getStringSet(Preferences.PreferenceKeys.EXCLUDED_AUTOUPLOAD_PATTEN_KEY, Preferences.PreferenceKeys.EXCLUDED_AUTOUPLOAD_PATTEN_DEFAULT_VALUES));
+
         boolean isBlacklisted=false;
-        for (String pattern : excludedFilesyncPatterns) {
+        for (String pattern : list) {
             if (fileName.matches(createRegexFromGlob(pattern))) {
                 Log_OC.d(TAG, "Blacklisted file: '"+fileName+"' matched with pattern :'"+pattern+"'");
                 isBlacklisted = true;
@@ -780,7 +788,7 @@ public class UploadFileOperation extends SyncOperation {
                 case '?': out += '.'; break;
                 case '.': out += "\\."; break;
                 case '\\': out += "\\\\"; break;
-                default: out += c;
+                default: out += c; break;
             }
         }
         out += '$';
--- a/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
+++ b/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
@@ -35,6 +35,7 @@ import com.evernote.android.job.util.Device;
 import com.google.gson.reflect.TypeToken;
 import com.nextcloud.client.device.PowerManagementService;
 import com.nextcloud.client.network.ConnectivityService;
+import com.nextcloud.client.preferences.AppPreferencesImpl;
 import com.owncloud.android.datamodel.ArbitraryDataProvider;
 import com.owncloud.android.datamodel.DecryptedFolderMetadata;
 import com.owncloud.android.datamodel.EncryptedFolderMetadata;
@@ -43,7 +44,6 @@ import com.owncloud.android.datamodel.OCFile;
 import com.owncloud.android.datamodel.ThumbnailsCacheManager;
 import com.owncloud.android.datamodel.UploadsStorageManager;
 import com.owncloud.android.db.OCUpload;
-import com.owncloud.android.db.PreferenceManager;
 import com.owncloud.android.files.services.FileUploader;
 import com.owncloud.android.lib.common.OwnCloudClient;
 import com.owncloud.android.lib.common.network.OnDatatransferProgressListener;
@@ -64,8 +64,6 @@ import com.owncloud.android.lib.resources.files.ReadFileRemoteOperation;
 import com.owncloud.android.lib.resources.files.UploadFileRemoteOperation;
 import com.owncloud.android.lib.resources.files.model.RemoteFile;
 import com.owncloud.android.operations.common.SyncOperation;
-import com.owncloud.android.ui.activity.Preferences;
-import com.owncloud.android.utils.ConnectivityUtils;
 import com.owncloud.android.utils.EncryptionUtils;
 import com.owncloud.android.utils.FileStorageUtils;
 import com.owncloud.android.utils.MimeType;
@@ -759,15 +757,15 @@ public class UploadFileOperation extends SyncOperation {
         return remoteOperationResult;
     }
 
-    private boolean isBlacklistedFile(String fileName){
+    private boolean isBlacklistedFile(String fileName) {
 
-        SharedPreferences sharedpreferences = PreferenceManager.getDefaultSharedPreferences(this.getContext());
-        List<String> list = new ArrayList<String>(sharedpreferences.getStringSet(Preferences.PreferenceKeys.EXCLUDED_AUTOUPLOAD_PATTEN_KEY, Preferences.PreferenceKeys.EXCLUDED_AUTOUPLOAD_PATTEN_DEFAULT_VALUES));
+        SharedPreferences sharedpreferences = android.preference.PreferenceManager.getDefaultSharedPreferences(this.getContext());
+        List<String> list = new ArrayList<>(sharedpreferences.getStringSet(AppPreferencesImpl.EXCLUDED_AUTOUPLOAD_PATTEN_KEY, AppPreferencesImpl.EXCLUDED_AUTOUPLOAD_PATTEN_DEFAULT_VALUES));
 
-        boolean isBlacklisted=false;
+        boolean isBlacklisted = false;
         for (String pattern : list) {
             if (fileName.matches(createRegexFromGlob(pattern))) {
-                Log_OC.d(TAG, "Blacklisted file: '"+fileName+"' matched with pattern :'"+pattern+"'");
+                Log_OC.d(TAG, "Blacklisted file: '" + fileName + "' matched with pattern :'" + pattern + "'");
                 isBlacklisted = true;
             }
         }
--- a/src/main/java/com/owncloud/android/ui/dialog/PatternBlacklistEditorDialogFragment.java
+++ b/src/main/java/com/owncloud/android/ui/dialog/PatternBlacklistEditorDialogFragment.java
@@ -1,6 +1,5 @@
 package com.owncloud.android.ui.dialog;
 
-import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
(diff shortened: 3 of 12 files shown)


newhinton:
This is a basic implementation for a filter mechanism which prevents the (auto)upload of a given file if the name matches a pattern. Currently there are following problems:

-Where should this get it's place in the ui? We could do the same as the desktop-client does.
-Where should i store the list of excluded patterns?
-I am using a snippet from stackoverflow to convert a glob style pattern to a regex one. I provided a link to the original in a comment over the snippet. Is that okay to do, license-wise?

I'd appreciate any help :)

newhinton:
I have added a new DialogFragment, which contains a list with all the patterns.

I still dont know on how to persist the list, any suggestions are appreciated.

src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -744,9 +751,47 @@ private RemoteOperationResult checkConditions(File originalFile) {
...
+    }
+
+    //https://stackoverflow.com/questions/1247772/is-there-an-equivalent-of-java-util-regex-for-glob-type-patterns
+    private static String createRegexFromGlob(String glob)
+    {
+        String out = "^";
+        for(int i = 0; i < glob.length(); ++i)
+        {
+            final char c = glob.charAt(i);
+            switch(c)
```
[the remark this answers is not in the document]
newhinton: Is this the missing break in the default case?

AndyScherzinger:
Hi @newhinton :) I'll try to mention the afaik right people to answer the questions:

> -Where should this get it's place in the ui? We could do the same as the desktop-client does.

likely a general question for @nextcloud/designers and @tobiasKaminsky if/how exclude patterns for the auto upload should be added to the UI. I am not sure if they should be added (I understand the desktop client has such a feature) or if we should rather have a fixed list of patterns (like implemented within this PR, to not upload *.thumbdata* files). So any feedback is appreciated and if added, also if this is a glbal setting (for all auto upload folders, just custom auto uploads (I guess so), per account?, 1 setting for any *)

> -Where should i store the list of excluded patterns?

Highly depends on the discussion and decision on the first question in my opinion. @tobiasKaminsky 

> -I am using a snippet from stackoverflow to convert a glob style pattern to a regex one. I provided a link to the original in a comment over the snippet. Is that okay to do, license-wise?

@schiessle can you answer this? :)

> -Which RemoteOperationResult should i use? Currently im using unautorized, but that results in a wrong notification, but a blacklisted file should not create a notification at all

@tobiasKaminsky can you answer this? :)

newhinton:
the thing with fixed lists are, well, they are fixed ^^
Imho we should keep this as close as possible to the desktop-client, so that a user should have no trouble at all to understand how those patterns work. Also the user cannot remove excluded pattern if he needs to, which can be a problem if he needs excluded files

tobiasKaminsky:
My opinion
- we try to have as few options as possible, but instead work with sane defaults
- an exclude list should be globally for any auto-upload folder, otherwise this will be totally confusing
- configurable list seems to be very low demanded; I cannot recall any other request like this on gplay comments, nor somewhere else
- existing issues point to .thumbnail file only

So my idea is to first have a simple hardcoded blacklisted pattern list to exclude.

tobiasKaminsky:
After deciding how this feature shall look like, we can discuss implementation detail (like licence, ResultCode, etc.)

newhinton:
I understand and agree withyour point regarding adding as few settings as possible, but i think excluding files without the ability to disable this behaviour may also be confusing/really bad. I tried to keep this function as close as possible to the desktop-client, so users dont have to deal with multiple different-working sync-features

newhinton:
Also a good reason to allow customizing: custom auto uploads can be from a gamefolder or a folder which creates a lot of tempfiles or so on, where users could want to prevent uploading of those files

newhinton:
@nextcloud/designers What do you think about this? 

Besides the obvious shortfalls of my current design (wrong buttons, listlayout broken, etc) i think this would serve just right and behaves like the desktop version


2019-02-07 22_47_11-window

newhinton:
@tobiasKaminsky Also i have a different problem: my build wont compile anymore. Master works fine, but this branch wont compile because it cannot find ProgressiveDataTransferer. (The same on my old repo, and a fresh copy from git) Basically com.owncloud.android.lib.* does not resolve anymore except on the master branch. Any ideas why that is? Or how i can solve this?

tobiasKaminsky:
try to rebase your branch on latest master, there were some name changes…

AndyScherzinger:
@newhinton I did a rebase for you ❗️ Should now work fine again.

newhinton:
@AndyScherzinger That did it, thanks!

AndyScherzinger:
@newhinton you are very welcome ❤️

Siggi0904:
when'll be these branch merged into master?

I need these very nice feature. :-)

newhinton:
@Siggi0904 This is currently in development. While i have figured out the basics, putting it together takes some additional time, and after that we may have to discuss changes. I hope this wont take too long, but this is by no means ready to release ;)

Siggi0904:
ok, if you need a tester, write me. :-)

newhinton:
@AndyScherzinger @tobiasKaminsky
This basically works now. Except for custom uploads, which do not seem to work at all. (They don't trigger. I curently have set up the Downloads-Folder as a custom one, and an auto-detected folder, but only the auto-detected works. @Siggi0904 Can you confirm this behaviour with this branch?)

Besides that: there are a few minor ui-issues:
-The fragment is to small in size for my taste, i'd like to have it the same size as the "set up custom upload" fragment
-The 'save'&'close' buttons are the wrong (default-blue) color
-The 'add' button is currently a black +. Is there a better icon i should use?

That, and the open questions regarding licensing and the sane defaults need to be discussed.

jancborchardt:
Only read the beginning of the discussion, here’s my design input – basically I agree with @tobiasKaminsky and especially @AndyScherzinger’s comment here:

> I am not sure if they should be added (I understand the desktop client has such a feature) or if we should rather have a fixed list of patterns (like implemented within this PR, to not upload _.thumbdata_ files).

These patterns to exclude thumbnail files and temp folders are just common sense, we don’t need anything in the UI for that. The desktop client is not a good comparison as it is a "we always had it and people will complain" issue rather than a specific design decision we ever made.

newhinton:
@jancborchardt Basically, i would agree. I am somewhat sure that i wouldn't use that feature heavily(From the ui/customizability). But there is a problem: what are sensible defaults? Do you know every filename(They may be different between androidforks), and what if the thumbnails are timestamped? What if an app adds a new tempfile which we do not cover? We could find many filenames which would be nessessary to exclude, but adding/removing one would need an app update. Also what if a user wants to keep a temp file and sync it? I am not sure how we would cover such cases, but i dont think we should ignore them either

newhinton:
Also, how would a user know that some files are excluded? If there is no userinterface at all, we hide this feature from the user. He would have no way to find it, and he would simply wonder why the heck his custom sync does not work properly because not all files are synced and some are missing. Which basically means that the sync which should sync everything does not work, which reduces the confidence in this app by users who don't know that filtering exists

jancborchardt:
> Also, how would a user know that some files are excluded? If there is no userinterface at all, we hide this feature from the user. He would have no way to find it, and he would simply wonder why the heck his custom sync does not work properly because not all files are synced and some are missing.

Even if we have an interface for it, regular people should not need to look through that. So we need sane defaults anyway. :)

What sane defaults actually are, I’ll leave up to the Android experts. On desktop there are certain filetypes and folders which do not have user data in them, like .DS_Store and such. Image thumbnails also belong in this category – there would be no expectation of sync since they do not show up anywhere.



> What are sensible defaults? Do you know every filename? (They may be different between androidforks)
> What if the thumbnails are timestamped?
> What if an app adds a new tempfile which we do not cover?

As usual, the perfect is the enemy of the good. ;) People don’t want to manage exclude rules. They just want their stuff safe.

newhinton:
Yeah, don't get me wrong, i am not arguing against sane defaults, i have the same opinion like you regarding that ;) sane defaults are good and important 

What i think is problematic is not providing an ui for users who want/nees the customizability and to inform them about the filtering

jancborchardt:
> What i think is problematic is not providing an ui for users who want/nees the customizability and to inform them about the filtering

I do see your point, but the issue is that everything has to be built _and_ maintained. :) Especially these specific settings only accounting for a tiny fraction of users (if at all – many times people only _think_ they need these things) are always very involved.

There are plenty of complicated open source sync apps which are customizable to the last level. Nextcloud aims to be an alternative for people who are not routinely fiddling with settings. And with the limited resources we have, we need to focus on what brings value to the majority of people using Nextcloud.

newhinton:
What i mean is:
we sync everything, except e.g. *.temp because it is a temp file.

User is using a custom (or even autoupload) to sync his images from SuperPhotoApp.
Then we update nextcloud, and suddently, .temp files aren't synced anymore, but user wants/needs those files for whatever reason. He does not know that we are filtering, so he thinks nextcloud is broken (which would be the best case here)

We need to at least inform the user about the filtering, and preferably what is beeing filtered. We cannot simply not sync some files and not tell the user

jancborchardt:
Right, but let’s start with not hypotheticals but actual cases. :)

You opened the original issue at https://github.com/nextcloud/android/issues/3418, saying:

> Filter Uploads to exclude files, quiet like the desktop app. That would enable the user to exclude files like the .thumbdataXX files android creates for dcim-folders.

So `.thumbdata*` would then be the first exclude rule as it’s a specific case you brought up.

> .temp files aren't synced anymore, but user wants/needs those files for whatever reason.

This is speculation, and we can’t design properly on that. :) Like: .temp files are temporary, so why would they need to be synced? Or we could ask the person why they need the files. If that user ever happens to exist.

But we should not take imaginary scenarios as a justification to introduce a management interface for what should _just work_ in the background. (As said – the desktop app is not a good benchmark for this.)

newhinton:
tbh, if i get my tempfiles included in this pr, im fine with not including any explanation, but i don't think that is the way we should go :D

newhinton:
its not quiet about the ability to manage it, its about filtering at all? shouldnt the user know that we are filtering at all?

something like "hey, some files were excluded in the upload because they are tempfiles"

jancborchardt:
> something like "hey, some files were excluded in the upload because they are tempfiles"

Regular people don’t even know what a "temporary file" is. :) Even less what `.thumbdata` or whatever is. It would expose technical minutiae that’s irrelevant and should just work.

(E.g. we also don’t require the protocol to be put in with the server or the full path, to the /dav or whatever – it just works, using https as default. And people are not informed about every single detail of it because it’s really not relevant.)

newhinton:
i get that ;) but you are sure there should be no hint whatsover? to me it seems really strange to not inform about whats going on, but you are the pro here ;)

jancborchardt:
Yes I’m quite sure as that’s also what other platforms are doing. :) In any case we always do things step by step and iterate – on the scale from "Nothing" to "Full-blown management interface", one small step is simply adding some automatic rules.

AndyScherzinger:
@newhinton are you willing to provide such a "simple" PR? That would be highly appreciated ❤️

newhinton:
its no problem whatsoever to do that, as long as my pattern is included ^^ however, there are still some technical questions open:
-how to supress the failed upload notification
-license stuff regarding stackoverflow
and how to apply this to custom uploads (that may be due to a broken android emulator, have to to test this, though)

long story short: im not done here ;)

jancborchardt:
cc @schiessle regarding the licensing question. :) @newhinton to make it easier for him, could you point out the specific commit or part of code you took, and the Stackoverflow thread?

AndyScherzinger:
I'll try from a not to technical point of view

- if it is a simple pattern matching than there is no need for using the code from stackoverflow
- I'd think that (not having looked into the code) that the pattern matching needs to take place before a file is added to the upload queue and thus there should then be no entry in the upload screen and also no notification for the file.

newhinton:
@schiessle 

//https://stackoverflow.com/questions/1247772/is-there-an-equivalent-of-java-util-regex-for-glob-type-patterns
    private static String createRegexFromGlob(String glob)
    {
        String out = "^";
        for(int i = 0; i < glob.length(); ++i)
        {
            final char c = glob.charAt(i);
            switch(c)
            {
                case '*': out += ".*"; break;
                case '?': out += '.'; break;
                case '.': out += "\\."; break;
                case '\\': out += "\\\\"; break;
                default: out += c; break;
            }
        }
        out += '$';
        return out;
    }

in src/main/java/com/owncloud/android/operations/UploadFileOperation.java

newhinton:
@AndyScherzinger  At the moment, i check the file in the block where it also checks if the upload would work, if it is locked or if the password&username is wrong. Logically it belongs there, but i will check if i can prevent it earlier

AndyScherzinger:
> Logically it belongs there, but i will check if i can prevent it earlier

I'd disagree respectively. The other checks are regarding the upload itself and if that one is possible. 
While your check would be if a file is even to be considered for an upload action thus that should happen before. 
While yes it could also happen where you put it now while that would then kind of being an input check for a IllegalArgumentException while using exceptions for control flows would be bad.

newhinton:
You are right, i will take a further look and try to capture it earlier :D

newhinton:
@jancborchardt You mentioned open source sync applications, can you name some? I am seemi gly unable to find any of them, at least not on fdroid ^^ (I hoped that i could replace my propietary solution with nextcloud because i couldn't find any open source ones :D)

src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -756,9 +765,51 @@ private RemoteOperationResult checkConditions(File originalFile) {
...
+    private boolean isBlacklistedFile(String fileName) {
+
+        SharedPreferences sharedpreferences = android.preference.PreferenceManager.getDefaultSharedPreferences(this.getContext());
+        List<String> list = new ArrayList<>(sharedpreferences.getStringSet(AppPreferencesImpl.EXCLUDED_AUTOUPLOAD_PATTEN_KEY, AppPreferencesImpl.EXCLUDED_AUTOUPLOAD_PATTEN_DEFAULT_VALUES));
+
+        boolean isBlacklisted = false;
+        for (String pattern : list) {
+            if (fileName.matches(createRegexFromGlob(pattern))) {
+                Log_OC.d(TAG, "Blacklisted file: '" + fileName + "' matched with pattern :'" + pattern + "'");
+                isBlacklisted = true;
```
bertwesarg: No need to check more patterns, if one matched

ezaquarii: Can we extract this into a predicate implementing `Function1<String, Bool>`? This should be tested as well.

src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -756,9 +765,51 @@ private RemoteOperationResult checkConditions(File originalFile) {
...
+            remoteOperationResult =  new RemoteOperationResult(ResultCode.UNAUTHORIZED);
+        }
+
         return remoteOperationResult;
     }
 
+    private boolean isBlacklistedFile(String fileName) {
+
+        SharedPreferences sharedpreferences = android.preference.PreferenceManager.getDefaultSharedPreferences(this.getContext());
+        List<String> list = new ArrayList<>(sharedpreferences.getStringSet(AppPreferencesImpl.EXCLUDED_AUTOUPLOAD_PATTEN_KEY, AppPreferencesImpl.EXCLUDED_AUTOUPLOAD_PATTEN_DEFAULT_VALUES));
```
ezaquarii: This is a private implementation detail. 

Please inject `AppPreferences` and do not use shared preferences directly.

src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -34,6 +35,7 @@
 import com.google.gson.reflect.TypeToken;
 import com.nextcloud.client.device.PowerManagementService;
 import com.nextcloud.client.network.ConnectivityService;
+import com.nextcloud.client.preferences.AppPreferencesImpl;
```
ezaquarii: This should go away.

src/main/java/com/owncloud/android/ui/dialog/PatternBlacklistEditorDialogFragment.java
```diff
@@ -0,0 +1,101 @@
...
+    private PatternBlacklistEditorDialogListFragment adapter;
+
+    @Inject AppPreferences preferences;
+
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        // keep the state of the fragment on configuration changes
+        setRetainInstance(true);
+
```
ezaquarii: Please adjust the formatting.

src/main/java/com/owncloud/android/ui/dialog/PatternBlacklistEditorDialogFragment.java
```diff
@@ -0,0 +1,101 @@
...
+        mView.findViewById(R.id.add_pattern_button).setOnClickListener(
+            new View.OnClickListener() {
+                @Override
+                public void onClick(View v) {
+                    adapter.addNewPattern(et.getText().toString());
+                }
+            });
+
+        mView.findViewById(R.id.cancel).setOnClickListener(
+            new View.OnClickListener() {
```
ezaquarii: This can be replaced with lambdas.

src/main/java/com/owncloud/android/ui/dialog/PatternBlacklistEditorDialogFragment.java
```diff
@@ -0,0 +1,101 @@
...
+                @Override
+                public void onClick(View v) {
+                   close();
+                }
+            });
+
+        mView.findViewById(R.id.save).setOnClickListener(
```
ezaquarii: lambda as well

src/main/java/com/owncloud/android/ui/dialog/PatternBlacklistEditorDialogFragment.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    private void close() {
+        this.dismiss();
+    }
+
+    public static PatternBlacklistEditorDialogFragment newInstance(int title) {
+        PatternBlacklistEditorDialogFragment frag = new PatternBlacklistEditorDialogFragment();
+        Bundle args = new Bundle();
+        args.putInt("title", title);
```
ezaquarii: No magic strings please. What is title?

src/main/java/com/owncloud/android/ui/dialog/PatternBlacklistEditorDialogListFragment.java
```diff
@@ -0,0 +1,81 @@
...
+
+        if (view == null) {
+            view = LayoutInflater.from(getContext()).inflate(R.layout.pattern_blacklist_editor_list_item, parent, false);
+        }
+
+        TextView label = view.findViewById(R.id.pattern_name);
+        ImageButton ib = view.findViewById(R.id.delete_pattern_button);
+
+        ib.setOnClickListener(
+            new View.OnClickListener() {
```
ezaquarii: lambda

src/main/java/com/owncloud/android/ui/dialog/PatternBlacklistEditorDialogListFragment.java
```diff
@@ -0,0 +1,81 @@
...
+
+import com.owncloud.android.R;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import androidx.annotation.NonNull;
+import lombok.Getter;
+
+public class PatternBlacklistEditorDialogListFragment extends ArrayAdapter<String> {
```
ezaquarii: This is de facto legacy API. Can we use modern `RecyclerView` for this?
Also, your data will be subject to UI lifecycle issues, which is difficult to maintain. Moving to `ViewModel` resolves those issues.

src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
```diff
@@ -72,11 +79,20 @@
     private static final String PREF__LOCK = SettingsActivity.PREFERENCE_LOCK;
     private static final String PREF__SELECTED_ACCOUNT_NAME = "select_oc_account";
     private static final String PREF__MIGRATED_USER_ID = "migrated_user_id";
+    public static final String EXCLUDED_AUTOUPLOAD_PATTEN_KEY = "EXCLUDED_AUTOUPLOAD_PATTEN_KEY";
```
ezaquarii: Those should be private.

src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
```diff
@@ -35,6 +35,13 @@
 import com.owncloud.android.ui.activity.SettingsActivity;
 import com.owncloud.android.utils.FileSortOrder;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import static com.owncloud.android.ui.activity.SettingsActivity.PreferenceKeys.EXCLUDED_AUTOUPLOAD_PATTEN_DEFAULT_VALUES;
```
ezaquarii: Preferences implementation should not depend on UI. This should be removed.

src/main/java/com/owncloud/android/operations/UploadFileOperation.java
AndyScherzinger: We need to put all this some place else since the blocking shouldn't happen within the upload operation but before even triggering it during the discovery of potentially, to be uploaded files!

src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
```diff
@@ -35,6 +35,13 @@
...
 import static com.owncloud.android.ui.fragment.OCFileListFragment.FOLDER_LAYOUT_LIST;
```
ezaquarii: Hm... I'm wondering how this one got here... it doesn not belong here.

src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
```diff
@@ -72,11 +79,20 @@
     private static final String PREF__LOCK = SettingsActivity.PREFERENCE_LOCK;
     private static final String PREF__SELECTED_ACCOUNT_NAME = "select_oc_account";
     private static final String PREF__MIGRATED_USER_ID = "migrated_user_id";
+    public static final String EXCLUDED_AUTOUPLOAD_PATTEN_KEY = "EXCLUDED_AUTOUPLOAD_PATTEN_KEY";
+    public static final Set<String> EXCLUDED_AUTOUPLOAD_PATTEN_DEFAULT_VALUES = new HashSet<>();
```
ezaquarii: private too

src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -24,6 +24,7 @@
 import android.accounts.Account;
 import android.annotation.SuppressLint;
 import android.content.Context;
+import android.content.SharedPreferences;
```
ezaquarii: Naked shared preferences should not be used - we have this responsibility already encapsulated in `AppPreferences`.

src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -228,6 +235,8 @@ public UploadFileOperation(UploadsStorageManager uploadsStorageManager,
         // Ignore power save mode only if user explicitly created this upload
         mIgnoringPowerSaveMode = mCreatedBy == CREATED_BY_USER;
         mFolderUnlockToken = upload.getFolderUnlockToken();
+
+        excludedFilesyncPatterns.add(".thumbdata*");
```
ezaquarii: I see this is duplicated in preferences class. There should be one place responsible for default values.

src/main/java/com/owncloud/android/ui/dialog/PatternBlacklistEditorDialogFragment.java
```diff
@@ -0,0 +1,101 @@
...
+
+        List<String> list = preferences.getAutoUploadPatternBlackList();
+
+        adapter = new PatternBlacklistEditorDialogListFragment(this.getContext(), list);
+        scrollViewContainer.setAdapter(adapter);
+
+
+
+        EditText et = mView.findViewById(R.id.add_new_edit_text);
+        mView.findViewById(R.id.add_pattern_button).setOnClickListener(
```
ezaquarii: lambdas

src/main/java/com/owncloud/android/ui/dialog/PatternBlacklistEditorDialogListFragment.java
AndyScherzinger: Probably won't need this since the decision has been to implement a static (fixed) list of patterns that are blocked from upload so no UI changes at all and a skip mechanism during detection of files.

th122:
I would be thrilled to see this going into production. I am using the Android Nextcloud client to upload logfiles to the server and, as of current, it keeps uploading the uncompressed daily log along with the archived compressed log files.

The quickly growing active log (up to 200m at the end of the day) makes for a rather high traffic load which I'd rather prevent by applying a filter. (On most occasions, the upload fails anyway due to the log file being appended to while the upload was under way)

As usual with log files, there's a lot of repetitive content in it, so at the end of the day it compresses fairly well from around 200MB to 10MB. Stream compression, which I assume the client will be using, is somewhat less efficient at detecting the larger repetitive blocks of text.

I'm completely content with a file-based solution such as sync-exclude.lst, but couldn't find anything like that by searching the sources.

AndyScherzinger:
@tobiasKaminsky @jancborchardt @ezaquarii what do you think? We could extend the list and add such a mechanism like the desktop client offers, see hint to sync-exclude.lst while I can't say how this could then be made editable but such a mechanism would bring us en par with the desktop and would lead to a similar behavior when it comes to ignoring files.

newhinton:
@AndyScherzinger

Isn't that what i wanted to do in the first place? Jan said that from a users perspective this would be bad, because making everything configurable leads to complicated ui's which are often not needed, at least i understood that from his arguments.

AndyScherzinger:
@newhinton actually I don't know for sure. My last comment was about adopting the desktop client mechanism which I wasn't aware of existed :/

Still I don't have an idea on how to handle such a file from a user perspective.

tobiasKaminsky:
Hm. Tricky, as this whole upload system evolves more and more. First, it was only upload image/video, then there was possibility to use any folder as upload source, which then generated demand of excluding files.
Additionally there is the idea to have a real two way sync…

I currently tend to have this auto upload feature as simple as possible, that means that if there is a custom folder, all files will be uploaded. 

If many vote for having this in, then it needs to be done in a good and user friendly way
- have a configurable list
- clear instructions how entry is defined (reg ex)
- reg ex check prior saving
- some basic sanity checks

th122:
I gave the 10935 QA Version linked above a try. 
Setting the pattern to ignore works like a charm, and the undesired file doesn't get uploaded either. Great so far! 
However.... (and this may be unrelated) every time the client decides that it would like to upload the file, it throws an error notification at the user claiming that authorization had failed.

Opening the app indeed asks for re-authorization. Looking at the nexctcloud browser interface from another device shows that the last successful upload has been some hours ago - and soon after the next (desired) upload happens without complaint.

tobiasKaminsky:
> However.... (and this may be unrelated) every time the client decides that it would like to upload the file, it throws an error notification at the user claiming that authorization had failed.
> 
> Opening the app indeed asks for re-authorization. Looking at the nexctcloud browser interface from another device shows that the last successful upload has been some hours ago - and soon after the next (desired) upload happens without complaint.

I guess this is unrelated, please open up a new issue.

jancborchardt:
Screenshots would be very useful for design review. :)

Also, the language should be "Allowlist / Blocklist" instead of "blacklist" as it is more understandable.

m-prol:
The last comments were about testing. Is anything checked in mainline yet?
Will this feature be available in the near future?
I'm looking for a feature to exclude the e.g. /tmp-Folder, when Auto-Uploading e.g. the whole Whatsapp-Folder. 
Thanx

Siggi0904:
+1 for adding these long missing feature.
Should have the same configuration settings like desktop app.

Siggi0904:
Hi, around 3 years of these feature request.
Now the changes/added lines are older the current branch.

Please check these and add to main tree for next version.
Should be merged with question about blacklist folders in instant upload.

Thanks.

PR #3515, opened 2019-01-29
Do not open ics within our app

### LINKED ISSUE #3417: .ics should not be opened as text file
### Actual behaviour
After clicking on an .ics file its content is being presented as the raw text.

### Expected behaviour
The app should let the user decide, if he wants to open the file with a fitting app and not show the text (preview).
It should become the same behaviour like when opening e.g. a `.mp3` or `.pdf`.

### Reason why
Viewing the text source of the calendar file is an uncommon use-case.
Everybody will try to view or import their calendar files with an appropriate app.
So many users need to click on the file in order to get it downloaded first, because else the "`open with`" dialog wouldn't be available.
After that you need to tap "back", to get out of the text preview. Finally you can click the three dots to open the context menu and open the "open with dialog" and use your calendar app.
### Environment data
Android version: 8.0.0
Nextcloud app version: 3.4.1

tobiasKaminsky:
Fix #3417 
ICS are calendar files, and while they are technically text files, it does not make sense to open them internally, as there is a calendar app which handles it way better.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
+++ b/src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
@@ -477,6 +477,7 @@ public class PreviewTextFragment extends FileFragment implements SearchView.OnQu
         final List<String> unsupportedTypes = new LinkedList<>();
         unsupportedTypes.add("text/richtext");
         unsupportedTypes.add("text/rtf");
+        unsupportedTypes.add("text/calendar");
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
👍 Nice + I'd vote for backport :)

tobiasKaminsky:
> I'd vote for backport

Then let us put it int RC2, because I want to get RC1 out of the door today, but CI is so overloaded that this will take ages…

PR #3521, opened 2019-01-30
Direct upload fix

tobiasKaminsky:
Fixes #3523

- add permission check for direct upload
- use yyyy instead of Y for date building


tobiasKaminsky:
> @tobiasKaminsky so which solution is "better"`--> #3523

Combination of both:
I'll add permission check like you did (move into FileOperationHelper as it is then always checked), but keep this PR (as it contains another fix).

AndyScherzinger:
> Combination of both:
I'll add permission check like you did (move into FileOperationHelper as it is then always checked), but keep this PR (as it contains another fix).

👍 perfect!

PR #3537, opened 2019-01-31
enhance notification buttons

### LINKED ISSUE #3533: [3.5.0 RC1] Actions in notifications look weird
screenshot_nextcloud_20190130-202930

Those action buttons are styled in similar way to the other buttons. Only I feel there is some significant spacing missing here ;)

tobiasKaminsky:
Fix #3533 

This is a possible fix, but as soon as I add padding, it looks strange again, as the height is not computed correctly.


mario:
Still looks like the buttons could be bigger?

tobiasKaminsky:
> Still looks like the buttons could be bigger?

Yeah, but it is set to "wrap_content" and if I add more padding, which normally would increase the button, I do get the image on the right:

 

Increasing height of the container of the buttons (LinearLayout) to e.g. 60dp also does not help. Then there is just a giant gap, but the button looks like the same.

Manually hardcoding width and height seems to help:


But is this something we want? There can be any text in any language, so this should be auto-computed…

tobiasKaminsky:
Added black border to second button:


Only downside is still, that width/height are hardcoded :/

tobiasKaminsky:
With longer text this then looks like:

mario:
Could you try LinearLayout.LayoutParams instead of ViewGroup? that should do it :)

mario:
@tobiasKaminsky in the code it looks like you added margin, and not padding?

AndyScherzinger:
Strange. Border color on the secondary action should be grey

mario:
Like I said on IRC (but to repeat it here) padding of 5 (like @tobiasKaminsky mentioned he set) is 5px. I would assume that is less than the default button padding which is why the button size gets smaller.

tobiasKaminsky:
> Strange. Border color on the secondary action should be grey

I set this via code to black. Which grey should I use?
standard_grey: #757575
grey_200: #EEEEEE

tobiasKaminsky:
Findbugs now thinks there is a circular dependency between NotificationsActivity and NotificationListAdapter, but those changes should not be the cause…

AndyScherzinger:
for the grey I don't know, would need to color pick 757575/grey600 seems to be to dark so yeah eith er EEEEEE or E0E0E0

tobiasKaminsky:
So this is fine and can go into RC2?

tobiasKaminsky:
Text bold
grey set to #EEEEEE
text color on secondary button is primary color

tobiasKaminsky:
To have this in RC2, I'll merge this now.
For RC3 or even final we can do minor UI enhancements.

AndyScherzinger:
Looks okay to me, maybe some really minor things. @tobiasKaminsky what do you mean by cancel looks different?

tobiasKaminsky:
In here https://github.com/nextcloud/android/pull/3537#issuecomment-459667406, the spacing between characters of "Cancel" have different spacing.
I even think it is a different font…?

AndyScherzinger:
Hard to tell on the smartphone, might be a different font, definitely different letter spacing, yes

PR #3563, opened 2019-02-04
Use lombok where fit

AndyScherzinger:
Solving some codacy issues (undocumented empty constructors) by making use of lombok for the contrusctors (and getter/setter).


tobiasKaminsky:
2 issues on Codacy: seems that variables need to be private.

AndyScherzinger:
Argh, I accidentially pushed to master :/

AndyScherzinger:
I fixed codacy though, be completely removing the class (not needs due to its using class not being needed...)

PR #3567, opened 2019-02-05
sort files after set/unset favorite

### LINKED ISSUE #2522: Set/unset favorite should refresh file sorting
When setting/unsetting a file as favorite, it should be sorted to top / in correct sorting.
Currently this is done only after a manual refresh.

With new recycler view there is maybe a way to only move the one file/folder in question.

tobiasKaminsky:
Fix #2522 


AndyScherzinger:
👍

PR #3578, opened 2019-02-06
Use getter instead of direct access of Activity

tobiasKaminsky:


AndyScherzinger:
👍 looks good, needs to be set back to lib master as soon as lib PR has been merged.

PR #3587, opened 2019-02-07
Show only media folders that exist

### LINKED ISSUE #3555: Auto Upload doesn't forget old directories
### Actual behaviour
when directories found once by Auto Upload disappear, they are not removed from the list of available media directories _and_ they show "phantom" content from other directories.

### Expected behaviour
Non-existing directories should disappear from the list.
 
### Steps to reproduce
1. create a new directory with photos
2. wait until Auto Upload has discovered it (e.g. restart Android or wait a long time)
3. remove the directory

### Environment data
Android version: 8.1 (LineageOS)

Device model: bq Aquaris Pro X

Stock or customized system: LineageOS

Nextcloud app version: 3.4.2 (FDroid)

Nextcloud server version: 15.0.0

#### Web server error log

n/a

n/a (I think, let me know if you need it)

tobiasKaminsky:
Fix #3555 

### DIFF
--- a/src/main/java/com/owncloud/android/datamodel/MediaProvider.java
+++ b/src/main/java/com/owncloud/android/datamodel/MediaProvider.java
@@ -118,8 +118,8 @@ public final class MediaProvider {
                         filePath = cursorImages.getString(cursorImages.getColumnIndexOrThrow(
                                 MediaStore.MediaColumns.DATA));
 
-                        // check if valid path
-                        if (filePath != null && filePath.lastIndexOf('/') > 0) {
+                        // check if valid path and file exists
+                        if (filePath != null && filePath.lastIndexOf('/') > 0 && new File(filePath).exists()) {
                             mediaFolder.filePaths.add(filePath);
                             mediaFolder.absolutePath = filePath.substring(0, filePath.lastIndexOf('/'));
                         }
--- a/src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
@@ -226,7 +226,13 @@ public class SyncedFoldersActivity extends FileActivity implements SyncedFolderA
         Account currentAccount = AccountUtils.getCurrentOwnCloudAccount(this);
         for (SyncedFolder syncedFolder : syncedFolderArrayList) {
             if (currentAccount != null && syncedFolder.getAccount().equals(currentAccount.name)) {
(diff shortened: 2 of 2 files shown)


AndyScherzinger:
@tobiasKaminsky would have to test this but I always thought that our notification trigger for a newly found folder will also write a db entry into our synced folders table which will then lead to the folder still being shown even after the deletion! So I think this fix won't help but I might be mistaken... :/

tobiasKaminsky:
You are partly right:
For non-enabled folders this fix is working.

But:
- setup a folder as AutoUpload
- delete it
- see 

Shall we silently remove them? What if this is just a folder on a removable sdcard and will re-appear later?
I would rather keep it but just show a proper thumbnail/default thumbnail?

tobiasKaminsky:
As this is persisted in database, this also survives restarts, @AndyScherzinger

AndyScherzinger:
> Shall we silently remove them? What if this is just a folder on a removable sdcard and will re-appear later?
I would rather keep it but just show a proper thumbnail/default thumbnail?

I guess to be discussed with @nextcloud/designers. From a technical point of view it definitely needs to be removed for several reasons: It clutters the ist over time since it'll have more and more dead/non-existing folders over time so they shouldn't be shown when not present (anymore). Talking about deleting the entries complpetely: you need to do a file check per db entries if still present on the file system every time you load this list and that is procecssing time not well spent and will also grow over time. So to me it leaves us with the decision on how to handle such entries for SD cards where you still can have the same scenario: folders that don't exist anymore. So even there I am fine with deleting such db entries. Entries that shouldn't be deleted (right away) are the one that are active (even though the folder doesn't ecist anymore). The would then be deleted after de-activation and reentrering the auto upload screen at a later point in time.

Just my 2 cents :)

tobiasKaminsky:
So to summarize:
- delete non-enabled folders
- keep enabled folders

even if the local folder is currently not available/deleted on purpose.

tobiasKaminsky:
Added :)

AndyScherzinger:
Correct while I'd say we still have an edge case :/

A folder with images that get created and deleted regularly will now work fine from a auto upload screen perspective but will now lead to a new folder notification in case the auto upload screen has been opened in between. 
So maybe we need to keep the DB entries but flag them to then prevent the notifications for that folder? Other than that I think the notification implementation is flaws since it should maintain its own list of folder for which the app launches a notification so that these two things operate indipendently...

tobiasKaminsky:
As discussed on IRC, we currently keep this approach, but if problems arises, we will use the new approach suggested by @AndyScherzinger

PR #3589, opened 2019-02-08
Check if app is excluded from battery optimization

### LINKED ISSUE #3553: Photo upload runs only when charging
Auto upload of photos works only when charging despite the settings (not checked) and the granted exception from battery-saving (phone). When charging starts, the auto upload starts immediately.

Related to old #300 and #355. 
3.5.0 RC1, I do not have access to server logs. 
Sony Xperia X Compact, Android 8. 0

tobiasKaminsky:
Fix #3553 

To test:
either add "google" (for emulator") in PowerManagementService -> OVERLY_AGGRESSIVE_POWER_SAVING_VENDORS
or use on of the problematic devices.


tobiasKaminsky:
First I would get this tested by @kotecky

mario:
So ...

1) This will work only sometimes, on certain phones. Xiaomi, Huawei and a few other phones have their own battery ignore mechanisms.

2) Battery ignore is a COMPLETELY different functionality from Power saving. I understand the use-case from the issue that lead to this PR, but this will wreak havoc on the phone and the battery with unintended side-effects, depending on the fact if the phone can be shut down on its own prior to the battery being completely exhausted.

tobiasKaminsky:
https://developer.android.com/reference/android/os/PowerManager.html#isIgnoringBatteryOptimizations(java.lang.String)

This is a stable and public API from Android.
Even if other vendors do use their own battery optimization, this should not hurt, as this then might just return always false.

AndyScherzinger:
@tobiasKaminsky The point seems to be that there is "battery optimization" and "power save mode", so during power save mode we should not trigger austo uploads uploads even when the app in whitelisted for battery saving (which is a different thing). Because now, if the app is whitelisted for battery optimization and the user or system activated the battery saviing mode since the battery is low we would do uploads draining the battery.

Gaeldrin:
@AndyScherzinger @mario I want to express my support for this PR as an user affected by this problem. Make it a setting if you really don't want this to be default behaviour but I believe you should enable it in some form as there are many people suffering from this (I have found at least 3 different issues on this without much looking).

I have had some power saving mode tweaked and turned on on every phone in past 5 years and my user experience was great. Now I am on S7 and uploading is completely disabled for me as I have a "Medium power saving mode" (the weakest form) enabled. Even when charging and triggering the upload manually in open app, it responds with "Waiting on termination of Power saving mode". This basically forces me to change the settings of my whole device because a single app does not work under it, and I think that is very wrong. Certainly more wrong than draining battery when user made the choice to disable battery optimizations. Now I have about a month's worth of photos and videos that never got uploaded and the only way to upload them is to change the CPU speed, screen resolution, background connections and probably more. There is no way this is correct behaviour. Moreover, this behaviour is inconsistent with every other Nextcloud feature as my contacts, position, sms.. get backed up every day.

tobiasKaminsky:
I would be fine for an option for special devices, like Samsung.
What do you think?

AndyScherzinger:
We can give it a try, sure

tobiasKaminsky:
> I would be fine for an option for special devices, like Samsung.
> What do you think?

I'll give auto upload a bit more love :heart: 
What do you think, where should I put this option (it is only shown for special vendors):
- in auto upload in three dot menu
- in settings with a new group "upload"?

tobiasKaminsky:
I changed this a bit:
On special devices (samsung, huawei, xiaomi) there is now an option to disable power check.

Wording of dialog can be enhanced.

src/main/java/com/owncloud/android/utils/PowerUtils.java
```diff
@@ -17,6 +20,12 @@ private PowerUtils() {
      */
     @TargetApi(Build.VERSION_CODES.LOLLIPOP)
     public static boolean isPowerSaveMode(Context context) {
+        AppPreferences preferences = AppPreferencesImpl.fromContext(context);
```
ezaquarii: I think it woudn't be a bad idea to convert this into a `PowerManagementService` or something like this. Access to `AppPreferencesImpl` should be generally exclude outside preferences package, as it leaks impl detail into a 3rd component.

It is also challenging to unit-test the code this way, as I'd need to mock internal `Context` details.

If you don't plan to refactor it yet, passing `AppPreferences` will do the trick for now. We can have another sweep as a separate ticket.

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
```diff
@@ -171,15 +172,46 @@ protected void onCreate(Bundle savedInstanceState) {
         if (ThemeUtils.themingEnabled(this)) {
             setTheme(R.style.FallbackThemingTheme);
         }
+
+        isSpecialManufacturer = SPECIAL_MANUFACTURER.contains(Build.MANUFACTURER.toLowerCase(Locale.ROOT));
```
ezaquarii: I think that this piece of logic should go to `PowerUtils`/`PowerManagerService`. Here UI is not only concerned about a phone model it runs on, it will behave differencly depending on test environment.

src/main/java/com/owncloud/android/utils/PowerUtils.java
tobiasKaminsky: Context is also used in line 30 to get PowerManager…
So passing AppPrefences does not help, but context is still needed.

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
tobiasKaminsky: Yeah :/ Currently there is no real distinction between UI, controller, etc :/
I plan to migrate this step by step…

If we move this to PowerUtils then this would become a static method for now and we can refactor it later, together with the above part?

ezaquarii: Yes, we can do it this way. Moving those statics is easy.

ezaquarii:
This can benefit from #4073 PR

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
ezaquarii: Since `PowerUtils` are single-method, I decided to bite the bullet. #4073

tobiasKaminsky:
This is in DeviceModule
https://github.com/nextcloud/android/pull/3589/files#diff-498ea5de527e47aeaf8fedc276c913e1R29

This is in AppModule
https://github.com/nextcloud/android/pull/3589/files#diff-5358f7de48f9b7bd08951b0c712858baR113

When I have it in DeviceModule Dagger complains about missing Factory class…

@ezaquarii do you have an idea?

src/main/java/com/nextcloud/client/device/PowerManagementServiceImpl.kt
```diff
@@ -36,10 +40,22 @@ internal class PowerManagementServiceImpl(
...
+            val powerManager = context.getSystemService(Context.POWER_SERVICE) as PowerManager
+            val preferences = AppPreferencesImpl.fromContext(context)
+
+            return PowerManagementServiceImpl(powerManager, DeviceInfo(), preferences)
+        }
     }
 
     override val isPowerSavingEnabled: Boolean
         get() {
+            if (preferences.isPowerCheckDisabled) {
```
ezaquarii: I think it is not covered by unit test yet.

src/main/java/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -982,16 +987,18 @@ public void onTransferProgress(long progressRate, long totalTransferredSoFar,
...
                         cancel(mCurrentUpload.getAccount().name, mCurrentUpload.getFile().getRemotePath()
                                 , ResultCode.DELAYED_FOR_WIFI);
                     } else if (mCurrentUpload.isChargingRequired() &&
-                            !Device.getBatteryStatus(MainApp.getAppContext()).isCharging()) {
+                        !Device.getBatteryStatus(context).isCharging()) {
                         cancel(mCurrentUpload.getAccount().name, mCurrentUpload.getFile().getRemotePath()
                                 , ResultCode.DELAYED_FOR_CHARGING);
                     } else if (!mCurrentUpload.isIgnoringPowerSaveMode() &&
-                            PowerUtils.isPowerSaveMode(MainApp.getAppContext())) {
+                        PowerManagementServiceImpl.fromContext(context).isPowerSavingEnabled()) {
```
ezaquarii: Would it be possible to inject this one too?

src/main/java/com/nextcloud/client/device/PowerManagementServiceImpl.kt
```diff
@@ -22,12 +22,16 @@
...
+import android.content.Context
 import android.os.Build
 import android.os.PowerManager
+import com.nextcloud.client.preferences.AppPreferences
+import com.nextcloud.client.preferences.AppPreferencesImpl
 
 internal class PowerManagementServiceImpl(
     private val powerManager: PowerManager,
-    private val deviceInfo: DeviceInfo = DeviceInfo()
+    private val deviceInfo: DeviceInfo = DeviceInfo(),
```
ezaquarii: Arguments with default value should go last or otherwise you need to provide all of them.

ezaquarii: The constructor change should cause `TestPowerManagementService` build failure. Surprisingly, the CI does not complain.

Are we running all tests?

src/main/java/com/owncloud/android/files/services/FileUploader.java
tobiasKaminsky: I failed to do so:
- added @Inject
        PowerManagementService powerManagementService; to FileUploaderBinder (which extends Binder)
- added FileUploaderBinder to ComponentsModule

But I do get a NPE

src/main/java/com/nextcloud/client/device/PowerManagementServiceImpl.kt
tobiasKaminsky: Test on CI failed due to timeout, so maybe this never got to this point.
I restarted test.

src/main/java/com/owncloud/android/files/services/FileUploader.java
tobiasKaminsky: I fear that this somehow the same problem as in #4076.

ezaquarii:
Binders, providers, broadcast receivers and services do not have lifecycle listeners so there is no automatic injection on creation - you need to manualy call AndroidInjection.inject(this) somewhere.

Also, sometimes it is not practical to inject using Dagger and plain parameters injection is enough. I did it quite a lot in utility helper functions.

src/main/java/com/nextcloud/client/device/PowerManagementServiceImpl.kt
tobiasKaminsky: I'll add one.

tobiasKaminsky:
> you need to manualy call AndroidInjection.inject(this) somewhere.

There is no function for Binder, so I keep it for now with this and if you have time/want to fiddle around, maybe you can have look.

tobiasKaminsky:
> > you need to manualy call AndroidInjection.inject(this) somewhere.
> 
> There is no function for Binder, so I keep it for now with this and if you have time/want to fiddle around, maybe you can have look.

As FileUploaderBinder is a class within FileUploader it can access powerManagementService from FileUploader, so nothing to do here…(silly me)

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
```diff
@@ -177,9 +178,38 @@ protected void onCreate(Bundle savedInstanceState) {
...
         inflater.inflate(R.menu.synced_folders_menu, menu);
+
+        if (powerManagementService.isPowerSavingExclusionAvailable()) {
+            MenuItem item = menu.findItem(R.id.action_disable_power_save_check);
+            item.setVisible(true);
+
+            item.setChecked(preferences.isPowerCheckDisabled());
+
+            item.setOnMenuItemClickListener(powerCheck -> {
+                if (!powerCheck.isChecked()) {
```
ezaquarii: How about extracting this lambda to a method `onDisablePowerSaveCheckClicked()`? We could avoid mixing `onCreate()` initialization code with business logic.

src/main/java/com/owncloud/android/jobs/NContentObserverJob.java
```diff
@@ -65,9 +70,14 @@ public boolean onStartJob(JobParameters params) {
     }
 
     private void checkAndStartFileSyncJob() {
-        if (!PowerUtils.isPowerSaveMode(getApplicationContext()) &&
-                new SyncedFolderProvider(getContentResolver(),
-                                         AppPreferencesImpl.fromContext(getApplicationContext())).countEnabledSyncedFolders() > 0) {
+        AppPreferences appPreferences = AppPreferencesImpl.fromContext(getApplicationContext());
```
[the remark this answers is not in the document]
tobiasKaminsky: NContentObserverJob is not created/accessed in NCJobCreator.
NContentObserverJob extends JobService
while the other
ContactsBackupJob extends Job

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
```diff
@@ -707,10 +737,8 @@ protected void onResume() {
     }
 
     private void showBatteryOptimizationInfo() {
-
-        boolean isSpecialManufacturer = SPECIAL_MANUFACTURER.contains(Build.MANUFACTURER.toLowerCase(Locale.ROOT));
-
-        if (isSpecialManufacturer && checkIfBatteryOptimizationEnabled() || checkIfBatteryOptimizationEnabled()) {
+        if (powerManagementService.isPowerSavingExclusionAvailable()
+            && checkIfBatteryOptimizationEnabled() || checkIfBatteryOptimizationEnabled()) {
```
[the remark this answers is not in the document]
tobiasKaminsky: Right, I did it wrong back then.
I want to show the info for battery optimization in these cases
- if battery optimization is enabled (simple)
- if power saving exclusion is available (e.g. Huawai) AND battery optimization is in any state, as a false value might be returned.

src/test/java/com/nextcloud/client/device/TestPowerManagementService.kt
```diff
@@ -101,4 +119,20 @@ class TestPowerManagementService {
         whenever(deviceInfo.vendor).thenReturn("some_other_nice_vendor")
         assertFalse(powerManagementService.isPowerSavingExclusionAvailable)
     }
+
+    @Test
+    fun isPowerSaveMode_assertCorrectlyReportsTrue() {
+        whenever(deviceInfo.apiLevel).thenReturn(Build.VERSION_CODES.O)
+        whenever(powerManagementService.isPowerSavingEnabled).thenReturn(true)
+        assertTrue("Incorrectly reported power saving mode on",
+            powerManagementService.isPowerSavingEnabled)
```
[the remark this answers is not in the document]
tobiasKaminsky: Right, this is somehow useless.

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
tobiasKaminsky: :+1: 
At some point we should take our time an split this all up into model, view, controller…

tobiasKaminsky:
@ezaquarii once you approve, I can get this in and do 3.7.0 RC1 :-)

ezaquarii:
> @ezaquarii once you approve, I can get this in and do 3.7.0 RC1 :-)

give me 5

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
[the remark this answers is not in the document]
ezaquarii: So it looks for me like you want this:
```
if (powerManagementService.isPowerSavingExclusionAvailable() || checkIfBatteryOptimizationEnabled())
```

as the current condition depends solely on `checkIfBatteryOptimizationEnabled()` and `isPowerSavingExclusionAvailable()` has no influence.

AndyScherzinger:
@tobiasKaminsky looking at the screenshot, the OK button on the dialog should be primary colored not grey. :)

tobiasKaminsky:
Backport is in #4120.
I'll resolve conflicts later and merge it then.
Priority has now 3.7.0 RC1.

tobiasKaminsky:
@AndyScherzinger @ezaquarii this is now rebased and ready for review again

AndyScherzinger:
@tobiasKaminsky rebase 'n' merge`?

tobiasKaminsky:
I would like to have a final review round by you and @ezaquarii.

src/main/res/values/strings.xml
```diff
@@ -884,10 +884,11 @@
     <string name="copy_internal_link">Copy internal link</string>
     <string name="copy_internal_link_subline">Only works for users with access to this folder</string>
     <string name="failed_to_download">Failed to pass file to download manager</string>
-
+    <string name="autoupload_disable_power_save_check">Disable power save check</string>
+    <string name="power_check_dialog_title">Disable power save check</string>
```
AndyScherzinger: duplicate string to translate should be just one string that is used in both places.

AndyScherzinger:
Just a minor issue with a duplicated string (while yes, the string could be changed later on but it should then be split into 2 distinct strings, not now :)

src/main/java/com/nextcloud/client/device/DeviceModule.java
```diff
@@ -19,23 +19,22 @@
...
 
-@Module
-class DeviceModule {
+import com.nextcloud.client.preferences.AppPreferences;
+
+import dagger.Module;
+import dagger.Provides;
 
+@Module
+public class DeviceModule {
```
ezaquarii: Is there a reason we reverted this to Java?

src/test/java/com/nextcloud/client/device/TestPowerManagementService.kt
```diff
@@ -120,6 +125,23 @@ class TestPowerManagementService {
...
+            //      power check is overridden by user
+            whenever(preferences.isPowerCheckDisabled).thenReturn(true)
+            whenever(platformPowerManager.isPowerSaveMode).thenReturn(true)
+
+            // WHEN
+            //      power save mode is checked
+            // THEN
+            //      power save mode is enabled
+            //      power saving is disabled
+            assertTrue(platformPowerManager.isPowerSaveMode)
```
ezaquarii: This test seems to be checking values returned by mock, which has little sense.

src/test/java/com/nextcloud/client/device/TestPowerManagementService.kt
```diff
@@ -61,13 +62,17 @@ class TestPowerManagementService {
 
         internal lateinit var powerManagementService: PowerManagementServiceImpl
 
+        @Mock
+        lateinit var preferences: AppPreferences
```
ezaquarii: We can `protect` it. Kotlin default scope is public, not package-private like in Java.

src/main/java/com/nextcloud/client/di/ComponentsModule.java
```diff
@@ -150,4 +151,7 @@
 
     @ContributesAndroidInjector abstract AccountManagerService accountManagerService();
     @ContributesAndroidInjector abstract OperationsService operationsService();
+
+    @ContributesAndroidInjector
+    abstract NContentObserverJob nContentObserverJob();
```
ezaquarii: Incinsistent placement of annotation (I don't mind tbh).

ezaquarii:
My concern is the test. I have a gut feeling that I've seen this code somewhere few months ago. Some other branch?

If you plan to add test later, we can `@Skip` it so we keep an eye on it.

AndyScherzinger:
@tobiasKaminsky can you comment on @ezaquarii's review comments. Besides them I think this is ready and would be nice if we get this in for 3.8.0 RC1.

AndyScherzinger:
@tobiasKaminsky can you comment on @ezaquarii's review comments + rebase due to conflicts with master. Besides them I think this is ready and would be nice if we get this in for 3.8.0 RC1. 🚀

AndyScherzinger:
ping @tobiasKaminsky

src/main/java/com/nextcloud/client/device/DeviceModule.java
tobiasKaminsky: Yes, there was an issue, but I do not remember what it was :/

src/test/java/com/nextcloud/client/device/TestPowerManagementService.kt
tobiasKaminsky: Right, I removed this line.

src/test/java/com/nextcloud/client/device/TestPowerManagementService.kt
tobiasKaminsky: Not really needed as it is "only" a test class, or?

src/main/java/com/nextcloud/client/di/ComponentsModule.java
tobiasKaminsky: This is due to auto format of (my) Android Studio. 
It always put it into two lines.

tobiasKaminsky:
Done
@ezaquarii I think we can change the test later, if needed.

tobiasKaminsky:
Merging due to unrelated CI error.

tobiasKaminsky:
We do not need a backport as 3.8.0 RC1 will be released today.

PR #3608, opened 2019-02-11
Actions in notifications

tobiasKaminsky:
This will add actions, like accept or dismiss, directly in push notifications.


tobiasKaminsky:
The only drawback currently is, that the user can do these actions also on the lock screen, which for e.g. 2fa via notification, a security problem.

- I can detect the lock screen and then prevent executing the action, but without visual feedback as this is not possible on lock screen.

- On creating the notification I also could check if lock screen is locked and hide actions, but then the action will still work if you get a notification and then lock the device

tobiasKaminsky:
Notification without lock screen:


Notification on lock screen with "hide sensitive data"


Notification on lock screen with "show all content"

---> here I am unsure if we should / need to prevent executing actions
(e.g. mark as read is also working on WhatsApp)

AndyScherzinger:
I'd be fine with having the actions shown even with security in mind. Other apps like GMail also should destructive actions like delete on the screen. The device should imho be considered the 2nd factor so if you have the device that should make it legit in my opinion.

But we should still get some feedback from @nextcloud/designers and maybe you could ask Maria @tobiasKaminsky?

tobiasKaminsky:
@rullzer security wise?

tobiasKaminsky:
Rebased

rullzer:
> I'd be fine with having the actions shown even with security in mind. Other apps like GMail also show destructive actions like delete on the screen. The device should imho be considered the 2nd factor so if you have the device that should make it legit in my opinion.
> 
> But we should still get some feedback from @nextcloud/designers and maybe you could ask Maria @tobiasKaminsky?

Yeah I'm a bit torn here. Google for example really opens a new screen. But this feels weird

But I tend to agree. If somebody gets ahold of your phone. And your password. I'm willing to bet they also have ways to unlock your phone.

tobiasKaminsky:
Good, and if someone is very privacy aware, he/she can set info level to "hide sensitive data" and then there are no actions at all.

So I'll work further on it :-)

rullzer:
Also the app requires explicit enabling. So people can chose not to use it :wink:

tobiasKaminsky:
On Android 7 we need a generic icon for all the actions.
Do you have in idea? https://material.io/tools/icons/?style=baseline

tobiasKaminsky:
image

maybe?

tobiasKaminsky:
Added correct icon:

AndyScherzinger:
All fine with me @tobiasKaminsky any reason this is not ready to review (or even merge?)?
@rullzer so you would be fine with the solution (as in fine with merging it in its current state) from a security point of view? `tending to something != approval` ;)

tobiasKaminsky:
From my side this is ready to review, test & merge.

rullzer:
@AndyScherzinger yes lets do this good enough for now.

rullzer:
On a more general note... @tobiasKaminsky @nickvergessen maybe iut would be good if we could notifications as senitive? Because then we could just do it similar to how google does it when you login for example?

tobiasKaminsky:
> On a more general note... @tobiasKaminsky @nickvergessen maybe iut would be good if we could notifications as senitive? Because then we could just do it similar to how google does it when you login for example?

What do you mean?

AndyScherzinger:
@tobiasKaminsky can be merged :)

> What do you mean?

If a notification is marked sensitive we could implement a diifferent mechanis, as-in not showing actions but launching an activity which then needs you to unlock the phone.

tobiasKaminsky:
> If a notification is marked sensitive we could implement a diifferent mechanis, as-in not showing actions but launching an activity which then needs you to unlock the phone.

Isn't this the visibility of lock screen? Then I already implemented it:
on lock screen:
- show all notification content: we show notification with actions
- hide sensitive content: we show "new notification" without actions
- don't show notifications at all: nothing to do on our side

(but this is already explained in this PR…)

src/main/java/com/owncloud/android/jobs/NotificationJob.java
```diff
@@ -191,6 +277,39 @@ public void onReceive(Context context, Intent intent) {
...
+
+                case "POST":
+                    method = new PostMethod(actionLink);
+                    break;
+
+                case "DELETE":
+                    method = new DeleteMethod(actionLink);
+                    break;
+
+                default:
```
nickvergessen: PUT is also allowed, although I'm not sure any app uses it:
https://github.com/nextcloud/server/blob/ba87db3fccb40aa58d84d12932424c83a4db411f/lib/private/Notification/Action.php#L137-L139

nickvergessen:
> On Android 7 we need a generic icon for all the actions.
Do you have in idea? https://material.io/tools/icons/?style=baseline

I think the bell doesn't look too good, but I will keep in optional image in mind for the next touch on the actions API. A new request-type "browser" or something similar is also on the todo, to allow apps to be redirected to a page (instead of a get request in the background).

AndyScherzinger:
>     * hide sensitive content: we show "new notification" without actions

Where would I find this in your code? This is afaik just a general Android OS feature you are using, you don't "know" if a notificatiion has sensitive data and hide the actions, which is what Rouland ment.

> How about check-circle for primary, and check-circle-empty for other actions?

can we distinguish primary and "secondary" actions based on the data?

nickvergessen:
> can we distinguish primary and "secondary" actions based on the data?

There is a primary flag on one of the actions

src/main/java/com/owncloud/android/jobs/NotificationJob.java
tobiasKaminsky: :+1:

tobiasKaminsky:
> I think the bell doesn't look too good, but I will keep in optional image in mind for the next touch on the actions API, added to [nextcloud/notifications#285](https://github.com/nextcloud/notifications/issues/285).

Icons are only shown in <= Android 6.
So I would not put too much effort into it, as the "new" version do not show icons.

tobiasKaminsky:
> Where would I find this in your code? This is afaik just a general Android OS feature you are using, you don't "know" if a notificatiion has sensitive data and hide the actions, which is what Rouland ment.

setPublicVersion:
> Supply a replacement Notification whose contents should be shown in insecure contexts (i.e. atop the secure lockscreen). 

https://github.com/nextcloud/android/blob/5f7b25f8a1a87fb909927194433fb46f1fe9ef27/src/main/java/com/owncloud/android/jobs/NotificationJob.java#L191-L203

tobiasKaminsky:
> > How about check-circle for primary, and check-circle-empty for other actions?
> 
> can we distinguish primary and "secondary" actions based on the data?

I'll add it :+1:

nickvergessen:
> So I would not put too much effort into it, as the "new" version do not show icons.

Ah i read it the other way around

jancborchardt:
Sorry for the late reply! :) 

The "Approve" and icon seem good! The "Cancel" however, shouldn’t that be called "Deny"?
And the icon should not be a checkmark but this cancel X: https://material.io/tools/icons/?search=cancel&icon=cancel&style=baseline
cancel

AndyScherzinger:
@jancborchardt it can't ;) We simply show the actions provided by the server, we only know what the primary action is (tagged as such) we have no idea what the action does/means... So we don't know that it is a cancel... (or else)

jancborchardt:
Aaah ok – in that case can we only have the icon for the primary action? Then only the primary action pulls the actual attention.

And so the rename of Cancel to Deny we’d have to do on the server?

rullzer:
> And so the rename of Cancel to Deny we’d have to do on the server?

in the app. Is already done. Release will follow one of these days.

AndyScherzinger:
> in that case can we only have the icon for the primary action?

Unfortunately no, all actions enforce an icon (legacy Android) while newer Android version enforce no-action ever for any action...

jancborchardt:
@AndyScherzinger so there’s no way at all we can know what icon to set, cause we only know the primary action (not even what it is) and we don’t know the meaning of the secondary one?

In that case I’d rather go without any icons. What do you think?

AndyScherzinger:
>@AndyScherzinger so there’s no way at all we can know what icon to set, cause we only know the primary action (not even what it is) and we don’t know the meaning of the secondary one?

We just get a list of actions and one of them is flag as primary

> In that case I’d rather go without any icons. What do you think?

Like I said impossible, APIs on Android for older Android Versions have the icon as a mandatory field so we a forced to set an icon. For newer Android versions we can't set an icon at all (again, in newer version the API forbids to set one).

PR #3616, opened 2019-02-13
add Android FindBugs plugin

tobiasKaminsky:
Locally I am also using this.


tobiasKaminsky:
Not sure why I closed this, so test again.

AndyScherzinger:
@tobiasKaminsky would this need to be replaced by SpotBugs?

tobiasKaminsky:
Seems that this version is not longer available…
You can give it a try, but I think otherwise we can close it.

AndyScherzinger:
I am fine with closing since I only check my PRs via CI

PR #3624, opened 2019-02-14
Test CreateFolderOperation without server

tobiasKaminsky:
We have CreateFolderRemoteOperation which is in library and is (or should be) tested there.
Goal of this PR is to test CreateFolderOperation without really using CreateFolderRemoteOperation or any other RemoteOperation, so that this is all Unit Testing.

It is meant to be a start, where we can discuss.

Some consideration:
- I could also have mocked FileDataStorageManager and test the class in a separate unit test
- I commented Log_OC.d(…) as I have no idea how to mock this: maybe same approach like with PreferencesManager? But we use this way more often…
- Tests for failure, multiple folder structure, etc. are missing
- I used a method "wrapper" for CreateFolderRemoteOperation and ReadFolderRemoteOperation, maybe there is something easier?

@ezaquarii maybe you have some input in this.

### DIFF
--- a/src/main/java/com/owncloud/android/operations/CreateFolderOperation.java
+++ b/src/main/java/com/owncloud/android/operations/CreateFolderOperation.java
@@ -60,11 +60,11 @@ public class CreateFolderOperation extends SyncOperation implements OnRemoteOper
 
 
     @Override
-    protected RemoteOperationResult run(OwnCloudClient client) {
-        RemoteOperationResult result = new CreateFolderRemoteOperation(mRemotePath, mCreateFullPath).execute(client);
+    public RemoteOperationResult run(OwnCloudClient client) {
+        RemoteOperationResult result = createCreateFolderRemoteOperation(mRemotePath, mCreateFullPath).execute(client);
 
         if (result.isSuccess()) {
-            RemoteOperationResult remoteFolderOperationResult = new ReadFolderRemoteOperation(mRemotePath)
+            RemoteOperationResult remoteFolderOperationResult = createReadFolderRemoteOperation(mRemotePath)
                 .execute(client, true);
 
             createdRemoteFolder = (RemoteFile) remoteFolderOperationResult.getData().get(0);
@@ -74,6 +74,14 @@ public class CreateFolderOperation extends SyncOperation implements OnRemoteOper
         }
 
         return result;
+    }
+
+    public CreateFolderRemoteOperation createCreateFolderRemoteOperation(String mRemotePath, boolean mCreateFullPath) {
+        return new CreateFolderRemoteOperation(mRemotePath, mCreateFullPath);
+    }
+
+    public ReadFolderRemoteOperation createReadFolderRemoteOperation(String mRemotePath) {
+        return new ReadFolderRemoteOperation(mRemotePath);
     }
 
     @Override
@@ -120,7 +128,7 @@ public class CreateFolderOperation extends SyncOperation implements OnRemoteOper
             newDir.setEncrypted(FileStorageUtils.checkEncryptionStatus(newDir, getStorageManager()));
             getStorageManager().saveFile(newDir);
 
-            Log_OC.d(TAG, "Create directory " + mRemotePath + " in Database");
+            // Log_OC.d(TAG, "Create directory " + mRemotePath + " in Database");
         }
     }
 
--- a/src/test/java/com/owncloud/android/FileDataStorageManagerLocal.java
+++ b/src/test/java/com/owncloud/android/FileDataStorageManagerLocal.java
@@ -1,0 +1,149 @@
+package com.owncloud.android;
+
+import android.accounts.Account;
+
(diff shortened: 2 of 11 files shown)


AndyScherzinger:
Looks good code-wise on an cellphone.
As for the logger, well I would expect a logger to be usable in a static way, so I wouldn't go for the same approach as for the preference manager. Maybe we could have a method injecting the writer within the logger class so that for testing it writes to something like a list or simply doesn't write or store anything anywhere.

tobiasKaminsky:
> Maybe we could have a method injecting the writer within the logger class so that for testing it writes to something like a list or simply doesn't write or store anything anywhere.

Right, like with FileDataStorageManagerInterface, we could have Log_OCInterface…
But is this the best approach…

src/main/java/com/owncloud/android/operations/CreateFolderOperation.java
```diff
@@ -76,6 +76,14 @@ protected RemoteOperationResult run(OwnCloudClient client) {
         return result;
     }
 
+    public CreateFolderRemoteOperation createCreateFolderRemoteOperation(String mRemotePath, boolean mCreateFullPath) {
```
ezaquarii: Those m prefixes are probably a left-overs after either refactoring or automated generation? It can be trimmed with Shift-F6 in no time.

src/test/java/com/owncloud/android/FileDataStorageManagerLocal.java
```diff
@@ -0,0 +1,149 @@
...
+    @Override
+    public boolean saveFile(OCFile file) {
+        pathToFile.put(file.getRemotePath(), file);
+
+        return true;
+    }
+
+    @Override
+    public boolean fileExists(long id) {
+        throw new IllegalArgumentException("to implement");
```
ezaquarii: How about `UnsupportedOperationException`?

ezaquarii:
> @ezaquarii maybe you have some input in this.

Pls allow me to digest it. It's pretty big PR and I need to familiarize myself with the code it touches.

src/test/java/com/owncloud/android/operations/CreateFolderOperationTest.java
```diff
@@ -0,0 +1,73 @@
...
+    @Mock CreateFolderRemoteOperation createFolderRemoteOperation;
+    @Mock ReadFolderRemoteOperation readFolderRemoteOperation;
+
+    @Before
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void testCreateFolder() {
```
ezaquarii: To be honest, I'm not sure what this test does, as the name of the test doesn't explain it.

It looks like it tests folder creation. How about `testNewFolderIsCreated()` or something similarly expressive?

tobiasKaminsky:
As this seems to be the right way, thanks @ezaquarii, I'll do three PRs:
- create FileDataStorageManagerInterface
- create Log Interface
- create Test PR based on both above

So this clean and easy to review.

AndyScherzinger:
So @ezaquarii and @tobiasKaminsky I'd like to start the interface naming discussion :D
My vote would be to no prefix or suffix the interface names. Usually we probably always work with the interface so the code is easier to read and less verbose. As for implementing classes I would even suggest the same thing as in to choose a proper name for the class which hints to the class' intent and implementation case. In edge cases where we can't find a name for the class we then could add an Impl suffix but rather as some kind of a last resort.

AndyScherzinger:
Example
* FileDataStorageManager (Inferface), 
* FileDataStorageManagerLocal (for Test) and
* FileDataStorageManagerDB (for SQLite/App)

ezaquarii:
> Example
> 
>     * FileDataStorageManager (Inferface),
> 
>     * FileDataStorageManagerLocal (for Test) and
> 
>     * FileDataStorageManagerDB (for SQLite/App)

It's not a bad practice to optimize for most frequent expected use case:

1. interface
2. there is only 1 default implementation
3. there is at least 1 mock

It boils down to choosing between 2 industry wide conventions:

1. `IFoo` + `Foo` + `FooTest` + some `Runnable`-style exceptions
2. `Foo` + `FooImpl` + `FooImplTest`/`FooTest` + (matches default `Runnable` style, but has `Impl` wart)

Each one of them has some corner aesthetic issues, philosophical foundations, fans and haters. I have my opinion of course, but it's not important as long as:

1. you are not intrusive
2. you are consistent across the stack.

I believe Android picked `I`-`Impl` variant.

AndyScherzinger:
Well, in that case I vote for variant 2 proposed by @ezaquarii since that should fit the most frequent case and as stated would also be the typical Android naming schema pick.

ezaquarii:
Is it going to be merged? Still WIP?

tobiasKaminsky:
> Is it going to be merged? Still WIP?

Yes, I just need to find a bit more time… :-)

PR #3625, opened 2019-02-14
Show snackbar after 10s when cool is loading to slow

tobiasKaminsky:


tobiasKaminsky:
> What if primary dark is black? I think we did go for white for the back action in the Snackbar in other places, no?

C&P from https://github.com/nextcloud/android/blob/c83230198db78e13b196f9e5dfaae660a4ee56c5/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java#L405-L414

AndyScherzinger:
Hmmm, I thought I did this differently somewhere since this isn't save for dark colors thus using white (should be like this somewhere...)

PR #3668, opened 2019-02-22
Refactor legacy instant upload prefs removal

ezaquarii:
- Refactored legacy instant upload preferences to instance methods
- Duplicate legacy preferences removal logic moved to `PreferenceManager`

This is purposefully small change to open discussion about next move. Impact around legacy code should have minimal influence on currently deployed clients.

### DIFF
--- a/src/main/java/com/nextcloud/client/preferences/PreferenceManager.java
+++ b/src/main/java/com/nextcloud/client/preferences/PreferenceManager.java
@@ -17,7 +17,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-package com.owncloud.android.db;
+package com.nextcloud.client.preferences;
 
 import android.accounts.Account;
 import android.content.Context;
@@ -36,7 +36,7 @@ import static com.owncloud.android.ui.fragment.OCFileListFragment.FOLDER_LAYOUT_
 /**
  * Helper to simplify reading of Preferences all around the app
  */
-public final class PreferenceManager {
+public final class PreferenceManager implements AppPreferences {
     /**
      * Constant to access value of last path selected by the user to upload a file shared from other app.
      * Value handled by the app without direct access in the UI.
@@ -69,7 +69,15 @@ public final class PreferenceManager {
     public static final String PREF__LOCK_TIMESTAMP = "lock_timestamp";
     private static final String PREF__SHOW_MEDIA_SCAN_NOTIFICATIONS = "show_media_scan_notifications";
 
-    private PreferenceManager() {
+    private final SharedPreferences preferences;
+
+    public static AppPreferences fromContext(Context context) {
+        SharedPreferences prefs = getDefaultSharedPreferences(context.getApplicationContext());
+        return new PreferenceManager(prefs);
+    }
+
+    PreferenceManager(SharedPreferences preferences) {
+        this.preferences = preferences;
     }
 
     public static void setKeysReInit(Context context) {
@@ -88,12 +96,14 @@ public final class PreferenceManager {
         return getDefaultSharedPreferences(context).getString(PREF__PUSH_TOKEN, "");
     }
 
-    public static boolean instantPictureUploadEnabled(Context context) {
-        return getDefaultSharedPreferences(context).getBoolean(PREF__INSTANT_UPLOADING, false);
+    @Override
+    public boolean instantPictureUploadEnabled() {
+        return preferences.getBoolean(PREF__INSTANT_UPLOADING, false);
     }
 
-    public static boolean instantVideoUploadEnabled(Context context) {
-        return getDefaultSharedPreferences(context).getBoolean(PREF__INSTANT_VIDEO_UPLOADING, false);
+    @Override
+    public boolean instantVideoUploadEnabled() {
+        return preferences.getBoolean(PREF__INSTANT_VIDEO_UPLOADING, false);
     }
 
     public static boolean instantPictureUploadPathUseSubfolders(Context context) {
@@ -592,5 +602,23 @@ public final class PreferenceManager {
 
     public static SharedPreferences getDefaultSharedPreferences(Context context) {
         return android.preference.PreferenceManager.getDefaultSharedPreferences(context.getApplicationContext());
+    }
+
+    @Override
+    public void removeLegacyPreferences() {
+        preferences.edit()
+                .remove("instant_uploading")
+                .remove("instant_video_uploading")
+                .remove("instant_upload_path")
+                .remove("instant_upload_path_use_subfolders")
+                .remove("instant_upload_on_wifi")
+                .remove("instant_upload_on_charging")
+                .remove("instant_video_upload_path")
+                .remove("instant_video_upload_path_use_subfolders")
+                .remove("instant_video_upload_on_wifi")
+                .remove("instant_video_uploading")
+                .remove("instant_video_upload_on_charging")
+                .remove("prefs_instant_behaviour")
+                .apply();
     }
 }
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -51,7 +51,8 @@ import com.owncloud.android.datamodel.SyncedFolderProvider;
 import com.owncloud.android.datamodel.ThumbnailsCacheManager;
 import com.owncloud.android.datastorage.DataStorageProvider;
 import com.owncloud.android.datastorage.StoragePoint;
-import com.owncloud.android.db.PreferenceManager;
+import com.nextcloud.client.preferences.AppPreferences;
+import com.nextcloud.client.preferences.PreferenceManager;
 import com.owncloud.android.jobs.MediaFoldersDetectionJob;
 import com.owncloud.android.jobs.NCJobCreator;
 import com.owncloud.android.lib.common.OwnCloudClientManagerFactory;
@@ -113,6 +114,7 @@ public class MainApp extends MultiDexApplication {
     private static boolean mOnlyOnDevice;
 
     private SharedPreferences appPrefs;
+
     @SuppressWarnings("unused")
     private boolean mBound;
 
@@ -507,23 +509,9 @@ public class MainApp extends MultiDexApplication {
 
     private static void updateToAutoUpload() {
             Context context = getAppContext();
-            if (PreferenceManager.instantPictureUploadEnabled(context) ||
-                    PreferenceManager.instantVideoUploadEnabled(context)){
-
-                // remove legacy shared preferences
-                SharedPreferences.Editor editor = PreferenceManager.getDefaultSharedPreferences(context).edit();
-                editor.remove("instant_uploading")
-                        .remove("instant_video_uploading")
-                        .remove("instant_upload_path")
-                        .remove("instant_upload_path_use_subfolders")
-                        .remove("instant_upload_on_wifi")
-                        .remove("instant_upload_on_charging")
-                        .remove("instant_video_upload_path")
-                        .remove("instant_video_upload_path_use_subfolders")
-                        .remove("instant_video_upload_on_wifi")
-                        .remove("instant_video_uploading")
-                        .remove("instant_video_upload_on_charging")
-                        .remove("prefs_instant_behaviour").apply();
+            AppPreferences preferences = PreferenceManager.fromContext(context);
+            if (preferences.instantPictureUploadEnabled() || preferences.instantVideoUploadEnabled()){
+                preferences.removeLegacyPreferences();
 
                 // show info pop-up
                 try {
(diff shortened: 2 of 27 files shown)


ezaquarii:
I'm not convinced that preferences should be kept in db package, as it is not a database code. Shall we move it to separate package? `com.nextcloud.android.preferences` maybe?

src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -507,23 +509,9 @@ private static String getUserAgent(@StringRes int agent) {
...
-                        .remove("instant_upload_path_use_subfolders")
-                        .remove("instant_upload_on_wifi")
-                        .remove("instant_upload_on_charging")
-                        .remove("instant_video_upload_path")
-                        .remove("instant_video_upload_path_use_subfolders")
-                        .remove("instant_video_upload_on_wifi")
-                        .remove("instant_video_uploading")
-                        .remove("instant_video_upload_on_charging")
-                        .remove("prefs_instant_behaviour").apply();
+            AppPreferences prefs = PreferenceManager.fromContext(context);
```
tobiasKaminsky: We always tend to not abbreviate variable names for better readability.
Especially if such a variable is used later.

src/main/java/com/owncloud/android/db/AppPreferences.java
```diff
@@ -0,0 +1,8 @@
+package com.owncloud.android.db;
```
tobiasKaminsky: Can you add (if you agree) this as a license header:
```
/*
 * Nextcloud Android client application
 *
 * @author Name
 * Copyright (C) 2019 Name
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program. If not, see <https://www.gnu.org/licenses/>.
 */
```

src/main/java/com/owncloud/android/db/PreferenceManager.java
```diff
@@ -69,7 +69,15 @@
     public static final String PREF__LOCK_TIMESTAMP = "lock_timestamp";
     private static final String PREF__SHOW_MEDIA_SCAN_NOTIFICATIONS = "show_media_scan_notifications";
 
-    private PreferenceManager() {
+    private final SharedPreferences mPrefs;
```
tobiasKaminsky: At some point we decided to not use m prefix as it is visible on all modern IDEs and thus not needed.

src/main/java/com/owncloud/android/db/PreferenceManager.java
```diff
@@ -69,7 +69,15 @@
...
+
+    public static AppPreferences fromContext(Context context) {
+        SharedPreferences prefs = getDefaultSharedPreferences(context.getApplicationContext());
+        return new PreferenceManager(prefs);
+    }
+
+    private PreferenceManager(SharedPreferences preferences) {
+        mPrefs = preferences;
```
tobiasKaminsky: Just out of curiosity, why do you split this up?

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -215,6 +217,8 @@ protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState); // this calls onAccountChanged() when ownCloud Account
         // is valid
 
+        mPreferences = PreferenceManager.fromContext(this);
```
tobiasKaminsky:
```suggestion
        preferenceManager = PreferenceManager.fromContext(this);
```

tobiasKaminsky:
> Shall we move it to separate package? `com.nextcloud.android.preferences` maybe?

Totally fine with it. Also to use nextcloud instead of owncloud, when doing refactoring. This way we can slowly migrate our app and see which parts need to be done; and also introduce a new package schema.

AndyScherzinger:
> Shall we move it to separate package? `com.nextcloud.android.preferences` maybe?

Also perfectly fine with me, while we could also discuss _if_ we should have something like a persistence package as-in a repository thing since preferences could be seen like database or services. In the end we read/write data from/to a source. While we can do such things at a later point in case it then still makes sense, so preferences package sound good 👍

tobiasKaminsky:
Can you also (if you agree) sign-off your github commits, this means that you agree too https://developercertificate.org/

tobiasKaminsky:
And my last remark: you can open your next PR/branch directly in our nextcloud repo, which makes it easier for checkout, update, etc.

src/main/java/com/owncloud/android/MainApp.java
ezaquarii: fixed

src/main/java/com/owncloud/android/db/AppPreferences.java
ezaquarii: Added licence header to `AppPreferences.java`

src/main/java/com/owncloud/android/db/PreferenceManager.java
ezaquarii: Removed `m` prefix

src/main/java/com/owncloud/android/db/PreferenceManager.java
ezaquarii: Constructed object should not be responsible for creating it's dependencies - those should be injected. Otherwise you woudn't be able to instantiate it in unit test without mocking `Context`.

Since we are not using any DI solution yet, I provided a factory method that takes care of it as an intermediate step. At some point those factory methods should be replaced with DI of some sort.

This loosely follows convention used in Android platform: https://developer.android.com/reference/android/view/LayoutInflater.html#from(android.content.Context)

Once some unit test is provided, constructor visibility should be relaxed to package-level (or public, if tests live in different package - not recommended).

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
ezaquarii: done

src/main/java/com/owncloud/android/db/PreferenceManager.java
ezaquarii: > Can you also (if you agree) sign-off your github commits, this means that you agree to https://developercertificate.org/

Sure, I'll sign it off. Should I also paste this document into a commit message?

There is one thing about git history that bothers me. I believe we'll create a series of small commits with "fix for this" and "review that" messages. At a merge time, this "creative nosie" can be either squashed into a 1 commit or merged as is with full history.

I'm a big fan of squashing noise of intermediate fixes and merging one, solid piece of work.

What would work for you? Shall I sign-off a final commit?

ezaquarii: Outside GitHub I usually `git commit --amend` any tiny fixes, but this completely breaks GitHub code review, as GitHub tracks conversations by commit sha, so this workflow is probably a no-go for us.

ezaquarii:
I run across manifest to and directory structure and I see few package names being in use:

1. `com.owncloud.android` - obviously we're not going to use it
2. `com.nextcloud.android.sso` - what is the origin of this one?
3. `com.nextcloud.client` - google play package; can't change it, so maybe this should be the candidate?

What would be the best package to keep our new source then?

src/main/java/com/owncloud/android/db/PreferenceManager.java
tobiasKaminsky: Thanks for the long explanation, always happy to learn something :+1: 
DI is/was also on my roadmap, but due to time…

Regarding DCO: it should be enought to have it signed via git commit -s (or the same in AndroidStudio); no need to copy/paste any document.

Regarding commits: If I have too many commits, I usually squash them all in one or fewer commits…
Same applies if a merge commit occurs, this is also better to rebase.
--> tl;dr: squash + rebase at a last step before merging is fine :-)

tobiasKaminsky:
> * `com.owncloud.android` - obviously we're not going to use it
> * `com.nextcloud.android.sso` - what is the origin of this one?

This is part of https://github.com/nextcloud/Android-SingleSignOn, but can be migrated to any other package.

> * `com.nextcloud.client` - google play package; can't change it, so maybe this should be the candidate?

I would be fine with this. "android" as a naming part is not needed and "com.nextcloud" is maybe a bit too short/lacking of info that it is a client.

--> we then could also migrate com.nextcloud.android.sso to com.nextcloud.client.sso (but this then needs to be done in sync with SSO-lib).

src/main/java/com/owncloud/android/db/PreferenceManager.java
AndyScherzinger: @tobiasKaminsky squash needs to be discussed imho sinice it likely removed authorship (sign-off might still be in the commit message, git blame won't work anymore...)

ezaquarii:
Squashed and signed-off.

It looks like GitHub managed to preserve the conversation, despite rewritten history.

src/main/java/com/nextcloud/client/preferences/PreferenceManager.java
```diff
@@ -36,7 +36,7 @@
 /**
  * Helper to simplify reading of Preferences all around the app
  */
-public final class PreferenceManager {
+public final class PreferenceManager implements AppPreferences {
```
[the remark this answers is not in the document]
AndyScherzinger: to be ignored since this is a known issue and already present (while codacy seems to not always understand that...)

AndyScherzinger:
@ezaquarii the build now fails with compile errors in the different build flavors due to the `PushUtils`whichh references `PreferenceManager`, I don't have a fast enough internet connection at the moment so I can't fix it for you. See https://github.com/ezaquarii/android/blob/refactor-legacy-instant-upload-prefs-removal/src/gplay/java/com/owncloud/android/utils/PushUtils.java

ezaquarii:
> @ezaquarii now the build fails with compile errors in the different build flavors due to the `PushUtils`which references `PreferenceManager`, I don't have a fast enough internet connection at the moment so I can't fix it for you (checkout would take forever). See https://github.com/ezaquarii/android/blob/refactor-legacy-instant-upload-prefs-removal/src/gplay/java/com/owncloud/android/utils/PushUtils.java That file still uses the old package

No problem, I'll fix it tomorrow morning.

AndyScherzinger:
> No problem, I'll fix it tomorrow morning.

Thanks a million @ezaquarii - I would have side winged you on this but unfortunately network connections are quite the road block at the hotel room :/

ezaquarii:
I added a legacy prefs removal test. I'll try to refactor it a bit next time and a test should be useful then.

AndyScherzinger:
@ezaquarii I pushed a fix for the compile error in the gplay buildflavor: https://github.com/nextcloud/android/pull/3668/commits/0e56b544b9360195143c2c60d3801abb45d129eb let's see if this fixes the CI build 😃

AndyScherzinger:
@tobiasKaminsky can you check CI, for some reasoon the build fails due to problems pushing the lint reports... :(

ezaquarii:
I fixed the conflict in `OCFileListAdapter` and squashed the previous fix.

ezaquarii:
> IT test failed: https://www.kaminsky.me/nc-dev/android-integrationTests/8079

How can I set up test environment to run those intrgration tests?

AndyScherzinger:
> How can I set up test environment to run those intrgration tests?

@tobiasKaminsky can you answer this question :)

ezaquarii:
I don't have access to the detailed CI logs. Can somebody check for me what went wrong?

AndyScherzinger:
@ezaquarii what I can see is 
```
/drone/src/github.com/nextcloud/android/src/versionDev/java/com/owncloud/android/utils/PushUtils.java:27: error: cannot find symbol
import com.owncloud.android.db.PreferenceManager;
                              ^
  symbol:   class PreferenceManager
  location: package com.owncloud.android.db
1 error

> Task :compileVersionDevDebugJavaWithJavac FAILED

FAILURE: Build failed with an exception.
```

other than that, can't you access, e.g. https://drone.nextcloud.com/nextcloud/android/8091/12

tobiasKaminsky:
@ezaquarii we have the different flavours: generic, devVersion, gplay.
And on devVersion (which is a nightly build flavor) the PreferencesManager change in PushUtils was missing.
(devVersion and generic (for fdroid) do not support google stuff, so they have a dummy class for this).

tobiasKaminsky:
> How can I set up test environment to run those intrgration tests?

Sorry, I was sick on Friday.
The integration tests are run against the settings found in gradle.properties, but can be overriden by ~/.gradle/gradle.properties
So you would need to have local Nextcloud server with an existing user.

ezaquarii:
> @ezaquarii we have the different flavours: generic, devVersion, gplay.
> And on devVersion (which is a nightly build flavor) the PreferencesManager change in PushUtils was missing.
> (devVersion and generic (for fdroid) do not support google stuff, so they have a dummy class for this).

Phew, it looks like we made it through. All issues fixed.

AndyScherzinger:
Woohoo... Aaaaaaand merged 🎉

Nice work @ezaquarii 👍

PR #3688, opened 2019-02-28
Fix federated sharing

### LINKED ISSUE #3666: Not able to choose remote (federated id) share option.
### Actual behaviour
When I try to share a folder with for example user@domain.example the app gives me two options. I can either choose between a remote account or an email address. Regardless what I'm selecting, remote or email, the app always creates a share by email and not by the federated id.
In the web UI I'm able to create the remote share without any issues.

### Expected behaviour
If I select `remote` a remote share should be created instead of an email share.
 

### Environment data
Android version: 6.0.1 

Device model: Motorola Moto X (xt1052)

Stock or customized system: Lineage OS 13.0

Nextcloud app version: 3.5.0

Nextcloud server version: 14.0.6 and 15.0.4

tobiasKaminsky:
Fix #3666 


AndyScherzinger:
@tobiasKaminsky findbugs is up by +1, thus the build fails

PR #3698, opened 2019-03-04
Move showDetailedTimestampPreference to AppPreferences interface

ezaquarii:
Please see review comments!

### DIFF
--- a/src/main/java/com/owncloud/android/ui/fragment/FileDetailFragment.java
+++ b/src/main/java/com/owncloud/android/ui/fragment/FileDetailFragment.java
@@ -23,6 +23,7 @@
 package com.owncloud.android.ui.fragment;
 
 import android.accounts.Account;
+import android.app.Activity;
 import android.content.Context;
 import android.graphics.Bitmap;
 import android.os.Bundle;
@@ -40,6 +41,7 @@ import android.widget.ProgressBar;
 import android.widget.TextView;
 
 import com.google.android.material.tabs.TabLayout;
+import com.nextcloud.client.preferences.AppPreferences;
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
 import com.owncloud.android.datamodel.FileDataStorageManager;
@@ -103,7 +105,7 @@ public class FileDetailFragment extends FileFragment implements OnClickListener 
     @BindView(R.id.size)
     TextView fileSize;
 
-    @BindView(R.id.modified)
+    @BindView(R.id.last_modification_timestamp)
     TextView fileModifiedTimestamp;
 
     @BindView(R.id.favorite)
@@ -139,6 +141,7 @@ public class FileDetailFragment extends FileFragment implements OnClickListener 
     private ProgressListener progressListener;
     private ToolbarActivity activity;
     private int activeTab;
+    private AppPreferences preferences;
 
     /**
      * Public factory method to create new FileDetailFragment instances.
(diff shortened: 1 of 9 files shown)


src/main/java/com/owncloud/android/ui/fragment/FileDetailFragment.java
```diff
@@ -483,9 +492,9 @@ public void onClick(View v) {
                 break;
             }
             case R.id.modified: {
-                boolean flag = !PreferenceManager.isShowDetailedTimestamp(getContext());
-                PreferenceManager.setShowDetailedTimestamp(getContext(), flag);
-                setFileModificationTimestamp(getFile(), flag);
+                boolean showDetailedTimestamp = !preferences.isShowDetailedTimestampEnabled();
```
ezaquarii: I tried to find a clickable way to trigger this code path but coudn't find it. I did a refactoring semi-automatically, but I think this might be a dead code. Shall we remove it in such case?

AndyScherzinger: This shouldn't be dead code actually. It should be triggered when you click on the modification date to toggle between relative and absolute date values.

AndyScherzinger: @ezaquarii https://github.com/nextcloud/android/pull/3399 should be the corresponding PR

ezaquarii: Ok, I refactored the ID name, as I found this feature non-discoverable. Now it should be clear that it refers to the timestamp label.

tobiasKaminsky:
Works for me :-)

To trigger it:
- open file list
- click "three dots" on file/folder -> details
- see detail screen
- click on modification date next to size
- toggles between relative timestamp and absolute timestamp

tobiasKaminsky:
@ezaquarii I invited you to our team, so that you can directly push the next branch/PR to nextcloud/android, which makes it easier for us to check/test.

src/main/java/com/owncloud/android/ui/fragment/FileFragment.java
```diff
@@ -43,7 +43,7 @@
 
     private OCFile mFile;
```
tobiasKaminsky: While we are already on this class: can you also remove m prefix here?
So it is consistent within this class?

tobiasKaminsky: I would leave the other classes this way and do maybe another PR, with only getting rid of m prefix, but this would then bloat up this PR.

AndyScherzinger:
@ezaquarii if you label a PR `2. to review` the will be tracked and show up here https://www.kaminsky.me/nc-dev/reviews.php

@tobiasKaminsky can we add @ezaquarii there too? Just in case you are fine with that of course @ezaquarii and also want to do PR review (which you already do 🎉)

src/main/java/com/owncloud/android/ui/fragment/FileFragment.java
ezaquarii: This is trivial PR and I already killed one `m` here. I think we can afford finishing it. :)

AndyScherzinger: I agree with @ezaquarii, killing prefixes should be done in a separate OR that only removes prefixes, that makes that one easier to review also :)

ezaquarii: I'll make sure `m`s are addressed as a separate PR in the future.

PR #3704, opened 2019-03-05
Integration Tests: Show log in case of error

tobiasKaminsky:

### DIFF
--- a/src/androidTest/java/com/owncloud/android/FileIT.java
+++ b/src/androidTest/java/com/owncloud/android/FileIT.java
@@ -29,7 +29,7 @@ public class FileIT extends AbstractIT {
         SyncOperation syncOp = new CreateFolderOperation(path, true);
         RemoteOperationResult result = syncOp.execute(client, getStorageManager());
 
-        assertTrue(result.isSuccess());
+        assertTrue(result.toString(), result.isSuccess());
 
         // folder exists
         assertTrue(getStorageManager().getFileByPath(path).isFolder());
@@ -46,7 +46,7 @@ public class FileIT extends AbstractIT {
 
         SyncOperation syncOp = new CreateFolderOperation(path, true);
         RemoteOperationResult result = syncOp.execute(client, getStorageManager());
-        assertTrue(result.isSuccess());
+        assertTrue(result.toString(), result.isSuccess());
 
         // folder exists
         assertTrue(getStorageManager().getFileByPath(path).isFolder());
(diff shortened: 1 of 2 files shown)


src/androidTest/java/com/owncloud/android/FileIT.java
```diff
@@ -46,7 +46,7 @@ public void testCreateNonExistingSubFolder() {
 
         SyncOperation syncOp = new CreateFolderOperation(path, true);
         RemoteOperationResult result = syncOp.execute(client, getStorageManager());
-        assertTrue(result.isSuccess());
+        assertTrue(result.getLogMessage(), result.isSuccess());
```
ezaquarii: That is going to work ok.

How about `RemoteOperationResult.toString()` to provide full insight into the object state? The additional benefit is that debugger uses `toString()`, making any code stepping-through very pleasant experience. And you can get free debug output in many other situations.

That is - of course - assuming that `toString()` is not used for human-readable output.

tobiasKaminsky: Nice idea.
This then needs to be changed on library, but I'll change it here to "toString()".

ezaquarii: The trick is, that if you are using an interface, you'll need to visit all implementations.

tobiasKaminsky: https://github.com/nextcloud/android-library/pull/256

ezaquarii: I'm wondering if this https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#toString(java.lang.Object) can do the job for you quickly. Apache has something similar, I believe.

AndyScherzinger: nextcloud/android-library#256 has been merged and jitpack build has succeeded

tobiasKaminsky: Then approve and merge? :tada:

AndyScherzinger: Well I thought you change it to `assertTrue(result.toString(), result.isSuccess());`?

tobiasKaminsky: Ha, right. I only changed it at one occurrence…

tobiasKaminsky:
So…let us see what we get if the IT tests ever fail again…

PR #3731, opened 2019-03-09
Fix button disabled state in folder sync preferences

ezaquarii:


ezaquarii:
@AndyScherzinger do we have more dialogs or buttons around that require similar theming support?

AndyScherzinger:
@ezaquarii not for the secondary action buttons (text button) we could think about adding a proper support for enabled/disabled state for primary action buttons (contained button) while they probably work okay-ish since @tobiasKaminsky implemented the "re-coloring" to grey when a button gets disabled while it woudl of course be nicer to have this via a color state list of course.

I'd say that (conatined buttons) should be done in a separate PR _if_ you want to tackle that.

AndyScherzinger:
Just did a quick test and it works like a charm 🥇 Thus merging

PR #3756, opened 2019-03-15
Rephrase term "Lock"

### LINKED ISSUE #2806: Rephrase "Lock" term to make it clearer that it is a app access restriction
see https://help.nextcloud.com/t/android-client-pin-fingerprint-protection-pointless/33590/2 

The thread shows that the access restriction we have is/can be misinterpreted as a file access restriction while it is a app access restriction. So I think we should thus show this fact as in maybe renaming "Lock" to "App access restriction" or something like that. Not a native speaker so maybe @nextcloud/designers or @mario @tobiasKaminsky have a better term to be used here :)

dan0xii:
Rephrase term "Lock" to "App passcode" and update dialog to reflect
change.

Closes #2806.

Screenshot_20190315_114539
Screenshot_20190315_114558
Screenshot_20190315_114634

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -37,8 +37,8 @@
     <string name="prefs_category_general">General</string>
     <string name="prefs_category_more">More</string>
     <string name="prefs_manage_accounts">Manage accounts</string>
-    <string name="prefs_lock">Lock</string>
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Thank you :+1:  and :tada: for screenshot :+1:

PR #3771, opened 2019-03-19
Combine sharedPreferences in MainApp

tobiasKaminsky:
In MainApp was a direct access to sharedPreferences, alongside with the new preferences done by @ezaquarii, so I combined both.

### DIFF
--- a/src/main/java/com/nextcloud/client/preferences/AppPreferences.java
+++ b/src/main/java/com/nextcloud/client/preferences/AppPreferences.java
@@ -274,4 +274,10 @@ public interface AppPreferences {
      * @implNote this clears only shared preferences, not preferences kept in account manager
      */
     void clear();
+
+    String getStoragePath(String defaultPath);
+
+    void setStoragePath(String path);
+
+    void removeKeysMigrationPreference();
 }
--- a/src/main/java/com/nextcloud/client/preferences/PreferenceManager.java
+++ b/src/main/java/com/nextcloud/client/preferences/PreferenceManager.java
@@ -20,9 +20,9 @@
 package com.nextcloud.client.preferences;
 
 import android.accounts.Account;
+import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.SharedPreferences;
-
 import com.owncloud.android.authentication.AccountUtils;
 import com.owncloud.android.datamodel.ArbitraryDataProvider;
 import com.owncloud.android.datamodel.FileDataStorageManager;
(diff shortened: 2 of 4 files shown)


ezaquarii:
IT tests reporting socket timeout on my PR too. Something is broken.

src/main/java/com/nextcloud/client/preferences/PreferenceManager.java
```diff
@@ -76,7 +77,7 @@ public static AppPreferences fromContext(Context context) {
         return new PreferenceManager(appContext, prefs);
     }
 
-    public static SharedPreferences getDefaultSharedPreferences(Context context) {
+    private static SharedPreferences getDefaultSharedPreferences(Context context) {
```
ezaquarii: I think this can be now removed. You can use `this.preferences` directly.

src/main/java/com/nextcloud/client/preferences/PreferenceManager.java
```diff
@@ -402,6 +408,29 @@ public void clear() {
...
+    public void setStoragePath(String path) {
+        preferences.edit().putString(STORAGE_PATH, path).commit();
+    }
+
+    @SuppressLint("ApplySharedPref")
+    @Override
+    /*
+     * commit is done on purpose to write immediately
+     */
+    public void removeKeysMigration() {
```
ezaquarii: I believe we could kill the comment with this approach:

```
@Override
@SuppressLint("ApplySharedPref")
public void removeKeysMigration() {
    preferences.edit().remove(...).commit(); // commit synchronously
}
```

src/main/java/com/nextcloud/client/preferences/AppPreferences.java
```diff
@@ -274,4 +274,10 @@
      * @implNote this clears only shared preferences, not preferences kept in account manager
      */
     void clear();
+
+    String getStoragePath(String defaultPath);
+
+    void setStoragePath(String path);
+
+    void removeKeysMigration();
```
ezaquarii: The name suggest hat it removes a migration, but it removes a pref from shared preferences. How about `removeKeysMigrationPref()`? I'm still not sure what is the final effect of it's call, so extra javadoc would help to understand how it works.

src/main/java/com/nextcloud/client/preferences/PreferenceManager.java
tobiasKaminsky: getDefaultSharedPreferences is used in fromContext.
I think you did this on purpose?

https://github.com/nextcloud/android/blob/1941547da49c0678e4b4387273ee3213f1dd8b02/src/main/java/com/nextcloud/client/preferences/PreferenceManager.java#L73-L82

tobiasKaminsky:
Thanks for your review, @ezaquarii.
I changed it accordingly.

PR #3772, opened 2019-03-20
Check for 500 char limit

tobiasKaminsky:
Gplay has limit of 500 char for each changelog.
While we use this only in stable-x.y branch, this should be in master for next stable-3.6 branch.


AndyScherzinger:
👍 @tobiasKaminsky nice job, seems to work just fine :)

tobiasKaminsky:
I deleted both test files, so this can be merged.

PR #3786, opened 2019-03-26
 ui:activity: Properly close InputStreamReader.

ardevd:
The `InputStreamReader` needs to be closed after use. Ideally, use try-with-resources to do this automatically.


AndyScherzinger:
Hi @ardevd! Welcome back 🎉 

Would you fancy signing the commits? See https://github.com/nextcloud/android/pull/3786/checks?check_run_id=85106119 - just asking it is not mandatory but makes sense imho :)

ardevd:
@AndyScherzinger thanks. Busy times. The commits were GPG signed so not entirely sure what the issue is/was but I ran the recommended commands so it's all good now I suppose?

AndyScherzinger:
> I ran the recommended commands so it's all good now I suppose?

Yes, all good 👍

PR #3787, opened 2019-03-26
authentication: Added missing variable assignment.

ardevd:
The `mAuthTokenType` variable is declared and assigned a null value but no value is subsequently assigned, making the null check later on redundant. The `AccountAuthenticator.KEY_AUTH_TOKEN_TYPE` extra value should therefore be assigned the `mAuthTokenType` value I'm assuming.

### DIFF
--- a/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -585,7 +585,7 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
         mAuthTokenType = null;
 
         if (extras != null) {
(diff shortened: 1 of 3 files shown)


AndyScherzinger:
@ardevd fancy signing the commit? 😁

ardevd:
@AndyScherzinger ok, I need to figure this out. I did GPG sign them before pushing them and GitHub shows them as verified. *confused*

AndyScherzinger:
> EDIT: ah, you want me to sign off on them. Gotcha.

Yes, sorry for the confusion, I should have been clearer 🙈

tobiasKaminsky:
Indeed, nice catch!
I have no idea why I removed this when adding #3383…

PR #3788, opened 2019-03-26
files: avoid possible dereference of null pointer.

ardevd:
~~itseems the `capability` object can be null. In which case it is probably a bad idea to pass a null object into `getRichDocumentsDirectEditing()`.~~

This might not be necessary but it looks very strange to me when you define a variable `endToEndEncryptionEnabled` based on whether the `capability` object is `null` first and then pass that same object to the `filterOpenAsRichDocument()` method later on. That function seemingly assumes the `capability` argument is not null too so something should be done here.

`capability` object cannot be null. Updated associated method with `@NonNull` annotation and removed unnecessary NPE check.

### DIFF
--- a/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
+++ b/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
@@ -62,6 +62,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Set;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import lombok.Getter;
 import lombok.Setter;
@@ -2023,6 +2024,7 @@ public class FileDataStorageManager {
         return c;
     }
 
+    @NonNull
     public OCCapability getCapability(String accountName) {
         OCCapability capability;
         Cursor c = getCapabilityCursorForAccount(accountName);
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
> it seems the `capability` object can be null.

Where do you see this? :thinking: 

https://github.com/nextcloud/android/blob/92bbcfb973594f9f0aafa16def97d9f003dcbe54/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java#L2026-L2037

Here we create explicitly a new "empty" object if we do not have yet capabilities.

ardevd:
I assumed as much due to this line:

https://github.com/nextcloud/android/blob/b96414d794bc2183765d1928491d36dc956a892f/src/main/java/com/owncloud/android/files/FileMenuFilter.java#L168

If capability is never null, then I can simply remove that check there instead?

tobiasKaminsky:
> I assumed as much due to this line:

This is indeed not neeeded. 

> If capability is never null, then I can simply remove that check there instead?

Yes, please and maybe you want to add @NonNull to the getCapability method in FileDataStorageManager to avoid such confusion (and also have IDE to warn about useless NPE checks)

ardevd:
Sounds good! I'll take care of it.

tobiasKaminsky:
Thanks for the fast change!

PR #3791, opened 2019-03-27
Inject AppPreferences into activities and fragments

ezaquarii:
Inject `AppPreferences` using Dagger 2 DI.

Dagger maintains a compile-time generated map of types and associated, generated injectors for specific activities and fragments. This map uses a class as a key, so if a type **inherits** injectable base, Dagger will not be able to find the injector, as it will be using instance type to find it - it will not try a**base** types, even if the base-type injector is registered in the map and could be used.

Btw, this is not a Dagger problem specifically, but `dagger.android.AndroidInjection` that uses simple resolution method.

As a result, since many activities inherit from a common, `Injectable` base - ie. abstract `DrawerActivity` via `FileActivity` - `Dagger` must be made aware of all their super-types to create separate injectors for all concrete types of activities - even if it injects only base fields.

This is a bit confusing, as a developer must track inheritance hierarchy to figure out which types should be registered. Currently, our code will throw exception if it fails to inject activity or fragment and I firmly believe we should leave it like this as you don't want to be left with silently uninitialized base members and blow up later.

To make our life easier, the proposed solution simply registers all activities listed in `AndroidManifest.xml` in `ComponentsModule`, regardless of DI usage, making sure that Dagger compiler will supply an injector for it.

The good:

1. Very easy to understand - we require all activities to be DI-enabled, period. No funky exceptions.
2. Probably much easier to refactor things, as we have less moving parts to consider.

The  bad:

Dagger will generate injectors for few activities that don't need them for now.

The  Ugly:

You must register your activitiy in 2 places: `AndroidManifest.xml` *and* `ComponentsModule`.

Optimizing it is IMO simply not worth the effort, as there is very little to gain here - so far DI increased overall binary size by 1%, which is modest price for the simplicity it provides. Also, we're not cranking out new activities too often.


ezaquarii:
Codacy warnings look like false positives in this case.

src/main/java/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -154,7 +154,7 @@ private int delete(SQLiteDatabase db, Uri uri, String where, String... whereArgs
                 count = db.delete(ProviderTableMeta.FILESYSTEM_TABLE_NAME, where, whereArgs);
                 break;
             default:
-                throw new IllegalArgumentException("Unknown uri: " + uri.toString());
+                throw new IllegalArgumentException(String.format(Locale.US, "Unknown uri: %s", uri.toString()));
```
ezaquarii: Findbugs cargo-culting.

tobiasKaminsky:
> The bad:
> 
> Dagger will generate injectors for few activities that don't need them for now.
> 
> The Ugly:
> 
> You must register your activitiy in 2 places: `AndroidManifest.xml` _and_ `ComponentsModule`.

Totally fine by me.
If we forget this too often, we can enhance and write a script which runs on every PR to check if every Activity is registered, but for now I do not assume we need this.

src/main/java/com/nextcloud/client/di/ComponentsModule.java
```diff
@@ -32,9 +70,45 @@
...
+    @ContributesAndroidInjector abstract UploadPathActivity uploadPathActivity();
+    @ContributesAndroidInjector abstract UserInfoActivity userInfoActivity();
+    @ContributesAndroidInjector abstract WhatsNewActivity whatsNewActivity();
 
-    @ContributesAndroidInjector
-    abstract FileDetailFragment fileDetailFragment();
+    @ContributesAndroidInjector abstract ExtendedListFragment extendedListFragment();
+    @ContributesAndroidInjector abstract FileDetailFragment fileDetailFragment();
+    @ContributesAndroidInjector abstract LocalFileListFragment localFileListFragment();
+    @ContributesAndroidInjector abstract OCFileListFragment ocFileListFragment();
```
tobiasKaminsky: Do we also have to manually register every fragment?

tobiasKaminsky:
@ezaquarii 

Codacy complains a lot regarding scoping, I also prefer explicit scoping…

https://app.codacy.com/app/nextcloud/android/pullRequest?prid=3332947

ezaquarii:
> @ezaquarii
> 
> Codacy complains a lot regarding scoping, I also prefer explicit scoping…
> 
> https://app.codacy.com/app/nextcloud/android/pullRequest?prid=3332947

Dagger requires package-level scoping, as it won't be able to inject fields. There is no explicit package-level scope, I'm affraid.

Is it possible to silence Codacy about `@Inject`-annotated elements?

AndyScherzinger:
> Is it possible to silence Codacy about @Inject-annotated elements? Alternatively we could use injection method

Unfortunately no, the injection method would also be package level and that would generate a codacy warning (or the method being public would solve it).

Can dagger do setter-injection?

ezaquarii:
> Can dagger do setter-injection?

Yes, it can. You annotate setters with `@Inject`. You'll end up with package-scoped setters this way.

ezaquarii:
BTW, I see that `Butterknife` does the same thing. Lack-of-keyword for a package scope was probably an unfortunate stylistic choice of Java creators and some people are so uncomfortable with it, that they prefix package-scoped elements with a comment:

```java
class Foo {
   /* package */ String member;
}
```

I understand why Codacy does it, as I myself find `private` and `protected` fields left package-scoped to be sort of a plague, but both `Dagger` and `ButerKnife` are actually valid exceptions. 

I also believe that they might be picked up by Codacy incorrectly, because the code that does access those fields is compile-time generated and not visible during static analysis.

AndyScherzinger:
I am fine with keeping it this way and I can also set any of these warnigns to ignore in codacy, while it will be on a per issue basis, not setting the while pattern to be ignored. (Which is also what we do for Butterknife).

Fine with you @tobiasKaminsky ? @ezaquarii ?

ezaquarii:
I'm not sure I understand you correctly. Does it mean you want to selectively disable it for a specific PR? I'm a bit confused...

AndyScherzinger:
> Does it mean you want to selectively disable it for a specific PR?

Yes, exactly :+1:

ezaquarii:
> Yes, exactly +1

Please bear in mind that you'll get quite a lot of those every time somebody touches injected fields (hint: upcoming refactorings).

Given that those fields are reliably identifiable by strict annotation pattern, I'm not convinced that this is the most convenient way to solve this - it feels more like a temporary workaround.

But if it works for you, I don't see a problem.

AndyScherzinger:
Yeah, it is probably rather a workaround but it would solve the issue for now and I am fine with manually maintaining that for now.
As long it helps turning your PRs to green I am fine doing it ;)

src/main/java/com/nextcloud/client/di/ComponentsModule.java
ezaquarii: Yes, for those fragments that require injection. I don't believe that we're using injectable base fragment, so there is no need to register the world.

tobiasKaminsky:
> `ComponentsModule` should be excluded from checks, as this is a Dagger-specifc DSL, not a normal class.

I think this is the way to go, and for the other injected fields in other classes we just keep the warning, but know that we can merge it.

AndyScherzinger:
Woohoo, merged 🎉 

Nice work as always @ezaquarii 👍

PR #3802, opened 2019-03-28
E2EE: directly encrypt/decrypt

tobiasKaminsky:
Previously it was first reading complete file into memory.

### DIFF
--- a/src/main/java/com/owncloud/android/utils/EncryptionUtils.java
+++ b/src/main/java/com/owncloud/android/utils/EncryptionUtils.java
@@ -38,15 +38,16 @@ import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.lib.resources.e2ee.GetMetadataRemoteOperation;
 
 import org.apache.commons.codec.binary.Hex;
+import org.spongycastle.util.io.Streams;
 
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.File;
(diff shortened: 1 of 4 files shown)


src/main/java/com/owncloud/android/utils/EncryptionUtils.java
```diff
@@ -328,30 +327,24 @@ public static EncryptedFile encryptFile(File file, byte[] encryptionKeyBytes, by
...
 
-        RandomAccessFile randomAccessFile = new RandomAccessFile(file, "r");
-        byte[] fileBytes = new byte[(int) randomAccessFile.length()];
-        randomAccessFile.readFully(fileBytes);
+        CipherInputStream cis = new CipherInputStream(new FileInputStream(file), cipher);
 
-        // check authentication tag
-        byte[] extractedAuthenticationTag = Arrays.copyOfRange(fileBytes,
-                fileBytes.length - (128 / 8), fileBytes.length);
+        FileOutputStream fileOutputStream1 = new FileOutputStream(output);
```
ezaquarii: This name looks like autogenerated by some bad generator. I'm sure we can avoid names like `someVar7`.

tobiasKaminsky: Thanks for your first review, however this is in dev by me and I mainly opened a PR to not forget about it ;-)
I'll enhance this prior setting it to "2. to review".

PR #3804, opened 2019-03-28
authentication: do not set unlock timestamp when prompting for passcode.

### LINKED ISSUE #3757: App lock easily bypassed
### Actual behaviour
In Nextcloud app setting under Lock, if Passcode is enabled, all you have to do is open the recent apps screen on the phone, close Nextcloud, relaunch Nextcloud and you're in. Video: https://youtu.be/WLRt3qUGThs

In Nextcloud app setting under Lock, if using Device Credentials (I'm using Fingerprint lock on my phone), all you have to do is press Back button quickly as the Nextcloud app is launching, do it a couple times and Nextcloud app will launch without asking for it. Video: https://youtu.be/x9rJH7XLof8

### Expected behaviour
App should stay locked unless pin or fingerprint is entered.
 
### Steps to reproduce
1. Listed above with video.

### Environment data
Android version: 9

Device model: Pixel 3 XL

Nextcloud app version: 3.5.0

Nextcloud server version: 15.0.5

ardevd:
Currently, every time an activity is stopped the `PassCodeManager` registers the current time and stores it as the last unlock time. This should not happen when the pass code activity is being stopped. To make matters worse, when the passcode activity is started whatever other activity was initially being started is then stopped which also registeres the last unlock time. By killing the app and restarting it before the 5 second timeout expires the lock will be bypassed entirely.

The only reason the lock works at all when not killing the app is simply because the PassCode activity is on top of the activity stack so the user is presented with it when resuming the app.

To combat the issue this commit introduces two changes.

1. Do not register unlock timestamp when the PassCodeActivity is stopped.
2. When the PassCodeActivity is started, set a `LOCK_PROMPTED` to `true`. When the PassCodeActivity is stopped we set the `LOCK_PROMPTED` to `false`.

Only when `LOCK_PROMPTED` is false and an activity other than the `PassCodeActivity` is stopped will we register the unlock time.

This fixes #3757

Signed-off-by: ardevd <edvard.holst@gmail.com>

### DIFF
--- a/src/main/java/com/owncloud/android/authentication/PassCodeManager.java
+++ b/src/main/java/com/owncloud/android/authentication/PassCodeManager.java
@@ -44,6 +44,8 @@ public final class PassCodeManager {
 
     public static final int PASSCODE_ACTIVITY = 9999;
 
+    private boolean LOCK_PROMPTED = false;
+
     static {
         exemptOfPasscodeActivities = new HashSet<>();
         exemptOfPasscodeActivities.add(PassCodeActivity.class);
@@ -84,6 +86,7 @@ public final class PassCodeManager {
             i.setAction(PassCodeActivity.ACTION_CHECK);
             i.setFlags(Intent.FLAG_ACTIVITY_REORDER_TO_FRONT);
             activity.startActivityForResult(i, PASSCODE_ACTIVITY);
+            LOCK_PROMPTED = true;
         }
 
         if (!exemptOfPasscodeActivities.contains(activity.getClass()) &&
@@ -100,7 +103,12 @@ public final class PassCodeManager {
         if (visibleActivitiesCounter > 0) {
             visibleActivitiesCounter--;
         }
-        setUnlockTimestamp(activity);
+        // Only set unlock timestamp if activity is not passcode related
+        if (!exemptOfPasscodeActivities.contains(activity.getClass()) && !LOCK_PROMPTED) {
+            setUnlockTimestamp(activity);
+        } else {
+            LOCK_PROMPTED = false;
+        }
         PowerManager powerMgr = (PowerManager) activity.getSystemService(Context.POWER_SERVICE);
         if ((isPassCodeEnabled() || deviceCredentialsAreEnabled(activity)) && powerMgr != null
                 && !powerMgr.isScreenOn()) {


ardevd:
Please test this and let me know what I've missed. It's been a long day so I'm relying on your keen eyes to double check my fix here.

tobiasKaminsky:
I tested it and when I click quickly back while opening the app I can still sometimes  directly access the files…

ardevd:
@tobiasKaminsky could you test with the two latest commits to this PR one by one? First check if it works without setting LOCK_PROMPTED to `true` by default and see if that works. If not, see what happens when we set LOCK_PROMPTED to `true` when instantiating it. 

The only downside I can imagine by setting LOCK_PROMPTED to true initially is the lock timer will not work in the following use case:

* Start the app, enter PIN. The FilesDisplay activity starts.
* Force kill the app
* Start the app
* Stop the app after > 5 seconds <-- unlock time will not be registered since `LOCK_PROMPTED` is now true.
* Start the app within 5 seconds <-- you will be prompted for PIN since the unlock time was not previously updated.

src/main/java/com/owncloud/android/authentication/PassCodeManager.java
```diff
@@ -44,6 +44,8 @@
 
     public static final int PASSCODE_ACTIVITY = 9999;
 
+    private boolean LOCK_PROMPTED = true;
```
ezaquarii: This formatting style and location suggests it is a constant. Pls change naming style and move it along other member variables around line 64.

src/main/java/com/owncloud/android/authentication/PassCodeManager.java
```diff
@@ -79,7 +81,7 @@ public void onActivityCreated(Activity activity) {
     public void onActivityStarted(Activity activity) {
         Long timestamp = PreferenceManager.fromContext(activity).getLockTimestamp();
         if (!exemptOfPasscodeActivities.contains(activity.getClass()) && passCodeShouldBeRequested(timestamp)) {
-
+            LOCK_PROMPTED = true;
```
ezaquarii: Again, formatting suggest this is a write to a const value... very confusing style.

ezaquarii:
Missing sign-off. You can add it with `git commit --amend -s` and force-push to update the PR.

src/main/java/com/owncloud/android/authentication/PassCodeManager.java
ardevd: thanks for pointing this out. Fixed.

ardevd:
@tobiasKaminsky did you get the chance to test this?

src/main/java/com/owncloud/android/authentication/PassCodeManager.java
```diff
@@ -59,6 +59,7 @@
 
     private AppPreferences preferences;
     private int visibleActivitiesCounter;
+    private boolean lockPrompted = true;
```
ezaquarii: I'm keen on finding if we can choose a more descriptive name for this variable. Maintainer will probably have no idea what a lock prompting concept is about.

ezaquarii: If you are unsure how to name it, it's a good idea to start with a JavaDoc. Once something is documented, it's easier to extract a meaningful terminology.

ezaquarii: I'm keen on finding if we can choose a more descriptive name for this variable. Maintainer will probably have no idea what a lock prompting concept is about.

If you are unsure how to name it, it's a good idea to start with a JavaDoc. Once something is documented, it's easier to extract a meaningful terminology.

ardevd: I agree. Firstly I'd like to ensure that it actually works though which is why I'm hoping someone can stress test it and report back.

ardevd:
@tobiasKaminsky bump

tobiasKaminsky:
I gave it a try and set it here to false:
https://github.com/nextcloud/android/blob/18a0929d6e8e5ddd20e436ffc22576e9919bcf09/src/main/java/com/owncloud/android/authentication/PassCodeManager.java#L83

Then when opening the app and you click back multiple times very fast and then re-open the app you can directly see file listing.

This also happens when lockPrompted is true.

ardevd:
hmmm, I'll have to test this on a real device since I'm unable to reproduce it in the emulator.

ardevd:
Glad you got it fixed!

PR #3815, opened 2019-03-31
activity: finish the activity when pressing back from the root folder.

### LINKED ISSUE #3811: Back-Navigation in File-Upload
### Actual behaviour
- Back-navigation in file upload activity does not work, it shows a storage chooser dialog

### Expected behaviour
- it goes back to previous activity
 
### Steps to reproduce
1. Open app
2. Click on '+'-Fab
3. Choose upload files
4. Click Back in navigation bar

### Environment data
Android version: 7.1, 1, 8.1.0, 9.0

Device model: 
Moto G5, Moto G4 Play, Samsung Tab S2

Stock or customized system:
Both

Nextcloud app version:
3.5.1, Dev 20190327

Nextcloud server version:
15.0

#### Web server error log
```
```

```
```

ardevd:
When browsing for files to upload a storage location selection prompt is shown when hitting the back toolbar button when in the root directory. It probably makes more sense to finish the activity.

Fixes #3811

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/UploadFilesActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/UploadFilesActivity.java
@@ -379,7 +379,7 @@ public class UploadFilesActivity extends FileActivity implements
 
             File parentFolder = mCurrentDir.getParentFile();
             if (!parentFolder.canRead()) {
-                showLocalStoragePathPickerDialog();
+                finish();
                 return;
             }
 


ardevd:
Not entirely sure how we can determine whether the folder is not accessible and not the root (of /sdcard) considering how that partition is mounted in a plethora of different ways across different devices.

tobiasKaminsky:
Right, this is indeed a bit tricky…
I am not sure when we want to 
a) show storage chooser
b) close activity

What about this:
- never show storage chooser
- close activity only when in "/"

So if you chose external sdcard, which might be in /emulated/1908-1233/ and you navigate up, you might see /emulated (read-only) with no content, but still can navigate up to "/".

tobiasKaminsky:
@AndyScherzinger we discussed this a while ago on irc…do you remember our conclusion?

AndyScherzinger:
> we discussed this a while ago on irc…do you remember our conclusion?

Unfortunately no :( But couldn't we just show the chooser except in cases where we can't navigate one-up -> then we close the screen.

ardevd:
> > we discussed this a while ago on irc…do you remember our conclusion?
> 
> Unfortunately no :( But couldn't we just show the chooser except in cases where we can't navigate one-up -> then we close the screen.

How do you differentiate between navigating one-up and navigating down?

AndyScherzinger:
> obviously, when pressing the back button you will be navigating up. So the current PR kinda works I guess?

Not sure because you can navigate one up (to the root) while you can't read it. So this PR will close the screen when being in the root folder instad of showing the chooser. _If_ you are in the root, canceled the chooser and then try to move one up (which is impossible of course) then I'd say we close the screen.

As for navigating down, well that is a different scenario in my opinion. _If_ you navigate down (into a non-readable folder) then we should rather either a) not navigate down but show a Snackbar that the folder is non-readable) or b) navigate into that folder and show an empty list with the information that  the folder is non-readable (navigatioin back up is then still possible).

What do you think @ardevd @tobiasKaminsky maybe also looping in @jancborchardt and @nextcloud/designers and please beware that the original solution proposed by @jancborchardt (displaying the choosers content instead of the file listing when hitting the root folder) is still valid but I never found the time to implement that since that would take more time than I had/have so we went for the actual implementation to at least solve the general problem and optimize for a nicer solution later.

tobiasKaminsky:
I am fine with either way.
For navigation down: please let us open up a new issue, if we further need to talk about it

AndyScherzinger:
> For navigation down: please let us open up a new issue, if we further need to talk about it

I also vote for separating that part

jancborchardt:
@ardevd could you post a screenshot for context?

Maybe it’s better then to indicate somewhere in the header of that dialog (like top right) which storage you are in (like icons for either "Internal storage" with a phone icon or "SD card" with SD card icon). Tapping it would open a picker for the storage, rather than having it as a parent view. That way you can also switch storages when somewhere lower down the hierarchy. Does that make sense?


> b) navigate into that folder and show an empty list with the information that the folder is non-readable (navigatioin back up is then still possible).

@AndyScherzinger agree with that one, better than the toast. But yes as has already been said it’s a separate issue. :)

PR #3818, opened 2019-03-31
datamodel: use equals for string comparison.

ardevd:
It is almost always a mistake to compare two instances of `java.lang.String` using reference equality == or !=, because it is not comparing the actual value but rather locations in memory.


src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -1471,7 +1471,7 @@ public void saveSharesDB(List<OCShare> shares) {
         // Reset flags & Remove shares for this files
         String filePath = "";
         for (OCShare share : shares) {
-            if (filePath != share.getPath()) {
+            if (filePath.equals(share.getPath())) {
```
ezaquarii: You've changed `!=` to `equals`. Are you sure the logic is good here?

ardevd: *Embarrassed*. Thank you. Nice catch. Fixed.

ezaquarii: And how do you test this code path?

AndyScherzinger:
@tobiasKaminsky to follwo-up on @ezaquarii's question: How can we get this to tested via a unit test? :)

ardevd:
@AndyScherzinger I can do some Powermock magic and write a unit test for this I think.

AndyScherzinger:
> @AndyScherzinger I can do some Powermock magic and write a unit test for this I think.

What do you think @tobiasKaminsky @ezaquarii? Powermock would be fine with me.

tobiasKaminsky:
Not sure if powermock is needed, as we already have mockito. 
So I would first try it with mockito. 

To test this function the only (maybe) tricky part is to catch applyBatch(…) but this should be possible with pure Mockito.

tobiasKaminsky:
Ping me, if you need any help :+1:

PR #3824, opened 2019-04-01
Use spotbugs for analytics

tobiasKaminsky:
Follow up of #3813, thanks to @ezaquarii!

SpotBugs finds a few more warnings, so I manually increased our results.


AndyScherzinger:
@tobiasKaminsky raise score and merge?

tobiasKaminsky:
> @tobiasKaminsky raise score and merge?

Score is already updated, otherwise SpotBug would have shown an error.

AndyScherzinger:
@tobiasKaminsky merged, does this have to be backported to 3.6? I guess so, right?

tobiasKaminsky:
No, I do not think so, as all the config/changes are within build.gradle / drone config and thus it should remain consistent.

PR #3839, opened 2019-04-03
Minor refactor of StringUtils class. Added unit tests.

ardevd:
Simplified the StringUtils class a bit and added unit tests for full test coverage (gotta start somewhere, right?).

- Removed android.text.TextUtils dependency
- Replaced silly string concatenation with String.fromat equivalent.
- Added unit tests.

### DIFF
--- a/src/main/java/com/owncloud/android/utils/StringUtils.java
+++ b/src/main/java/com/owncloud/android/utils/StringUtils.java
@@ -20,7 +20,6 @@
 
 package com.owncloud.android.utils;
 
-import android.text.TextUtils;
 
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -38,7 +37,7 @@ public final class StringUtils {
 
     public static String searchAndColor(String text, String searchText, @ColorInt int color) {
 
-        if (TextUtils.isEmpty(text) || TextUtils.isEmpty(searchText)) {
+        if (text.isEmpty() || searchText.isEmpty()) {
             return text;
         }
 
--- a/src/main/java/com/owncloud/android/utils/StringUtils.java
+++ b/src/main/java/com/owncloud/android/utils/StringUtils.java
@@ -38,23 +38,30 @@ public final class StringUtils {
 
     public static String searchAndColor(String text, String searchText, @ColorInt int color) {
 
-        if (text.isEmpty() || searchText.isEmpty()) {
-            return text;
+        if (text != null && searchText != null) {
+
+            if (text.isEmpty() || searchText.isEmpty()) {
+                return text;
+            }
+
+            Matcher matcher = Pattern.compile(searchText,
+                                              Pattern.CASE_INSENSITIVE | Pattern.LITERAL).matcher(text);
+
+            StringBuffer stringBuffer = new StringBuffer();
+
+            while (matcher.find()) {
+                String replacement = matcher.group().replace(
+                    matcher.group(),
+                    String.format(Locale.getDefault(), "<font color='%d'><b>%s</b></font>", color,
+                                  matcher.group())
+                );
+                matcher.appendReplacement(stringBuffer, Matcher.quoteReplacement(replacement));
+            }
+            matcher.appendTail(stringBuffer);
+
+            return stringBuffer.toString();
+        } else {
+            return "";
         }
-
-        Matcher matcher = Pattern.compile(searchText, Pattern.CASE_INSENSITIVE | Pattern.LITERAL).matcher(text);
-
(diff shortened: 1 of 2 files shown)


src/main/java/com/owncloud/android/utils/StringUtils.java
```diff
@@ -38,7 +38,7 @@ private StringUtils() {
 
     public static String searchAndColor(String text, String searchText, @ColorInt int color) {
 
-        if (TextUtils.isEmpty(text) || TextUtils.isEmpty(searchText)) {
+        if (text.isEmpty() || searchText.isEmpty()) {
```
AndyScherzinger: @ardevd this change removes the null-safety! Since TextUtils explicitely takes care of that!

ardevd: Good catch. I'm assuming with the TextUtils implementation, the `searchAndColor` method would return an empty string if `text` or `searchText` is null?

AndyScherzinger: Well, looking at the code, `searchAndColor` would simply return `text`, so either its value or null in case it is null.

ardevd: If the previous implementation could return `null` as you suggest then ideally you should check for that when calling it in `PreviewTextFragment`. 

So either, we can have null checks and return null, which would be similar to the previous implementation, or we can return empty strings. If the former, we should annotate the method as `@Nullable` and with the latter we should annotate it with `@NonNull` then I guess.

AndyScherzinger: Any preference? Since it is a utilities implementation I tend to say it needs to be null-safe since utilities methods never crash but basically be able to cope with any input. On the other hand I can't really say if it makes sense at all looking at the return value so maybe non-null annotation makes more sense in this case.

What do you think?
Also looping in @tobiasKaminsky @ezaquarii for feedback and opinion

tobiasKaminsky: I think it should be capable of handling also null input, but in that case return non-null, but an empty string "".

tobiasKaminsky: Otherwise we would get a crash when handling using return value

src/test/java/com/owncloud/android/utils/StringUtilsTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class StringUtilsTest {
+
+    @Test
+    public void searchAndColor_assertCorrectSingleSearchMatch() {
+
+        String text = "this is a simple test";
+        String searchText = "simple";
+        int dummyColorInt = 44221;
+        String expectedReturn = String.format("this is a <font color='%d'><b>%s</b></font> test", dummyColorInt, searchText);
```
tobiasKaminsky: Please wrap at 120 chars.

src/test/java/com/owncloud/android/utils/StringUtilsTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+        assertEquals("correctly parsed text returned",
```
tobiasKaminsky: According to doc:
> Asserts that two objects are equal. If they are not, an AssertionError is thrown with the given message. If expected and actual are null, they are considered equal.

the string message is to explain what happens if the assertion is false (error message), so in this case I would suggest "Parsed text is not correct".

src/main/java/com/owncloud/android/utils/StringUtils.java
ardevd: agreed with @tobiasKaminsky

src/test/java/com/owncloud/android/utils/StringUtilsTest.java
ardevd: Spot on! :)

src/main/java/com/owncloud/android/utils/StringUtils.java
```diff
@@ -38,23 +38,30 @@ private StringUtils() {
 
     public static String searchAndColor(String text, String searchText, @ColorInt int color) {
```
tobiasKaminsky:
```suggestion
    public static @NonNull String searchAndColor(@Nullable String text, @Nullable String searchText, @ColorInt int color) {
```

tobiasKaminsky:
I added some null and empty tests and fixed one issue when searchText was null/empty.

tobiasKaminsky:
Thank you for your work!

PR #3841, opened 2019-04-03
Unit tests for the OwnCloudSession class.

ardevd:
- Added unit tests for the OwnCloudSession class. 
- Restricted access declarations for the OwnCloudSession class. Follow the principle of least privileged.

### DIFF
--- a/src/main/java/com/owncloud/android/utils/OwnCloudSession.java
+++ b/src/main/java/com/owncloud/android/utils/OwnCloudSession.java
@@ -27,7 +27,7 @@ public class OwnCloudSession {
     private String mSessionUrl;
     private int mEntryId;
 
-    public OwnCloudSession(String name, String url, int entryId) {
+    OwnCloudSession(String name, String url, int entryId) {
         mSessionName = name;
         mSessionUrl = url;
         mEntryId = entryId;
@@ -49,7 +49,7 @@ public class OwnCloudSession {
         return mSessionUrl;
     }
 
-    public int getEntryId() {
+    int getEntryId() {
         return mEntryId;
     }
 }
--- a/src/main/java/com/owncloud/android/utils/OwnCloudSession.java
+++ b/src/main/java/com/owncloud/android/utils/OwnCloudSession.java
@@ -1,4 +1,4 @@
-/**
+/*
  * ownCloud Android client application
  *
  * @author Bartek Przybylski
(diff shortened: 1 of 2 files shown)


ezaquarii:
If `OwnCloudSession` is not used outside the package, it should not be exposed as public interface.

I'd turn that rule off honestly, as I find package scope is very handy in hiding internal package guts (helpers, utils, etc) that should not be used outside.

tobiasKaminsky:
Totally agree. 
We should limit the scope as needed, and then can extend it again if needed.
Package scope is a valid approach and thus we should use if, when it fits.

tobiasKaminsky:
@ardevd I added gpl3+ license to unit tests, I hope this is ok?
I also changed OwnCloudSession to use lombok, so we do not need to take care of boilerplate.
I hope you are fine with my changes.

ardevd:
Looks good to me.

tobiasKaminsky:
@ardevd codacy is complaining about underscore in test method names.
while I really like your naming, maybe a simple camel case is also sufficient? :-)

ardevd:
@tobiasKaminsky Well thanks. I'm just following what I've come to believe is the styling guidelines set my Google for test method names and is what I've used for ages. If you want to please Codacy then I won't object. Ideally I'd have an exception in the Codacy rule for test classes though. 

Up to you guys.

tobiasKaminsky:
> Ideally I'd have an exception in the Codacy rule for test classes though.

@AndyScherzinger is this possible?

AndyScherzinger:
> @AndyScherzinger is this possible?

we can exclude testclasses (via the package) from codacy check in general for _any_ kind of check. Codacy doesn't offer more fine grained ways to disable certain checks for certain naming patterns though. We already ignore anything wihtin the test folders while sometimes codacy goes nuts and uses @dan0xii's codacy accoutn/definition and pushes it into the PRs within this repo. So Warnings in test classes can be ignored since they come from the wrong codacy account.

AndyScherzinger:
@tobiasKaminsky any idea why the spoptbugs analysis fails within this PR?

AndyScherzinger:
aaaaaand merged 🎉

tobiasKaminsky:
> while sometimes codacy goes nuts and uses @dan0xii's codacy accoutn/definition

I asked codacy and they created a support ticket for this.
This seems to be indeed a problem, as normally a forked project is not related to the original one…

AndyScherzinger:
Nice, so let's see if this can get resolved 🎉

PR #3850, opened 2019-04-05
removed unused imports.

ardevd:
Leaving unused imports in reduces the code's readability, since their presence can be confusing.


tobiasKaminsky:
Nice catch :tada: 
Note to whoever removed the corresponding code: this can be added as a step in Android Studio to "organize imports"

PR #3853, opened 2019-04-05
fix drone.yml for drone 1.0

tobiasKaminsky:
Signed-off-by: Morris Jobke <hey@morrisjobke.de>

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,68 +1,108 @@
-pipeline:
-    compile:
-        image: nextcloudci/android:android-44
-        commands:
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@tobiasKaminsky seems to lead to an endless loop for the analysis task :(

tobiasKaminsky:
Strange.
I will have a look into it.
Also the "notify" part is removed, which I will also include again.

tobiasKaminsky:
Notify added.
Seems to work.

PR #3857, opened 2019-04-07
About resolving #2953

### LINKED ISSUE #2953: Update account list to recyclerview
> The account list is still a list view so we'd need to rewrite it to be a recyclerview to get rid of the dividers (can be done and should be done of course).

- update account list to recyclerview
- remove dividers

nantoniou:
Fix #2953 

Hello, I have been working on #2953 and it is almost ready.
However, the click listener on the add account button is not working, due to the different handling of on click listeners from RecyclerView, which, as far as I understand, binds the listener to a view and not the ListView.
Thus, I may have to edit and/ or move the createAccount on the ManageAccountsActivity method, but I am not sure. So any help would be appreciated and excuse my newbie question!

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/ManageAccountsActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/ManageAccountsActivity.java
@@ -130,7 +130,7 @@ public class ManageAccountsActivity extends FileActivity
 
         arbitraryDataProvider = new ArbitraryDataProvider(getContentResolver());
 
-        mAccountListAdapter = new AccountListAdapter(this, getUserAccountManager(), getAccountListItems(), mTintedCheck);
+        mAccountListAdapter = new AccountListAdapter(this, accountManager, getAccountListItems(), mTintedCheck);
 
         mRecyclerView.setAdapter(mAccountListAdapter);
         mRecyclerView.setLayoutManager(new LinearLayoutManager(this));
@@ -272,10 +272,10 @@ public class ManageAccountsActivity extends FileActivity
                               try {
                                   Bundle result = future.getResult();
                                   String name = result.getString(AccountManager.KEY_ACCOUNT_NAME);
-                                  AccountUtils.setCurrentOwnCloudAccount(getApplicationContext(), name);
+                                  accountManager.setCurrentOwnCloudAccount(name);
                                   mAccountListAdapter = new AccountListAdapter(
                                           this,
-                                          getUserAccountManager(),
+                                          accountManager,
                                           getAccountListItems(),
                                           mTintedCheck
                                   );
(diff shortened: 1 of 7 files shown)


tobiasKaminsky:
It should be possible to set an on click listener here: https://github.com/nextcloud/android/pull/3857/files#diff-a0ffffde3b457c958686e0b7c17948beR106

On master we use an interface for communication between AccountListAdapter and ManageAccounsActivity, this should also be possible for you.

tobiasKaminsky:
> excuse my newbie question!

You absolutely do not need to apologize!
It is great that you pick up this issue and want to resolve it :tada: 
Thanks for your work so far :-)

nantoniou:
@tobiasKaminsky Thank you for the quick response. I think I should (try to) clarify my problem. On the activity method, [here](https://github.com/nextcloud/android/blob/30cad20f86f25fa8f8152f370414ef52af3798c8/src/main/java/com/owncloud/android/ui/activity/ManageAccountsActivity.java#L133), we set the adapter that contains the click listener to the ListView. And then on line 137 we set the ItemClickListener to the ListView.
However, in RecyclerView we set the ItemClickListener to the specific View, in line 124 in AccountListAdapter in my commit. So I think that it is not using the methods from the AccountListAdapterListener that we have in the Activity class and that's why it is not working, even though the OnClickListener is there (in the onBindViewHolder).
Changing from `holder.itemView.setOnClickListener` to `mRecyclerView.setOnClickListener` in line 124 makes the click listener work for the "add account" and not the account itself- and I think is bad practice in the RecyclerView.
So do you have any suggestions? (it's probably an easy solution)

tobiasKaminsky:
We talk about this?
https://github.com/nantoniou/android/blob/7a787858736f5224abe52c3d9195a705f94c662b/src/main/java/com/owncloud/android/ui/adapter/AccountListAdapter.java#L123-L140

Then I would do it also with an interface like:
https://github.com/nantoniou/android/blob/7a787858736f5224abe52c3d9195a705f94c662b/src/main/java/com/owncloud/android/ui/adapter/AccountListAdapter.java#L164-L168

Then all three real methods are in ManageAccountsActivity, but are still accessible via AccountListAdapter.

nantoniou:
Thank you for the feedback again @tobiasKaminsky. The problem was a simple error in my code; it is fixed now. I also cleaned up some parts of the pre-existing code and added some minor documentation for some methods. Can I also add a test class?

tobiasKaminsky:
> Can I also add a test class?

This would be awesome!

nantoniou:
@tobiasKaminsky I added some basic tests. The PR is ready for review now. I'm not quite sure why Spotbugs is making the build Analysis fail though.

AndyScherzinger:
@tobiasKaminsky @nantoniou I took the liberty to finish-up this PR (at least from my point of view 😉) 

* I added a view swith for the add-accounts part, 
* took care of the theming, 
* cleaned up the variable names (no m-prefixes in all touches classes)

Tested on an OP3T/Android 9 with multiple accounts from 2 servers

PR #3859, opened 2019-04-08
tests: added unit tests for the ConnectivityUtils util class.

ardevd:
The `isOnlineWithWifi()` method is fully covered by these tests.

The `isInternetWalled()` method will require use of PowerMock to mock the static `AccountUtils.getCurrentOwnCloudAccount(context);` method. I'll deal with that later. 

Also, the ConnectivityUtils class is riddled with use of deprecated API's. Needs a make over. I'll look into it sometime if nobody else does it.

### DIFF
--- a/src/test/java/com/owncloud/android/utils/ConnectivityUtilsTest.java
+++ b/src/test/java/com/owncloud/android/utils/ConnectivityUtilsTest.java
@@ -1,0 +1,65 @@
+package com.owncloud.android.utils;
+
+import android.content.Context;
+import android.net.ConnectivityManager;
+import android.net.NetworkInfo;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+import static org.junit.Assert.*;
+import static org.mockito.ArgumentMatchers.contains;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class ConnectivityUtilsTest {
+
+    @Mock
+    private Context mContext;
+
+    @Mock
+    private ConnectivityManager mConnectivityManager;
+
+    @Mock
+    private NetworkInfo mNetworkInfo;
+
+    @Before
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
+        when(mContext.getSystemService(Context.CONNECTIVITY_SERVICE)).thenReturn(mConnectivityManager);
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
> The `isInternetWalled()` method will require use of PowerMock to mock the static `AccountUtils.getCurrentOwnCloudAccount(context);` method. I'll deal with that later.

Nice. I fear that we will have these static calls more often while testing…
Do you imagine another way? 



> Also, the ConnectivityUtils class is riddled with use of deprecated API's. Needs a make over. I'll look into it sometime if nobody else does it.

This would be awesome! But please keep our min supported sdk in mind :-)

ardevd:
> Nice. I fear that we will have these static calls more often while testing…
> Do you imagine another way?

Unfortunately, the code base is, as you know, not exactly structured to be testable. PowerMock gives us the tool to write tests for a lot of things we would not be able to do with regular Mockito. 

PowerMock is great though :)

tobiasKaminsky:
Then let us try to do it in two steps, first use power mock to test those static calls, and after that refactor so that we can test it better (e.g. change to a more testable structure).

Thank you very much for taking care of this!

Will you do the power mock test in a new PR? Then we can merge this?

ardevd:
I'll do it in a separate PR I reckon. 

I find writing tests extremely satisfying so I'm happy to do it. It's also generally a good idea to have someone write the code and someone else write the tests which is exactly what we're doing here ;)

ardevd:
So yeah, merge away!

tobiasKaminsky:
> So yeah, merge away!

I'll do once codacy/drone is satisfied. 
There are currently some hiccups, which I'll resolve.



> I find writing tests extremely satisfying so I'm happy to do it. It's also generally a good idea to have someone write the code and someone else write the tests which is exactly what we're doing here ;)

Cool, really really glad that you are working with us :-D

ardevd:
Happy to be of assistance. :)

tobiasKaminsky:
@AndyScherzinger codacy is complaining about "Method names should not contain underscores", which I tried to disable, but I cannot find it.
I think for tests it is a good way to use it.

AndyScherzinger:
@tobiasKaminsky I'll have a look later on, just merged it waiting for codacy to pick it up on the master branch and then I'll set them to ignore for the issues above. We can only deactivate it completely (the rule) or deactivate codacy for test packages (which I wouldn't do). So I'll maintain this manually on a per PR basis for now (fine with me!).

PR #3864, opened 2019-04-09
test: added PowerMocked unit tests for the ConnectivityUtils class.

ardevd:
Added necessary PowerMock dependencies in addition to a JSON dependency which is required in order to unit test with JSON objects. The default JSON library is part of AOSP and would need to be mocked which is not what we want.

Note: the test currently fails. The `isInternetWalled` method has a weird check that I do not quite understand:

https://github.com/nextcloud/android/blob/b252c6440d196409076b7a00c83ef4271cbae1ef/src/main/java/com/owncloud/android/utils/ConnectivityUtils.java#L76-L89

So only if we're on NextCloud v13 we do the `else` block?

My test fails if I set version to `OwnCloudVersion.nextcloud_14` and true if I set it to `OwnCloudVersion.nextcloud_13`.

Could someone shed some light on the intended functionality here? I'm confused.


ardevd:
@tobiasKaminsky Nvm, I'm fairly sure `if (serverVersion.compareTo(OwnCloudVersion.nextcloud_13) > 0)` will only be true for NextCloud versions > 13. With this assumption I'll update the tests and this PR.

ardevd:
this signoff requirement is driving me nuts. can someone fix this? My git fu is not strong enough it seems.

tobiasKaminsky:
> @tobiasKaminsky Thanks for the explanation. But what about version > NC13?

Sorry, that was a typo on my side, it is >= NC13

tobiasKaminsky:
> Sorry, that was a typo on my side, it is >= NC13

Mockup seems to have problem when using ">=" in a list. I added ">=" above, but only "=" was shown.

src/test/java/com/owncloud/android/utils/ConnectivityUtilsTest.java
```diff
@@ -45,11 +64,27 @@
     @Mock
     private NetworkInfo mNetworkInfo;
 
+    @Mock
+    private Account mAccount;
```
tobiasKaminsky: We try to get rid of the m prefix everywhere.
Can you remove them here? So we do not introduce now ones :-)

tobiasKaminsky:
:+1: nice, so now we have a blue print how to mock static methods.

src/test/java/com/owncloud/android/utils/ConnectivityUtilsTest.java
ardevd: done. :)

src/test/java/com/owncloud/android/utils/BitmapUtilsTest.java
```diff
@@ -0,0 +1,24 @@
...
+import java.net.URL;
+
+import static org.junit.Assert.*;
+
+public class BitmapUtilsTest {
+
+
+    @Test
+    public void calculateHSL_assertCorrectRGBArrayReturned() throws Exception {
+        int[] rgbArray = BitmapUtils.calculateHSL("Edvard");
```
tobiasKaminsky: This should be a real test with assert?

src/test/java/com/owncloud/android/utils/BitmapUtilsTest.java
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    @Test
+    public void HSLtoRGB_assertCorrectRGBArrayReturned() throws Exception {
+        int[] rgbArray = BitmapUtils.HSLtoRGB(300.0f, 90.0f, 99.0f, 1.0f);
+        String format = "hello";
```
tobiasKaminsky: Same here

ardevd: oopsies. that class should not have been committed. Removed.

src/test/java/com/owncloud/android/utils/BitmapUtilsTest.java
ardevd: my bad.

PR #3871, opened 2019-04-10
Show current codacy issue count

tobiasKaminsky:


AndyScherzinger:
@tobiasKaminsky Works fine it seems whiel we only see the master's current codacy wanrings count. So I think this is ready to be merged.

tobiasKaminsky:
True, but we should also see codacy's own status report, which will (on success) be 0 changes or -x changes. 
But at least we know how many open issues we have left.

PR #3879, opened 2019-04-11
PMD ruleset

tobiasKaminsky:
- rename ruleset
- add test_test to MainApp to test Codacy

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -169,7 +169,7 @@ android {
     }
 
     tasks.register("pmd", Pmd) {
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
I did a test and it failed to complain about test_test in Mainapp.java, but on the second test it worked.
After all it did not complained about _ methods in Test, so I guess we can merge it.

(Hopefully it will still use all other checks, and not only those 5 in ruleset.xml)

AndyScherzinger:
let's give it a try then 👍

AndyScherzinger:
I changed codacy to look for a ruleset so let's see what happens

PR #3889, opened 2019-04-12
Cleanup authentication

tobiasKaminsky:
- removed useNextcloudUserAgent -> we now use this everytime
- removed any oAuth and saml stuff, as we rely on weblogin flow


AndyScherzinger:
Fine with me, will kill any use in combination with a owncloud backend but I am fine with this since we never officially supported it anyways.

PR #3903, opened 2019-04-15
Update wording for store text

jancborchardt:
Please double-check @jospoortvliet @shironextcloud @nextcloud/android, I simply adjusted it to use the same wording as on the frontpage of https://nextcloud.com

Subline might need to be changed based on if we change the sub-headline wording: https://github.com/nextcloud/nextcloud.com/pull/1093

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -660,8 +660,8 @@
     <string name="resized_image_not_possible">No resized image available.</string>
     <string name="resized_image_not_possible_download">No resized image available. Download full image?</string>
 
-    <string name="store_short_desc">Protecting your data</string>
-    <string name="store_full_desc">Building self-hosted products that allow you to be productive without losing control.\n\nFeatures:\n* Easy, modern interface, suited to the theme of your server\n* Upload files to your Nextcloud server\n* Share them with others\n* Keep your favorite files and folders synced\n* Search across all folders on your server\n* Auto Upload for photos and videos taken by your device\n* Keep up to date with notifications\n* Multi-account support\n* Secure access to your data with fingerprint or PIN\n* Integration with DAVx5 (formerly known as DAVdroid) for easy setup of calendar &amp; Contacts synchronization\n\nPlease report all issues at https://github.com/nextcloud/android/issues and discuss this app at https://help.nextcloud.com/c/clients/android\n\nNew to Nextcloud? Nextcloud is a private file sync &amp; share and communication server. It is libre software, and you can host it yourself or pay a company to do it for you. That way, you are in control of your photos, your calendar and contact data, your documents and everything else.\n\nCheck out Nextcloud at https://nextcloud.com</string>
+    <string name="store_short_desc">The self-hosted productivity platform that keeps you in control</string>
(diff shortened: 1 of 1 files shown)


shironextcloud:
Looks good to me, what do you think @jospoortvliet ?

tobiasKaminsky:
These values are translated via Transifex and used for f-droid (gplay is done manually currently):

Short description:
image


Full description:
image

Do you think this fits?

jancborchardt:
Updated with new wording from frontpage as discussed and merged in https://github.com/nextcloud/nextcloud.com/pull/1093

What do you think @tobiasKaminsky @AndyScherzinger? I also used the short_desc sentence to start out the long_desc as on some platforms (like F-Droid) the short desc is not fully shown.

src/main/res/values/strings.xml
```diff
@@ -660,8 +660,8 @@
...
+    <string name="store_short_desc">The self-hosted productivity platform that keeps you in control</string>
```
comradekingu:
```suggestion
    <string name="store_short_desc">The self-hosted productivity platform keeping you in control</string>
```

src/main/res/values/strings.xml
```diff
@@ -690,8 +690,8 @@
     <string name="forward">Forward</string>
     <string name="what_s_new_image">What\'s new image</string>
 
-    <string name="store_short_dev_desc">The Nextcloud Dev app is a development snapshot and can be installed parallel.</string>
-    <string name="store_full_dev_desc">The copylefted libre Nextcloud Android app, gives you mobile freedom.\nThis is the official development version, featuring a daily sample of any new untested functionality, which may cause instability and data loss. The app is for users willing to test, and report bugs should they occur. Do not use it for your productive work!\n\nBoth the official dev and regular version are available on F-droid, and can be installed at the same time.</string>
+    <string name="store_short_dev_desc">The self-hosted productivity platform that keeps you in control (dev preview version)</string>
```
comradekingu:
```suggestion
    <string name="store_short_dev_desc">The self-hosted productivity platform keeping you in control (dev preview version)</string>
```

comradekingu:
I still think it is warranted to point out the app is copylefted, seeing as it is the only differentiating quality to anyone else making the same claims for themselves.

jancborchardt:
> I still think it is warranted to point out the app is copylefted, seeing as it is the only differentiating quality to anyone else making the same claims for themselves.
> Edit: As it relates to freedom, or specifically "control".

- I have never seen the word "copylefted" anywhere else, and find it difficult to understand even as someone who can guess what it means.
- "self-hosted" is the word giving agency/control here, and for people who know what this means they can likely infer it is open source. We could also use "open source" instead of "self-hosted" here, I don’t really mind.
- If you see this on F-Droid, you know it’s open source / free software / licensed under a copyleft license  anyway.

> The Android app is also not the "self-hosted platform", it is the client.

Nextcloud is never just one of the apps or clients, it’s always the full platform. Also this line is a general tagline to use across apps. We wouldn’t write "The open source Android app to the self-hosted productivity platform that keeps you in control." – because that’s just a mouthful, and people always get the full product.

Sorry but I feel often these detail wording discussions and needing to explain this are very draining. This was already the case on the Include program subsite. You clearly come from a different background and seem to want to speak to different people than we want to speak to, so I don’t know how valuable the discussion is.

comradekingu:
Feel free to skip any or all of this.

For how difficult to understand and concise a word "[copyleft](https://en.wikipedia.org/wiki/Copyleft)/[ed](https://en.wiktionary.org/wiki/copylefted)" is, it is the most prominent differentiating quality to that of other solutions. Many of which are self-hosted, and also have some marketing explaining how the user is (supposedly) in charge of things. "Self-hosted" does however not mean _open source_, and as you may well know, open source does not necessitate being copyleft. (Copyleft is the additional (free software freedom 3) clause ensuring continued reciprocal sharing.)

However one wants to word it, it is a meaningful way to distinguish how it not only copyleft is different, but also how it will remain so. As opposed to just Open Source/Free Software/Libre Software.

AGPLv3 preamble (from nextcloud.com license)

> The GNU Affero General Public License is a free, copyleft license for
> software and other kinds of works, specifically designed to ensure
> cooperation with the community in the case of network server software.

https://nextcloudappstore.readthedocs.io/en/latest/ cuts right to the chase:

> The App Store is available under the AGPLv3 or later which offers a strong copyleft so developers can profit from changes made in other versions

You don't actually know what my background is I think, just what I currently do. I digress.

There are many potential users for whom using Nextcloud is new, the bulk of which probably share very little in common with either of us, or the we of Nextcloud at large. It seemed fair to me to assume some might mistake a client app for something that also acts as a server if it purports to be, or even as a base assumption. The same reasoning would apply to not getting it from F-Droid, or not knowing what F-Droid is about.

They may not know what a server even is, but an all-in-one app fitting the description is a possible mental construction. When "The cloud" helps itself to user data and hides any language to that effect, why should Nextcloud try to fit that same mental model?

shironextcloud:
@comradekingu, as Jan pointed out very politely, such conversations you keep pushing after we already refused your wording edits are really draining. This discussion is now closed for us, and we will ignore all further comments of yours on this thread to focus on something more constructive.

Issue #3901, opened 2019-04-15
java.lang.NullPointerException in DocumentsProvider


jmue:
### Actual behaviour
- account is not shown in DocumentsUI as result of java.lang.NullPointerException

### Expected behaviour
- no crash should happen
 
### Steps to reproduce
1. Open StoragePicker after fresh deployment on device

### Environment data
Android version: 8.1

Device model: Android Emulator

Stock or customized system:

Nextcloud app version: dev-20190414

Nextcloud server version:

#### Web server error log
```
Insert your webserver log here
```

#### Nextcloud log (data/nextcloud.log)
```
2019-04-15 12:59:33.492 27887-27919/com.nextcloud.client E/DatabaseUtils: Writing exception to parcel
    java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String android.content.Context.getPackageName()' on a null object reference
        at android.preference.PreferenceManager.getDefaultSharedPreferencesName(PreferenceManager.java:537)
        at android.preference.PreferenceManager.getDefaultSharedPreferences(PreferenceManager.java:526)
        at com.nextcloud.client.preferences.AppPreferencesImpl.fromContext(AppPreferencesImpl.java:76)
        at com.owncloud.android.providers.DocumentsStorageProvider.queryRoots(DocumentsStorageProvider.java:108)
        at android.provider.DocumentsProvider.query(DocumentsProvider.java:786)
        at android.content.ContentProvider$Transport.query(ContentProvider.java:240)
        at android.content.ContentProviderNative.onTransact(ContentProviderNative.java:102)
        at android.os.Binder.execTransact(Binder.java:697)
2019-04-15 12:59:33.753 27887-27919/com.nextcloud.client E/DatabaseUtils: Writing exception to parcel
    java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String android.content.Context.getPackageName()' on a null object reference
        at android.preference.PreferenceManager.getDefaultSharedPreferencesName(PreferenceManager.java:537)
        at android.preference.PreferenceManager.getDefaultSharedPreferences(PreferenceManager.java:526)
        at com.nextcloud.client.preferences.AppPreferencesImpl.fromContext(AppPreferencesImpl.java:76)
        at com.owncloud.android.providers.DocumentsStorageProvider.queryRoots(DocumentsStorageProvider.java:108)
        at android.provider.DocumentsProvider.query(DocumentsProvider.java:786)
[...]

PR #3907, opened 2019-04-15
Initialize global context before MainApp.onCreate()

ezaquarii:

### DIFF
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -49,7 +49,6 @@ import com.nextcloud.client.di.ActivityInjector;
 import com.nextcloud.client.di.DaggerAppComponent;
 import com.nextcloud.client.preferences.AppPreferences;
 import com.nextcloud.client.preferences.AppPreferencesImpl;
-import com.owncloud.android.authentication.AccountUtils;
 import com.owncloud.android.authentication.PassCodeManager;
 import com.owncloud.android.datamodel.ArbitraryDataProvider;
 import com.owncloud.android.datamodel.MediaFolder;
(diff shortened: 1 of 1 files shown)


ezaquarii:
This could fix  #3901 and possibly some other issues.

build.gradle
```diff
@@ -247,9 +247,9 @@ dependencies {
     // dependencies for app building
     implementation 'androidx.multidex:multidex:2.0.1'
 //    implementation project('nextcloud-android-library')
-    genericImplementation 'com.github.nextcloud:android-library:master-SNAPSHOT'
-    gplayImplementation 'com.github.nextcloud:android-library:master-SNAPSHOT'
-    versionDevImplementation 'com.github.nextcloud:android-library:master-SNAPSHOT'
+    genericImplementation 'com.github.nextcloud:android-library:1.4.0'
```
ezaquarii: Sorry, you'll need to sort it out tomorrow. I'll leave it like this to compile.

mario:
Can I ask why not a module and then provideContext which can be injected everywhere to get application-wide context?

src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -164,9 +165,17 @@
     @SuppressWarnings("unused")
     private boolean mBound;
 
+    /**
+     * Temporary hack
+     */
+    private static void initGlobalContext(Context context) {
+        mContext = context;
```
ezaquarii: `getApplicationContext()` is `null`, as it is set somewhere between `Application.attachBaseContext()` and `Application.onCreate()`. However, because there is only 1 instance of `MainApp`, there is no leak, so no worries.

ezaquarii:
> Can I ask why not a module and then provideContext which can be injected everywhere to get application-wide context?

It is already done.

However, large portions of the code still call to `MainApp.getAppContext()` either directly or indirectly via other helper and it will be a long march to fix this.

PR #3909, opened 2019-04-16
Ensure that userId is always present

tobiasKaminsky:
- migrate userId
- always fetch userId on account creation

To clarify naming, as it is really confusing:
userId: internally id, needed for e.g. dav/trashbin/$userId/
loginName: used for authenticate on every login/network connection, determinated by first login (weblogin/old login): can be anything: email, name, name with withespaces
displayName: Name chosen by user, only displayed, no usage at all

I tried to have them now consistent across app, lib and SSO.

Side step:
removed dynamic session: simpleFactoryManager was used <OC9.1, so directly use SingleSessionManager and named it OwnCloudClientManager

Resolves:  #4006


src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
```diff
@@ -69,34 +71,30 @@ public Account getCurrentAccount() {
         return AccountUtils.getCurrentOwnCloudAccount(context);
     }
 
-    public void updateAccountVersion() {
-        Account currentAccount = AccountUtils.getCurrentOwnCloudAccount(context);
+    public void migrateUserId() {
+        boolean success = false;
```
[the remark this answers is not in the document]
AndyScherzinger: valid point

AndyScherzinger:
needs a rebase and codacy fix, otherwise fine with me

AndyScherzinger:
Checks are still not happy :/

src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
```diff
@@ -80,34 +82,30 @@ public Account getAccountByName(String name) {
         return null;
     }
 
-    public void updateAccountVersion() {
-        Account currentAccount = AccountUtils.getCurrentOwnCloudAccount(context);
+    public void migrateUserId() {
+        boolean success = false;
```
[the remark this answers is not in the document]
AndyScherzinger: fixed

src/androidTest/java/com/nextcloud/client/account/UserAccountManagerImplTest.java
```diff
@@ -0,0 +1,70 @@
...
+//        }
+//
+//        assertEquals(0, sut.getAccounts().length);
+//    }
+
+    @Test
+    public void updateOneAccount() {
+        AppPreferences appPreferences = AppPreferencesImpl.fromContext(targetContext);
+        UserAccountManagerImpl sut = new UserAccountManagerImpl(targetContext, accountManager);
+        assertEquals(1, sut.getAccounts().length);
```
AndyScherzinger: @tobiasKaminsky this test fails since `actual == 2`

src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -186,6 +186,12 @@ public void onCreate() {
 
         registerActivityLifecycleCallbacks(new ActivityInjector());
 
+        Thread t = new Thread(() -> {
```
ezaquarii: I smell subtle race conditions here. Can we make it a blocking call? It will run only once.

src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
```diff
@@ -80,34 +82,28 @@ public Account getAccountByName(String name) {
         return null;
     }
 
-    public void updateAccountVersion() {
-        Account currentAccount = AccountUtils.getCurrentOwnCloudAccount(context);
+    public void migrateUserId() {
+        AppPreferences appPreferences = AppPreferencesImpl.fromContext(context);
```
ezaquarii: I'm afraid this will create a circular dependency, as app references already depend on account manager.
IMO this should be moved into migration class/service/etc.

AndyScherzinger:
Fine with me to fix the review comments after merging the PR. What I can't say looking at the failing test is if it is save to merge since the test needs to be fixed or any future build will still fail due to the test not the compile issue though.

ezaquarii:
> Fine with me to fix the review comments after merging the PR. What I can't say looking at the failing test is if it is save to merge since the test needs to be fixed or any future build will still fail due to the test not the compile issue though.

Yea, tests are a sure thing - I just don't want to get @tobiasKaminsky into a rabbit hole of week  of further refactoring.

src/main/java/com/owncloud/android/MainApp.java
tobiasKaminsky: Depending how many accounts/how slow the network connection is, this will then freeze the UI, or?

src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
tobiasKaminsky: AppPreferencesImpl has only static calls to AccountUtils (which then uses AccountManager):
https://github.com/nextcloud/android/blob/87dd10323aafcd9bbed5a97b7a222b401393287a/src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java#L245-L247

I would say we can address this once we get rid of these static calls, or?

ezaquarii:
I guess the delay will be measured in few milliseconds and if you use isMigrated Boolean flag in settings, it will simply not be called more than once.

Few frames dropped, not a big deal IMO. I doubt that anybody has 100 accounts to notice it.

Once we migrate to migration service we'll be able to synchronize it correctly.

Otherwise you are at the mercy of scheduler that might not schedule the migration thread before something reads old data.

Those issues are very tricky to reproduce and debug.

tobiasKaminsky:
I'll merge this to have master up and running again.
We can continue discussion here and then create following PRs.

PR #3912, opened 2019-04-17
Setup.md: direct usage of library project

tobiasKaminsky:


AndyScherzinger:
I am unsure if we should call this a best practice since that hints to this being the best way to use the lib while that is only true when working on both at the same time and will only work locally...

tobiasKaminsky:
Right, any other idea how we can call it?

AndyScherzinger:
We could call it How-To for example

PR #3914, opened 2019-04-17
backport show reshares for NC16+

tobiasKaminsky:
On <NC16 it does not show reshares, but only those user one shared.


tobiasKaminsky:
> Needs a Lib merge, else fine with me

Hm. This should be shown by our bot…

PR #3915, opened 2019-04-17
use path separator constant

jmue:
Use already defined path separator constant.


tobiasKaminsky:
Nice!
But there are even more, if you search for "/" in total project.
So…if you want… :-)

jmue:
> Nice!
> But there are even more, if you search for "/" in total project.
> So…if you want… :-)
I'm unsure how to deal with slashes within uri context. From my point of view they are not logical part of OCFile.PATH_SEPARATOR constant.

tobiasKaminsky:
> But I'm unsure how to deal with slashes within uri context. From my point of view they are not logical part of OCFile.PATH_SEPARATOR constant.

I would keep them, as you are right that they are not related to OCFile/files at all.

tobiasKaminsky:
Thank you very much!

PR #3916, opened 2019-04-17
unify empty string compare

jmue:
Unify string test on emptyness, use string `Textutils.isEmpty()`. This shortens written code and improves readability.


tobiasKaminsky:
This change led our unit tests fail: https://www.kaminsky.me/nc-dev/android-integrationTests/8869/

As unit test run without Android SDK there is no TextUtils.
So ErrorMessageAdapterUnitTest and TestSorting needs to be moved within /src/androidTest

jmue:
> This change led our unit tests fail: https://www.kaminsky.me/nc-dev/android-integrationTests/8869/
> 
> As unit test run without Android SDK there is no TextUtils.
> So ErrorMessageAdapterUnitTest and TestSorting needs to be moved within /src/androidTest

Thank you for helping out with that.

AndyScherzinger:
One question: Why the change / rearrangement in the gradle build file?

Anything else looks good to me :+1:

tobiasKaminsky:
> One question: Why the change / rearrangement in the gradle build file?

With this PR Mockito was added, but on a different place and thus dependancies for instrumentationTests and unit tests were mixed.

So now everything should be in its own area and if we add new dependencies we should obey these categories :-)

AndyScherzinger:
well, merging then :)

❤️ @jmue thanks a lot for this PR 🎉

PR #3920, opened 2019-04-17
remove unused variable/getter/setter in OCFile

jmue:
Remove availableOffline variable/getter/setter in OCFile as it is not used nowhere.


AndyScherzinger:
Thanks again for keeping the house clean ❤️

PR #3969, opened 2019-05-07
Handle silent delete/delete-all push notifications

tobiasKaminsky:
TODO
- [x]  test silent push notification on older version: how is this handled?

If a notification is now dismissed, the server sends a silent push notification to all devices and removes the notification in status bar.

This will be in NC17 on server side, but can already merged as this then just does nothing…
Ref: https://github.com/nextcloud/server/pull/15040, https://github.com/nextcloud/notifications/pull/318


src/main/java/com/owncloud/android/jobs/NotificationJob.java
```diff
@@ -133,8 +134,13 @@ protected Result onRunJob(@NonNull Params params) {
...
 
-                        // We ignore Spreed messages for now
-                        if (!APP_SPREED.equals(decryptedPushMessage.getApp())) {
+
+                        if (decryptedPushMessage.delete) {
+                            notificationManager.cancel(decryptedPushMessage.nid);
+                        } else if (decryptedPushMessage.deleteAll) {
+                            notificationManager.cancelAll();
+                        } else if (!APP_SPREED.equals(decryptedPushMessage.getApp())) {
+                            // We ignore Spreed messages for now
```
nickvergessen: Please don't!

The server app is cleverly enough. You only get them when there is no Talk app installed for the user. In this case a "linking to the web" notification in the files app is better than nothing.

tobiasKaminsky: Hm. This is code done by Mario.
I did not touched it.
I'll test what happens when I remove it.

tobiasKaminsky:
This is what I do get if I do not have Talk app installed.

nickvergessen:
Jep, thats the expected result

AndyScherzinger:
Rebase might help with spot bugs?

AndyScherzinger:
Nope,increased scoring...

tobiasKaminsky:
```

Method com.owncloud.android.datamodel.DecryptedPushMessage$$Parcelable.write(DecryptedPushMessage, Parcel, int, IdentityCollection) excessively uses methods of another class
--
  | Bug type CE_CLASS_ENVY (click for details) In class com.owncloud.android.datamodel.DecryptedPushMessage$$ParcelableIn method com.owncloud.android.datamodel.DecryptedPushMessage$$Parcelable.write(DecryptedPushMessage, Parcel, int, IdentityCollection)At DecryptedPushMessage$$Parcelable.java:[lines 48-61]Value com.owncloud.android.datamodel.DecryptedPushMessage



```

This is the reason. But this is because Parceler, so we cannot do much here.
Except of fiddling around with spotbugs to hide these false alarms.

AndyScherzinger:
> This is the reason. But this is because Parceler, so we cannot do much here.
Except of fiddling around with spotbugs to hide these false alarms.

@tobiasKaminsky filter didn't seem to work. So raise the score? (I guess)

AndyScherzinger:
@tobiasKaminsky spotbugs filter change doesn't seem to work...

tobiasKaminsky:
> @tobiasKaminsky spotbugs filter change doesn't seem to work...

Yes, I am on it, but somehow all regex I am trying just do not work… :/

AndyScherzinger:
not a regex pro but shouldn't  `~.*?Parcelable.*` work?

tobiasKaminsky:
Finally I hope I got it.
But I do not get why "~" is needed: <Class name="~.*\$\$Parcelable.*" />

Escaping $ is ok
.* is also needed, as this just means any or zero character
but ~ is included in "."…

PR #3971, opened 2019-05-07
Use product name on rich document, if available

tobiasKaminsky:

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -794,7 +794,7 @@
     <string name="stream_not_possible_headline">Internal streaming not possible</string>
     <string name="stream_not_possible_message">Please download media instead or use external app.</string>
     <string name="folder_already_exists">Folder already exists</string>
-    <string name="actionbar_open_as_richdocument">Open with Collabora</string>
+    <string name="actionbar_open_as_richdocument_parameter">Open with %1$s</string>
     <string name="notification_icon">Notification icon</string>
     <string name="folder_confirm_create">Create</string>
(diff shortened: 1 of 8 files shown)


src/main/res/values/strings.xml
```diff
@@ -793,7 +793,7 @@
     <string name="stream_not_possible_headline">Internal streaming not possible</string>
     <string name="stream_not_possible_message">Please download media instead or use external app.</string>
     <string name="folder_already_exists">Folder already exists</string>
-    <string name="actionbar_open_as_richdocument">Open with Collabora</string>
+    <string name="actionbar_open_as_richdocument">Open with</string>
```
AndyScherzinger: Or work with a pattern to have a parameter to not manually concatenate strings since we don't know how that works for all languages? :)

tobiasKaminsky: :+1: changed

tobiasKaminsky:
I'll rebase this after #3989 is merged, as both touch database.

AndyScherzinger:
@tobiasKaminsky #3989 has been merged

AndyScherzinger:
Lint score increased

tobiasKaminsky:
> Lint score increased

This was due to the fact that I changed an existing translatable string and thus it is not yet updated on Transifex.
So I created just a new one, which results in a bit c&p for translators, but stays same for 3.6 and we do not get any strange conflicts, like Android trying to add a a parameter to a string without one.

AndyScherzinger:
@tobiasKaminsky can Rebase this one again?

PR #3989, opened 2019-05-09
Public share: if ask for password is set, show an optional password dialog

tobiasKaminsky:
Ref: 
- [x] https://github.com/nextcloud/server/pull/15461

This is a new capability, starting with NC17.
As the current behaviour is to not show any dialog, it is ok to add it right now.
If the flag is set (see below), then it will be used.

image


tobiasKaminsky:
Will wait for final decision on capability name.

AndyScherzinger:
@tobiasKaminsky needs another rebase to resolve latest conflicts

PR #4008, opened 2019-05-12
Refactor ConnectivityUtils

ezaquarii:
ConnectivityUtils static methods moved to a service class.
This should open path to migrate from Android's deprecated
network state getter API to new asynchronous API.

Additionally this change migrates more AccountUtils
calls to newer UserAccountManager API.


build.gradle
```diff
@@ -250,8 +250,8 @@ android {
 dependencies {
     // dependencies for app building
     implementation 'androidx.multidex:multidex:2.0.1'
-//    implementation project('nextcloud-android-library')
-    genericImplementation 'com.github.nextcloud:android-library:master-SNAPSHOT'
+    //implementation project('nextcloud-android-library')
+    genericImplementation 'com.github.nextcloud:android-library:1.4.2'
```
ezaquarii: This needs to be removed in final commit version

scripts/analysis/findbugs-results.txt
```diff
@@ -1 +1 @@
-438
\ No newline at end of file
+17
```
ezaquarii: I don't believe this is true:

```
@@ -1 +1 @@
438
17
```

src/main/java/com/nextcloud/client/network/ConnectivityService.java
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.nextcloud.client.network;
+
+import com.evernote.android.job.JobRequest;
+
+public interface ConnectivityService {
+    boolean isInternetWalled();
```
ezaquarii: I don't want to jump into conclusions too soon, but I have strong gut feeling that this poetic terminology could be simplified:

```
boolean isServerReachable();
boolean isNetworkUnmetered();
```

tobiasKaminsky: I would go with isNetworkUnmetered, as reachable could also imply no network at all.

src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
```diff
@@ -136,4 +153,29 @@ public static boolean isOnlineWithWifi(Context context) {
             return false;
         }
     }
+
+    @Override
+    public JobRequest.NetworkType getActiveNetworkType() {
+        NetworkInfo networkInfo;
+        try {
+            networkInfo = connectivityManager.getActiveNetworkInfo();
+        } catch (Throwable t) {
```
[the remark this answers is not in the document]
ezaquarii: This is not mine - it was here in old code that I **moved** from `ConnectivityUtils.java`.

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -947,14 +951,14 @@ public void moveFiles(Collection<OCFile> files, OCFile targetFolder) {
      */
     public void copyFiles(Collection<OCFile> files, OCFile targetFolder) {
         for (OCFile file : files) {
-            Intent service = new Intent(mFileActivity, OperationsService.class);
+            Intent service = new Intent(fileActivity, OperationsService.class);
```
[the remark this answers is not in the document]
ezaquarii: Not mine too.

src/main/java/com/nextcloud/client/network/ConnectivityService.java
ezaquarii: I'm not so sure about this. It looks like `isInternetWalled` performs server-reachability test, while general any-kind-of-network availability has been somehow banished to `Device.java`.

ezaquarii: Maybe I'll just leave it like this and clean it during next round of refactoring.

ezaquarii:
I haven't run it on real device yet. Pls review, but do not merge yet.

AndyScherzinger:
@tobiasKaminsky can you cover the review? I am pretty swamped at work right now :(

src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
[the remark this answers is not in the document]
tobiasKaminsky: What about this?

src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
```diff
@@ -70,6 +71,16 @@ public Account getCurrentAccount() {
         return AccountUtils.getCurrentOwnCloudAccount(context);
     }
 
+    @Override
+    public OwnCloudAccount getCurrentNextcloudAccount() {
```
tobiasKaminsky: I know it is strange, but can you name it to getCurrentOwncloudAccount() as this is what we return?

tobiasKaminsky: And add @Nullable

src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
```diff
@@ -82,9 +93,13 @@ public Account getAccountByName(String name) {
         return null;
     }
 
+    @Override
+    public OwnCloudVersion getServerVersion(Account account) {
```
tobiasKaminsky: This is doing a call to a static method?
For what do we need it then?

src/main/java/com/nextcloud/client/network/ClientFactoryImpl.java
```diff
@@ -0,0 +1,63 @@
...
+        this.context = context;
+    }
+
+    @Override
+    public OwnCloudClient create(Account account) throws OperationCanceledException, AuthenticatorException, IOException, AccountUtils.AccountNotFoundException {
+        return OwnCloudClientFactory.createOwnCloudClient(account, context);
+    }
+
+    @Override
+    public OwnCloudClient create(Account account, Activity currentActivity) throws OperationCanceledException, AuthenticatorException, IOException, AccountUtils.AccountNotFoundException {
```
tobiasKaminsky: line break

src/main/java/com/nextcloud/client/network/ClientFactoryImpl.java
```diff
@@ -0,0 +1,63 @@
...
+        return OwnCloudClientFactory.createOwnCloudClient(account, context, currentActivity);
+    }
+
+    @Override
+    public OwnCloudClient create(Uri uri, boolean followRedirects, boolean useNextcloudUserAgent) {
+        return OwnCloudClientFactory.createOwnCloudClient(uri, context, followRedirects);
```
tobiasKaminsky: Same here, I do not understand the idea of this. 
Why do you wrap this is an additional class?

src/test/java/com/nextcloud/client/network/ConnectivityServiceTest.kt
```diff
@@ -0,0 +1,238 @@
...
+        companion object {
+            fun mockNetworkInfo(connected: Boolean, connecting: Boolean, type: Int): NetworkInfo {
+                val networkInfo = mock<NetworkInfo>()
+                whenever(networkInfo.isConnectedOrConnecting).thenReturn(connected or connecting)
+                whenever(networkInfo.isConnected).thenReturn(connected)
+                whenever(networkInfo.type).thenReturn(type)
+                return networkInfo
+            }
+
+            const val SERVER_BASE_URL = "https://test.server.com"
```
tobiasKaminsky: Please use here the value from AbstracIT or fetch it again:
https://github.com/nextcloud/android/blob/87dd10323aafcd9bbed5a97b7a222b401393287a/src/androidTest/java/com/owncloud/android/AbstractIT.java#L49-L51

src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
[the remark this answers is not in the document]
AndyScherzinger: I'd be fine with setting this one to ignore (not the rule in general though) since we sometimes need this for hardening :)

src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
ezaquarii: 1. It belongs here, as the information about server's version is tied to the account
2. Mocking in tests
3. it calls static function because I wanted to restructure it without changing the behaviour - this will undergo more refactorings as we modernize account handling.

src/test/java/com/nextcloud/client/network/ConnectivityServiceTest.kt
ezaquarii: This is not an integration test - the URL is not really used for anything. I had to provide some value - it could be an empty string as well.

src/main/java/com/nextcloud/client/network/ClientFactoryImpl.java
ezaquarii: To eliminate the dependency on context and whole zoo of platform dependencies it uses underneath. It simplifies mocking in tests greatly and simplifies every component that uses it as you don't need to pass context through the whole stack.

Quite frankly this API is a bit hostile with the amount of exceptions it throws and I see more cleanup around it in the future. The factory however is in a library, so I can't do it easily, nor is it a scope of this task. Wrapping it seems like a good temporary solution.

src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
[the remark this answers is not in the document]
ezaquarii: Again, this is just **moved** verbatim from `Device.java`, as it belongs here.

src/main/java/com/nextcloud/client/network/ClientFactoryImpl.java
ezaquarii: Line breaks added.

src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
ezaquarii: @nullable added and renamed

ezaquarii:
@tobiasKaminsky @AndyScherzinger I checked the application by using it for some time and it looks like it works without regressions. Are there any special edge cases you'd like to verify?

ezaquarii:
Both Codacy issues are copy&pasted from legacy code and that confused static checker. There are no **new** issues here.

ezaquarii:
@tobiasKaminsky @AndyScherzinger I rebased on top of QA changes made by Tobias few days ago (no other changes) and now it fails GPlay version CI.

I eyeballed the output but honestly I don't see a single failure there. I restarted if few times and the error for GPlay variant is consistent. Can you pls take a look?

AndyScherzinger:
I also checked the builds and couldn't see any issue except the build turning red. Same thing happens for master builds and any other up to date branch/pr.

Hoping for @tobiasKaminsky's knowledge of the build pipeline because it would be nice to have this on master before 3.7.0 RC1

src/main/java/com/nextcloud/client/network/ClientFactoryImpl.java
tobiasKaminsky: Thanks for the explanation :-)

src/main/java/com/owncloud/android/operations/UpdateShareViaLinkOperation.java
```diff
@@ -70,8 +70,8 @@ protected RemoteOperationResult run(OwnCloudClient client) {
         UpdateRemoteShareOperation updateOp = new UpdateRemoteShareOperation(publicShare.getRemoteId());
         updateOp.setPassword(password);
         updateOp.setExpirationDate(expirationDateInMillis);
-        updateOp.setPublicUploadOnFolder(publicUploadOnFolder);
-        updateOp.setPublicUploadOnFile(publicUploadOnFile);
+//        updateOp.setPublicUploadOnFolder(publicUploadOnFolder);
```
tobiasKaminsky: Why is this commented? I fear that this is a breaking change that we do not want to have?

tobiasKaminsky: Reverted

PR #4009, opened 2019-05-12
Spelling: Strings reworked

comradekingu:
As per, 
https://en.wikipedia.org/wiki/Webmail
https://en.wikipedia.org/wiki/Input/output
https://en.wikipedia.org/wiki/End-of-life_(product)

Curiously https://www.britannica.com/technology/calendering means something else

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -85,7 +85,7 @@
     <string name="uploader_upload_files_behaviour">Upload option:</string>
     <string name="uploader_upload_files_behaviour_move_to_nextcloud_folder">Move file to %1$s folder</string>
     <string name="uploader_upload_files_behaviour_only_upload">Keep file in source folder</string>
-    <string name="uploader_upload_files_behaviour_upload_and_delete_from_source">Delete file from source folder.</string>
+    <string name="uploader_upload_files_behaviour_upload_and_delete_from_source">Delete file from source folder</string>
     <string name="file_list_seconds_ago">seconds ago</string>
     <string name="file_list_empty_headline">No files here</string>
     <string name="folder_list_empty_headline">No folders here</string>
@@ -96,10 +96,10 @@
     <string name="file_list_no_app_for_file_type">No app set up to handle this file type.</string>
     <string name="local_file_list_empty">There are no files in this folder.</string>
     <string name="local_folder_list_empty">There are no further folders.</string>
-    <string name="file_list_empty_headline_search">No results in this folder.</string>
+    <string name="file_list_empty_headline_search">No results in this folder</string>
     <string name="file_list_empty_headline_server_search">No results</string>
-    <string name="file_list_empty_favorite_headline">Nothing favorited yet.</string>
-    <string name="file_list_empty_shared_headline">Nothing shared yet.</string>
+    <string name="file_list_empty_favorite_headline">Nothing favorited yet</string>
+    <string name="file_list_empty_shared_headline">Nothing shared yet</string>
     <string name="file_list_empty_shared">Files and folders you share will show up here.</string>
     <string name="file_list_empty_headline_server_search_videos">No videos</string>
     <string name="file_list_empty_headline_server_search_photos">No photos</string>
@@ -112,7 +112,7 @@
     <string name="file_list_empty_text_photos_filter">No photos.</string>
     <string name="file_list_empty_text_videos">Upload some videos or activate auto upload.</string>
     <string name="file_list_empty_text_videos_filter">No videos.</string>
-    <string name="trashbin_empty_headline">No deleted files.</string>
+    <string name="trashbin_empty_headline">No deleted files</string>
     <string name="trashbin_empty_message">You will be able to recover deleted files from here.</string>
     <string name="upload_list_empty_headline">No uploads available</string>
     <string name="upload_list_empty_text_auto_upload">Upload some content or activate auto upload.</string>
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -641,7 +641,7 @@
 
     <string name="prefs_sourcecode">Get source code</string>
     <string name="prefs_license">License</string>
-    <string name="prefs_gpl_v2">GNU General Public License, version 2+</string>
+    <string name="prefs_gpl_v2">GNU General Public License, version 2</string>
     <string name="push_notifications_not_implemented">Push notifications turned off due to dependencies on proprietary Google Play services.</string>
     <string name="push_notifications_old_login">No push notifications due to outdated login session. Please consider re-adding your account.</string>
     <string name="push_notifications_temp_error">Push notifications currently unavailable.</string>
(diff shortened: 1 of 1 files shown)


src/main/res/values/strings.xml
```diff
@@ -751,47 +751,47 @@
...
     <string name="new_media_folder_videos">video</string>
-    <string name="outdated_server">The server has reached end of life, please upgrade!</string>
+    <string name="outdated_server">The server has reached end-of-life, please upgrade!</string>
     <string name="dismiss">Dismiss</string>
-    <string name="feedback_no_mail_app">No app available to send mails!</string>
+    <string name="feedback_no_mail_app">No app available to send e-mails.</string>
     <string name="drawer_item_trashbin">Deleted files</string>
     <string name="trashbin_activity_title">Deleted files</string>
     <string name="action_empty_trashbin">Empty trashbin</string>
-    <string name="trashbin_loading_failed">Loading trashbin failed!</string>
```
comradekingu: As in starting of filling up?

src/main/res/values/strings.xml
```diff
@@ -641,10 +641,10 @@
 
     <string name="prefs_sourcecode">Get source code</string>
     <string name="prefs_license">License</string>
-    <string name="prefs_gpl_v2">GNU General Public License, version 2</string>
-    <string name="push_notifications_not_implemented">Push notifications disabled due to dependencies on proprietary Google Play services.</string>
+    <string name="prefs_gpl_v2">GNU General Public License, version 2+</string>
```
comradekingu:
```suggestion
    <string name="prefs_gpl_v2">GNU General Public License, version 2</string>
```
Appears to attemptedly have been changed to GPLv2 only, and from v3+ to v2 only, and from v3+ to v3 only, after being adopted from ownCloud and respective copyright holders. Some of it is (still) GPLv3+ now… Transifex license says GPLv2+, but the version used is closed source now.

tobiasKaminsky: "old" owncloud code is gpl2, but newer Nextcloud code is v3+.

src/main/res/values/strings.xml
```diff
@@ -79,47 +79,47 @@
...
     <string name="uploader_error_title_file_cannot_be_uploaded">This file can not be uploaded</string>
-    <string name="uploader_error_message_read_permission_not_granted">%1$s is not allowed to read a received file</string>
+    <string name="uploader_error_message_read_permission_not_granted">%1$s is not allowed to read a received file.</string>
     <string name="uploader_error_message_source_file_not_found">File selected for upload not found. Please check whether the file exists.</string>
     <string name="uploader_error_message_source_file_not_copied">Could not copy file to a temporary folder. Try to resend it.</string>
     <string name="uploader_upload_files_behaviour">Upload option:</string>
     <string name="uploader_upload_files_behaviour_move_to_nextcloud_folder">Move file to %1$s folder</string>
     <string name="uploader_upload_files_behaviour_only_upload">Keep file in source folder</string>
-    <string name="uploader_upload_files_behaviour_upload_and_delete_from_source">Delete file from source folder</string>
+    <string name="uploader_upload_files_behaviour_upload_and_delete_from_source">Delete file from source folder.</string>
```
tobiasKaminsky: Having a "." here loosk strange.

src/main/res/values/strings.xml
```diff
@@ -79,47 +79,47 @@
...
     <string name="file_list_empty">Upload some content or sync with your devices.</string>
-    <string name="file_list_empty_favorites_filter_list">Files and folders you mark as favorites will show up here</string>
+    <string name="file_list_empty_favorites_filter_list">Files and folders you mark as favorites will show up here.</string>
     <string name="file_list_empty_favorites_filter">Your search returned no favorited files.</string>
     <string name="file_list_loading">Loading…</string>
     <string name="file_list_no_app_for_file_type">No app set up to handle this file type.</string>
     <string name="local_file_list_empty">There are no files in this folder.</string>
     <string name="local_folder_list_empty">There are no further folders.</string>
-    <string name="file_list_empty_headline_search">No results in this folder</string>
+    <string name="file_list_empty_headline_search">No results in this folder.</string>
```
tobiasKaminsky: Same here, this is somekind of a headline…

src/main/res/values/strings.xml
comradekingu: Moving that discussion to #4014.

src/main/res/values/strings.xml
comradekingu: I think removing dots from everything with `_headline_` will clear that up.

src/main/res/values/strings.xml
comradekingu: Done. Could be "No results in folder", but that is nitpicking at best.

src/main/res/values/strings.xml
tobiasKaminsky: Filling up, e.g. for some reason content of trash bin could not be retrieved from server.

tobiasKaminsky:
@jancborchardt can you review this as this is  UI/UX related

src/main/res/values/strings.xml
comradekingu: "Could not connect to trashbin?"

jancborchardt:
I cursorily looked over the changes, but in most of these cases the full stop does not make much sense because the sentence is very short and not a sentence in body text.

Similarly, we had this in the past: We put a space before ellipsis when we use it.

-------

Generally I'm not very fond of sweeping changes in text strings when we don't have the context – as can be seen from @tobiasKaminsky's feedback already. It should be done when working on the specific part, with knowledge of how it will look like.

comradekingu:
The dichotomy is lost on me. A sentence can be short, albeit still a sentence, adding another one would warrant an ending dot. At some point length could do it to (?), but what is the specified length at which this occurs?

My general issue is with translations, were length not only varies greatly, but are also subject to rules requiring sentences to always have dots. For sake of consistency it is favourable to apply dots, rather than to omit them. This has been broken for quite some time, resulting in sub-par translations. When Nexcloud eventually moves to Weblate, there are checks in place for consistency. It keep the workload down and helps consistency overall to do it this way.

<details>
<summary>
The ellipsis situation
</summary
The use of ellipsis in the Android client is actually unchanged in this PR, and for the whole of the Android client, also curiously consistent. To my knowledge that can not be said of any part of Nextcloud that purports to use space before ellipsis. If it isn't going to change for the whole of Nextcloud, and the translation remains on Transifex, doing it here will invalidate a lot of translations for no ultimate benefit. My arguments for changing it, and merit of actually being willing to do the work ended up being insufficient, so that is a challenge for someone else to do, for another day.</details>

jancborchardt:
As mentioned before, and in other pull requests by you, we generally make individual changes to text when working on the context anyway. Above for example where you even ask where a specific string occurs kind of proves the point.

> When Nexcloud eventually moves to Weblate

Not aware that this is a plan.

-----------

Sorry, but reviewing these, discussing them individually to reach a consensus, and making sure that there are no regressions is more work than it is worth. So I’m going to have to reject this.

PR #4028, opened 2019-05-15
use authenticated glide loader for svg

tobiasKaminsky:
- have external app installed
- previously: links in drawer had default icon
- now: they do get the real icon

This was due to a change that requires that icons can only be fetched if authenticated

### DIFF
--- a/src/main/java/com/owncloud/android/utils/glide/CustomGlideUriLoader.java
+++ b/src/main/java/com/owncloud/android/utils/glide/CustomGlideUriLoader.java
@@ -1,0 +1,46 @@
+/*
+ * Nextcloud Android client application
+ *
+ * @author Tobias Kaminsky
+ * Copyright (C) 2019 Tobias Kaminsky
+ * Copyright (C) 2019 Nextcloud GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


src/main/java/com/owncloud/android/utils/glide/CustomGlideUriLoader.java
```diff
@@ -0,0 +1,45 @@
+/**
+ * Nextcloud Android client application
+ *
+ * @author Alejandro Bautista
+ * @author Chris Narkiewicz
+ * <p>
+ * Copyright (C) 2017 Alejandro Bautista Copyright (C) 2019 Chris Narkiewicz <hello@ezaquarii.com>
```
AndyScherzinger: seems to be missing a proper line break (one auther per line), also comes with paragraph html tags

AndyScherzinger:
👍 looks good to me except for one minor comment format issue

src/main/java/com/owncloud/android/utils/glide/CustomGlideUriLoader.java
tobiasKaminsky: I'll change it to "my" header.

PR #4031, opened 2019-05-16
Use account.hashCode instead of name as account switching value

tobiasKaminsky:
To test you need an account where display name is not the same like user Id, e.g:
tobi in settings->users
"Tobias K" in display name -> setting of my user

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -138,21 +138,21 @@ steps:
   recipients_only: true
   recipients: [ $EMAIL_RECIPIENTS ]
   environment:
-    EMAIL_USERNAME:
+      EMAIL_USERNAME:
       from_secret: EMAIL_USERNAME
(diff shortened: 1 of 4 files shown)


AndyScherzinger:
Score issues :/

AndyScherzinger:
Backport?

sten-schmidt:
In 9236.apk you can only add one account , nothing happens when AddAccount-Command is selectet in the Menü for adding the second account.

tobiasKaminsky:
@sten-schmidt thank you for testing :tada: 

I indeed forgot to change add/manage behaviour.

Once there is a new version, can you please give it a try again?

sten-schmidt:
@tobiasKaminsky with 9288.apk the AccountSwitch works well. :)

PR #4040, opened 2019-05-18
Remove legacy FindBugs tasks from build script

ezaquarii:


build.gradle
```diff
@@ -246,7 +224,7 @@ android {
         }
     }
 
-    check.dependsOn 'checkstyle', 'findbugs', 'pmd', 'lint'
+    check.dependsOn 'checkstyle', 'pmd', 'lint'
```
ezaquarii: Not sure about this

AndyScherzinger: @tobiasKaminsky likely knows

tobiasKaminsky: This just means that ./gradlew check will start checkstyle, pmd, etc.
So having findbug/spotbug in it would be nice.

ezaquarii: Ok, I added spotbugs here.

build.gradle
```diff
@@ -200,27 +199,6 @@ android {
         }
     }
 
-    tasks.register("findbugs", FindBugs) {
```
AndyScherzinger: wouldn't we need this task but rather for spotbugs now? (since it has been completely removed here)

build.gradle
```diff
@@ -74,7 +73,7 @@ def versionPatch = 0
 def versionBuild = 0 // 0-50=Alpha / 51-98=RC / 90-99=stable
 
 def taskRequest = getGradle().getStartParameter().getTaskRequests().toString()
-if (taskRequest.contains("Gplay") || taskRequest.contains("findbugs") || taskRequest.contains("lint")) {
```
AndyScherzinger: so this doesn't have to be replaced with the spotbugs check instead of the check being removed?

ezaquarii: We're registering `spotbugs` task for all variants now, so google variant will apply `gplay.gradle`.

Since `findbugs` task is not used for some time now, this condition is false anyway and so far all works ok.

build.gradle
ezaquarii: It is registered. You can find it around line 205 below (new version, right column, unfold the collapsed code block).

AndyScherzinger:
@ezaquarii a rebase should help with the ci build now (latest merge to master fixed the pipeline, fix provided by @tobiasKaminsky)

PR #4056, opened 2019-05-22
Re-enable exception handler

### LINKED ISSUE #4030: Re-add Exception handler
Like in Deck app: if we encounter an Exception, we can show a nice dialog.
Pre-requisites are already at https://github.com/nextcloud/android/blob/9130635fa1b9ba7b18a08cb54177c2924c2a8d8c/src/main/java/com/owncloud/android/ui/errorhandling/ExceptionHandler.java#L30

tobiasKaminsky:
Resolves #4030


src/main/java/com/owncloud/android/ui/activities/ActivitiesActivity.java
```diff
@@ -109,6 +110,7 @@
     protected void onCreate(Bundle savedInstanceState) {
         Log_OC.v(TAG, "onCreate() start");
         super.onCreate(savedInstanceState);
+        Thread.setDefaultUncaughtExceptionHandler(new ExceptionHandler(this));
```
ezaquarii: I'd see this in `MainApp.onCreate()` so it can start working early, before activities are launched. Will it work there?

ezaquarii: How does it work if app is terminated by JVM, such as in case of OOM? Will it preserve the stack over process resurrection?

tobiasKaminsky: > I'd see this in `MainApp.onCreate()` so it can start working early, before activities are launched. Will it work there?

MainApp is not working, but I set it in BaseActivity.




> How does it work if app is terminated by JVM, such as in case of OOM? Will it preserve the stack over process resurrection?

I do not know. I'll give it a try.

tobiasKaminsky: I added it to Settings:
```
int i = Integer.MAX_VALUE;
        String[] array = new String[i];
```

resulting in

ezaquarii: Ok, one more thing to resolve is a crash when activity is not started. Let's imagine that Service or document provider is launched by OS. In this case, because the handler is registered in activity, it won't work.

Why it dodn't work in `MainApp`? I just checked what `ACRA` is doing and it seems they are registering it like we do, in `attachBaseContext`.

tobiasKaminsky: > Ok, one more thing to resolve is a crash when activity is not started. Let's imagine that Service or document provider is launched by OS. In this case, because the handler is registered in activity, it won't work.

True, I am unsure if this also can be registered in document provider…



> Why it didn't work in `MainApp`? I just checked what `ACRA` is doing and it seems they are registering it like we do, in `attachBaseContext`.

It just turns into a black screen and after a while "Nextcloud is not responding".
From log it seems as ExceptionHandler is started, but just the activity is not shown.

tobiasKaminsky: So I added a RuntimeException in DocumentStorageProvider, along with the ExceptionHandler, but:
```
2019-05-24 08:34:29.487 10380-10445/com.nextcloud.client E/JavaBinder: *** Uncaught remote exception!  (Exceptions are not yet supported across processes.)
    java.lang.RuntimeException: fail
        at com.owncloud.android.providers.DocumentsStorageProvider.queryDocument(DocumentsStorageProvider.java:154)
        at android.provider.DocumentsProvider.query(DocumentsProvider.java:797)
        at android.content.ContentProvider$Transport.query(ContentProvider.java:240)
        at android.content.ContentProviderNative.onTransact(ContentProviderNative.java:102)
        at android.os.Binder.execTransact(Binder.java:731)
```

tobiasKaminsky:
@ezaquarii do we wanna give this a try? 
I would then soon enhance layout a bit and then we can merge it.

AndyScherzinger:
> I would then soon enhance layout a bit and then we can merge it.

I could also help with that (maybe _after_ a merge via a separate PR), if that would help you :)

tobiasKaminsky:
> So I added a RuntimeException in DocumentStorageProvider, along with the ExceptionHandler, but:
```
2019-05-24 08:34:29.487 10380-10445/com.nextcloud.client E/JavaBinder: *** Uncaught remote exception!  (Exceptions are not yet supported across processes.)
    java.lang.RuntimeException: fail
        at com.owncloud.android.providers.DocumentsStorageProvider.queryDocument(DocumentsStorageProvider.java:154)
        at android.provider.DocumentsProvider.query(DocumentsProvider.java:797)
        at android.content.ContentProvider$Transport.query(ContentProvider.java:240)
        at android.content.ContentProviderNative.onTransact(ContentProviderNative.java:102)
        at android.os.Binder.execTransact(Binder.java:731)
```

This is the only remaining problem.

tobiasKaminsky:
@AndyScherzinger do you have time to add, before 3.8?
- a share button on toolbar
- a text this exception can be useful to create a new github issue with a link

Then we can merge this right now.

AndyScherzinger:
> @AndyScherzinger do you have time to add, before 3.8?

What's the date on the feature-freeze?

AndyScherzinger:
@tobiasKaminsky please check and merge at will, I did:
* [x] added a toolbar + title
* [x] implemented share action
* [x] extended the layout

as a bonus to make @ezaquarii happy
* [x] rewrote the activity in kotlin 😁

AndyScherzinger:
@tobiasKaminsky beware, I did a rebase to harmonize the share-string use (added `common_share`)

AndyScherzinger:
@tobiasKaminsky as for

> a text this exception can be useful to create a new github issue with a link

where how should that be added? at the top below the toolbar? copy the text to the clipboard?
We can probably do that in a separate PR

AndyScherzinger:
@tobiasKaminsky added sharing & report issue+copy-to-clipboard & proper theming, screenshots see https://github.com/nextcloud/android/pull/4056#issue-281071389

tobiasKaminsky:
This is great progress! :heart: 

I just realized that we do not have any info about our app, I'll add this.
```

java.lang.RuntimeException: Unable to start activity ComponentInfo{com.nextcloud.client/com.owncloud.android.ui.activity.FileDisplayActivity}: java.lang.RuntimeException
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2913)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3048)
	at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:78)
	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:108)
	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:68)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1808)
	at android.os.Looper.loop(Looper.java:193)
	at android.app.ActivityThread.main(ActivityThread.java:6669)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
Caused by: java.lang.RuntimeException
	at com.owncloud.android.ui.activity.FileDisplayActivity.onCreate(FileDisplayActivity.java:288)
	at android.app.Activity.performCreate(Activity.java:7136)
	at android.app.Activity.performCreate(Activity.java:7127)
	at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1271)
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2893)
	... 11 more

************ DEVICE INFORMATION ***********
Brand: google
Device: generic_x86
Model: Android SDK built for x86
Id: PSR1.180720.093
Product: sdk_gphone_x86

SDK: 28
Release: 9
Incremental: 5456446
```

tobiasKaminsky:
```
ID: com.nextcloud.client
Version: 30080000 (build #388)
Build flavor: gplay
```

Build number is set on branded versions.
ID is useful to see devVersion, qa or branded app.

AndyScherzinger:
@tobiasKaminsky ready to be merged 👍

ezaquarii:
Hold on. Why is AppInfoImpl public?

tobiasKaminsky:
> Hold on. Why is AppInfoImpl public?

I changed this to get version info in the exception info:
```
ID: com.nextcloud.client
Version: 30080000 (build #388)
Build flavor: gplay
```

src/main/java/com/nextcloud/client/appinfo/AppInfoImpl.java
```diff
@@ -21,7 +21,7 @@
 
 import com.owncloud.android.BuildConfig;
 
-class AppInfoImpl implements AppInfo {
+public class AppInfoImpl implements AppInfo {
```
ezaquarii: This is private implementation that should not be exposed. We can inject `AppInfo` via `ExceptionHandler` constructor the same way we inject `Context`.

src/main/java/com/nextcloud/client/errorhandling/ExceptionHandler.kt
```diff
@@ -0,0 +1,111 @@
...
+import android.os.Build
+import android.util.Log
+import com.nextcloud.client.appinfo.AppInfoImpl
+import com.owncloud.android.BuildConfig
+import com.owncloud.android.R
+import java.io.PrintWriter
+import java.io.StringWriter
+import kotlin.system.exitProcess
+
+class ExceptionHandler(private val context: Activity) : Thread.UncaughtExceptionHandler {
```
ezaquarii: We can narrow type down to `Context`. The handler's place in in `MainApp` anyway (which is `Context` subtype) and when we figure out why it didn't work we'll move it there.

src/main/java/com/nextcloud/client/errorhandling/ExceptionHandler.kt
```diff
@@ -0,0 +1,111 @@
...
+        val intent = Intent(context, ShowErrorActivity::class.java)
+        intent.putExtra(ShowErrorActivity.EXTRA_ERROR_TEXT, errorReport)
+        context.startActivity(intent)
+
+        android.os.Process.killProcess(android.os.Process.myPid())
+        exitProcess(STATUS)
+    }
+
+    private fun generateErrorReport(stackTrace: String): String {
+        val appInfo = AppInfoImpl()
```
ezaquarii: The problem with this is that if I want to add more stuff ot `AppInfoImpl` (let's say, another dependency), now I'll need to inject that dependency to `ExceptionHandler`... and possibly any other place that uses `AppInfoImpl`. I'd consider injecting this via contructor. `AppInfo` can be obtained from `BaseActivity` where the handler is constructed and it also is already available in `MainApp`, so this is safe move.

src/main/java/com/nextcloud/client/errorhandling/ShowErrorActivity.kt
```diff
@@ -0,0 +1,115 @@
...
+        setContentView(R.layout.activity_show_error)
+
+        text_view_error.text = intent.getStringExtra(EXTRA_ERROR_TEXT)
+
+        setSupportActionBar(toolbar)
+
+        val snackbar = DisplayUtils.createSnackbar(
+            error_page_container,
+            R.string.error_report_issue_text, Snackbar.LENGTH_INDEFINITE)
+            .setAction(R.string.error_report_issue_action
```
ezaquarii: How about `.setAction(R.string.error_report_issue_action) { reportIssue() }`?

src/main/java/com/nextcloud/client/errorhandling/ExceptionHandler.kt
```diff
@@ -0,0 +1,111 @@
...
+import com.owncloud.android.BuildConfig
+import com.owncloud.android.R
+import java.io.PrintWriter
+import java.io.StringWriter
+import kotlin.system.exitProcess
+
+class ExceptionHandler(private val context: Activity) : Thread.UncaughtExceptionHandler {
+    companion object {
+        private val TAG = ExceptionHandler::class.java.simpleName
+        private val LINE_SEPARATOR = "\n"
```
ezaquarii: `private const val`

src/main/java/com/nextcloud/client/errorhandling/ShowErrorActivity.kt
```diff
@@ -0,0 +1,115 @@
...
+            val intent = Intent(Intent.ACTION_VIEW, uriUrl)
+            DisplayUtils.startIntentIfAppAvailable(intent, this, R.string.no_browser_available)
+        }
+        Toast.makeText(this, R.string.copied_to_clipboard, Toast.LENGTH_LONG).show()
+    }
+
+    override fun onCreateOptionsMenu(menu: Menu?): Boolean {
+        menuInflater.inflate(R.menu.activity_error_show, menu)
+
+        ThemeUtils.tintDrawable(menu?.findItem(R.id.error_share)?.icon, ThemeUtils.fontColor(this))
```
ezaquarii: Does `tintDrawable` handle `null`? `?.icon` can potentially yeld `null`. Although it probably won't in this context, we need to be midful on nullability issues across Java/Kotlin. Java will happily swallow `null` and blow up.

src/main/java/com/nextcloud/client/errorhandling/ExceptionHandler.kt
```diff
@@ -0,0 +1,111 @@
...
+    private fun generateErrorReport(stackTrace: String): String {
+        val appInfo = AppInfoImpl()
+        val buildNumber = context.resources.getString(R.string.buildNumber)
+
+        var buildNumberString = ""
+        if (buildNumber.isNotEmpty()) {
+            buildNumberString = " (build #$buildNumber)"
+        }
+
+        return "************ CAUSE OF ERROR ************\n\n" +
```
ezaquarii: Using string builder or template string wouldn't hurt. I guess this is copied from old impl. Not that it matters in this context...

src/main/java/com/owncloud/android/ui/activities/ActivitiesActivity.java
ezaquarii: Remote process exception is caught and handled gracefully, so it doesn't crash entire process.

src/main/java/com/nextcloud/client/appinfo/AppInfoImpl.java
ezaquarii: I banished it back to darkness. :)

src/main/java/com/nextcloud/client/errorhandling/ExceptionHandler.kt
ezaquarii: Generally done.

src/main/java/com/nextcloud/client/errorhandling/ExceptionHandler.kt
ezaquarii: Also done.

ezaquarii:
@tobiasKaminsky @AndyScherzinger Can you pls re-review last commit?

src/main/java/com/nextcloud/client/errorhandling/ExceptionHandler.kt
```diff
@@ -21,43 +25,46 @@
...
-        Log.e(TAG, "An exception was thrown and handled by ExceptionHandler:", exception)
-
-        val intent = Intent(context, ShowErrorActivity::class.java)
-        intent.putExtra(ShowErrorActivity.EXTRA_ERROR_TEXT, errorReport)
-        context.startActivity(intent)
-
-        android.os.Process.killProcess(android.os.Process.myPid())
-        exitProcess(STATUS)
+        try {
+            val stackTrace = StringWriter()
```
ezaquarii: I removed calls to `Log.e` as they were printing exception extra 2 times. Normally the platform exception handler does it for us and this is happening now via `defaultExceptionHandler`, preserving original exception formatting.

src/main/java/com/nextcloud/client/errorhandling/ExceptionHandler.kt
```diff
@@ -73,7 +80,7 @@ class ExceptionHandler(private val context: Activity) : Thread.UncaughtException
             BuildConfig.APPLICATION_ID +
             LINE_SEPARATOR +
             "Version: " +
-            appInfo.formattedVersionCode +
+            (appInfoProvider.get()?.formattedVersionCode ?: "not available") +
```
ezaquarii: `appInfo` is not available immediately, so we make it lazy. This way we can catch early exception that happen during DI init.

Not that I expect them and if those happen, it's probably already hopeless to handle. We can experiment with dagger-less process startup in context of `:crash` only, as this process must support only 1 activity so we don't need to bring up the world.

src/main/java/com/nextcloud/client/errorhandling/ExceptionHandler.kt
AndyScherzinger: Yupp, simple copy of the original code but yeah should probably be fixed (either here or in a later PR)

src/main/java/com/nextcloud/client/errorhandling/ExceptionHandler.kt
```diff
@@ -0,0 +1,118 @@
...
+            exception.printStackTrace(PrintWriter(stackTrace))
+            val errorReport = generateErrorReport(stackTrace.toString())
+            val intent = Intent(context, ShowErrorActivity::class.java)
+            intent.putExtra(ShowErrorActivity.EXTRA_ERROR_TEXT, errorReport)
+            intent.flags = Intent.FLAG_ACTIVITY_NEW_TASK
+            context.startActivity(intent)
+            // Pass exception to OS for graceful handling - OS will report it via ADB
+            // and close all activities and services.
+            defaultExceptionHandler.uncaughtException(thread, exception)
+        } catch (fatalException: Exception) {
```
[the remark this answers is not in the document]
ezaquarii: Suppressed as this is what we want here.

src/main/java/com/nextcloud/client/errorhandling/ShowErrorActivity.kt
AndyScherzinger: Don't know in this context while yes in this case it won't be null.

AndyScherzinger:
@ezaquarii nice work, just one lint error which needs fixing:

```
../../../src/main/java/com/owncloud/android/MainApp.java:176: Call requires API level 28 (current min is 14): android.app.Application#getProcessName
```

ezaquarii:
> Call requires API level 28 (current min is 14): 
> android.app.Application#getProcessName
> ```

DOH! Thanks, working on a solution right now.

ezaquarii:
@tobiasKaminsky @AndyScherzinger The pop-up with stacktrace is nice, but I think it can be a bit surprising for a user. I believe that a more descriptive window title would be nice.

src/main/java/com/nextcloud/client/errorhandling/ShowErrorActivity.kt
```diff
@@ -0,0 +1,110 @@
...
+
+        val snackbar = DisplayUtils.createSnackbar(
+            error_page_container,
+            R.string.error_report_issue_text, Snackbar.LENGTH_INDEFINITE)
+            .setAction(R.string.error_report_issue_action) { reportIssue() }
+
+        val primaryColor = ThemeUtils.primaryColor(this)
+        val fontColor = ThemeUtils.fontColor(this)
+
+        ThemeUtils.colorSnackbar(this, snackbar)
```
ezaquarii: Is that accessing any app data, such as account info?

I understand the desire to be consistent with app theme, but depending on any app API under crash condition carries a risk of not getting stack trace at all.

AndyScherzinger: Yes, it reads the originally server side defined colors to theme the screen as seen on the screenshots.

@tobiasKaminsky what do you think? I get the vanilla Api argument and would be fine to drop theming in this case.

src/main/java/com/nextcloud/client/errorhandling/ShowErrorActivity.kt
tobiasKaminsky: We should do more annotation:
```
@Nullable 
    public static Drawable tintDrawable(Drawable drawable, int color) {
if (drawable != null) {
…
}
return null;
``` 

We should use @NotNull when we do not check it.

src/main/java/com/nextcloud/client/errorhandling/ExceptionHandler.kt
tobiasKaminsky: In fact it was a string builder, but AndroidStudio suggested me to change it.
https://github.com/nextcloud/android/commit/b5f38f000d9cc1a4edf7720b49e115abd86247cc#diff-50076ac20cdd0be4538afca0879f0cdfL45

src/main/java/com/nextcloud/client/errorhandling/ShowErrorActivity.kt
tobiasKaminsky: Yes, theming should be disabled.
(That was also the reason why I used AppInfoImpl directly as then there is no dep to dagger…)

AndyScherzinger: Removed themeing with latest commit 👍

AndyScherzinger:
> Something like "Nextcoud Crashed" or similar.

Done with latest commit (while Nextcloud has been made dynamic to support branding

AndyScherzinger:
@tobiasKaminsky @ezaquarii do we want to remove the dagger dependency here too (right now) or via a separate PR?

src/main/java/com/nextcloud/client/errorhandling/ShowErrorActivity.kt
AndyScherzinger: > We should do more annotation

While that should be done separately in a new PR then

AndyScherzinger:
> do we want to remove the dagger dependency here too (right now) or via a separate PR?

removed just waiting for the build

PR #4058, opened 2019-05-22
OCFileListFragment: Only scroll to top if directory changed

Unpublished:
Fixes jumping to top everytime the listview is updated, since 3.6.1.

TODO

- [x] This needs to be tested with talk app

### DIFF
--- a/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
+++ b/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
@@ -122,6 +122,7 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
 import androidx.appcompat.app.ActionBar;
+import androidx.core.util.ObjectsCompat;
 import androidx.drawerlayout.widget.DrawerLayout;
 import androidx.fragment.app.FragmentActivity;
 import androidx.recyclerview.widget.GridLayoutManager;
@@ -1190,6 +1191,8 @@ public class OCFileListFragment extends ExtendedListFragment implements
                     onlyOnDevice,
                     mLimitToMimeType
                 );
+
+                boolean changedDirectory = !ObjectsCompat.equals(mFile, directory);
                 mFile = directory;
 
                 updateLayout();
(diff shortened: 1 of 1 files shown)


Unpublished:
This needs to be tested with talk app.

src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1190,14 +1191,18 @@ public void run() {
                     onlyOnDevice,
                     mLimitToMimeType
                 );
+
+                boolean changedDirectory = !ObjectsCompat.equals(mFile, directory);
```
tobiasKaminsky: We should stick to equals from OCFile as this also compares if it is the same content, but maybe a different object:
https://github.com/nextcloud/android/blob/b69169f457c38abe50f8828d8c0e53a477b55b5e/src/main/java/com/owncloud/android/datamodel/OCFile.java#L435-L447

src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1190,14 +1191,18 @@ public void run() {
...
                 mFile = directory;
 
                 updateLayout();
 
                 mAdapter.setHighlightedItem(file);
                 int position = mAdapter.getItemPosition(file);
                 if (position == -1) {
-                    getRecyclerView().scrollToPosition(0);
+                    if (changedDirectory) {
```
tobiasKaminsky: As this is a really small comparison and the check please inline it here.
(in your implementation the comparison was done every time, regardless if it is needed)

tobiasKaminsky:
@Unpublished thanks for the finding of this, I did not thought about it :tada: 
After your changes, I'll test it with Talk.

src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
Unpublished: Done, though `Objects.equals` calls this equals after making sure both objects aren't the same (reference) and the first is not null.

src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
Unpublished: Thought about this, but couldn't completely inline it because `mFile` is updated in line 1198... I'm saving the previous value temporarily now.

tobiasKaminsky:
Tested and working :+1:

tobiasKaminsky:
Thank you very much @Unpublished :tada: :tada: :tada:

Unpublished:
Oh it works 😄 
You're very welcome and I'm happy to contribute :)

PR #4063, opened 2019-05-23
Add ktlint

tobiasKaminsky:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -65,7 +65,6 @@ steps:
       - ./gradlew combinedTestReport
       - curl -o codecov.sh https://codecov.io/bash
       - bash ./codecov.sh -t fc506ba4-33c3-43e4-a760-aada38c24fd5
-        - scripts/uploadArtifact.sh $LOG_USERNAME $LOG_PASSWORD $DRONE_BUILD_NUMBER $DRONE_PULL_REQUEST $GIT_USERNAME $GIT_TOKEN
   - name: notify
     image: drillster/drone-email
     settings:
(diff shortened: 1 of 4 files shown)


tobiasKaminsky:
There is ./gradlew ktlintFormat, which is also available within AS.
But I do not think that we should do this automatically, or?

ezaquarii:
> But I do not think that we should do this automatically, or?

It should reject commits if they are not formatted. This saves on useless review round-trips and suppresses inflammatory discussions.

Besides, Kotlin has some features that make Java style unfit for purpose and it will take time until the knowledge spreads.

tobiasKaminsky:
> It should reject commits if they are not formatted. This saves on useless review round-trips and suppresses inflammatory discussions.

I really like this, but this seems to be a github enterprise feature only.
While we can encourage people to run this check locally, we cannot ensure this.
Do you have a better idea?

ezaquarii:
> I really like this, but this seems to be a github enterprise feature only.

I was thinking about making it part of the CI job and "failing build" if it is not formatted according to the lint, the same way we fail on SpotBugs.

tobiasKaminsky:
Right, I misunderstood it.
Then this is already implemented:

image

AndyScherzinger:
seems to work while the report for ktlint ist really a html file but rather a `br` seperated list of issues that have been found (no body, no header, etc.)

tobiasKaminsky:
> seems to work while the report for ktlint ist really a html file but rather a `br` seperated list of issues that have been found (no body, no header, etc.)

Yeah, this is just a quick & easy list. 
Do you think we reall need the header/body stuff?

AndyScherzinger:
for now, I am completely fine with the solution 👍

PR #4073, opened 2019-05-27
Move PowerUtils to dedicated power management service

ezaquarii:

### DIFF
--- a/src/main/java/com/nextcloud/client/device/DeviceInfo.kt
+++ b/src/main/java/com/nextcloud/client/device/DeviceInfo.kt
@@ -1,3 +1,24 @@
+/*
+ * Nextcloud Android client application
+ *
+ * @author Chris Narkiewicz
+ *
+ * Copyright (C) 2019 Chris Narkiewicz <hello@ezaquarii.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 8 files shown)


src/main/java/com/nextcloud/client/device/Package.md
```diff
@@ -0,0 +1,4 @@
+# Package com.nextcloud.client.device
```
ezaquarii: This is Kotlin (KDoc to be specific) way of docimenting packages and modules.

src/main/java/com/nextcloud/client/device/DeviceInfo.kt
```diff
@@ -0,0 +1,8 @@
+package com.nextcloud.client.device
+
+import android.os.Build
+
+class DeviceInfo {
```
ezaquarii: This should save us from rolling some `PowerMock` kludge test API-level-dependent code.

Nevertheless there will be a point in time where mocking platform code will require `Robolectric`, but I'd rather delay that until it's really needed.

src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
ezaquarii: Classes are `final` in Kotlin by default, so mocking require specialized mock maker.

AndyScherzinger:
@ezaquarii nice work, looking good to me. Can you add the license header to files? :+1;

tobiasKaminsky:
@ezaquarii thanks for the detailed explanation on new kotlin stuff :-)

I added gpl license header, so that I can merge it once CI is green.

PR #4076, opened 2019-05-28
Support for Chromebook

tobiasKaminsky:
Chromebooks can have no camera, so this must be a non-required feature.
Therefore also the check if we should show the "upload from camera" button at all.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/fragment/OCFileListBottomSheetDialog.java
+++ b/src/main/java/com/owncloud/android/ui/fragment/OCFileListBottomSheetDialog.java
@@ -28,6 +28,7 @@ import android.widget.TextView;
 
 import com.google.android.material.bottomsheet.BottomSheetBehavior;
 import com.google.android.material.bottomsheet.BottomSheetDialog;
+import com.nextcloud.client.device.DeviceInfo;
 import com.owncloud.android.R;
 import com.owncloud.android.lib.resources.status.OCCapability;
 import com.owncloud.android.ui.activity.FileActivity;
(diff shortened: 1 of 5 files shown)


src/main/java/com/owncloud/android/ui/fragment/OCFileListBottomSheetDialog.java
```diff
@@ -97,6 +101,10 @@ protected void onCreate(Bundle savedInstanceState) {
             templates.setVisibility(View.VISIBLE);
         }
 
+        if (!getContext().getPackageManager().hasSystemFeature(PackageManager.FEATURE_CAMERA)) {
```
ezaquarii: I think this could go into our new `com.nextcloud.client.device.DeviceInfo` that looks like a good place to collect information about hardware.

tobiasKaminsky: Nice idea, I will do it tomorrow.

ezaquarii: For the time being you can just move it into a static helper for now. Cleanup can follow.

tobiasKaminsky: Done, though I could not get it to work with dagger…

tobiasKaminsky:
So, I moved DeviceInfo initialization to calling fragment.

AndyScherzinger:
Fix seems fine to me, so waiting for an approval form @ezaquarii - I'd say since it is to be considered a bugfix (to support chromebooks) it be nice to have it on master / 3.7.0 release and cleanups/housekeepings can be done separately :)

PR #4078, opened 2019-05-28
Battery charging status API

ezaquarii:
Battery charing checks are currently performed using evernote jobs API.
This ties our power management to rather unrelated 3rd party library.

This change provides battery check API via app power management
abstraction. It will also allow to access native battery API on
Android Lollipop and newer in the future.

This change does not migrate any current code to new API.

### DIFF
--- a/src/test/java/com/nextcloud/client/device/TestPowerManagementService.kt
+++ b/src/test/java/com/nextcloud/client/device/TestPowerManagementService.kt
@@ -21,84 +21,187 @@
 
 package com.nextcloud.client.device
 
+import android.content.Context
+import android.content.Intent
+import android.os.BatteryManager
 import android.os.Build
 import android.os.PowerManager
-import com.nhaarman.mockitokotlin2.never
-import com.nhaarman.mockitokotlin2.verify
-import com.nhaarman.mockitokotlin2.whenever
+import com.nhaarman.mockitokotlin2.*
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
+import org.junit.runners.Suite
 import org.mockito.Mock
 import org.mockito.MockitoAnnotations
-import org.mockito.junit.MockitoJUnitRunner
 
-@RunWith(MockitoJUnitRunner::class)
+@RunWith(Suite::class)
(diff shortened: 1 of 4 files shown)


ezaquarii:
I plan to remove those calls:

https://github.com/nextcloud/android/blob/a8829c6c68be2d453792e9988e2397105b4e7d8a/src/main/java/com/owncloud/android/files/services/FileUploader.java#L416

ezaquarii:
@tobiasKaminsky @AndyScherzinger I still need to massage ktlint but otherwise it's ready for review.

src/test/java/com/nextcloud/client/device/TestPowerManagementService.kt
```diff
@@ -21,84 +21,187 @@
...
 
+import android.content.Context
+import android.content.Intent
+import android.os.BatteryManager
 import android.os.Build
 import android.os.PowerManager
-import com.nhaarman.mockitokotlin2.never
-import com.nhaarman.mockitokotlin2.verify
-import com.nhaarman.mockitokotlin2.whenever
+import com.nhaarman.mockitokotlin2.*
```
AndyScherzinger: Do we want to use wildcard import statements?

ezaquarii: Damn, IDE is converted them to wildcard...

ezaquarii: I generally don't have anything against wildcards in tests, as the `Assert` and `Mockito` imports look quite bonkers.

ezaquarii:
@tobiasKaminsky @AndyScherzinger I'm ready for merge here.

src/test/java/com/nextcloud/client/device/TestPowerManagementService.kt
ezaquarii: Given that IDE does this by default (and is pretty smart about when do wildcard import and when not), I guess that trying to fight it is not going to be very productive in general, unless we force everybody to change their IDE settings.

ezaquarii:
@tobiasKaminsky @AndyScherzinger  Can we merge it? it adds some api, but not changes any existing code. Should be zero-risk merge.

AndyScherzinger:
Let's get it in then 👍

Nice work as always!

PR #4085, opened 2019-05-29
Copy internal link

tobiasKaminsky:
TODO
- [x] fine adjust layout
- [x] show copy/send link dialog

As discussed in https://github.com/nextcloud/server/issues/12877

Server mockup:

Android: 


tobiasKaminsky:
This will not get into 3.7.0 as then sharing view looks half finished.
Instead this will be in 3.8 when complete sharing view, including multiple share links, projects, etc. is in.

tobiasKaminsky:
Feedback by @jancborchardt 
> @tobiasKaminsky looks good! Only 2 details:
> 
> - The "Copy internal link" text color should be black / normal text color too. Sorry this was messed up in my mockup. :) The subline is fine.

There are multiple things not black
- "search users and groups"
- underline
- file size
- file date
- "no data shared with users yet"

> - The entry should be very last in the list (yes even after all the sharees) because otherwise it’s going to call undue attention again. Just like projects are also below sharees in the web interface.

I'll do.

tobiasKaminsky:
For further sharing enhancements, we should discuss this in a separate issue/PR.

src/main/java/com/owncloud/android/ui/fragment/FileDetailSharingFragment.java
```diff
@@ -200,6 +204,14 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
 
         setupView();
 
+        // todo extract
+        internalLinkIcon.getBackground().setColorFilter(getResources().getColor(R.color.grey_db),
+                                                        PorterDuff.Mode.SRC_IN);
+        internalLinkIcon.getDrawable().mutate().setColorFilter(getResources().getColor(R.color.black),
+                                                               PorterDuff.Mode.SRC_IN);
+
```
AndyScherzinger: Less line breaks would be enough 😉 in my opinion

src/main/java/com/owncloud/android/ui/fragment/FileDetailSharingFragment.java
```diff
@@ -325,6 +337,18 @@ public void toggleShareByLink() {
         }
     }
 
+    @OnClick(R.id.copy_internal_container)
+    public void copyInternalLink() {
+        try {
+            String baseUrl = AccountUtils.getBaseUrlForAccount(requireContext(), account);
+            String internalLink = baseUrl + OCFile.PATH_SEPARATOR + file.getLocalId();
```
AndyScherzinger: Why all the variables? I'd suggest to have a helper method for creating the link strings

ezaquarii: `AccountUtils` is deprecated since `UserAccountManager` arrived.
Base URI is property of `OwnCloudAccount`.

PR #4116, opened 2019-06-05
Bump to 3.7.0 RC1

tobiasKaminsky:
Library 1.5.0-rc1 is also a pre-release, so if we need to change something we do not "waste" stable versions.

@AndyScherzinger @ezaquarii please feel free to add anything I might miss to changelog.


AndyScherzinger:
Fine with me, also the changelog. 3.7.0 doesn't ship many features but is rather an under the hood release.

PR #4117, opened 2019-06-05
Remove legacy AccountUtils

ezaquarii:


src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
```diff
@@ -107,7 +105,38 @@ public boolean exists(Account account) {
 
     @Nullable
     public Account getCurrentAccount() {
-        return AccountUtils.getCurrentOwnCloudAccount(context);
+        Account[] ocAccounts = getAccounts();
```
ezaquarii: This is copied verbatimg from `AccountUtils.getCurrentOwnCloudAccount()`

src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
```diff
@@ -258,9 +281,7 @@ public void migrateUserId() {
             success = true;
         }
 
-        if (success) {
```
ezaquarii: This responsibility has to be moved to caller. At some point, a migration manager should be responsible for keeping track of applied migrations.

src/main/java/com/nextcloud/client/di/AppModule.java
```diff
@@ -63,11 +63,6 @@ Resources resources(Application application) {
         return application.getResources();
     }
 
-    @Provides
-    AppPreferences preferences(Application application) {
```
ezaquarii: Moved to `PreferencesModule` due to scope restrictions (package-private constructor).

src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
```diff
@@ -74,15 +75,18 @@
 
     private final Context context;
     private final SharedPreferences preferences;
+    private final CurrentAccountProvider currentAccountProvider;
 
     public static AppPreferences fromContext(Context context) {
-        SharedPreferences prefs = android.preference.PreferenceManager.getDefaultSharedPreferences(context);
-        return new AppPreferencesImpl(context, prefs);
+        final CurrentAccountProvider currentAccountProvider = UserAccountManagerImpl.fromContext(context);
```
ezaquarii: This is very sad. Due to circular dependency between preferences and account, I had to apply this temporary hack.

It's still *yuck*, but at least external code doesn't know about it anymore.

src/main/java/com/owncloud/android/utils/ThemeUtils.java
```diff
@@ -544,7 +544,8 @@ private static OCCapability getCapability(Account acc, Context context) {
         if (acc != null) {
             account = acc;
         } else if (context != null) {
-            account = AccountUtils.getCurrentOwnCloudAccount(context);
+            // TODO: refactor when dark theme work is completed
+            account = UserAccountManagerImpl.fromContext(context).getCurrentAccount();
```
ezaquarii: Making it right would cause too much impact on `ThemeUtils` at this point. This stinks, but I'm not ready to refactor this atm.

ezaquarii:
@AndyScherzinger @tobiasKaminsky I fixed all CI issues here. Ready for review.

scripts/uploadReport.sh
```diff
@@ -14,6 +14,9 @@ USER=$1
 PASS=$2
 TYPE=$4
 
+set -x
+set -e
```
tobiasKaminsky: Is this needed? I "fear" that this might expose password/sensitive data

src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
tobiasKaminsky: Totally agree!

tobiasKaminsky:
As an overall thing (as we touch code already here):
We should really distinguish between Account and OwncloudAccount:
Account: stored & handled by AccountManager from Android
OwnCloudAccount: created upon Account, with additional informations (BASE_URL, DISPLAY_NAME)

For instance on setCurrentOwncloudAccount(String accountName) we do not use OwncloudAccount at all.

src/main/java/com/owncloud/android/utils/ThemeUtils.java
tobiasKaminsky: Totally fine to do this later :+1:

scripts/uploadReport.sh
ezaquarii: I removed `set -x` which enables debug trace.

`set -e` causes script to fail fast on first non-zero exit code, preventing Jenkins from running `curl` when previous commands fail for whatever reason (we had multiple cases of `curl` trying to upload something for hours).

ezaquarii:
> As an overall thing (as we touch code already here):
> We should really distinguish between Account and OwncloudAccount:

I understand that we want to extend `android.accounts.Account` with extra information, but I fail to understand the benefit of having 2 entities responsible for 1 thing. It clearly breaks single responsibility principle for me, forces developers to perform conversions between types, causes extra pain around tests (`Account` pulls significant portion of platform API).

`OwncloudAccount` can surely have `android.accounts.Account` as an implementation detail and delegate calls to it.

Did I miss anything?

scripts/uploadReport.sh
tobiasKaminsky: :+1: for -e

tobiasKaminsky:
> `OwncloudAccount` can surely have `android.accounts.Account` as an implementation detail and delegate calls to it when needed.
> 
> Did I miss anything?

No, I think this is just a left-over from old code.
As this then will be a bigger change, let us keep this PR as is and then do a discussion and PR separate.

ezaquarii:
> > `OwncloudAccount` can surely have `android.accounts.Account` as an implementation detail and delegate calls to it when needed.
> > Did I miss anything?
> 
> No, I think this is just a left-over from old code.
> As this then will be a bigger change, let us keep this PR as is and then do a discussion and PR separate.

Yeah, I definitely don't plan to touch this issue in the near time - there are plenty of other structural problems to solve first.

PR #4122, opened 2019-06-06
use own getAccounts() instead of AccountManager from AndroidSDK

tobiasKaminsky:


tobiasKaminsky:
otherwise we try to delete every account, which fails as we are not allowed to delete other app's accounts.

Issue #3421, opened 2019-01-03
Cannot turn off app locking


tomasz-grobelny:
### Actual behaviour
- Cannot turn off application locking in settings. Every time I finish editing a file I am asked to unlock the application. Every time I start the application I am asked to unlock it.

### Expected behaviour
- Application should not ask for unlocking.
 
### Steps to reproduce
1. Run the application.
2. Unlock (I use swipe pattern)
3. Go to nextcloud app settings and set locking to none.
4. Exit app.
5. Run app - unlock screen appears again.

### Environment data
Android version: 8.0.0.162(C432)
Device model: FIG-LX1
Stock or customized system: Stock
Nextcloud app version: 3.4.1

tobiasKaminsky:
Strange, do you see at step 3 a confirmation?

tomasz-grobelny:
Do you mean if I need to enter my unlocking code to confirm resigning from unlocking code? Yes. The setting seems to be saved, but not respected in subsequent application runs.

One more observation: I recently installed dev version and it seems to work there just fine.

tobiasKaminsky:
> The setting seems to be saved, but not respected in subsequent application runs.

Strange…



> One more observation: I recently installed dev version and it seems to work there just fine.

So for some reason the saved setting is not respected in the stable version.

Unfortunately I have no idea what to do, as I do not have access to the device…

tomasz-grobelny:
Any idea what I could check or the only option that comes to mind is to set up development environment and attach debugger (which might not be worth it as it seems to work in dev version)? Maybe I could analyse the configuration? Where is this setting stored?

tobiasKaminsky:
Setting is stored in shared_preferences.xml, called "lock", with one of:
"none"
"passcode"
"device_credentials"

christianlupus:
I think I am affected by the same bug. I cannot disable the protection mode as well. As far as I remember, I had to use my fingerprint/swipe for the very first time I used the app. 

Where is this file `shared_preferences.xml` to be found?

**Environment data**
Android version: 9
Device model: Xiaomi Redmi A2 Lite
Stock or customized system: Stock
Nextcloud app version: 3.6.0 (stable from Google play store)

christianlupus:
@tobiasKaminsky Can you tell me, where to find this XML file? Do I need to use ADB to open it or can I use a normal file explorer to get it?

christianlupus:
[...]

PR #4133, opened 2019-06-10
Engineering Test Mode

ezaquarii:
If you are wondering where ETM (Engineering Test Mode) term came from, it's automotive industry. If you have a car, you can probably play with it there too. :)

This module allows developers and dev users to access some internal application data.
The module is indended to be a dumping ground for various diagnostic and development tools.

1. Go to Settings
2. Scroll to section "Dev" (bottom)
3. Select "Engineering Test Mode"

### DIFF
--- a/src/main/java/com/nextcloud/client/etm/pages/EtmPreferencesFragment.kt
+++ b/src/main/java/com/nextcloud/client/etm/pages/EtmPreferencesFragment.kt
@@ -1,0 +1,73 @@
+/*
+ * Nextcloud Android client application
+ *
+ * @author Chris Narkiewicz
+ * Copyright (C) 2019 Chris Narkiewicz <hello@ezaquarii.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.nextcloud.client.etm.pages
+
+import android.content.Intent
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.Menu
+import android.view.MenuInflater
+import android.view.MenuItem
+import android.view.View
+import android.view.ViewGroup
+import com.nextcloud.client.etm.EtmBaseFragment
+import com.owncloud.android.R
+import kotlinx.android.synthetic.main.fragment_etm_preferences.*
+
+class EtmPreferencesFragment : EtmBaseFragment() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setHasOptionsMenu(true)
+    }
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        return inflater.inflate(R.layout.fragment_etm_preferences, container, false)
+    }
+
+    override fun onResume() {
+        super.onResume()
+        val builder = StringBuilder()
+        vm.preferences.forEach { builder.append("${it.key}: ${it.value}\n") }
+        etm_preferences_text.text = builder
+    }
+
+    override fun onCreateOptionsMenu(menu: Menu?, inflater: MenuInflater?) {
+        super.onCreateOptionsMenu(menu, inflater)
+        inflater?.inflate(R.menu.etm_preferences, menu)
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem?): Boolean {
+        return when (item?.itemId) {
+            R.id.etm_preferences_share -> {
+                onClickedShare(); true
+            }
+            else -> super.onOptionsItemSelected(item)
+        }
+    }
+
+    private fun onClickedShare() {
+        val intent = Intent(Intent.ACTION_SEND)
+        intent.putExtra(Intent.EXTRA_SUBJECT, "Nextcloud preferences")
+        intent.putExtra(Intent.EXTRA_TEXT, etm_preferences_text.text)
+        intent.type = "text/plain"
+        startActivity(intent)
+    }
+}
--- a/src/main/res/layout/activity_etm.xml
+++ b/src/main/res/layout/activity_etm.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+    Nextcloud Android client application
+
+    @author Chris Narkiewicz
+    Copyright (C) 2019 Chris Narkiewicz <hello@ezaquarii.com>
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program. If not, see <http://www.gnu.org/licenses/>.
+-->
+<androidx.coordinatorlayout.widget.CoordinatorLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context=".etm.EtmActivity">
+
+    <include layout="@layout/toolbar_standard"/>
+
+    <FrameLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        app:layout_behavior="@string/appbar_scrolling_view_behavior"
+        android:id="@+id/etm_page_container">
+
+    </FrameLayout>
+
+</androidx.coordinatorlayout.widget.CoordinatorLayout>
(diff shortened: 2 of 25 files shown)


ezaquarii:
@tobiasKaminsky @AndyScherzinger I created this PR to help a user with #3421 that was unable to share relevant diagnostics with us.

Please review and comment. It's WIP as I don't want to invest more time into it before we make sure it's worth it.

christianlupus:
@ezaquarii Now things are gettign strange. The app is installed parallel to the main app. With the QA version, I can open the app without swipe code even if I set it up like this in the settings.I tried to quit the app by going into the task manager and there swiping out the app. Still the app opens directly without protection, content is shown in the task switcher and I can make screenshots.

So unfortunately this branch is not working as intended as a debug tool....

ezaquarii:
@christianlupus You are right that this *"QA"* version is installed as a separate app. QA versions are intended for evaluation without overwriting your daily-driver app, especially that it might not work or cause data corruption. This way your data is safe and the experiment is done in a "sandbox".

Regarding the locking mechanism - you have probably found another bug. We'll be investigating it when @tobiasKaminsky  comes back.

ezaquarii:
@tobiasKaminsky Let me know what you think about it - I don't want to invest more time into it if it's a bad idea. (-:

tobiasKaminsky:
> @tobiasKaminsky Let me know what you think about it - I don't want to invest more time into it if it's a bad idea. (-:

If we put this for now only on dev/daily version, I am fine with it.
So if an user experiences a bug where we need more infos, we can ask him to test dev version (which then is based on master, where the problem is maybe already fixed) and additional can ask them for infos via ETM.

Big question would be, which infos do we want to have collected…

ezaquarii:
> Big question would be, which infos do we want to have collected…

Well, when you browse the backlog of bugs, it is quite clear that we're lacking audit capabilities as there is no way of getting meaningful data from users currently. My intention was to create a dumping ground for such tools, as I needed to investigate preferences relevant to #3421 and there was no tool to do it.

I'm sure the tech support would love to have it.

Quite frankly, if you could reach them and ask them for feature requests, that would be great. I'm sure they are having tons of interesting stories to tell...

AndyScherzinger:
@tobiasKaminsky what do you think? I agree with @ezaquarii that we do hit certain scenarios every now and then where we can't reproduce an issue and depend on further input that not all reporters can easily provide or can't provide at all, so having such a feature (with on/off of course) would be helpful I think.

src/main/java/com/nextcloud/client/etm/pages/EtmPreferencesFragment.kt
```diff
@@ -0,0 +1,55 @@
...
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.Menu
+import android.view.MenuInflater
+import android.view.MenuItem
+import android.view.View
+import android.view.ViewGroup
+import com.nextcloud.client.etm.EtmBaseFragment
+import com.owncloud.android.R
+import kotlinx.android.synthetic.main.fragment_etm_preferences.*
```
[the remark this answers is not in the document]
ezaquarii: I really think this tool is not production quality... :(

ezaquarii:
@tobiasKaminsky @AndyScherzinger I polished the ETM so it is in mergable state.
ETM entry is available in Dev settings, so it should not be visible for normal users.

The biggest upside of this PR is that it enables us to use injectable `ViewModel`s. There is also a complete VM test suite as an example.

.idea/codeStyles/Project.xml
```diff
@@ -59,6 +59,10 @@
         </value>
       </option>
     </JavaCodeStyleSettings>
+    <JetCodeStyleSettings>
```
ezaquarii: Codacy does not accept wildcard imports, so I guess this should be disabled in IDE.

AndyScherzinger:
I am currently traveling without an ide at hand but will check it out when I am back home 👍

.idea/codeStyles/Project.xml
tobiasKaminsky: I changed this here, which then will be soon on master: https://github.com/nextcloud/android/pull/4253/files#diff-c0176795f4ea97568151067bbe2f0bb5R62

tobiasKaminsky: Ah, you also changed it, good.
(github was only showing line 62)

src/main/res/values/strings.xml
```diff
@@ -873,4 +873,8 @@
     <string name="copy_internal_link">Copy internal link</string>
     <string name="copy_internal_link_subline">Only works for users with access to this folder</string>
     <string name="failed_to_download">Failed to pass file to download manager</string>
+
+    <string name="etm_title" translatable="false">Engineering Test Mode</string>
+    <string name="etm_preferences" translatable="false">Preferences</string>
+    <string name="etm_share" translatable="false">Share</string>
```
tobiasKaminsky: Those should be translated via Transifex, thus please remove translatable

src/main/res/layout/generic_list_item_single_line.xml
```diff
@@ -0,0 +1,51 @@
...
+        android:textSize="16sp"
+        android:lines="1"
+        android:ellipsize="end"
+        android:textColor="?android:attr/textColorPrimary"
+        tools:text="Single line of text"/>
+
+    <androidx.appcompat.widget.AppCompatImageView
+        style="@style/GenericListItemSecondaryAction"
+        android:id="@+id/secondary_action"
+        android:src="@drawable/ic_placeholder_black_24dp"/>
```
tobiasKaminsky: This is not shown for me:


Can you also add some margin/spacing?

src/main/res/values/styles.xml
```diff
@@ -284,4 +285,33 @@
 		<item name="android:textSize">16sp</item>
 		<item name="textAllCaps">false</item>
 	</style>
+
+    <style name="GenericListItemSingleLine">
```
tobiasKaminsky: I am unsure if this should/needs to be here.
I "fear" that we do not have such generic lists, e.g. listing files, listing user account infos, activity stream; all are handled in a very different way (UI wise).

So I think we should move these "special" UI implementation to ETM. 
Drawback is that we might have to do copy&paste if we have such a plain list.
On the other hand we are sure that it looks as expected when changing something (as we do not change it indirectly on other places).

tobiasKaminsky:
> The biggest upside of this PR is that it enables us to use injectable `ViewModel`s. There is also a complete VM test suite as an example.

This is indeed great.
I'll have a look how to use my #3781 with this PR…

src/main/res/drawable/ic_placeholder_black_24dp.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
tobiasKaminsky: If we need this, then it should get a proper name, e.g. ic_android.xml

src/main/res/layout/generic_list_item_single_line.xml
AndyScherzinger: Padding schould be `standard_padding`

src/main/res/values/strings.xml
ezaquarii: Removed `translatable=false`

src/main/res/values/styles.xml
ezaquarii: I renamed those components into `MaterialSomethingSomething` - there should be no confusion about the purpose if it.

src/main/res/drawable/ic_placeholder_black_24dp.xml
ezaquarii: I removed this icon and replaced `android:src` with `tools:src`.

src/main/res/layout/generic_list_item_single_line.xml
ezaquarii: Added padding the preferences textview

ezaquarii: @tobiasKaminsky This view uses a simple textview, as it was enough to debug the situation, but we can improve it by adding a list with items.

AndyScherzinger:
@tobiasKaminsky can you re-review. Imho all review commetns have been fixed, I just optimized one padding value for toolbar text alignment, so this is imho ready to be merged now 👍 

Nice work @ezaquarii - as always 🎉

AndyScherzinger:
All things turned green, just waiting for @tobiasKaminsky's final approval :)

ezaquarii:
@tobiasKaminsky ping!

src/main/res/layout/activity_etm.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tobiasKaminsky: Can you rename this to etm_layout.xml?
Then we have this in sync to all other names.

src/main/res/layout/fragment_etm_menu.xml
```diff
@@ -0,0 +1,32 @@
+<!--
```
tobiasKaminsky: Maybe etm_fragment_menu? So that etm is in front

src/main/res/layout/fragment_etm_preferences.xml
```diff
@@ -0,0 +1,33 @@
+<!--
```
tobiasKaminsky: Same here, etm in front?

tobiasKaminsky:
Just litte naming ideas: have ETM in front of activity/fragment.
Other than that :+1: :+1: 

As always thank very very much for your valuable contributions!

AndyScherzinger:
@tobiasKaminsky I am with @ezaquarii's actual file names, se for example:
* https://medium.com/mindorks/android-resource-naming-convention-42e4e8026614
* https://github.com/ribot/android-guidelines/blob/master/project_and_code_guidelines.md

So files should start with the "type" So I think, this should be merged as-is.

tobiasKaminsky:
I was just looking at our current naming scheme, e.g:
activity_list_item.xml
contactlist_fragment.xml
file_details_fragment.xml

I am totally up for changing this. I just do not want that we have different naming schemes…
I guess we could also have subfolders in layout to split up activity, fragment, etc.

AndyScherzinger:
> I am totally up for changing this. I just do not want that we have different naming schemes…

absolutely, yes

> I guess we could also have subfolders in layout to split up activity, fragment, etc.

I don't think Android is support sub-folders with the layout structure!

AndyScherzinger:
@tobiasKaminsky opened #4278 for discussion of naming pattern(s) - I'd still vote to simply merge it (or rename and merge, but not further postpone the merge)

tobiasKaminsky:
As we will go with new naming scheme, we can merge this.

PR #4145, opened 2019-06-14
Allow "save as" and "print" to be downloaded from Collabora

tobiasKaminsky:


AndyScherzinger:
@tobiasKaminsky should this be backported?

PR #4146, opened 2019-06-14
Test login on integrationTests

tobiasKaminsky:
This is for starting & testing a real login with webview.
For now it is ok to have only one test.

This will fail as of missing #6476, but this is a good test to see if it works.


AndyScherzinger:
@tobiasKaminsky the login test is actually failing xD

Issue #4168, opened 2019-06-21
Nextcloud crashes


jmue:
### Actual behaviour
- Nextcloud crashes while selecting app drawer entries

### Expected behaviour
- Nextcloud should not crash
 
### Steps to reproduce
1. Open nextcloud app
2. Open app drawer, click on 'Activities'
3.  Open app drawer, click on 'Favorites'

### Environment data
Android version: 9.0, 7.0

Device model: 
Emulator, Moto G

Stock or customized system:

Nextcloud app version:
3.7 RC3

Nextcloud server version:

#### Web server error log
```
Insert your webserver log here
```

#### Nextcloud log (data/nextcloud.log)
```
    --------- beginning of crash
2019-06-21 22:30:30.917 25627-25627/com.nextcloud.client E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.nextcloud.client, PID: 25627
    org.parceler.ParcelerRuntimeException: Unable to find generated Parcelable class for com.owncloud.android.ui.events.SearchEvent, verify that your class is configured properly and that the Parcelable class com.owncloud.android.ui.events.SearchEvent$$Parcelable is generated by Parceler.
        at org.parceler.Parcels$ParcelCodeRepository.get(Parcels.java:154)
        at org.parceler.Parcels.wrap(Parcels.java:73)
        at org.parceler.Parcels.wrap(Parcels.java:57)
        at com.owncloud.android.ui.activity.DrawerActivity.handleSearchEvents(DrawerActivity.java:536)
        at com.owncloud.android.ui.activity.DrawerActivity.selectNavigationItem(DrawerActivity.java:450)
        at com.owncloud.android.ui.activity.DrawerActivity.access$300(DrawerActivity.java:119)
        at com.owncloud.android.ui.activity.DrawerActivity$3$1.run(DrawerActivity.java:344)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6669)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)

```
[...]

PR #4178, opened 2019-06-25
fix parcelable with using kapt

tobiasKaminsky:
Triggered by:
- create a new document via collabora


ezaquarii:
I'm not sure this will work. It looks like Lombok doesn't work with Kotlin. Honestly I didn't see this coming.

jmue:
Fixes #4168. Thank you!

PR #4179, opened 2019-06-26
fix wrong tinted button on upload file within collabora

### LINKED ISSUE #4177: Collabora: upload file: wrong tinted button
- open document with collabora
- upload file from cloud

tobiasKaminsky:
Fix #4177

 -> 


src/main/res/layout/files_picker.xml
```diff
@@ -53,15 +54,17 @@
...
             android:id="@+id/folder_picker_btn_cancel"
+            android:theme="@style/OutlinedButton"
             android:layout_width="0dp"
             android:layout_height="wrap_content"
             android:layout_marginEnd="@dimen/standard_half_margin"
             android:layout_marginRight="@dimen/standard_half_margin"
             android:layout_weight="1"
-            android:text="@string/common_cancel"
-            android:theme="@style/Button"/>
+            android:stateListAnimator="@null"
```
AndyScherzinger: `android:stateListAnimator="@null"` shouldn't be needed imho (we also don't have this elsewhere  and it creates another lint warning.

tobiasKaminsky: Lint warning is that this is only supported in >=5.0, but this is ok as this files_picker is only used within Collabora, which is also >=5.0.

We do not need it elsewhere as we use another theme for all other things.
But files_picker is a dialog and this is another theme with other default settings. I failed to do it in another way.

I would to get this in and maybe, if time permits, you can have a look to find a better approach? :heart:

AndyScherzinger: Fine with me 👍

AndyScherzinger:
@tobiasKaminsky need manual backport :/

tobiasKaminsky:
Backport in https://github.com/nextcloud/android/pull/4192

PR #4180, opened 2019-06-26
Improve app store screenshot designs

jancborchardt:
@tobiasKaminsky as discussed. :) And let’s chat about the wording later too.

- Change text from Light to better readable Regular
- Remove text shadow
- Use simpler device container in white
- Add slight drop shadow to device (Material ;))


tobiasKaminsky:
This needs changes from my/script side to get screenshots again.

AndyScherzinger:
> This needs changes from my/script side to get screenshots again.

Would then be nice to get this done for 3.8 💃

tobiasKaminsky:
@jancborchardt I sent you the screenshots, waiting for your feedback :-)

tobiasKaminsky:
This should be in "1. developing" until @jancborchardt apporves.

AndyScherzinger:
> This should be in "1. developing" until @jancborchardt apporves.

since labelled with "design" and Jan needs to approve that actually implies "review", at least to me it does.

jancborchardt:
Looking suuper awesome @tobiasKaminsky! :)

Only one thing: On the screenshot with DAVx5, instead of putting grey cover onto the other things, I would instead make those parts less opacity. (If you give me the screenshot I can do that too. :)

tobiasKaminsky:
> Only one thing: On the screenshot with DAVx5, instead of putting grey cover onto the other things, I would instead make those parts less opacity. (If you give me the screenshot I can do that too. :)

This is a hidden rectangle in svg, which I just make visible if needed.
I am not sure if you can do a blur effect as a layer, but if so it would be cool.

jancborchardt:
> > Only one thing: On the screenshot with DAVx5, instead of putting grey cover onto the other things, I would instead make those parts less opacity. (If you give me the screenshot I can do that too. :)
> 
> This is a hidden rectangle in svg, which I just make visible if needed.
> I am not sure if you can do a blur effect as a layer, but if so it would be cool.
> 
> https://github.com/nextcloud/android/blob/8435e210a16809983a8f86274c36708a08e214c4/scripts/screenshots/phone.svg?short_path=64c924e#L114-L120

Ah, easy then – if you just change the color from grey/black to white it should work then, as with white/#ffffff and slight transparency it should have the same effect I described.

jancborchardt:
Oh and the position of the grey rectangle is also off, it actually obscures the first entry which is the one about sync:
06_davdroid_1564122819456

tobiasKaminsky:
06_davdroid_1564565366008

I'll merge it so we have transifex.
Once we have enough translations, I'll update play store.

jancborchardt:
@tobiasKaminsky per-fect!

PR #4189, opened 2019-06-28
Provide a banal 'paste' postmessage implementation.

mmeeks:
Browsers will allow apps to populate the clipboard, but not read
from it without user input. This patch makes it possible for
context menus with 'Paste' in them to trigger a real OS paste
when hosted inside the webview.

Signed-off-by: Michael Meeks <michael.meeks@collabora.com>

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/RichDocumentsWebView.java
+++ b/src/main/java/com/owncloud/android/ui/activity/RichDocumentsWebView.java
@@ -35,8 +35,8 @@ import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.text.TextUtils;
-import android.view.View;
 import android.view.KeyEvent;
+import android.view.View;
 import android.webkit.JavascriptInterface;
 import android.webkit.ValueCallback;
 import android.webkit.WebChromeClient;
@@ -423,8 +423,10 @@ public class RichDocumentsWebView extends ExternalSiteWebView {
         @JavascriptInterface
         public void paste() {
            // Javascript cannot do this by itself, so help out.
-            webview.dispatchKeyEvent(new KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_PASTE));
-            webview.dispatchKeyEvent(new KeyEvent(KeyEvent.ACTION_UP, KeyEvent.KEYCODE_PASTE));
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                webview.dispatchKeyEvent(new KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_PASTE));
+                webview.dispatchKeyEvent(new KeyEvent(KeyEvent.ACTION_UP, KeyEvent.KEYCODE_PASTE));
+            }
         }
     }
 
--- a/src/main/java/com/owncloud/android/ui/activity/RichDocumentsWebView.java
+++ b/src/main/java/com/owncloud/android/ui/activity/RichDocumentsWebView.java
@@ -36,6 +36,7 @@ import android.os.Build;
 import android.os.Bundle;
 import android.text.TextUtils;
 import android.view.View;
+import android.view.KeyEvent;
 import android.webkit.JavascriptInterface;
 import android.webkit.ValueCallback;
 import android.webkit.WebChromeClient;
@@ -417,6 +418,13 @@ public class RichDocumentsWebView extends ExternalSiteWebView {
             } catch (JSONException e) {
                 Log_OC.e(this, "Failed to parse rename json message: " + e);
             }
+        }
+
+        @JavascriptInterface
+        public void paste() {
+           // Javascript cannot do this by itself, so help out.
+            webview.dispatchKeyEvent(new KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_PASTE));
+            webview.dispatchKeyEvent(new KeyEvent(KeyEvent.ACTION_UP, KeyEvent.KEYCODE_PASTE));
         }
     }
 


tobiasKaminsky:
Can you give me steps how to test this?
I did this
- opened up a document within NC app
- long pressed on text to select it & copy it
- long pressed on a white area to paste it -> nothing happened
- edit -> insert text -> worked
(but latest step also worked without this PR)

mmeeks:
Hi Tobias,

On 01/07/2019 10:45, Tobias Kaminsky wrote:
> Can you give me steps how to test this?
> I did this
> 
>   * opened up a document within NC app
>   * long pressed on text to select it & copy it

	Great.

>   * long pressed on a white area to paste it -> nothing happened

	So - you'll need an updated C'bra Online build that posts the
PostMessage on paste at the right time to make that work - we're just
doing some final testing on that.

	ATB,

		Michael.

-- 
michael.meeks@collabora.com <><, GM Collabora Productivity
Hangout: mejmeeks@gmail.com, Skype: mmeeks
(M) +44 7795 666 147 - timezone usually UK / Europe

tobiasKaminsky:
> So - you'll need an updated C'bra Online build that posts the PostMessage on paste at the right time to make that work - we're just doing some final testing on that.

Hm, but how/by whom is this postMessage triggered?
(just for understanding)

As this is not in official release, there is no need for backporting it to 3.7.

tobiasKaminsky:
Please ping me, once there is a Collabora version, I can test (it can also be a docker tag/snapshot).

tobiasKaminsky:
When I now try to long-press to select test, I get following:

mmeeks:
Ah - this is easy; you need to double-tap to select text, not long-press =)

tobiasKaminsky:
I did this on our collabora staging server:

- typed text
- double click on it
- select a snippet
- select copy
- go to a new line
- long press -> paste

- at least on my emulator stock keyboard  I cannot paste…

mmeeks:
Any console output ? I wonder - it could be a cross-site origin issue - we attempt to post directly to the top window, which may cause concern. Thanks.

tobiasKaminsky:
No meaningful error/info output.

On my OnePlus3 with Swiftkey I can use the keyboard to paste the text snippet.
Is this the desired way?

mmeeks:
Sure, using the system paste functionality should work, this is primarily intended to make non-system paste work ie. via the context menus or menu -> edit -> paste. And you're sure you're using the updated/instrumented app ? - I'd be curious to know if the postmessage arrives of course. Thanks.

tobiasKaminsky:
> And you're sure you're using the updated/instrumented app ?

I am using the android app with this patch and collabora.nc (our staging server) for testing. This is correct, or?

tobiasKaminsky:
@mmeeks @juliushaertl if you also want to test this, in this PR is a QR code where you can install the app as "nextcloud qa" in parallel to existing one.

juliushaertl:
@mmeeks The android app doesn't use postmessages for the integration but a call to `window.RichDocumentsMobileInterface.paste()` on the top frame.

See https://github.com/nextcloud/richdocuments/blob/842d5f2ef665edabaed581701a849736adcfde5c/js/documents.js#L940-L959 for the richdocuments implementation

tobiasKaminsky:
Tested together with @juliushaertl and @mmeeks and works :tada:

AndyScherzinger:
@tobiasKaminsky doesn't this need another rebase and fixes in the score-file?

tobiasKaminsky:
Yes, now that https://github.com/nextcloud/android/pull/4253 is merged, I'll do a rebase.

AndyScherzinger:
@tobiasKaminsky there are two lint warnings:
> Field requires API level 24 (current min is 21): android.view.KeyEvent#KEYCODE_PASTE

tobiasKaminsky:
@mmeeks this is only possible for at least Nougat (Android 7), but RichDocuments/Collabora works with at least Lollipop (5.0).
Do you have an idea how we can support back to 5.0?
What does happen if I just do nothing on 5.0/6.0? Then pasting will just not work, or?

mmeeks:
Ah - if we do nothing in older versions, then there will just be a popup saying "Please use a system paste button" or somesuch, which is fine - we can cope with that.

Ideally we would have an 'internal paste' option inside that dialog I guess.

Wrt. invoking a paste method dynamically, I guess there is a Java mechanism for that that will work if it is there (?) - but no expert; sorry for not noticing that.

mmeeks:
Oh, and ... it is possible that there is a 'Paste Key' - that could be inserted instead. Oddly that may work just as well and be more compatible (?).

tobiasKaminsky:
I only found https://developer.android.com/guide/topics/text/copy-paste#PastePlainText, but this is for accessing content of clip board.

Pasting into webview/RichDocument is another step, which requires a javascript interface?
Then it would also be possible to copy text in any other android app and paste it into collabora (as right now you can only copy&paste within the same document, or?)

mmeeks:
With the other pieces of this work - you should be able to paste form anywhere, whenever you like without  using a system-paste event. The problem is that a system paste event cannot be triggered from Javascript. In consequence we have to get it emitted by the native code. That could (probably) be done with a paste key injection instead. I guess some code like:

BaseInputConnection  mInputConnection = new BaseInputConnection( findViewById(R.id.main_content), true);
KeyEvent kd = new KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_PASTE);
KeyEvent ku = new KeyEvent(KeyEvent.ACTION_UP, KeyEvent.KEYCODE_PASTE);
mInputConnection.sendKeyEvent(kd);
mInputConnection.sendKeyEvent(ku);

Or somesuch; or even InputConnecction's sendKeyEvent ... not sure; for the case where that API is not there (?).

tobiasKaminsky:
The problem with your new code idea is that KeyEvent.KEYCODE_PASTE only exists beginning from API 24 only.
As long as we do not find a way to do a paste on <24, nothing will happen on Lollipop and Marshmallow (5.x and 6.x).

tobiasKaminsky:
@mmeeks I would merge it as is:
>= Nougat: we have a proper paste
< Nougat (and >= Lollipop as we only support Collabora since then): we do get the "please use system paste" message.

Or do you have another idea?

tobiasKaminsky:
Merging after CI is green as it is for >= Nougat an improvement.

Issue #3061, opened 2018-09-19
Media Streaming Doesn't Work


windows10luke:
### Actual behaviour
- Tell us what happens
When tapping on a media file in the latest version of the app(3.3.0), it opens a player in the app, and then says in a little popup "Could not play the file using the builtin media player" and then it gives up.
### Expected behaviour
- Tell us what should happen
 It should play mp3 and mp4 files when selecting them.
### Steps to reproduce
1.  Tap a media file in the Nextcloud app
2.  Media Player opens and produces the error

### Environment data
Android version: 8.0

Device model: LG V30

Stock or customized system: Stock

Nextcloud app version: 3.3.0

Nextcloud server version:14.0

#### Web server error log
```
Insert your webserver log here
```

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```
**NOTE:** Be super sure to remove sensitive data like passwords, note that everybody can look here! You can use the Issue Template application to prefill some of the required information: https://apps.nextcloud.com/apps/issuetemplate

gradinaruvasile:
I can confirm this bug.
Although streaming works with VLC at least (if you have it installed) by selecting the "Stream with..." option.

johninstockie:
I can confirm this bug too, however I am experiencing the opposite in that the internal video player works if you use the stream with option, but not with normal behavior, and VLC does not work by any method.

breisig:
I am having this exact same issue even with mp3's. Have to either download the file for it to play or stream with 'vlc' to play.

galosre:
I completely with the above Plus it would be nice to play files by folder in case of multiple tracks (album)

DPTJKKVH:
Is there something I can provide aside from a logcat? I wanted to upload a logcat but noticed it contains lots of confidential information like folder names, sync names and such.

BeGunman:
can also confirm this bug! having this exact same issue even with mp3's. Have to either download the file for it to play or stream with 'vlc' to play.

alkamen:
I can confirm this bug.
Plays mp4 some Time, but not mp4, or wav.

DPTJKKVH:
Has any of you force password on share link enabled? I have a hunch that this is an issue. At least for us that cannot stream anything.

tobiasKaminsky:
@rullzer this should not be related, or?

rullzer:
Nope unrelated

tiredthinker:
[...]

PR #4208, opened 2019-07-08
New audio media player service

ezaquarii:
This is a new media service based on state machine to address #3061. I tried to salvage the old solution but decided to do a complete re-work. I firmly believe that new architecture will allow us to fix any current and future edge cases easily.

However, the work is not complete yet and there are few issues to consider before we move forward.
Some featutres are cut out, some didn't work anyway, so we need to make Decisions(tm).

### Restoring minimized player exists preview

1. Open media preview
2. Minimize the screen (media stops)
3. Restore app from recent apps
Expected: media preview screen, maybe resume playback?
Actual: media preview fragment is closed

I discovered that this also happens on current master, so this is an existing issue, not a regression.
This is most likely caused by `FileDisplayActivity.onAccountSet(false)` which re-sets displayed fragments.

### Total time is not displayed when streaming media

I could not confirm if this is regression, as streaming media does not work on prod and master.
It seems that `android.media.MediaPlayer` returns 0. I can imagine that media file header is not available in streaming mode, but also the media player could do something smart to work around it, so I'd like to have a 2nd opinion.

Here are my results:

| | mp3 | ogg |
| --- | --- | ---
| total | 0:00 | 0:00 |
| progress bar | 0:00 | stuck at 0:02 |
|rewind| nope | nope |

I'll need some help with testing it in other environments, as I'm not sure if this is expected to work (which require more work, obviously) or not (which require setting some flags to disable those controls).

### Media control is a bit wonky

Play-pause is not propagated correctly as it is polling the player. This is something to be addressed with further refactoing.

### Audio manager

~Audio focus handling is still not implemented, as it is not required at this stage.~
Implemented.

### Auto-stop on file deletion

Auto-stop on file deletion is currently removed (I left the placeholder to trigger it). Closing preview stops the player and it seems that background playback was not implemented anyway.

### Notification is removed for now

~Again, no background playback `==` not needed at this stage. Besides, it didn't provide controls, so more work is needed if we want to bring that into usable state.~
Old notification behavior is restored.

### ~Last but not least is the StateMachine framework~

No longer an issue - I'm a maintainer and `stateless4j` is in good shape again.

@tobiasKaminsky @AndyScherzinger @DPTJKKVH Please give it a go and let me know how we'd like to proceed with that.


src/main/java/com/nextcloud/client/media/PlayerStateMachine.kt
```diff
@@ -0,0 +1,101 @@
+package com.nextcloud.client.media
+
+import com.github.oxo42.stateless4j.StateMachine
+import com.github.oxo42.stateless4j.StateMachineConfig
+import org.slf4j.Logger
```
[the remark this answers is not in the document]
ezaquarii: Yeah... doh.

DPTJKKVH:
@ezaquarii @nextcloud-android-bot  Amazing! I will give it a thorough test tomorrow after work. (Tuesday)

redtux:
Unortunately, it is not working. I have just installed the above build and successfully logged in, but when clicking play on an mp3 file the client immediately crashes. (Been trying several times, so it's reproducible.) 😕

ezaquarii:
> Unortunately, it is not working. I have just installed the above build and successfully logged in, but when clicking play on an mp3 file the client immediately crashes. (Been trying several times, so it's reproducible.) confused

Thanks for testing.

I'll try to ship another build with ACRA crash reporter later this evening, so people actually can get a chance to submit helpful bug reports.

In the meantime, if you can get any logs from logcat (I know few people have development tools installed, so no worries if you can't), it would help to figure it out.

redtux:
I have logcat installed, but my phone is not rooted and the adb command is not working. I'll try your next build and upload the resulting output here. Thank you in advance! 👍

DPTJKKVH:
I will wait for the new build as well. Thanks from me, too! 👍

ezaquarii:
> I have logcat installed, but my phone is not rooted and the adb command is not working. I'll try your next build and upload the resulting output here. Thank you in advance! +1

You don't need your phone to be rooted to use adb.
You need however to enable USB debugging via developer options:

1. Open `Settings`
2. Go into `About phone` page
3. Find the `Build number` entry
4. Tap it quickly few times
5. You might be asked to enter your PIN number

A new `Developer options` should appear in `System` page or somewhere else, depending on your phone model.

Enter `Developer options` and enable `USB debugging`.

I'm ready with ACRA version so you might as well just install it using QR code when Jenkins CI build is complete.

ezaquarii:
@redtux the error you reported should be fixed, you could give it another try. Thanks for the involvment.

redtux:
@ezaquarii Great, it works! 🤩

src/main/java/com/nextcloud/client/media/Player.kt
```diff
@@ -0,0 +1,226 @@
...
+import android.os.PowerManager
+import android.widget.MediaController
+import com.nextcloud.client.media.PlayerStateMachine.Event
+import com.nextcloud.client.media.PlayerStateMachine.State
+import com.owncloud.android.datamodel.OCFile
+import com.owncloud.android.lib.common.OwnCloudAccount
+import com.owncloud.android.lib.common.OwnCloudClient
+import com.owncloud.android.lib.common.OwnCloudClientManagerFactory
+
+/**
```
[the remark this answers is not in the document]
ezaquarii: @tobiasKaminsky @AndyScherzinger I don't know where this is coming from, but this rule effectively makes implementing many interfaces impossible.

redtux:
Good morning everybody! I have just tested build 9938 and everything is working as expected. 👍 @tobiasKaminsky it would be cool to have this included into the next GPlay and F-Droid releases, as not being able to stream media files might be a show stopper for many users. @ezaquarii Thank you again for this rewrite! 🍾 The only thing not working is the "Stream with…" option. When I click it, nothing happens. 😃

DPTJKKVH:
Doesn't work for me. With mp3 the player opens but shows 0 seconds play time and doesn't play the music. With mp4 it shows only a black screen. Player probably doesn't open at all.

redtux:
@DPTJKKVH How long did you wait? How big are those files? (No idea if this matters.) In any case I also have a (logical) delay of some seconds (using smaller files of some MB for testing). A note "Streaming is starting…" or something like this might make sense.

DPTJKKVH:
A 450 kilobytes and the other file was 3 megabytes . Waited more than 30 seconds.

Buffering should be visible indeed. However that was not the issue here.

ezaquarii:
Can you also try ogg vorbis file, just to confirm?

Hard to say what is going on. It can be a problem with OS media player unable to decode funky stream or bug in state machine.

I need to add extra diagnostics. I'll come back to you soon.

ezaquarii:
Yes, it makes perfect sense to confirm your findings on another device. This will rule out or confirm subset of possibilities.

Media player is built into system, so in theory there is only one, but 

1. each phone vendor can do modifications (Samsung is particularly dodgy here)

2. Media player can differ between OS versions

3. Phone can use hardware media decoder for certain formats (less CPU usage, less battery drain)

DPTJKKVH:
But shouldn't the player bundled with the Nextcloud App be always the same?

I'm on Lineage OS 16 btw.

redtux:
Test result on Android 6.0 Alcatel A5 LED: It works! 🎉 The ogg vorbis file starts almost immediately, while the m4a file starts after more than 30 seconds. Several mp3's (~ 5 MB) also took between some seconds and more than 30 seconds to start. The transfer rate is the same. And yes, it shows 00:00 on both sides and there is no indication whatsoever that the player is buffering. But maybe I should try with much bigger files. 🤷‍♂️ (The other phone is a Xiaomi Mi MIX 2, btw.)

ezaquarii:
> But shouldn't the player bundled with the Nextcloud App be always the same?

Nextcloud uses media player API provided by Android operating system. That API goes down below ART into a native platform media player - I guess it's a combination of C/C++/assembly/hardware down there.

Nobody would write complete media player in Java/C/C++ - it would be more complex than entire Nexcloud mobile app.

ezaquarii:
@AndyScherzinger @tobiasKaminsky SpotBugs decreased, but CI says it increased. Any idea why that might happen?

ezaquarii:
@redtux @DPTJKKVH I added some logs (search D/Player tag in logcat) and there should be error message pop-up if streaming fails.

Could you pls give it a try?

redtux:
Sorry, I have been away but have not forgotten about this. :smiley_cat: I am going to test the new build tonight. Is there a way on Android to reduce the transfer limit for this app in order for testing the race condition? Thus, I would have three different devices with different Android versions and different internet connectivities… I will also try different file types with different files sizes. Hopefully, I can give a somehow representative report tomorrow. :+1:

DPTJKKVH:
@ezaquarii I was busy with work as well. Sorry I also did not forget! I will test asap!

ezaquarii:
@redtux I don't think there is an easy way to limit it. You can try switching between Wi-Fi and GSM networks. Usually GSM has longer roundtrips.

ezaquarii:
@tobiasKaminsky @AndyScherzinger can you take a look? We should have some discussion regarding UX before moving forward.

Currently the player is shut down when the preview screen is closed, which is fairly simple.
Notification however does nothing interesting and I'm thinking about getting rid of it for the time being.

This is a preview after all, not a full-blown media player with background playback and playlist handling, so I think that having a notification controls is not needed really at this stage.

redtux:
@ezaquarii Working fine on both phones. Same scenarios as described above (wifi/lte, mp3, mp4, ogg, m4a).

AndyScherzinger:
@ezaquarii didn't have a chance to test it yet while I checked the master and downloaded mp3s do play in the background in case I close (as in minimize, not terminate) the app.

ezaquarii:
> @ezaquarii didn't have a chance to test it yet while I checked the master and downloaded mp3s do play in the background in case I close (as in minimize, not terminate) the app.

Yeah, I guess master is working as expected, ie. it doesn't work in streaming mode. :)
I need some decisions made on playing in background. The UX for this feature was seriously lacking and new implementation doesn't provide background playback, but if we want to have it, I guess we should come up with some sensible UI design for this. My main concern is that the notification doesn't provide any playback controls and since there is no playlist support, this would have little sense anyway.

IMO file preview feature has much lower bar than player should have, so I don't know what we want to do here. I'd rather limit the functionality for the time being as Nextcloud really struggles with supporting existing codebase already.

@tobiasKaminsky let me know what you think. I'm keen on moving forward with this PR.

AndyScherzinger:
I'm also interested in Tobias' pov. Removing the actual existing playback functionality will lead to bad user reviews for sure (as always when taking away previously existing things...) So yes when all need to come to a decision while personally I think it be nice to keep it and rather add a play/pause button to the notification. Just my opinion though.

ezaquarii:
> Removing the actual existing playback functionality will lead to bad user reviews for sure

I woudn't be so sure about it. Streaming didn't work for years and nobody really complained...
I guess people just sync their music and feed it into a proper player. Streaming is probably used only for edge cases.

@DPTJKKVH @redtux can you please share your use cases for this feature, so we can better understand how to proceed with it?

redtux:
Wow, difficult question! 😃 tl;dr? a. Streaming 🤷‍♂️ to 👍 b. Playing in background 👎 (at least as it is implemented atm)… 

In my case I have just clicked on the media file and wondered why play was not working. The player just showed 00:00 as described in the issue, while no media file was found. For me it might have been okay if the app asked me whether I wanted to download the file informing me that otherwise play would not work.

Only when having followed the discussions here I realized that a) this issue was related to streaming and b) the respective file was not yet available offline. At least some note that the file cannot be played because it has not yet been downloaded would have been enough for me, but streaming seems nice, of course. 🤩

Concerning the background play, I consider it beeing quite annoying and not intuitive at all. Not just once did I have the case that I clicked back expecting the media player to stop, and then had no option to stop it. 😂 I agree that some UI desicions need to be taken here.

Regarding streaming, the only unrestricted and ad free player I found on GPlay is CloudBeats, which unfortunately does not seem to be Free Software. (I found no info about its license).

All free players I have found so far (mostly on F-Droid) are not supporting WebDAV, so — as stated above — if streaming is an issue, I would definitely prefer this to playing in background.

(Sorry for this detailed novel, but you asked. 😇)

DPTJKKVH:
@ezaquarii I have a large collection of both music and videos/movies I want to watch and listen to when on business trips or with friends. I also use Nextcloud for work and quite often need to rather spontaneously show someone a video that can be a couple of hundreds of megabytes large.

I can't have all those videos and music stored locally on my smartphone of course so it would be very much appreciated if I could stream such files instead of having to completely download them one by one beforehand.

tobiasKaminsky:
I think we should try to keep the current feature set (also play in background), but on the other hand make sure that when we add something new, it is well tested with tests, etc., so that we do not have that many problems with maintaining of rarely used features.

Streaming should(tm) work, but heavily depends on Android version.

DPTJKKVH:
@tobiasKaminsky The problem is that Nextcloud simply advertises it's supposed streaming capabilities without actually mentioning how much it depends on the Android version used in conjunction. I follow this issue closely and still don't know what libraries (and what versions of those) are required for streaming mp3 audio, mp4 video and so on.

So the damage has already been done and there are multiple issues with users complaining about broken streaming and non functioning thumbnails on the mobile apps. **I wouldn't call it a "new feature" to fix a broken implementation and making it accessible to a broader range of users.**

Also I heavily doubt that the nonfunctional streaming can be tied exclusively to the Android version used. You might be right for the "internal player" that uses the OS provided player but not for external players. There are players like VLC on Android that (to my knowledge) come with their own libraries. The same files that play fine locally with VLC and can be streamed with VLC from others streaming sources cannot be streamed through Nextcloud with the "stream with..." feature.

You can call me a fool if I'm wrong but for me it is pretty obvious that there is an issue with giving external applications access to media files under certain (server) configurations. This is highlighted by the fact that VLC explicitly gives an error that says "Cannot access external resource at [URI]".

Maybe I totally miss your intentions and if so I apologize but for me it seems like you want to put this issue to the side. Instead you should provide more details on what is required to successfully stream both with the internal and with external players so @ezaquarii can improve his fix and I can work out why VLC cannot access files shared by the streaming feature.

Thanks in advance!

ezaquarii:
@DPTJKKVH I believe there is confusion between what is the actual problem with the current player and what is @tobiasKaminsky referring to.

We can't of course guarantee what the android media player does nor we'll be capable providing any workarounds for platform bugs, should there be any. I believe that the built-in media player is pretty reliable as many dedicated players depend on it, but you never know with 3rd party, hence the small print.

The current issues we're experiencing with streaming are not caused by android media player itself, but by incorrect usage of the offcial API - and this is certainly in our capacity to fix. I'm confident that this will improve the situation considerably.

VLC however is a different beast. It's a media player with entire company behind it dedicated to provide robust, multi-platform solution for media decoding. The level of media support VLC provides is - I'm affraid - unique for the project. The only known player that reached (suprassed?) VLS is MPlayer (both share a great deal of code, btw).

DPTJKKVH:
@ezaquarii Thanks for the clarification. I fully agree with you on the API issue. This is in line with what I observed on my end. I also think that fixing the API issue should be prioritized over any player related changes since it probably will fix streaming in most cases.

ezaquarii:
@tobiasKaminsky @AndyScherzinger The player is in mergable state.

Please note that I removed some unused transaltions due to simplified error handling. I decided to keep this removal as a separate commit for quick revert if you'd like to keep it.

@DPTJKKVH @redtux Please test the latest build using build from QR code.

ezaquarii:
New player implementation is based on a state machine.

PlayerStateMachine

ezaquarii:
@tobiasKaminsky @AndyScherzinger The player is in mergable state.

Please note that I removed some unused transaltions due to simplified error handling. I decided to keep this removal as a separate commit for quick revert if you'd like to keep it.

@DPTJKKVH @redtux Please test the latest build using build from QR code.

redtux:
@ezaquarii Works for me! Playing already downloaded files starts without delay (as expected), while streaming still takes a while to start. For video (mp4) I get a "Loading..." message like when opening other files like pictures, which would also be nice for audio files, wouldn't it?

(I have just tried with the default Nextcloud.mp4 from a fresh install.)

Playing in background as expected is _not_ working. (I thought that leaving the NC app media player open while switching to another app might work, but it does not, as this closes the player - returning to the file browser.)

One question remains: How can I know that streaming is _really_ working? Given the delay it feels like the whole file is being downloaded, but I have no idea how to check this. Anyway, the `nextcloud/tmp` folder stays empty (in case this is related).

ezaquarii:
I'm glad it worked for you.

> while streaming still takes a while to start.

This is expected. When we land #4275, we'll be able to measure those delays, so stay tuned.

> For video (mp4) I get a "Loading..." message

Video playback uses different player, which is not a scope of this PR.

> like when opening other files like pictures, which would also be nice for audio files, wouldn't it?

Entire UI should be reworked completely, as the current architecture has issues that are not fixable economically. Once that is done, we can add playing in background easily.

> One question remains: How can I know that streaming is _really_ working? Given the delay it feels like the whole file is being downloaded

Try playing (very) big file, preferably over a slow connection. You can also monitor your network traffic during playback either via router metrics, some applet or VPN. If connection is fast (WiFi, 10s of Mbit/s) and file is small (few MB), the media player can potentially buffer it whole - we have no control over it.

ezaquarii:
@tobiasKaminsky  ping

ezaquarii:
@tobiasKaminsky @AndyScherzinger Can we have it in 3.8?

AndyScherzinger:
I'd be fine with that knowing it will remove the possibility to play the file in the background. While playback functionality should then be brought back right after 3.8.0

ezaquarii:
Background playback is not working for now due to lack of UI to stop it.

tobiasKaminsky:
Sorry, we should not add in a major release something that has limited functions than before. 
Let us focus on it directly after 3.8., so that we can get it 3.9.
(I know it is most because of me, as I did not found any time yet to review it. Sorry for that! Once you implemented the missing part, I will directly review it!)

ezaquarii:
@tobiasKaminsky Background playback works more or less like it did before.

Please be aware that the previous player behavior was result of a bug - service player is not unbound in `FileDisplayActivity`, retaining player state until the app closed. I did not try to reproduce this bug 1:1. :-)

The UI for stopping is still not there (it wasn't on master - stopping experience is bad), so I hope we can start discussing it now and patch it here or as a follow up.

It would be nice if you could review it before end of Friday, as I have some bandwidth available saturday to continue working on any remarks.

ezaquarii:
@tobiasKaminsky @AndyScherzinger lint complains about unused strings, I'll clean it later.

ezaquarii:
I managed to finalize `stateless4j` transition to my github org, library cleanup and pushed new release to Maven Central. I updated our gradle to use new maintained lib.

redtux:
I have just been testing the two most recent releases, and both 10925 & 10926 still work fine for me — with backgroud play. 👍 .aac is not working (trying to download the file instead of opening the mediaplayer for streaming) but this might be another issue (as it also does not work with the current release 3.8.0).

ezaquarii:
@tobiasKaminsky @AndyScherzinger Can we start reviewing PRs? PRs from various authors are piling up. There is generally little reviewing activity for last 2 months.

AndyScherzinger:
> Can we start reviewing PRs? PRs from various authors are piling up. There is generally little reviewing activity for last 2 months.

Sure thing, I'll take a look at them (at least some, since work/private life are quite busy atm).

src/main/java/com/nextcloud/client/media/LoadUrlTask.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.owncloud.android.lib.common.OwnCloudClient
+
+internal class LoadUrlTask(
+    private val client: OwnCloudClient,
+    private val fileId: String,
+    private val onResult: (String?) -> Unit
+) : AsyncTask<Void, Void, String>() {
+
+    override fun doInBackground(vararg args: Void): String? {
+        val sfo = StreamMediaFileOperation(fileId)
```
tobiasKaminsky: Normally we do not use abbreviations, as they are harder to read.

src/main/java/com/nextcloud/client/media/LoadUrlTask.kt
```diff
@@ -0,0 +1,49 @@
...
+        val result = sfo.execute(client)
```
tobiasKaminsky:
```suggestion
        val result = StreamMediaFileOperation(fileId).execute(client)
```

This way it is shorter, and no additional variable is needed

src/main/java/com/nextcloud/client/media/Player.kt
```diff
@@ -0,0 +1,312 @@
...
+    private val mediaPlayerCreator: () -> MediaPlayer = { MediaPlayer() }
+) : MediaController.MediaPlayerControl {
+
+    private companion object {
+        const val DEFAULT_VOLUME = 1.0f
+        const val DUCK_VOLUME = 0.1f
+        const val MIN_DURATION_ALLOWING_SEEK = 3000
+    }
+
+    data class Error(val message: String)
```
tobiasKaminsky: Not sure about this and Listener, but having different classes inside another file is in my opinion always a bit tricky and can lead to misunderstandings.

src/main/java/com/nextcloud/client/media/Player.kt
```diff
@@ -0,0 +1,312 @@
...
+
+    interface Listener {
+        fun onRunning(file: OCFile)
+        fun onStart()
+        fun onPause()
+        fun onStop()
+        fun onError(error: Error)
+    }
+
+    private var fsm: PlayerStateMachine
```
tobiasKaminsky: fsm? Maybe there is a better name for it?

src/main/java/com/nextcloud/client/media/Player.kt
```diff
@@ -0,0 +1,312 @@
...
+    private val focusManager = AudioFocusManager(audioManager, this::onAudioFocusChange)
+
+    private val delegate = object : PlayerStateMachine.Delegate {
+        override val isDownloaded: Boolean get() = playedFile?.isDown ?: false
+        override val isAutoplayEnabled: Boolean get() = autoPlay
+        override val hasEnqueuedFile: Boolean get() = enquedFile != null
+
+        override fun onStartRunning() {
+            trace("onStartRunning()")
+            enquedFile.let {
```
tobiasKaminsky: Just for my understanding, why do you use let here?
The function block could also be written directly via if (enquedFile != null) { … } else { … }?

src/main/java/com/nextcloud/client/media/Player.kt
```diff
@@ -0,0 +1,312 @@
...
+
+    private fun onAudioFocusChange(focus: AudioFocus) {
+        when (focus) {
+            AudioFocus.FOCUS -> fsm.post(Event.FOCUS_GAIN)
+            AudioFocus.DUCK -> fsm.post(Event.FOCUS_DUCK)
+            AudioFocus.LOST -> fsm.post(Event.FOCUS_LOST)
+        }
+    }
+
+    // this should be refactored into a proper, injectable factory
```
tobiasKaminsky: I am currently working on replacing our network library, there we can also tackle this :+1:

src/main/java/com/nextcloud/client/media/PlayerError.kt
```diff
@@ -0,0 +1,94 @@
...
+object PlayerError {
+
+    /** Error code for specific messages - see regular error codes at [MediaPlayer]  */
+    const val OC_MEDIA_ERROR = 0
+
+    @JvmStatic
+    fun toString(context: Context?, what: Int, extra: Int): String {
+        val messageId: Int
+
+        if (what == OC_MEDIA_ERROR) {
```
tobiasKaminsky: Maybe a switch is "nicer" / more readable here?

src/main/java/com/nextcloud/client/media/PlayerStateMachine.kt
```diff
@@ -0,0 +1,229 @@
...
+import com.github.oxo42.stateless4j.StateMachineConfig
+import com.github.oxo42.stateless4j.delegates.Action
+import com.github.oxo42.stateless4j.transitions.Transition
+import java.util.ArrayDeque
+
+/*
+ * To see visual representation of the state machine, install PlanUml plugin.
+ * http://plantuml.com/
+ *
+ * @startuml
```
tobiasKaminsky: Very nice! :tada: :tada: :tada: 

image

src/main/java/com/nextcloud/client/media/PlayerStateMachine.kt
```diff
@@ -0,0 +1,229 @@
...
+        ERROR,
+        FOCUS_LOST,
+        FOCUS_GAIN,
+        FOCUS_DUCK,
+        IMMEDIATE_TRANSITION,
+    }
+
+    private var pendingEvents = ArrayDeque<Event>()
+    private var isProcessing = false
+    private val fsm: StateMachine<State, Event>
```
tobiasKaminsky: fsm? Maybe just stateMachine?

src/main/java/com/owncloud/android/media/MediaControlView.java
```diff
@@ -259,6 +264,21 @@ public void updatePausePlay() {
         } else {
             mPauseButton.setImageResource(android.R.drawable.ic_media_play);
         }
+
+        final boolean canSeekFfd = mPlayer.canSeekForward();
+        if (canSeekFfd) {
+            mFfwdButton.setVisibility(View.VISIBLE);
```
tobiasKaminsky: While we are already touching this class, maybe we want to get rid of "m" prefix? And fwd -> forward, bwd -> backward
I can do this if you want.

ezaquarii:
Finite State Machine aka FSM is a well recognized abbreviation used in CS literature.

https://en.m.wikipedia.org/wiki/Finite-state_machine

I get that we don't do abbreviations to be more friendly to contributors so as this is our cultural standard I'll change it to stateMachine. Np.

tobiasKaminsky:
Bug(?):
- start an audio
- suspend app
- restart app
- click on pause
- click on back
- still see "is playing" notification

Old behaviour was that when pausing (as we do not have a stop) and leaving player the notification is gone away.

tobiasKaminsky:
Except this, it looks very very good and is working fine :tada: :tada: :tada: :tada: 
I guess this is a huge enhancement.
Thanks again for your wonderful work!

ezaquarii:
> Bug(?):

Weeeeell, yes and no, aka *it's complicated*.

Previous implementation simply turned off service foreground mode, but did not care about clearing the player state nor even releasing the binders.

This had a side effect of clearing the icon, but the player was left in paused state. It was usually also bound to `FileListActivity` (and derivatives) as well, preventing the service from being destroyed by the OS most of the time.

Relying on a background service to retain it's state guarantees to have funky edge issues caused by Android resources reclamation.

Current implementation preserves the backgrounding behaviour of old player, but does not disable foreground mode, which manifests itself with the icon being still on screen (because the player is technically still up and running, just paused). Otherwise it works as the old one MINUS possible breakages on service kill.

I added a STOP button to the notification, so you can stop the player if needed.

I was hoping that we simply add proper media player controls to this notification, like every other media player does, allowing the user to go back, pause, resume and stop from there.

If the old behavior is important to retain, we could:
1. bind the service to `MainApp` to protect it against Grim Reaper in paused-and-not-foreground state
2. turn toggle foreground mode when entering/exiting paused state

This won't play nice if we want to have media controls in notification, as pausing it would close the media controls (and then you can resume it again of course).

We have some complaints about the current pause/stop experience, and I think this "magical" pausing is not a good UX at all.

Let me know what we want to do here.

src/main/java/com/nextcloud/client/media/PlayerError.kt
ezaquarii: This is previous player code that I copy-pasted verbatim to preserve error messages UX.

I can modernise it using Kotlin `when`.

src/main/java/com/nextcloud/client/media/Player.kt
ezaquarii: That would be nice.

src/main/java/com/nextcloud/client/media/Player.kt
ezaquarii: Prevents the `enquedFile` to change beetween `if` expression evaluation and body. Linter complains about it. It is a syntax sugar for:

```
var enqueuedFile: OCFile? = ...
val immutableEnqueuedFile = enqueuedFile
if (immutableEnqueuedFile != null) {
    ...
    enqueuedFile could be null by now, but immutableEnqueuedFile is not
    ...
}
```

src/main/java/com/nextcloud/client/media/LoadUrlTask.kt
ezaquarii: Fixed

src/main/java/com/nextcloud/client/media/LoadUrlTask.kt
ezaquarii: I'd leave it separate as there should be 1 statement per line.

Not that it matter in this case...

src/main/java/com/nextcloud/client/media/Player.kt
ezaquarii: Listeners are generally quiet often declared as nested classes to indicate strong relationship with the enclosing class.

Also, you want to name the listener `Listener` - plain and simple.

src/main/java/com/nextcloud/client/media/Player.kt
ezaquarii: renamed `fsm` to `stateMachine`

src/main/java/com/owncloud/android/media/MediaControlView.java
ezaquarii: renamed members

src/main/java/com/nextcloud/client/media/PlayerStateMachine.kt
ezaquarii: renamed to `stateMachine`

src/main/java/com/nextcloud/client/media/PlayerError.kt
ezaquarii: It uses 2 values, so can't do. I'll leave `if-else`. This is marked as deprecated and generally should go at some point.

tobiasKaminsky:
> I added a STOP button to the notification, so you can stop the player if needed.

Nice, I think this is enough. 



> I was hoping that we simply add proper media player controls to this notification, like every other media player does, allowing the user to go back, pause, resume and stop from there.

This would indeed be very cool, but should be done in another PR :+1:

tobiasKaminsky:
Thanks again for this great PR :tada: :tada: :tada:

Issue #4228, opened 2019-07-13
Log_OC doesn't collect logs in file


ezaquarii:
### Actual behaviour
Logs don't show up in `Settings`->`Logs`

### Expected behaviour
Logs should be visible there

### Steps to reproduce
1. Launch Nextcloud app
2. Go to `Settings`->`Logs`
3. Logs are outdated

### Environment data
Android version: 9.0
Device model: One Plus 6T
Stock or customized system: Stock
Nextcloud app version: master
Nextcloud server version: 16.0.2

#### Nextcloud log (data/nextcloud.log)
```
This is exactly the stuff that is missing!
```

ezaquarii:
Logging must be enabled:
```
synchronized private static void appendLog(String text) {
    if (isEnabled) {
        ...
```

However, the logger is enabled in a somewhat peculiar way:

```
synchronized public static void startLogging(Context context) {
    ...
    try {
        // Create the current log file if does not exist
        isEnabled = mLogFile.createNewFile();
```

`File.createNewFile` returns `false` if the file already exists, which means that logging works only for the very first ime. On all subsequent app runs logging is disabled.

ezaquarii:
To solve this problem, there are few things that are piling up:

1. ETM PR for injectable view models: https://github.com/nextcloud/android/pull/4133
2. Pluggable logger implementation in Library: https://github.com/nextcloud/android-library/pull/324
3. New logger - https://github.com/nextcloud/android/pull/4275

tobiasKaminsky:
Thanks for the nice summary and bug finding/fixing!
I'll have a look at the PRs now.

PR #2250, opened 2018-02-28
Use reloading on photo view

tobiasKaminsky:
Fix #4235 
Fix #1972 
Fix #1682 
Fix #1901 

*  [x] Needs: https://github.com/nextcloud/android-library/pull/134

Once this is merged, revert: https://github.com/nextcloud/android-library/pull/322

~~Stalled by:~~
~~https://github.com/nextcloud/server/issues/15029~~
~~https://github.com/nextcloud/server/issues/15048~~

This first loads 3x display items and on each bottom reach loads the next bunch of images.
Due to limitation of webdav this is currently only working if search is sorted by date, so not working on file names (for (favorite) search).

**BUGS:**
- [x] no loading dialog on start
- [x] swipe to refresh on photo search results in "/" file list
- [x] loading additional photos duplicates items, so after 3x we have 3x the first image
- [x] sort order is totally arbitrary
- [x] click on an image, return, sometimes show file list

**TODOs**
- [x] "cache" search: show photos, click on image, go back within 30s should show directly "old" search
- [x] try to re-use stored information, e.g. of parent folder
- [x] add check for NC14
- [x] check/fix navigation
- [x] maybe enhance loading with http://www.devexchanges.info/2017/02/android-recyclerview-dynamically-load.html

Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

PR #4233, opened 2019-07-15
Re-add content search type

tobiasKaminsky:
Depends on https://github.com/nextcloud/android-library/pull/319

:exclamation: Stalled by https://github.com/nextcloud/android-library/pull/319 :exclamation: 
Should be based upon photoSearchLimit branch.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -471,6 +471,10 @@ public abstract class DrawerActivity extends ToolbarActivity
                                                    SearchEvent.UnsetType.UNSET_BOTTOM_NAV_BAR),
                                    menuItem.getItemId());
                 break;
+            case R.id.nav_recently_added:
+                handleSearchEvents(new SearchEvent("%", SearchRemoteOperation.SearchType.CONTENT_TYPE_SEARCH,
+                                                   SearchEvent.UnsetType.UNSET_BOTTOM_NAV_BAR), menuItem.getItemId());
+                break;
             case R.id.nav_recently_modified:
                 handleSearchEvents(new SearchEvent("",
                                                    SearchRemoteOperation.SearchType.RECENTLY_MODIFIED_SEARCH,
--- a/src/main/java/com/nextcloud/client/di/ComponentsModule.java
+++ b/src/main/java/com/nextcloud/client/di/ComponentsModule.java
@@ -105,9 +105,7 @@ abstract class ComponentsModule {
     @ContributesAndroidInjector abstract ManageAccountsActivity manageAccountsActivity();
     @ContributesAndroidInjector abstract ManageSpaceActivity manageSpaceActivity();
     @ContributesAndroidInjector abstract NotificationsActivity notificationsActivity();
-
-    @ContributesAndroidInjector
-    abstract CommunityActivity participateActivity();
+    @ContributesAndroidInjector abstract CommunityActivity participateActivity();
     @ContributesAndroidInjector abstract PassCodeActivity passCodeActivity();
     @ContributesAndroidInjector abstract PreviewImageActivity previewImageActivity();
     @ContributesAndroidInjector abstract PreviewVideoActivity previewVideoActivity();
--- a/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
+++ b/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
@@ -377,6 +377,13 @@ public class OCFileListFragment extends ExtendedListFragment implements
                     currentSearchType = SearchType.FILE_SEARCH;
                     break;
 
+                case CONTENT_TYPE_SEARCH:
+                    if ("video/%".equals(event.getSearchQuery())) {
+                        currentSearchType = SearchType.VIDEO_SEARCH;
+                    }
+                    break;
+
+
                 case FAVORITE_SEARCH:
                     currentSearchType = SearchType.FAVORITE_SEARCH;
                     break;


src/main/java/com/owncloud/android/ui/EmptyRecyclerView.java
```diff
@@ -34,7 +34,7 @@
  */
 public class EmptyRecyclerView extends RecyclerView {
     private View mEmptyView;
-    private boolean hasFooter;
+    private boolean hasFooter = false;
```
ezaquarii: Since `false` is a default value, it's redundant. Since being explicit is better than implicit, I actually find this a good move, but `Codacy` doesn't. 

It would be nice to disable codacy check if possible.

src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -451,8 +451,16 @@ private void selectNavigationItem(final MenuItem menuItem) {
                         SearchEvent.UnsetType.NO_UNSET), menuItem.getItemId());
                 break;
             case R.id.nav_photos:
-                handleSearchEvents(new SearchEvent("image/%", SearchRemoteOperation.SearchType.CONTENT_TYPE_SEARCH,
-                        SearchEvent.UnsetType.NO_UNSET), menuItem.getItemId());
+                SearchEvent searchEvent = new SearchEvent("image/%",
```
ezaquarii: Can we move this into a method?

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -2576,6 +2593,25 @@ public void showFiles(boolean onDeviceOnly) {
...
+            fragment = new PhotoFragment();
+        } else {
+            Log_OC.d(this, "Switch to OCFileListFragment");
+
+            fragment = new OCFileListFragment();
+        }
+
+        FragmentTransaction transaction = getSupportFragmentManager().beginTransaction();
+        transaction.add(R.id.left_fragment_container, fragment, TAG_LIST_OF_FILES);
+        transaction.commit();
```
ezaquarii: This is UI code called on background thread. Although `commit()` will schedule transaction on main thread, that aspect of `FragmentManager` is not documented very well and I'd expect no guarantees.

It's safer to keep all UI ops on main thread, IMO.

src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -801,13 +807,16 @@ private void parseVirtuals(List<Object> objects, ExtendedListFragment.SearchType
             case PHOTO_SEARCH:
                 type = VirtualFolderType.PHOTOS;
                 onlyImages = true;
+                lastTimestamp = ((RemoteFile) objects.get(objects.size() - 1)).getModifiedTimestamp() / 1000;
```
ezaquarii: what if `objects.size()` is `0`?

src/main/java/com/owncloud/android/ui/asynctasks/PhotoSearchTask.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    protected Boolean doInBackground(Void... voids) {
+        if (photoFragmentWeakReference.get() == null) {
+            return false;
+        }
+        PhotoFragment photoFragment = photoFragmentWeakReference.get();
+        OCFileListAdapter adapter = photoFragment.getAdapter();
+
+        photoFragment.setPhotoSearchQueryRunning(true);
```
ezaquarii: Setting UI in background thread is asking for trouble.

src/main/java/com/owncloud/android/ui/asynctasks/PhotoSearchTask.java
```diff
@@ -0,0 +1,123 @@
...
+
+        if (!isCancelled()) {
+            int limit = -1;
+
+            limit = 15 * columnCount;
+
+            long timestamp = -1;
+            if (adapter.getLastTimestamp() > 0) {
```
ezaquarii: I doubt `adapter` is thread-safe. Expect weird, random issues and glitches here.

src/main/java/com/owncloud/android/ui/asynctasks/PhotoSearchTask.java
```diff
@@ -0,0 +1,123 @@
...
+                                                                                        photoFragment.requireContext());
+
+            if (remoteOperationResult.isSuccess() && remoteOperationResult.getData() != null
+                && !isCancelled()) {
+                if (remoteOperationResult.getData() == null || remoteOperationResult.getData().size() == 0) {
+                    photoFragment.setPhotoSearchNoNew(true);
+                    // todo how distinguish to show
+                    // setEmptyView(event);
+                } else {
+                    adapter.setData(remoteOperationResult.getData(),
```
ezaquarii: Setting UI data on background thread.

src/main/java/com/owncloud/android/ui/asynctasks/PhotoSearchTask.java
```diff
@@ -0,0 +1,123 @@
...
+                                    ExtendedListFragment.SearchType.PHOTO_SEARCH,
+                                    storageManager,
+                                    null,
+                                    false);
+                    Log_OC.d(this, "Search: count: " + remoteOperationResult.getData().size());
```
ezaquarii: BTW, see #4228.

src/main/java/com/owncloud/android/ui/asynctasks/PhotoSearchTask.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @Override
+    protected void onPostExecute(Boolean result) {
+        if (photoFragmentWeakReference.get() != null) {
+            PhotoFragment photoFragment = photoFragmentWeakReference.get();
+
+            final ToolbarActivity fileDisplayActivity = (ToolbarActivity) photoFragment.getActivity();
+            if (fileDisplayActivity != null) {
+                fileDisplayActivity.runOnUiThread(() -> fileDisplayActivity.setIndeterminate(false));
```
ezaquarii: `onPostExecute` is actually running on a main thread, so this `runOnUiThread` seems to be reduntant.

src/main/java/com/owncloud/android/ui/fragment/PhotoFragment.java
```diff
@@ -0,0 +1,187 @@
...
+
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
```
ezaquarii: I think it needs `@Override`

src/main/java/com/owncloud/android/ui/fragment/PhotoFragment.java
```diff
@@ -0,0 +1,187 @@
...
+        View v = super.onCreateView(inflater, container, savedInstanceState);
+
+        onScrollChangeListener = new RecyclerView.OnScrollListener() {
```
ezaquarii: I'd avoid doing inline class here. It seems that considerable amount of UI logic is part of `onCreateView`. It's generally ok for trivial lambda's, but this is serious piece of logic.

src/main/java/com/owncloud/android/ui/fragment/PhotoFragment.java
```diff
@@ -0,0 +1,187 @@
...
+                }
+            }
+        };
+
+        Log_OC.i(this, "onCreateView() in PhotoFragment end");
+        return v;
+    }
+
+    @Override
+    public void onItemClicked(OCFile file) {
```
ezaquarii: Redundant override.

src/main/java/com/owncloud/android/ui/fragment/PhotoFragment.java
```diff
@@ -0,0 +1,187 @@
...
+    @Override
+    public void onRefresh(boolean ignoreETag) {
+        super.onRefresh(ignoreETag);
+
+        // TODO: refresh not working
+
+        onMessageEvent(searchEvent, true);
+    }
+
+    public void onMessageEvent(final SearchEvent event, boolean clear) {
```
ezaquarii: Where is this event coming from?

Actually, there is considerable amount of UI logic that sould be named somehow. This function would benefit from a better name.

Or JavaDoc.

src/main/java/com/owncloud/android/ui/fragment/PhotoFragment.java
```diff
@@ -0,0 +1,187 @@
...
+                                mContainerActivity.getStorageManager())
+                .execute();
+        }
+    }
+
+    public void setPhotoSearchQueryRunning(boolean bool) {
+        photoSearchQueryRunning = bool;
+    }
+
+    public void setPhotoSearchNoNew(boolean bool) {
```
ezaquarii: This concept of `no new` is not obvious. JavaDoc would help here.

src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
ezaquarii: what if `objects` is not `RemoteFile` (`ClassCastException`)

ezaquarii:
This is completely new fragment and I believe we could greatly benefit from using `ViewModel` here. View models are gratly improving testability.

src/main/java/com/owncloud/android/ui/EmptyRecyclerView.java
AndyScherzinger: I am fine either way. Any opinion @tobiasKaminsky ?

tobiasKaminsky: I also think it is better to understand…
Either we can disable this on Codacy, or we just ignore it.

tobiasKaminsky:
@ezaquarii I messed something up. I wanted to split up #2250 so that #2250 does not contain any "content search type".
Now it seems that both PRs have the same changes to master and so your comments need to be in #2250. 
I'll check them and change them in #2250 and after that try to rebase this PR to be ontop of #2250, also changing the base branch here in github.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
tobiasKaminsky: As we have this on several places in FileDisplayActivity, I would like to change this in a new PR, so it is easier to test.

src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
tobiasKaminsky: True, I'll change it. (it is already checked in line 764, but better to be safe.

objects is result from RemoteOperation and it is always a List<Object>. 
At some point we should make this typesafe, but for now we are using the same approach in many many other places.

src/main/java/com/owncloud/android/ui/asynctasks/PhotoSearchTask.java
tobiasKaminsky: it is in onPreExecute, which says:
Runs on the UI thread before doInBackground.

src/main/java/com/owncloud/android/ui/asynctasks/PhotoSearchTask.java
tobiasKaminsky: It is a direct (although weak) reference, so there can be no other Activity changing this, or?

src/main/java/com/owncloud/android/ui/asynctasks/PhotoSearchTask.java
tobiasKaminsky: True, this needs to be moved to onPostExecute, which is run in UI thread.

src/main/java/com/owncloud/android/ui/fragment/PhotoFragment.java
tobiasKaminsky: :+1:

src/main/java/com/owncloud/android/ui/fragment/PhotoFragment.java
tobiasKaminsky: Shall I create a new Class for this then?

src/main/java/com/owncloud/android/ui/fragment/PhotoFragment.java
tobiasKaminsky: onMessagEvent is for EventBus, but in this case I c&p the name.
I changed it.

tobiasKaminsky:
> This is completely new fragment and I believe we could greatly benefit from using `ViewModel` here. View models greatly improve testability.

While this is a new fragment, it is "only" a subset of OCFileListFragemnt, which is a subclass of ExtendedListFragment.

Indeed, we should change this, but please not in this PR.

tobiasKaminsky:
All further discussion should go in #2250 as this PR will be based on it.

PR #4245, opened 2019-07-20
Fix upload list comparator to make it respect the comparator contract

fogninid:
the current comparator implementation is correct only if there is only one "isUpdating" file in the array.
This can happen for example as a result of #4244, or if concurrent uploads will be impelmented.

Change the implementation to make it follow the simpler Comparator.thenComparing pattern

### DIFF
--- a/src/main/java/com/owncloud/android/db/OCUploadComparator.kt
+++ b/src/main/java/com/owncloud/android/db/OCUploadComparator.kt
@@ -1,0 +1,75 @@
+/*
+ * NextCloud Android client application
+ *
+ * @copyright Copyright (C) 2019 Daniele Fognini <dfogni@gmail.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.owncloud.android.db
+
+/**
+ *  Sorts OCUpload by (uploadStatus, uploadingNow, uploadEndTimeStamp, uploadId).
+ */
+class OCUploadComparator : Comparator<OCUpload?> {
+    @Suppress("ReturnCount")
+    override fun compare(upload1: OCUpload?, upload2: OCUpload?): Int {
+        if (upload1 == null && upload2 == null) {
+            return 0
+        }
+        if (upload1 == null) {
+            return -1
+        }
+        if (upload2 == null) {
+            return 1
+        }
+
+        val compareUploadStatus = compareUploadStatus(upload1, upload2)
+        if (compareUploadStatus != 0) {
+            return compareUploadStatus
+        }
+
+        val compareUploadingNow = compareUploadingNow(upload1, upload2)
+        if (compareUploadingNow != 0) {
+            return compareUploadingNow
+        }
+
+        val compareUpdateTime = compareUpdateTime(upload1, upload2)
+        if (compareUpdateTime != 0) {
+            return compareUpdateTime
+        }
+
+        val compareUploadId = compareUploadId(upload1, upload2)
+        if (compareUploadId != 0) {
+            return compareUploadId
+        }
+
+        return 0
+    }
+
+    private fun compareUploadStatus(upload1: OCUpload, upload2: OCUpload): Int {
+        return upload1.fixedUploadStatus.compareTo(upload2.fixedUploadStatus)
+    }
+
+    private fun compareUploadingNow(upload1: OCUpload, upload2: OCUpload): Int {
+        return upload2.isFixedUploadingNow.compareTo(upload1.isFixedUploadingNow)
+    }
+
+    private fun compareUpdateTime(upload1: OCUpload, upload2: OCUpload): Int {
+        return upload2.fixedUploadEndTimeStamp.compareTo(upload1.fixedUploadEndTimeStamp)
+    }
+
+    private fun compareUploadId(upload1: OCUpload, upload2: OCUpload): Int {
+        return upload1.fixedUploadId.compareTo(upload2.fixedUploadId)
+    }
+}
--- a/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
+++ b/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
@@ -49,6 +49,7 @@ import com.owncloud.android.datamodel.ThumbnailsCacheManager;
 import com.owncloud.android.datamodel.UploadsStorageManager;
 import com.owncloud.android.datamodel.UploadsStorageManager.UploadStatus;
 import com.owncloud.android.db.OCUpload;
+import com.owncloud.android.db.OCUploadComparator;
 import com.owncloud.android.db.UploadResult;
 import com.owncloud.android.files.services.FileUploader;
 import com.owncloud.android.lib.common.utils.Log_OC;
@@ -59,7 +60,6 @@ import com.owncloud.android.utils.ThemeUtils;
 
 import java.io.File;
 import java.util.Arrays;
-import java.util.Comparator;
 
 import androidx.annotation.NonNull;
 import butterknife.BindView;
@@ -714,7 +714,7 @@ public class UploadListAdapter extends SectionedRecyclerViewAdapter<SectionedVie
             for (OCUpload upload : array) {
                 upload.setDataFixed(binder);
             }
-            Arrays.sort(array, comparator);
+            Arrays.sort(array, new OCUploadComparator());
 
             setItems(array);
         }
@@ -722,46 +722,5 @@ public class UploadListAdapter extends SectionedRecyclerViewAdapter<SectionedVie
         private int getGroupItemCount() {
             return items == null ? 0 : items.length;
         }
-
-        Comparator<OCUpload> comparator = new Comparator<OCUpload>() {
-            @Override
-            public int compare(OCUpload upload1, OCUpload upload2) {
-                if (upload1 == null && upload2 == null) {
-                    return 0;
-                }
-                if (upload1 == null) {
-                    return -1;
-                }
-                if (upload2 == null) {
-                    return 1;
-                }
-                if (UploadStatus.UPLOAD_IN_PROGRESS == upload1.getFixedUploadStatus()) {
-                    if (UploadStatus.UPLOAD_IN_PROGRESS != upload2.getFixedUploadStatus()) {
-                        return -1;
-                    }
-                    // both are in progress
-                    if (upload1.isFixedUploadingNow()) {
-                        return -1;
-                    } else if (upload2.isFixedUploadingNow()) {
-                        return 1;
-                    }
-                } else if (upload2.getFixedUploadStatus() == UploadStatus.UPLOAD_IN_PROGRESS) {
-                    return 1;
-                }
-                if (upload1.getFixedUploadEndTimeStamp() == 0 || upload2.getFixedUploadEndTimeStamp() == 0) {
-                    return compareUploadId(upload1, upload2);
-                } else {
-                    return compareUpdateTime(upload1, upload2);
-                }
-            }
-
-            private int compareUploadId(OCUpload upload1, OCUpload upload2) {
-                return Long.compare(upload1.getFixedUploadId(), upload2.getFixedUploadId());
-            }
-
-            private int compareUpdateTime(OCUpload upload1, OCUpload upload2) {
-                return Long.compare(upload2.getFixedUploadEndTimeStamp(), upload1.getFixedUploadEndTimeStamp());
-            }
-        };
     }
 }
--- a/src/main/java/com/owncloud/android/db/OCUpload.java
+++ b/src/main/java/com/owncloud/android/db/OCUpload.java
@@ -124,10 +124,10 @@ public class OCUpload implements Parcelable {
     /**
      * temporary values, used for sorting
      */
-    @Getter private UploadStatus fixedUploadStatus;
-    @Getter private boolean fixedUploadingNow;
-    @Getter private long fixedUploadEndTimeStamp;
-    @Getter private long fixedUploadId;
+    private UploadStatus fixedUploadStatus;
+    private boolean fixedUploadingNow;
+    private long fixedUploadEndTimeStamp;
+    private long fixedUploadId;
 
     /**
      * Main constructor.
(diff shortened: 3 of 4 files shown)


src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
```diff
@@ -681,6 +681,59 @@ private void openFileWithDefault(String localPath) {
         CURRENT, FINISHED, FAILED
     }
 
+    static Comparator<OCUpload> comparator = new Comparator<OCUpload>() {
```
ezaquarii: This can be made a normal class, not a nested member. There are 2 benefits of this:

- resulting code would be less complex due ot comparator name stating what it actually does (I have no idea what it does, TBH)
- `comparator` field is an implementation detail that we can hide from other package members

Also, a JavaDoc woudn't hurt, as it's pretty much impossible to understand what this code really does.

src/test/java/com/owncloud/android/ui/adapter/UploadListAdapterTest.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Comparator;
+
+import static org.mockito.Mockito.mock;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.when;
+
+public class UploadListAdapterTest {
+    @Test
+    public void testUploadListComparator() {
```
ezaquarii: There are 3 comparator cases:

- greater than
- equal
- less than

I don't see this behaviour tested here. It could be indirectly, through array sorting, but reading the code I have no confidence that we're covering all of them.

I'm sure we can improve this.

src/test/java/com/owncloud/android/ui/adapter/UploadListAdapterTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.Arrays;
+import java.util.Comparator;
+
+import static org.mockito.Mockito.mock;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.when;
+
+public class UploadListAdapterTest {
```
ezaquarii: This test does not cover list adapter functionality - it's a comparator test only. The name should reflect this fact, as it confuses other people reading it.

src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
fogninid: The implementation is really just easier than a verbose javadoc, I wrote the java8 translation into it, which is much easier than trying to describe what it does.

The comparator is tailored to the uploadlist view, there is no reason to reuse it somewhere else, it is package-private only for testing, added an annotation.

src/test/java/com/owncloud/android/ui/adapter/UploadListAdapterTest.java
fogninid: > While being here, I'm sure we can apply Scout's Rule `#1` - leave the place cleaner than we found it. ;)
> 
> Also, if possible, could we write the test in Kotlin? The goal is to accelerate and encourage it's adoption, as Kotlin is de facto primary development language for Android now.

rewrote it in Kotlin with generic tests of the comparator contract

src/test/java/com/owncloud/android/ui/adapter/UploadListAdapterTest.java
fogninid: Well, it tests how uploads are sorted in the uploadview (and that it does not crash as it does now).

It just happens to be the only tested functionality

src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
```diff
@@ -681,6 +682,72 @@ private void openFileWithDefault(String localPath) {
...
+     *
+     *  Comparator.nullsFirst(
+     *      Comparator.comparing(OCUpload::getFixedUploadStatus)
+     *                .thenComparing(Comparator.comparing(OCUpload::isFixedUploadingNow).reversed())
+     *                .thenComparing(Comparator.comparing(OCUpload::getFixedUploadEndTimeStamp).reversed())
+     *                .thenComparing(OCUpload::getFixedUploadId))
+     *
+     */
+    @VisibleForTesting
+    static Comparator<OCUpload> uploadListComparator = new Comparator<OCUpload>() {
```
ezaquarii: We'd still like to extract it into a separate class. 

1. We don't want to encourage other contributors to create helpers inside classes anymore, unless trivial lambda-like functions. People tend to create large nested classes and grow them out of control.

2. We don't want to set examples for `@VisibleForTesting` or similar measures that can be mitigated by better code organization. Those are abused and you end up with `public String mVeryPrivateStuff`.

I completely understand why google created this tag, after so many years of nurturing a culture of untestable code and ignorance of design patterns, but I find it a crude work-around that we'd rather avoid than promote.

Hm... I think that we can even consider using kotlin `object` to avoid instance allocations. That would create even more encouragement to use Kotlin. Now, when we have a complete test suite, we can even refactor it into more functional style without fear (if your implementation from JavaDoc can be reflected in Kotlin).

```
object UploadsListComparator : Comparator<OCUpload> { ... }
```

Conversion from Java to Kotlin is pretty much automatic, althouth I won't insist on it, as Java is ok too.

src/test/java/com/owncloud/android/ui/adapter/UploadListAdapterTest.kt
```diff
@@ -0,0 +1,186 @@
...
+import org.junit.runners.Suite
+import org.mockito.MockitoAnnotations
+
+@RunWith(Suite::class)
+@Suite.SuiteClasses(
+    UploadListAdapterTest.UploadListComparator::class,
+    UploadListAdapterTest.UploadListComparatorGeneric2::class,
+    UploadListAdapterTest.UploadListComparatorGeneric3::class
+)
+class UploadListAdapterTest {
```
ezaquarii: > Well, it tests how uploads are sorted in the uploadview (and that it does not crash as it does now).
> It just happens to be the only tested functionality

I tend to still disagree with this. This is not an adapter test and should not be named like this.

Adding a base class just made the problem more explicit, as it yelds risk of becoming a dumping ground for all sorts of things unrelated to the comparator test cases in the future.

I'd propose the following structure:

```
/**
 * <pls put copyright info here>
 */
package ...
<imports and other stuff>

class UploadsListComparatorTest { // this could be fold under adapter tests suite later

    class Base { ... } // could be Mocks, Fixture, or similar too

    class Ordering {
        ... // here are tests for fluffy UI rules, something UX ppl care about
    }

    class ComparatorContract2 {
        ... // here we check if UX ppl did proper engineering work
    }

    class ComparatorContract3 {
        ... // here we check if UX ppl did proper engineering work
    }
```

It's probably a bit overengieered but I'd rather have a solid example I can point to than going the same circles again on other ocasions. ::-)

src/test/java/com/owncloud/android/ui/adapter/UploadListAdapterTest.kt
```diff
@@ -0,0 +1,186 @@
+package com.owncloud.android.ui.adapter
```
ezaquarii: Copyright notice pls.

ezaquarii:
@fogninid sorry if you feel annoyed by the amount of massaging of this simple code, but as you can clearly see, we have tech-debt we're trying to address and that won't succeed with attitude. :)

src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
fogninid: I think the problem is that in my mind that comparator in not worth of a class on its own, coming from java8 I would have written the `Comparator.comparing...` in a static initializer.

I will extract it into a class on its own and see how it looks like if written in kotlin.
Should I move it in the same package as `OCUpload` or do you have a better place? Leaving it in the adapter package has no sense if we do not consider it part of the UploadList.

ezaquarii: I won't insist on doing Kotlin here if you don't want to sink time into it, I'm ok with Java - the test is already a massive improvement. My main concern across the project is to push hard for untangling the spaghetti of classes.

The logic behind the proposed refactoring comes from a "Separation of concerns" principle - the adapter only role is to bind the collection of items to views, not to handle ordering logic.

Quite frankly, the `UploadGroup`  does not belong to the adapter either and this entire ball of mud could be broken down.

But this is for another time. We're just making a small dent here.

fogninid: You're right most of this code belongs in the view, I will move the comparator to the same package as `OCUpload`.

While attempting to write it in kotlin my IDE complained that `Long.compare()` is only available with `API>=19`, but gradle has `minSdkVersion 14`.

Do you have any idea why this is not a problem for the current java version?
Should I write the long comparator explicitly, ignore the problem, or change the minSdk requirements to 19?

AndyScherzinger: minSdk needs to stay at 14 since that is the least version the app supports which I guess means long comparator needs to be done explicitly.

ezaquarii: @fogninid Android runtime does not implement all Java8 APIs. Some APIs will be unavailable on earlier phone models, some won't be available at all on any model, causing `ClassNotFoundException`.

Since we're supporting API level 14, the only way around it is:

- do not use the API
- use Kotlin

There is nothing we can do about it - those APIs are simply not available on all phones we want to support.

fogninid: My question why is it working now!

It was changed in master 6 months ago: https://github.com/nextcloud/android/commit/cb7134817edd33c7188e6b3559a3207c7f55412a#diff-24801376b7c7edeb4e3089df542a4de8R676

I assume that nobody is actually using such old devices anymore or caring to update them.

I will change it to use the older version, but you should probably consider to drop support or at least try to compile with such old sdk versions.

ezaquarii:
Yes this is possible that nobody hit that path on an old phone. The app will compile without trouble, but it  will fail at runtime.

@tobiasKaminsky I believe the app is tested on API 14. Is it possible that test is not covering this path?

src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
AndyScherzinger: @tobiasKaminsky can you comment on this please ❤️

fogninid: I rewrote it with `java.lang.valueOf(...)`. A pure kotlin alternative would be using `kotlin.Boolean.compareTo`: for example `upload2.isFixedUploadingNow.compareTo(upload1.isFixedUploadingNow)`

I have no experience with kotlin, are the two variants actually different when compiled and executed? I could not find any documentation about the api level of `kotlin.Boolean`, is this always going to work? I would go pure kotlin if you are sure old runtimes will have the class.

src/test/java/com/owncloud/android/ui/adapter/UploadListAdapterTest.kt
fogninid: I put my name in both author and Copyright. Or should it be Copyright Nextcloud?

src/test/java/com/owncloud/android/ui/adapter/UploadListAdapterTest.java
ezaquarii: Your name according to "_Contribution Guidelines & License_" https://github.com/nextcloud/android

src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
ezaquarii: Compiler links Kotlin `stdlib`, so it's safe to use it regardless of API level.

src/test/java/com/owncloud/android/ui/adapter/UploadListAdapterTest.kt
ezaquarii: Yes, your name. The template is in _Contribution Guidelines & License_: https://github.com/nextcloud/android -> Readme

You can also see other files.

src/main/java/com/owncloud/android/db/OCUploadComparator.kt
```diff
@@ -0,0 +1,74 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.owncloud.android.db
+
+/**
+ *  Sorts OCUpload by (uploadStatus, uploadingNow, uploadEndTimeStamp, uploadId).
+ */
+class OCUploadComparator : Comparator<OCUpload?> {
+
+    override fun compare(upload1: OCUpload?, upload2: OCUpload?): Int {
```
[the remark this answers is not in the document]
fogninid: @ezaquarii what should I do about this warning? Can I suppress it anywhere?

There is IMHO no readable way to write it without using multiple `return`s. 

I can only come up with such an abomination:

```
        return when (upload1) {
            null -> when (upload2) {
                null -> 0
                else -> -1
            }
            else -> when (upload2) {
                null -> 1
                else -> when (val compareUploadStatus = compareUploadStatus(upload1, upload2)) {
                    0 -> when (val compareUploadingNow = compareUploadingNow(upload1, upload2)) {
                        0 -> when (val compareUpdateTime = compareUpdateTime(upload1, upload2)) {
                            0 -> compareUploadId(upload1, upload2)
                            else -> compareUpdateTime
                        }
                        else -> compareUploadingNow
                    }
                    else -> compareUploadStatus
                }
            }
        }
```

src/test/java/com/owncloud/android/ui/adapter/UploadListAdapterTest.kt
fogninid: I changed mine to follow the template from the Readme, but I could not find a single occurrence in the source code which uses that format!

ezaquarii: Every source code has a header:

https://github.com/nextcloud/android/blob/43ac61409ac01e711ffe42ff2bdc2f2a148e78d5/src/main/java/com/owncloud/android/MainApp.java

I don't think they are strict about formatting. Just follow the convention. Licence is the most important thing.

src/main/java/com/owncloud/android/db/OCUploadComparator.kt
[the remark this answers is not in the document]
ezaquarii: Well, I thought that we can make something like the initially proposed Java implementation using `Comparator` API, but if that is not feasible, we can suppress the warning.

Just annotate the method with `@Suppress("ReturnCount")` and we can ship the old style implementation with early returns.

Detekt linter is still a new thing and the config has default, not always sane settings.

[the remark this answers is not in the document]
fogninid: my IDE directs me here when I ask where kotlin defines Comparator.
```
@SinceKotlin("1.1") public actual typealias Comparator<T> = java.util.Comparator<T>
```

I believe that mean that even if we write the function in kotlin it will crash at runtime with the java class does not provide those methods (i.e. on old API levels)

fogninid:
@nextcloud-android-bot well done for deleting the unresolved conversation /s

I added the `@Suppress("ReturnCount")` because I understand that `kotlin.Comparator` is just a typealias for the java class, so `Comparator.nullsFirst(...)` would crash at Runtime because java on android does not have it for old API levels

ezaquarii:
> I added the `@Suppress("ReturnCount")` because I understand that `kotlin.Comparator` is just a typealias

Ok, let's go with the old impl and not spend more time massaging it into unreachable perfection.

Thanks. Awesome work with valuable analysis as a byproduct. The extra comment mentioned earlier would be a cherry on top.

tobiasKaminsky:
Thanks for your awesome work!

PR #4253, opened 2019-07-23
Master: get lint/warnings down

tobiasKaminsky:


tobiasKaminsky:
:-D 
I could have expected that you directly see this PR.
I just changed something so that our "check if drone needs to be run" is allowing drone.
Now we will get lint/kotlin error/increased warnings.
Then I can fix them :-)

tobiasKaminsky:
All green now.
This came from the problem of "check if drone should run" which was doing false positive.

tobiasKaminsky:
I wanted to rebase/squash this, as commit history looks awful.
Note to myself: better do it directly ;-)

PR #4262, opened 2019-07-25
Fix local state if fragment is initialized before services

Unpublished:
Download, upload and operations services can be null at initialization of fragment and adapter:

https://github.com/nextcloud/android/blob/3949238de048ec18fba2e0bb518268d1738134ef/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java#L155-L157

Behavior before (without syncing status):
before

With this PR (with syncing status):
after
The sync icon is finally there :)


tobiasKaminsky:
:+1: 
Nice catch!

PR #3905, opened 2019-04-15
enhance DocumentsProvider

jmue:
This commit enhances DocumentProvider implementation in several ways:
- Get rid of 'currentStorageManager' and 'client' member variable. They are assigned (and only there) if one browsed over the document root of an account. But there are problems with this implementation. If a client request a child document information via content uri after some time again, these variables may be set to a different account.
- Delayed update of folder contents (faster). Instead of sending a network request and waiting for response before displaying folder contents, provide the least recent offline version. If the information is outdatet a update is queued and the UI is updated after response.
- Update folder contents in UI after create/rename/copy/move operations (which are currently not)
- Implement 'isChildDocument()' and expose it to support 'ACTION_OPEN_DOCUMENT_TREE' intent
- Implement 'removeDocument()' in addition to 'deleteDocument()' to support delete operations on Android >= 7.0

PR #4263, opened 2019-07-25
remove document (with relation to parent)

tobiasKaminsky:
This was missing (and is only used when opening via DOCUMENT_TREE)

@grote


grote:
Nice! Looks like a small and straightforward change.

tobiasKaminsky:
I'll merge this, although it is also in #3905 
But having #3905 a bit smaller is easier to review.

Issue #4259, opened 2019-07-25
Brainstorm how to prevent lib branch merged too early


tobiasKaminsky:
Sometimes we have change that requires a PR both on client and library.

If we merge library PR too early, then every other PR on client will fail until we merge client PR (given that we have a breaking change, e.g. rename of a class).

On client our drone system checks if library is set to master, else it will refuse to merge.
On library PR we can always merge it directly and thus break it.

We would need some way to make sure that both PRs are ready and merge-able and only then allow them to merge them within a short time span.

On client side we are somewhat safe, as this PR can only be merged once we switch to master library branch.

So the only thing is how to prevent that we merge library PR too early.

My idea is:
- add "Needed by #LinkToPRonClientRepo
- drone will check this and then check if the client PR is approved and all checks are passed and only then will allow it to merge.

Drawback: we have to remember to add the keyword to PR description.

Do you think this will work?
@AndyScherzinger @ezaquarii

AndyScherzinger:
any kind of check would be nice while I haven't understood how we would block a lib-PR from being merged breaking the app-master (or any other branch/PR)

tobiasKaminsky:
> 1. pin the library version on app master to some rc or daily build that works

This would then mean a change every time we change library or we sometimes might get out of sync

> 2. merge app and library into a monorepo that would be kept in sync and build library out of it

This would revert our idea with having them independant.

> 3. keep the snapshot, but still provide timestamped library builds, so I can switch to something that works for me in case of breakage

This is already possible as we use jitpack: https://jitpack.io/#nextcloud/android-library
There you can browse any recent build and on "get it" you see how to integrate it.

> 4. use git submodules and default to source library

How would this help? git submodule or jitpack do the same, if we target them to latest master.

I would go for "my" idea above as this does not change anything in current workflow.

tobiasKaminsky:
> any kind of check would be nice while I haven't understood how we would block a lib-PR from being merged breaking the app-master (or any other branch/PR)

This will also not work reliable.

After thinking about it again and again, Chris' idea
> 1. pin the library version on app master to some rc or daily build that works

might be the only one working:

- for master client branch we switch from -SNAPSHOT to buildId, e.g. 3652654454
- this ensures that master will always work

When creating a new feature that needs client/lib change:
- create lib PR -> will create jitpack buildId
- use this in client PR
- downside: if change to lib is needed, also a change on client PR is needed
[...]

PR #4275, opened 2019-07-28
New logger implementation

### LINKED ISSUE #4228: Log_OC doesn't collect logs in file

ezaquarii:
Fixes #4228


build.gradle
```diff
@@ -245,10 +249,13 @@ dependencies {
     // dependencies for app building
     implementation 'androidx.multidex:multidex:2.0.1'
 //    implementation project('nextcloud-android-library')
-    genericImplementation "com.github.nextcloud:android-library:$androidLibraryVersion"
-    gplayImplementation "com.github.nextcloud:android-library:$androidLibraryVersion"
-    versionDevImplementation "com.github.nextcloud:android-library:$androidLibraryVersion"
-    qaImplementation "com.github.nextcloud:android-library:$androidLibraryVersion"
+
+    implementation 'org.apache.jackrabbit:jackrabbit-webdav:2.12.6'
```
ezaquarii: This whole mess is temporary to build with custom library and will be removed once we land https://github.com/nextcloud/android-library/pull/324 and build snapshot.

src/main/java/com/nextcloud/client/logger/FileLogHandler.kt
```diff
@@ -0,0 +1,108 @@
...
+    fun rotateLogs() {
+        val rotatatingOpenedLog = isOpened
+        if (rotatatingOpenedLog) {
+            close()
+        }
+
+        val existingLogFiles = logDir.listFiles().associate { it.name to it }
+        existingLogFiles[rotationList.first()]?.delete()
+
+        for (i in 0..rotationList.size - 2) {
```
ezaquarii: use `until` - it's more intuitive

src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.java
```diff
@@ -141,112 +130,37 @@ void deleteHistoryLogging() {
      */
     @OnClick(R.id.sendLogHistoryButton)
```
ezaquarii: All of this deserve proper `ViewModel`

build.gradle
tobiasKaminsky: You can use
```
ext {
    ...
    androidLibraryVersion = "plug-custom-impl-into-legacy-logger-SNAPSHOT"
	...
}
```

This way we do not have to include any binary blob into git.

src/main/java/com/nextcloud/client/logger/FileLogHandler.kt
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    fun close() {
+        writer?.close()
+        writer = null
+        size = 0L
+    }
+
+    fun delete() {
```
tobiasKaminsky: maybe deleteAll?

src/main/java/com/nextcloud/client/logger/ThreadLoop.kt
```diff
@@ -0,0 +1,57 @@
+package com.nextcloud.client.logger
+
+/**
+ * This utility runs provided loop body continuously in a loop on a background thread
+ * and allows start and stop the loop thread in a safe way.
+ */
+internal class ThreadLoop {
```
tobiasKaminsky: Do you know if this will be killed by Android, if app is in background?
Or if it will be detected as a long running process and thus produce a warning "high battery usage"?

src/main/java/com/nextcloud/client/logger/ui/LogsAdapter.kt
```diff
@@ -0,0 +1,48 @@
...
+
+class LogsAdapter(context: Context) : RecyclerView.Adapter<LogsAdapter.ViewHolder>() {
+
+    class ViewHolder(view: View) : RecyclerView.ViewHolder(view) {
+        val header = view.findViewById<TextView>(R.id.log_entry_list_item_header)
+        val message = view.findViewById<TextView>(R.id.log_entry_list_item_message)
+    }
+
+    private val timestampFormat = SimpleDateFormat("yyyy-MM-dd HH:mm:ss'Z'", Locale.US).apply {
+        timeZone = TimeZone.getTimeZone("UTC")
```
tobiasKaminsky: This will have logs in UTC, but not in local time, which will make it harder for user to find the correct log lines?

src/main/java/com/nextcloud/client/logger/ui/LogsAdapter.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    var entries: List<LogEntry> = listOf()
+        set(value) {
+            field = value
+            notifyDataSetChanged()
+        }
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): ViewHolder {
+        val view = inflater.inflate(R.layout.log_entry_list_item, parent, false)
+        return ViewHolder(view)
```
tobiasKaminsky: Inline this?
return ViewHolder(inflater.inflate(R.layout.log_entry_list_item, parent, false))

src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.java
tobiasKaminsky: So this is not yet ready to merge?

build.gradle
ezaquarii: That is very useful trick! Thanks.

I'll remove temporary AAR and dependencies.

src/main/java/com/nextcloud/client/logger/ThreadLoop.kt
ezaquarii: If the process is shut down, it will be killed of course and restarted on process start as well.
If process is in sleep, the thread will sleep too it won't get CPU time.

This thread is blocking on a mutex most of the time.

But good point! I need to investigate it and maybe restart the loop on demand.

src/main/java/com/owncloud/android/ui/activity/LogHistoryActivity.java
ezaquarii: Not yet, just a preview but I need viewmodels from ETM branch to properly handle this.

src/main/java/com/nextcloud/client/logger/ui/LogsAdapter.kt
ezaquarii: Thanks, I'll change it to use local time.

src/main/java/com/nextcloud/client/logger/FileLogHandler.kt
ezaquarii: renamed to `deleteAll`

src/main/java/com/nextcloud/client/logger/ui/LogsAdapter.kt
ezaquarii: using local time now

src/main/java/com/nextcloud/client/logger/ui/LogsAdapter.kt
ezaquarii: done

ezaquarii:

This is new logs browser WIP/proposal. I propose to move the send/delete buttons to toolbar and here is the prototype:

logs_browser_screenshot

The space freed by bottom button bar can be used for log filtering (likely as a follow-up PRs, as I don't want to get stuck in rabbit hole here).

There are 2 things that are bothering me:

1. Icons are tinted and tint shades differ; generally icons should be shipped as monochrome b/w and tintend at runtime - is there are reason we do it this way? Some are b/w, some are tintend. Not sure how to follow with it as it might be something to do with themes.
2. Delete icon has invalid size - 16dp instead of standard 24dp and won't be correctly scaled by the toolbar. It also differ from material design icon slightly and I coudn't find original source to fix it. I placed both icons for comparision so we can make a decision what to do.

Let me know how we can progress with it.

AndyScherzinger:
just my thought, not having a final view on the points raised

> The space freed by bottom button bar can be used for log filtering (likely as a follow-up PRs, as I don't want to get stuck in rabbit hole here).

ATM any kind of search/filrtering is done via the toolbar, so maybe we could put that into the toobar, not sure though. In case send would be the primary action then we could also put that into a FAB, not sure though if we want to introduce a FAB here.

> There are 2 things that are bothering me:

> Icons are tinted and tint shades differ; generally icons should be shipped as monochrome b/w and tintend at runtime - is there are reason we do it this way? Some are b/w, some are tintend. Not sure how to follow with it as it might be something to do with themes.

That is due to the dynamic theming/tinting, yes.

> Delete icon has invalid size - 16dp instead of standard 24dp and won't be correctly scaled by the toolbar. It also differ from material design icon slightly and I coudn't find original source to fix it. I placed both icons for comparision so we can make a decision what to do.

I can take care of the icon work if you want me to. The bin-icon is the nextcloud icon of it and should be somewhere in the server repo. 

> Let me know how I should progress with it.

I think besides the UI points raised this could be a first step (to be merged after UI glitches have been resolved) and the filtering as mentioned by you would be a sweet extension ❤️

tobiasKaminsky:
> > The space freed by bottom button bar can be used for log filtering (likely as a follow-up PRs, as I don't want to get stuck in rabbit hole here).
> 
> ATM any kind of search/filrtering is done via the toolbar, so maybe we could put that into the toobar, not sure though. In case send would be the primary a

Let us do this discussion in a new issue and then changes in new PR to not get lost within this PR.

Just my opinion in short:
- search/filter should be on top, as this is most common in Android/web
- send should be first item
- delete is not often used and should be in three dots

ezaquarii:
Ok, I decided not to ship any UI changes (aside from code lint cleanups) in this PR. The UX stays the same and I'll work on improvements in another change.

The priority is to ship a working logger in 3.8, so we can start investigating long-standing issues such as power and sync.

ezaquarii:
I propose we revert order of logs from newest to oldest. Otherwise getting to any logs after few hours is impossible, unless we add search.

I see 2 use cases for logs screen atm:
1.  browsed immediately after some interesting event
2. sent via e-mail for off-line analysis

@tobiasKaminsky @AndyScherzinger let me know if this is acceptable change.

ezaquarii:

```
The following classes needed for analysis were missing: 
<few 100s class names>
...
FAILURE: Build failed with an exception.
...
BUILD FAILED in 36s 
found warnings: 410 
previous warnings: 410 
SUCCESS: count stayed the same 
```

Quite frankly spotbugs report looks normal and the count decreased from 412 to 410. I'm not sure what is happening.

.idea/codeStyles/Project.xml
```diff
@@ -60,6 +60,11 @@
       </option>
     </JavaCodeStyleSettings>
     <JetCodeStyleSettings>
+      <option name="PACKAGES_TO_USE_STAR_IMPORTS">
```
ezaquarii: Disable star imports as `ktlint` complains about wildcards.

build.gradle
```diff
@@ -57,7 +57,7 @@ configurations {
 ext {
     jacocoVersion = "0.8.2"
     daggerVersion = "2.24"
-    androidLibraryVersion = "master-SNAPSHOT"
+    androidLibraryVersion = "plug-custom-impl-into-legacy-logger-SNAPSHOT"
```
ezaquarii: @tobiasKaminsky not sure what to do with this one.

AndyScherzinger:
I'd be fine with that change

ezaquarii:
@tobiasKaminsky Can you explain how this new merging works? The description in https://github.com/nextcloud/android/issues/4259 is not entirely clear to me.

.idea/codeStyles/Project.xml
tobiasKaminsky: :+1:

tobiasKaminsky:
> Quite frankly spotbugs report looks normal and the count decreased from 412 to 410. I'm not sure what is happening.

I have also no idea, but spotBugs always fails if there is error/warning > 0:
```
> Task :spotbugsGplayDebug FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':spotbugsGplayDebug'.
> A failure occurred while executing com.github.spotbugs.internal.spotbugs.SpotBugsRunner
   > SpotBugs rule violations were found. See the report at: file:///home/tobi/projekt/nextcloud/android/build/reports/spotbugs/spotbugs.html

```

src/main/java/com/nextcloud/client/core/AsyncRunner.kt
```diff
@@ -0,0 +1,14 @@
+package com.nextcloud.client.core
+
+typealias TaskBody<T> = () -> T
+typealias OnResultCallback<T> = (T) -> Unit
+typealias OnErrorCallback = (Throwable) -> Unit
+
+/**
+ * This interface allows to post background tasks that report results via callbacks invoked on main thread.
+ *
+ * It is privided as an alternative for heavy, platform specific and virtually untestable [android.os.AsyncTask]
```
tobiasKaminsky:
```suggestion
 * It is provided as an alternative for heavy, platform specific and virtually untestable [android.os.AsyncTask]
```

src/main/java/com/owncloud/android/ui/activity/LogsActivity.java
```diff
@@ -0,0 +1,120 @@
...
+import butterknife.ButterKnife;
+import butterknife.OnClick;
+import butterknife.Unbinder;
+
+
+public class LogsActivity extends ToolbarActivity {
+
+    private Unbinder unbinder;
+
+    @BindView(R.id.deleteLogHistoryButton)
```
tobiasKaminsky: We can merge this, but I would like then to do a following PR to use LiveData and DataBindings to get rid of Butterknife (e.g. https://medium.com/@umang.burman.micro/login-example-with-mvvm-databinding-with-livedata-81319048afb0)

I am happy to do this.

tobiasKaminsky:
Please add in every new file the gpl3+ header.

tobiasKaminsky:
> @tobiasKaminsky Can you explain how this new merging works? The description in #4259 is not entirely clear to me.

I took the chance to document it: https://github.com/nextcloud/android/pull/4295

build.gradle
AndyScherzinger: nothing ;) _if_ this PR is fine (except for this line of code) _and_ the lib PR is fine:
1. merge lib PR
2. build lib via jitpack
3. revert back this line of code to master and push to branch
4. everything should now turn green (except for flaky tests...)
5. merge this PR

src/main/java/com/nextcloud/client/core/AsyncRunner.kt
```diff
@@ -0,0 +1,14 @@
+package com.nextcloud.client.core
```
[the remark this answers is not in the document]
ezaquarii: Added GPL headers to all files.

src/main/java/com/nextcloud/client/logger/Level.kt
```diff
@@ -0,0 +1,25 @@
...
+        @JvmStatic
+        fun fromTag(tag: String): Level? = when (tag) {
+            "U" -> UNKNOWN
+            "V" -> VERBOSE
+            "D" -> DEBUG
+            "I" -> INFO
+            "W" -> WARNING
+            "E" -> ERROR
+            "A" -> ASSERT
+            else -> null
```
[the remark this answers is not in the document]
ezaquarii: Not really. I simply needed some value in a non-nullable code and I didn't want to re-use existing value to hide potential errors.

If we ever see U in our logs, that means a bug, and U will be a warning.

src/main/java/com/nextcloud/client/logger/LogEntry.kt
```diff
@@ -0,0 +1,74 @@
...
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+import java.util.TimeZone
+
+data class LogEntry(val timestamp: Date, val level: Level, val tag: String, val message: String) {
+
+    companion object {
+        private const val DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
+        private val TIME_ZONE = TimeZone.getTimeZone("UTC")
```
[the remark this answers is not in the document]
ezaquarii: Yes, you seen it in UI layer where I changed it to TZ-aware format - that means that user will see TZ-aware timestamps.

Logger stores logs with unambiguous UTC timestamps.
Presentation layer uses timezone-aware formatting.

src/main/java/com/nextcloud/client/logger/LogEntry.kt
```diff
@@ -0,0 +1,74 @@
...
+        private val LEVEL_GROUP_INDEX = 2
+        private val TAG_GROUP_INDEX = 3
+        private val MESSAGE_GROUP_INDEX = 4
+
+        /**
+         *  <iso8601 date>;<level tag>;<entry tag>;<message>
+         *  1970-01-01T00:00:00.000Z;D;tag;some message
+         */
+        private val ENTRY_PARSE_REGEXP = Regex(
+            pattern = """(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z);([ADEIVW]);([^;]+);(.*)"""
```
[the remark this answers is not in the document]
ezaquarii: In practice, thanks to the initial regexp filtering, log level parsing will never return `null` , unless a maintainer adds new level and forgets to be consistent.

`Level.fromTag` is simply more robust than this code requires. In case of a maintainer's mistake, sombebody will see `U` in logs instead of `NPE`, but that extra safety layer is just a byproduct of Kotlin null-safety. I simply had to provide a non-null value here.

src/main/java/com/owncloud/android/ui/activity/LogsActivity.java
```diff
@@ -0,0 +1,120 @@
...
+import androidx.lifecycle.ViewModelProvider;
+import androidx.recyclerview.widget.LinearLayoutManager;
+import androidx.recyclerview.widget.RecyclerView;
+import butterknife.BindView;
+import butterknife.ButterKnife;
+import butterknife.OnClick;
+import butterknife.Unbinder;
+
+
+public class LogsActivity extends ToolbarActivity {
```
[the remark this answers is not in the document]
ezaquarii: This is an old `LogHistoryActivity`, just renamed. I'm happy to rewrite it in Kotlin and add data bindings in layout.

I'd prefer to do it as a follow up as I don't know if I can manage to ship it before 3.8.

src/main/java/com/owncloud/android/ui/activity/LogsActivity.java
ezaquarii: I'm happy to give it another round of refactoring. However, I believe that shipping working logs ASAP is going to benefit us in other areas more than UI.

src/main/java/com/owncloud/android/ui/activity/LogsActivity.java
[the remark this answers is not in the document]
AndyScherzinger: > I'd prefer to do it as a follow up as I don't know if I can manage to ship it before 3.8.

I'd second that

src/main/java/com/owncloud/android/ui/activity/LogsActivity.java
AndyScherzinger: I agree with @ezaquarii on this

src/main/java/com/nextcloud/client/logger/Level.kt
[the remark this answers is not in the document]
ezaquarii: Actually, you are right. I'll change this and make the whole function non-null.

src/main/java/com/nextcloud/client/logger/LogEntry.kt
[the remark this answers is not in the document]
ezaquarii: I changed the `Level` parser so it does not return `null` anymore.

src/main/java/com/nextcloud/client/logger/LogEntry.kt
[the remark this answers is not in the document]
ezaquarii: Entries are **stored** using UTC timestamps (no ambiguities == no problems).
**UI** presents dates using TZ-aware formatting. See `LogsAdapter`.

The e-mail formatter is a bit controversial. Currently it uses UTC formatting, as I believe that logs sent to us will be analyzed by experts and there are some edge cases I'm concerned about, such as summer/winter time changes, moves across timezones. Also, I see some benefits in preserving original format in case we want to perform some automated logs parsing (there is a lot of noise there), although having TZ added in timestamps won't hurt that use case.

src/main/java/com/nextcloud/client/core/AsyncRunner.kt
ezaquarii: Fixed typo in javadoc.

src/main/java/com/nextcloud/client/core/AsyncRunnerImpl.kt
```diff
@@ -0,0 +1,46 @@
+package com.nextcloud.client.core
```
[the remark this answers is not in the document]
ezaquarii: added headers to all sources

build.gradle
ezaquarii: @tobiasKaminsky @AndyScherzinger Ok, so it looks like you need to sort out library merge first. :)

src/main/java/com/nextcloud/client/logger/LogEntry.kt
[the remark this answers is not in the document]
tobiasKaminsky: > Let me know if you want to have TZ in logs sent via e-mail too.

Somewhere in the email we should have the info which TZ the user reported at.
I guess it is the best to have it in each line, though it is overhead and noisy.

src/main/java/com/owncloud/android/ui/activity/LogsActivity.java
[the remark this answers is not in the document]
tobiasKaminsky: Yeah, no rush. Small PRs and small changes are good :-)

src/main/java/com/owncloud/android/ui/activity/LogsActivity.java
tobiasKaminsky: of course, my idea was not meant to block merging this PR, but to not forget about the potential next refactoring round.

src/main/java/com/nextcloud/client/core/ClockImpl.kt
```diff
@@ -0,0 +1,35 @@
+/*
+ * Nextcloud Android client application
+ *
+ * @author Chris Narkiewicz
+ *
+ * Copyright (C) 2019 Chris Narkiewicz <hello@ezaquarii.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
```
tobiasKaminsky: This is GPL3+, an AppModule for example you used AGPL3+.
While the changes are minor, we encourage APGL3+.
(sorry if I mixed this up and falsly suggested GPL3+…)

ezaquarii: I think I copied the header from other source file - some are GPL3.

It should be all converted to AGPL3 now.

src/main/java/com/nextcloud/client/logger/LogEntry.kt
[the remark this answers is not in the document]
ezaquarii: I added TZ information in logs attached to e-mail. 5 extra chars should not be a problem.

src/main/java/com/nextcloud/client/core/ClockImpl.kt
ezaquarii: Now it should be really done. (-:

tobiasKaminsky:
> https://www.kaminsky.me/nc-dev/android-integrationTests/10299/classes/com.nextcloud.client.logger.TestLogger.html

I tried to run these unit tests locally, but I always get:
``` 
java.lang.RuntimeException: Method d in android.util.Log not mocked. See http://g.co/androidstudio/not-mocked for details.

	at android.util.Log.d(Log.java)
	at com.nextcloud.client.logger.LoggerImpl.d(LoggerImpl.kt:66)
	at com.nextcloud.client.logger.TestLogger.queue overflow warning is logged(TestLogger.kt:231)
``` 

Am I missing something?

ezaquarii:
@tobiasKaminsky
The only thing coming to my mind is 

```
        testOptions {
            unitTests.returnDefaultValues = true
        }
``` 

in gradle. If you restored the library by restoring entire `build.gradle` by accident, this is what will happen.

ezaquarii:
@tobiasKaminsky Testing platform code is tricky. I'm on the fence between having tests on android devices (`androidTests`) vs grabbing `Robolectric`. I decided to keep any extra dependencies out as long as we really don't require them.

tobiasKaminsky:
> If you restored the library dependency by restoring entire `build.gradle` by accident, this is what will happen.

I did not (yet) changed anything, but only checked it out.
I restart drone and give it also a new clean test locally.



> I decided to keep any extra dependencies out as long as we really don't require them.

Good :-) If we come to the point where we have to change this, we should really be sure about the change.

tobiasKaminsky:
Interesting. Running TestLogger via Android Studio results in error above.

But running "./gradlew testGplayDebugUnitTest" says that everything works:
> com.nextcloud.client.logger.TestLogger > queue overflow warning is logged PASSED

--> so false alarm, but still strange

tobiasKaminsky:
After CI, I will merge lib, and do all the needed stuff to get this in.

PR #4286, opened 2019-07-31
FIDO U2F Support

dschuermann:
As discussed by email, this is our FIDO U2F implementation for Android, which is 100% GPLv3 and not depending on Google Play Services.

This implementation works by adding the Javascript FIDO API to the WebView and bridging it to our Java calls. More details: https://hwsecurity.dev/guide/fido-webview/

I tried this with two different nextloud instances where the nextcloud-twofactor_u2f plugin is activated.

I am interested in Feedback and discussions around this PR.

## Source Code
https://github.com/cotechde/hwsecurity

## Context regarding FIDO Support on Android
* Google's proposes to use their [Google Play Services for FIDO](https://developers.google.com/android/reference/com/google/android/gms/fido/u2f/U2fApiClient). These are closed source and not available on phones that do not ship Google Play. Especially due to the recent US sanctions, this is even more important.
* The WebView itself will probably not support FIDO in the near future: ["This feature will be supported on all platforms except for Android WebView."](https://groups.google.com/a/chromium.org/d/msg/blink-dev/wfIVkXvQ7kQ/VfuOr_FhBwAJ)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -284,6 +284,8 @@ dependencies {
     implementation 'com.google.code.gson:gson:2.8.5'
     implementation 'org.jetbrains:annotations:17.0.0'
 
+    implementation 'com.github.cotechde.hwsecurity:hwsecurity-fido:2.4.1'
+
     spotbugsPlugins 'com.h3xstream.findsecbugs:findsecbugs-plugin:1.9.0'
     spotbugsPlugins 'com.mebigfatguy.fb-contrib:fb-contrib:7.4.6'
 
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -94,6 +94,8 @@ import androidx.multidex.MultiDexApplication;
 import dagger.android.AndroidInjector;
 import dagger.android.DispatchingAndroidInjector;
 import dagger.android.HasAndroidInjector;
+import de.cotech.hw.SecurityKeyManager;
+import de.cotech.hw.SecurityKeyManagerConfig;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 import static com.owncloud.android.ui.activity.ContactsPreferenceActivity.PREFERENCE_CONTACTS_AUTOMATIC_BACKUP;
@@ -210,6 +212,13 @@ public class MainApp extends MultiDexApplication implements HasAndroidInjector {
     @Override
     public void onCreate() {
         super.onCreate();
+
+        SecurityKeyManager securityKeyManager = SecurityKeyManager.getInstance();
+        SecurityKeyManagerConfig config = new SecurityKeyManagerConfig.Builder()
+            .setEnableDebugLogging(BuildConfig.DEBUG)
+            .build();
+        securityKeyManager.init(this, config);
+
         registerActivityLifecycleCallbacks(new ActivityInjector());
 
         Thread t = new Thread(() -> {
--- a/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -51,6 +51,7 @@ import android.content.ServiceConnection;
 import android.content.SharedPreferences;
 import android.content.pm.ActivityInfo;
 import android.content.pm.PackageManager;
+import android.graphics.Bitmap;
 import android.graphics.Rect;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
@@ -76,6 +77,8 @@ import android.webkit.CookieManager;
 import android.webkit.CookieSyncManager;
 import android.webkit.HttpAuthHandler;
 import android.webkit.SslErrorHandler;
+import android.webkit.WebResourceRequest;
+import android.webkit.WebResourceResponse;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
 import android.widget.EditText;
@@ -145,6 +148,7 @@ import androidx.fragment.app.DialogFragment;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentManager;
 import androidx.fragment.app.FragmentTransaction;
+import de.cotech.hw.fido.WebViewFidoBridge;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /**
@@ -233,6 +237,8 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
     private ImageButton mTestServerButton;
 
     private WebView mLoginWebView;
+
+    private WebViewFidoBridge webViewFidoBridge;
 
     private String mAuthStatusText = EMPTY_STRING;
     private int mAuthStatusIcon;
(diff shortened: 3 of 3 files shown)


dschuermann:
Not sure why CI's analysis is failing, maybe someone from the team can provide insights.

AndyScherzinger:
cc @tobiasKaminsky 

Regarding the raised lint score:
* 33 warnings due to `Logging call to Log instead of Timber` (is the implementation making use of Timber since the Nextcloud app doesn't)
* 1 warning due to `Lint found an issue registry (timber.lint.TimberIssueRegistry) which is older than the current API level; these checks may not work correctly.` which again points to the introduction of Timber

These two issues account for all additional warnigns within this PR, my guess: 
`implementation 'de.cotech:hwsecurity-fido:2.3.3'` comes with a dependency to Timber which is detected by the linter and thus the linter activates further checks when using Timber which lead to these issues. The simplest thing would maybe to just deactivate these Timber related checks (if that is possible).

tobiasKaminsky:
Thank you very much for this PR.

I tested it and it works great :+1: 
On the very first use I had to unplug and plug in the the usb device to get the prompt for allowing it.

Would it possible for you to use Log instead of Timber?

mario:
I also see a mention of licence on your website. Can you please clarify that?

In addition, I'm not sure we want dependencies which require user/pass to be able to be pulled.

ezaquarii:
We probably don't want to use android.util.Log neither, as it's not for an external library to make a logging strategy decision at all.

We'd like to have a possibility to plug our own logging callback and decide on app side what to do with it.

We'll most likely want to redirect those logs to file with our own tagging and verbosity settings or disable it to avoid leaks of private data via adb, so something as simple as a lambda (or few for log levels) would do the job just fine. By default it can be android.util.Log, but it should be swappable to give us control.

I don't see much value in Timber, TBH.

dschuermann:
Thanks for the valuable feedback.

## Logging
We removed the dependency on Timber and instead include one file in our library that is responsible for logging.

To enable simple debug logging using Android's Log class:
```
SecurityKeyManagerConfig.Builder().setEnableDebugLogging(BuildConfig.DEBUG)
```

If you like to use your own custom logger you can use something like this:
```
SecurityKeyManagerConfig.Builder()
    .setLoggingTree(new HwTimber.DebugTree() {
        @Override
        protected String createStackElementTag(@NonNull StackTraceElement element) {
            if (element.getClassName().startsWith("de.cotech.hw")) {
                return super.createStackElementTag(element);
            } else {
                return null;
            }
        }
    
        @Override
        protected boolean isLoggable(String tag, int priority) {
            return tag != null;
        }
    
        @Override
        protected void log(int priority, String tag, @NonNull String message, Throwable t) {
            // TODO: add your own logging implementation here.
        }
    })
```

## Other questions
@mario 

> I also see a mention of licence required on your website. Can you please clarify that?

The license is GPLv3: https://github.com/cotechde/hwsecurity/blob/master/LICENSE.md

> In addition, I'm not sure we want dependencies which require user/pass to be able to be pulled.

User/password is not used anywhere in our SDK. Nothing is "pulled" from the WebView. What we do: We inject the FIDO U2F Javascript API into the WebView and bridge it to our FIDO implementation in Java. You can inspect the provided API here: https://github.com/cotechde/hwsecurity/blob/master/hwsecurity-fido/src/main/assets/fidobridge.js

dschuermann:
>  On the very first use I had to unplug and plug in the the usb device to get the prompt for allowing it.

@tobiasKaminsky What hardware setup are you using (security key? smartphone? USB adapter?). Any other dialogs that popped up and asked for permission for other apps?

dschuermann:
>     In addition, I'm not sure we want dependencies which require user/pass to be able to be pulled.

Sry, I misunderstood your question. If this is a blocker we can think of other ways to provide the dependency.

ezaquarii:
> We removed the dependency on Timber and instead include one file in our library that is responsible for logging.

Thank you for a quick change. I reviewed the code and I see the comment stating that you vendored latest `Timber` into your project. I personally have nothing against it, but from engineering point of view this means that we'll be susceptible to any `Timber` API changes in the future or you'll be blocked from `Timber` updates.

You'd need to maintain this API regradless of changes made by `Timber` upstream. I'd reconsider exposing it as your public API.

dschuermann:
> > We removed the dependency on Timber and instead include one file in our library that is responsible for logging.
> 
> Thank you for a quick change. I reviewed the code and I see the comment stating that you vendored latest `Timber` into your project. I personally have nothing against it, but from engineering point of view this means that we'll be susceptible to any `Timber` API changes in the future or you'll be blocked from `Timber` updates.
> 
> You'd need to maintain this API regradless of changes made by `Timber` upstream. I'd reconsider exposing it as your public API.

Timber is just one file and it perfectly fits the purpose of having a simple logging mechanism that also allows other to set their own ``DebugTree``. So this was the easiest way for us to provide your logging requirement. If I would start developing my own Tree interface it would probably look similar. Maybe I'll later move the ``HwTimber`` logging class into our internal package and only expose ``HwTimber.Tree``.

mario:
> > In addition, I'm not sure we want dependencies which require user/pass to be able to be pulled.
> 
> @mario Sry, I misunderstood your question. If this is a blocker we can think of other ways to provide the dependency.

Yes, please :)

dschuermann:
> > > In addition, I'm not sure we want dependencies which require user/pass to be able to be pulled.
> > 
> > 
> > @mario Sry, I misunderstood your question. If this is a blocker we can think of other ways to provide the dependency.
> 
> Yes, please :)

We now have a repo at maven-free.cotech.de that will always match the Github source at https://github.com/cotechde/hwsecurity for GPL projects to use.

mario:
Any chance to just use Jitpack @dschuermann? See part of the reasoning here: https://github.com/nextcloud/talk-android/issues/534

AndyScherzinger:
I second @mario's suggestion. Unfortunately https://github.com/cotechde/hwsecurity/tags is empty so one would have to go for the commit so for `2.4.1`I'd have to use:
```
dependencies {
	        implementation 'com.github.cotechde.hwsecurity:hwsecurity:7b5f1af'
	}
```
via jitpack builds: https://jitpack.io/#cotechde/hwsecurity

dschuermann:
> I second @mario's suggestion. Unfortunately https://github.com/cotechde/hwsecurity/tags is empty so one would have to go for the commit so for `2.4.1`I'd have to use:
> 
> ```
> dependencies {
> 	        implementation 'com.github.cotechde.hwsecurity:hwsecurity:7b5f1af'
> 	}
> ```
> 
> via jitpack builds: https://jitpack.io/#cotechde/hwsecurity (just created them). Other than that this would also be a mvn repo developers are used to / is known to them.

great, thanks for the pointer. Jitpack is a better solution for the GPL release. I added a tagged release and changed this PR.

mario:
Thanks @dschuermann. I'd appreciate the same PR for Nextcloud Talk! :)

tobiasKaminsky:
> @tobiasKaminsky What hardware setup are you using (security key? smartphone? USB adapter?). Any other dialogs that popped up and asked for permission for other apps?

OnePlus3 with usb adapter
key is Key-ID FIDO U2F (https://www.amazon.de/Key-ID-FIDO-U2F-USB-Sicherheitsschl%C3%BCssel/dp/B01N6XNC01)

No other popup.
But it seems that it was only the very very first start…
Now it seems to be working just fine.

mario:
Might also need commits signing, but otherwise looks good.

dschuermann:
> Might also need commits signing, but otherwise looks good.

done :)

tobiasKaminsky:
Thank you very much.
I am looking forward to see people using it :+1:

AndyScherzinger:
@tobiasKaminsky build turned green, so I took the liberty to merge like you stated 😊

Thanks again for this PR, looking forward to release it with the upcoming feature release 🚀

dschuermann:
Great, can't wait to see it released :)

AndyScherzinger:
Should be around 3 week from now, give or take

PR #4287, opened 2019-08-01
Fix upload and delete to actually delete

### LINKED ISSUE #3840: File is not deleted after upload
### Actual behaviour
- when i upload a file i choose the option "delete file from original folder". It worked a while ago but now it doesn t anymore.

### Expected behaviour
- the file should be deleted from my local folder after being uploaded 
 
### Steps to reproduce
1. 
2. 
3. 

### Environment data
Android version:

Device model: 

Stock or customized system:

Nextcloud app version:

Nextcloud server version:

#### Web server error log
```
```

```
```

Unpublished:
The removed logic results in falling back to LOCAL_BEHAVIOUR_FORGET and was introduced here: https://github.com/nextcloud/android/pull/3467/files#diff-ecdca245e1aa615e4581732a6606684bR1038-R1044

It seems like this is a left over from the logic below:
https://github.com/nextcloud/android/blob/ea44c595a8d7b43bb9f1e6787b1540cd85791a8f/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java#L1073-L1089

Fixes #3840


tobiasKaminsky:
Nice one!
Thank you very much!

PR #4309, opened 2019-08-11
Log search functionality and log browser refactoring

### LINKED ISSUE #4311: Log: do not quit view after reset
### Actual behaviour
- Log view quits after clearing

### Expected behaviour
- only clear log, but do not exist
 
### Steps to reproduce
1. go into log view
2. click on clear logs

ezaquarii:
* added log search in logs browser
* added logs browser view model test
* added universal async filtering utility
* refactored async runner and added manual runner
* migrated logs browser to Android DataBinding and Kotlin
* disabled imports ordering Ktlint rule as IDE does not
  support ktlint ordering
* added some missing tests around logger

Closes #4311


ezaquarii:
@tobiasKaminsky @AndyScherzinger This is a follow-up on https://github.com/nextcloud/android/pull/4275 with discussed UI improvements.

AndyScherzinger:
@ezaquarii I haven't had a chance to test it yet, first review looks good code-wise. Did you take care of themeing (search icon / search view)? If not @tobiasKaminsky might be able to help :)

ezaquarii:
> Did you take care of themeing (search icon / search view)?

@AndyScherzinger I tried to call `ThemeUtils.themeSearchView(...)` but the font remains dimmed. I see that in other parts we use 100% opaque white font. Any idea where I can find a suitable example of working setup?

ezaquarii:
@AndyScherzinger @tobiasKaminsky I tried to copy the solution from `ExtendedListFragment.onCreateOptionsMenu` but the search view won't budge - it's still dimmed. I'm not sure how the theming is done there, but it looks like it depends on some hidden state set elsewhere.

AndyScherzinger:
@tobiasKaminsky can you provide @ezaquarii with the infos for search bar theming (which you already implement4ed elsewhere in the app)? :)

tobiasKaminsky:
Regarding translation strings:
Please do not rename keys that are already in use, otherwise they will conflict with older version (we translate master and 1-2 older versions within the same transifex repository)
Just create a new string, even when it is with the same content. It causes overhead to the translation team, but this way we are bulletproof.

As transifex is using only values/strings.xml as source, we do not need to change the translated files. In best case they will be overwritten by next transifex update, and in worst case they result in a conflict.
Modifying only values/strings.xml makes review also a bit easier.

After that I am happy to review it (currently my browser refuses a bit to show that many changes…)

I will add both to our contributing.md

tobiasKaminsky:
> @AndyScherzinger @tobiasKaminsky I tried to copy the solution from `ExtendedListFragment.onCreateOptionsMenu` but the search view won't budge - it's still dimmed. I'm not sure how the theming is done there, but it looks like it depends on some hidden state set elsewhere.

> @tobiasKaminsky can you provide @ezaquarii with the infos for search bar theming (which you already implement4ed elsewhere in the app)? :)

Sorry, I do not get the term "budge". What is wrong?
At a quick test, only three dots are tinted wrong on a light theme?
(not sure if this is even a problem with this PR, I doubt so…)

AndyScherzinger:
@tobiasKaminsky -  could you test if it works with (very) light Background and dark text/icons? Else it seems fine (except for the 3-dot icon)

tobiasKaminsky:
I reverted translation files and rebased.
I'll review & test it today and then let's merge it.

After that I'll have a look again at the tinting problematic, as I think that the ThemeUtils.themeSearchView() is not working and thus it might be changed also in other places.

src/main/java/com/nextcloud/client/core/Task.kt
```diff
@@ -0,0 +1,38 @@
+package com.nextcloud.client.core
```
tobiasKaminsky: AGPL3+ please

src/main/res/layout/logs_activity.xml
```diff
@@ -16,57 +16,47 @@
   You should have received a copy of the GNU General Public License
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
```
tobiasKaminsky: As you changed nearly everything you can change this also to AGPL3+

ezaquarii:
Damn, I copied it from other places in our source tree. It seems like we're having some inconsistencies there because every time I copy the header we find issues. :)

@tobiasKaminsky, could you please put a golden sample in Readme? That would limit the risk of copying wrong header form more "legacy" code.

I'll try to change this one this evening, as I guess it must be me.

Otherwise, I have bandwidth shortages this week so please free to address other little things to speed up things as you see fit.

Cheers,
Chris

AndyScherzinger:
Ready, steady, merge 🌟

PR #4326, opened 2019-08-15
Centralize theming of search bar

tobiasKaminsky:
In the last screenshot I am not sure if we should tint the search input field. I think it is better to not use primary color here, but use black for all.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -318,7 +318,7 @@
         <activity android:name=".ui.activity.ConflictsResolveActivity"/>
         <activity android:name=".ui.activity.ErrorsWhileCopyingHandlerActivity"/>
 
-        <activity android:name="com.nextcloud.client.logger.ui.LogsActivity"/>
+        <activity android:name="com.nextcloud.client.logger.ui.LogsActivity" />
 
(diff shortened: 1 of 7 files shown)


AndyScherzinger:
> In the last screenshot I am not sure if we should tint the search input field. I think it is better to not use primary color here, but use black for all.

👍

tobiasKaminsky:
> > In the last screenshot I am not sure if we should tint the search input field. I think it is better to not use primary color here, but use black for all.

> :+1: 

Then I'll change this.

tobiasKaminsky:
> > > In the last screenshot I am not sure if we should tint the search input field. I think it is better to not use primary color here, but use black for all.
> 
> > +1
> 
> Then I'll change this.

This is also used in PinCodeActivity…
For very light themes this just looks strange, so it is better to have it black on white.

AndyScherzinger:
still fine with me 👍 

So awaiting @ezaquarii's review :)

AndyScherzinger:
Pinging @ezaquarii to maybe get this in before Monday (rc branching)

AndyScherzinger:
Pinging @ezaquarii to maybe get this in for RC1

PR #4345, opened 2019-08-19
Push: no need to use ocClient

tobiasKaminsky:
Requires https://github.com/nextcloud/android-library/pull/328


tobiasKaminsky:
I'll merge this now as other drone jobs will not work as the lib PR is already merged.

PR #4348, opened 2019-08-19
hide user account from upload view items for single account setups

### LINKED ISSUE #1943: Uploads view doesn't need a user info line if only one account is in the app
The row with the username for every upload is only necessary as soon as there are at least 2 accounts in the app.

AndyScherzinger:
Resolves #1943

e.g.
device-2019-08-19-140634

### DIFF
--- a/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
+++ b/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
@@ -228,17 +228,21 @@ public class UploadListAdapter extends SectionedRecyclerViewAdapter<SectionedVie
 
         // upload date
         long updateTime = item.getUploadEndTimestamp();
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@tobiasKaminsky as described by @jancborchardt - I am aware that this is a starter issue but since nobody took care of it since beginning of 2018 I thought it's fine for me to pick-up the task and it helps with the UI since most people probably only have one account anyways :)

tobiasKaminsky:
I just added a whitespace in "if(showUser)" and formatted all parameters to be in one line each, if line > 120 chars.

AndyScherzinger:
@tobiasKaminsky any idea why tests keep failing (imho unrelated to the PR's changes)?

tobiasKaminsky:
No idea. In both cases they were different tests, so… just merge ;-)

PR #4355, opened 2019-08-20
drawer: show only server address

### LINKED ISSUE #4337: Navigation drawer shows full name instead of username
### Actual behaviour
When opening the navigation drawer, the displayed Cloud ID doesn't consist of `<username>@<host>` but instead `<full name>@<host>` (see screenshot). In the "Manage accounts" section, the Cloud ID is displayed correctly.

### Expected behaviour
The Cloud ID should contain the username, not the full name. I remember it showing correctly, not sure what update update caused this. Maybe this is somehow related to #157 ?

### Environment data
Android version: 7.1.2 (LineageOS)
Nextcloud app version: 3.7.1
Nextcloud server version: 16.0.4

tobiasKaminsky:
Fix #4337 


tobiasKaminsky:
Maybe we should do the same in account list? (in a later PR)

AndyScherzinger:
Spotbugs again... :/

tobiasKaminsky:
> Spotbugs again... :/

This is unrelated to this PR, as it is a warning about a list, so I just increased count here.

AndyScherzinger:
> Maybe we should do the same in account list? (in a later PR)

Yes, I'd think so 👍

tobiasKaminsky:
> > Maybe we should do the same in account list? (in a later PR)
> 
> Yes, I'd think so

This is a bit tricky as in account list it is a menu within NavigationView.
There we cannot easily another line.

AndyScherzinger:
> This is a bit tricky as in account list it is a menu within NavigationView.
There we cannot easily another line.

Than I miunderstood you, I meant `full account name@server`

tobiasKaminsky:
> > This is a bit tricky as in account list it is a menu within NavigationView.
> > There we cannot easily another line.
> 
> Than I missunderstood you, I meant `full account name@server`

I meant that we have also two lines
- Display Name (full name)
- server address

Currently we already use display Name, so it is ok, I guess.

AndyScherzinger:
> Currently we already use display Name, so it is ok, I guess.

yes, than we are perfectly fine 👍

PR #4363, opened 2019-08-21
drone only on specific worker

tobiasKaminsky:


tobiasKaminsky:
Nope :/
My server is too slow for this.

PR #4372, opened 2019-08-22
throttle avatar lookup

tobiasKaminsky:
We do eTag comparison, if available for avatars.
But this means that we still check every time when displaying an avatar.

This enhances it to check only
- if no avatar is in cache
- last lookup is >1h ago

It of course means that the avatar might be outdated, but to be honest, how often is this changed? ;-)

### DIFF
--- a/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.java
+++ b/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.java
@@ -411,7 +411,7 @@ public class PreviewImageActivity extends FileActivity implements
                 }
 
                 // Call to reset image zoom to initial state
-                ((PreviewImagePagerAdapter) mViewPager.getAdapter()).resetZoom();
+                // ((PreviewImagePagerAdapter) mViewPager.getAdapter()).resetZoom();
             }
         }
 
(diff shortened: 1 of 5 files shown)


tobiasKaminsky:
/backport to stable-3.8

(this is not a bugfix, but we have to do RC2 anyways and with showing sharees in list view, we will have many many avaters, and this will slow down app otherwise)

tobiasKaminsky:
IT test is unrelated.

src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.java
```diff
@@ -411,7 +411,7 @@ public void onPageSelected(int position) {
                 }
 
                 // Call to reset image zoom to initial state
-                ((PreviewImagePagerAdapter) mViewPager.getAdapter()).resetZoom();
+                // ((PreviewImagePagerAdapter) mViewPager.getAdapter()).resetZoom();
```
PhilLab, line 414: @tobiasKaminsky while reading through the code I stumbled upon this line which is still commented out on `master`. I wonder if this got committed by accident?

The original zoom reset was added in bbbc486d532878095a3133639641e8a6e529d22a

PR #4387, opened 2019-08-23
fix possible NPE on stopping contact backup job

tobiasKaminsky:


tobiasKaminsky:
Something in this line is null. I think it is extras, as account is used previously and then it would blow up there.

PR #4389, opened 2019-08-23
Fix build warning

tobiasKaminsky:
> Configure project :
539	src: 'annotationProcessor' dependencies won't be recognized as kapt annotation processors. Please change the configuration name to 'kapt' for these artifacts: 'org.projectlombok:lombok:1.18.8'.


AndyScherzinger:
Rebased, resolved conflicts, waiting for CI results

tobiasKaminsky:
Hm. I guess this is somewhat silly, as lombok is working on java code, whereas kapt is for kotlin.
We need to live with this warning until we replaced all Lombok and thus can drop it.

PR #4394, opened 2019-08-23
3.9.0 Alpha

AndyScherzinger:
Bump version on master

cc @tobiasKaminsky Should be add this to the release roadmap issues (template) since this should be done whenever we create a stable branch.


tobiasKaminsky:
> cc @tobiasKaminsky Should be add this to the release roadmap issues (template) since this should be done whenever we create a stable branch.

added

Issue #4272, opened 2019-07-27
Odd Activities Behavior


Boballen55:
### Actual behaviour
If I click on the Activities menu options it loads a recent list of activities and all looks well. If I scroll down it attempts to pull more history from the server but instead adds the same bit of entries to the bottom over and over.

### Expected behaviour
Ideally new data would be pulled from server. Alternatively just the limited set of initial data would be logical if limited. 
 
### Steps to reproduce
1. Install Android 3.7.0
2. Connect to server 
3. Open activities and keep scrolling down, the entries should repeat 

### Environment data
Android version: 8.1.0

Device model: Nexus 6P

Stock or customized system: Lineagos

Nextcloud app version: 3.7.0

Nextcloud server version: 15.0.10snap2

### Logs
Where do I find my Nexcloud log in the snap version?

tobiasKaminsky:
@Boballen55 thanks for reporting.
This have been slipped through, but now I fixed it.

In https://github.com/nextcloud/android/pull/4419 you will soon find a APK you can test, if you want.

PR #4419, opened 2019-08-26
activity: use lastGiven header

tobiasKaminsky:
Fix https://github.com/nextcloud/android/issues/4272


tobiasKaminsky:
@Boballen55 thanks for testing, this was very important!
I indeed added a limit of 5 for fetching activities, instead of default of 50.

I changed it, so it will be fast again.

src/main/java/com/owncloud/android/ui/activities/ActivitiesActivity.java
```diff
@@ -125,8 +125,8 @@ protected void onCreate(Bundle savedInstanceState) {
         swipeListRefreshLayout.setOnRefreshListener(() -> {
             // We set the nextPageUrl variable to null here since when manually refreshing
             // activities data we want to clear the list and reset the pagination.
-            nextPageUrl = null;
-            mActionListener.loadActivities(nextPageUrl);
+            lastGiven = -1;
```
AndyScherzinger: It seems to me "-1" is a magic number and should be defined as a constant.

tobiasKaminsky: Nice catch, this makes it indeed more readable

PR #4434, opened 2019-08-28
Update MimeTypes

### LINKED ISSUE #2073: nextcloud doesn't know how to open .love files
### Actual behaviour
nextcloud cannot open .love files

### Expected behaviour
nextcloud should open .love files with the love2d engine
https://play.google.com/store/apps/details?id=org.love2d.android
 
### Steps to reproduce
1. put a .love game into nextcloud
2. try to open it

### Environment data
Android version:
7.1 

Device model: 
ZTE Axon 7

Nextcloud app version:
newest

### LINKED ISSUE #2081: Matlab source files not opened as text files
### Actual behaviour
When opening a file with a `.m` extension (Matlab source file), the app downloads it and prompts you to open with apps that don't even deal with text files - https://i.imgur.com/tXdv7PA.jpg. 

Perhaps consider adding "Open As" button the same way the Total Commander app does (it lets you open files as if they were a text file/image/video/music no matter what the extension or mime type is, in case the app doesn't recognise it properly)? 

### Expected behaviour
Either it should open as a text file, or it should let you to open it with an app that supports text files
 
### Steps to reproduce
1. Have a .m file on NextCloud
2. Try to open it with the Android app

### Environment data
Android version: 8.0.0

Device model: OnePlus 3

Nextcloud app version: 2.0.0

Nextcloud server version: v13.0.0b4

AndyScherzinger:
Updated MimeType list to Nc17 list, also implemented 2 minor extensions to support opened issues.

* Resolve #2073
* Resolve #2081

I vote for simply merging the change (if the code itself is fine and works) and _if_ as discussed we want to move to a mime type mapping fetched from the server than we should do that in a later, separate PR.


AndyScherzinger:
@tobiasKaminsky I did some refactoring/extension of thhe mime type detection to not just rely on the OS but also use the ever exisitng local mime type maps as a fallback :)

AndyScherzinger:
@tobiasKaminsky I shared a folder with you with plain/empty files with all the extensions added via this PR :)

PR #4458, opened 2019-09-04
rename participate -> community

tobiasKaminsky:
Also fixed button on white screen


AndyScherzinger:
@tobiasKaminsky doesn't the rename lead to the need that all strings need to be translated again? (shouldn't be too much work though)

tobiasKaminsky:
> @tobiasKaminsky doesn't the rename lead to the need that all strings need to be translated again? (shouldn't be too much work though)

Yes, but we have plenty of time as it will be in 3.9 and for translators it should be just one click on each as the value did not changed and therefore transifex suggests the correct translation.

I was thinking if I should change it, but I think it is better to have all in a common term.

AndyScherzinger:
@tobiasKaminsky spotbugs increased while not being increased?!

PR #4482, opened 2019-09-09
(temporary) fix for navigation issue

tobiasKaminsky:
For 3.9. I'll do a complete rewrite of navigation flow:
- get rid of event bus
- each search in an own fragment
- lifecycle aware


AndyScherzinger:
It seems better than before. The original issue seems fixed so that'll work for 3.8.0 while I agree with Tobias that navigation needs a proper post 3.8.0 rewrite

AndyScherzinger:
❗️ rebased and fixed a spotbugs issue to fix build.

Issue #184, opened 2016-08-01
Upload whole folder


tobiasKaminsky:
When uploading make it possible to choose the complete folder with all subfolder.

AndyScherzinger:
Sounds good to me!

I see two scenarios:
* select a folder: upload the folder and all its content including all sub folders (so the selected folder also gets created on the server)
* go into a folder with sub folders and choose "select all": no separate folder gets created, all selected files, folders, the content within the sub folders gets uploaded to the chosen location.

Any thought? @tobiasKaminsky @jancborchardt ?

tobiasKaminsky:
:+1:

jancborchardt:
Yup, sounds good!

AndyScherzinger:
Dev hint - corresponding activity: 
https://github.com/nextcloud/android/blob/master/src/com/owncloud/android/ui/activity/UploadFilesActivity.java

kollix:
I don't see the second scenario as needed:

"go into a folder with sub folders and choose "select all": no separate folder gets created, all selected files, folders, the content within the sub folders gets uploaded to the chosen location."

If I have already managed my files in subfolders on the mobile device, why would I NOT want to create exactly this structure on the server ?

tobiasKaminsky:
Was requested today again via email.

seanlynch:
Just tried to do this myself and was quite surprised to find it isn't supported.

xXSTrikeXx:
refs: #285 #1328

AndyScherzinger:
@xXSTrikeXx nope ;) this is related to manual upload where atm you can only choose files but not a folder for upload.

QuelonaSec:
This feature is really mandatory. Along with a checkbox allowing full (initial) upload of a folder in the auto upload section.

Use case: automatic backup of a app folder (which is almost working thanks to the improvement in 2.0)

avanc:
+1 for this Feature Request.
I have a lot of data in my phone which I want to backup.

leonardsaers:
+1
I'm now trying next cloud and I'm happy with how it works. However uploading an entire directory from the mobile is a feature which I'm really missing.

subterfugium:
[...]

PR #4505, opened 2019-09-14
Allow whole folders to be uploaded to the server

robb4:
Long press the folder to select it
Refs: #184

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -52,6 +52,7 @@ import android.view.MenuInflater;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewTreeObserver;
+import android.util.Pair;
 
 import com.google.android.material.snackbar.Snackbar;
 import com.nextcloud.client.appinfo.AppInfo;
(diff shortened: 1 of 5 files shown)


tobiasKaminsky:
Instead of adding a non-visible long press action, I would simply use the multiple choice checkbox for this.

robb4:
Hmm, your screeshot looks a bit off, this is how it looks for me:
Screenshot_1568143119

Currently, for files, the checkbox gets activated by a single tap, but a tap on a folder navigates to it's content. It would make a bit more sense UX-wise to select both types on a tap, but that would interfere with the current behaviour. Is it acceptable to change the navigation option to a long tap?

tobiasKaminsky:
Sorry, I somehow was on the wrong branch.
Adding long press is something no user will directly know/understand, so I think we can omit it.
Having a checkbox for a folder should be enough @jancborchardt?

tobiasKaminsky:
@robb4 thanks for contributing to this long wished feature :tada:

tobiasKaminsky:
I created such a folder structure
- 1
- 1/2
- 1/2a
- 1/2/test.pdf

and select 1 to be uploaded, but "only":
- 1/2/test.pdf was uploaded
Empty 2a was not created.
I guess it as we only upload files.
I fear that people expect that also empty folders are created, which then unfortunately complicates it a lot…

robb4:
@tobiasKaminsky it does indeed only upload files, and it does not create empty folders. I'll take a look over this use-case when I'll have some free time.

jancborchardt:
Sorry for the late reply – yes the checkbox seems good to select folders! :+1: 

@robb4 any update on the other parts? :)

robb4:
@jancborchardtS adly I haven't had the time to work on this lately. 
I was initially hoping that creating empty folders too would be a quick and easy fix, but at a closer look it showed to be a bit more problematic. 
I tried piggybacking off the 'Create Folder' action available in the same activity, but that got messy quickly, so last time I was investigating different solutions to this. I could probably use some hints from someone who's more familiar to the architecture :)
I hope to be able to get back to this in the following weeks.

kilves:
@robb4 Are you still working on this? I accidently started this issue without even noticing that it had been under development for a few months already. I also completely missed the parameter on `requester.uploadNewFile()` that allows you to auto-create folders, so I ended up implementing my own folder creation. My redundant-but-useful solution might be of some help: https://github.com/kilves/android/commit/0297601f993f09912a2c14976536036c34626781 the folder creation code here could probably be adapted to also create empty folders with just a few lines of extra code.

There's also a $300 bounty on this in bountysource if you're interested in splitting it: https://www.bountysource.com/issues/36543232-upload-whole-folder

robb4:
@kilves  No, I haven't really worked on the core issue with my PR (the empty folders part), I only have a few local changes like not using long press to select and some other messed up code written while investigating.
To be honest, if your solution already ticks all the requirements go for it and raise a PR. If it's not complete and you think I can help you with something then sure, drop me a line on IRC or something and we can collaborate, but from what I've looked over your code I don't believe there's any benefit of merging the two solutions together.

AndyScherzinger:
superseeded by #7877

PR #4544, opened 2019-09-23
create a new activity for full screen slide show

hcvcastro:
Introduce openWebView function that it will create a new
RichDocumentsWebView activity and it will load the URL (SVG file
resource).

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/RichDocumentsWebView.java
+++ b/src/main/java/com/owncloud/android/ui/activity/RichDocumentsWebView.java
@@ -199,6 +199,10 @@ public class RichDocumentsWebView extends ExternalSiteWebView {
         if (TextUtils.isEmpty(url)) {
             new LoadUrlTask(this, getAccount()).execute(file.getLocalId());
         } else {
+            if (file.getFileName().equals("slideShow.svg")) {
+                hideLoading();
+            }
+
             webview.loadUrl(url);
         }
     }
@@ -274,6 +278,13 @@ public class RichDocumentsWebView extends ExternalSiteWebView {
         Intent intent = new Intent(this, ShareActivity.class);
         intent.putExtra(FileActivity.EXTRA_FILE, file);
         intent.putExtra(FileActivity.EXTRA_ACCOUNT, getAccount());
+        startActivity(intent);
+    }
+
+    private void openWebView(String url) {
+        Intent intent = new Intent(this, RichDocumentsWebView.class);
+        intent.putExtra(EXTRA_URL, url);
+        intent.putExtra(EXTRA_FILE, new OCFile("/slideShow.svg"));
         startActivity(intent);
     }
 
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@tobiasKaminsky can you take a look at this
@hcvcastro any particular reason why the file name is hard-coded?

tobiasKaminsky:
First, thanks for contributing to our app :+1: 
We normally follow this approach: create an issue with bug/enhancement, discuss how to fix it, create a PR.
Doing it in another way might lead to not accepting the PR.
In this case I fail to understand the use case.
Can you explain why you want to show a svg within a web view?
Normally every svg should be shown in our image system.

tobiasKaminsky:
I close this for now, if you update this and show a valid use case for it, we can discuss how to proceed.

hcvcastro:
Hi,
This is required for https://cgit.freedesktop.org/libreoffice/online/commit/?id=ca4fc297545d07ab9ab306925a431194d6be499f

The idea is to create a new activity for full-screen presentation for "Impress" app.
The commit has minimal changes to achieve the purpose, but you it can be improved

tobiasKaminsky:
@hcvcastro is there a docker snapshot I can test this with?

src/main/java/com/owncloud/android/ui/activity/RichDocumentsWebView.java
```diff
@@ -277,6 +281,13 @@ private void openShareDialog() {
         startActivity(intent);
     }
 
+    private void openWebView(String url) {
+        Intent intent = new Intent(this, RichDocumentsWebView.class);
```
tobiasKaminsky: It should not be necessary to reload the class completely from scratch, but instead just pass the the new url to the same instance.
Or is the intention of this to have it in a new separate activity? So that on back pressing the regular collabora (edit mode) is shown?

thebearon:
> @hcvcastro is there a docker snapshot I can test this with?

@tobiasKaminsky the following tag should yield you the latest snapshot that already contains the fix `collabora/code:4.2-snapshot`

juliushaertl:
Could you maybe also explain why the existing logic doesn't work in the mobile view?

https://cgit.freedesktop.org/libreoffice/online/commit/?id=ca4fc297545d07ab9ab306925a431194d6be499f
```
this._slideShow = L.DomUtil.create('iframe', 'leaflet-slideshow', this._map._container);
```

thebearon:
> Could you maybe also explain why the existing logic doesn't work in the mobile view?

My experience is that it downloads the slideshow file instead of showing it.

src/main/java/com/owncloud/android/ui/activity/RichDocumentsWebView.java
hcvcastro: Yes, the intention is a new separate activity, with the browser with no buttons or header to show a full-screen, and the browser will download the presentation.svg and run it, when the user press back the button it will close the presentation and back to regular collabora (edit mode)

hcvcastro:
>> Could you maybe also explain why the existing logic doesn't work in the mobile view?

There is a bug my list to fix, I need to investigate, it does not work in mobile view, but instead of i frame element maybe we can create a popup window or something related.

tobiasKaminsky:
Can you please update this and give me an instruction how to test this?

tobiasKaminsky:
Done via #5557

PR #4545, opened 2019-09-23
Fix for NPE caused by Log.d in ThumbnailsCacheManager

ezaquarii:
Fixes #4545 

### DIFF
--- a/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -74,6 +74,7 @@ import java.io.InputStream;
 import java.lang.ref.WeakReference;
 import java.net.URLEncoder;
 import java.util.List;
+import java.util.Locale;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -142,7 +143,7 @@ protected Void doInBackground(File... params) {
                         mThumbnailCache = new DiskLruImageCache(diskCacheDir, DISK_CACHE_SIZE, mCompressFormat,
                                                                 mCompressQuality);
                     } catch (Exception e) {
-                        Log_OC.d(TAG, e.getMessage());
+                        Log_OC.d(TAG, String.format(Locale.US, "Disk cache init failed: %s", e.getMessage()));
```
ezaquarii: I suspect message could be `null`, causing `Log.d` to throw inside native platform implementation.

ezaquarii: Actually, I managed to reproduce it by passing null to `Log.d`.

Strangely it does handle null under debugger, but crashes without debugger. :)

AndyScherzinger:
To me it looks like might solve the matter while we probably have such an issue in more than one place within the thumbs class.

ezaquarii:
Yeah, there is a library fix that solves this for good.

tobiasKaminsky:
Nice, doubled preventing npe :-)
Plus we get at least here the "init failed" part.

PR #4577, opened 2019-09-28
Fix NPE in PreviewMediaFragment

ezaquarii:
Fragment can be detached when async task finishes.
Detached fragment has no context and getString() fails with NPE.


ezaquarii:
It will be fixed by #4208, but this little fix should go into 3.8 IMO.

tobiasKaminsky:
Thanks for the finding!

PR #4600, opened 2019-10-02
Initial support for v2 library

tobiasKaminsky:
This is a very very first version of our new v2 library, which uses dav4jvm and okhttp.

This uses new GetUserInfoRemoteOperation with a NextcloudClient.
Please have first look at it, @AndyScherzinger @ezaquarii 
Same goes for https://github.com/nextcloud/android-library/pull/342

### DIFF
--- a/src/main/java/com/owncloud/android/operations/GetUserProfileOperation.java
+++ b/src/main/java/com/owncloud/android/operations/GetUserProfileOperation.java
@@ -19,9 +19,9 @@
  */
 package com.owncloud.android.operations;
 
-import android.accounts.Account;
 import android.accounts.AccountManager;
 
+import com.nextcloud.common.NextcloudClient;
 import com.owncloud.android.MainApp;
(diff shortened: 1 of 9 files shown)


src/main/java/com/owncloud/android/operations/GetUserProfileOperation.java
```diff
@@ -44,13 +44,28 @@
      *
      * Stored account is implicit in {@link #getStorageManager()}.
      *
-     * @return      Result of the operation. If successful, includes an instance of
+     * @return Result of the operation. If successful, includes an instance of
      *              {@link String} with the display name retrieved from the server.
      *              Call {@link RemoteOperationResult#getData()}.get(0) to get it.
      */
     @Override
+    @Deprecated
```
AndyScherzinger: If it is deprecated it should still work, so this just leads to an exception at runtime, so removing the method (leading to a compiler error) would be preferable imho.

What do you think @ezaquarii ?

Issue #4620, opened 2019-10-07
Android Nextcloud


Oceania2301:
### Actual behaviour
Crash at start

### Expected behaviour
All should Run

### Environment data
Android version:

Device model: ZenFone 3

Stock or customized system:

Nextcloud app version:

Nextcloud server version:

#### Nextcloud log (data/nextcloud.log)
```************ CAUSE OF ERROR ************

java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.get(ArrayList.java:437)
	at com.owncloud.android.ui.adapter.ActivityListAdapter.getItemViewType(ActivityListAdapter.java:392)
	at com.owncloud.android.ui.adapter.ActivityListAdapter.isHeader(ActivityListAdapter.java:461)
	at com.owncloud.android.ui.adapter.ActivityListAdapter.getHeaderPositionForItem(ActivityListAdapter.java:438)
	at com.owncloud.android.ui.activities.StickyHeaderItemDecoration.getHeaderViewForItem(StickyHeaderItemDecoration.java:99)
	at com.owncloud.android.ui.activities.StickyHeaderItemDecoration.onDrawOver(StickyHeaderItemDecoration.java:55)
	at androidx.recyclerview.widget.RecyclerView.draw(RecyclerView.java:4223)
	at android.view.View.updateDisplayListIfDirty(View.java:18073)
	at android.view.View.draw(View.java:18851)
	at android.view.ViewGroup.drawChild(ViewGroup.java:4214)
	at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4000)
	at android.view.View.draw(View.java:19126)
	at android.view.View.updateDisplayListIfDirty(View.java:18073)
	at android.view.View.draw(View.java:18851)
	at android.view.ViewGroup.drawChild(ViewGroup.java:4214)
	at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4000)
	at android.view.View.updateDisplayListIfDirty(View.java:18064)
	at android.view.View.draw(View.java:18851)
	at android.view.ViewGroup.drawChild(ViewGroup.java:4214)
	at android.view.ViewGroup.dispatchDraw(ViewGroup.java:4000)
	at android.view.View.updateDisplayListIfDirty(View.java:18064)
	at android.view.View.draw(View.java:18851)
	at android.view.ViewGroup.drawChild(ViewGroup.java:4214)
[...]

PR #4629, opened 2019-10-08
Remove sticky headers from activities list view

ezaquarii:
This will prevent application from crashing on adapter update.


ezaquarii:
@AndyScherzinger @tobiasKaminsky This is a low-risk mitigation for #4620 suitable for 3.8 branch.

How to address this in 3.9.x is another story.

AndyScherzinger:
Wouldn't we then simply loose the headers?

We had headers before, made them sticky with the last release and now temporarily remove them to fix the crash?

ezaquarii:
Headers are still there, just not sticky.

AndyScherzinger:
Then I am fine with the quick fix for 3.8.1 while we should then have another issue ticket scheduled for 3.9.0 since we should get this working.

tobiasKaminsky:
:+1: for the quick fix.
:tada: as always many thanks for your help @ezaquarii

PR #4630, opened 2019-10-09
Note about bug fix release schedule

tobiasKaminsky:
Added that bug fix release is 4 weeks after stable release.


tobiasKaminsky:
Nice note about security fixes. Absolutely correct!

AndyScherzinger:
Yeah, I though twe better add this (mmight be obvious and implict for us) so that people don't imply that we would wait for 4 weeks before we release critical patches.

PR #4659, opened 2019-10-11
3.8.1 release notes

AndyScherzinger:


tobiasKaminsky:
Nice!
I should really automate this.

tobiasKaminsky:
Do you think it is really needed to have it also in stable-3.8 branch?
There we have the changelogs

AndyScherzinger:
> Do you think it is really needed to have it also in stable-3.8 branch?

Not mandatory I guess, but still nice in case someone is looking at the stable branch

> I should really automate this.

Sure, but please do it after thet Hacktoberfest 😉

tobiasKaminsky:
I will merge this as soon as 3.8.1 is released

PR #4677, opened 2019-10-13
Deprecate setAccount logic and remove accounWasSet flag

ezaquarii:

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
@@ -38,11 +38,6 @@ public abstract class BaseActivity extends AppCompatActivity implements Injectab
     private OCCapability mCapabilities;
 
     /**
-     * Flag to signal when the value of mAccount was set.
-     */
-    protected boolean mAccountWasSet;
-
-    /**
      * Flag to signal when the value of mAccount was restored from a saved state.
      */
     protected boolean mAccountWasRestored;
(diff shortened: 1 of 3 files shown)


src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
```diff
@@ -37,11 +37,6 @@
      */
     private OCCapability mCapabilities;
 
-    /**
-     * Flag to signal when the value of mAccount was set.
-     */
-    protected boolean mAccountWasSet;
```
ezaquarii: This can boils down to `account == null`, so it's duplicated piece of information.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -450,10 +451,8 @@ protected void onAccountSet(boolean stateWasRecovered) {
             }
             setFile(file);
 
-            if (mAccountWasSet) {
```
ezaquarii: This boils down to `if (account != null)`, which is checked at the function entry. It is safe to omit this condition.

src/main/java/com/owncloud/android/ui/activity/UploadListActivity.java
```diff
@@ -382,20 +392,6 @@ public void onReceive(Context context, Intent intent) {
         }
     }
 
-    /**
-     * Called when the ownCloud {@link Account} associated to the Activity was just updated.
-     */
-    @Override
-    protected void onAccountSet(boolean stateWasRecovered) {
-        super.onAccountSet(stateWasRecovered);
```
ezaquarii: This code was called in `onStart()` if `account != null`. 2nd flag is not used, so we can eliminate `onAccountSet` call.

tobiasKaminsky:
@ezaquarii please set it always to "1. developing" or "2. to review".
There is a script that shows me what I should review, and without any labelling it ignores those PRs.
Also other contributor then do not know the state.

tobiasKaminsky:
Is this ready?

ezaquarii:
Yes. I tested it and it worked ok for me with account switching and deletion from system settings.

ezaquarii:
Blocking #4685 and #4686.

PR #4685, opened 2019-10-15
Remove onAccountSet from UploadPathActivity 

ezaquarii:


ezaquarii:
@tobiasKaminsky @AndyScherzinger I can't find a way to launch this activity. Is it even used? Maybe we can just delete if if we don't use it.

ezaquarii:
@tobiasKaminsky I can't help the coverage decrease here.

tobiasKaminsky:
Obsolete.

PR #4686, opened 2019-10-15
Remove onAccountSet from image and video preview activities

ezaquarii:


ezaquarii:
@tobiasKaminsky I can't help the coverage decrease here.

PR #4700, opened 2019-10-17
Remove onAccountSet from ReceiveExternalFilesActivity

ezaquarii:
1. `onAccountSet` logic inlined in `onStart` and `changeAccount`
2. call to `onAccountSet` eliminated

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
@@ -231,12 +231,13 @@ public class ReceiveExternalFilesActivity extends FileActivity
 
     public void changeAccount(Account account) {
         setAccount(account, false);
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -231,12 +231,13 @@ private Activity getActivity() {
...
     public void changeAccount(Account account) {
         setAccount(account, false);
-        onAccountSet(mAccountWasRestored);
+        initTargetFolder();
+        populateDirectoryList();
     }
 
     @Override
-    protected void onAccountSet(boolean stateWasRecovered) {
-        super.onAccountSet(mAccountWasRestored);
```
ezaquarii: This flag from `BaseActivity` is not modified in the subclass so it's safe to remove it.

PR #4692, opened 2019-10-16
Markdown rendering

AndyScherzinger:
initial rendering for markdown in the TextPreview fragment

Resolves #3302

Issue #3302, opened 2018-11-30
Preview markdown files


jangmarker:
### Is your feature request related to a problem? Please describe.
Markdown files are an easy way of structuring some quick notes.
Although their plain textural representation is already well structured, adapted rendering for phones would allow these files to be even more useful.
That way one could take down notes on the computer, sync them with Nextcloud to the phone and have an easy to read representation on the phone.

### Describe the solution you'd like
The android app is currently displaying text files within the app. It would be nice if it could render Markdown files in preview mode, i.e. similar to HTML.

### Describe alternatives you've considered
There are apps on the Android app store that allow editing and preview of Markdown files, however using them for preview is rather cumbersome compared to previewing Markdown files within the Nextcloud Android app.

AndyScherzinger:
@tobiasKaminsky I am fine with the assessement that this is a feature while I'd like to disuss it and it's implications so I'd also like to get @stefan-niedermann's input on adding "just the rendering".

* adding markdown rendering will likely lead to the next feature request: editing capability and then we basically copy the notes client 
* I wouldn't want to re-invent the wheel or create a competitor here for no good reason since the notes client is part of the eco system and a good prodcut where I don't see a reason to attack that

So to be discussed (from my point of view) and rendering would be fine with me (scope of this feature request) while editing capabilities is somethign I'd rather not add to the files client. (same arguments would also apply for deck, bookmarks, you name it)

thomas-mc-work:
> There's no smoke without fire

I think write support would still be different. Notes are notes and can live in the notes app, which provides more specialized features for that task. But if there is much demand for simple read/write support in the client – why not?

But that demand hasn't been claimed yet. I'd also love to see the markdown rendering feature in the android app.

Also, there is no right to deliver all feature requests. So personally I wouldn't care too much about this virtual threat.

Another point for possible subsequent feature requests for write support: There isn't even write support for simple plain text files. So then write support for markdown files is much further away.

AndyScherzinger:
PR adding initial support is #4692

PR #4701, opened 2019-10-17
Markdown preview syntax highlighting

AndyScherzinger:
adds syntax highlighting capabilities within the markdown rendering

to be merged _after_ #4692

extends #3302 by adding syntax highlighting


gradle.properties
```diff
@@ -5,4 +5,4 @@ NC_TEST_SERVER_PASSWORD=test
 android.enableJetifier=true
 android.useAndroidX=true
 android.debug.obsoleteApi=true
-
+org.gradle.jvmargs=-Xmx4g -XX:MaxPermSize=2048m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
tobiasKaminsky: This should not be in public grade.properties, but only in your local one.

AndyScherzinger:
Hi @stefan-niedermann,

in generall, yes, that would be the prefered way. In this case the situation is slightly diffrent, so IMHO we rather won't move for such a solution due to the next iterations of this feature.

Since Nextcloud now comes with Nextcoud Text, the client will implement support for this (similar to the collabora support). So Markdown suppurt read/write will come with supporting Nextcloud Text (end of this year, beginning of next year). So this is a rather temporary solution to suppoort markdown previewing for the time being. Other than that the advantage of Markwon (for preview-only scenrarios) is that is works natively as in rendering spannedText based on the  markdown.

So _if_ Nextcloud Text wouldn't be there I'd also vouched for using Markor 👍

gradle.properties
AndyScherzinger: removed 👍

stefan-niedermann:
Did i understand you right, that the Files app will support editing markdown files in tge future based on a webview-solution (instead of a native implememtation)?

AndyScherzinger:
> support editing markdown files in the future based on a webview-solution (instead of a native implememtation)

That is what I understood from a discussion with @tobiasKaminsky while it'll rather be a _"any-text-based"_ file editing. It is than based on a webview _and_ will of course online work when being online.

stefan-niedermann:
Thank you for clarification :)

spotbugs-filter.xml
```diff
@@ -33,6 +33,11 @@
             <Class name="~.*\.DataBinderMapperImpl"/>
         </Or>
     </Match>
+	
+	<!-- Third-party library code -->
+	<Match>
+	    <Package name="~io\.noties\..*" />
```
AndyScherzinger: @tobiasKaminsky mind having a look? I just don't get this exclusion right :/

AndyScherzinger: @tobiasKaminsky nevermind, got it now :)

build.gradle
```diff
@@ -327,6 +329,10 @@ dependencies {
     implementation ("io.noties.markwon:html:$markwonVersion")
     implementation ("io.noties.markwon:linkify:$markwonVersion")
 
+    implementation ("io.noties.markwon:syntax-highlight:$markwonVersion")
+    implementation ("io.noties:prism4j:${prismVersion}")
```
tobiasKaminsky: Just for consistent reasons, I guess we do not need { } here, or?

build.gradle
```diff
@@ -327,6 +329,10 @@ dependencies {
...
+    kapt "io.noties:prism4j-bundler:${prismVersion}"
```
tobiasKaminsky: Same here

tobiasKaminsky:
@AndyScherzinger 
tested and works :tada: 
Can you have a look at my two (small) questions?

build.gradle
AndyScherzinger: Done https://github.com/nextcloud/android/pull/4701/commits/f5ebba9efffef00912302fd493538767da424de0

tobiasKaminsky:
Two spotbugs are this:


  | com.owncloud.android.ui.preview.MarkwonGrammarLocator$1.name() may return null, but is declared @Nonnull
-- | --
  | Bug type NP_NONNULL_RETURN_VIOLATION (click for details) In class com.owncloud.android.ui.preview.MarkwonGrammarLocator$1In method com.owncloud.android.ui.preview.MarkwonGrammarLocator$1.name()At MarkwonGrammarLocator.java:[line 47]
NP | com.owncloud.android.ui.preview.MarkwonGrammarLocator$1.tokens() may return null, but is declared @Nonnull
  | Bug type NP_NONNULL_RETURN_VIOLATION (click for details) In class com.owncloud.android.ui.preview.MarkwonGrammarLocator$1In method com.owncloud.android.ui.preview.MarkwonGrammarLocator$1.tokens()At MarkwonGrammarLocator.java:[line 53]


There is nothing we can do, so I will increase spotbugs.

AndyScherzinger:
Yeah, I tried to put the file on the exclude list but failed to get the regex right 😞

PR #4710, opened 2019-10-20
Eliminate account status flags from BaseActivity

ezaquarii:
onAccountSet(stateWasRecovered) is no longer used anywhere,
so the flag mAccountWasRestored can be safely removed.

This intermediary step opens path to eliminating
out of band calls to onAccountSet().


ezaquarii:

PR #4712, opened 2019-10-21
Fix possible IndexOutOfBoundsException if returned array is empty

tobiasKaminsky:
Found via google dev console.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/asynctasks/FetchRemoteFileTask.java
+++ b/src/main/java/com/owncloud/android/ui/asynctasks/FetchRemoteFileTask.java
@@ -26,6 +26,7 @@ import android.os.AsyncTask;
 import android.text.TextUtils;
 
 import com.owncloud.android.MainApp;
+import com.owncloud.android.R;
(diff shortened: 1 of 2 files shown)


AndyScherzinger:
Perfectly fine, only thing is with the string.
We usually don't use exclamation marks throughout NC so it should be dropped.

tobiasKaminsky:
> We usually don't use exclamation marks throughout NC so it should be dropped.

I found several ones, e.g:
<string name="account_not_found">Account not found!</string>
<string name="uploads_view_upload_status_virus_detected">Virus detected. Upload cannot be completed!</string>
<string name="file_version_restored_error">Error restoring file version!</string>
    <string name="trashbin_loading_failed">Loading trashbin failed!</string>
    <string name="trashbin_file_not_deleted">File %1$s could not be deleted!</string>
    <string name="trashbin_file_not_restored">File %1$s could not be restored!</string>
    <string name="trashbin_not_emptied">Files could not be deleted permanently!</string>

All have in common, that they are severe warnings or errors.

PR #4716, opened 2019-10-22
DavX5: support login flow

### LINKED ISSUE #4708: Use Login Flow with DAVx⁵
DAVx⁵ supports the Nextcloud Login Flow since 2.6-beta1. The intent is documented here: https://www.davx5.com/manual/technical_information.html#api-integration (updated recently so you may have to refresh the browser cache)

I suggest to call the new Intent from the Nextcloud app so that Login Flow is used.

Compatibility:

* old Nextcloud app / new DAVx⁵ versions: no problems, because the URL/username/password extras are still actively supported
* old DAVx⁵ / new Nextcloud app versions: You could set both the url, username, password extras and the data URI + loginFlow/davPath extras. In this case, new DAVx⁵ versions will see the loginFlow extra and switch to Login Flow, while old DAVx⁵ versions will still use the URL/username extras.

If you need any assistance or anything else, please just write here or there: https://forums.bitfire.at/topic/2121/please-support-nextclouds-singlesignon-sso

tobiasKaminsky:
Resolves #4708 

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
@@ -62,6 +62,7 @@ import com.nextcloud.client.preferences.AppPreferencesImpl;
 import com.owncloud.android.BuildConfig;
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
+import com.owncloud.android.authentication.AuthenticatorActivity;
 import com.owncloud.android.authentication.PassCodeManager;
 import com.owncloud.android.datamodel.ArbitraryDataProvider;
 import com.owncloud.android.datamodel.ExternalLinksProvider;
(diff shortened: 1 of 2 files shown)


src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
```diff
@@ -750,6 +751,11 @@ private void launchDavDroidLogin() {
                 davDroidLoginIntent.putExtra("url", serverBaseUri.toString() + DAV_PATH);
             }
             davDroidLoginIntent.putExtra("username", UserAccountManager.getUsername(account));
+
+            davDroidLoginIntent.putExtra("loginFlow", 1);
```
AndyScherzinger: 1 should be a named constant 😊

AndyScherzinger:
Just one spotbug to go @tobiasKaminsky 😉

PR #4725, opened 2019-10-24
SSO: return response headers

tobiasKaminsky:


src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
```diff
@@ -157,6 +196,18 @@ public int read() {
...
 
+    private <T extends Serializable> ByteArrayInputStream serializeNewObjectToInputStream(T obj,
+                                                                                          String headers)
+        throws IOException {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        ObjectOutputStream oos = new ObjectOutputStream(baos);
+        oos.writeObject(obj);
+        oos.writeObject(headers);
+        oos.flush();
+        oos.close();
```
AndyScherzinger: stream(s) aren't closed properly in finally block(s)

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
```diff
@@ -112,6 +115,42 @@ public InputStreamBinder(Context context, UserAccountManager accountManager) {
...
+        try {
+            // Start request and catch exceptions
+            NextcloudRequest request = deserializeObjectAndCloseStream(is);
+            response = processNewRequest(request, requestBodyInputStream);
+        } catch (Exception e) {
+            Log_OC.e(TAG, "Error during Nextcloud request", e);
+            exception = e;
+        }
+
+        try {
```
AndyScherzinger: does this block make sense in case `exception` is null?

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
```diff
@@ -350,4 +466,67 @@ private static boolean isEqual(byte[] a, byte[] b) {
...
+    private class Response {
+        private InputStream body;
+        private Header[] headers;
+
+        public Response() {
+            headers = new Header[0];
+            body = new InputStream() {
+                @Override
+                public int read() {
+                    return 0;
```
AndyScherzinger: Magic number, should be a named constant

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
```diff
@@ -350,4 +466,67 @@ private static boolean isEqual(byte[] a, byte[] b) {
...
+            };
+        }
+
+        public Response(InputStream inputStream, Header[] headers) {
+            this.body = inputStream;
+            this.headers = headers;
+        }
+
+        public String getPlainHeadersString() {
+            ArrayList<PlainHeader> arrayList = new ArrayList<>();
```
AndyScherzinger: Should be a `List`

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
```diff
@@ -350,4 +466,67 @@ private static boolean isEqual(byte[] a, byte[] b) {
...
+
+            for (Header header : headers) {
+                arrayList.add(new PlainHeader(header.getName(), header.getValue()));
+            }
+
+            Gson gson = new Gson();
+            return gson.toJson(arrayList);
+        }
+
+        public InputStream getBody() {
```
AndyScherzinger: can ge generated via Lombok

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
```diff
@@ -350,4 +466,67 @@ private static boolean isEqual(byte[] a, byte[] b) {
...
+            return body;
+        }
+    }
+
+    private class PlainHeader implements Serializable {
```
AndyScherzinger: getters can be generated via Lombok

AndyScherzinger: Classes implementing Serializable should set a serialVersionUID

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
```diff
@@ -350,4 +466,67 @@ private static boolean isEqual(byte[] a, byte[] b) {
...
+            headers = new Header[0];
+            body = new InputStream() {
+                @Override
+                public int read() {
+                    return 0;
+                }
+            };
+        }
+
+        public Response(InputStream inputStream, Header[] headers) {
```
AndyScherzinger: Consider using varargs for methods or constructors which take an array the last parameter.

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
tobiasKaminsky: We always write excecption stream, even if it s null. With this approach we ensure that the "parameter" order and numbers are the same.

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
AndyScherzinger: Try catch finally for oos

src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
```diff
@@ -157,6 +196,18 @@ public int read() {
...
+    private <T extends Serializable> ByteArrayInputStream serializeObjectToInputStreamV2(T obj,
+                                                                                         String headers)
+        throws IOException {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        ObjectOutputStream oos = new ObjectOutputStream(baos);
+        oos.writeObject(obj);
+        oos.writeObject(headers);
+        oos.flush();
+        oos.close();
+        return new ByteArrayInputStream(baos.toByteArray());
```
AndyScherzinger: baos would then needed to be closed before returning the result (so the result needs to be saved to a variable and send afterwards.

PR #4741, opened 2019-10-28
Brand feature to hide provider buttons but keep slider

Shagequi:
Why hide the slider step if you want to just hide the buttons through `show_provider_or_own_installation` ? 

Here is a suggestion


tobiasKaminsky:
Please open up first an issue to discuss this and not directly a PR.
As said in the other PR, if you want to do changes for your own branded clients this is totally fine, but this does not mean that your expectations fit the needs of this official client.

PR #4748, opened 2019-10-29
re-add CONTENT_TYPE_SEARCH

tobiasKaminsky:
Successor of https://github.com/nextcloud/android/pull/4233


tobiasKaminsky:
This is currently not yet enabled, question is, if we should? :thinking:

AndyScherzinger:
I don't know what it would be used for. What I find "strange" is that the content type is relevant for the recently added menu item 🤔

PR #4753, opened 2019-10-29
Added brand feature to hide notifications from Drawer in setup

Shagequi:
When push server is unavailable, it's useless to have a "notifications tab", here is an enhancement :)

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -381,6 +381,7 @@ public abstract class DrawerActivity extends ToolbarActivity
         DrawerMenuUtil.removeMenuItem(menu, R.id.nav_shared, !getResources().getBoolean(R.bool.shared_enabled));
         DrawerMenuUtil.removeMenuItem(menu, R.id.nav_contacts, !getResources().getBoolean(R.bool.contacts_backup)
                 || !getResources().getBoolean(R.bool.show_drawer_contacts_backup));
+        DrawerMenuUtil.removeMenuItem(menu, R.id.nav_notifications, !getResources().getBoolean(R.bool.notifications_enabled));
 
         DrawerMenuUtil.removeMenuItem(menu, R.id.nav_synced_folders,
                 getResources().getBoolean(R.bool.syncedFolder_light));
--- a/src/main/res/values/setup.xml
+++ b/src/main/res/values/setup.xml
@@ -109,6 +109,8 @@
     <!-- Files becomes Home -->
     <bool name="use_home">false</bool>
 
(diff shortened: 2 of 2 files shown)


Shagequi:
Note : you should replace by "true" by default (misstake from me ^^')

AndyScherzinger:
Why should you hide the notifications from the drawer? The have nothing to do with push notifications...

Shagequi:
If you don't have a push server available, the notifications Activity (and Drawer tab) is useless because you get an error.

It would be better if we could hide the notifications tab if we don't have push server available.

tobiasKaminsky:
if you do not have a push server, you can still see a list of notifications.
You only do not  get an info in sysbar, if a new nofification is generated on server.

tobiasKaminsky:
> it is a brand feature... to allow companies to hide some un-needed things

If you want to brand your own client this is totally fine, but then these changes should not be part of official clients (as those might/will have different requirements).

Shagequi:
It's a feature that I added on my forked client, and after having made an overview, I thought I would give some of these changes to potential companies that would like to save analysis time by simply adding settings in `setup`.

(This is free of charge, it does not imply any technical debt, nor any modification in the pure functioning of your code).

If you don't want this type of feature, so i'll only create pull requests to fix bugs and enhance this app :)

AndyScherzinger:
@Shagequi the only thing your PR is doing is to hide the menu item, nothing else. Still if the server has access to the push server (which you'd need to block via firewall on the server) you'd still get push notification which the app would open and like @tobiasKaminsky said even when you don't have push you'd still have notifications. Else this would be true for any feature of the app and feature toggles would actually have to be (regression) tested.

We sure do appreciate the effort while feature changinig PRs should first be discussed in an issue to prevent efforts made being declined (which isn't the case here since you use this change anyways of course).

Shagequi:
Ok ! I see what you wanted to explain.
So, I'll convert this one to an issue, and (if I have time) work on a better pull request about this.

AndyScherzinger:
Just out of curiosity, why would you wan tto hide the notifications on the client? You'd still see them being logged into the web UI.

Shagequi:
Here is the issue : https://github.com/nextcloud/android/issues/4759

Shagequi:
@AndyScherzinger We're a hosting company, and we create a Drive solution with our own API but with NextCloud clients (Android, iOS, and Desktop).

So we have specific needs, and we create what we need to setup that project, and we would like to share our work on Clients with other companies that have the same type of project.

AndyScherzinger:
> the push server

is run by Nextcloud themselves, except in case you actually modified the server too.

Shagequi:
The push server is customized too

AndyScherzinger:
> The push server is customized too

Ah, now I get the push notification part (its motivation)

Issue #1150, opened 2017-06-28
Need new button for complete re-scan-re-upload of folders selected for synchronization


thsnielsen:
### Steps to reproduce
1. While having folders checked for synchronization - unknown problems arise, and for some odd reason synchronization stops (as i have seen a number of times) then normal troubleshooting starts - are there blocked files on the server - delete them on server - these files are not re-uploaded once back to normal and next time the android client syncs the folder from where the picture or file came from.
You may even chose to delete the account on the device, and recreate it - then no more files older that the date of the creation of the account are uploaded - no matter if the files are on the server or not
2. Normally synchronization picks up again if you manually tags a file for upload to server, but all new files arrived since the stop of sync until manual upload will not never synced.
3.i am sure there are other scenarios, where files on the device are left un-uploaded

### Expected behaviour
At next scan the non-uploaded files should be detected, and uploaded, but for the case where a file was purposely deleted via the web interface on the server, it creates a conflict - should this file be re-uploaded or not - rather not i would say. But sometimes you just wish to make sure, that the complete folder marked for sync on the phone IS on the %é_è-'çàçe(bloody) server.
Therefore the need for a "redo complete deep dive sync of folders marked for sync" "with/without overwrite" of existing files (the overwrite could be interesting in case of corruption on the server (sd card about to hang up, cryptolock etc)

### Actual behaviour
not possible to get missing files onto the server once sync has decided to omit them, unless finding out yourself which files are missing, and manually tagging them for upload

### Server configuration

**Operating system**:Raspbian 4.9.24-v7+ #993 SMP Wed Apr 26 18:01:23 BST 2017 armv7l GNU/Linux

**Web server:**
apt-cache show nginx      
Package: nginx
Version: 1.10.3-1
Installed-Size: 91

**Database:**
Package: mariadb-server-core-10.1
Source: mariadb-10.1
Version: 10.1.22-3
Architecture: armhf

**PHP version:**
php -v
PHP 7.0.19-1 (cli) (built: May 11 2017 14:04:47) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.19-1, Copyright (c) 1999-2017, by Zend Technologies

**Server version:** (see your admin page)
12.0.0
Here connected remotely i discovered my USB storage is down, so i have to provide the rest of the details later, when i have cooled down my storage (bloody heatwave)- sorry about that

[...]

PR #4787, opened 2019-11-03
Make synced folder init/enable date persistent

ArisuOngaku:
As seen in #1150 :

When the app crashes, is not run for some reason, and that files are added in auto upload folders during this downtime, even if the folders were enabled before and successfully synced some files, there are some that get ignored because the sync start time of an auto upload folder is reset when the application restarts.

This pull request fixes this bug/undesirable behavior.

### DIFF
--- a/src/main/java/com/owncloud/android/datamodel/SyncedFolder.java
+++ b/src/main/java/com/owncloud/android/datamodel/SyncedFolder.java
@@ -30,23 +30,23 @@ import lombok.Setter;
 /**
  * Synced folder entity containing all information per synced folder.
  */
-@Getter
-@Setter
 @AllArgsConstructor
 public class SyncedFolder implements Serializable, Cloneable {
     public static final long UNPERSISTED_ID = Long.MIN_VALUE;
+    public static final long EMPTY_ENABLED_TIMESTAMP_MS = -1;
     private static final long serialVersionUID = -793476118299906429L;
 
-    private long id = UNPERSISTED_ID;
-    private String localPath;
-    private String remotePath;
-    private Boolean wifiOnly;
-    private Boolean chargingOnly;
-    private Boolean subfolderByDate;
-    private String account;
-    private Integer uploadAction;
-    private boolean enabled;
-    private MediaFolderType type;
+    @Getter @Setter private long id;
+    @Getter @Setter private String localPath;
+    @Getter @Setter private String remotePath;
+    @Getter @Setter private Boolean wifiOnly;
+    @Getter @Setter private Boolean chargingOnly;
+    @Getter @Setter private Boolean subfolderByDate;
+    @Getter @Setter private String account;
+    @Getter @Setter private Integer uploadAction;
+    @Getter private boolean enabled;
+    @Getter private long enabledTimestampMs;
+    @Getter @Setter private MediaFolderType type;
 
     /**
      * constructor for new, to be persisted entity.
@@ -59,11 +59,25 @@ public class SyncedFolder implements Serializable, Cloneable {
      * @param account         the account owning the synced folder
      * @param uploadAction    the action to be done after the upload
      * @param enabled         flag if synced folder config is active
+     * @param timestampMs     the current timestamp in milliseconds
      * @param type            the type of the folder
      */
     public SyncedFolder(String localPath, String remotePath, Boolean wifiOnly, Boolean chargingOnly,
                         Boolean subfolderByDate, String account, Integer uploadAction, Boolean enabled,
-                        MediaFolderType type) {
+                        long timestampMs, MediaFolderType type) {
+        this(UNPERSISTED_ID, localPath, remotePath, wifiOnly, chargingOnly, subfolderByDate, account, uploadAction,
+             enabled, timestampMs, type);
+    }
+
+    /**
+     * constructor for wrapping existing folders.
+     *
+     * @param id id
+     */
+    protected SyncedFolder(long id, String localPath, String remotePath, Boolean wifiOnly, Boolean chargingOnly,
+                           Boolean subfolderByDate, String account, Integer uploadAction, Boolean enabled,
+                           long timestampMs, MediaFolderType type) {
+        this.id = id;
         this.localPath = localPath;
         this.remotePath = remotePath;
         this.wifiOnly = wifiOnly;
@@ -71,8 +85,16 @@ public class SyncedFolder implements Serializable, Cloneable {
         this.subfolderByDate = subfolderByDate;
         this.account = account;
         this.uploadAction = uploadAction;
-        this.enabled = enabled;
+        this.setEnabled(enabled, timestampMs);
         this.type = type;
+    }
+
+    /**
+     * @param timestampMs the current timestamp in milliseconds
+     */
+    public void setEnabled(boolean enabled, long timestampMs) {
+        this.enabled = enabled;
+        this.enabledTimestampMs = enabled ? timestampMs : EMPTY_ENABLED_TIMESTAMP_MS;
     }
 
     public Object clone() {
--- a/src/main/java/com/owncloud/android/datamodel/SyncedFolderDisplayItem.java
+++ b/src/main/java/com/owncloud/android/datamodel/SyncedFolderDisplayItem.java
@@ -56,9 +56,11 @@ public class SyncedFolderDisplayItem extends SyncedFolder {
      */
     public SyncedFolderDisplayItem(long id, String localPath, String remotePath, Boolean wifiOnly, Boolean chargingOnly,
                                    Boolean subfolderByDate, String account, Integer uploadAction, Boolean enabled,
-                                   List<String> filePaths, String folderName, long numberOfFiles, MediaFolderType type)
+                                   long timestampMs, List<String> filePaths, String folderName, long numberOfFiles,
+                                   MediaFolderType type)
     {
-        super(id, localPath, remotePath, wifiOnly, chargingOnly, subfolderByDate, account, uploadAction, enabled, type);
+        super(id, localPath, remotePath, wifiOnly, chargingOnly, subfolderByDate, account, uploadAction, enabled,
+              timestampMs, type);
         this.filePaths = filePaths;
         this.folderName = folderName;
         this.numberOfFiles = numberOfFiles;
@@ -66,8 +68,9 @@ public class SyncedFolderDisplayItem extends SyncedFolder {
 
     public SyncedFolderDisplayItem(long id, String localPath, String remotePath, Boolean wifiOnly, Boolean chargingOnly,
                                    Boolean subfolderByDate, String account, Integer uploadAction, Boolean enabled,
-                                   String folderName, MediaFolderType type) {
-        super(id, localPath, remotePath, wifiOnly, chargingOnly, subfolderByDate, account, uploadAction, enabled, type);
+                                   long timestampMs, String folderName, MediaFolderType type) {
+        super(id, localPath, remotePath, wifiOnly, chargingOnly, subfolderByDate, account, uploadAction, enabled,
+              timestampMs, type);
         this.folderName = folderName;
     }
 }
--- a/src/main/java/com/owncloud/android/datamodel/SyncedFolderProvider.java
+++ b/src/main/java/com/owncloud/android/datamodel/SyncedFolderProvider.java
@@ -27,6 +27,7 @@ import android.content.Context;
 import android.database.Cursor;
 import android.net.Uri;
 
+import com.nextcloud.client.core.Clock;
 import com.nextcloud.client.preferences.AppPreferences;
 import com.nextcloud.client.preferences.AppPreferencesImpl;
 import com.owncloud.android.db.ProviderMeta;
@@ -36,6 +37,8 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Observable;
+
+import javax.inject.Inject;
 
 import androidx.annotation.NonNull;
 
@@ -49,6 +52,8 @@ public class SyncedFolderProvider extends Observable {
 
     private ContentResolver mContentResolver;
     private AppPreferences preferences;
+
+    @Inject protected Clock clock;
 
     /**
      * constructor.
@@ -162,7 +167,7 @@ public class SyncedFolderProvider extends Observable {
                 // read sync folder object and update
                 SyncedFolder syncedFolder = createSyncedFolderFromCursor(cursor);
 
-                syncedFolder.setEnabled(enabled);
+                syncedFolder.setEnabled(enabled, clock.getCurrentTime());
 
                 // update sync folder object in db
                 result = updateSyncFolder(syncedFolder);
@@ -347,11 +352,13 @@ public class SyncedFolderProvider extends Observable {
                     ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_UPLOAD_ACTION));
             Boolean enabled = cursor.getInt(cursor.getColumnIndex(
                     ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_ENABLED)) == 1;
+            long enabledTimestampMs = cursor.getLong(cursor.getColumnIndex(
+                    ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_ENABLED_TIMESTAMP_MS));
             MediaFolderType type = MediaFolderType.getById(cursor.getInt(cursor.getColumnIndex(
                     ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_TYPE)));
 
             syncedFolder = new SyncedFolder(id, localPath, remotePath, wifiOnly, chargingOnly, subfolderByDate,
-                    accountName, uploadAction, enabled, type);
+                    accountName, uploadAction, enabled, enabledTimestampMs, type);
         }
         return syncedFolder;
     }
@@ -370,6 +377,7 @@ public class SyncedFolderProvider extends Observable {
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_WIFI_ONLY, syncedFolder.getWifiOnly());
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_CHARGING_ONLY, syncedFolder.getChargingOnly());
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_ENABLED, syncedFolder.isEnabled());
+        cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_ENABLED_TIMESTAMP_MS, syncedFolder.getEnabledTimestampMs());
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_SUBFOLDER_BY_DATE, syncedFolder.getSubfolderByDate());
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_ACCOUNT, syncedFolder.getAccount());
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_UPLOAD_ACTION, syncedFolder.getUploadAction());
--- a/src/main/java/com/owncloud/android/datamodel/SyncedFolderProvider.java
+++ b/src/main/java/com/owncloud/android/datamodel/SyncedFolderProvider.java
@@ -38,8 +38,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Observable;
 
-import javax.inject.Inject;
-
 import androidx.annotation.NonNull;
 
 import static com.owncloud.android.datamodel.OCFile.PATH_SEPARATOR;
@@ -50,22 +48,22 @@ import static com.owncloud.android.datamodel.OCFile.PATH_SEPARATOR;
 public class SyncedFolderProvider extends Observable {
     static private final String TAG = SyncedFolderProvider.class.getSimpleName();
 
-    private ContentResolver mContentResolver;
-    private AppPreferences preferences;
-
-    @Inject protected Clock clock;
+    private final ContentResolver mContentResolver;
+    private final AppPreferences preferences;
+    private final Clock clock;
 
     /**
      * constructor.
      *
      * @param contentResolver the ContentResolver to work with.
      */
-    public SyncedFolderProvider(ContentResolver contentResolver, AppPreferences preferences) {
+    public SyncedFolderProvider(ContentResolver contentResolver, AppPreferences preferences, Clock clock) {
         if (contentResolver == null) {
             throw new IllegalArgumentException("Cannot create an instance with a NULL contentResolver");
         }
         mContentResolver = contentResolver;
         this.preferences = preferences;
+        this.clock = clock;
     }
 
     /**
--- a/src/main/java/com/owncloud/android/db/ProviderMeta.java
+++ b/src/main/java/com/owncloud/android/db/ProviderMeta.java
@@ -31,7 +31,7 @@ import com.owncloud.android.MainApp;
  */
 public class ProviderMeta {
     public static final String DB_NAME = "filelist";
-    public static final int DB_VERSION = 49;
+    public static final int DB_VERSION = 50;
 
     private ProviderMeta() {
         // No instance
@@ -220,6 +220,7 @@ public class ProviderMeta {
         public static final String SYNCED_FOLDER_WIFI_ONLY = "wifi_only";
         public static final String SYNCED_FOLDER_CHARGING_ONLY = "charging_only";
         public static final String SYNCED_FOLDER_ENABLED = "enabled";
+        public static final String SYNCED_FOLDER_ENABLED_TIMESTAMP_MS = "enabled_timestamp_ms";
         public static final String SYNCED_FOLDER_TYPE = "type";
         public static final String SYNCED_FOLDER_SUBFOLDER_BY_DATE = "subfolder_by_date";
         public static final String SYNCED_FOLDER_ACCOUNT = "account";
--- a/src/main/java/com/owncloud/android/providers/FileContentProvider.java
+++ b/src/main/java/com/owncloud/android/providers/FileContentProvider.java
@@ -63,6 +63,7 @@ import java.util.Locale;
 import javax.inject.Inject;
 
 import androidx.annotation.NonNull;
+import dagger.android.AndroidInjection;
 
 /**
  * The ContentProvider for the ownCloud App.
@@ -419,6 +420,7 @@ public class FileContentProvider extends ContentProvider {
 
     @Override
     public boolean onCreate() {
+        AndroidInjection.inject(this);
         mDbHelper = new DataBaseHelper(getContext());
         mContext = getContext();
 
--- a/src/main/java/com/owncloud/android/providers/FileContentProvider.java
+++ b/src/main/java/com/owncloud/android/providers/FileContentProvider.java
@@ -42,9 +42,11 @@ import android.net.Uri;
 import android.os.Binder;
 import android.text.TextUtils;
 
+import com.nextcloud.client.core.Clock;
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
 import com.owncloud.android.datamodel.OCFile;
+import com.owncloud.android.datamodel.SyncedFolder;
 import com.owncloud.android.db.ProviderMeta;
 import com.owncloud.android.db.ProviderMeta.ProviderTableMeta;
 import com.owncloud.android.lib.common.accounts.AccountUtils;
@@ -57,6 +59,8 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Locale;
+
+import javax.inject.Inject;
 
 import androidx.annotation.NonNull;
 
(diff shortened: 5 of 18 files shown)


src/main/java/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -2013,6 +2014,24 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
...
             }
+
+            if (oldVersion < 50 && newVersion >= 50) {
+                Log_OC.i(SQL, "Entering in the #50 add persistent enable date to synced_folders table");
+                db.beginTransaction();
+                try {
+                    db.execSQL(ALTER_TABLE + ProviderTableMeta.SYNCED_FOLDERS_TABLE_NAME +
+                                   ADD_COLUMN + ProviderTableMeta.SYNCED_FOLDER_ENABLE_DATE + " INTEGER ");
+
+                    upgraded = true;
```
AndyScherzinger: After adding the timestamp for the sync items themselves wouldn't the former key's value need to be migrated into the column field right here at this piece of code? Else the timestamp is 0 for all existing db entries.

AndyScherzinger:
@ArisuOngaku Thanks for this PR 🙏 

@tobiasKaminsky @mario fancy taking a look/review ❤️ Code-wise it seems fine to me, besides one minor question regarding data migration for existing db entries.

src/main/java/com/owncloud/android/providers/FileContentProvider.java
ArisuOngaku: Yes that's true, I completely forgot about that. I'm working on that ! :)

AndyScherzinger: Again thanks a lot for your effort @ArisuOngaku ❤️

ArisuOngaku: No problem, it's an honor to contribute to this awesome project!

ArisuOngaku:
There you go, disabled folders get enable_date = -1 and enabled get enable_date = System.currentTimeMillis()

This avoids enabled folders from uploading old files, which when actually desired, should be satisfied by #4788

src/main/java/com/owncloud/android/db/ProviderMeta.java
```diff
@@ -220,6 +220,7 @@ private ProviderMeta() {
         public static final String SYNCED_FOLDER_WIFI_ONLY = "wifi_only";
         public static final String SYNCED_FOLDER_CHARGING_ONLY = "charging_only";
         public static final String SYNCED_FOLDER_ENABLED = "enabled";
+        public static final String SYNCED_FOLDER_ENABLE_DATE = "enabled_date";
```
tobiasKaminsky: If it is "enableD" date, then we should use this everywhere.

ArisuOngaku: It's a typo, I meant "enable_date", though maybe "enabled_date" is better ?
Let me know what's best and I'll make the switch (english is not my native language)

tobiasKaminsky: If it is enable"D" date, we should use it everywhere.

ArisuOngaku: After a quick search, it looks like "enable date" is the right grammar, so I'll push an update to fix my typo

src/main/java/com/owncloud/android/datamodel/SyncedFolder.java
```diff
@@ -46,6 +46,7 @@
     private String account;
     private Integer uploadAction;
     private boolean enabled;
+    private long enableDate;
```
ezaquarii: `enableData` does not mean too much. Also, what is the time unit?
 How about `enabledTimestampMs`?

src/main/java/com/owncloud/android/datamodel/SyncedFolderDisplayItem.java
```diff
@@ -58,7 +58,8 @@ public SyncedFolderDisplayItem(long id, String localPath, String remotePath, Boo
                                    Boolean subfolderByDate, String account, Integer uploadAction, Boolean enabled,
                                    List<String> filePaths, String folderName, long numberOfFiles, MediaFolderType type)
     {
-        super(id, localPath, remotePath, wifiOnly, chargingOnly, subfolderByDate, account, uploadAction, enabled, type);
+        super(id, localPath, remotePath, wifiOnly, chargingOnly, subfolderByDate, account, uploadAction, enabled,
+              enabled ? System.currentTimeMillis() : -1, type);
```
ezaquarii: 1. Magic numbers are hard to understand. Please conver this to a named constant.
2. Why `-1`?
3. This parameters list is very long and inlining extra logic there does not help at all. Can we move this `enabled ? System.currentTimeMillis() : -1` ifology to the constructor body? Put default value in `super`, override in the body.

src/main/java/com/owncloud/android/datamodel/SyncedFolderProvider.java
```diff
@@ -163,6 +163,7 @@ public int updateSyncedFolderEnabled(long id, Boolean enabled) {
                 SyncedFolder syncedFolder = createSyncedFolderFromCursor(cursor);
 
                 syncedFolder.setEnabled(enabled);
+                syncedFolder.setEnableDate(enabled ? System.currentTimeMillis() : -1);
```
ezaquarii: 1. This is 3rd time we have this code snippet. Can we extract this to a well named (static?) method?
2. Use of `System.currentTimeMillis` makes it untestable. Can we inject `Clock` component via contructor and use it instead?

src/main/java/com/owncloud/android/datamodel/SyncedFolderProvider.java
```diff
@@ -370,6 +373,7 @@ private ContentValues createContentValuesFromSyncedFolder(SyncedFolder syncedFol
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_WIFI_ONLY, syncedFolder.getWifiOnly());
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_CHARGING_ONLY, syncedFolder.getChargingOnly());
         cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_ENABLED, syncedFolder.isEnabled());
+        cv.put(ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_ENABLE_DATE, syncedFolder.getEnableDate());
```
ezaquarii: I find `SYNCED_FOLDER_ENABLE_DATE` also a bit confusing. How about `SYNCED_FOLDER_ENABLE_TIMESTAMP_MS`?

src/main/java/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -2013,6 +2014,30 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
...
+                Log_OC.i(SQL, "Entering in the #50 add persistent enable date to synced_folders table");
+                db.beginTransaction();
+                try {
+                    db.execSQL(ALTER_TABLE + ProviderTableMeta.SYNCED_FOLDERS_TABLE_NAME +
+                                   ADD_COLUMN + ProviderTableMeta.SYNCED_FOLDER_ENABLE_DATE + " INTEGER ");
+
+                    db.execSQL("UPDATE " + ProviderTableMeta.SYNCED_FOLDERS_TABLE_NAME + " SET " +
+                                   ProviderTableMeta.SYNCED_FOLDER_ENABLE_DATE + " = CASE " +
+                                   " WHEN enabled = 0 THEN -1 " +
+                                   " ELSE " + System.currentTimeMillis() +
```
ezaquarii: `Clock`?

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
```diff
@@ -700,6 +682,7 @@ private SyncedFolderDisplayItem updateSyncedFolderItem(SyncedFolderDisplayItem i
         item.setSubfolderByDate(subfolderByDate);
         item.setUploadAction(uploadAction);
         item.setEnabled(enabled);
+        item.setEnableDate(enabled ? System.currentTimeMillis() : -1);
```
ezaquarii: I see this pattern repeats itself over and over. Can we consider folding this logic into `setEnabled`? It looks like those 2 fields are linked together logically and `setEnabled(false)` overrides the timestamp.

src/main/java/com/owncloud/android/ui/adapter/SyncedFolderAdapter.java
```diff
@@ -148,18 +148,22 @@ public void onBindHeaderViewHolder(SectionedViewHolder commonHolder, final int s
         holder.syncStatusButton.setVisibility(View.VISIBLE);
         holder.syncStatusButton.setTag(section);
         holder.syncStatusButton.setOnClickListener(v -> {
-            mSyncFolderItems.get(section).setEnabled(!mSyncFolderItems.get(section).isEnabled());
-            setSyncButtonActiveIcon(holder.syncStatusButton, mSyncFolderItems.get(section).isEnabled());
-            mListener.onSyncStatusToggleClick(section, mSyncFolderItems.get(section));
+            SyncedFolderDisplayItem item = mSyncFolderItems.get(section);
+            item.setEnabled(!item.isEnabled());
+            item.setEnableDate(item.isEnabled() ? System.currentTimeMillis() : -1);
```
ezaquarii: DRY

src/main/java/com/owncloud/android/ui/adapter/SyncedFolderAdapter.java
```diff
@@ -148,18 +148,22 @@ public void onBindHeaderViewHolder(SectionedViewHolder commonHolder, final int s
...
 
         holder.syncStatusButton.setVisibility(View.VISIBLE);
         holder.syncStatusButton.setTag(section);
         holder.syncStatusButton.setOnClickListener(v -> {
-            mSyncFolderItems.get(section).setEnabled(!mSyncFolderItems.get(section).isEnabled());
-            setSyncButtonActiveIcon(holder.syncStatusButton, mSyncFolderItems.get(section).isEnabled());
-            mListener.onSyncStatusToggleClick(section, mSyncFolderItems.get(section));
+            SyncedFolderDisplayItem item = mSyncFolderItems.get(section);
+            item.setEnabled(!item.isEnabled());
+            item.setEnableDate(item.isEnabled() ? System.currentTimeMillis() : -1);
```
ezaquarii: DRY

src/main/java/com/owncloud/android/datamodel/SyncedFolderDisplayItem.java
ArisuOngaku: 1. I'll convert that -1 to a constant.

2. Well actually the value of enable_date when enabled=false doesn't matter, it's never used and is always replaced with a System.currentTimeMillis() so i went for -1 as an "invalid" / "null-like" value and [0; +inf[ the rest of the time (0 is valid).

3. I'll move that if logic in the constructor as requested.

src/main/java/com/owncloud/android/datamodel/SyncedFolderProvider.java
ArisuOngaku: I agree with this new name, I'll change it.

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
ArisuOngaku: This is true, what was I thinking when I did this I'm sorry xD

src/main/java/com/owncloud/android/providers/FileContentProvider.java
ArisuOngaku: You mean something like `Clock.systemUTC().millis()` ?

src/main/java/com/owncloud/android/datamodel/SyncedFolderDisplayItem.java
ArisuOngaku: Is `DISABLED_ENABLE_TIMESTAMP_MS` a good name for that constant ?

src/main/java/com/owncloud/android/providers/FileContentProvider.java
ArisuOngaku: Also `build.gradle`'s min api is 16, and Clock was added in 26 https://developer.android.com/reference/java/time/Clock

ArisuOngaku: Nevermind, I found the Clock class in nextcloud core

src/main/java/com/owncloud/android/datamodel/SyncedFolder.java
ArisuOngaku: I agree with the name `enableTimestampMs`.

Can a timestamp be timezoned? (I guess you said that because of my bad choice of "date" in the name) Anyway, no, this should only be UTC.

src/main/java/com/owncloud/android/datamodel/SyncedFolder.java
```diff
@@ -35,9 +38,20 @@
...
+    /**
+     * @param enabled whether the target synced folder is enabled or not
+     * @return the timestamp in ms for enabled folders, -1 for disabled ones.
+     */
+    public static final long getNewEnableTimestampMs(boolean enabled) {
+        return enabled ? CLOCK.getCurrentTime() : DISABLED_ENABLE_TIMESTAMP_MS;
+    }
+
     private static final long serialVersionUID = -793476118299906429L;
+    private static final Clock CLOCK = new ClockImpl();
```
ArisuOngaku: @ezaquarii I was unable to find a better way of using this class, got any tip/suggestion?
I also found that System.currentTimeMillis() is used multiple times across the whole project, though that is of course not a reason to keep using it.

ArisuOngaku: I read your comment again, and now I think I understand what was confusing to me.

"Use of System.currentTimeMillis makes it untestable. Can we inject Clock component via contructor and use it instead?"

Now that the enableTimestampMs value is generated by a static method in the SyncedFolder class, is this still relevant to use the Clock component ?

ezaquarii: > Now that the enableTimestampMs value is generated by a static method in the SyncedFolder class, is this still relevant to use the Clock component ?

Actually, after giving it a 2nd thought, I was wrong about this one, sorry. (-:

This is a plain POJO, a dumb data structure without any logic. It should not have any external dependencies at all. Let's just inject the timestamp value and we can move forward.

Apart from the setter that requires enabled and timestamp pair, I see no need to make it more complex.

src/main/java/com/owncloud/android/datamodel/SyncedFolder.java
```diff
@@ -35,9 +38,11 @@
 @AllArgsConstructor
 public class SyncedFolder implements Serializable, Cloneable {
     public static final long UNPERSISTED_ID = Long.MIN_VALUE;
+    public static final long DISABLED_ENABLE_TIMESTAMP_MS = -1;
```
ezaquarii: How about `EMPTY_ENABLED_TIMESTAMP_MS`? "Disabled enable" is a bit weird...

src/main/java/com/owncloud/android/datamodel/SyncedFolderDisplayItem.java
ezaquarii: `EMPTY_ENABLED_TIMESTAMP_MS`?

src/main/java/com/owncloud/android/providers/FileContentProvider.java
ezaquarii: Yes, it's our own `Clock` implementation. Please try using one that is `@Inject`-ed somewhere and pass it down.

Sadly, Dagger is not yet available everywhere. If that is the case, we exceptionally create `*Impl` as an intermediate step before further refactoring "when the time comes".

src/main/java/com/owncloud/android/datamodel/SyncedFolderDisplayItem.java
ArisuOngaku: Deal.

src/main/java/com/owncloud/android/providers/FileContentProvider.java
ArisuOngaku: Ok thanks for that info!

src/main/java/com/owncloud/android/datamodel/SyncedFolder.java
ArisuOngaku: No problem! I'm going to make the modifications.

src/main/java/com/owncloud/android/datamodel/SyncedFolder.java
```diff
@@ -46,6 +51,7 @@
     private String account;
     private Integer uploadAction;
     private boolean enabled;
+    private long enableTimestampMs;
```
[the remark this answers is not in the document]
ArisuOngaku: I'll make the change.

src/main/java/com/owncloud/android/datamodel/SyncedFolder.java
```diff
@@ -64,22 +70,48 @@
...
         this.chargingOnly = chargingOnly;
         this.subfolderByDate = subfolderByDate;
         this.account = account;
         this.uploadAction = uploadAction;
-        this.enabled = enabled;
+        this.setEnabled(enabled);
         this.type = type;
     }
 
+    public void setEnabled(boolean enabled) {
```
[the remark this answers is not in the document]
ArisuOngaku: I'll do that. Nothing can be perfect. Though I find this solution quite elegant.

ArisuOngaku:
So I discovered dagger on my last push; I tested my code of course, but still, how can you tell whether dagger is available in a class? I'm not 100% sure for in `SyncedFolderProvider`, `FileContentProvider` and `SyncedFolderAdapter`.

ezaquarii:
> So I discovered dagger on my last push; I tested my code of course, but still, how can you tell whether dagger is available in a class? 

Please take a look at `com.nextcloud.client.di.ComponentsModule`. If the class is listed there, the compiler will generate an injector - fancy name for a boring [visitor](https://en.wikipedia.org/wiki/Visitor_pattern) that sets the values.

If the class is not there, and it is Android component, you can add it and mark the class as `Injectable`.

The trouble is that the injector/vistor must be called somehow to set all dependencies, something like (pseudocode):

```
RubeGoldbergMachine myRubeGoldbergMachine = new RubeGoldbergMachine();
// now the machine is emtpy, with most of it's internal dependencies == null
injectorInstance.injectInto(myRubeGoldbergMachine); // this sets internal members
```

Now, for most major components, we try to rely on Android platform hooks to call the injector during component lifecycle for us. For Activities and Fragments we have a application-wide listener that is called for each new created component. Just take a look at `com.nextcloud.client.di` package  [where this automagic lives](https://github.com/nextcloud/android/blob/9e15f5ea398244e411e7b0ee93f8df4761c18155/src/main/java/com/nextcloud/client/di/FragmentInjector.java#L30)

For `Service`s and `BroadcastReceiver`s there is a 1-line DI helper that calls appropriate injector but it must be called manually in `onCreate()`. This sucks, but there is no other way around it as those components don't have suitable lifecycle hooks where we could apply the injector before the component is used.

For anything lesser we generally don't use autogenerated Dagger injector, as it's a bit too invasive for our taste, as you need to instantiate and call some ugly boilerplate: https://github.com/nextcloud/android/blob/9e15f5ea398244e411e7b0ee93f8df4761c18155/src/main/java/com/owncloud/android/MainApp.java#L228).

As long as Android calls this magic piece for us automatically via lifecycle methods, we're happy to just mark our class with `Injectable` and take for granted that all `@Inject`-annotated members will be automagically set somewhere inside `super.onCreate()`. This is real gain.

For anything else, not so easy. The boilerplate turns out to quickly infest the code, so we prefer to inject dependencies in vanilla way - manually into contructors. Please note that we still use the term "dependency injection" in such case - we apply a design pattern, not a specific library (we can get rid of Dagger easily, btw). The gist of the idea is that you don't call constructors inside constructors - you pass your deps via arguments.

Funny thing is that since we used to liberally instantiate various components everywhere, some of the components ended up having loooong lists of implicit dependencies (yuck!), in many cases circular (eww!). Once such component is refactored by applying dependency injection pattern, the constructor argument list can become huge, revealing the true scale of the problem. For deeply buried dependecies you need to drill down through few layers of objects, causing avalanche effect.

Enough background context. Now the bad news is that the component you just work on is one of those troubling ones and sadly in the most challenging area of the app. `SyncedFolderProvider` for example has 10 places of instantiation, so any dependency injected into it will spill over those 10 areas, which in turn might cause ripple effect in other parts of the system, causing combinatorial explosion.

We've done it before, so ["DONT'T PANIC](https://en.wikipedia.org/wiki/The_Hitchhiker%27s_Guide_to_the_Galaxy_(fictional))" in large, friendly letters.

ezaquarii:
> So I discovered dagger on my last push; I tested my code of course, but still, how can you tell whether dagger is available in a class? 

@ArisuOngaku
Please take a look at `com.nextcloud.client.di.ComponentsModule`. If the class is listed there, the compiler will generate an injector - fancy name for a boring [visitor](https://en.wikipedia.org/wiki/Visitor_pattern) that sets the values.

If the class is not there, and it is Android component, you can add it and mark the class as `Injectable`.

The trouble is that the injector/vistor must be called somehow to set all dependencies, something like (pseudocode):

```
RubeGoldbergMachine myRubeGoldbergMachine = new RubeGoldbergMachine();
// now the machine is emtpy, with most of it's internal dependencies == null
injectorInstance.injectInto(myRubeGoldbergMachine); // this sets internal members
```

Now, for most major components, we try to rely on Android platform hooks to call the injector during component lifecycle for us. For Activities and Fragments we have a application-wide listener that is called for each new created component. Just take a look at `com.nextcloud.client.di` package  [where this automagic lives](https://github.com/nextcloud/android/blob/9e15f5ea398244e411e7b0ee93f8df4761c18155/src/main/java/com/nextcloud/client/di/FragmentInjector.java#L30)

For `Service`s and `BroadcastReceiver`s there is a 1-line DI helper that calls appropriate injector but it must be called manually in `onCreate()`. This sucks, but there is no other way around it as those components don't have suitable lifecycle hooks where we could apply the injector before the component is used.

For anything lesser we generally don't use autogenerated Dagger injector, as it's a bit too invasive for our taste, as you need to instantiate and call some ugly boilerplate: https://github.com/nextcloud/android/blob/9e15f5ea398244e411e7b0ee93f8df4761c18155/src/main/java/com/owncloud/android/MainApp.java#L228).

As long as Android calls this magic piece for us automatically via lifecycle methods, we're happy to just mark our class with `Injectable` and take for granted that all `@Inject`-annotated members will be automagically set somewhere inside `super.onCreate()`. This is real gain.

For anything else, not so easy. The boilerplate turns out to quickly infest the code, so we prefer to inject dependencies in vanilla way - manually into contructors. Please note that we still use the term "dependency injection" in such case - we apply a design pattern, not a specific library (we can get rid of Dagger easily, btw). The gist of the idea is that you don't call constructors inside constructors - you pass your deps via arguments.

Funny thing is that since we used to liberally instantiate various components everywhere, some of the components ended up having loooong lists of implicit dependencies (yuck!), in many cases circular (eww!). Once such component is refactored by applying dependency injection pattern, the constructor argument list can become huge, revealing the true scale of the problem. For deeply buried dependecies you need to drill down through few layers of objects, causing avalanche effect.

Enough background context. Now the bad news is that the component you just work on is one of those troubling ones and sadly in the most challenging area of the app. `SyncedFolderProvider` for example has 10 places of instantiation, so any dependency injected into it will spill over those 10 areas, which in turn might cause ripple effect in other parts of the system, causing combinatorial explosion.

We've done it before, so ["DONT'T PANIC](https://en.wikipedia.org/wiki/The_Hitchhiker%27s_Guide_to_the_Galaxy_(fictional))" in large, friendly letters.

ArisuOngaku:
@ezaquarii 
First of all, thank you for your time and this wonderful explanation. Now I have a way better understanding of how the code of this project is folded, and I feel confident in making more contributions in the future (in a more casual pace though).

Now, as I get it, the overall idea is to use dagger :
- For Activities and Fragments (automagic, just implement `com.nextcloud.client.di.Injectable`)
- For anything else that has a method in `dagger.android.AndroidInjection` (must be called in `onCreate()`)

Otherwise avoid using it but in rare case because it requires a `DaggerAppComponent.builder()` which is, at this point, less clean than the vanilla way.

Anywhere else must be done by passing dependencies in constructor parameters (which, because of the way the project was historically built, sometimes lead to needing to update a dozen of calls, and then same with the classes where the calls are).

Now I think we share the same pain, but here we are. I am going to NOT PANIC and just do it!

Also I feel lucky because the only class which needs this work in the current state of this PR is the `SyncedFolderProvider`.

ArisuOngaku:
I don't know if that's useful, but I thought maybe I should have mentioned I'm done working on the last review comments.

AndyScherzinger:
> I don't know if that's useful, but I thought maybe I should have mentioned I'm done working on the last review comments.

Sweet, yes that s indeed a valuable info.

@tobiasKaminsky @ezaquarii can you have a final look (and merge if okay)? I don't have enough spare time at the moment :(

ArisuOngaku:
Yeah sorry for that, please be sure I learnt from this situation.

AndyScherzinger:
> Yeah sorry for that, please be sure I learnt from this situation.

Nothing to be sorry for! Just happy about your ping 🥇

ezaquarii:
I'll take another look evening.

src/main/java/com/owncloud/android/utils/FilesSyncHelper.java
```diff
@@ -147,20 +120,17 @@ public FileVisitResult visitFileFailed(Path file, IOException exc) {
                             return FileVisitResult.CONTINUE;
                         }
                     });
-
+                } catch (IOException e) {
+                    Log.e(TAG, "Something went wrong while indexing files for auto upload " + e.getLocalizedMessage());
```
ezaquarii: While being here, can we write it using our logging framewrk?

Currently it is using Android log output to write an error message that nobody will ever read (because even developers don't carry a laptop with `adb logcat` wired up all the time).

We can safely use `Log_OC.e` logger here. It has identical syntax and will redirect logs to our own logger which allows us to browse, filter and  extract past logs from our embedded logs browser (somewhere in Settings, enabled in QA and dev builds).

ezaquarii:
@tobiasKaminsky @AndyScherzinger @ArisuOngaku I have only 1 minor comment - not sure if this is important so feel free to ignore it. Rest looks LGTM.

src/main/java/com/owncloud/android/utils/FilesSyncHelper.java
ezaquarii: I'm not sure if this checked exception is realistically going to happen here. Feel free to ignore it if this inot the case and we have this block only to satisfy the compiler.

ArisuOngaku: This is going to happen, it's IO, and even though a software bug is unlikely, there could be cases of corrupted file permissions or hardware failures. Or files that get deleted during the execution of this procedure.

I'm going to push `Log_OC.e(TAG, "Something went wrong while indexing files for auto upload", e);`, feel free to answer if you're not into logging the stacktrace or resolve this conversation.

ezaquarii:
Ok, so I'd this is not likely to happen due to our bugs, we don't care about stacktrace too much although it won't hurt to have it.

Feel free to close the review and carry on as I won't be available for the rest of the business day.

AndyScherzinger:
@tobiasKaminsky LGTM code-wise from @ezaquarii's and my perspective. Fancy taking a look? ❤️

tobiasKaminsky:
I did following test:
- fresh install with this branch
- enabled camera folder
- took some images --> uploaded fine
- disabled camera folder
- took some images
- re-enabled camera folder
--> nothing happened
- took a new image --> only this was uploaded

As I understand it, this PR should then also upload those images within the disabled timespan, or?

But how should this work, if we have "only" enabled timestamp?
E.g:
1000s (just to have short timestamps): enable folder
1500s disable folder
1600s take some images
2000s enable folder again: then we search for images that are enabled since 2000s (https://github.com/nextcloud/android/blob/65da616f812760362ecc0b0e02b288188babe8cc/src/main/java/com/owncloud/android/utils/FilesSyncHelper.java#L174)

But instead we would need the 1500s (disabled timestamp), or?
So we would need an "enabled/disabled" state and a "changedTimestamp"?

ArisuOngaku:
No, this pull request does not aim to upload files in a synced folder that were created while the folder was disabled. This pull request only aims to correctly upload all files that are create while a synced folder is enabled, even when the nextcloud app is closed (aka not monitoring changes) and later opened again.

Before this PR, the app effectively reseted this timestamp at each launch, meaning that when it crashed or wasn't started, any file created during this period would never get uploaded even if the synced folder was enabled.

tobiasKaminsky:
Ah, then I misunderstood the idea of it and now it makes sense and this PR fits totally fine.

Thanks for your great enhancement :heart:

ArisuOngaku:
No problem, it's been a pleasure ! :heart:

tobiasKaminsky:
As you are a member of nextcloud, you can open up a branch next time directly in this repo, so you do not have to your own.
This makes reviewing/checkout a bit more easier for us :-)

I am looking forward to "see" you next time :-)

ArisuOngaku:
Alright, thanks for that ! I'm most likely going to work on #4788 now.

Issue #1271, opened 2017-08-05
Merge notification, activities and uploads


mario:

tobiasKaminsky:
Merging Notifications and activities :+1: as they are both provided by the server.
Upload is only device specific and should stay separate:
If this is all sorted by date the failed/pending uploads will vanish somehow on the bottom of the list.
And if they stay on top of the list, then we have it just like the current way.

But maybe we just need a detailed input of @nextcloud/designers or from you if you thought already deeper about this.

mario:
@tobiasKaminsky this was actually jointly designed by me and @jancborchardt, and there is a pen&paper wireframe in one of the issues (I'll find it for you later). I believe you also commented on it. :)

Also, stuff that needs actions have their own group so nothing really disappears.

mario:
(P.S. Many many months ago)

jancborchardt:
Reposting the spec with mockups from https://github.com/nextcloud/android/issues/557#issuecomment-274776899:

Here’s the mockup of the activity stream merged with notifications as @mario and I talked about:
![img_20170124_113357](https://cloud.githubusercontent.com/assets/925062/22245246/47e67ea4-e22f-11e6-99ba-a2d21fc4ae3b.jpg)


From the writeup in the sidebar issue: https://github.com/nextcloud/android/issues/556#issuecomment-274772899

> **Activity**: Merged Activity & notifications etc, much like the unified timeline in the sidebar of files https://github.com/nextcloud/server/issues/658. Most of these entries will at some point be interactive, like accepting a share, reverting to a version, or commenting etc. Actionable items like failed uploads or shares to be accepted will be shown up top. They will be separated by intermittent headers:
> 	- **Failed uploads** is first cause it’s local and most important cause it can lead to data loss. Will have a »Retry« and »Clear« in the header with text+icons. Then icons for each individual item.
> 	- **Requires attention** (working title) are the notifications and activities with interactivity which come from the server. For example incoming shares you need to accept. (Or in the future comments where you were mentioned with an action to reply.)
> 	- **Activities / History** where all the activites go in the end, and where also the things you act upon from **Failed uploads** and **Requires attention** go to → which is a big reason why this is a single view instead of separate ones. It’s the same things.

jancborchardt:
To also answer your remark @tobiasKaminsky:

> Upload is only device specific and should stay separate:
> If this is all sorted by date the failed/pending uploads will vanish somehow on the bottom of the list.
[...]

PR #4788, opened 2019-11-03
Setting to also upload existing files

AndyScherzinger:
successor for #2873

----
Originally created by @koying

Please be kind, first time I put my nose into the code.

This adds a setting to the instant uploads to also upload files already present on the device at the time instant upload is setup.
Ref: https://help.nextcloud.com/t/why-doesnt-android-client-upload-existing-pictures/10365

95% of the PR is for handling of the setting itself.
It might be worth wondering whether this deserves a setting at all, really, ie whether not uploading existing files is worthwhile, as it's a bit counter-intuitive, imo


ArisuOngaku:
In the original PR it was suggested that this should not be a setting but rather a one time action with a prompt when exiting the settings view.

However, this would mean that when enabling/disabling the synced folder through the button from the synced folder list (and not the full parameters view) would skip files created while disabled, and not resync them when reenabled.

Maybe this setting should be on by default ? This might also mean needing to warn the user that files with the same name at destination will be overridden while there is not conflict detection/management.

AndyScherzinger:
> However, this would mean that when enabling/disabling the synced folder through the button from the synced folder list (and not the full parameters view) would skip files created while disabled, and not resync them when reenabled.

> Maybe this setting should be on by default ? This might also mean needing to warn the user that files with the same name at destination will be overridden while there is not conflict detection/management.

Fully agree and so would probably @jancborchardt and @nextcloud/designers and @jospoortvliet. The user shouldn't have to think about it. It even say when a db entry is initially created the user could be asked if they want to upload existing files. So I'd rather see it an an initial opt-out thing. If initially the user decided to not upload existing files then they are left out, else the users simply want to always have them uploaded (which is likely the default: Backup anything and don't bother me).

tobiasKaminsky:
When enabling (don't matter how), this then takes all images from existing folder and uploads it?
How do we prevent cases where files already exists?
(sorry if this is again the same discussion).

We could do it this way:
- show an info that this might lead to duplicated data
- upload and do not overwrite any file with the same name
- let user remove the possible duplicated files

szaimen:
In my opinion 
> upload and do not overwrite any file with the same name

Is the best option in combination with a warning that already existing files with the same name don't get overwritten. Then a user has the choice to choose a different upload folder when he/her isn't sure about whether there are already files with the same name in this upload folder.

dgtlmoon:
FWIW https://github.com/nextcloud/android/pull/2873#issuecomment-549124650 the APK works well for me

cpw:
Sadly, https://github.com/nextcloud/android/pull/4788#issuecomment-549124896 doesn't seem to work. I look forward to this feature.

dgtlmoon:
@cpw please be more helpful/mindful, "doesnt seem to work" does really not help anyone to improve the topic, how exactly? please fill in a complete error report, device type, what you tried, messages you saw (paste the text as it is exactly), version number of client, android and nextcloud server, connection type etc

cpw:
Nevermind. I reinstalled this morning, and it seems to be trying to sync existing files this time. :woman_shrugging:

Excel1:
> APK file: https://www.kaminsky.me/nc-dev/android-artifacts/11517.apk
> 
> qrcode
> 
> To test this change/fix you can simply download above APK file and install and test it in parallel to your existing Nextcloud app.

https://github.com/nextcloud/android/pull/2873#issuecomment-549124650

These APKs didnt work for me. (Note 8 - Android 9) After clicking the cloud icon below the tab autoupload the app restart.

ArisuOngaku:
@Excel1 this is an APK from **another (older) pull request** that is not relevant anymore for testing. Please try with the newer https://github.com/nextcloud/android/pull/4788#issuecomment-549124896

Also keep in mind that this PR modifies the database which means in order to test in realistic conditions, you should perform a full manual uninstall of the previous test versions before installing this new APK.

Also this PR is still currently in discussion and development.

Excel1:
@ArisuOngaku thats absolutley right. I only wanna give feedback :) https://github.com/nextcloud/android/pull/4788#issuecomment-549124896 i recently tested this apk. Same issues: (Note 8 - Android 9) After clicking the cloud icon below the tab autoupload the app restart.

ArisuOngaku:
Thanks for that! I can't reproduce this crash. Would you mind sharing your logs?

(settings -> logs -> 3 dots up right corner -> send by email, and you can actually upload this as a file if you select the nextcloud app itself)

Excel1:
@ArisuOngaku sure. here it is. (i censored sensitive data)
[logs.txt](https://github.com/nextcloud/android/files/3845986/logs.txt)

dgtlmoon:
@Excel1 did you censor this line? the URL looks really broken to me `Creating SSL Socket with remote to.nextcloud.url:443`

ArisuOngaku:
@Excel1 I see no stack trace though if I'm not mistaken, this is the part where you experience the crash:
```
2019-11-14T10:44:32.137+0100;V;SyncedFolderProvider;Inserting /storage/0000-0000/DCIM/Camera with enabled=true
2019-11-14T10:44:33.397+0100;D;Debug;start logging
 ```

My best guess is that the local database was not wiped, did you fully uninstall the Nextcloud QA app before installing the new APK? If yes, then try that again but clean the app's cache before.

What indicates that the database was correctly wiped is that you need to login again when you start the app.

**Comment by Excel1** [comment]
@ArisuOngaku Yes i did. And now i did it again (and deleted cache and app data). Should i also uninstall the original Nextcloud App ? 
[logs.txt](https://github.com/nextcloud/android/files/3846280/logs.txt)

**Comment by ArisuOngaku** [comment]
@Excel1 No, I see no reason the original Nextcloud app would interfere.
Though your logs are going back to 2019-10-13 which means it wasn't wiped. I'm sorry but we have to be sure: did you have to log back in when starting the app? If no, then the actual app cache is not getting properly wiped.

When I uninstall (manually) then install again the same APK (on emulator), my cache and logs get deleted. By any chance, are you correctly opening the `Nextcloud QA` app (not to be confused with `Nextcloud`)?

**Comment by Excel1** [comment]
@ArisuOngaku Okay. I definitely uninstalled the right version (QA) and deleted the right cache (QA). Here are the log (definitely from the QA version) :thinking: 

[logs.txt](https://github.com/nextcloud/android/files/3846403/logs.txt)

**Comment by ArisuOngaku** [comment]
@Excel1 Thanks a lot for your patience. I looked deeper in your logs and it looks like your cache actually gets properly deleted.

Are you familiar with android debug tools (adb) and the logcat utility? Recording the logs from there while reproducing the crash would give us a more precise and definitive answer to what's happening.

**Comment by Excel1** [comment]
@ArisuOngaku No problem, thanks for helping me. I will try it and inform you about it as soon as I have done.

**Comment by ArisuOngaku** [comment]
@AndyScherzinger I rebased this PR on the latest master to fix conflicts (and also removed the typo commit and integrated its changes in the first one by koying), am I allowed to force push and work on this PR?
(what I want to push https://github.com/nextcloud/android/compare/master...ArisuOngaku:instantupload_all)

**Comment by AndyScherzinger** [comment]
> am I allowed to force push and work on this PR?

@ArisuOngaku absolutely yes! ❤️ 👍 Only when you label it `2. to review` it shouldn't be rebased again since other might have checked out the branch for QA.

**Comment by ArisuOngaku** [comment]
### Code and bugs
An update to what I've done for the last 3 force pushes :
1. Rebased this PR on the latest master and squashed a typo fixing commit
2. Squashed the other fixing commits on the first one + probably fixed @cpw 's problem (my theory is that it was working but it actually would have taken up to 15min to detect new changes, I fixed this problem by moving the instant job scheduling in `FilesSyncHelper.java`) + made the new setting on by default
3. DCO (I really got to get used to signing off my own commits...)

@Excel1 I would advise that you try the latest generated [Nextcloud QA apk](https://github.com/nextcloud/android/pull/4788#issuecomment-556053416) to check if your crash still happens, though I doubt my new changes fixed it.


### Discussion
I would suggest an alternative frontend for the setting: `Only upload new(ly created) files`, off by default instead of `(Also) Upload existing files`, on by default.

Imho, conflict management for uploads should be done in `FilesSyncJob` (in a centralized way), that is to say in a future pull request (note that we already correctly keep track of "already tried to be uploaded" files).

Thus, a confirm dialog stating something like `Caution! Files with matching names that are already present on the server will be overwritten without verification.` should be added after saving settings from the synced folder detailed settings view (and the said future pull request should also remove that dialog).

**Comment by Excel1** [comment]
@ArisuOngaku Thank you for the update. I tried it but it directly crashes if i click at the cloud icon or the save button after the dialog:

ADB Error Report


2019-11-20 18:27:52.191 25611-25611/com.nextcloud.android.qa V/SyncedFolderProvider: Inserting /storage/0000-0000/DCIM/Camera with enabled=true
2019-11-20 18:27:52.193 25611-25611/com.nextcloud.android.qa E/SQLiteLog: (1) table synced_folders has no column named enabled_timestamp_ms
2019-11-20 18:27:52.197 25611-25611/com.nextcloud.android.qa E/SQLiteDatabase: Error inserting existing=true enabled_timestamp_ms=1574270872178 subfolder_by_date=false wifi_only=true upload_option=2 local_path=/storage/0000-0000/DCIM/Camera remote_path=/SofortUpload/Camera charging_only=false type=1 enabled=true account=User@nextcloud.test
    android.database.sqlite.SQLiteException: table synced_folders has no column named enabled_timestamp_ms (code 1 SQLITE_ERROR[1]): , while compiling: INSERT INTO synced_folders(existing,enabled_timestamp_ms,subfolder_by_date,wifi_only,upload_option,local_path,remote_path,charging_only,type,enabled,account) VALUES (?,?,?,?,?,?,?,?,?,?,?)
        at android.database.sqlite.SQLiteConnection.nativePrepareStatement(Native Method)
        at android.database.sqlite.SQLiteConnection.acquirePreparedStatement(SQLiteConnection.java:1229)
        at android.database.sqlite.SQLiteConnection.prepare(SQLiteConnection.java:703)
        at android.database.sqlite.SQLiteSession.prepare(SQLiteSession.java:588)
        at android.database.sqlite.SQLiteProgram.<init>(SQLiteProgram.java:59)
        at android.database.sqlite.SQLiteStatement.<init>(SQLiteStatement.java:31)
        at android.database.sqlite.SQLiteDatabase.insertWithOnConflict(SQLiteDatabase.java:2019)
        at android.database.sqlite.SQLiteDatabase.insert(SQLiteDatabase.java:1890)
        at com.owncloud.android.providers.FileContentProvider.insert(FileContentProvider.java:335)
        at com.owncloud.android.providers.FileContentProvider.insert(FileContentProvider.java:255)
        at android.content.ContentProvider$Transport.insert(ContentProvider.java:266)
        at android.content.ContentResolver.insert(ContentResolver.java:1594)
        at com.owncloud.android.datamodel.SyncedFolderProvider.storeSyncedFolder(SyncedFolderProvider.java:80)
        at com.owncloud.android.ui.activity.SyncedFoldersActivity.onSyncStatusToggleClick(SyncedFoldersActivity.java:562)
        at com.owncloud.android.ui.adapter.SyncedFolderAdapter.lambda$onBindHeaderViewHolder$1$SyncedFolderAdapter(SyncedFolderAdapter.java:165)
        at com.owncloud.android.ui.adapter.-$$Lambda$SyncedFolderAdapter$cFk0LH91zGGKKwylS1aaEpKJLYk.onClick(Unknown Source:6)
        at android.view.View.performClick(View.java:7359)
        at android.view.View.performClickInternal(View.java:7325)
        at android.view.View.access$3200(View.java:846)
        at android.view.View$PerformClick.run(View.java:27807)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7073)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:965)
2019-11-20 18:27:52.198 25611-25611/com.nextcloud.android.qa D/AndroidRuntime: Shutting down VM
    
    
    --------- beginning of crash
2019-11-20 18:27:52.200 25611-25611/com.nextcloud.android.qa E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.nextcloud.android.qa, PID: 25611
    android.database.SQLException: ERROR content://org.nextcloud.qa.provider/synced_folders
        at com.owncloud.android.providers.FileContentProvider.insert(FileContentProvider.java:340)
        at com.owncloud.android.providers.FileContentProvider.insert(FileContentProvider.java:255)
        at android.content.ContentProvider$Transport.insert(ContentProvider.java:266)
        at android.content.ContentResolver.insert(ContentResolver.java:1594)
        at com.owncloud.android.datamodel.SyncedFolderProvider.storeSyncedFolder(SyncedFolderProvider.java:80)
        at com.owncloud.android.ui.activity.SyncedFoldersActivity.onSyncStatusToggleClick(SyncedFoldersActivity.java:562)
        at com.owncloud.android.ui.adapter.SyncedFolderAdapter.lambda$onBindHeaderViewHolder$1$SyncedFolderAdapter(SyncedFolderAdapter.java:165)
        at com.owncloud.android.ui.adapter.-$$Lambda$SyncedFolderAdapter$cFk0LH91zGGKKwylS1aaEpKJLYk.onClick(Unknown Source:6)
        at android.view.View.performClick(View.java:7359)
        at android.view.View.performClickInternal(View.java:7325)
        at android.view.View.access$3200(View.java:846)
        at android.view.View$PerformClick.run(View.java:27807)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7073)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:965)

**Comment by ArisuOngaku** [comment]
@Excel1 I think this is caused by google's automatic backup function. The actual database gets restored to its previous state when you reinstall the QA app.

@AndyScherzinger @tobiasKaminsky What do you think about disabling the android backup feature in the Nextcloud QA app?

`qa/AndroidManifest.xml`
```AndroidManifest.xml
<manifest xmlns:android="http://schemas.android.com/apk/res/android">
    <application android:allowBackup="false" />
</manifest>
```

AndyScherzinger:
> What do you think about disabling the android backup feature in the Nextcloud QA app?

I'd be fine with that, what do you think @tobiasKaminsky ?

AndyScherzinger:
The override discussion needs a larger audience I guess, so @nextcloud/designers @jancborchardt @tobiasKaminsky @jospoortvliet I my opinion putting a caution sign that we might override files (with the same name) seems to be desatructive, while a caution sign which does the ..(1) ..(2) renaming logik wouldn't while that means that when repeaqting this process you might end up with several copies of a file, but yopu'll never loose one due to overrides.

szaimen:
Sry for spamming the chat but as I stated [above](https://github.com/nextcloud/android/pull/4788#issuecomment-551231396) should IMO already existing files with the same name never get overwritten nor duplicated during the initial upload:

> In my opinion 
> > upload and do not overwrite any file with the same name
> 
> Is the best option in combination with a warning that already existing files with the same name don't get overwritten. Then a user has the choice to choose a different upload folder (or delete files in this folder before uploading) if he/her wants to upload everything (and he/her isn't sure about whether there are already files with the same name in this upload folder). Best thing is, that this initial upload then can be triggered more then one time and only uploads files that are (based on the name) not already uploaded into this upload folder.

So in conclusion is the best way to deal with this problem:

- never overwrite already existing files with the same name during the initial upload (and don't upload the file with this name at all)
- notify the user after the setup that this is the case but that he/her can choose/create another folder or delete files inside this folder first
- then if the initial upload fails for whatever reason you could easily redo the procedure without having to deal with duplicate/overwritten files -> it is more robust

AndyScherzinger:
> So in conclusion is the best way to deal with this problem IMO

This doesn't fix the issue properly, because:
* some OEMs decided to go for a camera like naming schema (like `DSC0001.jpg`) so in case of several devices or same oem devices you'll definitely have name collisions and ignore all these photos which are definitely different. SO you'd loose data which isn't okay. SO you would atr least have to check name+date+size and in case it differs uploade them with a renamed file name (adding (1), (2), etc.). That way you can be sure you don't loose any data, while you still might create duplicated (as in one imnage thatr has been reworked over time).

@tobiasKaminsky @jancborchardt @jospoortvliet can you please join this discussion so we can be sure @ArisuOngaku heads into the right direction so no effort get lost due to decisions not made. ? ❤️

dgtlmoon:
Can of worms - for the first version please just alert the user that files with the same name will be overriden. otherwise I fear we will lose momentem on this truely wonderful merge

true negotiation of synced files is a lot more complex and should be handled in another issue.  (Checksum provision from nextcloud API etc)

dgtlmoon:
By the way, what kind of comment is "so we can be sure @ArisuOngaku heads into the right direction so no effort get lost due to decisions not made. "  

You mean "let us make a decision for someone else, but not contribute any code?"

Are you saying Arisu cant' comprehend the implications of the code or what? geezus man, grow up!

szaimen:
> some OEMs decided to go for a camera like naming schema (like DSC0001.jpg) so in case of several devices or same oem devices you'll definitely have name collisions and ignore all these photos which are definitely different. SO you'd loose data which isn't okay. SO you would atr least have to check name+date+size and in case it differs uploade them with a renamed file name (adding (1), (2), etc.). That way you can be sure you don't loose any data, while you still might create duplicated (as in one imnage thatr has been reworked over time).

I am okay with that but if it doesn't differ for a picture I think it shouldn't get uploaded at all to make it more robust.

ArisuOngaku:
> Can of worms - for the first version please just alert the user that files with the same name will be overriden. otherwise I fear we will lose momentem on this truely wonderful merge
> 
> true negotiation of synced files is a lot more complex and should be handled in another issue. (Checksum provision from nextcloud API etc)

@dgtlmoon while I agree with this, this is not that much more work to not actually override matching existing files but instead, as @AndyScherzinger wisefully suggested, rename with a (1..n) suffix. Making a more precise file conflict management then could be reasonably handled in another issue.



> By the way, what kind of comment is "so we can be sure @ArisuOngaku heads into the right direction so no effort get lost due to decisions not made. "
> 
> You mean "let us make a decision for someone else, but not contribute any code?"
> 
> Are you saying Arisu cant' comprehend the implications of the code or what? geezus man, grow up!

@dgtlmoon You're missing out the point of Andy's message. He meant and I think we all mean to make this project progress as fast as possible, because it's our common goal. This wasn't about making decisions about code at this point, but rather design, for which I'm not competent enough alone - thus the call to discussion for decisions.

AndyScherzinger:
> Are you saying Arisu cant' comprehend the implications of the code or what? geezus man, grow up!

@dgtlmoon please read the CoC

> By the way, what kind of comment is "so we can be sure @ArisuOngaku heads into the right direction so no effort get lost due to decisions not made. "

> You mean "let us make a decision for someone else, but not contribute any code?"

Please don't (wrongly) imply things. A PR won't just be merged because it exists. It will be merged when the projecct agrees on it being the right solution (read: governance, vision, etc.). I just want to prevent work from being rejected in case decisions on the behavior of the app have been made before agreeing on it (which has absolutely nothing to do with code). As for contributions, please see https://github.com/nextcloud/android/graphs/contributors before making wrong allegations 👓 

> true negotiation of synced files is a lot more complex and should be handled in another issue.

absolutely

>  for the first version please just alert the user that files with the same name will be overriden

destructive behavior, it should rather be the other way round

@ArisuOngaku Exactly my intensions, thanks you ❤️ While I think you do great contributions and your designb is just fine. I just think that we haven't yet decided on the specifics of the behavior (override, rename, wanrings, etc.) which will lead to discussion and I would like them to happen before the logic is implemented (read: the time spent) so that it doesn't have to be changed (read: wasting efforts being made before). 

Just to make this clear and explicit: I also fully support and love what you are doing here (and in the former PR).

jancborchardt:
Sorry, there’s a lot of text and comments here now so I can’t go through them all. :\ Please keep this in mind when commenting, as it just gets harder and harder to see what’s important. Stay concise and relevant.

@AndyScherzinger regarding your question, I’d say that duplicating files is nearly as bad as losing data, it’s frustrating as you have to clean them up again. So this case should also be avoided as far as possible.
I would say it’d be best for @AndyScherzinger @tobiasKaminsky @ArisuOngaku to talk through the options for this before moving on. I trust you make a good decision – keep in mind that we can do small steps towards a goal as well.

ArisuOngaku:
### Development
Last commit detailed:
- When file already exists server side, the upload fails and registers a conflict in the Uploads tab
- When clicking on a conflict, the conflict prompt opens to let the user chose what to do
- [Bug] When selecting "KEEP SERVER" (and once the download is finished), sometimes the conflict list is not visually refreshed

I still have to add a flag to specify whether to save the file locally or not in the conflict activity (uploaded files aka KEEP LOCAL and KEEP BOTH actually save a copy of the file just like if the user pressed "Download" on the target file, which is undesired in this case).

[Design] Uploads and downloads should be merged together in a new class named `Transfer`.

### Discussion
On the desktop app, conflicts are shown in the "Activity" tab. I think the "Uploads" tab should also contain Downloads. Should we rename the "Uploads" tab to "Transfers"? (which also means adding downloads)

ArisuOngaku:
Hey @Excel1 I rebased this on latest master which contains #4885, so your problem should be fixed.
Please try https://www.kaminsky.me/nc-dev/android-artifacts/11749.apk with same procedure, full manual uninstall then install this new APK.

jancborchardt:
@ArisuOngaku nice! Very relevant to what you mentioned in the comment seems to be the issue & specification at **Merge notification, activities and uploads #1271** – your input would be appreciated!

ArisuOngaku:
I rebased on master + fixed the local behaviour problem.
One change I've made (and I have no idea if it's good or not) is that for generic file syncing, a conflict resolved by "Keep Both" no longer move the local file to the app's folder (it should already be there).
I tested conflicts for file uploads, but not for file/folder syncing, and I have no idea how to do that.

@AndyScherzinger I consider the base development of this PR done. If you agree with me, I think it's time for design review!

### Discussion
In the conflict dialog, I removed the "Keep Server" option when the local version is supposed to be deleted, because the result is just the deletion of the local file. Please let me know if this makes sense, I'm still unsure of this.

ArisuOngaku:
Should existing auto upload folders have the "Upload existing files" enabled or disabled when updating the app? The current state is disabled, and I think this should stay like that to avoid taking action without the user knowing/agreeing.

AndyScherzinger:
> I rebased on master + fixed the local behaviour problem.

Nice 👍 

> One change I've made (and I have no idea if it's good or not) is that for generic file syncing, a conflict resolved by "Keep Both" no longer move the local file to the app's folder (it should already be there).
I tested conflicts for file uploads, but not for file/folder syncing, and I have no idea how to do that.

@tobiasKaminsky can you answer this for @ArisuOngaku ?

> @AndyScherzinger I consider the base development of this PR done. If you agree with me, I think it's time for design review!

Yes, sounds good to me, just loop in @jancborchardt and @nextcloud/designers when posting screenshots :) - I now flag it for review 👍 

> **Discussion**

> In the conflict dialog, I removed the "Keep Server" option when the local version is supposed to be deleted, because the result is just the deletion of the local file. Please let me know if this makes sense, I'm still unsure of this.

@tobiasKaminsky can you comment on this please?

> Should existing auto upload folders have the "Upload existing files" enabled or disabled when updating the app? The current state is disabled, and I think this should stay like that to avoid taking action without the user knowing/agreeing.

I agree with your assessment, what do you think @tobiasKaminsky @jancborchardt ?

jancborchardt:
> > Should existing auto upload folders have the "Upload existing files" enabled or disabled when updating the app? The current state is disabled, and I think this should stay like that to avoid taking action without the user knowing/agreeing.
> 
> I agree with your assessment, what do you think @tobiasKaminsky @jancborchardt ?

Agree on this as well, good call @ArisuOngaku

tobiasKaminsky:
> > In the conflict dialog, I removed the "Keep Server" option when the local version is supposed to be deleted, because the result is just the deletion of the local file. Please let me know if this makes sense, I'm still unsure of this.
> 
> @tobiasKaminsky can you comment on this please?

In which scenario is this happening? 
I remember that we named it "keep server" to have an explicitly description of what is happening. Previously there was no option and thus users were a bit left alone.
Removing this, will then lead again to this trap, or?
"Keep server" in this case does nothing more than just closing it, or? But it still is better understandable for the user…

Or did I understand you wrong?

tobiasKaminsky:
> > > Should existing auto upload folders have the "Upload existing files" enabled or disabled when updating the app? The current state is disabled, and I think this should stay like that to avoid taking action without the user knowing/agreeing.
> > 
> > 
> > I agree with your assessment, what do you think @tobiasKaminsky @jancborchardt ?
> 
> Agree on this as well, good call @ArisuOngaku

Yes, please have it as an "opti-ìn"  :+1:

ArisuOngaku:
> In which scenario is this happening?

@tobiasKaminsky When you set a synced folder to delete the local version on successful upload or when you upload a file and choose to delete the local version. Uploading a file with this setting and choosing "Keep server" :
- Should delete the local file (but this is counter intuitive imho, so I just disabled this option)
- Before this PR actually did not delete the local file, instead downloaded the remote version to overwrite the local version (I didn't fix this behavior because it cannot happen now)

Note that the "Keep Server" option is still available in other cases (keep local version or move it to the local Nextcloud folder).

Also, there is a 4th "hidden" option which is to close the dialog. @jancborchardt [conflict resolution dialog] I think it's important to give the user (when it's an upload) the choice to "cancel upload" (aka do nothing), and maybe another explicit button "resolve later" (which is basically the same as pressing the "back" navigation button = cancel the dialog but not the conflict/action). Would you recommend a separate PR to redesign that dialog or should we do it here?

ArisuOngaku:
Last force push : fixed a bug where clicking on a file conflict in the upload list didn't open the dialog and instead retried the upload. Now if it doesn't find the target remote file on the local cache, it will try to refresh the parent folder once and then retry before concluding that the remote file doesn't exist anymore.

ArisuOngaku:
I was going `static fields, static methods, fields, constructor, methods, inner classes`... What's the correct order?

ArisuOngaku:
@jancborchardt @nextcloud/designers Here is a full report on what this PR changes with (folded `▶`) screenshots.

# Auto upload existing files

<details>
<summary>
Auto upload configuration dialog<br>
---> New setting `Also upload existing files` on by default (except for existing enabled auto upload folders before app update).</summary>
Auto upload folder settings
</details>

When this setting is on, files that existed before the folder was enabled will be synced. Activating this setting on an already enabled folder after app update will also upload files that were previously missed.

# File conflict management

Instead of saving a renamed version of the file directly, a conflict is reported in the Uploads tab with a notification.

<details>
<summary>
Notification<br>
---> Clicking on the notification opens the Uploads tab.</summary>
Auto upload file conflict notification
</details>

<details>
<summary>
Uploads tab<br>
---> Clicking on the file conflict item opens the conflict resolution dialog.</summary>
Auto upload file conflict in Uploads tab
</details>

<details>
<summary>Conflict resolution dialog</summary>
Auto upload file conflict resolution
</details>

<details>
<summary>Note that if the upload is supposed to result in the deletion of the local file, the `Server version` option is not available.</summary>
Conflict resolution dialog no Server version
</details>

This way of managing file conflicts now also happens with manually uploaded files (from within or outside the app).

AndyScherzinger:
@tobiasKaminsky @ezaquarii can you please also have a look (code/behavior-wise)? I am quite blocked by private life at the moment 😢 and I think @ArisuOngaku did a great you getting this feature done ❤️

ArisuOngaku:
@AndyScherzinger There's actually a commit I want to split because it is not atomic, so I'm going to do that right now as quick as possible.

AndyScherzinger:
> There's actually a commit I want to split because it is not atomic, so I'm going to do that right now as quick as possible.

Thanks and no worries and no pressure! The changeset in total is probably still the same. Just wanted to ping dev folks early on so they have it on their radar since I am not of much help right now (which sucks)

ArisuOngaku:
Np, it's just that I would recommend reviewers to actually go through commits one by one because the last commits are heavy code cleanup that interferes with legibility on logic changes.

dgtlmoon:
how about a memory usage profile test where there is 5000-10000 photos/files? (typical load for modern phone storage and a reasonable use case in my opinion)

ArisuOngaku:
@dgtlmoon regarding the changes that have been made here, this would only be relevant on this PR to file upload conflicts listed in the Uploads tab. But 10000 or even 5000 file conflicts in this case is unlikely to happen.

However, this can be a good point to test this outside the context of this PR.

dgtlmoon:
yes - 

> However, this can be a good point to test this outside the context of this PR.

In other words, a big change in functionality will be merged to master, which could introduce a quite nasty crash for all users using the app... better to be safe than sorry

yes - I can say that i'm a very light phototaker with my smart devices and they all have atleast 5,000 or more photos... a real selfie maniac would have more :)

ArisuOngaku:
@dgtlmoon I'm sorry, I think I was not very clear. I answered too quickly.

- For file uploads, this PR does not introduce such a risk, but it still is relevant to the master branch which already has this risk (I said to test this outside this PR for this specific case, though this is probably outdated because:)
- For upload conflicts, your statement is totally relevant and should be tested here

jancborchardt:
Good stuff! :) Some design review:

> ### Auto upload configuration dialog
> ---> New setting `Also upload existing files` on by default (except for existing enabled auto upload folders before app update).
> Auto upload folder settings

Looking good. Only a detail: Is it normal that the checkboxes don’t align horizontally with the folder icon and toggle switch above?

-------------------

> ### File conflict management
> 
> Instead of saving a renamed version of the file directly, a conflict is reported in the Uploads tab with a notification.
> Notification
> ---> Clicking on the notification opens the Uploads tab.
> Auto upload file conflict notification

Could we have more informative notification text? "Upload failed" and "Could not upload `filename`" is a bit light on info. :)

Something like:
> **Conflict during auto upload**
> Pick which version to keep of `filename`

------------------------

> ### Uploads tab
> ---> Clicking on the file conflict item opens the conflict resolution dialog.
> Auto upload file conflict in Uploads tab

Looking good. But the icon on the right is the Delete trash can? Not sure if it’s obvious that tapping the entry will go for resolution. So maybe we should do a 3-dot menu rather than direct delete action, and in the 3-dot menu duplicate the "Resolve conflict" action and also have the "Delete" action.

That way Delete is not as easy to accidentally get to, and we also have "Resolve conflicts" written.


-----------------------


> ### Conflict resolution dialog
> Auto upload file conflict resolution

We need more info on both the files here to be able to decide which one to keep. See the conflict resolution dialog on the server, it gives you:
- thumbnails of both files so you can quickly see
- timestamp and file size of the files (ideally both, not sure why it currently doesn’t do the file you are uploading …)

Not saying that dialog is perfect, but thumbnails and the basic metadata go a long way already to decide whether it’s the same file, or a newer one, or a downscaled/cropped one, etc. to help inform your decision.
file conflict server

tobiasKaminsky:
The last "conflict dialog" is definitive out of scope, please open up a new issue.

ArisuOngaku:
@jancborchardt 

Auto upload folder settings fixed:
Auto upload folder settings

Notification strings changed (different title than yours though, because it also happens for non-auto upload conflicts):
Auto upload folder settings

Uploads list file conflicts three-dot menu:
Auto upload folder settings

jancborchardt:
Looking great! :+1: 

Agree with @tobiasKaminsky as well, a proper conflict resolution dialog should be a separate PR. :)

AndyScherzinger:
@tobiasKaminsky can we get this sheduled for 3.11? ❤️

AndyScherzinger:
@ArisuOngaku Do you have the time to rebase this, resolving its conflicts?

ArisuOngaku:
@AndyScherzinger yup I can do that give me an hour

AndyScherzinger:
> @AndyScherzinger yup I can do that give me an hour

@ArisuOngaku No hurries, it is not that time critical! (Since @tobiasKaminsky is working on the 3.10 RC3 at the moment!). So take all the time you need, no pressure!

ArisuOngaku:
@AndyScherzinger Don't worry, I said that because these were minor conflicts and I had some spare time.

slackfan:
I tested the latest QA build and this basically works. Which is great! The only flaw I found is that IF the feature is enabled on an existing shared folder and IF using subfolders during upload is enabled and IF the file is already on the server (same checksum), a Synchronisation error appears requesting to manually fix the issue, while ideally just nothing should be done. The workaround for this issue would be to enable the new "Also upload existing files" checkbox once, let everything already uploaded fail, and disable the checkbox again.

ArisuOngaku:
@slackfan thanks for your testing! :) As far as I know, the app does not do any checksum checking when comparing local+remote files. This missing feature would be relevant to the new file conflicts PR/issue that was previously suggested here. Unless I'm wrong or not up to date, this is working as intended in this scope.

AndyScherzinger:
Correct, neither client nor server do checksums

src/main/java/com/owncloud/android/jobs/ContactsBackupJob.java
```diff
@@ -159,18 +159,18 @@ private void backupContact(Account account, String backupFolder) {
...
+            account,
+            file.getAbsolutePath(),
+            backupFolder + filename,
+            FileUploader.LOCAL_BEHAVIOUR_MOVE,
+            null,
+            true,
+            UploadFileOperation.CREATED_BY_USER,
+            false,
+            false,
+            FileUploader.NameCollisionPolicy.ASK_USER
```
tobiasKaminsky: What happens if there is a conflict? 
As ContactBackupJob is run in background?

src/main/java/com/owncloud/android/jobs/FilesSyncJob.java
```diff
@@ -199,24 +203,24 @@ private void syncFolder(Context context, Resources resources, boolean lightVersi
                 remotePath = syncedFolder.getRemotePath();
             }
 
-            requester.uploadFileWithOverwrite(
-                    context,
-                    account,
-                    file.getAbsolutePath(),
-                    FileStorageUtils.getInstantUploadFilePath(
+            FileUploader.uploadNewFile(
```
tobiasKaminsky: Why did you changes this to a static call?
In general we should avoid static calls, as we cannot test them easily.
With requester as an object, we could simply mock this and response accordingly.

src/main/java/com/owncloud/android/ui/activity/ConflictsResolveActivity.java
```diff
@@ -25,59 +25,90 @@
...
+
+    private OCUpload conflictUpload;
+    private int localBehaviour = FileUploader.LOCAL_BEHAVIOUR_FORGET;
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+
+        if (savedInstanceState != null) {
+            this.conflictUpload = savedInstanceState.getParcelable(EXTRA_CONFLICT_UPLOAD);
```
tobiasKaminsky: Can you remove "this" here as it is not needed and complicates reading?

src/main/java/com/owncloud/android/ui/activity/ConflictsResolveActivity.java
```diff
@@ -25,59 +25,90 @@
...
-                intent.putExtra(FileDownloader.EXTRA_FILE, getFile());
-                startService(intent);
-                finish();
-                return;
-            default:
-                Log_OC.e(TAG, "Unhandled conflict decision " + decision);
-                return;
+        // Upload
+        if (decision == Decision.KEEP_LOCAL || decision == Decision.KEEP_BOTH) {
+            FileUploader.NameCollisionPolicy collisionPolicy = FileUploader.NameCollisionPolicy.OVERWRITE;
```
tobiasKaminsky: This is a bit confusing.
Can you change this so that we have two cases keep_local and keep_both?
I think it is not bad to have here duplicated code, if it helps readability and understanding.

If you want to have it in one if, then at least use
FileUploader.NameCollisionPolicy collisionPolicy;
            if (decision == Decision.KEEP_BOTH) {
                collisionPolicy = FileUploader.NameCollisionPolicy.RENAME;
            } else { 
 collisionPolicy = FileUploader.NameCollisionPolicy.OVERWRITE;
}

In this case it is also more clear what happens.

src/main/java/com/owncloud/android/ui/activity/ConflictsResolveActivity.java
```diff
@@ -25,59 +25,90 @@
...
+
+            FileUploader.uploadUpdateFile(this, getAccount(), file, localBehaviour, collisionPolicy);
+
+            if (this.conflictUpload != null) {
+                uploadsStorageManager.removeUpload(this.conflictUpload);
+            }
+        }
+
+        // Download
+        if (decision == Decision.KEEP_SERVER && !this.shouldDeleteLocal()) {
```
tobiasKaminsky: What is "shouldDeleteLocal" about? 
You refer to setting of File Uploader?

I would expect that there is no reference to an  "unknown"/not remembered setting, but instead that it downloads the file, if I choose "keep server".

src/main/java/com/owncloud/android/ui/activity/ConflictsResolveActivity.java
```diff
@@ -25,59 +25,90 @@
...
+            // Overwrite local file
+            Intent intent = new Intent(this, FileDownloader.class);
+            intent.putExtra(FileDownloader.EXTRA_ACCOUNT, getAccount());
+            intent.putExtra(FileDownloader.EXTRA_FILE, file);
+            if (this.conflictUpload != null) {
+                intent.putExtra(FileDownloader.EXTRA_CONFLICT_UPLOAD, this.conflictUpload);
```
tobiasKaminsky: Why do we need this?
Can't we remove upload like in L95: uploadsStorageManager.removeUpload(this.conflictUpload);

src/main/java/com/owncloud/android/jobs/ContactsBackupJob.java
ArisuOngaku: This method delegates the file upload job to another service, so it will behave exactly the same as any file upload. See FileUploader L930.

It would register a conflict and generate a notification.

src/main/java/com/owncloud/android/jobs/FilesSyncJob.java
ArisuOngaku: These are static methods from an OOP perspective so I made them actual static methods.

However I would prefer turning this into a builder pattern `FileUpload.Builder` if you're ok with that. Final method `upload` or `execute` which starts the intent and locks itself so it isn't accidentally made twice. Maybe call it `FileUpload.Request` but this is less straightforward imho.

src/main/java/com/owncloud/android/ui/activity/ConflictsResolveActivity.java
ArisuOngaku: Of course; should I remove it for method calls too?

src/main/java/com/owncloud/android/ui/activity/ConflictsResolveActivity.java
ArisuOngaku: For example, this is used with auto upload folders where the user chooses to delete the local version. In this case, no version of the file should be kept after the upload is complete.

So, if you choose "keep server", the local file should be deleted.

However I chose to disable this option in this specific case, see https://github.com/nextcloud/android/pull/4788#issuecomment-560359959

In the end, such a case doesn't happen (at the moment) and so I didn't implement the behavior of deleting the local file. But in no way should the remote file be downloaded.

Do you want me to add details about this case in the documentation of this method?

ArisuOngaku: I expect the "KEEP_SERVER" option to be used as a "delete local version" option in the future versions of the file conflict resolution dialog.

src/main/java/com/owncloud/android/ui/activity/ConflictsResolveActivity.java
ArisuOngaku: Well, tbh, the switch does look better with duplicated code. I took it the wrong way. I'll change that back to a switch.

src/main/java/com/owncloud/android/ui/activity/ConflictsResolveActivity.java
ArisuOngaku: I did this because if the file download fails, there is no way to retry it. So I only remove the upload (conflict) on download success.

AndyScherzinger:
@tobiasKaminsky fancy checking the failed test? This seems to be very unrelated to this PR :)

AndyScherzinger:
@tobiasKaminsky any way we get this in for the next feature release?

tobiasKaminsky:
Bug:
- have 3 images in Camera
- upload them
- install this branch
- enable "upload previously"
- enable upload for folder "Camera"
- see that three items as conflict in upload list view :+1: 
- click on one, select "keep server"
- still see three items
--> it should be two

ArisuOngaku:
> Bug:
> 
>     * have 3 images in Camera
> 
>     * upload them
> 
>     * install this branch
> 
>     * enable "upload previously"
> 
>     * enable upload for folder "Camera"
> 
>     * see that three items as conflict in upload list view +1
> 
>     * click on one, select "keep server"
> 
>     * still see three items
>       --> it should be two

This is a display bug, if you refresh up to date list is displayed.

`FileDownloader L645` and `UploadsStorageManager L238` -> UploadListAdapter doesn't register as an Observer of `UploadsStorageManager`. But there are multiple instances (see screenshot) of  `UploadsStorageManager` so it doesn't work anyway. I need your help @tobiasKaminsky on how you would make this class an effective singleton; else there is no way to properly update the view.

image

@tobiasKaminsky up on my answers to your comments btw.

tobiasKaminsky:
> This is a display bug, if you refresh up to date list is displayed.

True.
Let us figure this out in another issue/PR.

I will rebase it and then merge it :-)

tobiasKaminsky:
Thank you for this awesome PR!

ArisuOngaku:
You're welcome! :)

koying:
@ArisuOngaku Congrats for pushing this to the end :)

bastian-src:
I am using version 3.14 of the App and wonder why this feature is not included although it was already part of the 3.11 milestone?

AndyScherzinger:
@bastian-src just checked and the feature is there! Depending on your screen resolution you might have to scroll down in the auto upload config dialog (item is at the bottom) how to handle existing files in such a case and the 3rd checkbox from the top should say (upload existing).

So can you post a screenshot of the auto upload config dialog of your device?

bastian-src:
Sorry, my fault. The option is in the menu but I thought there would be another option/action to trigger the upload of existing files. So, I think the appropriated behaviour would be that it uploads the existing files right after clicking on **Save**?
In my case, another dialogue pops up which gives me a warning about the power saving (which I excluded the app from) and afterwards nothing happens. The manual file upload works fine for me.
Screenshot_20201124-102448_Nextcloud
Screenshot_20201124-102456_Nextcloud

AndyScherzinger:
Hmmm, the upload should happen within 15 minutes

bastian-src:
After 1h, the upload suddenly started. Thanks for your immediate response!

AndyScherzinger:
Sweet and glad to hear :)

PR #4789, opened 2019-11-03
Improve menu file name pattern

### LINKED ISSUE #4774: Improve menu file name pattern
Adapt naming pattern based on https://github.com/ribot/android-guidelines/blob/master/project_and_code_guidelines.md

#### 1.2.2.3 Menu files

Similar to layout files, menu files should match the name of the component. For example, if we are defining a menu file that is going to be used in the `UserActivity`, then the name of the file should be `activity_user.xml`

A good practice is to not include the word `menu` as part of the name because these files are already located in the `menu` directory.

AndyScherzinger:
Resolves #4774 

* correcting all menu file names _and_ 
* adding the corresponding guidelines chapter to `CONTRIBUTING.md` including a reference to the original Ribot guidelines.


CONTRIBUTING.md
```diff
@@ -236,6 +238,24 @@ Source code of app:
 -->
 ```
 
+## File naming
+
+The file naming patterns are inspired and based on [Ribot's Android Project And Code Guidelines](https://github.com/ribot/android-guidelines/blob/master/project_and_code_guidelines.md).
```
tobiasKaminsky: https://github.com/ribot/android-guidelines/blob/c1d8c9c904eb31bf01fe24aadb963b74281fe79a/project_and_code_guidelines.md

Please take this one, as it points to a commit, so that the content stays the same.
Otherwise we might have a diff, if in masters breaking changes occur.

src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
```diff
@@ -293,11 +293,7 @@ protected void onPause() {
     @Override
     public boolean onCreateOptionsMenu(Menu menu) {
         MenuInflater inflater = getMenuInflater();
-        inflater.inflate(R.menu.main_menu, menu);
-        menu.findItem(R.id.action_switch_view).setVisible(false);
-        menu.findItem(R.id.action_sync_account).setVisible(false);
-        menu.findItem(R.id.action_select_all).setVisible(false);
-        // menu.findItem(R.id.action_sort).setVisible(false);
+        inflater.inflate(R.menu.activity_folder_picker, menu);
```
tobiasKaminsky: Good idea!

src/main/res/layout/drawer.xml
```diff
@@ -31,7 +31,7 @@
         android:background="@color/bg_default"
         android:theme="@style/NavigationView_ItemTextAppearance"
         app:headerLayout="@layout/drawer_header"
-        app:menu="@menu/drawer_menu">
+        app:menu="@menu/partial_drawer">
```
tobiasKaminsky: I have a hard time understanding "partial", it is mainly the layout of the entries within the drawer and not the xml for other parts of the drawer, so maybe "@menu/drawer_entries"?

AndyScherzinger: Hard to say,since the names usually refer to the layout location, so this wouldn't fit any of the defined name patterns. So yeah, this one is a bit weird :(

AndyScherzinger: renamed to `partial_drawer_entries` (since it is still a "partial", naming pattern wise)

CONTRIBUTING.md
AndyScherzinger: DONE

AndyScherzinger:
@tobiasKaminsky please re-review. All requested changed have been done 😃 Thanks for the review ❤️

Issue #4757, opened 2019-10-29
Obey dark/light setting by OS


tobiasKaminsky:
In Android 10 one can enable dark mode in quick settings, this should be reflected by our app.
Something like:
Theme
- dark
- light
- determinated by system

tobiasKaminsky:
Nearly done. PR will be coming this week.

AndyScherzinger:
@tobiasKaminsky pllease be aware of https://github.com/nextcloud/android/pull/4797#discussion_r342792918

PR #4797, opened 2019-11-04
Replace BaseActivity.sharedPreferences with app preferences

### LINKED ISSUE #4791: Bypassing AppPreferences in BaseActivity
### Actual behaviour
`BaseActivity` uses `SharedPreferences` directly and cherry-picks private implementation details from `AppPreferencesImpl`.

### Expected behaviour
Application should use application abstraction `AppPreferences`.
`AppPreferencesImpl` is private implementation and shall not be touched.
 
### Steps to reproduce
This regression has been introduced in 1615130093d382211c3c84d33ff2b6206d5f5728 and c1aa005db929d08af4453489de603938e9d0c9c2

ezaquarii:
1. eliminate access to shared preferences when setting dark theme.
2. extend `AppPreferences` interface to allow subscription for preference changes
3. improve preferences interface documentation

@AndyScherzinger @tobiasKaminsky @dan0xii

This is my initial proposal to deal with `AppPreferences` interface bypass.
I tried to address access to `AppPreferencesImpl` but it looks like we still have
many points where we do cross-cutting calls to private implementation for various
reasons (dodgy workarounds, circular dependencies, etc). Despite my tries, I conclude
that closing `AppPreferencesImpl` is a bit too much work for a single change and I exhausted
my bandwidth for today.

This PR will address only theme related concerns. I'll follow up with improvements around
`*Impl` access as my time allows.

Fixes #4791


src/main/res/xml/preferences.xml
```diff
@@ -27,8 +41,9 @@
             android:title="@string/prefs_storage_path"
             android:key="storage_path"/>
         <com.owncloud.android.ui.ThemeableSwitchPreference
+            android:id="@+id/dark_theme_preference"
             android:defaultValue="@string/prefs_value_theme_light"
-            android:key="darkTheme"
+            android:key="ui_key_dark_theme_enabled"
```
AndyScherzinger: Now there are two different key for the theme!

AndyScherzinger:
In general it looks good while there are now two keys for the theme which should imho be just one.

src/main/res/xml/preferences.xml
AndyScherzinger: My mistake, just realized you intercept the persistence and use your internal key.

ezaquarii:
> In general it looks good while there are now two keys for the theme which should imho be just one.

Yeah,  I just checked the settings XML and it looks like Android is putting the UI key into the store,
which is not good. I'll roll this back as I have no good solution for this atm.

The rationale behind this experiment is that conceptually we have 1 key somewhere behind `AppPreferences` and **nobody** should touch it in any way other than calling strongly typed API. This is on purpose:

1. people should not liberally write into the `SharedPreferences` bypassing the interface we provide, with all sort of potential sanitation and safeguards; we'd like to discoruage people from fiddling with the store on their own, so we can safely evolve this area in the future.
2. UI should not be concerned how or where we store app settings - we might find that JSON stored on Nextcloud server is the way to go at some time and I'd argue that this should be the goal eventually - I have some vested interest here, as I'm kinda tired of reconfiguring my autoupload each time I reinstall the app (;
3. tracking preferences access via interface is as easy as tracking a method call; dictionary keyed with raw strings - not so fun

Currently keys must be maintained in sync in 3 distinct places:
1. `AppPreferencesImpl` - to access the store
2. `settings.xml` - so the UI has backdoor into the store layer
3. `SettingsActivity` - so we can access UI controls

BTW, those sets are not 100% overlapping.

Now, you could argue that we can just move those keys to `AppPreferences` interface to unify 1. and 2., but my fear is that we again start seeing stuff like this:

```
void outOfTheBlue() {
    SharedPreferences p = MainApp.getInstance().getSharedPreferences(); // circular deps again
    p.edit().setString(AppPrefernces.PREFS__DODGY_KEY, "magic value").apply(); // out of band calls

    MainApp.getInstance().getSharedPreferences().setBoolean(AppPrefernces.PREFS__WORKAROUND_FLAG, !MainApp.getInstance().getSharedPreferences().setBoolean(AppPrefernces.PREFS__WORKAROUND_FLAG, false));
}
```

and we just got rid if this stuff recently.

src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
```diff
@@ -68,7 +78,7 @@
     private static final String PREF__AUTO_UPLOAD_INIT = "autoUploadInit";
     private static final String PREF__FOLDER_SORT_ORDER = "folder_sort_order";
     private static final String PREF__FOLDER_LAYOUT = "folder_layout";
-    public static final String PREF__THEME = "darkTheme";
+    static final String PREF__DARK_THEME_ENABLED = "dark_theme_enabled";
```
AndyScherzinger: Will likely chance when @tobiasKaminsky pushes his PR to the repo adding a 3rd option for the theme values (dark, light, system)

AndyScherzinger:
Fine with me. Just one minor comment that this PR and a soon to be openend PR by @tobiasKaminsky will have conflicts while both PRs will change the type of the preference and thus (fine since master is in alpha state for v3.10) both PRs will break the activated dark theme (since the key changed twice then)

src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
ezaquarii: This will be interesting. I wild guess that the change will be based on the `SharedPreferences` code that I just removed and the preference will become some kind of `enum` instead of `boolean`?

I'm fine with resolving conflicts on my side, let's just make sure we don't have to do complete rewrite. (-:

Is there any public branch I could potentially try?

AndyScherzinger: > Is there any public branch I could potentially try?

Don't know, I just saw this earlier this evening, which is why I mentioned @tobiasKaminsky here --> https://github.com/nextcloud/android/issues/4757#issuecomment-549990621

tobiasKaminsky:
We merge this one, and then I change it within my upcoming PR.
Sometimes it will just happen that we create conflicts…

PR #4819, opened 2019-11-11
Custom spotbugs rules to detect restricted API calls and impl abuse

ezaquarii:
This is experimental try to verify Spotbugs suitability to detect common
mistakes specific to Nextcloud project.

2 experimental rules are introduced:
* `NXRAC` - Restricted API Call - checks for access to low-level Android platform APIs that have Nextcloud specific wrappers / equivalents
* `NXPIC` - Private Implementation Call - checks for direct calls to `*Impl` that correlate with bypassing dependency injection mechanism

Since SpotBugs works on JVM bytecode, rules are quite low level and ruels maintenance require some level of JVM bytecode understanding.

Kotlin is not supported and would require a parallel plugin for `Detekt`. Since Kotlin is used in just a tiny fracion of our codebase, it is not an immediate priority to develop such co-plugin. On a positive note, `Detekt` analysis is based on Abstract Syntax Tree, so it is easier to do.

Rule violations are given the highest possible priority, as - contrary to other, lesser quality issues, such as locking or `null`ability - design defects are rather expensive to fix.

It is an open question if the cost of implementation and maintenance of a custom extention for our static analyzers is worth the effort if cheaper alternatives are possible, such a manual review process supported by `grep | wc -l`-based notification and a warning script in GitHub. I'm happy to have a discussion - this plugin is more of a proof of concept than a comprehensive solution.


tobiasKaminsky:
This is really cool and I think it is worth the effort.
Hopefully we will not have that much custom rules.

It seems that also auto-generated files, like dagger, are verified/found:
`Bug type NEXTCLOUD_PRIVATE_IMPL_CALL (click for details)
In class com.nextcloud.client.account.UserAccountManagerImpl_Factory
In method com.nextcloud.client.account.UserAccountManagerImpl_Factory.newInstance(Context, AccountManager)
At UserAccountManagerImpl_Factory.java:[line 35]
`
which is something we cannot prevent, or?

But this is a valid finding:
`
Bug type NEXTCLOUD_PRIVATE_IMPL_CALL (click for details)
In class com.owncloud.android.jobs.MediaFoldersDetectionJob$NotificationReceiver
In method com.owncloud.android.jobs.MediaFoldersDetectionJob$NotificationReceiver.onReceive(Context, Intent)
At MediaFoldersDetectionJob.java:[line 250]
`

tobiasKaminsky:
@ezaquarii in addition: do you think it is worth to have this in a separate repo, so that it could also be used for library?

We then also should move to a deploy system (like jitpack, if possible) so that we do not have a binary dependency in here.

tobiasKaminsky:
As this is not yet finished, I revert to "1. developing".

ezaquarii:
@tobiasKaminsky I was thinking about this PR and I have some doubts that it delivers the value for comitted effort.

Although automating this is a good thing, I'm worried about false-positives and general maintenance effort required to keep this plugin going. Not to mention that Detekt plugin needs to be created as well.

So, instead of maintaining full-blown analyzer plugin, I'm thinking about a simpler approach.

What if we just scan the diff looking for specific regexp and then mark such PR in a way visible by the reviewers? Something like "This PR contains calls to restricted API. Please review carefully".

Does GitHub allow us some flexibility to add extra review step?

ezaquarii:
Interesting problem, but there is no bandwidth to do it properly. I'm closing it for now.

PR #4859, opened 2019-11-19
Upload folder support

### LINKED ISSUE #184: Upload whole folder

kilves:
https://github.com/nextcloud/android/pull/4505 develeper stopped updating his PR, so here's an alternate solution that should implement the requirements mentioned in old PR.

Fixes #184


kilves:
IT test fail might be unrelated, the exact same error has happened before on a completely differnet PR: https://kaminsky.me/nc-dev/android-integrationTests/10775/index.html

AndyScherzinger:
hi @kilves,

first of all welcome to the Nextcloud Android client project and thank for taking an interest in it and for picking up this task ❤️ 

Tests are actually a bit flaky, so no worries. Can you do a rebase against the latest master please. See https://github.com/nextcloud/android/pull/4859#issuecomment-557289432 which shows an increase in linbt warnings (breaking the build) which might be due to the not-up-to-date master branch of yours (or needs fixing, hard to tell right now). The rebase would also remove all changes not made within this PR itself 😃 

cc @tobiasKaminsky for review

AndyScherzinger:
Also cc @ezaquarii for review :)

src/main/java/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -482,7 +487,8 @@ public void onCheckAvailableSpaceFinish(boolean hasEnoughSpaceAvailable, String.
 
                 preferences.setUploaderBehaviour(FileUploader.LOCAL_BEHAVIOUR_MOVE);
             } else {
-                data.putExtra(EXTRA_CHOSEN_FILES, mFileListFragment.getCheckedFilePaths());
+                data.putExtra(EXTRA_CHOSEN_FILES, (ArrayList<File>)mFileListFragment.getCheckedFilePaths());
```
ezaquarii: We should not perform type casting. Can we change `getCheckedFilePaths()` return type to explicitly return `ArrayList<>`? There should be no problem type-wise when narrowing the return type.

src/main/java/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -659,7 +665,7 @@ public void onConfirmation(String callerTag) {
             // return the list of selected files to the caller activity (success),
             // signaling that they should be moved to the ownCloud folder, instead of copied
             Intent data = new Intent();
-            data.putExtra(EXTRA_CHOSEN_FILES, mFileListFragment.getCheckedFilePaths());
+            data.putExtra(EXTRA_CHOSEN_FILES, (ArrayList<File>)mFileListFragment.getCheckedFilePaths());
```
ezaquarii: not type casting pls

src/main/java/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -703,6 +709,41 @@ private ExtendedListFragment getListOfFilesFragment() {
...
+        String targetPath = targetDirectory.getAbsolutePath();
+        String currentPath = currentDirectory.getAbsolutePath();
+        if (targetPath.startsWith(currentPath)) {
+            return new File(targetPath.replace(targetPath.replace(currentPath, ""), ""));
+        }
+        return traverseToCommonAncestor(targetDirectory, currentDirectory.getParentFile());
+    }
+
+    /** Tries to guess where replication root is by using common ancestor of all files */
+    private String determineReplicationRootDir() {
```
ezaquarii: `getReplicationRootDir()`. Verb vocabulary is pretty set in stone and people have expecations. Verbs such as "perform", "obtain", "materialize" (yeah, seriously), "determine", "establish", etc, are not helping readability.

src/main/java/com/owncloud/android/ui/adapter/LocalFileListAdapter.java
```diff
@@ -163,7 +156,7 @@ public void onBindViewHolder(@NonNull RecyclerView.ViewHolder holder, int positi
                 setThumbnail(file, gridViewHolder.thumbnail);
 
                 if (file.isDirectory()) {
-                    gridViewHolder.checkbox.setVisibility(View.GONE);
+                    // gridViewHolder.checkbox.setVisibility(View.GONE);
```
ezaquarii: let's remove dead code and dead branch

src/main/java/com/owncloud/android/ui/asynctasks/CheckAvailableSpaceTask.java
```diff
@@ -41,6 +43,15 @@ public CheckAvailableSpaceTask(CheckAvailableSpaceListener callback, String... p
         this.callback = callback;
     }
 
+    public CheckAvailableSpaceTask(CheckAvailableSpaceListener callback, List<File> files) {
```
ezaquarii: `Collection<File>` is more sutable, as we don't use random access API here.

src/main/java/com/owncloud/android/ui/activity/UploadFilesActivity.java
kilves: Automatic QA complained about using it as a return type of `getCheckedFilePaths()` so that's why I made this change. It'll probably complain anyway even with this though. Will refactor to something more sane.

ezaquarii: Than this is automatic QA bug ("false positive" as we call it), not your fault.

We should not be cargo-culting static code anslysis. It's just a dumb automaton after all.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1020,57 +1032,166 @@ private void exitSelectionMode() {
...
+
+            default:
+                behaviour = FileUploader.LOCAL_BEHAVIOUR_FORGET;
+                break;
+        }
+
+        FileUploader.UploadRequester requester = new FileUploader.UploadRequester();
+        requester.uploadNewFile(
+            this,
+            getAccount(),
```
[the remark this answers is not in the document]
ezaquarii: BTW, you'll probably need to rebase on master to have it :)

kilves:
Ok, so I did some refactoring and the code should be more readable now. I also moved the logic to separate classes. I'm not sure about unit tests in this case. The code is still largely dependent on the existence of a context, because you still need intents to create folders before uploading files. So it would really be an integration test. I can write some though if they are really necessary. I'm also open to suggestions about `FilesUploader` class if there's not enough decoupling for example.

src/main/java/com/owncloud/android/files/FilesUploader.java
```diff
@@ -0,0 +1,238 @@
...
+
+        FileUploader.UploadRequester requester = new FileUploader.UploadRequester();
+        Account account = user.toPlatformAccount();
+
+        String[] localPaths = getLocalFilePaths(localFiles).toArray(new String[]{});
+        String[] remotePaths = getRemoteTargetPaths(remoteTargets).toArray(new String[]{});
+
+        requester.uploadNewFile(
+            fileActivity,
+            account,
```
ezaquarii: Let's not create `Account` variables.

We're running a continuous refactoring effort where all calls to `toPlatformAccount()` are being chased. Because we have tons of it, it's much easier to browse the analysis report if it's called in point of use rather than spilled into variables, as IDE can tell us how many items we have in each class.

For the record, `toPlaformAccount()` will be retired in a week or 2. Hopefully.

src/main/java/com/owncloud/android/files/FilesUploader.java
```diff
@@ -0,0 +1,238 @@
...
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import androidx.annotation.NonNull;
+
+public class FilesUploader {
+
+    private FileActivity fileActivity;
```
ezaquarii: I don't see any `Activity` APIs being used, so we shall not depend on it.

src/main/java/com/owncloud/android/files/FilesUploader.java
```diff
@@ -0,0 +1,238 @@
...
+            // Folder already being created, don't create again
+            return;
+        }
+        this.foldersToAutoCreate.add(folder);
+        Intent intent = new Intent(fileActivity, OperationsService.class);
+        intent.setAction(OperationsService.ACTION_CREATE_FOLDER);
+        intent.putExtra(OperationsService.EXTRA_ACCOUNT, user.toPlatformAccount());
+        intent.putExtra(OperationsService.EXTRA_REMOTE_PATH, folder);
+        intent.putExtra(OperationsService.EXTRA_CREATE_FULL_PATH, true);
+        fileActivity.getOperationsServiceBinder().queueNewOperation(intent);
```
ezaquarii: Few issues here:
1. we inject `FileActivity` only to call a getter.
2. 2 statements per line

We shall inject the `OperationsServiceBinder` instead.

src/main/java/com/owncloud/android/files/FilesUploader.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        FileUploader.UploadRequester requester = new FileUploader.UploadRequester();
+        Account account = user.toPlatformAccount();
+
+        String[] localPaths = getLocalFilePaths(localFiles).toArray(new String[]{});
+        String[] remotePaths = getRemoteTargetPaths(remoteTargets).toArray(new String[]{});
+
+        requester.uploadNewFile(
+            fileActivity,
```
ezaquarii: `Context` is needed here.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -286,6 +288,7 @@ protected void onCreate(Bundle savedInstanceState) {
         }
 
         mPlayerConnection = new PlayerServiceConnection(this);
+        filesUploader = new FilesUploader(this);
```
ezaquarii: Constructor-inject all dependencies:

1. `Context` from `getApplicationContext()`
2. `OperationsServiceBinder`

src/main/java/com/owncloud/android/files/FilesUploader.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    /**
+     * Call this from activity when "create folder" operation finishes
+     * @param operation operation
+     * @param result    result
+     * @param user      current user
+     * @return          true if this operation was consumed, false otherwise
+     */
+    public boolean onCreateFolderOperationFinish(CreateFolderOperation operation,
```
ezaquarii: This name (and I know it's a delegate from legacy code, so no blames) suggests that we're creating some mysterious folder operation finish (a finish event?).

JavaDoc tells that it's `onFinish<what/noun>` callback.

Let's rename both methods.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -218,6 +219,8 @@
     private PlayerServiceConnection mPlayerConnection;
     private Account mLastDisplayedAccount;
 
+    private FilesUploader filesUploader;
```
ezaquarii: This component name suggests that it uploads files.

The UI can be destroyed at any given time - what happens with the files being uploaded by `FilesUploader`? Does it cancel?

I see 2 options:

1. the design in broken
2. this is not an uploader :)

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1014,62 +1017,37 @@ public void run() {
...
-                remotePaths,
-                null,           // MIME type will be detected from file name
-                behaviour,
-                false,          // do not create parent folder if not existent
-                UploadFileOperation.CREATED_BY_USER,
-                false,
-                false
-            );
+    private void uploadFromFileSystem(List<File> files, int resultCode) {
+        filesUploader.uploadFromFileSystem(files, resultCode, null, getCurrentDir(),
```
ezaquarii: When args list is too long, we make it vertical to boost readability.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1014,62 +1017,37 @@ public void run() {
...
+            DisplayUtils.showSnackMessage(this, R.string.filedisplay_no_file_selected);
+            return;
+        }
 
-                default:
-                    behaviour = FileUploader.LOCAL_BEHAVIOUR_FORGET;
-                    break;
-            }
+        String replicationStartDir = data.getStringExtra(UploadFilesActivity.EXTRA_REPLICATION_START_DIR);
+        filesUploader.uploadFromFileSystem(files, resultCode, replicationStartDir, getCurrentDir(),
```
ezaquarii: When args list is too long, we make it vertical to boost readability.

src/main/java/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -534,6 +536,40 @@ public static boolean checkEncryptionStatus(OCFile file, FileDataStorageManager
...
+        }
+        return currentDirectory.getAbsolutePath() + File.separator;
+    }
+
+    /** Traverses currentDirectory upward until a common ancestor with targetDirectory is reached, and returns it */
+    private static File traverseToCommonAncestor(@NonNull File targetDirectory, @NonNull File currentDirectory) {
+        String targetPath = targetDirectory.getAbsolutePath();
+        String currentPath = currentDirectory.getAbsolutePath();
+        if (targetPath.startsWith(currentPath)) {
+            return new File(targetPath.replace(targetPath.replace(currentPath, ""), ""));
```
ezaquarii: It's generally bad practice to have more than 1 statement per line.
We have 3 here.

src/main/java/com/owncloud/android/files/FilesUploader.java
```diff
@@ -0,0 +1,238 @@
...
+            foldersToAutoCreate.remove(createdPath);
+            if (foldersToAutoCreate.isEmpty()) {
+                completeFileUploadOperation(user);
+            }
+            return true;
+        }
+        return false;
+    }
+
+    private List<UploadTarget> getRemoteTargets(@NonNull List<File> localFiles,
```
ezaquarii: Javadoc pls

src/main/java/com/owncloud/android/files/FilesUploader.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    private boolean isCreatingFolders() {
+        return !this.foldersToAutoCreate.isEmpty();
+    }
+
+    private void setFilesToUpload(List<File> localFiles, List<UploadTarget> remoteTargets, int resultCode) {
+        this.localFiles = localFiles;
+        this.remoteTargets = remoteTargets;
+        this.resultCode = resultCode;
```
ezaquarii: is this expected result code?

src/main/java/com/owncloud/android/files/FilesUploader.java
```diff
@@ -0,0 +1,238 @@
...
+            case UploadFilesActivity.RESULT_OK_AND_DO_NOTHING:
+                behaviour = FileUploader.LOCAL_BEHAVIOUR_FORGET;
+                break;
+
+            default:
+                behaviour = FileUploader.LOCAL_BEHAVIOUR_FORGET;
+                break;
+        }
+
+        FileUploader.UploadRequester requester = new FileUploader.UploadRequester();
```
ezaquarii: This should be injected as a dependency.

src/main/java/com/owncloud/android/ui/asynctasks/CheckAvailableSpaceTask.java
```diff
@@ -41,6 +43,15 @@ public CheckAvailableSpaceTask(CheckAvailableSpaceListener callback, String... p
         this.callback = callback;
     }
 
+    public CheckAvailableSpaceTask(CheckAvailableSpaceListener callback, Collection<File> files) {
```
ezaquarii: It would be beneficial to kill the legacy constructor that uses `String[]`. Is it feasible?

src/main/java/com/owncloud/android/files/FilesUploader.java
```diff
@@ -0,0 +1,238 @@
...
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import androidx.annotation.NonNull;
+
+public class FilesUploader {
```
ezaquarii: Could you please supply JavaDoc for this component? 

Overall, while order of magnitude better, I still have a sneaky suspicion that the design does not match what I'd find in this JavaDoc, but I'm not yet ready to jump into conclusions.

JavaDoc would help us to pinpoint it, I believe.

ezaquarii: Also, have you considered converting it into Kotlin? Java is no longer default language, according to google.

src/main/java/com/owncloud/android/files/FilesUploader.java
```diff
@@ -0,0 +1,238 @@
...
+
+    private FileActivity fileActivity;
+
+    private Set<String> foldersToAutoCreate = new HashSet<>();
+    private List<UploadTarget> remoteTargets = new ArrayList<>();
+    private List<File> localFiles = new ArrayList<>();
+    private int resultCode;
+
+    public static class UploadTarget {
```
ezaquarii: In Kotlin, this would be 1 line data class. :)

src/main/java/com/owncloud/android/files/FilesUploader.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    /**
+     * Requests an upload from local filesystem.
+     * @param localFiles            Local files to upload
+     * @param resultCode            result code
+     * @param currentDir            the directory user currently is in
+     * @param user                  current user
+     */
+    public void uploadFromFileSystem(List<File> localFiles,
```
ezaquarii: JavaDoc tells me it's places a request.
Name tells me it uploads.

Shall we bring the function name on par with the description?

src/main/java/com/owncloud/android/files/FilesUploader.java
ezaquarii: Hm... this is suspicious.

src/main/java/com/owncloud/android/files/FilesUploader.java
kilves: Yeah, the whole design is a bit spotty. The class is named `FilesUploader` but it doesn't really upload anything. It's more of a manager for two intents. Essentially the context gives it a bunch of files and where it wants them to go, and it starts an Intent to create folders for those files. After that the context notifies it that the process is done. Then it finally asks another class to upload files.

I smell some problems in the design as well, but I can't put my finger on it. Do you have any suggestions for making this into something more sensible? A service perhaps?

kilves:
@ezaquarii did you delete the long comment chain about javadocs and refactoring?

ezaquarii:
> @ezaquarii did you delete the long comment chain about javadocs and refactoring?
Sorry, it looks like my review was published before I finished it. This one is final.

tobiasKaminsky:
@ezaquarii i totally agree with your analysis, but fear that there is (at least from my side) no time in nearer future to refactor this.

So I would go for 1b as a quick win (then we also do not have Android warning if UI takes too long) and we can also add this enhancement.

Meanwhile we still can & should discuss how to refactor this (in a new issue) with lots of smaller steps, which can be easily done and tested.

AndyScherzinger:
I'd also say for now 1b seems to be feasable given the circumstances.

tobiasKaminsky:
I switched label back to "1. developing", please change it once it is ready for review

AndyScherzinger:
superseeded by #7877

PR #4885, opened 2019-11-25
Disable android backup on the QA app

ArisuOngaku:
Sometimes, there are multiple versions of the QA app that share the same database versions for different migrations. When the android backup feature is enabled by the user's devices, this leads to invalidly restored database state and then crashes that would normally not happen in production.

This PR simply disables the android backup feature for the QA version of the app.

(as seen in #4788)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -105,7 +105,12 @@ android {
         abortOnError false
         htmlReport true
         htmlOutput file("$project.buildDir/reports/lint/lint.html")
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
I just disabled one lint check.

Thanks for the fix and the idea!

ArisuOngaku:
No problem!

AndyScherzinger:
All checks ✅ 

merging...

❤️ Thanks for the fix @ArisuOngaku 🥇

Issue #4852, opened 2019-11-18
Files list doesn't refresh after file restore


Shagequi:
Hi,

When you restore a file from "deleted files" (the trashbin), and return in "all files" Activity, the restored folder/file is not directly displayed, because the list is not refreshed after restore operation.

So, we need to manually reload the list. Is it a wanted behaviour or is it really a bug ?

Thanks :)

App Version : 3.9.0 , all devices, Nextcloud
Server Version : 17.0.1 - 17.0.1.1 (Nextcloud)

AndyScherzinger:
Cc @tobiasKaminsky AFAIK a bug since the list should get a different etag for the folder from the server and refresh the list.

Shagequi:
So, it could be relative to this issue too : https://github.com/nextcloud/android/issues/4851 ?

AndyScherzinger:
Maybe... it might also be related to `onResume` or something like that

AndyScherzinger:
fixed via #4898

Issue #4851, opened 2019-11-18
Folder size not refreshed after file moving


Shagequi:
Hi,

When I create a folder, move a file in it and go back to the main activity (where the folder has been initialy created), the folder size is not refreshed (although the file is well inside).

So, I think we could refresh the "folder" (file) onResume by refreshing only the folder during the `onResume`.

Steps : 
1) Create a folder where you want, name it like you want
2) Move a file in the created folder
3) Return where you were during the folder creation (so, go back through one level)
4) You can see the folder size at 0.0MB (despite the file in it)

**Version : 3.9.0 for Android app (NextCloud)**

AndyScherzinger:
Sounds right to me, cc @tobiasKaminsky for feedback

AndyScherzinger:
fixed via #4898

PR #4898, opened 2019-11-28
Refresh issue for file movment/file restore operations

siddharths2710:
The pull request fixes related issues  #4851 and #4852, and was achieved by invoking refresh list method on resume, and the same for file move operation result since the **onResume** method of `FileDisplayActivity` isn't touched during the file movement operation.

Tested on Android versions `6.0.1` and `8.1.0`

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -1212,6 +1212,7 @@ public class FileDisplayActivity extends FileActivity
     protected void onResume() {
         Log_OC.v(TAG, "onResume() start");
         super.onResume();
+        refreshList(true);
 
         OCFile startFile = null;
         if (getIntent() != null && getIntent().getParcelableExtra(EXTRA_FILE) != null) {
@@ -1884,6 +1885,7 @@ public class FileDisplayActivity extends FileActivity
                                            RemoteOperationResult result) {
         if (result.isSuccess()) {
             refreshListOfFilesFragment(false);
+            refreshList(true);
         } else {
             try {
                 DisplayUtils.showSnackMessage(
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -321,7 +321,7 @@ public class FileDisplayActivity extends FileActivity
             }
         } else {
             createMinFragments(savedInstanceState);
-            refreshList(true);
+            syncAndUpdateFolder(true);
         }
 
         setIndeterminate(mSyncInProgress);
@@ -394,7 +394,7 @@ public class FileDisplayActivity extends FileActivity
                         && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
                     // permission was granted
                     EventBus.getDefault().post(new TokenPushEvent());
-                    refreshList(true);
+                    syncAndUpdateFolder(true);
                     // toggle on is save since this is the only scenario this code gets accessed
                 } else {
                     // permission denied --> do nothing
@@ -685,7 +685,7 @@ public class FileDisplayActivity extends FileActivity
         updateActionBarTitleAndHomeButton(null);
     }
 
-    public void refreshListOfFilesFragment(boolean fromSearch) {
+    public void updateListOfFilesFragment(boolean fromSearch) {
         OCFileListFragment fileListFragment = getListOfFilesFragment();
         if (fileListFragment != null) {
             fileListFragment.listDirectory(MainApp.isOnlyOnDevice(), fromSearch);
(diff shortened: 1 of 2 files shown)


siddharths2710:
@AndyScherzinger and @Shagequi  please tell me if I have to [Create another pull request](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#3-create-another-pull-request) as well.

AndyScherzinger:
@siddharths2710 no need for another pull request except when you want to do another fix or feature.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1884,6 +1885,7 @@ private void onMoveFileOperationFinish(MoveFileOperation operation,
                                            RemoteOperationResult result) {
         if (result.isSuccess()) {
             refreshListOfFilesFragment(false);
+            refreshList(true);
```
ezaquarii: The difference between `refreshList` and `refreshListOfFilesFragment()` is not obvious.  While being here, can we `Shift-F6` the clarity here?

siddharths2710: @ezaquarii from my understanding of the code, the `refreshListOfFilesFragment` mainly takes care of layout updates and view changes. It mainly boils down to the `listDirectory` method defined in `OCFileListFragment` and an extra method for setting up the toolbar.  

`refreshList` however performs the remote operation `RefreshFolderOperation` explicitly, so that any changes to the list and file properties are accommodated, along with updating the local database. 

I've set the argument for `refreshList` to `true` since I observed that the etag of the folder was not getting changed, and it gets updated only after invocation of `refreshList`.

tobiasKaminsky: From my understanding it then should be enough to cal refreshList(true);

Can you test this?

And we can rename 
- refreshListOfFilesFragment to updateListOfFilesFragment
- refreshList to syncAndUpdateFolder()

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1212,6 +1212,7 @@ protected void onSaveInstanceState(Bundle outState) {
     protected void onResume() {
         Log_OC.v(TAG, "onResume() start");
         super.onResume();
+        refreshList(true);
```
tobiasKaminsky: In both cases: are you sure that we need "true" to ignore eTag?
As folder content changed, also eTag should change and thus we can rely on it?

siddharths2710: @tobiasKaminsky I'm not sure why, but for the file movement operation the refresh works only if `ignoreETag` is set to true.
I just tested the file restoration case now ( the one in `onResume` method ) and it seems to work if set to `false`. So I can make a new commit for this case if required.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
siddharths2710: @tobiasKaminsky Yes, calling `refreshList` alone seems to do the job.
Please give the go ahead for a new commit.

tobiasKaminsky: You do not have to ask :-)
Just do it

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
tobiasKaminsky: Hm. That is strange.
Can you describe what is not working?
As when I am in a folder and move a file, this file gets removed for me.

siddharths2710: @tobiasKaminsky I meant to say that when I test the file movement operation with `ignoreETag` set to false, I do not observe the change in the folder size, but the change is reflected if i switch the value to true instead.

Does the file get completely removed for you after the movement operation? It occurs like this for me:

siddharths2710: ezgif com-video-to-gif

siddharths2710: Please ignore the alignment. It was caused by the running video recorder.

tobiasKaminsky: > @tobiasKaminsky I meant to say that when I test the file movement operation with `ignoreETag` set to false, I do not observe the change in the folder size, but the change is reflected if i switch the value to true instead.

Then let us set it to true :+1:

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
ezaquarii: > `refreshList` however performs the remote operation `RefreshFolderOperation` explicitly

`requestRefreshFolder` is more suitable then. This method does not refresh anything as it only launches async op, returns immediately and remaining flow is handled in callbacks.

siddharths2710: @ezaquarii I've already updated the name for `eefreshList` to `syncAndUpdateFolder`. Do you suggest I rename it to `startFolderRefresh` instead?

tobiasKaminsky:
Thank you and :tada: for your first PR here :tada:

PR #4909, opened 2019-12-02
update Login testcase

dassio:
update Login testcase


tobiasKaminsky:
Thanks for taking care of this.
I fear that those tests are not yet executed, as they are in /automationTest and with our new package scheme, they must be in /src/androidTest

When I moved them over, I do get a AndroidDriver not found.
Maybe something on build.gradle is missing?

(I set label to "1. to developing", if you think it is ready to review, just switch it to "2. to review").

If you need any help meanwhile, just ping me :-)

dassio:
the testcase is using appium instead of Androidx , that's why it is not in the androidTest folder.
i guess the code is written in the owncloud age and some choose Appium instead of espresso. i am thinking of refactoring this one first for blackbox testing , maybe later we can add espresso test.

here is a link of the two tools' comparision :https://www.pcloudy.com/appium-vs-espresso-the-most-popular-automation-testing-framework-in-2019/

dassio:
you can run the test with appium and emulator running by typing :
`mvn -Dtest=LoginTestSuite.java#test1LoginPortrait  test`

tobiasKaminsky:
From the blog post I do not see that we need the advantages of appium (iOS testing, no access to source code needed).

As we already have (too few) Espresso tests running, maybe this is the way to go?
Some times ago I already created a simple login UI test: https://github.com/nextcloud/android/pull/4146

dassio:
true, but it is an option, and it is working. and we don't have to choose here, we can have both. 
i don't see the point of leave some malfunctioning test code lying  there, just want to clean up.

tobiasKaminsky:
My concern was mainly about the additional dev time needed to get this up and running in our CI (drone.yml).

dassio:
i will try to get it up and running, i will first try my own CI environment

PR #4926, opened 2019-12-03
remove all checks for <= JELLY_BEAN as we have this as our new minimum.

tobiasKaminsky:

### DIFF
--- a/src/main/java/com/owncloud/android/ui/ThemeableSwitchPreference.java
+++ b/src/main/java/com/owncloud/android/ui/ThemeableSwitchPreference.java
@@ -23,7 +23,6 @@ package com.owncloud.android.ui;
 import android.content.Context;
 import android.content.res.ColorStateList;
 import android.graphics.Color;
-import android.os.Build;
 import android.preference.SwitchPreference;
 import android.util.AttributeSet;
 import android.view.View;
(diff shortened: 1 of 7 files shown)


src/main/java/com/owncloud/android/ui/ThemeableSwitchPreference.java
```diff
@@ -58,12 +56,9 @@ public ThemeableSwitchPreference(Context context, AttributeSet attrs) {
     protected void onBindView(View view) {
         super.onBindView(view);
 
-        if (view instanceof ViewGroup && Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
-            findSwitch((ViewGroup) view);
-        }
+        findSwitch((ViewGroup) view);
```
AndyScherzinger: if (view instanceof ViewGroup) would still be valid, no?

src/test/java/com/nextcloud/client/device/TestPowerManagementService.kt
```diff
@@ -175,47 +175,6 @@ class TestPowerManagementService {
             }
         }
 
-        @Test
-        fun `battery charging status on API 14-16`() {
```
AndyScherzinger: We are targeting  >= 16 so the tests are still valid, since we still support 16 (!)

AndyScherzinger:
@tobiasKaminsky just 2 minor comments (which need fixing imho)

tobiasKaminsky:
@AndyScherzinger I changed both, thanks for finding them :-)

PR #4932, opened 2019-12-04
Use android test orchestrator

tobiasKaminsky:
This should make it a bit more robust, e.g. no context is shared between different tests and if one test fails the other will still be executed.


AndyScherzinger:
@tobiasKaminsky needs a rebase and see https://github.com/nextcloud/android/pull/4932#issuecomment-564899614 tests are reported to be failing but aren't 😁

PR #4935, opened 2019-12-04
Bump to 3.9.1

tobiasKaminsky:


tobiasKaminsky:
Lint erorr is because transifex added a new string on some languages, which is not yet used in stable-3.9.
We can safely ignore this.

PR #4982, opened 2019-12-11
only add account, but not go to main app

tobiasKaminsky:
Usage: start com.nextcloud.client/com.owncloud.android.authentication.AuthenticatorActivity with "onlyAdd" set to true as boolean

Wished by @grote

### DIFF
--- a/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -179,6 +179,7 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
     private static final String KEY_SERVER_AUTH_METHOD = "SERVER_AUTH_METHOD";
     private static final String KEY_WAITING_FOR_OP_ID = "WAITING_FOR_OP_ID";
     private static final String KEY_AUTH_TOKEN = "AUTH_TOKEN";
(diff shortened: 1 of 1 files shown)


grote:
Thanks a lot @tobiasKaminsky for the quick PR!

Unfortunately, it doesn't seem to work for me. I tried to debug it, but for some reason I can't even build master from a fresh checkout with `./gradlew assembleGeneric`. Tried debugging this, but gave up.

So I need to guess here: Maybe `onNewIntent()` isn't called and the flag needs to be checked in `onCreate()` instead?

tobiasKaminsky:
> Unfortunately, it doesn't seem to work for me. I tried to debug it, but for some reason I can't even build master from a fresh checkout with `./gradlew assembleGeneric`. Tried debugging this, but gave up.

My fault :-( I changed library but did not merged the corresponding client PR in time.



> So I need to guess here: Maybe `onNewIntent()` isn't called and the flag needs to be checked in `onCreate()` instead?

True. I just tested it with having NC app up and running.
Now it should also work if the app never has been started.

grote:
Awesome thank! Now both works, compiling and only adding the account. The workflow I need is now perfect, right after adding the account, the user gets back to where they started.

Ready to merge from my side!

AndyScherzinger:
@tobiasKaminsky feel free to merge,Looks good Code-wise

PR #4992, opened 2019-12-12
Ui comparison

tobiasKaminsky:
We can record tests with screenshots:
image

And later we can compare them:
image

And if it fails, it should get uploaded:
image

@AndyScherzinger @ezaquarii I appreciate your feedback


AndyScherzinger:
Like they idea 👍

PR #3948, opened 2019-05-04
Showing Unicode encoded IDN strings in drawer navigation

AndyScherzinger:
Resolves #1046

Issue #1046, opened 2017-05-27
IDN-domains not displayed correctly


kevinkk525:
### Actual behaviour
-IDN-domains (https://www.denic.de/en/know-how/idn-domains/) do work but in the account name they are displayed in their ACE string instead of IDN version.
Of course this is no bug, just a visual improvement.

### Expected behaviour
-Display IDN-domains correctly
 
### Steps to reproduce
1. Get an IDN-domain
2. Log in
3. check account name

### Environment data
Android version: 7.1.2

Device model:  Oppo Find 7

Nextcloud app version: Nextcloud-dev 20170525

Nextcloud server version: 12.0.0

AndyScherzinger:
Hi @kevinkk525, thanks for the input! And yes, while it is technically not a but I agree that this doesn't look very nice and see below, I also think this be fixed (even though it might be helpful if you could provide us with a test account so we can test the code change once it is done, waiting for the other folks approval first though).

@mario @tobiasKaminsky sounds good to me, a quick search revealed that this might be achievable using 
http://download.java.net/jdk7/archive/b123/docs/api/java/net/IDN.html#toUnicode(java.lang.String) - not sure if the setting of the flags might be necessary.

AndyScherzinger:
@kevinkk525 Or if you are willing to test we can first deploy it to the dev builds :) ...which might be easier!

kevinkk525:
You can deploy it to dev builds as I'm using those already anyways. I will gladly test it.
Additionally I can still create a test account for you if needed. 
Am 28.05.17, 00:30, Andy Scherzinger <notifications@github.com> schrieb:
@kevinkk525 Or if you are willing to test we can first deploy it to the dev builds :) ...which might be easier!

kevinkk525:
Just bumping this up as no progress was made during the last 2 years. (I know it's not a bug and has low priority).

AndyScherzinger:
Hi @kevinkk525, thanks for the reminder 👍
Would you be willing to create a test account for us and send it to either @tobiasKaminsky or me via email? 
I'd then work on a solution for it 😊

kevinkk525:
Thanks for working on this. I created a testaccount and sent the login to Tobias as I couldn't find a way to contact you personally.

AndyScherzinger:
Thanks! Today is a bank holiday in Germany, so @tobiasKaminsky will likely forward it to me by tomorrow :)

tobiasKaminsky:
[...]

PR #5006, opened 2019-12-15
Introduced user account id type

ezaquarii:
Since the application logic expects account id in a
specific format and relies on it, AccountId provides
dedicated type to represent id.

It uses RFC2396-compliant parser to validate id
format. It does not cover all possible edge cases,
but offers very cheap way to improve reliability over
existing free-form String used so far.

The type also offers reliable API to access host
and username parts, allowing us to remove ad-hoc
parsing logic present in various UI places.

AccountId provides CharSequence interface, preserving
String-like semantics whenever it is needed: mostly SQL
queries but also few UI cases.

### DIFF
--- a/src/androidTest/java/com/nextcloud/client/account/UserTest.kt
+++ b/src/androidTest/java/com/nextcloud/client/account/UserTest.kt
@@ -1,0 +1,77 @@
+package com.nextcloud.client.account
+
+import android.accounts.Account
+import android.net.Uri
+import android.os.Parcel
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.owncloud.android.lib.common.OwnCloudAccount
+import com.owncloud.android.lib.common.OwnCloudBasicCredentials
+import com.owncloud.android.lib.resources.status.OwnCloudVersion
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotSame
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.net.URI
+
+@RunWith(AndroidJUnit4::class)
+class UserTest {
+
+    @Test
+    fun anonymousUserImplementsParcelable() {
+        // GIVEN
+        //      anonymous user instance
+        val original = AnonymousUser("test_account")
+
+        // WHEN
+        //      instance is serialized into Parcel
+        //      instance is retrieved from Parcel
+        val parcel = Parcel.obtain()
+        parcel.setDataPosition(0)
+        parcel.writeParcelable(original, 0)
+        parcel.setDataPosition(0)
+        val retrieved = parcel.readParcelable<User>(User::class.java.classLoader)
+
+        // THEN
+        //      retrieved instance in distinct
+        //      instances are equal
+        assertNotSame(original, retrieved)
+        assertTrue(retrieved is AnonymousUser)
+        assertEquals(original, retrieved)
+    }
+
+    @Test
+    fun registeredUserImplementsParcelable() {
+        // GIVEN
+        //      registered user instance
+        val uri = Uri.parse("https://nextcloud.localhost.localdomain")
+        val credentials = OwnCloudBasicCredentials("user", "pass")
+        val account = Account("test@nextcloud.localhost.localdomain", "test-type")
+        val ownCloudAccount = OwnCloudAccount(uri, credentials)
+        val server = Server(
+            uri = URI(uri.toString()),
+            version = OwnCloudVersion.nextcloud_17
+        )
+        val original = RegisteredUser(
+            account = account,
+            ownCloudAccount = ownCloudAccount,
+            server = server
+        )
+
+        // WHEN
+        //      instance is serialized into Parcel
+        //      instance is retrieved from Parcel
+        val parcel = Parcel.obtain()
+        parcel.setDataPosition(0)
+        parcel.writeParcelable(original, 0)
+        parcel.setDataPosition(0)
+        val retrieved = parcel.readParcelable<User>(User::class.java.classLoader)
+
+        // THEN
+        //      retrieved instance in distinct
+        //      instances are equal
+        assertNotSame(original, retrieved)
+        assertTrue(retrieved is RegisteredUser)
+        assertEquals(original, retrieved)
+    }
+}
--- a/src/main/java/com/nextcloud/client/account/AccountId.kt
+++ b/src/main/java/com/nextcloud/client/account/AccountId.kt
@@ -1,0 +1,77 @@
+package com.nextcloud.client.account
+
+import java.io.Serializable
+import java.lang.IllegalArgumentException
+import java.net.URI
+import java.net.URISyntaxException
+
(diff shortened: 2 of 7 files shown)


ezaquarii:

The purpose of this type is to:

1. migrate existing `account.name` calls
2. eliminate existing ad-hoc `account.name.split("@")[0]...` logic
3. provide (more) reliable parsing of user ID

Ad 3, I haven't found any official specification for the user account tyle. It looks for me like it is just `username@nextcloud.instance.url`, hence RFC2396 seems to be pretty good fit to handle it, but please review this decision with somebody from server team.

I already rely on it in the upcoming PRs where I clean up existing logic, so I decided to extract it and ship it as a separate change.

src/main/java/com/nextcloud/client/account/AccountId.kt
```diff
@@ -0,0 +1,77 @@
...
+        @JvmStatic
+        private val serialVersionUid = 0L
+
+        /**
+         * Create account id from username and backend server URI.
+         *
+         * @throws IllegalArgumentException if input username and host are empty
+         */
+        fun create(username: CharSequence, backendUri: URI): AccountId {
+            if (username.length == 0 || backendUri.host.length == 0) {
```
tobiasKaminsky: I think I remember that we have sometimes, for various reasons, a null username. Then this check will return in a NPE?

src/test/java/com/nextcloud/client/account/AccountIdTest.kt
```diff
@@ -0,0 +1,153 @@
...
+
+        data class Row(val username: String, val host: String, val id: String = "$username@$host")
+
+        private val validIds = listOf(
+            Row("username", "hostname.com"),
+            Row("user_name", "hostname.com"),
+            Row("user_name_100", "hostname.com"),
+            Row("username", "hostname"),
+            Row("username", "host.domain.tld"),
+            Row("name.surname", "host.domain.tld")
```
tobiasKaminsky: also valid hostnames are 
- host.domain.tld/subfolder
- domain.tld/subfolder
- domain.tld:9999
- domain.tld:9999/subfolder

Can you add these, so that we make sure the tests work?

AndyScherzinger: And it needs to be able to handle encoded umlauts (which look weird...)

src/androidTest/java/com/nextcloud/client/account/UserTest.kt
```diff
@@ -0,0 +1,77 @@
+package com.nextcloud.client.account
```
AndyScherzinger: @tobiasKaminsky should test classes also have the license header?

AndyScherzinger:
Looks good to me

src/test/java/com/nextcloud/client/account/AccountIdTest.kt
ezaquarii: Hm... Indeed, I haven't accounted for those and I think it is not handled yet.

I'll extend the logic to cover those cases.

src/androidTest/java/com/nextcloud/client/account/UserTest.kt
tobiasKaminsky: Yes, otherwise we cannot distinguish if there is a file that has a different license (e.g. copied from old git history, or c&p from somewhere else)

ezaquarii: Yes. I'll add it.

src/main/java/com/nextcloud/client/account/AccountId.kt
ezaquarii: I was thinking about making it non-null across the stack. I've seen multiple places where such null username would simply crash. Also, in some places we parse the ID assuming there is a username, so I don't think this is a genuine case - sounds more like a bug with a workaround possibly?

src/test/java/com/nextcloud/client/account/AccountIdTest.kt
ezaquarii: @AndyScherzinger @tobiasKaminsky Can you please expand on this encoded characters thing?

Here is what I have in the test so far:

```
        private val validAddresses = listOf(
            "host.domain.tld/sp%C3%A9ci%C3%A4l%20s%C3%BCbf%C3%B3l%C4%99", // escaped "spéciäl sübfólęr"
            "host.domain.tld:9999/sp%C3%A9ci%C3%A4l%20s%C3%BCbf%C3%B3l%C4%99", // escaped "spéciäl sübfólęr"
            "host.domain.tld/spéciäl%20sübfólęr",
            "host.domain.tld:9999/spéciäl%20sübfólęr",
            "host.domain.tld/special%20subfolder",
            "host.domain.tld:9999/special%20subfolder",
            "host.domain.tld/subfolder",
            "domain.tld/subfolder",
            "domain.tld:9999",
            "domain.tld:9999/subfolder",
            "host.domain.tld",
            "domain.tld",
            "hostname"
        )

        private val validUsernames = listOf(
            "üsérnąmë",
            "username",
            "user.name",
            "user_name",
            "1234567890"
        )
```

How shall we map them to account IDs then?

ezaquarii: It looks like we do it here:
https://github.com/nextcloud/android-library/blob/1b57251eb61b5810f2b8da6ce68bb88d4f5d2210/src/main/java/com/owncloud/android/lib/common/accounts/AccountUtils.java#L184-L196

tobiasKaminsky: Indeed :+1:

AndyScherzinger: Umlauts on folders is one thing, Umlauts on domains another (known as IDN domains, see https://www.denic.de/en/know-how/idn-domains/) corresonding issue is #1046 and PR has been #3948 

überall-ist.de
```
xn--berall-ist-8db.de
```

ezaquarii: IDN domains should be handled "for free". Paths are also handled in both escaped and not escaped forms as well.

My main concern is that android `Uri` parser is not entirely RFC-compliant and accepts rubbish URIs while `java.net.URI` is known to be stricter.

Since we're using server base URL as ID suffix, this sould in theory guarantee it is valid (otherwise you wouldn't be able to contact the server).

But we wouldn't be doing our job without asking about this .01% chance of having some weird hipster case with pseudo-URI containing emojis or something like this. Possible even?

AndyScherzinger: Not sure if even possible

ezaquarii:
I underestimated the complexity of account name concept - it requires more research. I don't feel confident I have all information to ship it in backward compatible fashion, so I'll put it on hold until we clarify if this is even possible to implement.

ezaquarii:
I'll come back to this when it's time. For the time being, this is not what we want to do.

PR #5012, opened 2019-12-16
fix dark theme preference for dev version: use different preference name

tobiasKaminsky:
Fix https://github.com/nextcloud/android/issues/4997

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
@@ -695,7 +695,7 @@ public class SettingsActivity extends ThemedPreferenceActivity
 
         loadStoragePath();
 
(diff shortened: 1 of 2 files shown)


AndyScherzinger:
@tobiasKaminsky I might be wrong here, but this changes the preference name for any build, not just the dev build, right?So I am not sure how this fixes the issue.

tobiasKaminsky:
Previous setting was boolean, but now we have a string/enum.
As this will cause a typecast exception (which we cannot really handle, as it is internally in Android), I chose to just have a new setting/name.

You are right, that this will also remove the existing value of latest official version.
But I guess this is not that bad, as then users will go to settings and can use the third option, if available.

AndyScherzinger:
@tobiasKaminsky broken, uses 2 different keys now! see call
https://github.com/nextcloud/android/blob/master/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java#L720

PR #5014, opened 2019-12-16
Opt-out of scoped storage

ArisuOngaku:
Fix denied access (throwing FileNotFoundExceptions) to "out of scoped storage" files.

https://developer.android.com/training/data-storage/compatibility

The Nextcloud android app needs to access any external storage in the case of a custom folder sync or external media file folders.

While I named the branch "-workaround", it looks like the app will never use scoped storage unless it implements some on-demand by path access permission (correct me if I'm wrong).


tobiasKaminsky:
> it looks like the app will never use scoped storage unless it implements some on-demand by path access permission (correct me if I'm wrong).

I did not had the time to look into scoped storages…

PR #5018, opened 2019-12-16
Only trigger media scans for media files

### LINKED ISSUE #4999: Only trigger media scanner for media files
There's several places in the app where [`triggerMediaScan()` is called](https://github.com/nextcloud/android/search?q=triggerMediaScan&type=Code). This happens irrespective of the mimeType which is available in some of these places.

For huge non-media files, this can slow down the system significantly and even crash with OOM errors. Furthermore when writing these files with a DocumentsProvider, maybe the user doesn't want these files get indexed by the media scanner?

Anyway, a simple fix would be to check the mimetype for media files and only trigger a scan for those.

grote:
This also removes the last instance of generateResizedImage() which was
crashing NextCloud with OOM errors.

Fixes #4999

### DIFF
--- a/src/main/java/com/owncloud/android/operations/RemoveFileOperation.java
+++ b/src/main/java/com/owncloud/android/operations/RemoveFileOperation.java
@@ -33,6 +33,7 @@ import com.owncloud.android.lib.common.operations.RemoteOperationResult;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult.ResultCode;
 import com.owncloud.android.lib.resources.files.RemoveFileRemoteOperation;
 import com.owncloud.android.operations.common.SyncOperation;
+import com.owncloud.android.utils.MimeTypeUtil;
 
 
 /**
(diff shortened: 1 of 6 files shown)


src/main/java/com/owncloud/android/operations/RemoveFileOperation.java
```diff
@@ -91,9 +91,6 @@ protected RemoteOperationResult run(OwnCloudClient client) {
 
         fileToRemove = getStorageManager().getFileByPath(remotePath);
 
-        // store resized image
-        ThumbnailsCacheManager.generateResizedImage(fileToRemove);
-
```
grote: I am not sure why this was called here. Why do we need to save resized thumbnails when the files is about to get removed anyway?

It seems to be the only call for this method and caused OOM errors for me, because it was called on non-media files.

AndyScherzinger: > Why do we need to save resized thumbnails when the files is about to get removed anyway?

my guess and please correct me if I am wrong @tobiasKaminsky so that it can still be displayed in other places (like the uploads list) or maybe the trash bin. Not sure.
While if we can remove it then we probably have a thumbnail cahces anyways already...

tobiasKaminsky: This is not about thumbnail, but about resized image.
Idea of resized images is to show a screen adapted image (~70-100kb) instead of 5-10Mb for an image.
Now imagine that you download an image (maybe for printing), use it, and remove it afterwards.
If you then open this image in preview, you would have to wait for fetching the resized images (though you had the full sized image already).

Therefore I created the resized image.

--> do not remove it, but check for mimetype and do this only on images.

grote: Thanks for clarifying @tobiasKaminsky! I re-added the thumbnail resizing and added an image check to it.

AndyScherzinger:
Thanks for the fix @grote ❤️

PR #5022, opened 2019-12-17
iconify README.md

AndyScherzinger:
...like the server repo, to be more visually consistent.

https://github.com/nextcloud/android/blob/enh/iconify_readme/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# [Nextcloud](https://nextcloud.com) Android app 
+# [Nextcloud](https://nextcloud.com) Android app :iphone:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-# [Nextcloud](https://nextcloud.com) Android app 
+# [Nextcloud](https://nextcloud.com) Android app :iphone:
```
tobiasKaminsky: iphone? Really? :see_no_evil:

AndyScherzinger: Yes, that is the only phone emoji code (while it looks like a windows phone...)

PR #5031, opened 2019-12-18
Fix NPE crash on startup when storage manager is not set

### LINKED ISSUE #4941: Crash on first installation due to missing storageManager
```
java.lang.NullPointerException: Attempt to invoke virtual method 'com.owncloud.android.datamodel.OCFile com.owncloud.android.datamodel.FileDataStorageManager.getFileByPath(java.lang.String)' on a null object reference
        at com.owncloud.android.ui.activity.FileDisplayActivity.onStart(FileDisplayActivity.java:2573)
        at android.app.Instrumentation.callActivityOnStart(Instrumentation.java:1425)
        at android.app.Activity.performStart(Activity.java:7825)
        at android.app.ActivityThread.handleStartActivity(ActivityThread.java:3294)
        at 
```

1. uninstall app
2. install app
3. set up user account
-> boom

@ezaquarii I suspect that this is from the user account changes and that storage manager is not set after adding a new account.

(Normally we do not set milestone, but this must be fixed before next major release).

ezaquarii:
Fixes #4941

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
@@ -70,6 +70,13 @@ public abstract class BaseActivity extends AppCompatActivity implements Injectab
     }
 
     @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        Account account = accountManager.getCurrentAccount();
+        setAccount(account, false);
+    }
+
+    @Override
     protected void onPostCreate(@Nullable Bundle savedInstanceState) {
         super.onPostCreate(savedInstanceState);
         preferences.addListener(onPreferencesChanged);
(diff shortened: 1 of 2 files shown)


ezaquarii:
@tobiasKaminsky When creating new account, I see that we jump through 5 activities.

`FileDisplayActivity 1` -> `AuthenticatorActivity` -> `FirstRunActivity` -> `FileDisplayActivity 2` -> `FileDisplayActivity 3`

`FileDisplayActivity 2` is for some reason resumed and paused few times before `FileDisplayActivity 3` comes to life.

The whole flow is pretty complicated and I believe we could at least simplify it by removing the authenticator callback in `FileDisplayActivity 1`, as it dies immediately after launching `AuthenticatorActivity`.

src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
```diff
@@ -177,6 +177,7 @@ protected void swapToDefaultAccount() {
         if (newAccount == null) {
             /// no account available: force account creation
             createAccount(true);
+            finish();
```
ezaquarii: I tried to run it without it. I managed to create the account properly, but it landed in empty file list activity as a result, requiring extra pull-to-refresh.

tobiasKaminsky: We have somewhere a delay of 500ms to do a very first loading of folder content, maybe it just takes a bit too long for you.

tobiasKaminsky:
> The whole flow is pretty complicated

it is indeed.
As old login flow is deprecated since NC12, I will remove this for 3.11.
This then should simplify it a lot and I will take your idea into account.

Issue #5030, opened 2019-12-18
Markdown Preview: Make hyperlinks clickable


thomas-mc-work:
At the moment the hyperlinks are just formatted as links (underlined and blue font colour) but not clickable. Would be nice to have them functionally.

tobiasKaminsky:
@AndyScherzinger do you have an idea?

AndyScherzinger:
I'll have a look, while it might come out of the box when the client supports Nextcloud Text.

AndyScherzinger:
https://noties.io/Markwon/docs/v4/core/configuration.html#linkresolver / https://github.com/noties/Markwon/issues/118 should do the trick while this doesn't seem to get triggered, so maybe somebody has an idea why, maybe @BaptisteRFoxIntel can shed some light onto this? 🙏 

https://github.com/nextcloud/android/blob/dbe12e6d20cd86a007001709a77dd938bf35c9d4/src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java#L286-L300

noties:
Hey! It is always `MovementMethod` when links are not clickable. But `Markwon` will set one automatically if it's missing after text is rendered. And it works just fine in most of the cases until `textIsSelectable` is added. So, try to set `LinkMovementMethod` to your TextView before applying markdown content, like this:

```java
// in onCreateView for example, or right before applying markdown
textView.setMovementMethod(LinkMovementMethod.getInstance());

/* ... */

markwon.setMarkdown(textView, markdown);
```

At least this worked for me

AndyScherzinger:
@noties Thanks a lot for the hint! 🥇 Works for me too just fine.

@thomas-mc-work @tobiasKaminsky will open a PR is just a minute.

PR #4890, opened 2019-11-26
Direct editing support

tobiasKaminsky:
- abstract EditorWebView
- support direct editing endpoint
- needs Text app
- behaviour: 
 - if downloaded show it locally, use edit for remote editing
 - if not downloaded, open it remote directly

Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

PR #5035, opened 2019-12-19
Make hyperlinks clickable in markdown previews

AndyScherzinger:
Resolves: #5030


src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -291,6 +296,19 @@ public void configureTheme(@NonNull MarkwonTheme.Builder builder) {
...
                 }
+
+                @Override
+                public void configureConfiguration(@NonNull MarkwonConfiguration.Builder builder) {
+                    builder.linkResolver((view, link) -> {
+                        try {
+                            Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(link));
+                            DisplayUtils.startIntentIfAppAvailable(intent, getActivity(), R.string.no_browser_available);
+                        } catch (Throwable throwable) {
+                            Toast.makeText(context, R.string.error_opening_link, Toast.LENGTH_SHORT).show();
```
tobiasKaminsky: I prefer snackbar instead of toast?

src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -291,6 +299,19 @@ public void configureTheme(@NonNull MarkwonTheme.Builder builder) {
...
                     builder.linkColor(ThemeUtils.primaryColor(context, true));
                 }
+
+                @Override
+                public void configureConfiguration(@NonNull MarkwonConfiguration.Builder builder) {
+                    builder.linkResolver((view, link) -> {
+                        try {
+                            Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(link));
+                            DisplayUtils.startIntentIfAppAvailable(intent, getActivity(), R.string.no_browser_available);
+                        } catch (Throwable throwable) {
```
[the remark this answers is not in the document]
tobiasKaminsky: From where do you get this throwable?
And why do we catch this?

src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
AndyScherzinger: fine with me, yes. I'll change that

AndyScherzinger:
@tobiasKaminsky can you please rebase _and_ fix this thing? Your PR #4890 breaks the whole concept here. We only display markdown rendering if it has the right mimetype, which you just killed via your rich-thing PR 😠

tobiasKaminsky:
> Your PR #4890 breaks the whole concept here.

Idea is to have the same as in office documents or audio/video:
if the file is downloaded, show it locally.
If it is not downloaded, show it online.

I will take care of this.

Other than that your PR is working fine :+1:

src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -291,6 +296,19 @@ public void configureTheme(@NonNull MarkwonTheme.Builder builder) {
...
```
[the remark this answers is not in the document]
tobiasKaminsky: Why yo we need a try/catch here?

AndyScherzinger:
> if the file is downloaded, show it locally.
If it is not downloaded, show it online.

That's not the point. Point is your removed/hide the OCFile (I know I can call getFile()) so markdown rendering is always triggered! Markdown should not be triggered for non markdown mimetypes!

> Other than that your PR is working fine

Don't know, since we now have TextPreview and TextFilePreview while I haven't checked if both would still do proper markdown rendering and would also just do this in case of markdown present.

tobiasKaminsky:
> TextPreview and TextFilePreview

TextPreview is an abstract class, which is used by PreviewTextFile and PreviewTextString

I'll take care of it next year.
Just as an hint for me: What happens if I show plain text with Markdown? (we do the same on web with text app).

src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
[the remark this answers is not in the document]
ezaquarii: And why do we catch *World(tm)*?

src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
ezaquarii: `Toast` is deprecated in Material Design in favour of `Snackbar`.

src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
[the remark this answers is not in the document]
AndyScherzinger: We don't, this can be removed. I just checked the API code and the only exception they might throw is a NPE in case `link` is `null`

src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
AndyScherzinger: Jupp! 👍

src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -175,22 +183,36 @@ private void performSearch(final String query, int delay) {
...
+                    builder.linkColor(ThemeUtils.primaryColor(activity, true));
+                }
+
+                @Override
+                public void configureConfiguration(@NonNull MarkwonConfiguration.Builder builder) {
+                    builder.linkResolver((view, link) -> {
+                        try {
+                            Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(link));
+                            DisplayUtils.startIntentIfAppAvailable(intent, activity, R.string.no_browser_available);
+                        } catch (Throwable throwable) {
```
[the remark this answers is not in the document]
AndyScherzinger: Could be removed (I know I added it myself) since we don't expect exceptions here.

src/main/java/com/nextcloud/client/di/ComponentsModule.java
```diff
@@ -139,6 +140,9 @@
     @ContributesAndroidInjector abstract ContactListFragment chooseContactListFragment();
     @ContributesAndroidInjector abstract PreviewMediaFragment previewMediaFragment();
     @ContributesAndroidInjector abstract PreviewTextFragment previewTextFragment();
+
+    @ContributesAndroidInjector
+    abstract PreviewTextFileFragment previewTextFileFragment();
```
AndyScherzinger: Should be one line

tobiasKaminsky: This is due to code formatting of Android Studio.
Not sure, if we want to do this then all the time manually, or stick 100% with code formatting and having line wraps…

AndyScherzinger: I'd then say stick with AS code formatting, doing it manually leads to manual work.

src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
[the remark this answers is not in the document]
tobiasKaminsky: Removed.

PR #5054, opened 2019-12-22
Fix account switch buttons in drawer header

### LINKED ISSUE #5053: Account switch button in drawer header is not updated properly
### Steps to reproduce
1. open app
2. add 2 or 3 accounts
3. switch between accounts

### Expected behaviour
1 bit button show current account
2 miniature buttons in drawer show other accounts

### Actual behaviour
Miniature buttons are stuck and show one account

Nextcloud app version: master

ezaquarii:
This is a regression caused by using User.equals() to compare against
platform Account object.

- Migrated this functionality to User
- Refactored drawer setup code using CollectionsKt.filter()
  List<User>

Fixes #5053


src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -1360,33 +1369,22 @@ protected void onAccountCreationSuccessful(AccountManagerFuture<Bundle> future)
     }
 
     /**
-     * populates the avatar drawer array with the first three ownCloud {@link Account}s while the first element is
-     * always the current account.
+     * Get list of users suitable for displaying in navigation drawer heder.
```
tobiasKaminsky:
```suggestion
     * Get list of users suitable for displaying in navigation drawer header.
```

PR #5061, opened 2019-12-24
Resolved the crash when account is not new (Auth)

### LINKED ISSUE #5058: Error when you try to connect on an already logged server/user
Hi,

When you're connected on a server and want to connect on the SAME server with the SAME user, AND ONLY if you use a `webview_login_url` by default, the app will crash because `mAuthStatusView` is not defined in AuthenticatorActivity.

A "solution" would be to create a try-catch in `showAuthStatus` method, but it's ugly because the AuthenticatorActivity musts show the login input to show the error message.

If a webview_login_url is defined, the wanted behaviour is ALWAYS to connect through this specified URL and we don't want to show the URL input Activity for login. 

So what I suggest, is to create a Toast containing the error, and if the login fail, we finish Authenticator Activity (only if a `webview_login_url` is defined).

@AndyScherzinger @tobiasKaminsky I've figured out that everything related to branding was really not necessary, are you "interested" in this kind of issue or not? Thanks for your answer and have a happy holiday :D

Shagequi:
Hi,

it fixes #5058 :)

thanks.
Signed-off-by: Kilian Périsset <kilian.perisset@infomaniak.com>

### DIFF
--- a/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -253,6 +253,7 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
     private long mWaitingForOpId = Long.MAX_VALUE;
 
     private boolean webViewLoginMethod;
+    private boolean showWebViewLoginUrl;
     private String webViewUser;
     private String webViewPassword;
     private boolean forceOldLoginMethod;
@@ -320,6 +321,7 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
             showLegacyLogin = false;
         } else {
             webViewLoginMethod = !TextUtils.isEmpty(getResources().getString(R.string.webview_login_url));
+            showWebViewLoginUrl = getResources().getBoolean(R.bool.show_server_url_input);
             showLegacyLogin = true;
         }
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -748,4 +749,9 @@ static public void startIntentIfAppAvailable(Intent intent, Activity activity, @
             DisplayUtils.showSnackMessage(activity, error);
         }
     }
+
+    static public void showErrorAndAbort(Context context, String errorMessage) {
```
ezaquarii: 1. The term `abort` has a specific meaning associated with a fatal process error, causing process termination. Could you please consider `showErrorAndFinishActivity`?
2. Type cast should be removed - it requires `Activity` so it should accept `Activity` as parameter
3. This should generally belong to activity class - display utils shall not collect fragments of activity business logic.

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1840,17 +1865,22 @@ private void showServerStatus() {
...
-            mAuthStatusView.setVisibility(View.VISIBLE);
+        try {
+            if (mAuthStatusIcon == NO_ICON && EMPTY_STRING.equals(mAuthStatusText)) {
+                mAuthStatusView.setVisibility(View.INVISIBLE);
+            } else {
+                mAuthStatusView.setText(mAuthStatusText);
+                mAuthStatusView.setCompoundDrawablesWithIntrinsicBounds(mAuthStatusIcon, 0, 0, 0);
+                mAuthStatusView.setVisibility(View.VISIBLE);
+            }
+        } catch (Exception e) {
```
ezaquarii: Why do we blindly catch all type of exceptions just to ignore them? This is not a good practice:

1. it hides bugs that are present inside `try-catch`
2. unhandled errors cause undefined cascade issues in other components

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1478,43 +1480,66 @@ private void updateServerStatusIconAndText(RemoteOperationResult result) {
...
+                    } else {
+                        mAuthStatusText = errorMessage;
+                    }
+                    break;
+                case UNHANDLED_HTTP_CODE:
+                default:
+                    mAuthStatusText = ErrorMessageAdapter.getErrorCauseMessage(result, null, getResources());
+                    mAsyncTask.cancel(true);
+            }
+        } catch(Exception e) {
```
ezaquarii: Let's catch onlyh those exceptions that can occur here.

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1478,43 +1480,66 @@ private void updateServerStatusIconAndText(RemoteOperationResult result) {
...
+            Log_OC.e(TAG, "An error occured during auth status refresh : \n" + e.getMessage());
+            String errorMessage = getString(R.string.auth_unknown_error_title);
+            mAsyncTask.cancel(true);
```
ezaquarii: This method updates UI, but I see that it also cancels some async tasks, which does not belong here. I'd handle this elsewhere.

Also, this method is called in 3 different places - are you sure this behaviour is proper for all 3 use cases?

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1636,17 +1655,23 @@ public void onAuthenticatorTaskCallback(RemoteOperationResult result) {
...
-                    EditText serverAddressField = findViewById(R.id.hostUrlInput);
-                    serverAddressField.setText(mServerInfo.mBaseUrl);
+                        initOverallUi();
 
-                    findViewById(R.id.server_status_text).setVisibility(View.GONE);
-                    mAuthStatusView = findViewById(R.id.auth_status_text);
+                        EditText serverAddressField = findViewById(R.id.hostUrlInput);
+                        serverAddressField.setText(mServerInfo.mBaseUrl);
+
+                        findViewById(R.id.server_status_text).setVisibility(View.GONE);
```
ezaquarii: Some UI controls are accessed via members, some by `findViewById` - can we decide one approach?

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1636,17 +1655,23 @@ public void onAuthenticatorTaskCallback(RemoteOperationResult result) {
...
+                    } else {
+                        setContentView(R.layout.account_setup);
 
-                    EditText serverAddressField = findViewById(R.id.hostUrlInput);
-                    serverAddressField.setText(mServerInfo.mBaseUrl);
+                        initOverallUi();
 
-                    findViewById(R.id.server_status_text).setVisibility(View.GONE);
-                    mAuthStatusView = findViewById(R.id.auth_status_text);
+                        EditText serverAddressField = findViewById(R.id.hostUrlInput);
```
ezaquarii: While here, we can fix the ID to conform to our existing naming convention, ie use `_` instead of `camelCase`.

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1567,33 +1592,27 @@ public void onAuthenticatorTaskCallback(RemoteOperationResult result) {
...
+                    initOverallUi();
 
-                findViewById(R.id.server_status_text).setVisibility(View.GONE);
-                mAuthStatusView = findViewById(R.id.auth_status_text);
+                    EditText serverAddressField = findViewById(R.id.hostUrlInput);
+                    serverAddressField.setText(mServerInfo.mBaseUrl);
 
-                showAuthStatus();
+                    findViewById(R.id.server_status_text).setVisibility(View.GONE);
+                    mAuthStatusView = findViewById(R.id.auth_status_text);
```
ezaquarii: This looks like it belongs to `initOverallUi()` where we extract all relevant UI elements into member variables.

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -253,6 +253,7 @@
     private long mWaitingForOpId = Long.MAX_VALUE;
 
     private boolean webViewLoginMethod;
+    private boolean showWebViewLoginUrl;
```
ezaquarii: Are those 2 flags exclusive? If so, we should convert it to `enum`.

Shagequi:
@ezaquarii 
Hi,
Thank you for your detailed and built up feedback, it's very pleasant to read.

I agree with all the points you've made... however, some of the code snippets aren't my own, and I went with a "bug fixing" logic, not a "class refactor" logic.

The points that concern my code will be corrected :) Concerning the "code that was there before", I propose rather to make a new issue on this subject and to refactor the class in second time

src/main/java/com/owncloud/android/utils/DisplayUtils.java
Shagequi: About point 3,
I prefer keeping it in DisplayUtils.
If you see other functions in this class, they all collect the Activity/context '^^

ezaquarii: > If you see other functions in this class, they all collect the Activity/context '^^

I'd call it an example of [broken windows theory](https://en.wikipedia.org/wiki/Broken_windows_theory) and not worthy following.

In recent time we killed quite some of those `SomethingSomethingUtil` dumping-ground classes, but the road is still long and bumpy. What I'd like to encourage you to do is to at least refrain from contributing to this problem.

Shagequi: I think differently. 
Not all Utils class are "dump", we are supposed to use them in order to respect the DRY principle. 

To put a method to show an error and close an Activity in ONE activity, and rewrite this method each time we need is worse than having a "junkyard" class where reusable methods are stored.
In addition, this function could be reused in the future.

Shagequi:
Branch has been updated.

I fixed a good part of the problems you brought up, and took the opportunity to reformat the code a bit :)

src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
Shagequi: They're not exclusive.
Maybe we would like to show the login URL but "force user" to use a specific one to connect,

Shagequi:
@tobiasKaminsky @AndyScherzinger @ezaquarii 
Good evening, may we merge this now ? 
Thank you :)

Shagequi:
@tobiasKaminsky 
Good morning, may we merge this now ?
Thank you :)

tobiasKaminsky:
@Shagequi there are some conflicting files, can you rebase it? :heart:

Shagequi:
Hi @tobiasKaminsky,

This PR is 2 months old, the code was approved by Ezaquarii and Andy 1-2 months ago. 
So...now that the master branch and the FileDisplayActivity file have changed, I don't have the ability to pick exactly what I need to solve the issue, and I don't do double work.

I don't have time today to fix this bug, whereas 2 months ago I had it. I'm disappointed that it took so long to be merged when everything was fine and unconflicted.

I'm sorry, but I can't work on this ATM... if I have time to work on the issue, I will.

ezaquarii:
I'll try to pick it up.

tobiasKaminsky:
@Shagequi totally understandable, and I was not aware that this PR is so old.
Sorry for that, I try now to catch up more quickly :heart:

ezaquarii:
@Shagequi @tobiasKaminsky @AndyScherzinger 

I rebased this:
https://github.com/nextcloud/android/pull/5509/

I had to open new PR as I don't have right to push to the original repo, but I retained commit author.

Shagequi:
Due to a new PR (https://github.com/nextcloud/android/pull/5509), I close this one

PR #5064, opened 2019-12-25
Extract account logic from BaseActivity into a mixin

ezaquarii:
This PR moves some account-related shared logic in `BaseActivity` into a [mix-in](https://en.wikipedia.org/wiki/Mixin).

My intention is to collect various account handling logic there, although it is still in a research phase.

@tobiasKaminsky @AndyScherzinger I tried to bring a test coverage to some acceptable level, but this code is so untestable that the task proven to be counter-productive. This code generally must go anyway.

The session mixin has the original, account-related code moved *verbatim*, without functional changes, so the mixin should behave indentically to the original and 
https://github.com/nextcloud/android/pull/5031#discussion_r359570614 is also accounted for.

### DIFF
--- a/src/main/java/com/nextcloud/client/di/ActivityInjector.java
+++ b/src/main/java/com/nextcloud/client/di/ActivityInjector.java
@@ -23,7 +23,6 @@ package com.nextcloud.client.di;
 import android.app.Activity;
 import android.app.Application;
 import android.os.Bundle;
-
 import androidx.fragment.app.FragmentActivity;
 import androidx.fragment.app.FragmentManager;
 import dagger.android.AndroidInjection;
@@ -72,4 +71,3 @@ public class ActivityInjector implements Application.ActivityLifecycleCallbacks 
         // not needed
     }
 }
-
--- a/src/main/java/com/nextcloud/client/mixins/SessionMixin.kt
+++ b/src/main/java/com/nextcloud/client/mixins/SessionMixin.kt
@@ -1,0 +1,133 @@
+/*
+ * Nextcloud Android client application
+ *
+ * @author Chris Narkiewicz
+ * Copyright (C) 2020 Chris Narkiewicz <hello@ezaquarii.com>
+ * Copyright (C) 2020 Nextcloud GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.nextcloud.client.mixins
+
+import android.accounts.Account
+import android.app.Activity
+import android.content.ContentResolver
+import android.content.Intent
+import android.os.Bundle
+import com.nextcloud.client.account.User
+import com.nextcloud.client.account.UserAccountManager
+import com.nextcloud.java.util.Optional
+import com.owncloud.android.datamodel.FileDataStorageManager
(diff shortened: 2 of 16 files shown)


AndyScherzinger:
@ezaquarii looks good to me code-wise. 👍 
Only thing missing are the license headers in the newly created files.

src/main/java/com/nextcloud/client/di/ActivityInjector.java
```diff
@@ -32,6 +33,7 @@
 
     @Override
     public final void onActivityCreated(Activity activity, Bundle savedInstanceState) {
+        Log.d("mytag", String.format("ActivityInjector.onActivityCreated(): %s", activity));
```
tobiasKaminsky: Why do we use Log instead of Log_OC here?
"mytag" is also too generic. You could generate a tag from the activity?
Or call it "ActivityInjector"

src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -1359,12 +1358,13 @@ private View findQuotaViewById(int id) {
      */
     protected abstract void restart();
 
-    @Override
-    protected void onAccountCreationSuccessful(AccountManagerFuture<Bundle> future) {
-        super.onAccountCreationSuccessful(future);
-        updateAccountList();
-        restart();
-    }
+//    @Override
```
tobiasKaminsky: Can you remove this?

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -227,15 +228,20 @@
...
+    @Override
+    public void finish() {
+        super.finish();
+    }
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         Log_OC.v(TAG, "onCreate() start");
-
+        Log.d("mytag", "FileDisplayActivity.onCreate(): before super");
```
tobiasKaminsky: Same questions regarding log :-)

ezaquarii:
Oupsie! This is my private debug code - it should not be there. I'll fix it.

src/main/java/com/nextcloud/client/di/ActivityInjector.java
ezaquarii: Yeah, this is some leftover from development. This code should not be there and I'll remove it.

src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
ezaquarii: Removed.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
ezaquarii: Anther leftover... removed.

src/main/java/com/nextcloud/client/mixins/SessionMixin.kt
```diff
@@ -0,0 +1,113 @@
+package com.nextcloud.client.mixins
```
tobiasKaminsky: License header

src/test/java/com/nextcloud/client/mixins/MixinRegistryTest.kt
```diff
@@ -0,0 +1,48 @@
+package com.nextcloud.client.mixins
```
tobiasKaminsky: License header

src/test/java/com/nextcloud/client/mixins/SessionMixinTest.kt
```diff
@@ -0,0 +1,47 @@
+package com.nextcloud.client.mixins
```
tobiasKaminsky: License header

tobiasKaminsky:
Did a smoke test and it works :+1:

src/main/java/com/nextcloud/client/mixins/SessionMixin.kt
ezaquarii: done

src/test/java/com/nextcloud/client/mixins/SessionMixinTest.kt
ezaquarii: done

Issue #4754, opened 2019-10-29
Wrong setting header color after enabling dark


tobiasKaminsky:
![2019-10-29-114643](https://user-images.githubusercontent.com/5836855/67760551-cebfb800-fa41-11e9-9ed3-f4966d1c5d80.png)

- install for first time
- go to setting
- enable dark theme

tobiasKaminsky:
Did not happened to me with #4757 
If happens again, we can re-open.

tobiasKaminsky:
This happens for me, when I set server theming to white and enable dark  mode.

AndyScherzinger:
Hmmm, but isn't the header the same for dark/light? At least I don't knwo how to handle this or when to decide for which color (black/white hheader text/icons)

tobiasKaminsky:
I think it is just missing that black text should be used instead of white.
I'll have a look.

PR #5107, opened 2020-01-07
Fix white header with dark mode

tobiasKaminsky:
Fix https://github.com/nextcloud/android/issues/4754

 

White background of header will not be changed, as this is desired via theming api.
But at least we can now see the text again ;-)

(also applies for all other activities)

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -254,7 +254,7 @@ public abstract class DrawerActivity extends ToolbarActivity
 
         setupDrawerToggle();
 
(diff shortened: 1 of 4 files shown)


AndyScherzinger:
@tobiasKaminsky I tried the latest QA build but it renders the "+" on the fab button in white (for dark) so I can see the icon (which seems to work according to your screenshots).

tobiasKaminsky:
> @tobiasKaminsky I tried the latest QA build but it renders the "+" on the fab button in white (for dark) so I can see the icon (which seems to work according to your screenshots).

can you open up a new issue for this? Also which primary color do you use?

AndyScherzinger:
Done: https://github.com/nextcloud/android/issues/5110

PR #5121, opened 2020-01-08
new background icon

tobiasKaminsky:
TODO
- [x] @jancborchardt I need a background image for drawer: 984x656 as PNG


tobiasKaminsky:
also using this as new background, if server is >= 18 and theming app is disabled.

jancborchardt:
Background image for the drawer looks good! :+1: :)

AndyScherzinger:
I'll give it a spin in a second, mom

AndyScherzinger:
@tobiasKaminsky looks good to me, only thing I find strange is that with this change the launcher xml-drawables move from v26 down to v24 which I am unsure will work (but might have been auto-gerenated by AS, so it is fine)?

tobiasKaminsky:
> @tobiasKaminsky looks good to me, only thing I find strange is that with this change the launcher xml-drawables move from v26 down to v24 which I am unsure will work (but might have been auto-gerenated by AS, so it is fine)?

Yes, this was done by AS, so I assume they know what they do :rofl:

PR #5122, opened 2020-01-08
show first run again

### LINKED ISSUE #5112: FirstRun is not shown
### Steps to reproduce
1. fresh installation

### Expected behaviour
- see first run wizard

### Actual behaviour
- directly see authenticator activity

Tested on latest master / 3.10

tobiasKaminsky:
Fix #5112 

I also removed finish as it was not in there before.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
@@ -8,14 +8,13 @@ import android.accounts.OperationCanceledException;
 import android.content.Intent;
 import android.os.Bundle;
 import android.os.Handler;
-import android.os.PersistableBundle;
 
 import com.nextcloud.client.account.User;
 import com.nextcloud.client.account.UserAccountManager;
 import com.nextcloud.client.di.Injectable;
 import com.nextcloud.client.preferences.AppPreferences;
-import com.nextcloud.java.util.Optional;
 import com.nextcloud.client.preferences.DarkMode;
+import com.nextcloud.java.util.Optional;
 import com.owncloud.android.MainApp;
 import com.owncloud.android.datamodel.FileDataStorageManager;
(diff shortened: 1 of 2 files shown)


src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
```diff
@@ -69,13 +68,6 @@ public UserAccountManager getUserAccountManager() {
         return accountManager;
     }
 
-    @Override
-    protected void onCreate(@Nullable Bundle savedInstanceState) {
```
ezaquarii: This part was moved in this commit:

https://github.com/nextcloud/android/commit/d7d1c475aebb7b7e92b98c3d36dbf66c7a244d86

If you loot at it, you'll see that this code was previously in a wrong `onCreate(@Nullable Bundle savedInstanceState, @Nullable PersistableBundle persistentState)` method.

It was a bug that I introduced while refactoring `Account`-related code in `BaseActivity`, caused by IDE autocompletion - I simply typed wrong method signature while overriding `onCreate`. We didn't notice this as it works by accident simply becuase numerious calls to `setAccount` in derived activities.

src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
```diff
@@ -178,7 +170,6 @@ protected void swapToDefaultAccount() {
         if (newAccount == null) {
             /// no account available: force account creation
             createAccount(true);
-            finish();
```
ezaquarii: This is what closes the `FirstRunActivity`. 

Removing it wil break `FileListActivity`, causing empty file list after creating new account:
https://github.com/nextcloud/android/pull/5031#discussion_r359570614

mario: Yup, that's indeed what happens :-/  @tobiasKaminsky ?

PR #5143, opened 2020-01-10
launcher: correct folder position according to AS

tobiasKaminsky:
We had two times ic_launcher_foreground.xml


tobiasKaminsky:
> Shouldn't this then also be v24 instead of the general folder?

That is what AS is doing… I have no idea why…

Issue #5086, opened 2020-01-02
Very slow processing when first download lot of file in folder


JorisBodin:
Hello,

I've noticed that for processing more than 1000 files in a folder. The application processes the folder for about **a minute**. And only the first time. Afterwards it's not so long, but it's long anyway.

I'm making the optimization. **I'm getting to about a second now**. But I'd rather improve my code before proposing the pull request.

https://github.com/Infomaniak/android-infomaniak-drive/tree/optimization

JorisBodin:
![image](https://user-images.githubusercontent.com/2796439/71663869-2555b500-2d57-11ea-94e4-eb37dc00449f.png)
@tobiasKaminsky 
I do not understand the addition of this line. It adds a lot of processing time. What is its purpose?

JorisBodin:
@tobiasKaminsky 
I found it useful
https://github.com/nextcloud/android/issues/4502#issuecomment-540430544

I was trying to optimize this, but I can't reproduce the bug to understand exactly what's going on.
Because it adds 10 seconds of processing for the first sync of a folder of + 1000 files.

I realized that on the emulator it was very fast. But I was on an s8 galaxy

tobiasKaminsky:
> I was trying to optimize this, but I can't reproduce the bug to understand exactly what's going on.

This is not a bug, but you can reproduce it like this:
- have a fresh installation (with no database content of files)
- open up "photo" section 
- click on an image
- this needs then parentId and thus it needs to be fetched

As this is exists on second refresh, it is a lot faster.

For photos this might be possible to change, but the concept is also true for search, where you can get a file, where no ancestor is known, and thus it (currently) needs to be fetched.

PR #5150, opened 2020-01-10
Optimization for very slow processing after downloading folder content

JorisBodin:
Hello,

Correction for this issue: https://github.com/nextcloud/android/issues/5086

_My tests were performed with a folder of 1044 files and Galaxy S8_

Benchmark
**On master:**
Insert (first download):
**50** seconds.

Update (refresh):
**20** seconds

**With this pull request:**
Insert:
**20** seconds

Update:
**6** seconds

There are still 2 things to improve
First, the doubleCheck in FileContentProvider:
image

This check is here because _saveFolder_ (in _FileDataStorageManager_) is sometimes called twice for an insert.
I've found two cases where that happens, one of which is simple to fix.
The first time the application is started, the _WRITE_EXTERNAL_STORAGE_ permission is requested. When the user accepts, _syncAndUpdateFolder_ is called in _onResume_ and in _onRequestPermissionsResult_.
Delete _syncAndUpdateFolder_ in _onRequestPermissionsResult_ is solution.

Also the first time the application is started, if refreshing the folder, _syncAndUpdateFolder_ is called twice

The solution would be to block the simultaneous call of _saveFolder_.

Benchmark
**If doubleCheck is removed:**
Insert (first download):
**13** seconds.

Update (refresh):
**6** seconds

The other thing to improve is this check https://github.com/nextcloud/android/pull/4647
(It is discussed in https://github.com/nextcloud/android/issues/5086)
I haven't found an optimization solution for this case. 😞 

Benchmark
**But if we remove it**
Insert (first download):
**1** seconds.

Update (refresh):
**5** seconds

I also took the time to refactor the FileDataStorageManager class, I have removed the inconsistencies and duplications that have accumulated since the beginning. 2298 lines to 1887 lines
It allowed me to better understand the code and better optimize it.

Next week, we're thinking of coming to hackweek in Berlin. Will you be there?


src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -75,8 +75,8 @@
 public class FileDataStorageManager {
     private static final String TAG = FileDataStorageManager.class.getSimpleName();
 
-    private static final String AND = "=? AND ";
-    private static final String FAILED_TO_INSERT_MSG = "Fail to insert insert file to database ";
+    private static final String AND = " = ? AND ";
+    private static final String FAILED_TO_UPDATE_MSG = "Fail to update file to database ";
```
tobiasKaminsky: It seems that it same String is used for insert and update.
Changing this to "update" is as wrong as using only "insert".

So either, we use both and change it accordingly, or we generify the error message and us "failed to insert/update file to…"

JorisBodin: With my refactor, I've removed a lot of code duplication. So FAILED_TO_INSERT_MSG was used more than once.
However, for update, it is used in 2 places. So I changed that message to this.
We can use "insert/update", but it will be less exact.
"Fail to insert" is used in the insertFile method, if you want to merge this :D

PR #5175, opened 2020-01-13
UI enhancements to rich workspaces

tobiasKaminsky:


src/main/java/com/owncloud/android/ui/preview/PreviewTextFragment.java
```diff
@@ -195,10 +193,8 @@ protected static Spanned getRenderedMarkdownText(Activity activity, String markd
             .usePlugin(new AbstractMarkwonPlugin() {
                 @Override
                 public void configureTheme(@NonNull MarkwonTheme.Builder builder) {
-                    TextPaint textPaint = new TextPaint();
```
tobiasKaminsky: @AndyScherzinger seems that this is not used at all, so I removed it.

PR #5209, opened 2020-01-15
OnlyOffice: make download possible

tobiasKaminsky:


juliushaertl:
@tobiasKaminsky OnlyOffice 5.5 (latest docker) should now have the Content-Disposition set but it does not seem to pickup the file name. Could you check that?

Issue #4993, opened 2019-12-12
Weird state management in FirstRunActivity


Kilian-Perisset:
Hi,

On the first start (without being connected), The FirstRunActivity is launched after a few Intents. 
And, when you access this Activity, and lock your phone, and reopen it, you should go back to this "FirstRunActivity", but instead you're redirected directly to login and a thread crashes.

So why don't we redirect immediatly to FirstRunActivity through MainApp instead of passing from FileDisplayActivity, and why we don't relaunch this activity onRestart ? And why is it a MediaFolder detection before ANY connection (which fails and kill the thread)? : 

``` 
2019-12-12 16:47:53.267 16477-25164/XXXXXXXXX E/JobExecutor: Crashed job{id=27, finished=true, result=FAILURE, canceled=false, periodic=true, class=MediaFoldersDetectionJob, tag=MediaFoldersDetectionJob}
    java.lang.SecurityException: Permission Denial: reading com.android.providers.media.MediaProvider uri content://media/external/images/media from pid=16477, uid=10468 requires android.permission.READ_EXTERNAL_STORAGE, or grantUriPermission()
        at android.os.Parcel.createException(Parcel [...]
``` 

So, what is the "wanted" behaviour for the first launch ? And why do we pass through FileDisplayActivity without being connected ?

V. 3.9.2

AndyScherzinger:
Maybe you can shed some light on this @tobiasKaminsky?

tobiasKaminsky:
>  The FirstRunActivity is launched after a few Intents.
> And, when you access this Activity, and lock your phone, and reopen it, you should go back to this "FirstRunActivity", but instead you're redirected directly to login and a thread crashes.

Indeed. I guess I never tried this…Thanks for this findind!



> So why don't we redirect immediatly to FirstRunActivity through MainApp instead of passing from FileDisplayActivity,

Good question, I do not remember why I(?) did this. But your idea sounds better.

>  and why we don't relaunch this activity onRestart ?

I guess I just forgot about this case…




> And why is it a MediaFolder detection before ANY connection (which fails and kill the thread)? :

This is failing because of missing permission. And it is run as we simply start this as a periodic background check.
I would go make this in MediaFoldersDetectionJob more robust and exit early if permission is missing.

Do you want to to tackle these? :heart: 
(if so, please in separate (as possible) PRs)

Shagequi:
I'll see what I can do :)

tobiasKaminsky:
[...]

PR #5218, opened 2020-01-16
Fix #5213 app is excluded from recent/current apps

### LINKED ISSUE #5213: Since 3.10.0, app is excluded from recent/current apps
Hi,
Since the 3.10.0 version, during the first launch/connection process only,  the app is excluded from the "recent apps".
Here is an example of "recent apps" view : 
HFCji

Here is the steps to reproduce :
1. Disconnect your account(s) in order to retreive the "connection status" process OR uninstall and reinstall the app
2. During the process, try to go on another app, like an Authenticator or similar app
3. Try to return in Nextcloud app through the "recent apps" view (by clicking on bottom-bar button) 
**You'll see that you're unable to return to the app through this view.**

JorisBodin:
After talking with @tobiasKaminsky, we've decided to remove this finish()
Because "onResume" is called after requestPermissionsResult for PERMISSIONS_WRITE_EXTERNAL_STORAGE

https://github.com/nextcloud/android/issues/5213

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
@@ -182,10 +182,6 @@ public abstract class BaseActivity extends AppCompatActivity implements Injectab
         if (newAccount == null) {
             /// no account available: force account creation
             createAccount(true);
-
-            if (enableAccountHandling) {
-                finish();
-            }
         } else {
             currentAccount = newAccount;
         }


ezaquarii:
I tested account creation and login and I see no problem described in https://github.com/nextcloud/android/pull/5031#discussion_r359570614

Please test the following scenarios:

1. login
2. adding new account

In both cases application should not display empty file list.

mario:
This is actually how it is for every permission request - the thing is that when you request permissions via requestPermissions the activity is paused. After answering, onResume gets called (again).

mario:
So, to summarize: I'd suggest setting up all the views and everything and then ask for permissions.

JorisBodin:
I tested login and adding new account. I have note empty files.

**With finish()**
withfinish

**Wihout finish()**
wihoutFinish

JorisBodin:
I encountered the problem without the finish.  I had the request for authorization to write toi storage file on the login activity.  So the activity that lists the files is launched while we are on that of login.  Here is the problem.  I'm watching this next week.

ezaquarii:
That doesn't seem right given that:

1) this code path is hit only when there is no user
2) without user you can't instantiate storage manager to access files

tobiasKaminsky:
Hm.
This permission is not needed, if you 
- have new installation
- login for very first time
- browse files
- download file

It is only needed, if you want to upload a file from external storage, eg. external sdcard.
So we maybe should delay this permission question to the point we open up "+" and select "upload files"?

JorisBodin:
> That doesn't seem right given that:
> 
> 1. this code path is hit only when there is no user
> 2. without user you can't instantiate storage manager to access files

Screenshot_1579688132

Just to let you know it can happen when the app is launched.

JorisBodin:
@ezaquarii 
The problem is explained in more detail here: https://github.com/nextcloud/android/issues/4993

I think that adding this finish is not necessary (especially because it generates a bug). It is better to fix the problem at source, as described in this issue https://github.com/nextcloud/android/issues/4993

tobiasKaminsky:
@ezaquarii this bug prevents user from switching during login to another app, e.g. for generating one-time password.

As the only drawback right now might be that we get an empty list on first start, I am fine with this.

@JorisBodin can you then remove the accountHandling boolean es we this then do not need it.

How to proceed, we should discuss in #4993

JorisBodin:
@tobiasKaminsky if remove _enableAccountHandling_ in _baseActivity_, _accountManager.getCurrentAccount()_ is called.
So we go directly to the login view.

untitled

But it will have to be removed during the big correction of https://github.com/nextcloud/android/issues/4993

tobiasKaminsky:
ah true.
Then let us keep it for now and do it right with #4993.

JorisBodin:
@tobiasKaminsky So, do we merge it or do we close this MR?

PR #5231, opened 2020-01-18
images: slightly rounded corners

tobiasKaminsky:
@jancborchardt what do you think?


jancborchardt:
Nice in list view, but for grid view it either needs to be veeeery small of a rounding or none at all. :)

AndyScherzinger:
I agree fine @jancborchardt, fine for the list view and the grid view shouldn't have an rounding

tobiasKaminsky:
Hm. I really like it on grid view…But I will change it, as you said, masters of design :-)

tobiasKaminsky:
@AndyScherzinger @jancborchardt

JorisBodin:
I also prefer no rounding in the grid view. It looks good in your latest screenshot.

tobiasKaminsky:
@AndyScherzinger you were right, 20dp was too much.
I changed it to 10dp.
What do you think?

AndyScherzinger:
Multiples of 8 would be typical, so how about 8dp?

tobiasKaminsky:
Looks also fine :-)

AndyScherzinger:
@tobiasKaminsky looks good, just SpotBugs has a raised score :(

tobiasKaminsky:
> HTTP_PARAMETER_POLLUTION: HTTP Parameter Pollution

no idea why this suddenly pops up.
I increase spotbugs.

PR #5246, opened 2020-01-20
Fix #2753 "Unresponsive App" error popup when syncing

### LINKED ISSUE #2753: "Unresponsive App" error popup when syncing
### Actual behaviour
- When manually syncing a large and partially-synced directory, Android O will complain about "App not responding [Close/Wait/Report]" after about 5-20 seconds. This does not happen when the process encounters new/modified files early enough, i.e. before the "unresponsive"-popup occurs. I think, every encountered new/changed file resets some kind of timeout of the popup.
- Even when selecting "wait", the Nextcloud app seems to be closed after some moments when not encountering new/changed files (even if there are plenty to be found later down the tree), thus stopping all sync activity. So the wild guess is that Android misinterprets the idle-ness of "nothing to download" as "app is stuck" and gives up waiting anyways after asking the user once?

### Expected behaviour
- No error popup about unresponsive app when syncing large directories
 
### Steps to reproduce
1. Have a large directory tree with mostly un-synced files
2. Tap it -> Menu -> Synchronize
3. wait for error popup

### Environment data
Android version: 8.0.0

Device model: Moto Z Play

Nextcloud app version: 3.2.0

Nextcloud server version: 13.0.4

siebica:
Always start SyncFolderOperations as background services to prevent
ANRs when they take a little more time to complete.

### DIFF
--- a/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
+++ b/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
@@ -494,11 +494,7 @@ public class SynchronizeFolderOperation extends SyncOperation {
         intent.setAction(OperationsService.ACTION_SYNC_FOLDER);
         intent.putExtra(OperationsService.EXTRA_ACCOUNT, mAccount);
         intent.putExtra(OperationsService.EXTRA_REMOTE_PATH, path);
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Sorry for coming back so late.
I tested it on Android 10 and Android 5, both are working.

tobiasKaminsky:
Thank you for your contribution!

PR #5290, opened 2020-01-23
Layout 2.0

tobiasKaminsky:
:exclamation: NO REBASE ONTO MASTER :exclamation: 
(to get newest master/resolve conflict, use merge master)

Fix https://github.com/nextcloud/android/issues/5207

As changing layout is a big step, it will include lots of PRs. 
If we would merge the first PR, then it will happen that dev version is not functional.
If we would do it in one giant PR, it would be un-reviewable.

This way we can merge sub-PRs into this, and once everything is working, this will get rebased onto master and merged.
Only reviewed PRs will go into this PR, so that we can directly merge this, once we are finished.

**TODO**
- [ ] multi select
- [ ] search
- [ ] change between back/hamburger menu
- [ ] ShowErrorActivity: handle new "actionBar"
- [ ] find replacement for progressBar

Please keep this PR clean and create new issues/PR like normal.
If you notice a bug/improvement, related to layout changing, add it as TODO.

Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

PR #5291, opened 2020-01-23
New header

tobiasKaminsky:
This is the first initial PR for layout 2.0.
It is "intended"/known that not everything will work, but for keeping the PR small, I chose to let some things be broken.

Please add broken things to #5290.

This changes
- new header, scrollable
- moves FAB to files.xml, otherwise it is not visible on start, but only when scrolling list


AndyScherzinger:
Looks good, while for the hamburger icon color I'd rather go with `#616161` (as in `grey700`)

AndyScherzinger:
Alignment isn't perfect but fine for this PR (papercut to be fixed later)
72968052-fb763800-3dc2-11ea-87f8-583e9b10d348

Other than that we could (at some point) replace the sorting info with just the icon (from the sorting options list)

tobiasKaminsky:
So then let us merge this as a fist start, then open up individual PRs and issues.

PR #5293, opened 2020-01-23
Status bar: white/black 

tobiasKaminsky:
I found only a way to set status bar to white with >= Android M.
On <M it will be primary color.

### DIFF
--- a/src/main/java/com/nextcloud/client/device/DeviceInfo.kt
+++ b/src/main/java/com/nextcloud/client/device/DeviceInfo.kt
@@ -23,6 +23,7 @@ package com.nextcloud.client.device
 
 import android.content.Context
 import android.content.pm.PackageManager
+import android.content.res.Configuration
 import android.os.Build
 import java.util.Locale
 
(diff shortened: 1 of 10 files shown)


AndyScherzinger:
This likely fails to work for the dark theme, no?

tobiasKaminsky:
:see_no_evil: I totally forgot about dark theme…
Thanks for pointing out :-)

tobiasKaminsky:
These are screenshots based on newHeader (in case you wonder…)

src/main/java/com/nextcloud/client/device/DeviceInfo.kt
```diff
@@ -38,4 +39,10 @@ class DeviceInfo {
     fun editorSupported(): Boolean {
         return apiLevel < Build.VERSION_CODES.LOLLIPOP
     }
+
+    fun isDarkModeEnabled(context: Context): Boolean {
+        val nightModeFlag: Int = context.getResources().getConfiguration().uiMode and Configuration.UI_MODE_NIGHT_MASK
+
```
AndyScherzinger: Does this include system based night mode?

AndyScherzinger:
Settings page has black set not white (using system settings), uploads view is blue with black icons, same for any other Activity class.

src/main/java/com/nextcloud/client/device/DeviceInfo.kt
tobiasKaminsky: Sort of.
Instead of checking settings, it checks actually state of dark mode.
So this is always up to date.

PR #5296, opened 2020-01-23
Fixed defective search

Shagequi:
- Fixed onbackpressed bug
- Fixed the ever-present cross
- Optimized/cleaned the code
~- Added translations (EN, FR, IT, ES, DE) for next step : search by direct edit~

### DIFF
--- a/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.java
+++ b/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.java
@@ -49,6 +49,7 @@ import android.view.inputmethod.InputMethodManager;
 import android.widget.AdapterView;
 import android.widget.AdapterView.OnItemClickListener;
 import android.widget.GridView;
+import android.widget.ImageButton;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.ProgressBar;
@@ -61,6 +62,7 @@ import com.nextcloud.client.preferences.AppPreferences;
 import com.nextcloud.client.preferences.AppPreferencesImpl;
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
+import com.owncloud.android.datamodel.OCFile;
 import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.lib.resources.files.SearchRemoteOperation;
 import com.owncloud.android.ui.EmptyRecyclerView;
@@ -229,6 +231,7 @@ public class ExtendedListFragment extends Fragment implements
             }
         }));
 
+        // On close -> empty field, show keyboard and
         closeButton.setOnClickListener(view -> {
             searchView.setQuery("", true);
             searchView.requestFocus();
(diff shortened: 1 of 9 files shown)


src/main/res/values-fr/strings.xml
```diff
@@ -307,7 +307,7 @@ Attention la suppression est irréversible.</string>
     <string name="file_migration_cleaning">Nettoyage…</string>
     <string name="file_migration_dialog_title">Mise à jour du chemin de stockage</string>
     <string name="file_migration_directory_already_exists">Le dossier de données existe déjà. Choisissez une des options suivantes :</string>
-    <string name="file_migration_failed_dir_already_exists">Le dossier Nextcloud existe déjà</string>
+    <string name="file_migration_failed_dir_already_exists">Le dossier Infomaniak Drive existe déjà</string>
```
AndyScherzinger: will be overridden by the bot, rather set a placeholder and replace it via code (that way the translations will come in and it'll work for your fork too!

src/main/res/values-es/strings.xml
```diff
@@ -554,7 +552,7 @@
     <string name="screenshot_05_autoUpload_heading">Subida automática</string>
     <string name="screenshot_05_autoUpload_subline">para sus fotos &amp; vídeos</string>
     <string name="screenshot_06_davdroid_heading">Calendario y contactos</string>
-    <string name="screenshot_06_davdroid_subline">Sincronizar con DAVx5</string>
+    <string name="search_instruction">Buscar archivo (mínimo 2 caracteres)</string>
```
AndyScherzinger: will be overridden by the transifex bot

Shagequi:
@AndyScherzinger 
Ok, I've fixed the translations (it was clearly a mistake, it shouldn't be present in this branch, sorry ^^') 

And i formated the code in the last commit (you may see code changes in previous commits)

src/main/res/values-es/strings.xml
Shagequi: It's clearly a mistake, sorry

src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.java
```diff
@@ -135,7 +136,8 @@
 
     private EmptyRecyclerView mRecyclerView;
 
-    protected SearchView searchView;
+    public SearchView searchView;
+    public ImageView closeButton;
```
tobiasKaminsky: both are not needed to be public, according to AndroidStudio.
searchView can remain protected and closeButton even private

tobiasKaminsky:
Tested & works :+1:

Shagequi:
@tobiasKaminsky 

Thank you for your review.
There's no problem if you want to change my code, you know ^^

If these are impacting changes that require me to do a "heavy" review, I will gladly go over them :)
On the other hand, if it's a concern about the accessibility of a variable, or formatting, or obsolete import, don't hesitate to change things for me (we'll both save time).

PR #5300, opened 2020-01-24
Re-enable shared search and add ui test for list shares

### LINKED ISSUE #4845: The "Shared" item in Drawer (Shared Search Event) doesn't work ?
Hi,

The Shared Search Event (accessible from Drawer) doesn't seem to work on a complete NextCloud package

**App : NextCloud 3.9.0**
**Server : NextCloud 17.0.1 - 17.0.1.1**

The error returned on server is : 
```
Error | PHP | Undefined offset: 0 at [...]/drive/3rdparty/icewind/searchdav/src/XML/Operator.php#71 |  
```
(I didn't change server configuration, and used default files provided with NextCloud)

Could you try to reproduce ?

tobiasKaminsky:
Fix #4845

This helps as a quick win to get shared search back again.

Later we should separate this from remote search and create its own fragment.


tobiasKaminsky:
> SpotBugs increased!

No idea, where and why.
This is just too hard to compare… 
We really should aim to get this down a lot :-)

Issue #5178, opened 2020-01-13
Folder synchronization 150GB - The device allow the sync of too much data


Shagequi:
Hi,

When you want to synchronize a folder (of 150GB+) manually (by click on contextual menu of the folder, then "Synchronize"), the device starts the sync/download even if the phone has only 2gb of free space. 

I think we should add a condition, in order to check if the device is able to assume a "heavy" synchronization. Or to propose synchronization variants to the user (i.e. synchronize only a certain folder in the wanted folder).

What do you think about that ?

Shagequi:
After a talk with @jancborchardt and @tobiasKaminsky, the solution would be to create a dialog which allow three options : 
1) **Choose what to sync** -> Close dialog, and navigate down to the folder (main action)
2) **Free up space**  -> Close dialog, wipe space on phone through native process
3) **Cancel**  -> close the dialog, do not sync

The title of the dialog should be : "**Not enough space**", and the text :
"[file name] is [file size]mb but there is only [free space size]mb/gb available."

PR #5328, opened 2020-01-27
Added "not enough space dialog"

Shagequi:
Hi,

It fixes : https://github.com/nextcloud/android/issues/5178 (pre-approved by @jancborchardt IRL).

When we try to synchronize a "too big" folder/file (with a superior length than the phone available storage), the app will allow to synchronize it without any problem.

With this PR, it's now no longer allowed.

(I don't know how translations are handled through transifex, but the English and French are already translated)


AndyScherzinger:
> (I don't know how translations are handled through transifex, but the English and French are already translated)

only English is maintained "manually" within the repo. Any other language comes via the bot pushed from transifex, so french will be overridden as soon as this PR lands on master (overnight).

Can you post a screenshot? ❤️

Shagequi:
@AndyScherzinger,
Good morning 😄  :

Screenshot_20200128-092247_Nextcloud

AndyScherzinger:
Looks good :+1: - just some minor details

* Why are the strings marked non-translatable? (should be translated I think)
* Button texts shouldn't be capitalized (anymore, I am aware that we still haven't changed all dialogs/buttons)
* Dialog should have some whitespace left/right and not fill the screen horizontally

Shagequi:
@AndyScherzinger 

Thanks for your feedback 👍 

1) The `translatable` attribute has been removed
2) Based on Material chart : https://material.io/components/dialogs/#alert-dialog, the buttons are always in uppercase, so... Doesn't sound abnormal to me.
3) This dialog is based on `RemoveFile` one :
 
Screenshot_20200128-130955_Nextcloud

Which is created in `ConfirmDialogFragment`.
If you want to add white spaces, we need to add it in another feature/fix on another branch

AndyScherzinger:
@Shagequi thankls for the changes and yes, the other stuff can be changed alter via separate PRs 👍 

> Doesn't sound abnormal to me.

Both ways are fine material-wise. As for the capitalized button texts, that is rather depending on the theme you choose, like https://material.io/components/dialogs/#theming -> "Reply Theme" which at least for non-capitalized button texts is what we agreed on with @jancborchardt a while ago (haven't had the chance to document this in some kind of guideline I think 😞 Sorry for that)

As for the whitespace, thanks for the explanation, fine with me, would then rather be a papercut issue I can create later to harmonize the dialog stylings throughout the complete app 👍 

Other than that, fancy fixing the codacy warnings? If not and fine with you I can also take care of it. (Whatever works better for you ❤️)

Shagequi:
@AndyScherzinger 

Done :)

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -1043,5 +1039,22 @@ public static String getCapturedImageName() {
         return new SimpleDateFormat("yyyy-MM-dd_HHmmss", Locale.US).format(new Date()) + ".jpg";
     }
 
+    public static Pair<Boolean, Long> isSpaceEnough(OCFile file) {
```
ezaquarii: Please consider `hasEnoughSpace`.

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -1043,5 +1039,22 @@ public static String getCapturedImageName() {
...
+            android.os.Build.VERSION_CODES.JELLY_BEAN_MR2) {
+            availableBytesOnDevice = stat.getBlockSizeLong() * stat.getAvailableBlocksLong();
+        }
+        else {
+            availableBytesOnDevice = (long) stat.getBlockSize() * (long) stat.getAvailableBlocks();
+        }
+
+        boolean isSpaceEnough = file.getFileLength() < availableBytesOnDevice;
+
+        return new Pair<>(isSpaceEnough, availableBytesOnDevice);
```
ezaquarii: I'd consider a data class with named fields:
`data class AvailableSpace(val hasEnoughSpace: Boolean, val availableSpace: Long)`
or somehting similar (this is Kotlin, btw).

`Pair` is pretty useful concept in general, but the `first` and `second` field names are less than helpful when reading. I guess this is why Java did not ship `Pair` at all, argumenting that one should just create a well-named type instead.

If you believe that we should use `Pair` anyway, please consider using the class available in Kotlin stdlib instead of Android platform API, which gives us headaches during testing.

Shagequi:
@ezaquarii 
Done.
May we merge this now ?

Shagequi:
@ezaquarii 
Done.
May we merge this now ?

ezaquarii:
I'll review this evening.

Shagequi:
@tobiasKaminsky 

Good afternoon,

Would you please check this ? :D

Thanks

src/main/java/com/owncloud/android/ui/dialog/SyncFileNotEnoughSpaceDialogFragment.java
```diff
@@ -0,0 +1,132 @@
...
+    public static SyncFileNotEnoughSpaceDialogFragment newInstance(OCFile file, long availableDeviceSpace) {
+        Bundle args = new Bundle();
+        SyncFileNotEnoughSpaceDialogFragment frag = new SyncFileNotEnoughSpaceDialogFragment();
+        String properFileSize = DisplayUtils.bytesToHumanReadable(file.getFileLength());
+        String properDiskAvailableSpace = DisplayUtils.bytesToHumanReadable(availableDeviceSpace);
+
+        // Defining title, message and resources
+        args.putInt(ARG_TITLE_ID, R.string.sync_not_enough_space_dialog_title);
+        args.putInt(ARG_MESSAGE_RESOURCE_ID, R.string.sync_not_enough_space_dialog_placeholder);
+        args.putStringArray(ARG_MESSAGE_ARGUMENTS, new String[]{file.getFileName(), properFileSize, properDiskAvailableSpace});
```
tobiasKaminsky: please line wrap

src/main/java/com/owncloud/android/ui/dialog/SyncFileNotEnoughSpaceDialogFragment.java
```diff
@@ -0,0 +1,132 @@
...
+        args.putParcelable(ARG_PASSED_FILE, file);
+
+        // Defining buttons
+        if (file.isFolder()) {
+            args.putInt(ARG_POSITIVE_BTN_RES, R.string.sync_not_enough_space_dialog_action_choose);
+        }
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.N_MR1) {
```
tobiasKaminsky: What happens on older Android versions?

src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1720,6 +1724,33 @@ private boolean isSearchEventSet(SearchEvent event) {
...
     }
 
+    private void syncAndCheckFiles(Collection<OCFile> files) {
+        for (OCFile file : files) {
+            // Get the remaining space on device (after file download)
+            long remainingSpaceOnDevice = FileOperationsHelper.getRemainingSpaceOnDevice(file);
+            // Get the total available space on device (before file download)
+            long availableSpaceOnDevice = file.getFileLength() + remainingSpaceOnDevice;
+            // Determine if space is enough to download the file
+            boolean isSpaceEnough = remainingSpaceOnDevice > 0;
```
tobiasKaminsky: This is correct, but for me it is not intuitive to read, what about this:
getRemainingSpaceOnDevice -> should return how much space is left in this folder/partition
(no need of passing OCFile)

then it is just a 
boolean isSpaceEnough = remainingSpaceOnDevice > file.getFileLenght();

src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -1043,5 +1038,19 @@ public static String getCapturedImageName() {
         return new SimpleDateFormat("yyyy-MM-dd_HHmmss", Locale.US).format(new Date()) + ".jpg";
     }
 
+    public static Long getRemainingSpaceOnDevice(OCFile file) {
```
tobiasKaminsky: As written above, return remaining space as method name suggests that, do not subtract file length

tobiasKaminsky:
I changed storage location for computing remaining space left, as this can be changed by user via settings.

src/main/java/com/owncloud/android/ui/dialog/SyncFileNotEnoughSpaceDialogFragment.java
Shagequi: The "Free up space" button will simply not appear, because this function is unavailable before Jelly Bean. 
And, I take this opportunity to ask you why you stay on a minimal level of compilation on Android 4.1 (API level 16), it's a system that is 8 years old today. 
Without getting ahead of myself, I would say that the application is too complex today to run successfully (all the features) on Jelly Bean, are you planning to change the  minSdkVersion over time?

AndyScherzinger: > are you planning to change the minSdkVersion over time?

Yes, we do. The rule (introduced with the latest feature release): https://github.com/nextcloud/android/issues/150 --> `check if oldest Android version is EOL by Google and user usage <= 0.5% and drop it if so `

tobiasKaminsky:
@Shagequi I know you are quite busy, but can you fix/change my remarks?
Otherwise I am happy to take this over :heart:

PR #5336, opened 2020-01-28
Fds tests

tobiasKaminsky:
This splits up https://github.com/nextcloud/android/pull/5150 and has only refactoring in it.
I will add some more test cases.


src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -412,7 +412,7 @@ private void synchronizeData(List<Object> folderAndFiles) {
 
         Log_OC.d(TAG, "Remote folder " + mLocalFolder.getRemotePath() + " changed - starting update of local data ");
 
-        ArrayList<OCFile> updatedFiles = new ArrayList<>(folderAndFiles.size() - 1);
```
JorisBodin: Why prefer List to ArrayList?

PR #5337, opened 2020-01-28
Optimization after split up

tobiasKaminsky:
This is second PR after splitup of #5150 
It includes only code optimizations, done by @JorisBodin 

Plan is to have all possible/needed tests in #5336 and then they need to work also on this PR before merging.

For checking that everything was split up correctly, you can compare this PR with #5150 

### DIFF
--- a/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
+++ b/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
@@ -100,6 +100,114 @@ public class FileDataStorageManager {
         this.account = account;
     }
 
+    private Cursor executeQuery(Uri uri, String[] projection, String selection,
+                                String[] selectionArgs, String sortOrder, String errorMessage) {
+        Cursor cursor;
+        ContentResolver contentResolver = getContentResolver();
+
+        if (contentResolver != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -100,6 +100,114 @@ public FileDataStorageManager(Account account, ContentProviderClient contentProv
         this.account = account;
     }
 
+    private Cursor executeQuery(Uri uri, String[] projection, String selection,
```
tobiasKaminsky: We have the code style convention that if a line with parameters needs to be wrapped, that we then put all parameters in a separate line, so
``` 
private Cursor executeQuery(Uri uri, 
String[] projection, 
String selection,
String[] selectionArgs, 
String sortOrder, 
String errorMessage) 
{
```

I will change it, but just for your info :-) (it is not that bad, if we sometimes forget this, but it makes diff a bit easier, if a paremeter gets changed)

JorisBodin: Hi,

I take note for next time :)

tobiasKaminsky:
No longer valid, I fear.

PR #5345, opened 2020-01-28
Allows deep linking to open files

### LINKED ISSUE #2562: Option to open direct links in the app
### Actual behaviour
- Opening direct links like https://cloud.example.com/f/12345 is not possible

### Expected behaviour
- It should be possible to open direct links like https://cloud.example.com/f/12345 in the app, e.g. if I receive one via email.

I'm not sure if this is somehow possible on android, since Nextcloud instances are hosted on custom domains, but maybe someone from @nextcloud/android has an idea. Maybe we could extend the web ui with some button like "Show in the client" that is shown on mobile and calls some custom url like: `nc://directlink/https://cloud.example.com/f/12345`

Charon77:
Link: nc://directlink/f/{fileID} from direct linking

Resolves #2562

Signed-off-by: Evans Jahja <evansjahja13@gmail.com>

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -117,6 +117,19 @@
             <intent-filter>
                 <action android:name="android.intent.action.SEARCH" />
             </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="http"
+                    android:host="*"
+                    android:pathPattern="/.*index.php/f/.*"
+                    />
+                <data android:scheme="https"
+                      android:host="*"
+                      android:pathPattern="/.*index.php/f/.*"
+                    />
+            </intent-filter>
             <meta-data android:name="android.app.searchable"
                 android:resource="@xml/users_and_groups_searchable"/>
         </activity>
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -135,6 +135,8 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.annotation.Nullable;
 import javax.inject.Inject;
@@ -176,6 +178,8 @@ public class FileDisplayActivity extends FileActivity
     public static final String TAG_PUBLIC_LINK = "PUBLIC_LINK";
     public static final String FTAG_CHOOSER_DIALOG = "CHOOSER_DIALOG";
     public static final String KEY_FILE_ID = "KEY_FILE_ID";
+    public static final String KEY_ACCOUNT = "KEY_ACCOUNT";
+
 
     private static final String KEY_WAITING_TO_PREVIEW = "WAITING_TO_PREVIEW";
     private static final String KEY_SYNC_IN_PROGRESS = "SYNC_IN_PROGRESS";
@@ -2620,27 +2624,51 @@ public class FileDisplayActivity extends FileActivity
     private void handleOpenFileViaIntent(Intent intent) {
         showLoadingDialog(getString(R.string.retrieving_file));
 
-        String accountName = intent.getStringExtra("KEY_ACCOUNT");
+        String userName = intent.getStringExtra(KEY_ACCOUNT);
+        String fileId = intent.getStringExtra(KEY_FILE_ID);
 
-        Account newAccount;
-        if (accountName == null) {
-            newAccount = getAccount();
-        } else {
-            newAccount = getUserAccountManager().getAccountByName(accountName);
+        if (userName == null && fileId == null && intent.getData() != null) {
+            // Handle intent coming from URI
 
-            if (newAccount == null) {
+            Pattern pattern  = Pattern.compile("(.*)/index\\.php/([f])/([0-9]+)$");
+            Matcher matcher = pattern.matcher(intent.getData().toString());
+            if (matcher.matches()) {
+                String uri = matcher.group(1);
+                if ("f".equals(matcher.group(2))) {
+                    fileId = matcher.group(3);
+                    findAccountAndOpenFile(uri, fileId);
+                    return;
+                }
+            } else {
                 dismissLoadingDialog();
-                DisplayUtils.showSnackMessage(this, getString(R.string.associated_account_not_found));
+                DisplayUtils.showSnackMessage(this, getString(R.string.invalid_url));
                 return;
             }
+        }
+        openFile(userName, fileId);
 
-            setAccount(newAccount, false);
-            updateAccountList();
+    }
+    private void openFile(String userName, String fileId) {
+        Optional<User> optionalNewUser;
+        User user;
+
+        if (userName == null) {
+            optionalNewUser = getUser();
+        } else {
+            optionalNewUser = getUserAccountManager().getUser(userName);
         }
 
-        String fileId = String.valueOf(intent.getStringExtra(KEY_FILE_ID));
+        if (optionalNewUser.isPresent()) {
+            user = optionalNewUser.get();
+            setUser(user);
+            updateAccountList();
+        } else {
+            dismissLoadingDialog();
+            DisplayUtils.showSnackMessage(this, getString(R.string.associated_account_not_found));
+            return;
+        }
 
-        if ("null".equals(fileId)) {
+        if (fileId == null) {
             dismissLoadingDialog();
             DisplayUtils.showSnackMessage(this, getString(R.string.error_retrieving_file));
             return;
@@ -2649,13 +2677,56 @@ public class FileDisplayActivity extends FileActivity
         FileDataStorageManager storageManager = getStorageManager();
 
         if (storageManager == null) {
-            storageManager = new FileDataStorageManager(newAccount, getContentResolver());
+            storageManager = new FileDataStorageManager(user.toPlatformAccount(), getContentResolver());
         }
 
-        FetchRemoteFileTask fetchRemoteFileTask = new FetchRemoteFileTask(newAccount,
+        FetchRemoteFileTask fetchRemoteFileTask = new FetchRemoteFileTask(user.toPlatformAccount(),
                                                                           fileId,
                                                                           storageManager,
                                                                           this);
         fetchRemoteFileTask.execute();
+
+    }
+
+    private void findAccountAndOpenFile(String uri, String fileId) {
+
+        ArrayList<User> validUsers = new ArrayList<>();
+
+        for (User user : getUserAccountManager().getAllUsers()) {
+            if (user.getServer().getUri().toString().equals(uri)) {
+                validUsers.add(user);
+            }
+        }
+
+        if (validUsers.size() == 0) {
+            dismissLoadingDialog();
+            DisplayUtils.showSnackMessage(this, getString(R.string.associated_account_not_found));
+            return;
+        }
+
+        if (validUsers.size() == 1) {
+            openFile(validUsers.get(0).getAccountName(), fileId);
+            return;
+        }
+
+        ArrayList<String> validUserNames = new ArrayList<>();
+
+        for (User user : validUsers) {
+            validUserNames.add(user.getAccountName());
+        }
+
+        AlertDialog.Builder builder = new AlertDialog.Builder(getActivity());
+        builder
+            .setTitle(R.string.common_choose_account)
+            .setItems(validUserNames.toArray(new CharSequence[validUserNames.size()]),
+                      new DialogInterface.OnClickListener() {
+                public void onClick(DialogInterface dialog, int which) {
+                    openFile(validUsers.get(which).getAccountName(), fileId);
+                    showLoadingDialog(getString(R.string.retrieving_file));
+                }
+            });
+        AlertDialog dialog = builder.create();
+        dismissLoadingDialog();
+        dialog.show();
     }
 }
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -925,4 +925,6 @@
     <string name="sync_not_enough_space_dialog_action_free_space">Free up space</string>
     <string name="sync_not_enough_space_dialog_placeholder">%1$s is %2$s, but there is only %3$s available on device.</string>
     <string name="sync_not_enough_space_dialog_title">Not enough space</string>
+
+    <string name="invalid_url">Invalid URL</string>
 </resources>


AndyScherzinger:
Hi @Charon77 - welcome to the Nextcloud Android App project and thank you for your contribution! ❤️ I took the liberty to loop in some folks for review and feedback

looping in @tobiasKaminsky @juliushaertl @rullzer and @ezaquarii for review and feedback

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -512,6 +512,9 @@ protected void onNewIntent(Intent intent) {
             setFile(intent.getParcelableExtra(EXTRA_FILE));
             showDetails(file);
         } else if (Intent.ACTION_VIEW.equals(intent.getAction())) {
+            if (!intent.hasExtra(KEY_FILE_ID)) {
+                intent.putExtra(KEY_FILE_ID, intent.getData().getLastPathSegment());
```
ezaquarii: I'm not exactly sure how this extra is set, but rewriting it before handling does not feel right. Also, I do not understand why we use last path segment as file id - is it the same thing?

AndyScherzinger:
@ezaquarii I agree with your review, also looped in the other folks to make sure the url pattern is (still) fine since the corresponding issue is relatively old and the issues hasn't been labeled approved. So hoping to get some feedback from the rest of the mentioned people ❤️

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
mario: It's set from the Talk app at the moment when clicking on a shared file. The author of this PR seems to take advantage of this by rewriting it to the last segment of "nc://directlink/f/{fileID}" which would pretty much be the file id. Now, the issue is that the currently proposed nc:// syntax doesn't allow switching accounts if needed, so that should be fixed.

Charon77: @ezaquarii > but rewriting it before handling does not feel right
Thank you for your review. After viewing the code, I realize that this part of the code can be moved to:

```
// FileDisplayActivity.java
private void handleOpenFileViaIntent(Intent intent)
```
@mario I think we can pass the username in the URI, and then when we can do

`Account newAccount = getUserAccountManager().getAccountByName({username})`

I believe we need to establish the required URI format. Any ideas?

I'm thinking of "nc://directlink/f/{fileID}?username={username}", making username optional, which defaults to current user, by calling `BaseActivity.getAccount()`, but this potentially leads to problems as other users who gets the link might get totally different file when the fileID clashes

An alternative would be to make the username mandatory in forms like "nc://directlink/u/{username}/f/{fileID}", or maybe even "nc://directlink/{username}/f/{fileID}"

Charon77:
Question:

I can't seem to find the mechanism that finds a file/folder based on shares. I am hoping to implement this as well. Can someone point me to the right class/method? Also, what do we need in order to get shares to work? do we need the server address & port as well, or do we just need the share 'code' like '.../s/7wYtSxt7n3XTXox'

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
ezaquarii: There are 2 problems here:

1. KEY_FILE_ID is re-used with 2 types of values, which complicates handling
2. it doesn't handle user

I believe that the ideal solution woild be to change the contract/interface and don't re-use extra keys with different arugment formats. Since both apps are already on-line this might not be the most practical appraoch, but since the URI needs a change, maybe there is a window of opportunity for this?

ezaquarii: > An alternative would be to make the username mandatory in forms like 
> "nc://directlink/u/{username}/f/{fileID}", or maybe even "nc://directlink/{username}/f/{fileID}"

I believe URI specification has support for users, so: `nc://user@authority/path` should be ok. This is one thing.

Another one is `directlink` -  generally it is expected (by reader) that the authority part tells which app is handling it. So a link of `nc://user@app_package_name/file_id` would match the convention better.

We could - at some point in time - want to implement direct links in other Nextcloud apps, such as Talk (send file from files app to talk) or in News. In such case, you'll need to come up with creative URI schemes to distinguish between apps.

@mario is this URI still in development phase? If so, could we consider making some alterations before it's stable, relleased and untouchable?

ezaquarii: `com.nextcloud.client` is a generalized (no flavour suffix) form of app package for the files app, btw. We might be willing to have deep links into `com.nextcloud.talk` or `com.nextcloud.news` in the future as well.

tobiasKaminsky:
Opening a file from Talk in Files is assumed to work as both "should" have the same accounts already set up.
The direct link is generated via web UI, so we cannot (without big effort) change this, so we need to stick with: nc://directlink/f/{fileID}
I guess it will work this way:
- First we should check if the base url matches any installed account
  - if not: error out
  - if yes: As the file id is unique, we can also switch accounts with this: We loop over all accounts and try to fetch the file. 

Regarding shares: Let us keep this out of this PR and discuss it in a separate issue.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
mario: handleOpenFileViaIntent already handles the case where account is not set, but we should definitely have it in the URI. I'm not sure authority is needed as the prefix actually decides which app(s) can open an URI, but feel free to convince me otherwise @ezaquarii. Keep in mind that user should be in form of "userId@server" so with your suggestion it would be:

"nc://userId@server/file/${file_id}"

As for Talk @ezaquariil, since both Files and Talk are Android app I take advantage of an Intent with the required parameters

mario:
>  The direct link is generated via web UI, so we cannot (without big effort) change this, so we need to stick with: nc://directlink/f/{fileID}

I am not aware of any code that does this already on the server, but I didn't follow this part much. Can you point me to lines of code in the server that already generate this link? Otherwise, I think we should get it right, even if it means breaking the interface as not many people will know about this anwyay.

juliushaertl:
>> The direct link is generated via web UI, so we cannot (without big effort) change this, so we need to stick with: nc://directlink/f/{fileID}

> I am not aware of any code that does this already on the server, but I didn't follow this part much. Can you point me to lines of code in the server that already generate this link? Otherwise, I think we should get it right, even if it means breaking the interface as not many people will know about this anwyay.

This is not used anywhere afaik, it was just a proposal when discussing this with @tobiasKaminsky. And I agree with @mario Just sending the file id to the client doesn't help much without the server name.

rullzer:
> > > The direct link is generated via web UI, so we cannot (without big effort) change this, so we need to stick with: nc://directlink/f/{fileID}
> 
> > I am not aware of any code that does this already on the server, but I didn't follow this part much. Can you point me to lines of code in the server that already generate this link? Otherwise, I think we should get it right, even if it means breaking the interface as not many people will know about this anwyay.
> 
> This is not used anywhere afaik, it was just a proposal when discussing this with @tobiasKaminsky. And I agree with @mario Just sending the file id to the client doesn't help much without the server name.

Me neither.
But you will need both

1. server
2. userid

Else you can't uniquely identify accounts.

mario:
@juliushaertl @rullzer thanks for confirming! :)

Charon77:
By the way, I just found out that we could set the intent filter `android:host=*` which would match all hosts. we could then either use `android:pathPattern=/index.php/f/*' and hopefully it's nextcloud specific. We could also just use the normal direct link, replace the http scheme with nc, and it will become `nc://{accountname}/index.php/f/{fileId}`. This seems to solve all issues, because we now have specific accountname and fileId, and we can use what the nextcloud/server already use. Users just need to copy the direct link, and change the `http` into `nc`.

Thoughts?

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -2628,8 +2628,27 @@ private void handleOpenFileViaIntent(Intent intent) {
         showLoadingDialog(getString(R.string.retrieving_file));
 
         String accountName = intent.getStringExtra("KEY_ACCOUNT");
+        String fileId = String.valueOf(intent.getStringExtra(KEY_FILE_ID));
+
+        if (accountName == null && "null".equals(fileId) && intent.getData() != null) {
```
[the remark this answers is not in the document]
ezaquarii: There is no guarantee that `valueOf` will resolve to expected string. Java is strongly typed language and a proper type should be used. `instanceof` can be used.

ezaquarii: There is no guarantee that valueOf will resolve to expected string. Java is strongly typed language and a proper type should be used. instanceof can be used.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -2628,8 +2628,27 @@ private void handleOpenFileViaIntent(Intent intent) {
...
+
+            try {
+                // Matches /{accountName}/f/{fileId}
+                accountName = pathSegments.get(0);
+                if ("f".equals(pathSegments.get(1))) {
+                    fileId = pathSegments.get(2);
+                }
+            } catch (IndexOutOfBoundsException e) {
+                dismissLoadingDialog();
+                DisplayUtils.showSnackMessage(this, "Invalid URL");
```
ezaquarii: This string should be translated.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -2628,8 +2628,27 @@ private void handleOpenFileViaIntent(Intent intent) {
...
+            // Handle intent coming from URI
+            List<String> pathSegments = intent.getData().getPathSegments();
+
+            try {
+                // Matches /{accountName}/f/{fileId}
+                accountName = pathSegments.get(0);
+                if ("f".equals(pathSegments.get(1))) {
+                    fileId = pathSegments.get(2);
+                }
+            } catch (IndexOutOfBoundsException e) {
```
ezaquarii: Please properly validate list size before using it. Exceptions are not meant to be used as a flow control mechanism.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
Charon77: @ezaquarii Okay, I will do that. The `String fileId = String.valueOf(intent.getStringExtra(KEY_FILE_ID));` part confuses me to, as I did not write it. I only move the expression upwards.

AndyScherzinger: @tobiasKaminsky can you comment on this?

szaimen:
> In addition, the intent filter now uses the url `http://{accountName}/index.php/f/{fileId}`, which is what the Web UI uses for direct link, meaning users should now be able to click on the direct link and be directed to the app.

I've tested this and it doesn't work with `https://` links only `http://`

Also the default internal link of the webinterface and android app doesn't include the userid afaik, so it doesn't work with normal links. Wouldn't it work if you just iterate over available user accounts of the android app? Because most users should probably just have one account in their app. 
Another solution would be using the current active account in the app and iterate over the other accounts if more are available and it didn't find any match for the currently active account.

tobiasKaminsky:
> This is not used anywhere afaik, it was just a proposal when discussing this with @tobiasKaminsky. And I agree with @mario Just sending the file id to the client doesn't help much without the server name.

Well, it is somewhere(?) used, as if you click on such a direct link, it redirects you to the correct file.
So at least this must be changed, or?

I'll create an issue on server repo, to discuss this and then we can use the new link syntax also here.

tobiasKaminsky:
Server repo to discuss new direct link: https://github.com/nextcloud/server/issues/19294

src/main/AndroidManifest.xml
```diff
@@ -117,6 +117,15 @@
             <intent-filter>
                 <action android:name="android.intent.action.SEARCH" />
             </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="http"
+                      android:host="*"
+                      android:pathPattern="/index.php/f/.*"
```
tobiasKaminsky: @rullzer do you know if the directLink always contains "index.php" or can there be setups where no index.php is used?

AndyScherzinger: highly depends on rewrite rules I guess

juliushaertl: The URL with index.php should always work, we use that in some other places where we don't know about the actual rewriting settings as a save fallback.

szaimen:
What if you use the domain, (e.g. in the case of `http://localhost/index.php/f/1106` would that be `localhost`) to identify all matching accounts that are signed in into the app and iterate over those found accounts (the active one first)?
That should work in 99.99% of all cases and wouldn't need to change the internal link.

So a bit shorter:
1. check for all links that contain `index.php/f/`
2. get the domain, in this example `localhost`
3. get all accounts that are existant in the Nextcloud-Android app that match the domain, e.g. `admin@localhost` or `user@localhost`
4. try to open the file with the currently active account, so e.g. if `admin@localhost` is currently active, you could try to open the file or folder with this account first.
5. If no file or folder was found iterate over all other found users (that have ` @localhost` in their account name).
6. if no file or folder was opened send a message, that you don't have access to the file/folder or it is no longer existant.

src/main/AndroidManifest.xml
```diff
@@ -117,6 +117,15 @@
             <intent-filter>
                 <action android:name="android.intent.action.SEARCH" />
             </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="http"
```
szaimen: Maybe add an additional `android:scheme="https"` here?
Then it works with https links, too.

mario:
> What if you use the domain, (e.g. in the case of `http://localhost/index.php/f/1106` would that be `localhost`) to identify all matching accounts that are signed in into the app and iterate over those found accounts (the active one first)?

99,99% means it's not secure enough.

src/main/AndroidManifest.xml
mario: Yup, that's what needs doing.

szaimen:
> 99,99% means it's not secure enough.

Actually I would say it is 100% secure. This was just a phrase.

mario:
> > 99,99% means it's not secure enough.
> 
> This was just a phrase.
> Actually I would say it is 100% secure.

See my edit - like I said the approach has its issues in theory, but might indeed be applicable in the real world.

szaimen:
> But even if we accept this, it's non-trivial - it PROBABLY means N search queries (where N == number of accounts).
> 
> (Might not be a bad idea overall in reality, but needs proper UX)

I would say by far most users just use one account with the app (have no proof for that statement).

And for everyone else which uses a lot accounts, this should work as well but be in the worst situation  it would be a bit slow. ( e.g. if you are signed in with 10 accounts on the same server at the same time with the app and the last one is the one that has access to the file). 
But that is very unlikely imo because most of the time the account that is active should be the standard working account and therefore will be checked first.

Maybe you could remember the last used accounts or the most used accounts and check them in this order.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
ezaquarii: @Charon77 Thanks. We try to do cleanups whenever possible.

mario:
It's irrelevant if an account owns a file as long as one has access to it AFAIK. So I guess that's indeed the way to go - iterate all accounts, do a search and present a UI/popup asking which account you want to use if more than 1 account matches/has file access.

szaimen:
@Charon77 is there a way to test the newest commits? Would be really nice! :)

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -2664,5 +2689,59 @@ private void handleOpenFileViaIntent(Intent intent) {
...
+
+    }
+
+    private void findAccountAndOpenFile(String hostName, int port, String fileId) {
+        String strPort = Integer.toString(port);
+
+        ArrayList<String> validAccounts = new ArrayList<>();
+
+        for (Account account : accountManager.getAccounts()) {
+            String accountHostPort = account.name.split("@")[1];
```
ezaquarii: We have discussed the format of account name extensively when trying to formalize account ID:

https://github.com/nextcloud/android/pull/5006

AFAIR, the conclusion was that we cannot rely on this format, but maybe we should take a step back and look at it again.

@tobiasKaminsky @mario @AndyScherzinger

AndyScherzinger: I thought we can rely on the format as in <userId>@<server-address>, no?

szaimen: Actually I am also not quite sure. This is from the docs:
sketch1581462075813
But for now I would say it is good enough?

AndyScherzinger: Well, with this info we would have to look for the last @ character to split the string I guess.

Charon77: Solved in bcd277f

tobiasKaminsky: Sorry for seeing this too late, but we have:
https://github.com/nextcloud/android-library/blob/master/src/main/java/com/owncloud/android/lib/common/accounts/AccountUtils.java#L101

This should be enough and we stay consistent with our other account handling.

src/main/AndroidManifest.xml
tobiasKaminsky: But it is the other way around:
We want to check for every share link, but they are created by server and can be either
http://cloud.com/index.php/f/123 or
http://cloud.com//f/123
or?

So we cannot rely on "/index.php/f/.*" pattern?

tobiasKaminsky: We also have to think about subdirectories, so full potential list would be:
http://cloud.com/index.php/f/123
http://cloud.com//f/123
http://cloud.com/subfolder/index.php/f/123
http://cloud.com/subfolder/f/123

(subdomains should not be affected)

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
tobiasKaminsky: Better is to use: UserAccountManager, as this is our own implementation and we try to get rid of AccountManager.
From userAccountManager you can get User, which then has Server and this has Uri, which is also the baseUrl.

Charon77: 891916e

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -2664,5 +2691,48 @@ private void handleOpenFileViaIntent(Intent intent) {
...
         fetchRemoteFileTask.execute();
+
+    }
+
+    private void findAccountAndOpenFile(String scheme, String authority, String fileId) {
+
+        ArrayList<String> validAccounts = new ArrayList<>();
+        String queryBaseUrl = scheme + "://" + authority;
+
+        for (Account account : accountManager.getAccounts()) {
```
ezaquarii: Please do not use platform API. We try to get rid of it. We have our own abstraction of account: `UserAccountManager` and `User`.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -2664,5 +2691,48 @@ private void handleOpenFileViaIntent(Intent intent) {
...
+            try {
+                if (AccountUtils.getBaseUrlForAccount(this, account).equals(queryBaseUrl)) {
```
ezaquarii: I'm not sure what we do here.

Is there any reason why we try to build the URL ourselves and then use `getBaseUrlForAccount` only to see if we built it correctly?

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -2664,5 +2691,48 @@ private void handleOpenFileViaIntent(Intent intent) {
...
+                    validAccounts.add(account.name);
+                }
+            } catch (AccountUtils.AccountNotFoundException ignored) {
+            }
+        }
+
+        if (validAccounts.size() == 0) {
+            openFile(getAccount().name, fileId);
```
ezaquarii: Please use `User` instead of `Account`.

ezaquarii:
Rebase is needed as well.

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
Charon77: given direct link from the intent in forms of 

http://example.com:567/index.php/f/123

I'd like to extract the 
http://example.com:567
part.

queryBaseUrl is from intent. getBaseUrlforAccount is from the account in getAccounts()

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
Charon77: Not so sure why, but alright

04fc793

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
Charon77: 6d4ec4e

src/main/AndroidManifest.xml
szaimen: In my testing have the internal links always contained "/index.php/f/" no matter if you have pretty urls enabled or not. So we should can reliably rely on this pattern I would say.

But yes, subdomains also should considered here, too.
That was also the reason why I suggested to use the hostdomain here, so everything between "http://" or "https://" and "/index.php/f/"

Also shouldn*t the username then also contain this path, that wen can check against, so e.g. "user@cloud.com/subfolder"?

src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
ezaquarii: Because we do this: https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#custom-platform-apis

Charon77: Ah, my bad. Thanks for pointing out

src/main/AndroidManifest.xml
szaimen: @Charon77 Is this review thread already solved?

Charon77: not so sure what's going on here. can you give a test case?

szaimen: @tobiasKaminsky has provided some (but not all are possible afaik)
Correct ones are e.g.:
http://cloud.com/index.php/f/123
But also:
http://cloud.com/subfolder/index.php/f/123

So would this PR in the current state also work with link like the second one? Because it is actually possible to install Nextcloud like this (in a subfolder) which would create such links.

mario: Yes it would.

szaimen: Nice! And btw: is there a possibility to test this PR? (The apk-creation seems to have failed...)

szaimen:
Any chance to get this in for 3.11?

src/main/AndroidManifest.xml
Charon77: 7538fc0c6

@szaimen try this out:

https://github.com/Charon77/android/releases/tag/7538fc0c6

szaimen: @Charon77 Thanks for linking me, I've tried the app and there seems to be a bug:
now all subdomains work, but also other domains, that are not the same like the domain, the user is from:
So if my user is admin@localhost, links like https://localhost/index.php/f/39503 work but also e.g. https://cloud.com/index.php/f/35903 work and I think this shouldn't be the case here. The only different to the localhost link is, that it opens the file directly with the current account without asking which account you want to use when clicking in that link.
But I think it should say, that the link isn't working, and/or open that link directly in the Browser...

Charon77: Can you test this out:

This no longer defaults to current user when account is not found, and instead says that the account is not found. Opening the link in browser might be a bit tricky.

https://github.com/Charon77/android/releases/tag/aff6be0e

szaimen: I would like to, but I cannot install the apk you have linked.
Maybe one of those links can help you achieving the opening in the browser if no account was found? 
https://stackoverflow.com/questions/2201917/how-can-i-open-a-url-in-androids-web-browser-from-my-application
https://stackoverflow.com/questions/3004515/sending-an-intent-to-browser-to-open-specific-url
Maybe it should then say to the user, that "No Account matching the URL was found, opening the link in the browser..."  (Or something like that)
And directly open the link in the standard-browser afterwards.

Charon77: @szaimen Try the updated apk:
https://github.com/Charon77/android/releases/tag/aff6be0e
the file is **aff6be0e_2.apk**

Currently, if no matching user is found, it will say Associated user is not found.

There will be problems by opening the link in the browser (and kind of defeats the purpose of  the app). I think the case when no account matches the URL should be a separate issue.

Charon77:
Hello, this PR has been going for quite a while, I think I'd ask for another review and maybe get this merged.

I've posted the latest APK in:

https://github.com/Charon77/android/releases/tag/aff6be0e

with the filename:

aff6be0e_2.apk

tobiasKaminsky:
I wanted to test this right now, so I added a website with 
`<a href="http://10.0.2.2/nc/index.php/f/22994">file id</a>`
and clicked on it, but it opened the website instead of letting me choose Nextcloud app.
I used Chrome on Android 10.

Charon77:
> I wanted to test this right now, so I added a website with
> `<a href="http://10.0.2.2/nc/index.php/f/22994">file id</a>`
> and clicked on it, but it opened the website instead of letting me choose Nextcloud app.
> I used Chrome on Android 10.

I'm making a test link based on your URL. Let me see if it works...

[test link](http://10.0.2.2/nc/index.php/f/22994)

Charon77:
ncdl

@tobiasKaminsky 

I guess there's something I don't quite understand about how intent filters work. In the meantime, try scanning this. On my Android 10, scanning this brings up the dialog to choose Nextcloud

Charon77:
@tobiasKaminsky 

I have several apps that allow me to type in the URL and gets converted into links. Some apps show the dialog to open in nextcloud while others don't.

Does the dialog to open in Nextcloud shown?

* Discord: yes

* Naver LINE: yes

* Facebook Messenger: you have to click the triple dots and click "Open with", then it allows you to choose Nextcloud

* Chrome: no

* Firefox: hold click on the link and there's  "Open with Nextcloud app" on the menu

Charon77:
@tobiasKaminsky

Hi, I figured it out. Because on how Chrome handle stuffs their own way, in order for any link to be recognized by chrome, you need to follow their URI spec from [here](https://developer.chrome.com/multidevice/android/intents)

You should be able to click this 
<a href="intent://10.0.2.2/nc/index.php/f/22994#Intent;scheme=http;package=com.nextcloud.client;end">this link</a>

src/main/AndroidManifest.xml
szaimen: @Charon77 I don't think that it would defeat the purpose of the app. Because if no user is found, is it probably not a link from your Nextcloud that you have clicked on and/or no Nextcloud at all.
For those cases it would make much sense if the app would be able to open the link in the browser directly.

And please tell me which problems do you see for opening the link directly in the browser (if no account was found). Maybe they have a good foundation but I have to know them. Otherwise we cannot discuss them.

I will try the updated apk later on.

Charon77: First of all, I have not tried the link you suggested but I have read them.

Here's a scenario I think of where it's problematic. If a user opens a "potentially nextcloud URL", the user would be presented with choices: Would you like to open the URL in your browser, or nextcloud? And user would click "Always" or "Just once". Let's assume the user clicks nextcloud, and click "Always". Then, we realize that the URL is **not** a Nextcoud URL. However, since the Nextcloud app is set to perform Intent.ACTION_VIEW on said URL, it will relaunch the same activity, and says that there's no account, and the whole thing will run forever.

szaimen: @Charon77 Thanks for sharing your thoughts!
I can understand your concerns, but the idea was to prevent the loop from happening through launching directly a browser activity. I don't know if it works, but it theoratically should.

IMO, launching the browser directly through the activity would be the best UX, in the case, that no matching account was found, but I could also imagine a UI which is asking the user if you want to continue launching the link in the browser. Then you could click on no or yes.

tobiasKaminsky:
Tested and works for me.
Can you rebase, resolve conflict and squash? 

Great work!
(I am off tomorrow/Wed, but at latest I will merge it on Thursday)

src/main/AndroidManifest.xml
szaimen: @Charon77 maybe one of those helps?
https://stackoverflow.com/a/12013755
https://stackoverflow.com/a/18682857
https://stackoverflow.com/a/23268821
https://stackoverflow.com/a/23622337
https://developer.android.com/training/basics/intents/sending#java

so maybe something lilke this:
(In theory you could first try to get the package name of the default browser and the parse the URL with "setPackage")

```
Intent browserIntent = new Intent("android.intent.action.VIEW", Uri.parse("http://"));  
ResolveInfo resolveInfo = getPackageManager().resolveActivity(browserIntent,PackageManager.MATCH_DEFAULT_ONLY);

// This is the default browser's packageName
String packageName = resolveInfo.activityInfo.packageName;
```

and afterwards use that string to set a Package:
```
   String url = "http://www.example.com";
    Intent i = new Intent(Intent.ACTION_VIEW, Uri.parse(url));
    i.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
    i.setPackage("com.android.chrome");
    try {
        startActivity(i);
    } catch (ActivityNotFoundException e) {
        // Chrome is probably not installed
        // Try with the default browser
        i.setPackage(null);
        startActivity(i);
    }
```

actually the code is not complete, because the string url would be the provided URL in the first place, and of course instead of using "com.android.chrome" you would use the default browser aka "packageName"
also, I am not sure, if you need the catch (ActivityNotFoundException e) since you are always using the default browser which is always existing...

szaimen:
I've upgraded my server today to 18.0.1 and it seems like at least the webinterface does respect the pretty urls now. So the internal links from the webinterface looks like this now: `https://localhost/f/585482`. to make it work with those linke we probably need another:
```
                <data android:scheme="http"
                      android:scheme="https"
                      android:host="*"
                      android:pathPattern="/.*f/.*"
                    />
            </intent-filter>
```

btw: I think that should work, too (put everything into one data section)
```
                <data android:scheme="http"
                    android:scheme="https"
                    android:host="*"
                    android:pathPattern="/.*index.php/f/.*"
                    android:pathPattern="/.*f/.*"
                    />
```

Charon77:
> I've upgraded my server today to 18.0.1 and it seems like at least the webinterface does respect **pretty urls** now. So the internal links from the webinterface look like this now: `https://localhost/f/585482`. to make it work with those links we probably need another:
> 
> ```
>                 <data android:scheme="http"
>                       android:scheme="https"
>                       android:host="*"
>                       android:pathPattern="/.*f/.*"
>                     />
> ```
> 
> btw: I think that this should work, too (putting everything into one data section)
> 
> ```
>                 <data android:scheme="http"
>                     android:scheme="https"
>                     android:host="*"
>                     android:pathPattern="/.*index.php/f/.*"
>                     android:pathPattern="/.*f/.*"
>                     />
> ```
> 
> another btw: Because the new pattern is also inside the index.php pathpattern, the index.php pathpattern is probably no longer needed. So this is probably enough:
> 
> ```
>                 <data android:scheme="http"
>                     android:scheme="https"
>                     android:host="*"
>                     android:pathPattern="/.*f/.*"
>                     />
> ```
> 
> If filtering for "/f/", I think that beeing able to open that link in the browser if no matching account is found has gotten more important. See [#5345 (comment)](https://github.com/nextcloud/android/pull/5345#discussion_r383312633) and downwards...

Yeah, I think matching against /f/ is a bit too catch-all.

I think it's true that handling matching account is important, I think this should be handled in separate pull request, along with options to sign in with new account, if the direct link goes to a new server.

szaimen:
One last question: the app seems to get opened on top of the current used app. Is this intentional?
Screenshot_20200226-185538_Nextcloud QA

Charon77:
@szaimen

huh? that's weird. Can you reproduce it?

szaimen:
Yes, every time

Charon77:
I mean, how do you reproduce the bug? What are the steps? Sre you sure it's related to this PR?

szaimen:
Yes, I have downloaded and installed the latest apk.
i just click on that link in the tasks app I have installed and it launches on top of the tasks app. (the Nextcloud QA app doesn't launch, it is just here to show that it seems to be another instance...)

Maybe you need to add that flag?` i.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);`

szaimen:
Maybe a better screenshot is this:
photo5384461690403400827

Charon77:
> Yes, I have downloaded and installed the latest apk.
> i just click on an internal-nextcloud-link in the tasks app, that I have installed and it launches Nectcloud QA on top of the tasks app. (the Nextcloud QA app doesn't launch inits own window, it is just here to show that it seems to be another instance...)
> 
> Maybe you need to add that flag?` i.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);`

@szaimen 

That is the intended behaviour, as you are calling an app via direct linking. Similar behaviours can be observed such as when you click on a PDF link from browser, and it opens the PDF in the browser window.

adding intent flags is irrelevant here, as those are used when you're calling activity from another activity. In this case, we're calling the activity from just the link, so it's up to the calling application's implementation, not ours.

Charon77:
@tobiasKaminsky any comments?

szaimen:
> adding intent flags is irrelevant here, as those are used when you're calling activity from another activity. In this case, we're calling the activity from just the link, so it's up to the calling application's implementation, not ours

Okay you are right. It works in other apps as expected.

@tobiasKaminsky So this is ready to go?

I just wanted to throw in again, that if you are using pretty urls on your NC and are on NC18 and above, this PR will change nothing. But as you sad: maybe this should get fixed by another PR.

tobiasKaminsky:
Great work :tada:

jakobroehrl:
@Charon77 
It's working great, waited a long time for this feature, thanks!

PR #5399, opened 2020-02-05
extract ArbitraryDataSet class

tobiasKaminsky:
- cleanup

(just something little to free up my mind)


src/main/java/com/owncloud/android/datamodel/ArbitraryDataSet.java
```diff
@@ -0,0 +1,43 @@
...
+
+package com.owncloud.android.datamodel;
+
+import lombok.Getter;
+
+/**
+ * Data set for {@link ArbitraryDataProvider}
+ */
+@Getter
+class ArbitraryDataSet {
```
AndyScherzinger: should this be java or since it is a new file being re-created in kotlin?

tobiasKaminsky: As it is only refactoring, I would like to keep it Java #lazyness :see_no_evil:

AndyScherzinger: OK

PR #5400, opened 2020-02-05
l10n: Grammar: url -> URL

rakekniven:
Signed-off-by: rakekniven <mark.ziegler@rakekniven.de>


tobiasKaminsky:
Thank you :-)

tobiasKaminsky:
/backport to stable-3.10

(otherwise transifex will fail, as we then have same key, but different value)

PR #5414, opened 2020-02-06
Fix crash: Caused by: java.lang.ClassCastException: java.lang.Object[…

tobiasKaminsky:
…] cannot be cast to java.lang.Void[]

### Testing 
This is unfortunately too complex to test it in an easy way…


AndyScherzinger:
@tobiasKaminsky lint score need to be raised (nothing we can do about it)

PR #5450, opened 2020-02-11
Screenshot tests for ETM

tobiasKaminsky:

### Testing 

- [x] [unit tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#unit-tests)
- [x] [instrumented tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#instrumented-tests)
- [x] [UI tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#ui-tests)


src/main/java/com/nextcloud/client/etm/EtmActivity.kt
```diff
@@ -74,7 +74,7 @@ class EtmActivity : ToolbarActivity(), Injectable {
         }
     }
 
-    private fun onPageChanged(page: EtmMenuEntry?) {
+    fun onPageChanged(page: EtmMenuEntry?) {
```
tobiasKaminsky: @ezaquarii I had to set this function to public in Kotlin.
In Java I thought it would have been enough to set it to protected.
Do you have an idea?

ezaquarii:
This callback is driven from the ViewModel. If you call if by hand, UI instance will be out of sync with the VM data model.

Page switching should be driven via VM API.

Since VM is injected and has package private access, this should be fairly easy.

src/main/java/com/nextcloud/client/etm/EtmActivity.kt
ezaquarii: Yeah, generally this is a callback that is driven from the `ViewModel`.

https://github.com/nextcloud/android/blob/bc0b9e78f18ca468b20c4b740646084517673b2f/src/main/java/com/nextcloud/client/etm/EtmActivity.kt#L54-L56

So, by calling it outside VM, the UI is no longer in sync with view model.

The proper way to do it is to initiate the page change via `EtmViewModel.onPageSelected(pageIndex: Int)` API, which is wired to a list item tap event. By calling it you "simulate" user tapping the list item and everything will stay in sync.

src/androidTest/java/com/nextcloud/client/etm/EtmActivityTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        waitForIdleSync()
+
+        Screenshot.snapActivity(sut).record()
+    }
+
+    @Test
+    fun preferences() {
+        val sut: EtmActivity = activityRule.launchActivity(null)
+
+        UiThreadStatement.runOnUiThread(Runnable { sut.onPageChanged(sut.vm.pages[0]) })
```
ezaquarii: Try converting it to something like 

```
UIThreadStatement.runOnUiThread(Runnable {
    vm.onPageSelected(0)`
})
```
`vm` is internal so it's accessible.

tobiasKaminsky: Nice :+1:

PR #5458, opened 2020-02-12
Disable ANDROIDX_TEST_ORCHESTRATOR

tobiasKaminsky:
It seems that with enabled ANDROIDX_TEST_ORCHESTRATOR we do not test all screenshot tests, but only the latest, as the in between generated screenshots get removed.

This is for testing.


ezaquarii:
> It seems that with enabled ANDROIDX_TEST_ORCHESTRATOR we do not test all screenshot tests, but only the latest, as the in between generated screenshots get removed.

Can you elaborate on this? Where do we store the screenshots?

AFAIK, for very long time Android suffered from poor test isolation in general, to a point where in bigger projects ppl just refused to do tests. `Androidx` tried to ease this problem somehow - does it mean that they clean something between test runs?

If so, mybe we could fix this somehow?

tobiasKaminsky:
I created an issue in our library's repo, with a sample project.
https://github.com/Karumi/Shot/issues/78

As far as I read this flag leads to a clean situation, as it somehow(tm) sandboxed and thus a previous state of the app will not affect the next test.
I assume that this is the problem, but did not digged into it.

tobiasKaminsky:
Seems to work with Shot:4.x

PR #5471, opened 2020-02-14
Show "add folder info" only if text app is available

### LINKED ISSUE #5467: Failed to start editor
I just upgraded to version 3.10.1 from f-droid.
There's a new option in the up-right menu "Add folder info", but every time I select it, I just get a "Failed to start editor" message.
Am I missing something on android device (version 5.1) or it's a server side problem (missing app on 16.0.7 version)?

tobiasKaminsky:
Fix #5467 

@juliushaertl I test if an editor is available for "text/markdown" and then show/not show the "add folder info".
This should be correct for any scenario
- old server
- 18 server, but text disabled

Is my assumption correct?

### DIFF
--- a/src/main/java/com/owncloud/android/ui/fragment/OCFileListBottomSheetDialog.java
+++ b/src/main/java/com/owncloud/android/ui/fragment/OCFileListBottomSheetDialog.java
@@ -39,6 +39,7 @@ import com.nextcloud.client.device.DeviceInfo;
 import com.owncloud.android.R;
 import com.owncloud.android.datamodel.ArbitraryDataProvider;
 import com.owncloud.android.datamodel.OCFile;
+import com.owncloud.android.files.FileMenuFilter;
 import com.owncloud.android.lib.common.Creator;
 import com.owncloud.android.lib.common.DirectEditing;
 import com.owncloud.android.lib.resources.status.OCCapability;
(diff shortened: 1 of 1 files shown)


juliushaertl:
> I test if an editor is available for "text/markdown" and then show/not show the "add folder info". This should be correct for any scenario
> - old server
> - 18 server, but text disabled

Yes except for one case: We also have a server setting to disable the rich workspaces per user or on the whole instance, but it is not exposed in the capabilities yet. It will still work fine on Android even if it is disabled on the server side, but for feature parity I guess we should also respect that setting on mobile.

tobiasKaminsky:
> Yes except for one case: We also have a server setting to disable the rich workspaces per user or on the whole instance, but it is not exposed in the capabilities yet.

We then should (at some point) expose it via directEditing, and then this would work out of the box here, or?

juliushaertl:
> We then should (at some point) expose it via directEditing, and then this would work out of the box here, or?

Well, no, since this is different from direct editing. You can have rich workspaces disabled but still use direct editing with text. I guess we would just add a capability for this one and if set the markdown preview on top of the files list is hidden.

tobiasKaminsky:
> > We then should (at some point) expose it via directEditing, and then this would work out of the box here, or?
> 
> Well, no, since this is different from direct editing. You can have rich workspaces disabled but still use direct editing with text. I guess we would just add a capability for this one and if set the markdown preview on top of the files list is hidden.

This will be handled via https://github.com/nextcloud/server/issues/19548

PR #5482, opened 2020-02-17
Migrate contacts backup job to new background job manager API

ezaquarii:
@tobiasKaminsky @AndyScherzinger @mario This is still WIP but a review would be handy nevertheless.

This is rather big PR containing contacts backup job migration to new `androidx.work.WorkerManager`. To migrate it, I had to implement some missing infrastructure, hence the scope of this PR is much larger than  I intitially anticipated. To make matters more interesting, it will get bigger (see last 2 sections).

## New abstraction over `WorkManager`

`BackgroundJobManager` is our application-specific abstraction, exposing business-specific API to control running jobs. It hides all job handling machinery behind a very goal-oriented, verb-based API.

Important difference between old job manager API is that all `WorkManager` API is asynchronous. Exposed API is based on Java `Future` primitive and `LiveData<T>`. New semantics make some operations more tricky and porting of previously synchronous code needs some thinking, but no problems has been spotted so far.

One limitation of `WorkManager` API is that it does not provide elegant way of storing job metadata - like the previously used `com.evernote.android.job.JobRequest.getExtras()` - and provides onlylimited API in form of `Set<String> WorkInfo.getTags()`. The missing functionality has been implemented in as `JobInfo` abstraction that replaces `WorkInfo`. `JobInfo` is created by parsing tags stored in `WorkInfo`. Type translation is handled by the `BackgroundJobManagerImpl` and those complexities are not visible for the end user of our API.

Observable collection of all jobs is exposed using `LiveData<T>` interface, so it should play nice with the UI code.

### Background Jobs

Viewing current background job status is quite important do debug any potential issues, so I added `EtmBackgroundJobsFragment` with a list of all registered jobs and handy menu where we can dump development actions. It currently allows us to cancel all tasks and toggle `TestJob` that demonstrates progress indicator and is used to experiment with platform API.

### Migrations

This PR provies a small screen with migrations status. It lists migrations and allows to clear migrations status. This function can be very useful when creating new migrations.

## `BackupContactsJob`

Old code has been copied almost verbatim from old job class, with some minor tweaks around dependencies and few cleanups. There is no unit test for it, but it seems to work ok.

## `ContentObserverWork`

Yikes! We have a bug in our current manager implementation and the observer job is added to the queue each time the app starts, which means that users are accumulating those jobs in their queues. This is not a big deal, but might eat up resources. Since the `ContentObserverWork` only starts legacy job implementation, the damange is contained.

This issue is addressed and the observer job is started only once.

Old job is cleared.

## `User.nameEquals`

Extended user API with account name equality check, as we need to do a log of `String` comparisions.

## Missing things

- [x] Robolectric / test location decision
- [x] Remove Robolectric dependency fom tests
- [x] Lombok workaround must be fixed
- [x] Merge #5447  so I can rebase this work on `master`
- [x] ~Maybe refactor `ContactsBackupWork` to enable testing~
- [x] ~Remove legacy `ContactsBackupJob` class~
- [x] ~Design migration module~ See #5546 
- [x] ~Decide how to handle failing migrations~ See #5546
- [x] ~Add migration module to DI and enable it in `MainApp`~ See #5546
- [x] Linter and CI cleanups
- [x] Remove draft migrations code from this PR and rebase on top of #5546

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -884,6 +884,18 @@
     <string name="etm_title">Engineering Test Mode</string>
     <string name="etm_accounts">Accounts</string>
     <string name="etm_preferences">Preferences</string>
+    <string name="etm_background_jobs">Background jobs</string>
+    <string name="etm_background_jobs_cancel_all">Cancel all jobs</string>
+    <string name="etm_background_jobs_prune">Prune inactive jobs</string>
+    <string name="etm_background_jobs_start_test_job">Start test job</string>
+    <string name="etm_background_jobs_stop_test_job">Stop test job</string>
+    <string name="etm_background_job_uuid">UUID</string>
+    <string name="etm_background_job_name">Job name</string>
+    <string name="etm_background_job_user">@string/user_icon</string>
+    <string name="etm_background_job_state">State</string>
+    <string name="etm_background_job_started">Started</string>
+    <string name="etm_background_job_progress">Progress</string>
+    <string name="etm_migrations">Migrations (app upgrade)</string>
 
     <string name="logs_status_loading">Loading…</string>
     <string name="logs_status_filtered">Logs: %1$d kB, query matched %2$d / %3$d in %4$d ms</string>
--- a/src/androidTest/java/com/nextcloud/client/jobs/BackgroundJobManagerTest.kt
+++ b/src/androidTest/java/com/nextcloud/client/jobs/BackgroundJobManagerTest.kt
@@ -1,0 +1,371 @@
+/*
+ * Nextcloud Android client application
+ *
+ * @author Chris Narkiewicz
+ * Copyright (C) 2020 Chris Narkiewicz <hello@ezaquarii.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.nextcloud.client.jobs
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.Observer
+import androidx.test.annotation.UiThreadTest
+import androidx.work.Data
+import androidx.work.ExistingPeriodicWorkPolicy
+import androidx.work.ExistingWorkPolicy
+import androidx.work.OneTimeWorkRequest
+import androidx.work.PeriodicWorkRequest
+import androidx.work.WorkInfo
+import androidx.work.WorkManager
+import com.nextcloud.client.account.User
+import com.nextcloud.client.core.Clock
+import com.nhaarman.mockitokotlin2.KArgumentCaptor
+import com.nhaarman.mockitokotlin2.any
+import com.nhaarman.mockitokotlin2.argThat
+import com.nhaarman.mockitokotlin2.argumentCaptor
+import com.nhaarman.mockitokotlin2.eq
+import com.nhaarman.mockitokotlin2.mock
+import com.nhaarman.mockitokotlin2.verify
+import com.nhaarman.mockitokotlin2.whenever
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.Suite
+import org.mockito.ArgumentMatcher
+import java.util.Date
+import java.util.UUID
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+
+/**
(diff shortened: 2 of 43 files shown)


tobiasKaminsky:
Wow, this is big, a bit too big… ;-)

Let me first answer your post

New abstraction looks very very good!
Same goes for ETM, which is really a big win for testing/RC phase.

Lombok:
This seems to be only for modificationTimestamp needed, or? So to avoid a (for other confusing) helper class, we could write explicitly boilerplate code.
If we see that a class is getting too much boilerplate code, we should migrate it to Kotlin.
(as far as I understand, this would solve the problem, or?)

Testing:
I hesitate to add in this PR Robolectric. What about using regular androidTests here and in a later PR we can introduce/test Robolectric?
(maybe we can then also move already existing instrumentation tests, which then speeds overall test duration)

Migration:
I fail to see what account migration has to do with job management? Sorry, if I miss the obvious :-)

I will start reviewing it now.

src/main/java/com/nextcloud/client/jobs/Helper.java
```diff
@@ -0,0 +1,13 @@
+package com.nextcloud.client.jobs;
+
+import com.owncloud.android.datamodel.OCFile;
+
+public class Helper {
```
AndyScherzinger: fine for now while we should probably find a richer name at some point

tobiasKaminsky: As described by me and Chris, let us better go with removing Lombok for the problematic parameters and if needed change to Kotlin for those files.
Adding this, even temporary, just asks for trouble, or harder understanding of code.

ezaquarii: I'll remove it of course. This was just a temporary hack.

ezaquarii: Removed. Now the offending property has explicit accessors.

ezaquarii:
@tobiasKaminsky I fixed lint and rebased.

Robolectric is still enabled in this branch until I finish the implementation. I'll remove it at later stage.

src/main/res/values/strings.xml
```diff
@@ -887,6 +887,17 @@
     <string name="etm_title">Engineering Test Mode</string>
     <string name="etm_accounts">Accounts</string>
     <string name="etm_preferences">Preferences</string>
+    <string name="etm_background_jobs">Background jobs</string>
+    <string name="etm_background_jobs_cancel_all">Cancel all jobs</string>
+    <string name="etm_background_jobs_prune">Prune inactive jobs</string>
+    <string name="etm_background_jobs_start_test_job">Start test job</string>
+    <string name="etm_background_jobs_stop_test_job">Stop test job</string>
```
ezaquarii: Those strings are really needed only for development. Shall I mark them as `translatable=false`?

src/main/res/values/strings.xml
```diff
@@ -887,6 +887,17 @@
     <string name="etm_title">Engineering Test Mode</string>
     <string name="etm_accounts">Accounts</string>
     <string name="etm_preferences">Preferences</string>
+    <string name="etm_background_jobs">Background jobs</string>
+    <string name="etm_background_jobs_cancel_all">Cancel all jobs</string>
+    <string name="etm_background_jobs_prune">Prune inactive jobs</string>
+    <string name="etm_background_jobs_start_test_job">Start test job</string>
```
ezaquarii: @tobiasKaminsky Those strings are really needed only for development. Shall I mark them as `translatable=false`? Is it costly to be liberal about adding new strings? Linter enforces use of strings.xml.

build.gradle
```diff
@@ -390,6 +390,7 @@ dependencies {
 //    androidJacocoAnt "org.jacoco:org.jacoco.agent:${jacocoVersion}"
 
     implementation "com.github.stateless4j:stateless4j:2.6.0"
+    testImplementation "org.robolectric:robolectric:4.3.1"
```
ezaquarii: This will go in the final iteration.

src/main/res/values/strings.xml
AndyScherzinger: @ezaquarii would they be visible somewhere in any build? (like the nightly aka dev build which is published?)

ezaquarii:
Those strings are visible in ETM screen. I think ETM is disabled in release builds (f-droid and gplay). ETM is mostly for developersand maybe Dev build users.

AndyScherzinger:
We should be fine with adding new strings imho. 

They get pushed to transifex and for the languages that are very actively maintained the translations come in quite quickly.

ezaquarii:
I know. I was worried that it incurs extra cost and since those strings are development only, that might not be desired thing.

src/main/res/values/strings.xml
tobiasKaminsky: As this will be shown within dev and RC version, we should translate it :+1:

ezaquarii:

It looks like my initial appraoch was a bit too simplistic and we need - after all - async migration support.

The account name migration is run in a thread and makes network requests, which can take long time. It does not block BTW, wich raises question about unhappy path. Can it work if this migration failed?

2nd migration - content observer job - in theory cannot fail, so we won't care too much.

Anyway, the new proposal has `LiveData<Status>` that allow us to monitor the migration status. I haven't made a decision what to do if we find migration failing, which is something we should discuss.

AndyScherzinger:
Can really say much about the account migration while I think (rather guess) that failed account migrations might lead to issues afterwards.

ezaquarii:
It looks like the emulator is failing. Otherwise should be green, I hope.

build.gradle
ezaquarii: Robolectric has been removed.

ezaquarii:
@tobiasKaminsky This is ready for final review.

Screenshot tests are failing. The trouble is that those tests are failing 100% if time and I'm pretty sure it's not my code. What is the purpose of those tests? I haven't managed to pass tests a single time for months now. Can we disable them?

tobiasKaminsky:
Wow!
This is really really great.

tobiasKaminsky:
The trash bin should be tinted, but I merged it so that I do not delay further work :-)

Issue #5487, opened 2020-02-18
Multiple instances of UploadsStorageManager


tobiasKaminsky:
> Bug:
> 
>     * have 3 images in Camera
> 
>     * upload them
> 
>     * install this branch
> 
>     * enable "upload previously"
> 
>     * enable upload for folder "Camera"
> 
>     * see that three items as conflict in upload list view +1
> 
>     * click on one, select "keep server"
> 
>     * still see three items
>       --> it should be two

This is a display bug, if you refresh up to date list is displayed.

`FileDownloader L645` and `UploadsStorageManager L238` -> UploadListAdapter doesn't register as an Observer of `UploadsStorageManager`. But there are multiple instances (see screenshot) of  `UploadsStorageManager` so it doesn't work anyway. I need your help @tobiasKaminsky on how you would make this class an effective singleton; else there is no way to properly update the view.

![image](https://user-images.githubusercontent.com/35401485/74518393-0af52680-4f14-11ea-9dd5-58680e34c4b1.png)

@tobiasKaminsky up on my answers to your comments btw.

_Originally posted by @ArisuOngaku in https://github.com/nextcloud/android/pull/4788#issuecomment-586173105_

tobiasKaminsky:
> But there are multiple instances (see screenshot) of `UploadsStorageManager` so it doesn't work anyway.

This should per se be no problem, as all queries are against a database, which synchronizes it.
I guess that we just do not have a callback (or obsverable, as you said) to detect that we have to reload the list.

ArisuOngaku:
Registering a `ContentObserver` on the database means we have to unregister it once we're done with the instance of `UploadsStorageManager`. A new instance of this almost every time an activity is opened (and more), this gets hard to be tracked.

Making this a singleton (should be fine) means that there's only one `ContentObserver` on the db.

Anyway UploadListAdapter has to either:
- register an observer on UploadsStorageManager because it currently doesn't get updated on notifyObservers()
- receive an "upload message" in some way (through `UploadMessagesReceiver`)

See #5488 for 1st solution

PR #5488, opened 2020-02-18
Uploads storage manager sync

ArisuOngaku:
For #5487

### Testing 

- [ ] [unit tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#unit-tests)
- [ ] [instrumented tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#instrumented-tests)
- [ ] [UI tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#ui-tests)


tobiasKaminsky:
> Observing the whole uploads section of the database doesn't seem to make any difference regarding #5487

So this does not help? Or do I understand it wrong?

ArisuOngaku:
If anything directly modifies the database without going through `UploadsStorageManager`, Observing the database directly is important. However this PR actually fixes the original specific issue with its first 2 commits alone, but that's the Observable solution on `UploadsStorageManager` "accessors". There is another mechanism, "messages", that is an alternative and I haven't studied it yet, hence the *draft*.

ArisuOngaku:
I found a todo that says the "Message" system is to get rid of in favor of Observables.

Pending build then I'll request reviewing

tobiasKaminsky:
@ArisuOngaku I am really sorry for coming back so late :S
Is this still valid, as you say that Observables might get removed?

ArisuOngaku:
> @ArisuOngaku I am really sorry for coming back so late :S
> Is this still valid, as you say that Observables might get removed?

@tobiasKaminsky I think I didn't mean that; my point is that I found a "TODO" comment somewhere that said that:
- We should get rid of the "Message" system
- We should use the "Observable" instead

I believe that this PR is still relevant unless something new came since I originally submitted that PR.

Issue #5280, opened 2020-01-22
Enhance file conflict dialog


tobiasKaminsky:
![2020-01-22-123113](https://user-images.githubusercontent.com/5836855/72890829-17220580-3d13-11ea-83bd-5a0bcd991b36.png)

For #4788 we should enhance this dialog a bit, and as a quick win:
- title "click on file you want to keep"
- show files side by side
   - show file size
   - show modification date
   - show image
- click on file keeps it
- two secondary actions on bottom:
  - cancel
  - keep both

@AndyScherzinger @jancborchardt

AndyScherzinger:
@tobiasKaminsky details for have been described by @jancborchardt here: https://github.com/nextcloud/android/pull/4788#issuecomment-563875618

so c&p:
Something like:
**Conflict during auto upload**
Pick which version to keep of `filename`

### Conflict resolution dialog
![Auto upload file conflict resolution](https://camo.githubusercontent.com/e7d270a9fd12a6eba282cab1e22ed0da3bb781b5/68747470733a2f2f707573682e61726973752e66722f7065726d616e656e742f53637265656e73686f745f313537353838323639392e706e673f31)

We need more info on both the files here to be able to decide which one to keep. See the conflict resolution dialog on the server, it gives you:
- thumbnails of both files so you can quickly see
- timestamp and file size of the files (ideally both, not sure why it currently doesn’t do the file you are uploading …)

Not saying that dialog is perfect, but thumbnails and the basic metadata go a long way already to decide whether it’s the same file, or a newer one, or a downscaled/cropped one, etc. to help inform your decision.
![file conflict server](https://user-images.githubusercontent.com/925062/70499256-5c1b1f80-1b4b-11ea-919c-67efb5b38266.png)

tobiasKaminsky:
Having a checkbox is overhead and it is not that easy to use on touch.
Let us go with clicking on it, which highlights it instead.

PR #5491, opened 2020-02-19
Enhance conflict dialog

tobiasKaminsky:
Closes https://github.com/nextcloud/android/issues/5280

- fix bug
- add ui tests
- add instrumented tests

image

### Testing 

- [x] [unit tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#unit-tests)
- [x] [instrumented tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#instrumented-tests)
- [x] [UI tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#ui-tests)


src/main/res/values/strings.xml
```diff
@@ -328,7 +328,7 @@
     <string name="instant_upload_path">/InstantUpload</string>
     <string name="auto_upload_path">/AutoUpload</string>
     <string name="conflict_title">File conflict</string>
-    <string name="conflict_message">Which files do you want to keep? If you select both versions, the local file will have a number appended to its name.</string>
+    <string name="conflict_message"></string>
```
AndyScherzinger: empty string? (for backwards compatibility?)

src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
```diff
@@ -40,49 +65,129 @@
...
         KEEP_LOCAL,
         KEEP_SERVER,
     }
 
-    private final OnConflictDecisionMadeListener listener;
-    private final boolean canKeepServer;
+    public ConflictsResolveDialog() {
+        // needed by Android
+        existingFile = null;
+        newFile = null;
```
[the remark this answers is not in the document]
AndyScherzinger: why does Android need this? (according to your comment)

src/main/res/values/strings.xml
```diff
@@ -925,4 +925,10 @@
     <string name="sync_not_enough_space_dialog_action_free_space">Free up space</string>
     <string name="sync_not_enough_space_dialog_placeholder">%1$s is %2$s, but there is only %3$s available on device.</string>
     <string name="sync_not_enough_space_dialog_title">Not enough space</string>
+    <string name="conflict_message_headline">Which files do you want to keep?</string>
+    <string name="conflict_message_description">If you select both versions, the local file will have a number appended to its name.</string>
+    <string name="conflict_new_file">New file</string>
+    <string name="conflict_already_existing_file">Already existing file</string>
```
AndyScherzinger: strings are unused resources according to lint?

src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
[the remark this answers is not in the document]
tobiasKaminsky: Was needed as default/empty constructor, when rotating, but now I use fragment manager and it is not needed.
I will remove it :+1:

src/main/res/values/strings.xml
tobiasKaminsky: Not needed, as split up.

src/main/res/values/strings.xml
tobiasKaminsky: This is a lie ;-)
I will have a look.

tobiasKaminsky: You were right with lint, but referred to the wrong conflict messages :-)
I removed the not used ones now.

AndyScherzinger: Whoops 🙈

AndyScherzinger:
@tobiasKaminsky Test keeps failing :/

src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
```diff
@@ -40,49 +65,120 @@
...
+        alertDialog.getButton(AlertDialog.BUTTON_POSITIVE).setTextColor(color);
+        alertDialog.getButton(AlertDialog.BUTTON_NEGATIVE).setTextColor(color);
+    }
+
+    @Override
+    public void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        if (savedInstanceState != null) {
+            newFile = (File) savedInstanceState.getSerializable("file");
```
ezaquarii: `static final String KEY_FILE`?

src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
```diff
@@ -99,7 +195,7 @@ public void showDialog(AppCompatActivity activity) {
     }
 
     @Override
-    public void onCancel(DialogInterface dialog) {
+    public void onCancel(@NotNull DialogInterface dialog) {
```
ezaquarii: `NotNull` -> `NonNull`.

We are using a proprietary `org.jetbrains.annotations.NotNull` annotation. We should use the one from `androidx` package, called `NonNull`.

src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
```diff
@@ -40,49 +65,120 @@
...
         KEEP_BOTH,
         KEEP_LOCAL,
         KEEP_SERVER,
     }
 
-    private final OnConflictDecisionMadeListener listener;
-    private final boolean canKeepServer;
-
-    public ConflictsResolveDialog(OnConflictDecisionMadeListener listener, boolean canKeepServer) {
+    public ConflictsResolveDialog(OnConflictDecisionMadeListener listener,
```
ezaquarii: Do we have a good reason to do it this way?

We are using non-default constructor, whoch makes default constructor unavailable; default constructor is used by `FragmentManager`. I expect it can crash if `FragmentManager` tries to restore the dialog.

How it is normally done would be:
1. pass arguments via arguments `Bundle`
2. implement listener in parent activity and use type cast to obtain reference to it

src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
```diff
@@ -40,49 +65,120 @@
...
+                                  OCFile file,
+                                  OCUpload conflictUpload, Optional<User> user) {
         this.listener = listener;
-        this.canKeepServer = canKeepServer;
+        this.existingFile = file;
+        this.newFile = new File(conflictUpload.getLocalPath());
+        this.user = user.get();
```
ezaquarii: This `Optional<T>.get()` can throw NPE.
If we really, really, really know it won't be null, let's use `getOrElse(IllegalStateException::new)`.

The main reason we introduced `Optional<User>` was to replace nullable `getAccount()` call. The trouble with `getAccount()` was that callers were liberally assuming valid return value. Since the code was then full of `if (account == null) { some ad-hoc remediation }` workarounds, this assumption was not really holding up. My suspicion was that we were experiencing many edge cases of NPEs and were trying to patch them in some simple way.

Hence the `Optional<T>` which provides more explicit and harder to ignore API.

Having said all of the above, I think that the best idea would be to pass `User` instead of `Optional<User>` and just make sure that if the user dependency is not available, we never get here in the first place.

src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
```diff
@@ -40,49 +65,120 @@
...
+    }
+
+    @Override
+    public void onStart() {
+        super.onStart();
+
+        AlertDialog alertDialog = (AlertDialog) getDialog();
+
+        if (alertDialog == null || user == null) {
```
ezaquarii: This will never be called - if `user` is null, it would blow up with NPE in the constructor.

src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
```diff
@@ -108,4 +204,137 @@ public void onCancel(DialogInterface dialog) {
     public interface OnConflictDecisionMadeListener {
         void conflictDecisionMade(Decision decision);
     }
+
+    private void setThumbnail(OCFile file, ImageView thumbnailView) {
```
ezaquarii: Cyclomatic complexity of thos method is high. I'm not sure if this is a new code or moved from somewhere else.

If we have bandwidth, we could plit those `if`'s into methods.

src/main/java/com/owncloud/android/ui/activity/ConflictsResolveActivity.java
```diff
@@ -137,12 +151,21 @@ protected void onStart() {
...
+                if (prev == null) {
+                    if (getStorageManager().fileExists(file.getRemotePath())) {
+                        ConflictsResolveDialog dialog = new ConflictsResolveDialog(this,
+                                                                                   getFile(),
+                                                                                   conflictUpload,
+                                                                                   getUser()
+                        );
+                        dialog.show(fragmentTransaction, "conflictDialog");
+                    } else {
+                        // Account was changed to a different one - just finish
```
ezaquarii: Enclosing `if(getAcount != null`) could be updated:
```java
User optionalUser = getUser();
if (!user.isPresent()) {
    finish();
    return;
}
OCFile file = getFile();
...
```
This way we can eliminate 1 level of ifology and decrease cyclomatic complexity a bit.

src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
tobiasKaminsky: Argh. Again and again…
I removed lots of those, but seems that still some are left…

tobiasKaminsky: Nope, not my fault ;-)
Method in DialogFragment uses andoidx.annotation.NonNull
but auto suggestion from AndroidStudio only suggests org.jetbrains.annotations.NotNull

image
this can be changed here, no idea why jetbrains is on top…

src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
tobiasKaminsky: Ok, will pass User and fail early, if no user is available.

src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
tobiasKaminsky: This is half c&p from ThumbnailGenerationClass.
I will try to get this get this back, so that we do not have code duplication.

src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
tobiasKaminsky: Changed as above

tobiasKaminsky:
Thanks a lot for CR @ezaquarii!

src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
tobiasKaminsky: This was already the case, so I did not changed/touched it.
But let me change it :-)

AndyScherzinger:
@tobiasKaminsky Conflict resolve tests still keep failing :(

tobiasKaminsky:
Failure is unrelated

PR #5530, opened 2020-02-27
Improve dev edition changelog generation

### LINKED ISSUE #5525: Improve dev edition changelog
I find that the changelog for the F-Droid dev edition isn't really helpful; it's dense and hard to read. I haven't looked into it closely but I'm guessing we automatically generate it - here are some suggestions for how to clean it up:

* Omit commit IDs. These don't actually mean anything to end users
* Rewrite Transifex commits to "update translations" or something
* Omit version bump commits like 1dd355d
* Don't show the library commit, instead just say "updated library" if the commit changed since last time
* Filter out commits that touch only tests
* Filter out merges probably?

I'll look into this myself when I have time but anyone should feel free to pick it up themselves if you get to it before me 👍

strugee:
We:

* No longer include commit ids
* Exclude merges, individual Dependabot commits, commits including only non-user-facing changes, or version bump commits
* Collapse all Dependabot commits into a single "Update dependencies" changelog entry
* Do the same for Transifex updates, which get turned into a single "Update translations" changelog entry
* Say if the Android Nextcloud library has changed instead of including its commit id

The list of paths to ignore is probably incomplete, but it's easier to just revisit this script over time and add to it as new contenders pop up, rather than trying to find everything now.

Fixes #5525

### Testing 

N/A; I manually tested this

### DIFF
--- a/scripts/buildDev
+++ b/scripts/buildDev
@@ -33,13 +33,33 @@ mv latest.apk ~/apks/
 /bin/ls -t ~/apks/*.apk | awk 'NR>6' | xargs rm -f
 
 lastBuildTime=$(date --date="$(git log -1 --format=%ai $(git tag | grep dev | tail -n1))" +%s)
-changelog=$(git log --after=$lastBuildTime --pretty=oneline --abbrev-commit)
+# Show only the commit subject in the changelog and filter out:
+#  * Merges
+#  * Dependabot commits
+#  * Commits touching only non-user-facing stuff like tests
+#  * Version bump commits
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@strugee could you maybe do a sign-off, see https://github.com/nextcloud/android/pull/5530/checks?check_run_id=471641788 ❤️

strugee:
I've pushed a signoff commit, but the PR took about 10 minutes to update due to an ongoing [GitHub incident](https://www.githubstatus.com/incidents/q07bfjh7jf1t)... so who knows :sweat_smile:

tobiasKaminsky:
Nice, as this is mainly for fdroid, I am good to go :+1:

strugee:
Thank you!

PR #5531, opened 2020-02-27
Update to Android Studio 3.6

AndyScherzinger:
- [x] merge after 3.11 branch off 

see $subject


AndyScherzinger:
@tobiasKaminsky compile is (for unknown reasons) completely broken while it works just fine locally in Android Studio. Haven't seen any file I missed to push :/

scripts/analysis/findbugs-results.txt
```diff
@@ -1 +1 @@
-383
+0
```
tobiasKaminsky: :see_no_evil: :see_no_evil: :see_no_evil: :see_no_evil:

tobiasKaminsky:
For me our project is completely broken with new gradle as it cannot build anything due to missing lombok…

AndyScherzinger:
Hmm, I just updated an existing checkout being compiler before, so maybe that is the issue. 😔

tobiasKaminsky:
There is a change in gradle-plugin which seems to first compile Kotlin and then Java. 
But as Lombok generates its boilerplate (getter, setter) only during Java compilation all these functions are missing during Kotlin compile.

There is a "delombok" which removed Lombok within 5min, but then we have lots of boilerplate code again.
But moving all of this to Kotlin is also not feasible.
So I would suggest
- delombok to get gradle 3.6 running
- slowly over time migrate data classes to Kotlin
  -  do this everytime we would change/add a getter/setter

This way we prevent to add new Java/boilerplate, but do not have to do the Kotlin move at once.

AndyScherzinger:
Sounds good to me!

tobiasKaminsky:
@AndyScherzinger I tend to just increase lint, as we cannot handle all the fixes right now.
We should aim to remove Java over time and this will also decrease lint again…

AndyScherzinger:
> I tend to just increase lint, as we cannot handle all the fixes right now.

I fully agree. I removed the lint-errors (so just new warnings), so we should be fine 👍

tobiasKaminsky:
I will merge this *after* 3.11 branch off, so that we have more time to test.
Although I hope that this is not needed :-)

AndyScherzinger:
@tobiasKaminsky can you please bump the scores so we can get this PR merged? ❤️

PR #5537, opened 2020-02-27
Split up share internal link

tobiasKaminsky:
Needed as syntax is different on other languages, thus it is better to have separate strings.

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -905,7 +905,8 @@
     <string name="file_detail_activity_error">Error retrieving activities for file</string>
     <string name="no_mutliple_accounts_allowed">Only one account allowed</string>
     <string name="remote_file_fetch_failed">Failed to find file!</string>
-    <string name="share_internal_link_text">Only works for users with access to this %1$s</string>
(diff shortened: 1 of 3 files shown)


src/main/res/values/strings.xml
```diff
@@ -905,7 +905,8 @@
     <string name="file_detail_activity_error">Error retrieving activities for file</string>
     <string name="no_mutliple_accounts_allowed">Only one account allowed</string>
     <string name="remote_file_fetch_failed">Failed to find file!</string>
-    <string name="share_internal_link_text">Only works for users with access to this %1$s</string>
+    <string name="share_internal_link_to_folder_text">Only works for users with access to this folder</string>
```
ezaquarii: English is not a [pro drop language](https://en.wikipedia.org/wiki/Null-subject_language#Impersonal_constructions) so the subject should be added:

I'd guess that "Sharing only works..." is what we meant here?

tobiasKaminsky: image

As I am not a native speaker, I am not sure if we need this, if the context is clear via the headline?

PR #5546, opened 2020-02-28
Migrations manager

ezaquarii:
# Migrations now

Currently there is no centrilized migration logic. We have 1 ad-hoc user model migration launched
in `MainApp.java:264` that performs some account changes, but:
1. it does not block user interactions until the migration is completed
2. it ignores errors

Although we didn't have much problems with this simple solution, it is no longer scaling given the upcoming needs in `WorkManager` (#5482) migration and upcoming refactoring work.

# Migrations manager

## Changes

This PR provides a migration manager extracted from PR #5482 with few modifications done to satisfy `androidTests` environment. Changes are for the better anyway, so those changes will stay.

1. `MigrationManager` runs and keeps track of past migrations and failures
2. `Migrations` is a dumping ground for all sort of migration code for today and the future (#5482) and supplies migraiton procedures for manager to run
3. Local `tests` are still using `Robolectric`, but `androidTest` are supplied as well so we can compare and make an informed decision about; see section below about problems discovered during implementation.
3. Added `mockito-android` for `androidTests` dependencies

## Capabilities

- we can now run migrations only once
- we know when the migration failed and this fact can be used to force application re-installation
- UI can be blocked until all migrations are applied (using the infrastructure provided in https://github.com/nextcloud/android/pull/5064, this can be as easy as adding a mixin to base activity)
- migration status flag is observable

# Issues with test environment

Things are not straightforward, unfortunately and I discovered some problems with `androidTest` environment.

- Contrary to local tests running in vanilla JVM, Mockito bytecode generator has some limitations on Android:
 - `spy()` didn't work at all for me; this could be related to  https://github.com/nhaarman/mockito-kotlin/issues/374
 - Cannot mock `invoke()` (so can't use mocked functions)
 - Cannot mock classes in Kotlin by default, only interfaces; Kotlin classes are `final` to discourage
 inheritance abuse and must be `open`ed for mocking.
- For some rason, I cannot run all tests in IDE, only clicking them 1-by-1; instrumentation terminates unexpectedly; This could be just my IDE issue so I'd like to ask other devs for confirmation.
- Cannot run `androidTests` in QA/Dev build due to incompatible user ID (somehting related to screenshots); capability to run tests in non-GPlay variant is important for contributors who use their phone and don't want to overwrite their gplay daily driver app with development build.
- Cannot use backticks to name test functions as test titles and linter enforces `camelCaseNames` which are imo less readable as test titles (this is easily solvable by `Detekt` config).

# TODO

- [x] Decide what to do when migration fails - we still don't have a valid strategy for now, so the code only supplies a status flag and carries on - this is the current behaviour
- [x] Decide if we still wan to use `androidTest` for it; in the context of mocking troubles, it's not so black-white anymore
- [x] Linter cleanup for sure
- [x] Replace `MainApp.java:264` call with a call to migration manager
- [x] Address migrations that can be re-run after failure ("optional")
- [x] `build.gradle` cleanup
- [x] Verify if user id migration can be applied only once or needs to be re-applied after new account creation

@AndyScherzinger @tobiasKaminsky review pls


ezaquarii:
@tobiasKaminsky I looks like screenshots cannot be uploaded and the build fails because of it.

tobiasKaminsky:
> [ ] Decide what to do when migration fails - we still don't have a valid strategy for now, so the code only supplies a status flag and carries on - this is the current behaviour

When a migration fails it is a serious problem. We then should redirect to github for bug reporting , stop the app and recommend to reinstall it (after all only some settings are lost, but no data).

I know this is a rough step, but if we would allow to continue the app, it might lead to hard reproducable bugs.
(and every migration should be so bullet-proof that this does not really happen).

tobiasKaminsky:
Regarding tests:
If Robolectric is not working, let us use emulator / instrumented tests.

ezaquarii:
It's working. We have issues with instrumented tests and device tests.

ezaquarii:
@tobiasKaminsky I added flag to mark migrations mandatory or not. Optional migrations will not cause application failure.

The use case is the exisitng account migration procedure that can fail, but will try again on subsequent run. This change will preserve existing app behaviour which has been deployed in the field for long time and seems to be working.

ezaquarii:
> When a migration fails it is a serious problem. We then should redirect to github for bug reporting , stop the app and recommend to reinstall it (after all only some settings are lost, but no data).

As we currently do not have any hard-failing migrations (existing one is restarted and ignores failures), shall we address this separately, when such issue arises?

My #5482 PR will add a migration that is "bulletproof", so no need to over-implement at this stage as there will be direct need for it now. On the other hand, having a complete solution is beneficial as well. Let me know what strategy wil work for us here.

ezaquarii:

1. removed robolectric
2. migrations are now called in `MainApp`

I guess it's shippable stage now let's do a final review with that intention.

ezaquarii:
@tobiasKaminsky User id migration returns error when any account migration fails. Previous implementation was returning succeess if any account succeeded. This was ok assuming migration being re-run on every start.

From now on, once migration succeeds for all accounts, we mark it as applied and no longer re-try.

This has an important implication. Could you tell me if the user id is set properly on account creation? If the purpose of this migration was to fix legacy accounts only - this implementation is correct.

But in case it was intentedd to continuously patch new accounts - we need a to fix authentication flow as well.

tobiasKaminsky:
> My #5482 PR will add a migration that is "bulletproof", so no need to over-implement at this stage as there will be no need for it now. On the other hand, having a complete solution up front has it's benefits as well. Let me know what strategy will be best here.

Let us tackle failing migrations, once this might occur.
Keep it simple right now :-)

tobiasKaminsky:
> This has an important implication. Could you tell me if the user id is set properly on account creation? If the purpose of this migration was to fix legacy accounts only - this implementation is correct.

Yes this is only for legacy accounts, as we save this for every new account:
https://github.com/nextcloud/android/blob/f8ff0e106af11708c7e1cd4974c7b420800f6d8d/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java#L1777

src/main/java/com/nextcloud/client/migrations/Migrations.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * List of migration steps. Those steps will be loaded and run by [MigrationsManager]
+     */
+    val steps: List<Step> = listOf(
+        Step(0, "migrate user id", Runnable { migrateUserId() }, false)
+    ).sortedBy { it.id }
+
+    fun migrateUserId() {
+        val allAccuntsHaveUserId = userAccountManager.migrateUserId()
```
tobiasKaminsky: typo: account

ezaquarii: fixed the typo

Issue #5512, opened 2020-02-24
dossier personnalisé


ArthurLamiral:
### Steps to reproduce
1. Auto upload
2. Dossier personnalisé
3. Dossier source

### Expected behaviour
- Tell us what should happen
Sélection du dossier source
### Actual behaviour
- Tell us what happens
Aucune sélection mémorisée
### Environment data
Android version:
9
Device model: 
Ulefone Armor x5
Stock or customized system:

Nextcloud app version:
3.10
Nextcloud server version:
18.1
#### Web server error log
```
Insert your webserver log here
```

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```
**NOTE:** Be super sure to remove sensitive data like passwords, note that everybody can look here! You can use the Issue Template application to prefill some of the required information: https://apps.nextcloud.com/apps/issuetemplate

AndyScherzinger:
Hi @ArthurLamiral,

canb you post a screenshot, do you see any error message?

ArthurLamiral:
Bonjour,

ci-jointe une capture écran de la page qui refuse d'enregistrer la configuration source et destination des dossiers à sauvegarder.

Le serveur Nextcloud utilisé fonctionne très bien avec un PC et synchronise plusieurs répertoires du PC.

Le problème est uniquement avec Android.

Cordialement.

Arthur
![_20200224_140414](https://user-images.githubusercontent.com/61087128/75155888-41e7eb00-5711-11ea-9fba-4f0d6122b90f.png)

AndyScherzinger:
You might have to click several times on the item (unfortunately).

This is an issue -don't know why the action isn't triggered continuously all the time :( just yet

AndyScherzinger:
cc @tobiasKaminsky @ezaquarii @ArisuOngaku for input ❤️

ezaquarii:
I'm not sure I follow:

> You might have to click several times on the item (unfortunately).

@AndyScherzinger do you know what the problem is?

AndyScherzinger:
Unfortunately no, it seems that the click event on the text works sometimes while sometimes it just doesn't (for the folder items) while it seems to always work for the line items with the check boxes. So I am a bit confused how that could happen...

ArisuOngaku:
[...]

PR #5593, opened 2020-03-06
Synced folder dialog: fix summary click event intercepting parent

ArisuOngaku:
From #5512 

When clicking on the remote or local folder selection row, it works except if you click on the summary.

This is because the summary has a click listener to expand on click (i.e. it contains a very long path that was cut out).

### What this PR does

~~Behavior change: to expand the summary, long click on the row (anywhere). Simple clicking on the summary now correctly opens the folder selection dialog.~~

~~[image](https://camo.githubusercontent.com/7ff070f0a196cff6c36544c3677a4f8daae7ea10/68747470733a2f2f707573682e61726973752e66722f73637265656e5f64656d6f2e706e67)~~

Update:

- Remove the expand on click on the path summaries
- Remove the folder icon on the right of the local and remote folder settings
- Display the local folder path (under "Local folder") with a user friendly nickname

Examples for the new local folder display name :
- `/storage/emulated/0/DCIM -> Internal Storage / Camera`
- `/storage/emulated/0/non/standard/folder -> Internal Storage /non/standard/folder`
- `/storage/SD_CARD/Movies -> SD_CARD / Movies`

### Testing

- [ ] `FileStorageUtils#pathToUserFriendlyDisplay`
- [ ] `FileStorageUtils.StandardDirectory#fromPath`


tobiasKaminsky:
Thank you for your PR and a  suggestion how to make this better.
I am not a big fan of long press, as this is UI element that is hidden for user…

Not sure if this works, but I would only use the click listener if the text needs to be expanded, otherwise it opens up the path chooser.

What do the other think?

jancborchardt:
> I am not a big fan of long press, as this is UI element that is hidden for user…

Yeah, agree with @tobiasKaminsky here. :) Each row here should have only 1 tap handler, and tapping anywhere on that row should perform that action. Could you clarify why a second different action is needed on this element?

What I find a bit confusing is the folder icon on the right. It’s a bit unnecessary and also looks very similar to an unfilled checkbox. If we cut it, more of a path can also be shown?

AndyScherzinger:
> Could you clarify why a second different action is needed on this element?

* action 1: expand (too) long folder path
* action 2: launch folder chooser

> What I find a bit confusing is the folder icon on the right. It’s a bit unnecessary and also looks very similar to an unfilled checkbox. If we cut it, more of a path can also be shown?

Yes, that would be possible

jancborchardt:
> * action 1: expand (too) long folder path
> 
> * action 2: launch folder chooser

Expanding a too long folder path seems to be unnecessary? As:
- Launching the folder chooser does land you in the picked folder, effectively enabling you to view the full path via the dropdown in the header
- The path already shows in at least 3 rows for me?
- Currently we show nerdy info like "/storage/emulated/0/" and the like in the path – instead it should just be shortened to "Internal storage" or "SD card", also cutting down on the space used.

AndyScherzinger:
Agree with the first point you made @jancborchardt 

> Currently we show nerdy info like "/storage/emulated/0/" and the like in the path – instead it should just be shortened to "Internal storage" or "SD card", also cutting down on the space used.

That is rather tricky, since probably _all_ auto-upload entries would then show the same value (as in source/main storage point)

jancborchardt:
> That is rather tricky, since probably _all_ auto-upload entries would then show the same value (as in source/main storage point)

Right – the first 2 points are meant to illustrate why a second action is not necessary, the last point is also more of an enhancement proposal. :) If by showing the same value you mean that this issue would arise for phones which only have the main storage: Simply don’t show that first part at all if there is only one storage. :)

AndyScherzinger:
@jancborchardt yes, perfectly fine 👍

ArisuOngaku:
@jancborchardt So if I understand correctly, we should remove the "expand/show full path" action? (and thus keep only one single main action)

AndyScherzinger:
@ArisuOngaku yes, to sum things up:

* removal of the expand action
* removal of the folder icons/layout-part to have more space for the path text

ArisuOngaku:
@AndyScherzinger @jancborchardt so I did that + the nerdy path removal, please see the PR's first message new screenshot

I will later add one last commit to add some unit tests.

Also I don't really get what's the point of this "Avoid instantiating new objects inside loops" (code PR quality review).

AndyScherzinger:
> Also I don't really get what's the point of this "Avoid instantiating new objects inside loops" (code PR quality review).

Please feel free to ignore this one :) This check does have quite some false-positives. The idea is that within a loop objects should be reused which works in some cases, while it doesn't for others. 

@tobiasKaminsky what do you think? In most cases the mentioned code check is a false positive, so we might rather catch this in code reviews than having to ignore the detected issues in 98% of the cases...

Issue #4779, opened 2019-11-01
Release 3.11.0 - Roadmap


AndyScherzinger:
Planned RC1: 02.03.20

**Create stable-x.y branch & use it for release**
- [x] create branch ```stable-3.11``` (no ".x" anymore due to backport bot)
	- [x] make it protected
	- [x] set up Codacy, Drone as required status
	- [x] enable branch to be analyzed on codacy
- [x] change on App
	- [x] ```drone.yml```
	- [x] ```getBranchName.sh```
	- [x] ```analysis-wrapper.sh```
	- [x] add to backport transifex stable branch, use latest two stable branches
	- [x] check that transifex bot still works
- [x] change on Lib
	- [x] ```drone.yml```
	- [x] ```getBranchName.sh```
	- [x] ```analysis-wrapper.sh```

**Bump**
- [x] release new library version with rc1 tag & use this is in app
- [x] release new SSO if needed
- [x] Bump version name and version code in ```build.gradle``` files
- [x] Update ```CHANGELOG.md``` adding new features
- [x] Update version on master to next major version
- [x] ~~check if oldest Android version is EOL by Google and user usage <= 0.5% and drop it if so (**--> drop Android 4.1**) #5594~~
- [x] ~~With `minSDK=17` #4480 should be reactivated to get that depedency updated (which requires Android `minSDK=17`)~~
- [ ] check if code is outdated due to last supported server version ```@NextcloudServer``` annotation
- [ ] bump version in MainApp
	- [ ] ```OUTDATED_SERVER_VERSION``` according https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule (e.g. NC12)
	- [ ] ```MINIMUM_SUPPORTED_SERVER_VERSION```: latest end of life + 1 year (e.g. NC13 end of life 2019-02-28 + 1 year = 2020-02-28)
	- [ ] add these versions to changelog
- [ ] remove support for server older ```MINIMUM_SUPPORTED_SERVER_VERSION```
- [ ] do the same on library
- [x] add what's new slides

**Test app with all supported Android version**
- access all activities
- auto upload
- upload / download files/folders
- use external sdcard
- what's new slides correctly shown on first start

All tests on
  - [ ] Android 4
  - [ ] Android 5
  - [ ] Android 6
  - [ ] Android 7
  - [ ] Android 8
  - [ ] Android 9
  - [ ] Android 10

**RC1**
- [x] Verify that the new version number is shown in settings
[...]

PR #5594, opened 2020-03-06
change minimum supported Android version to 4.2

AndyScherzinger:
follow-up to #4779

Bumping min. Android version from 4.1 to 4.2
This should have had happened for v3.11.0 which is already in RC phase, so making sure it happens for 3.12 at least.


src/main/java/com/owncloud/android/ui/dialog/SyncedFolderPreferencesDialogFragment.java
```diff
@@ -180,15 +180,9 @@ private void setupDialogElements(View view) {
...
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR1) {
-            view.findViewById(R.id.setting_instant_upload_on_charging_container).setVisibility(View.GONE);
-        } else {
-            view.findViewById(R.id.setting_instant_upload_on_charging_container).setVisibility(View.VISIBLE);
-
-            mUploadOnChargingCheckbox = view.findViewById(
-                    R.id.setting_instant_upload_on_charging_checkbox);
-            ThemeUtils.tintCheckbox(mUploadOnChargingCheckbox, accentColor);
-        }
+        view.findViewById(R.id.setting_instant_upload_on_charging_container).setVisibility(View.VISIBLE);
```
tobiasKaminsky: This is now always shown, or? So we can change it in xml and remove this line.

AndyScherzinger: Nice catch, it is already like that in the xml 👍 
Removed the line!

PR #5607, opened 2020-03-09
Use view bindings: https://developer.android.com/topic/libraries/view-binding

tobiasKaminsky:
This has several advantages:
- null safety
- type safety
- faster

Once this is in, we can slowly migrate the rest.


AndyScherzinger:
LGTM, lint score should be raised, unused layout resource seems to be a false positive

PR #5624, opened 2020-03-11
Gradle wrapper validation

tobiasKaminsky:
See: https://github.com/gradle/wrapper-validation-action

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -1,0 +1,10 @@
+name: "Validate Gradle Wrapper"
+on: [push, pull_request]
+
+jobs:
+    validation:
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
cc @hannesa2 created/merged on the Nc repo (for obvious security reasons as stated by you). Thanks for the inspiration ❤️

hannesa2:
Gerne !
I see you are _much_ faster then the others 👍

hannesa2:
Is the NC app compatible with OC ?

tobiasKaminsky:
> Is the NC app compatible with OC ?

I have not tried it the last half year, but judging from bug/gplay reports it seems to still work.
(we do not really aim for it, but as many things are still the same, it should work).

PR #5738, opened 2020-03-25
Fix #5736 Readd enableAccountHandling for show slider

### LINKED ISSUE #5736: 3.11 RC4 : On first run, the slider is bypass
Hello,

The first time you open the application, you expect to get this:
Screenshot_20200324-185617_Nextcloud

But it comes down to this:
Screenshot_20200324-185624_Nextcloud

It used to happen only if the application was paused.

JorisBodin:
Hello,

Quick fix for https://github.com/nextcloud/android/issues/5736
I put enableAccountHandling back on, so I can see the slider from the beginning again. This is not a permanent fix. Because it doesn't fix when the application is paused.
Maybe @ezaquarii has imagined a good solution to fix it permanently?


tobiasKaminsky:
:+1: 
Many thanks for this quick fix.
Will be backported for tomorrows 3.11 :-)

Issue #5353, opened 2020-01-30
Enhance Dark Mode


tobiasKaminsky:
@jancborchardt and me went over dark mode and found some things to do

- [x] #5739 heading, blue background: text should stay white
- [x] #5739 triangle in drawer -> white
- [x] #5739 share icons: more users -> white
- [x] #5739 file detail sharing: above act/sharing -> white line should be same like receive external files line
- [x] activity/notifications: 
    - [x] invert all except "file_created" and "file_deleted" 
    - [x] remove background circle
    - [x] remove white border from shared_via_link
- [x] #5739 notifications: button: 
    - [x] #5739 light: background (light gray), no border
    - [x] #5739 dark: background: ligth gray, no border
- [x] #5739 + button: white color
- [x] long press: 
    - [x] weird header, status bar same dark gray
    - [x] on selected item: subline lighter
    - [x] highlight: darker
- [x] #5739 In the file list, the favorite icon needs a dark outline instead of white. Also the "Person" icon on the right should be light (very barely visible now): 
  ![Dark mode user share icon in file list and favorite icon](https://user-images.githubusercontent.com/925062/77441538-d49ba880-6de9-11ea-98b9-664eab98d9bd.png)
- [x] #5739 The triangle for account switching in the sidebar should be light:
  ![Dark mode triangle in sidebar](https://user-images.githubusercontent.com/925062/77441539-d5343f00-6de9-11ea-9885-543afd2d8420.png)
- [x] #5739 The lines in the "Add" menu should be dark as well:
  ![Dark mode add menu lines should be dark as well](https://user-images.githubusercontent.com/925062/77441529-d2d1e500-6de9-11ea-8dfe-46f5fd80ad7d.png)
- [x] #5739 The refresh spinner could be dark as well?
  ![Dark mode refresh spinner should be dark too](https://user-images.githubusercontent.com/925062/77441532-d4031200-6de9-11ea-9ca7-78a37c6982cc.png)
- [x] ~~Toasts could be dark as well:~~
  ![Dark mode toasts should also be dark](https://user-images.githubusercontent.com/925062/77441540-d5343f00-6de9-11ea-97ff-338f8c338e4d.png)
- [x] The header color in the app switcher looks off – this is not themed:
  ![Dark mode header color looks off](https://user-images.githubusercontent.com/925062/77441535-d4031200-6de9-11ea-98fa-7d555ed01424.png)
@AndyScherzinger @dan0xii maybe you have time and want to help out? :heart:

[...]

PR #5739, opened 2020-03-25
Dark Mode Improvements II

dan0xii:
* Fix list divider in upload file dialog
* The drawer account chooser toggle reacts to dark server themes
* ic_people in the file list changes depending on device dark mode
* file detail sharing: above act/sharing -> white line should be same like receive external files line 
* In the file list, the favorite icon needs a dark outline instead of white
* fix theming of swipe refresh layout
* fix theming of text/icons in toolbar
* grey background for secondary action button in in-app display of notifications)

relates to #5353.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -288,6 +288,9 @@ public abstract class DrawerActivity extends ToolbarActivity
             }
         };
 
+        // proper coloring of drawer menu icon
+        mDrawerToggle.getDrawerArrowDrawable().setColor(ThemeUtils.toolbarTextColor(this));
+
         // Set the drawer toggle as the DrawerListener
         mDrawerLayout.addDrawerListener(mDrawerToggle);
         mDrawerToggle.setDrawerIndicatorEnabled(true);
--- a/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -302,7 +302,7 @@ public abstract class DrawerActivity extends ToolbarActivity
         mAccountEndAccountAvatar = (ImageView) findNavigationViewChildById(R.id.drawer_account_end);
 
         mAccountChooserToggle = (ImageView) findNavigationViewChildById(R.id.drawer_account_chooser_toggle);
-        mAccountChooserToggle.setColorFilter(ThemeUtils.fontColor(this, true));
+        mAccountChooserToggle.setColorFilter(ThemeUtils.fontColor(this, !ThemeUtils.darkTheme(getApplicationContext())));
 
         if (getResources().getBoolean(R.bool.allow_profile_click)) {
             mAccountChooserToggle.setImageResource(R.drawable.ic_down);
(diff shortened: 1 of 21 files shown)


src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -302,7 +302,7 @@ private void setupDrawerHeader() {
         mAccountEndAccountAvatar = (ImageView) findNavigationViewChildById(R.id.drawer_account_end);
 
         mAccountChooserToggle = (ImageView) findNavigationViewChildById(R.id.drawer_account_chooser_toggle);
-        mAccountChooserToggle.setColorFilter(ThemeUtils.fontColor(this, true));
+        mAccountChooserToggle.setColorFilter(ThemeUtils.fontColor(this, !ThemeUtils.darkTheme(getApplicationContext())));
```
AndyScherzinger: if I understand it correctly then we simply don't need to tint the ChooserToggle (as in always-white)

src/main/res/values-night/colors.xml
```diff
@@ -22,6 +22,7 @@
 
     <color name="text_color">#ffffff</color>
     <color name="text_color_inverse">#000000</color>
+    <color name="list_divider_background">#222</color>
```
AndyScherzinger: let's not use short-codes but rather `#222222`

jancborchardt:
Goood stuff @dan0xii! :) Could you post screenshots so I can review?

tobiasKaminsky:
> Goood stuff @dan0xii! :) Could you post screenshots so I can review?

Better is to try the APK in https://github.com/nextcloud/android/pull/5739#issuecomment-603760294
You can install it paralell, but this way you can test dark/light/follow system and not only judge from screenshots :-)

src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
dan0xii: I can see you've reverted this in https://github.com/nextcloud/android/pull/5739/commits/92c56acbd02cf0fd25ded14dadaf6750a64b2649 but for info, I had it there because in my testing the toggle was always black. I presume that must have been operator error on my behalf.

src/main/res/values-night/colors.xml
dan0xii: Resolved in https://github.com/nextcloud/android/pull/5739/commits/04d5bb64de6a13824442b69b5eac369b101a0e52.

AndyScherzinger:
@dan0xii I also started pushing fixes to the branch :)

src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
AndyScherzinger: Hmm, then let me have another look. The toogle should always be white, so if that isn't working yet than I'll take care of it

AndyScherzinger: Just checked and `ic_down` is `#FFFFFF` and this was the only code that tinted the icon so we should be fine now with simply having the tinting removed :)

AndyScherzinger:
@dan0xii @jancborchardt @tobiasKaminsky to get this PR in, shouldn't we just bring back the grey circular background on the notification icons in the list view since inverting them seems to be not that easy (at least we haven't figured it out just yet), thus the PR is stalling and given that @JorisBodin and @Shagequi are working on the redesign also fixing dark/theme issues it be great to have it in so we prevent (larger) conflicts.

AndyScherzinger:
@dan0xii I rebased the branch and reverted the background change for activity items (background can only be removed when inverting of thew icons is in place which is missing atm). This way the PR can be merged to master and shipped with 3.12. Also this way the fixes will be available to master and to @Shagequi and @JorisBodin working on the UI revamp 🚀 Godspeed folks!

Will merge, when CI is done.

PR #5761, opened 2020-03-28
Migrate FileDownloader server to User model

ezaquarii:

### DIFF
--- a/src/main/java/com/owncloud/android/services/OperationsService.java
+++ b/src/main/java/com/owncloud/android/services/OperationsService.java
@@ -37,6 +37,9 @@ import android.os.Process;
 import android.text.TextUtils;
 import android.util.Pair;
 
+import com.nextcloud.client.account.User;
+import com.nextcloud.client.account.UserAccountManager;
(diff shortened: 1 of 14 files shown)


src/main/java/com/owncloud/android/services/OperationsService.java
```diff
@@ -698,6 +706,20 @@ private void nextOperation() {
...
+    /**
+     * This is a temporary compatibility helper to convert legacy {@link Account} instance
+     * to new {@link User} model.
+     *
+     * @param account Account instance
+     * @return User model that corresponds to Account
+     * @throws RuntimeException if account cannot be converted
+     */
+    @NonNull
+    private User toUser(Account account) {
```
ezaquarii: This helper `OperationsService` allows us to avoid immediate refactoring of `OperationsService`, which is rather invasive due to extensive usage of it's API across the app.

If account is valid, conversion to user will not fail.

ezaquarii:
Resolved merge conflic.

PR #5825, opened 2020-04-07
send push to deck app, if installed

tobiasKaminsky:

### Testing 


src/main/java/com/owncloud/android/jobs/NotificationJob.java
```diff
@@ -161,15 +161,31 @@ private void sendNotification(Notification notification, User user) {
         RichObject file = notification.subjectRichParameters.get("file");
 
         Intent intent;
-        if (file == null) {
-            intent = new Intent(context, NotificationsActivity.class);
+        if ("deck".equalsIgnoreCase(notification.app)) {
```
AndyScherzinger, line 164: should be a constant

tobiasKaminsky: True, currently this is a PoC to see if this works out with Deck…
After that I'll enhance code :+1:

src/main/java/com/owncloud/android/jobs/NotificationJob.java
```diff
@@ -243,6 +251,36 @@ private void sendNotification(Notification notification, User user) {
...
+        final String baseDeckApplicationId = "it.niedermann.nextcloud.deck";
+        final String activityToStart = "it.niedermann.nextcloud.deck.ui.PushNotificationActivity";
+        final String[] flavors = new String[]{"", ".play", ".dev"};
+        for (String flavor : flavors) {
+            final Intent intent = new Intent().setClassName(baseDeckApplicationId + flavor, activityToStart);
+            if (context.getPackageManager().resolveActivity(
+                intent, 0) != null) {
+                Log.i(TAG, "Found deck app flavor \"" + flavor + "\"");
+                return intent
+                    .putExtra("account", user.getAccountName())
```
tobiasKaminsky, line 268: I am not sure if we want do make this all constants, @AndyScherzinger ?

AndyScherzinger: Well, if it is kind of a key / fixed value than I'd say it should be a constant

ezaquarii: Entire Deck interaction should be extracted into it's own module with a well defined API.

stefan-niedermann: @ezaquarii what exactly do you propose? We have limited data that make actually sense to pass to a 3rd-party client. Basically the data consists of a `Notification` and a `User`-object. Some ideas:

- Even if Deck is currently the only 3rd-party-app subscribing to push notifications this way, i believe, that other apps will follow in the future
- Define the API for each 3rd-party client the same by passing "all" the data that we have
  - This means *one* clear contract
  - Letting the 3rd-party-client define the contract will result in chaos, and the main app will be the pain point to update.
  - The main app should be the provider, the consumers should have to follow
- We could name each argument like the correspond variable name (that's how it currently is)
- Instead of reading and passing each property on its own, pass the whole objects (`User` is a `Parcelable` and `Notification` should be a `Serializable`). For users of the `android-library` this would make stuff even easier, because they can reuse the classes (Deck does *not* use `android-library`).
- Some kind of `swagger`/`open-api`-contract, but that seems too blown up for me

stefan-niedermann: About an own feature module: This makes sense, but i am not experienced enough with android / gradle modules to make such an integration. I tried to swap out the Deck related stuff into an own method, but i would be grateful if you could swap it out into an own module like you proposed @ezaquarii :slightly_smiling_face:

ezaquarii: @stefan-niedermann Let's not overengineer it. This is simply not the place to have some 3rd party plumbing, especially that this `Job` object will be refactored in few days (by me), and I'm vitally interested in not dealing with any extra complexities here. :)

Let's just create a package (`com.nextcloud.client.integrations` or something like this) and put a class there with all Deck-specific plumbing. Let this component export a business-oriented interface that anybody who doesn't know anything about Deck can use:

```kotlin
interface Deck {
    /**
      * Tell us what the it does and what we can do with it
      */ 
    fun frobnicateDeck()
}
```

Since I have no clue what Deck is, please treat this example as a figure of speech.
It just illustrates the level of complexity I'd like to see when dealing with your app.

Maybe frobnication requires an intent with 137 magic keys and values carefully chosen during 7th full moon night, or maybe it hides a Rube Goldberg machine. We don't care here - it's sealed and maintained by somebody who knows better.

I'd start your work from the interface design. Come up with something and we'll guide you from there.

ezaquarii: > I tried to swap out the Deck related stuff into an own method,

That's a good start. Put in into a class and that's the level of sophistication that will work for us here. :)

stefan-niedermann: I'll try it, thanks :)

src/main/java/com/nextcloud/client/integration/NotificationHandler.java
```diff
@@ -0,0 +1,21 @@
...
+
+import androidx.annotation.NonNull;
+
+public interface NotificationHandler {
+
+    @NonNull
+    Intent handleNotification(@NonNull final Notification notification,
+                              @NonNull final User user) throws AppNotInstalledException, AppCannotHandelNotificationException;
+
+    class AppNotInstalledException extends Exception {
```
tobiasKaminsky, line 16: Please have this as a separate class and file.

src/main/java/com/nextcloud/client/integration/NotificationHandler.java
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    class AppCannotHandelNotificationException extends Exception {
```
tobiasKaminsky, line 19: Same here :-)
(and typo in handle)

src/main/java/com/nextcloud/client/integration/NotificationHandler.java
```diff
@@ -0,0 +1,21 @@
+package com.nextcloud.client.integration;
```
tobiasKaminsky, line 1: Can you add GPL3+ header for all new files?

src/main/java/com/owncloud/android/jobs/NotificationJob.java
stefan-niedermann: This is also fixed in the meantime, it's a constant in `DeckNotificationHandler`.

src/main/java/com/nextcloud/client/integration/deck/DeckNotificationHandler.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @NonNull
+    @Override
+    public Intent handleNotification(@NonNull Notification notification, @NonNull User user) throws AppNotInstalledException, AppCannotHandleNotificationException {
+        if (!APP_NAME.equalsIgnoreCase(notification.app)) {
+            throw new AppCannotHandleNotificationException();
+        }
+        for (String flavor : DECK_APP_ID_FLAVOR_SUFFIXES) {
+            final Intent intent = new Intent().setClassName(DECK_APP_ID_BASE + flavor, DECK_ACTIVITY_TO_START);
```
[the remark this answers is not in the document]
stefan-niedermann: I think this is inevitable. We have a fixed number of entries in this array and i do not expect to grow it. @tobiasKaminsky is this okay or do you have a better idea to cover all flavors?

src/main/java/com/nextcloud/client/integration/NotificationHandler.java
```diff
@@ -0,0 +1,15 @@
...
+
+import com.nextcloud.client.account.User;
+import com.owncloud.android.lib.resources.notifications.models.Notification;
+
+import androidx.annotation.NonNull;
+
+public interface NotificationHandler {
+
+    @NonNull
+    Intent handleNotification(@NonNull final Notification notification,
```
ezaquarii, line 13: Ok, so this is agood start, but we'd like to further massage it into shape.

First, "handle" means nothing, so this interface is not better than a `Runnable`.

It says that it "handles" something, but it the code is a factory... it doesn't make sense for me.

Lets start with the following questions:
1. what is the business goal of this module?
2. let's put a javadoc that explains what we use it for

Once we have those 2 established, we can think about a better structure.

src/main/java/com/nextcloud/client/integration/deck/DeckNotificationHandler.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private final PackageManager packageManager;
+
+    public DeckNotificationHandler(@NonNull Context context) {
+        this.packageManager = context.getPackageManager();
+    }
+
+    @NonNull
+    @Override
+    public Intent handleNotification(@NonNull Notification notification, @NonNull User user) throws AppNotInstalledException, AppCannotHandleNotificationException {
```
ezaquarii, line 53: This method throws 2 checked exception, making it cumbersome to use. If you really need to handle excepton, please use 1 type.

But honestly I'd not throw anything. People don't know how to handle them. I'd consider using `com.nextcloud.java.util.Optional` instead.

src/main/java/com/owncloud/android/jobs/NotificationJob.java
```diff
@@ -160,16 +164,21 @@ private void sendNotification(Notification notification, User user) {
...
 
+        final NotificationHandler deckNotificationHandler = new DeckNotificationHandler(context);
         Intent intent;
-        if (file == null) {
-            intent = new Intent(context, NotificationsActivity.class);
-        } else {
-            intent = new Intent(context, FileDisplayActivity.class);
-            intent.setAction(Intent.ACTION_VIEW);
-            intent.putExtra(FileDisplayActivity.KEY_FILE_ID, file.id);
+        try {
```
ezaquarii, line 169: We still have a bunch of deck related logic in a wrong place. Also, I have no idea what it does. 

Pleae bear in mind that this `Job` will be deleted.

src/main/java/com/nextcloud/client/integration/deck/DeckNotificationHandler.java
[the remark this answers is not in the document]
ezaquarii: 1. You don't need to allocate new intent every time - just set a class name and re-query package manager.
2. Package resolution and intent creation should be split - not coalesced into a loop

src/main/java/com/nextcloud/client/integration/deck/DeckNotificationHandler.java
```diff
@@ -0,0 +1,77 @@
...
+                    .putExtra("subject", notification.getSubject())
+                    .putExtra("subjectRich", notification.getSubjectRich())
+                    .putExtra("message", notification.getMessage())
+                    .putExtra("messageRich", notification.getMessageRich())
+                    .putExtra("user", notification.getUser())
+                    .putExtra("nid", notification.getNotificationId())
+                    .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+            }
+        }
+        Log.v(TAG, "Couldn't find any installed deck app.");
```
ezaquarii, line 74: Please remove this - we don't need to spam logs.

src/main/java/com/nextcloud/client/integration/deck/DeckNotificationHandler.java
ezaquarii: @stefan-niedermann 
Havind said that - I'm still not convinced that this `handleNotification` is a right abstraction for the business scenario you are implementing. Let's clarify that before we move forward.

src/main/java/com/nextcloud/client/integration/deck/DeckNotificationHandler.java
stefan-niedermann: Fixed

src/main/java/com/nextcloud/client/integration/deck/DeckNotificationHandler.java
stefan-niedermann: Optional is used now ✔️

src/main/java/com/owncloud/android/jobs/NotificationJob.java
stefan-niedermann: Think we solved it at our phone call :)

stefan-niedermann:
So, i think, we're done here basically. Not sure though if @ezaquarii wants to do the Dependency injection stuff in this Pull Request or if we want to merge first?

Here is a complete video demonstrating the workflow:

Workflow

And for documentation purposes the architectural image:

architecture

Thank you all for the help so far! :slightly_smiling_face:

src/main/java/com/nextcloud/client/integration/NotificationHandler.java
stefan-niedermann: I think we found a proper name: `DeckApi` with a method called `createForwardToDeckActionIntent`. I also added some java doc in commit https://github.com/nextcloud/android/pull/5825/commits/6a2536f81b2b36dad32ddbd143a8077c2383c27e

szaimen:
@mario If this is possible with the Deck app, couldn't you do the same with the talk app, could you?

As a result theoratically you should also be able to also use the singlesignon-library, wouldn't you?

mario:
Files app cannot be a requirement for the Talk app I am afraid.

src/main/java/com/owncloud/android/jobs/NotificationJob.java
ezaquarii: LGTM!

ezaquarii:
> Files app cannot be a requirement for the Talk app I am afraid.

We could extract the Deck API into a library and use it there as well.

ezaquarii:
> So, i think, we're done here basically. Not sure though if @ezaquarii wants to do the Dependency injection stuff in this Pull Request or if we want to merge first?

I can do a follow-up PR, np problem.

mario:
> > Files app cannot be a requirement for the Talk app I am afraid.
> 
> @mario Just food for thoughts: we could extract the Deck API into a library and use it there as well.

What the guy wants (I think) is to use Single Sign-On for Talk which won't work because the business case of having JUST Talk exists.

That being said, we *could* support both ways (single sign-on and app-specific accounts) but that just won't work due to workload that I'm facing already.

szaimen:
> What the guy wants (I think) is to use Single Sign-On for Talk.

Yes 👍

> That being said, we could support both ways (single sign-on and app-specific accounts) but that just won't work due to workload that I'm facing already.

Didn't know that this is the case. And of course I respect the decision to only support app-specific accounts; I just thought there would be no blocker anymore since you can obviously pass notifications over to other apps now...

mario:
> Didn't know that this is the case. And of course I respect the decision to only support app-specific accounts; I just thought there would be no blocker anymore since you can obviously pass notifications over to other apps now...

We always knew that was an option. It's just that somebody had to do it :P

mario:
> > What the guy wants (I think) is to use Single Sign-On for Talk.
> 
> Yes 👍
> 
> > That being said, we could support both ways (single sign-on and app-specific accounts) but that just won't work due to workload that I'm facing already.
> 
> Didn't know that this is the case. And of course I respect the decision to only support app-specific accounts; I just thought there would be no blocker anymore since you can obviously pass notifications over to other apps now...

Also, ++ for your understanding, it's rare these days.

ezaquarii:
On Thu, Apr 16, 2020 at 10:19:54AM -0700, Mario Đanić wrote:
> What the guy wants (I think) is to use Single Sign-On for Talk which won't work because the business case of having JUST Talk exists.

Not sure - I thought it's about forwarding push notification to Deck app... Am I
confusing some concepts here?

stefan-niedermann:
> Just food for thoughts: we could extract the Deck API into a library and use it there as well.

Not a bad idea. This way i could use the library in the Deck app too and we would have safe keys for the argument bundles. But it has some drawbacks, e.g. we have to keep the lib up to date both and we would generate more maintenance efforts.

All in all i think it's not worth the effort for now. I don't expect the Api to change again.

Feel free to merge @tobiasKaminsky or @ezaquarii if you are happy with the current state and let me know in which version this will be released :slightly_smiling_face:  (I won't dare to press the button myself)

>> What the guy wants (I think) is to use Single Sign-On for Talk which won't work because the business case of having JUST Talk exists.

> Not sure - I thought it's about forwarding push notification to Deck app... Am I
confusing some concepts here?

I am also confused. I thought what @szaimen meant was, that in an environment where a user has *only* Talk but *no* Files app installed, that the Talk app should also be able to forward notifications to the Deck app. But since the Deck app definitely needs the Files app as a requirement, it is obvious that i was wrong with this assumption :smile: But seems like @mario was able to understand the request better and already answered it. :+1:

tobiasKaminsky:
Having a deck lib is imho a bit overengineering, as we only pass push infos to deck.
This will maybe later also valid for other 3rd party apps, which are using SSO.

Having Talk supporting SSO would of course be nice, but they main big point for SSO is for devs to not have the whole network stack in their apps.
And as Mario said, this will be always needed in Talk, as it must be possible to be a standalone app.
(so the only advantage would be to have a bit reduced login process, but Talk is doing a great job to detect files accounts and re-use username and server-address, so I do not think that SSO in Talk is worth the effort…)

src/main/java/com/nextcloud/client/integration/deck/DeckApi.java
```diff
@@ -0,0 +1,27 @@
+package com.nextcloud.client.integration.deck;
```
tobiasKaminsky, line 1: Please here also license header

src/main/java/com/nextcloud/client/integration/deck/DeckApi.java
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * Creates a PendingIntent that can be used in a NotificationBuilder to start the Deck app
+     *
+     * @param notification Notification object that should be processed
+     * @param user         The user that is affected by the notification
+     * @return Optional with a PendingIntent or an empty Optional if the notification is not from the
+     * <a href="https://apps.nextcloud.com/apps/deck">Deck server app</a>.
+     */
+    @NonNull
+    Optional<PendingIntent> createForwardToDeckActionIntent(@NonNull final Notification notification, @NonNull final User user);
```
tobiasKaminsky, line 26: linewrap after 120

tobiasKaminsky:
Some (very minor) things and after that let us get CI green and then merge :-)

stefan-niedermann:
@tobiasKaminsky i added the license header and fixed the formatting :)

src/main/java/com/nextcloud/client/integration/deck/DeckApi.java
```diff
@@ -0,0 +1,48 @@
...
+import androidx.annotation.NonNull;
+
+/**
+ * This API is for an integration with the <a href="https://github.com/stefan-niedermann/nextcloud-deck">Nextcloud
+ * Deck</a> app for android.
+ */
+public interface DeckApi {
+
+    /**
+     * Creates a PendingIntent that can be used in a NotificationBuilder to start the Deck app
```
ezaquarii, line 38: Please calrify what does it mean to "start the deck app" in this context? Does it merely open the app, acting like a launcher of a main screen or does it pass some content?

ezaquarii, line 38: Please calrify what does it mean to "start the deck app" in this context? Does it merely open the app, acting like a launcher of a main screen or does it pass some content?

src/main/java/com/nextcloud/client/integration/deck/DeckApiImpl.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @NonNull
+    @Override
+    public Optional<PendingIntent> createForwardToDeckActionIntent(@NonNull Notification notification, @NonNull User user) {
+        if (APP_NAME.equalsIgnoreCase(notification.app)) {
+            final Intent intent = new Intent();
+            for (String flavor : DECK_APP_ID_FLAVOR_SUFFIXES) {
+                intent.setClassName(DECK_APP_ID_BASE + flavor, DECK_ACTIVITY_TO_START);
+                if (context.getPackageManager().resolveActivity(intent, 0) != null) {
+                    Log.i(TAG, "Found deck app flavor \"" + flavor + "\"");
```
ezaquarii, line 67: This log stanza should be removed.
1. we don't need to spam logs with this information, at least not at info level
2. We have our own logger that can be accessed by `Log_OC` helper or (better) `Logger` interface.

src/main/java/com/nextcloud/client/integration/deck/DeckApi.java
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * @param notification Notification object that should be processed
+     * @param user         The user that is affected by the notification
+     * @return Optional with a PendingIntent or an empty Optional if the notification is not from the
+     * <a href="https://apps.nextcloud.com/apps/deck">Deck server app</a>.
+     */
```
ezaquarii, line 44: The diagram tells that it passes account and some kind of "event".


Can we clarify what this API is doing exactly?

My posposal (after consulting the source code):

```suggestion
    /**
    * Creates a PendingIntent that can be used in a NotificationBuilder to open the notification link in Deck app 
    * @see Deck Server App - https://apps.nextcloud.com/apps/deck
    *
    * @param notification Notification that could be forwarded to Deck
    * @param user         The user that is affected by the notification
    * @return If notification can be consumed by Deck, a PendingIntent openinig notification link in Deck app; empty value otherwise
    */
```

src/main/java/com/nextcloud/client/integration/deck/DeckApiImpl.java
stefan-niedermann: I removed logging

src/main/java/com/nextcloud/client/integration/deck/DeckApi.java
ezaquarii: I don't want to block this PR - we can follow-up when merging #5892

stefan-niedermann: I adapted your proposal (but kept the `<a href...`-style for the `@see`-annotation ([because it will then be rendered as a link](https://stackoverflow.com/a/10683345))

> The diagram tells that it passes account and some kind of "event".

This can currently be for example

- User A assigns a card to User B
- User A mentions User B in a comment on a card
- ...

But this are internals of the Deck app and can change. The actual payload / content of the event is not relevant for the File app and should therefore omitted in the documentation IMHO.

ezaquarii:
@stefan-niedermann DCO needs to be set. It's signed-off stanza you can see in commits messages.

You can do it with `git commit --amend --signoff`. Also, some unused imports will make liner unhappy.

src/main/java/com/nextcloud/client/integration/deck/DeckApi.java
ezaquarii: > But this are internals of the Deck app and can change. The actual payload / content of the event is not relevant for the File app

Agree, if Deck content is opaque, it's implementation detail that should not be leaking into the caller.

ezaquarii:
@stefan-niedermann I just noticed that we didn't supply any unit tests for this. Can we think of anything? Given that this uses a `Notifcation` and `PendingIntent` APIs, I guess it must be `androidTest` running on emulator...

ezaquarii:
@stefan-niedermann Here is a basic test for 2 scenarios:
1. deck is installed
2. deck is not available

One thing that struck me when writing this code is that the test is not sensitive to `Notification` contents (apart `app` attribute, obviously). This opens some questions about undefined behaviour when the `Notification` object contains garbage data and - once forwarded to Deck - crashes the app because of some `null` or other malformed stuff.

This might be intentional or not - depending on your point of view.

Pending intent contents are not tested as one cannot do it without some britle reflection kung-fu, but if the contents of outgoing intent shall be tested as well, we can do it after some minor restructuring of the code, as this aspect is not testable a.t.m.

stefan-niedermann:
> This opens some questions about undefined behaviour when the Notification object contains garbage data

No, the content do not have to be tested explicitely. I designed the "other" side very robust. It checks what contents are really available and does not make any assumptions. It trys to use as much data as possible but hast fallbacks if some or all contents are missing.

Besides the contents it might make sense to check if the `notification`- and `user`-objects itself are `null` which would lead to a crash. Don't know how you handle this in the app elsewhere. I added the `@NonNull`-annotations, but we can make them `@Nullable` with an explicite check as well.

Also it's an "Intent" - it's up to the receiver to handle it or not.

Nice test, i completely understand it 🙂

tobiasKaminsky:
Rebased, @stefan-niedermann please one last check, as I had to re-add it in NotificationWork

stefan-niedermann:
As i told you @tobiasKaminsky i don't receive any push notifications at all with the current state of this branch. (Yes, i used the `gplay` flavor and i had installed play services).

Therefore i can't retest it.

If i manually open the notifications in the Android app, i see the following snackbar:

grafik

stefan-niedermann:
Successfully tested and confirmed to work!

grafik

stefan-niedermann:
:heart: :heart: :heart:

PR #5879, opened 2020-04-15
Migrate files sync related jobs to WorkManager API

ezaquarii:
* FilesSyncJob.java converted to FilesSyncWork.kt
* FilesSyncJob removed from code
* Extended ETM screen to allow scheduling immediate
  TestJob (for development experimentation)
* OfflineSyncJob.java conveted to OfflineSyncWork.kt
* OfflineSyncJob removed from code
* Fixed re-scheduling of content observer work
* MediaFoldersDetectionJob.java converted to
  MediaFoldersDetectionWork.kt
* MediaFolderDetectionJob removed from code

Fixes #5881

Signed-off-by: Chris Narkiewicz <hello@ezaquarii.com>

PR #5885, opened 2020-04-16
Do not run insertAllDBEntriesForSyncedFolder directly

### LINKED ISSUE #5882: Enabling auto-upload on Camera folder crashes app with ANR
### Steps to reproduce
1. have a decently sized camera folder
2. go to Auto upload
3. toggle camera folder (cloud icon)

### Expected behaviour
- setting toggles, folder is enabled for auto upload

### Actual behaviour
- Application freezes, ANR follows shortly after

### Environment data
Android version: 10
Device model: OnePlus 6T
Nextcloud app version: master
Nextcloud server version: not relevant

https://github.com/nextcloud/android/blob/5d428fe9a2f7870505d462672956a83266e0b241/src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java#L624-L628

Inside this `if` a massinve database I/O on main thread occurs:
https://github.com/nextcloud/android/blob/5d428fe9a2f7870505d462672956a83266e0b241/src/main/java/com/owncloud/android/utils/FilesSyncHelper.java#L88

This won't work. We need to move this massive I/O to a background thread. For the time being, I propose we don't enable upload of existing files by default.

tobiasKaminsky:
Fix #5882 

Previously we ran insertAllDBEntriesForSyncedFolder on UI, which causes a lag on very big folders.
Now we use FilesSyncJob, which does the same, but is a Job, which runs in background.

@ezaquarii @AndyScherzinger 

### Testing 


ezaquarii:
@tobiasKaminsky This will conflict with https://github.com/nextcloud/android/pull/5879 where `FileSyncJob` has been removed.

tobiasKaminsky:
> @tobiasKaminsky This will conflict with #5879 where `FileSyncJob` has been removed.

True.
Then I will base this onto stable-3.11 only.
Can you make sure that #5879 behaves the same (no lagging UI on saving)?

ezaquarii:
> Then I will base this onto stable-3.11 only.
> Can you make sure that #5879 behaves the same (no lagging UI on saving)?

@tobiasKaminsky This patch cannot be applied there as we're using `BackgroundJobManager` API instead of `FilesSyncHelper`. Semantics seem to be identical, so I'll make analoguous change in my branch using new API.

src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -768,15 +768,7 @@ private static void initiateExistingAutoUploadEntries(Clock clock) {
...
-                    new SyncedFolderProvider(MainApp.getAppContext().getContentResolver(), preferences, clock);
-
-                for (SyncedFolder syncedFolder : syncedFolderProvider.getSyncedFolders()) {
-                    if (syncedFolder.isEnabled()) {
-                        FilesSyncHelper.insertAllDBEntriesForSyncedFolder(syncedFolder, true);
-                    }
-                }
-
+                FilesSyncHelper.startFilesSyncJobNow(null);
                 preferences.setAutoUploadInit(true);
```
ezaquarii, line 772: @tobiasKaminsky I'm not sure what this flag is supposed to do in this context?

tobiasKaminsky: No idea neither, but it is just c&p, so should not be problematic…

Issue #5881, opened 2020-04-15
Content observer work is not re-scheduled after triggered by media update


ezaquarii:
### Steps to reproduce
1. Configure instant upload of camera images
2. take a photo few times

### Expected behaviour
- Instant upload should be triggered shortly after taking a photo

### Actual behaviour
- Nothing happens - media does not trigger content observer task

ezaquarii:
@tobiasKaminsky this should be fixed before the release

tobiasKaminsky:
Is this on current master? Or on your new PR?

I think we changed behaviour from instant sync to ~15min sync, so it would be an expected behaviour for me?

ezaquarii:
It's on current master. This is a regression introduced by migration to `ContentObserverWork` in 808c9098ea7476fd7e6311aa838e81d90c9e0e99.

The work is re-scheduled using `KEEP` flag. Since the work is running when hile re-scheduling occurs, it cannot replace itself.

The fix is included https://github.com/nextcloud/android/pull/5879

tobiasKaminsky:
> The work is re-scheduled iteself using `KEEP` flag. Since the work is unique and running when re-scheduling occurs, it cannot replace itself.
> 
> The fix is included in #5879
> 
> If you'd like to port it to other branch, just replace `KEEP` -> `REPLACE`.

I am a bit lost: is this included in stable-3.11 branch?
If so, we need to fix it, as I want to release 3.11.1 on Thursday.

ezaquarii:
> > The work is re-scheduled iteself using `KEEP` flag. Since the work is unique and running when re-scheduling occurs, it cannot replace itself.
> > The fix is included in #5879
> > If you'd like to port it to other branch, just replace `KEEP` -> `REPLACE`.
> 
> I am a bit lost: is this included in stable-3.11 branch?
> If so, we need to fix it, as I want to release 3.11.1 on Thursday.

@tobiasKaminsky 

Stable uses different version of this task, with a different bug:

https://github.com/nextcloud/android/blob/stable-3.11.0/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt

Here, we are scheduling unconditionally. Whenever the app starts, it schedules a new job. As a side effect, the app will be accumulating new `ContentObserverWork` jobs scheduled to run after restart restart.

#5879 uses `sheduleUniqueWork` to have no more than 1. We could do something similar here as well.

ezaquarii:
[...]

PR #5912, opened 2020-04-21
Make sure only 1 content observer job is scheduled

ezaquarii:
Since content observer is not unique job, each application
restart enqueues another job one-time job that will keep
itself re-scheduled when started.

This change makes sure there is only 1 instance of content
observer scheduled at any given time.

When old. accumulated one-time jobs re-schedule, they will
terminate leaving only 1 instance left, so the system will
self correct.

### DIFF
--- a/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
+++ b/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
@@ -23,6 +23,7 @@ import android.os.Build
 import android.provider.MediaStore
 import androidx.annotation.RequiresApi
 import androidx.work.Constraints
+import androidx.work.ExistingWorkPolicy
 import androidx.work.OneTimeWorkRequest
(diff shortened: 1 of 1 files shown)


ezaquarii:
@tobiasKaminsky This is for stable only. Do not merge to master.

AndyScherzinger:
@ezaquarii so this is fixed on master already?

ezaquarii:
On Wed, Apr 22, 2020 at 02:21:42AM -0700, Andy Scherzinger wrote:
> @ezaquarii so this is fixed on master already?

@AndyScherzinger
Well, it's a bit embarrasing. I "fixed" it and introduced #5881 in turn 🤦.
That will be fixed by pending PR #5879.

ezaquarii:
Please bear in mind that our WorkManager changes
diverged significanlty from master, so some expectations
no longer apply.

AndyScherzinger:
@ezaquarii nothing to be embarrassed ❤️

Issue #5207, opened 2020-01-15
Some design refresh


jancborchardt:
These are some possible design polishing items we could do – some more easy and some more difficult. :) What we assume to be quick wins are marked with ⚡

(No mockups yet as it’s just a quick list, and as it can mostly be seen by just checking out Gmail and Google Drive.)

- [x]  **⚡ Splash screen polishing:** During the splash screen, the status bar on top and the controls bar on the bottom is black. Can we make it blue as well? Will make it look much less boxed in. :) https://github.com/nextcloud/android/pull/5229
- [x] **⚡ Colorless header bar:** The new material design doesn’t have colored header bars or status bars. The branding would only be via FAB and folder icons, which is fine. :)
- [x] **⚡ Slight rounding of the file previews:** Some very slight rounded corners on the image file previews just like we do on server, makes it appear much friendlier. https://github.com/nextcloud/android/pull/5231
- [x] **Get rid of 3-dot menu in top right:** The "Refresh account" option can go away (right?) :D. The Sort as well as Grid view can go below the search field in a specific row, just like Google Drive does.
  - [x] Add sort / switch list view to ReceiveExternalFiles and Trashbin
- [x] **Account switcher in top right:** Again, same as Gmail and other Google Apps, we show the avatar in the top right instead of the drawer.

- [x] **Search bar open by default:** Just like Gmail as well, the bar could be there by default and say **Search in Nextcloud**. When you are in a subfolder, it would change "Nextcloud" to the subfolder name
  - [x] use new header in subfolders
- [x] **⚡ Use bottom sheet for actions menu:** Instead of the small popover we can use the proper bottom sheet which is nicely reachable with the thumb. It should of course show the file in question on top so you know what the action will be carried out on.
- [x] **⚡ Transparent bottom bar:** The Home/Back/Switcher bar on the bottom currently has a black background. If we make it transparent like in Gmail, it will look much lighter and like "more space for your own stuff"
- [x] **⚡ Hide header bar on scrolling:** In Gmail and Material Files (from F-Droid) for example, when you scroll down, the whole header bar hides, and shows again when you scroll up a bit. This gives you a lot of space for your files and is really nice.
- [x] centralize fonts, e.g. difference between switch account popup and drawer
- [x] FirstRunActivity (and follow up while login): white navigation bar, should be blue
- [x] header in e.g. Activities/Notifications looks old, what can we do about it?
- [x] add new header to trash bin, sort is then a filter as sever does not support searching in trash bin
[...]

PR #5966, opened 2020-04-29
New design (UI) - Replaced some icons

Shagequi:
Replaced some icons by Material ones.

about `ic_unshared.xml` icon, it's not in 24dp but in 20dp, it's due to the size of the "shared" icon which is too massive compared to the "three dots icon".

Relative to : https://github.com/nextcloud/android/issues/5207
Based on this : https://github.com/nextcloud/android/issues/5207#issuecomment-617270183


AndyScherzinger:
Since I haven't been involved in the discussions offline (don't worry). Beware that this PR also replaces the "Nextcloud icons" (e.g. for the empty screen), this would mean that the empty screens of web and Android differ (which in general is fine with me, I just don't know if this is clear to everyone involved in the revamp discussions, like @jancborchardt for example)

jancborchardt:
I'm in trains today all day, screenshots would be appreciated to review quicker. :)

(Just as a note, we are also considering switching to Material Design icons for the server, see https://github.com/nextcloud/nextcloud-vue/issues/1051 )

tobiasKaminsky:
> (Just as a note, we are also considering switching to Material Design icons for the server, see [nextcloud/nextcloud-vue#1051](https://github.com/nextcloud/nextcloud-vue/issues/1051) )

Nice idea, but what about iOS?
Material icons might look not well integrated there…?
And having on web/Android material design, but other icons on iOS is also strange.

Shagequi:
About iOS, I let you watch with @immortal79 
And about screenshots, i'll create some comparison screenshots for quicker/better review.

jancborchardt:
> > (Just as a note, we are also considering switching to Material Design icons for the server, see [nextcloud/nextcloud-vue#1051](https://github.com/nextcloud/nextcloud-vue/issues/1051) )
> 
> Nice idea, but what about iOS?
> Material icons might look not well integrated there…?
> And having on web/Android material design, but other icons on iOS is also strange.

Everything is a trade-off. Just like with the fonts, it's best to use the native stuff for the platform. Alas there's no "native" icons for web, so we just have to choose the best ones, which happen to be Material Design.

Shagequi:
Before : 
Capture d’écran 2020-04-30 à 11 39 31

After :
Capture d’écran 2020-04-30 à 11 58 01


___

Before : 
Capture d’écran 2020-04-30 à 11 40 52

After : 
Capture d’écran 2020-04-30 à 11 58 43

AndyScherzinger:
@Shagequi took the liberty to reformat your screenshot post using tables for a side-by-side visual diff :)

Shagequi:
Thank you @AndyScherzinger  <3

Shagequi:
Hi @jancborchardt , 

I followed the mockups : https://infomaniak.invisionapp.com/share/3VWA11OK9E8#/410365622_App_Nextcloud_-_Drive_2 present in this issue : https://github.com/nextcloud/android/issues/5207 

Concerning the "favorite icon" (star), the SVG used with the "white halo" is not the one from the Material library, and I don't find the current icon unreadable. If you want me to put the old one back it's possible but don't recommand it, or I recommand to find another way to display the folder as a "favorite". 

About the "share" icon, I'll ask the person who made the mockups to find a solution, and to be honest, I didn't know what was the purpose of the low opacity before you told me 😅

I come back as soon as I get some answers ! Don't hesitate to tell me if you have some ideas to transmit

Shagequi:
@jancborchardt I just wanted to check something before going to talk to the UX about it... 
When the file is shared with an user, the icon/pp of the user is shown directly instead of the "share icon". 

So, when you share the file with someone on your cloud, you see this : 
Capture d’écran 2020-05-04 à 05 12 59

When you share with link, you see this : 
Capture d’écran 2020-05-04 à 05 13 23

When you do both, you see this : 
Capture d’écran 2020-05-04 à 05 10 11

We don't need to make the opacity lower, because the icon "share" is no longer displayed once the file shared.
What do you think about it ?

tobiasKaminsky:
> We don't need to make the opacity lower, because the icon "share" is no longer displayed once the file shared.
> What do you think about it ?

I assume that Jan is mixing this up with server, where there are some shares which are shown with the "share icon" only.

As Kilian pointed out correctly, we do not have any "greyed out share" symbol anymore.
So we can keep it in 100% black.

AndyScherzinger:
I agree with @Shagequi and @tobiasKaminsky while the icon color is rather grey600, not black 😀

Shagequi:
@jancborchardt Hi, do you have some news ? :D

Shagequi:
@tobiasKaminsky Good morning, may we merge this please?

tobiasKaminsky:
Unrelated errors on drone.

tobiasKaminsky:
@Shagequi where did you get the svg from?
We store in /drawable_resources the svgs and now I need to use the new folder icon as sources for our encrypted folder icon.
But these files are outdated?

Shagequi:
@tobiasKaminsky 
These icons provide from Android Studio clip arts.

When you create a vector asset, instead of selecting one on your computer, you choose "Clip Art", this will give you access to the recent Material icons.

AndyScherzinger:
@tobiasKaminsky @Shagequi I'll take care of the folder icon and will see that I add the svg versions.

Shagequi:
Okay, tell me if you need more informations ^_^ @AndyScherzinger

AndyScherzinger:
@Shagequi I guess I am fine, only thing would be: Did you add any "non-standard" material icons?

Shagequi:
@AndyScherzinger 
I only took the clip arts provided by Google and watched the mockups : https://github.com/nextcloud/android/issues/5207
So, I'd say no ;)

PR #5979, opened 2020-04-30
fix #4540

### LINKED ISSUE #4540: 'Send to Nextcloud' files remain in tmp folder, never moved to correct location
### Actual behaviour
- Some files stay in the accounts tmp/ directory indefinitely and are never moved to the correct storage location. Thus, an external Gallery app for example will show two different "collections" (folders/albums/...). One with the files affected by this bug (found in tmp/...) and another one with the rest (found in the correct storage location).

This is very similar to #4341, but my problem is not so much about wasted space in tmp/ than it is about the files not appearing in the correct location.

### Expected behaviour
- all uploaded files should eventually land in the correct storage location, so external Apps can give a consistent view.
 
### Steps to reproduce
1. Upload a file through "Send to -> Nextcloud" (from Gallery app->Camera folder for example)
2. Open NC app and synchronize the file
3. File now lives in tmp/ folder forever.

#### Bonus badness
In NC app's Gallery view, the uploaded photo is even displayed correctly alongside all other photos of the designated folder that are in the correct location - I guess it's merging tmp/ and correct location. And yup, that means you cannot see the problem inside Nextcloud app at all. However, using any other app now suffers from having the designated directory twice on the file system.

### The bug can be mitigated manually:
1. Navigate to the affected file in NC app.
2. Select "Delete -> Local".
3. Synchronize file again.
4. File will live in correct location from now on.
But this is far from practical to do for every upload.

Deleting the tmp/ directory structure with a file manager and un/reinstalling Nextcloud app do not solve the problem.

Thanks!

### Environment data
Android version: 9

Device model: Samsung S9 and A40

Nextcloud app version: 3.8.0

Nextcloud server version: 16.0.4

tobiasKaminsky:
fix #4540

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
+++ b/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
@@ -981,6 +981,7 @@ public class UploadFileOperation extends SyncOperation {
 
             case FileUploader.LOCAL_BEHAVIOUR_DELETE:
                 originalFile.delete();
+                mFile.setStoragePath("");
                 getStorageManager().deleteFileInMediaScan(originalFile.getAbsolutePath());
                 saveUploadedFile(client);
                 break;


AndyScherzinger:
tested successfully by issue reporters

tobiasKaminsky:
Also tested by bug reporter.

Issue #5891, opened 2020-04-16
UploadsStorageManager: Couldn't read row 1872, col 1 from CursorWindow.


andreapx:
### Steps to reproduce
1.  Open the app

### Actual behaviour
- The app crashes

### Environment data
Android version: 9

Device model: OnePlus 6T

Stock or customized system: Stock

Nextcloud app version: 30110090

Nextcloud server version: 18.0.3

Reverse proxy: nginx

************ CAUSE OF ERROR ************

java.lang.IllegalStateException: Couldn't read row 1872, col 1 from CursorWindow. Make sure the Cursor is initialized correctly before accessing data from it.
at android.database.CursorWindow.nativeGetString(Native Method)
at android.database.CursorWindow.getString(CursorWindow.java:465)
at android.database.AbstractWindowedCursor.getString(AbstractWindowedCursor.java:51)
at android.database.CursorWrapper.getString(CursorWrapper.java:137)
at com.owncloud.android.datamodel.UploadsStorageManager.createOCUploadFromCursor(UploadsStorageManager.java:321)
at com.owncloud.android.datamodel.UploadsStorageManager.getUploads(UploadsStorageManager.java:302)
at com.owncloud.android.datamodel.UploadsStorageManager.getFailedUploads(UploadsStorageManager.java:368)
at com.owncloud.android.files.services.FileUploader.retryFailedUploads(FileUploader.java:1005)
at com.owncloud.android.utils.FilesSyncHelper.lambda$restartJobsIfNeeded$0(FilesSyncHelper.java:222)
at com.owncloud.android.utils.-$$Lambda$FilesSyncHelper$fa2LBpPOxRVQOm4F_Glue5u4hdM.run(Unknown Source:10)
at java.lang.Thread.run(Thread.java:764)

************ APP INFORMATION ************
ID: com.nextcloud.client
Version: 30110090
Build flavor: generic

************ DEVICE INFORMATION ************
Brand: OnePlus
Device: OnePlus6T
Model: ONEPLUS A6013
Id: PKQ1.180716.001
Product: OnePlus6T

************ FIRMWARE ************
SDK: 28
Release: 9
Incremental: 1909112330

tobiasKaminsky:
When  did this happen?

andreapx:
A couple of days ago when I tried to open it.
Now that I think about it, the last thing that I've done it's been to add the WhatsApp folder to the autoupload. But that seems to be working fine.
[...]

PR #5985, opened 2020-04-30
Add pagination to SQL query in UploadsStorageManager

gjedeer:
I've added pagination to the SQL query in order to fix issue #5891 which exposed a bug in Android - cursors larger than 2MB threw an exception.

### DIFF
--- a/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
+++ b/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
@@ -39,6 +39,8 @@ import com.owncloud.android.lib.common.operations.RemoteOperationResult;
 import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.operations.UploadFileOperation;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Observable;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -284,35 +286,78 @@ public int removeUploads(String accountName) {
...
-        OCUpload[] list;
-
-        Cursor c = getDB().query(
-                ProviderTableMeta.CONTENT_URI_UPLOADS,
-                null,
-                selection,
-                selectionArgs,
-                null
-        );
+        ArrayList<OCUpload> uploads = new ArrayList<>();
```
AndyScherzinger, line 289: Bets would be to use the interface `List`

AndyScherzinger:
Hi @gjedeer and welcome to Nextcloud Android app! 

Just a minor comment regarding types, else looks good to me code-wise 👍

src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
gjedeer: @AndyScherzinger I'm not that proficient with Java, if I do that do I still have to convert the ArrayList to an array when returning data, or is the interface not compatible with OCUpload[] ?

AndyScherzinger: Will have to check

src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -284,35 +286,78 @@ public int removeUploads(String accountName) {
...
+        final long pageSize = 100;
```
ezaquarii, line 290: We can probably make this number bigger to issue smaller amount of queries.

src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -284,35 +286,78 @@ public int removeUploads(String accountName) {
...
+                    pageSelection = "(" + selection + ") AND _id < ?";
+                } else {
+                    pageSelection = "_id < ?";
+                }
+                if(selectionArgs != null) {
+                    pageSelectionArgs = Arrays.copyOf(selectionArgs, selectionArgs.length + 1);
+                } else {
+                    pageSelectionArgs = new String[1];
+                }
+                pageSelectionArgs[pageSelectionArgs.length - 1] = String.valueOf(lastRowID);
```
ezaquarii, line 311: Instead of doing ANSI C style array acrobatics, we could use `ArrayList` here and convert that into `[]String` when supplying the selection to query.

src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -284,35 +286,78 @@ public int removeUploads(String accountName) {
...
+                Log_OC.d(TAG, String.format("QUERY: %s ROWID: %d", pageSelection, lastRowID));
```
ezaquarii, line 312: Those logs are probably not really needed and will bombard our log file with content.

src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -284,35 +286,78 @@ public int removeUploads(String accountName) {
...
+                    } while (c.moveToNext() && !c.isAfterLast());
+                }
+                c.close();
+                Log_OC.v(TAG, String.format("getUploads() got %d rows from page %d, %d rows total so far, last ID %d",
+                            rowsRead,
+                            page,
+                            rowsTotal,
+                            lastRowID
+                ));
+                offset += rowsRead;
```
ezaquarii, line 347: `offset` is unused.

src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -284,35 +286,78 @@ public int removeUploads(String accountName) {
...
-                    OCUpload upload = createOCUploadFromCursor(c);
-                    if (upload == null) {
-                        Log_OC.e(TAG, "OCUpload could not be created from cursor");
-                    } else {
-                        list[c.getPosition()] = upload;
-                    }
-                } while (c.moveToNext() && !c.isAfterLast());
+            if (c != null) {
+                if (c.moveToFirst()) {
+                    do {
```
ezaquarii, line 328: This can be probably extracted into a method to make this loopy block more readable. Nested loops will cause linter anxiety.

ezaquarii:
@gjedeer Please let us know if you'd like to work on it. I remeber you quoted that your time is limited so we can jump in and take over if not.

gjedeer:
@ezaquarii Thank you for your review, I don't think I have much more time to work on it. So please do take over.

Just a note though. This PR fixes the cursor-too-big crash, but it makes the whole phone crawl. The background uploader process makes the phone so slow that unlocking the screen takes 30 seconds.

I think the process needs much bigger refactoring, perhaps the background process doesn't need to query all pending uploads after every file?

gjedeer:
> Since database contents can change between queries, this multi-query should wrap those selects in transaction.

No, it doesn't have to. Concurrency is the reason why I went away from LIMIT/OFFSET pagination to ORDER BY _id and storing latest _id in the loop.

ezaquarii:
@gjedeer I plan to take over this issue and finish it. What is the data fixture that would allow me to test this?

I'm quite surprised that this PR causes the phone to crawl - DB queries should be performed on background thread and I wouldn't expect it to impact whole OS.

ezaquarii:
Internalized here: https://github.com/nextcloud/android/pull/6037

PR #5991, opened 2020-05-01
New design (UI) - loader thumbnail

sirambd:
### Description

A shimmer effect is played while a thumbnail is being downloaded

effect
effect2


AndyScherzinger:
Nice work @sirambd 👌

Only thing from my side, the contrast should probably be dialed down quite a bit. Right know it seems to mimic shiny metal a bit too much while I'd say it should have a rather mate feel. (not sure if I found the right words to describe what I mean)

src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -514,9 +515,15 @@ protected Bitmap doInBackground(ThumbnailGenerationTaskObject... params) {
 
             } catch(OutOfMemoryError oome) {
                 Log_OC.e(TAG, "Out of memory");
```
ezaquarii, line 517: Logging `oome` would give us stacktrace. This log line is pretty useless as it is now. While beinghere, can we fix it?

JorisBodin:
@jancborchardt 
Like this? https://github.com/elye/loaderviewlibrary

jancborchardt:
> @jancborchardt
> Like this? https://github.com/elye/loaderviewlibrary

Exactly, perfect! :) Thanks for the link

AndyScherzinger:
The lib is rather legacy since it doesn't support AndroidX but the deprecated AppCompat implementation :(

JorisBodin:
> The lib is rather legacy since it doesn't support AndroidX but the deprecated AppCompat implementation :(

We use this commit :)
_Bump up from 1.5.0 to 2.0.0, as we have migrated to AndroidX, upgrade…_

AndyScherzinger:
@JorisBodin Nice! 👍 I didn't check a maven repo, just saw they didn't do a release on Github.

@jancborchardt the lib is fine with me (see @JorisBodin's comment)

JorisBodin:
It's better :D
untitled

AndyScherzinger:
Nice 👍

src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
AndyScherzinger: @ezaquarii what would you suggest? Simply removing the log statement and be done?

JorisBodin: And remove `//} else {  do nothing` just above

ThumbnailsCacheManager.java need a good refactor.

src/main/res/layout/grid_sync_item.xml
```diff
@@ -50,15 +49,6 @@
         android:gravity="center_horizontal"
         android:orientation="horizontal">
 
-        <TextView
```
AndyScherzinger, line 53: @JorisBodin this textview is needed! as is it's non-translatable string!

src/main/res/layout/grid_sync_item.xml
```diff
@@ -18,29 +17,29 @@
...
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:id="@+id/grid_item_container"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
     android:layout_gravity="center_horizontal"
     android:background="@drawable/list_selector"
     android:gravity="center_horizontal">
 
-    <com.owncloud.android.ui.SquareImageView
+    <ImageView
```
AndyScherzinger, line 28: this needs to be SquareImageView since the images aren't square anymore due to this change

src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
tobiasKaminsky: My aim is to replace this at some point by a well known lib, e.g. glide.

JorisBodin:
https://github.com/nextcloud/android/pull/6042 broke this PR :(
We're looking to fix it

sirambd:
> #6042 broke this PR :(
> We're looking to fix it

It's fixed in the new PR https://github.com/nextcloud/android/pull/6044#issuecomment-628386871

PR #5992, opened 2020-05-01
New design (UI) - Sort and Gridview switch Bar

### LINKED ISSUE #6009: Displaying SORT option on the action bar
I use SORT very often and changing sort mode currently requires 3 taps. Given the importance of the function, I suggest to put it on the action bar next to search.

Solution pictured: (no other changes to the function or design)

sort

JorisBodin:
Hello :)

I added the sort button and Gridview switch like mockup: https://github.com/nextcloud/android/issues/5207#issuecomment-617270183

This new bar and the FAB hides during scroll.
Why FAB? Because you can't press the menu button of the last item in the list.

effect1

I haven't removed the menu. To already validate this feature.

Normally it should have less spotbugs.

Resolves #6009


AndyScherzinger:
Nice work @JorisBrodin 👍

The grid/list switch should be a bit more to the right to be vertically, center aligned with the 3dot menu icons. 

As for the Fab, the list/grid has a larger area at the end of the lust/grid with the file/folder count so the fab doesn't hide any menu or else if you just scroll by :)
Fine though to hide it on scroll down, while I guess it should show again on scroll up (haven't tested, just looked at the gif)

src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1293,9 +1307,45 @@ private void invalidateActionMode() {
...
 
     public void sortFiles(FileSortOrder sortOrder) {
+        setSortButton(sortOrder);
         mAdapter.setSortOrder(mFile, sortOrder);
     }
 
+    private void setSortButton(FileSortOrder sortOrder) {
+        int nameId;
+        switch (sortOrder.name) {
+            case "sort_new_to_old":
```
ezaquarii, line 1317: Can we move those into consts?

JorisBodin:
@jancborchardt 
> The sorting dropdown should use the same wording in the dropdown as it does in the modal. That is "A–Z", "Newest first" etc.

It doesn't match the mockup...
The arrow on the right indicates the sort. So there's no need to indicate "A-Z" or "Z-A"
It's same google drive :)
And "Newest first" etc it's to long and not pretty

jancborchardt:
> It doesn't match the mockup...
The arrow on the right indicates the sort. So there's no need to indicate "A-Z" or "Z-A"
It's same google drive :)
And "Newest first" etc it's to long and not pretty

We will largely follow the mockup, but in some cases it might need to be adjusted or it’s things we forgot to talk about. The app should be internally consistent, not use 2 different kind of languages for the same thing. :)

We can write "A–Z", "Z–A", "Newest", "Oldest", "Biggest", "Smallest". Short and clear

JorisBodin:
> > It doesn't match the mockup...
> The arrow on the right indicates the sort. So there's no need to indicate "A-Z" or "Z-A"
> It's same google drive :)
> And "Newest first" etc it's to long and not pretty
> 
> We will largely follow the mockup, but in some cases it might need to be adjusted or it’s things we forgot to talk about. The app should be internally consistent, not use 2 different kind of languages for the same thing. :)
> 
> We can write "A–Z", "Z–A", "Newest", "Oldest", "Biggest", "Smallest". Short and clear

Currently on the nextcloud app on Google play. 
French translation is easier to understand
Screenshot_20200504-100228_Nextcloud

JorisBodin:
I fixed the padding, the lowercase, and the arrows.
@jancborchardt you prefer?
Screenshot_1588585800

src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
JorisBodin: It's fix :)

src/main/res/values-fr/strings.xml
```diff
@@ -633,6 +633,9 @@ Attention la suppression est irréversible.</string>
     <string name="signup_with_provider">Se connecter avec un fournisseur</string>
     <string name="single_sign_on_request_token" formatted="true">Autoriser %1$s à accéder à votre compte Nextcloud  %2$s ?</string>
     <string name="sort_by">Trier par</string>
+    <string name="sort_by_date">Date</string>
```
tobiasKaminsky, line 636: This is not needed and will be deleted after next transifex sync.
Translations (to other languages) need to be done on transifex.com

jancborchardt:
As said it can just be "Newest" when shown in the list view, not "Newest first". :)

Also let's put a downward-pointing riangle right of the text (not an arrow) it so it's more obvious it's a control and interactable kind of like a dropdown.

"Date" is as if it will open a date picker, which in this context makes no sense.

JorisBodin:
> As said it can just be "Newest" when shown in the list view, not "Newest first". :)

I liked to have a short word. Cause in French, for example, "Newest" = "Les plus récents"

The arrow indicates if it is ascending or descending. (like google drive)

JorisBodin:
@jancborchardt 
untitled

src/main/res/values-fr/strings.xml
JorisBodin: Fixed

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
```diff
@@ -233,7 +232,6 @@ private boolean onDisablePowerSaveCheckClicked(MenuItem powerCheck) {
 
     private void showPowerCheckDialog() {
         AlertDialog alertDialog = new AlertDialog.Builder(this)
-            .setView(findViewById(R.id.root_layout))
```
tobiasKaminsky, line 236: Why is this removed?
If it is a valid point, please still have it in a separate PR, otherwise revert please.

src/main/java/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -740,4 +745,12 @@ static public void showErrorAndFinishActivity(Activity activity, String errorMes
         Toast.makeText(activity, errorMessage, Toast.LENGTH_LONG).show();
         activity.finish();
     }
+
+    static public void openSortingOrderDialogFragment(FragmentManager supportFragmentManager, FileSortOrder sortOrder) {
+        FragmentTransaction fragmentTransaction = supportFragmentManager.beginTransaction();
+        fragmentTransaction.addToBackStack(null);
+
+        SortingOrderDialogFragment mSortingOrderDialogFragment = SortingOrderDialogFragment.newInstance(sortOrder);
```
tobiasKaminsky, line 753: Just a minor nitpick:
- m is a legacy way of showing a member variable of a class (which is accessible in whole class)
- we try to get rid of the m prefix, as it is shown by AndroidStudio

You can also combine it to SortingOredDialogFragment.newInstance(…).show(…).
(with a wrap after .newInstance(…) for readability

tobiasKaminsky:
Some minor things, then ready to go.

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
JorisBodin: Because it's useless.
But okay, we can go back with warnings...

JorisBodin:
@tobiasKaminsky 
It's now good for you?

tobiasKaminsky:
I'll merge as soon as drone is done :)

tobiasKaminsky:
Last thing, according to lint:
> ../../../src/main/res/drawable/ic_folder_black_24dp.xml:1: The resource R.drawable.ic_folder_black_24dp appears to be unused

jancborchardt:
Btw @JorisBodin, now the "Sort" and "Grid view" entries in the 3-dot menu on the top right can be removed, right?

JorisBodin:
> Btw @JorisBodin, now the "Sort" and "Grid view" entries in the 3-dot menu on the top right can be removed, right?

@jancborchardt Yes, but not in same PR :)

PR #6002, opened 2020-05-04
Fix appbar color in ReceiveExternalFilesActivity

JorisBodin:
Remove useless setcolor. Otherwise the appbar was the wrong color.


AndyScherzinger:
Works like a charm @JorisBodin

PR #6029, opened 2020-05-08
New design (UI) - Folders icons are now smaller

Shagequi:
Hi,

All is in the title. Based on this : https://github.com/nextcloud/android/issues/5207#issuecomment-617270183. 

|Before|After|
|---|---|
|Capture d’écran 2020-05-08 à 15 16 24|Capture d’écran 2020-05-08 à 15 15 13

It'll be enhanced by other changes to come.

Signed-off-by: Kilian Périsset <kilian.perisset@infomaniak.com>

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
+++ b/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
@@ -37,6 +37,7 @@ import android.graphics.drawable.Drawable;
 import android.os.Handler;
 import android.os.Looper;
 import android.text.TextUtils;
+import android.util.TypedValue;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
@@ -626,6 +627,11 @@ public class OCFileListAdapter extends RecyclerView.Adapter<RecyclerView.ViewHol
                                     Context context,
                                     LoaderImageView shimmerThumbnail) {
         if (file.isFolder()) {
+            // Applying a smaller size for thumbnail, if it's a folder
+            if (!gridView) {
+                int folderThumbnailSize = (int) context.getResources().getDimension(R.dimen.folder_thumbnail_size);
+                thumbnailView.setLayoutParams(new FrameLayout.LayoutParams(folderThumbnailSize, folderThumbnailSize));
+            }
             thumbnailView.setImageDrawable(MimeTypeUtil
                                                .getFolderTypeIcon(file.isSharedWithMe() || file.isSharedWithSharee(),
                                                                   file.isSharedViaLink(), file.isEncrypted(),
--- a/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
+++ b/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
@@ -37,7 +37,6 @@ import android.graphics.drawable.Drawable;
 import android.os.Handler;
 import android.os.Looper;
 import android.text.TextUtils;
-import android.util.TypedValue;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
--- a/src/main/res/layout/list_item.xml
+++ b/src/main/res/layout/list_item.xml
@@ -54,7 +54,6 @@
                 android:id="@+id/thumbnail"
                 android:layout_width="@dimen/file_icon_size"
                 android:layout_height="@dimen/file_icon_size"
-                android:layout_marginStart="@dimen/standard_half_margin"
                 android:contentDescription="@null"
                 android:src="@drawable/folder"/>
         </FrameLayout>
--- a/src/main/res/values/dims.xml
+++ b/src/main/res/values/dims.xml
@@ -142,4 +142,5 @@
     <dimen name="button_corner_radius">24dp</dimen>
     <integer name="media_grid_width">4</integer>
     <dimen name="copy_internal_link_padding">10dp</dimen>
+    <dimen name="folder_thumbnail_size">30dp</dimen>
 </resources>


src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -37,6 +37,7 @@
 import android.os.Handler;
 import android.os.Looper;
 import android.text.TextUtils;
+import android.util.TypedValue;
```
tobiasKaminsky, line 40: Is this needed?

src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -626,6 +627,11 @@ private static void setThumbnail(OCFile file,
                                     Context context,
                                     LoaderImageView shimmerThumbnail) {
         if (file.isFolder()) {
+            // Applying a smaller size for thumbnail, if it's a folder
+            if (!gridView) {
+                int folderThumbnailSize = (int) context.getResources().getDimension(R.dimen.folder_thumbnail_size);
+                thumbnailView.setLayoutParams(new FrameLayout.LayoutParams(folderThumbnailSize, folderThumbnailSize));
+            }
```
tobiasKaminsky, line 634: I am unsure if this is peformance wise the best idea.
(or if we at all need to check performance here…)

But if we want to, it might be faster to just load an icon of the same size, but with a smaller folder symbol on it.

Shagequi: Using a different icon for each situation under the pretext of performance is not a good idea in my opinion.
In this case, it doesn't imply a "heavy" process, it's a simple check with a simple view change.

src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
Shagequi: No, sorry, a mistake

Shagequi: Has been fixed.

Shagequi:
Hi @jancborchardt 

Thanks for your feedback.
I apply what's designed in mockups right here : https://github.com/nextcloud/android/issues/5207#issuecomment-617270183 (https://infomaniak.invisionapp.com/share/3VWA11OK9E8#/screens/410365622_App_Nextcloud_-_Drive_2). 

I'm willing to start over as many times as needed to keep everyone satisfied.
But, What would you like to do right now ? 
Keep the same size for all the folders is possible, but it doesn't match the mockups :/

jancborchardt:
@Shagequi sorry for the miscommunication. The folder icon size is great in the grid view as in the mockup – but for the list view, the current folder icon size is better, so we should keep that.

I checked the current mockup again to make sure there’s no inconsistencies, and 2 more things came up (which I gave as feedback before as well):
- The search requires 2 characters minimum – that is unnecessarily restrictive, instead the current view should be filtered first for quick results.
- The "Menu" navigation sliding in from the right but then sticking to the left is strange. Would look nicer if it is a full view list.

The mockup also doesn’t have other details like left/right animation for going into folders, so I didn’t assume it was so set in stone but a basis for what we’ll work on. Sorry if that was the wrong assumption, and hope the comment clears it up? :)

Shagequi:
@jancborchardt 

Concerning your comments about mockups and general design, I've copied them in the main UI thread : https://github.com/nextcloud/android/issues/5207 

Concerning the comments about this PR, I've pushed a fix to vertically align the folder icon (I forgot to remove a margin, sorry). 

Here is the new alignment : 

Capture d’écran 2020-05-11 à 12 34 32

|Mode|Before|After|
|---|---|---|
|Light mode|Capture d’écran 2020-05-11 à 12 42 12|Capture d’écran 2020-05-11 à 12 40 03|
|Dark mode|Capture d’écran 2020-05-11 à 12 41 35|Capture d’écran 2020-05-11 à 12 40 47


___

To be honest, I don't really approve all of your comments, if you allow me to show you Google Drive, here is their files and their folders : 

Screenshot_20200511-124649_Drive

You can see : 
- Small file/folders thumbnails (not in "a circle" like Material says)
- A small "shared" icon in file description (we don't need a giant symbol to determine if a folder is shared or not)

So... We don't need to base an entire design on public Material guideline, these are recommendations, not rules. Google doesn't base everything on its public guideline.

> Just noticed however: The folder icon of the regular folder and the shared folder (with the icon) is not the same, that we need to fix of course. :)

Yes we'll talk about that in another issue 👍

Shagequi:
image

jancborchardt:
> Small file/folders thumbnails (not in "a circle" like Material says)

Yeah, no matter what Material says – my point is how Google Drive does it with the tiny file type icons just looks bad.

We can look at other big players out there:
### Dropbox

### OneDrive

### iOS Files
While we are at it. Yes it’s iOS, but small filetype icons looking bad transcends operating systems:



> A small "shared" icon in file description (we don't need a giant symbol to determine if a folder is shared or not)

The difference between our icon and their icon is that ours is an action button and theirs is purely an indicator. In Nextcloud, you can quickly share, and you can quickly see who else has access to a file with the avatars.

> So... We don't need to base an entire design on public Material guideline, these are recommendations, not rules. Google doesn't base everything on its public guideline.

Exactly – this is why I’m giving the feedback and doing the design reviews, and also look at other apps and platforms. Material design is often great, sometimes not, for the rest we can still look at cognitive science and user research.

tobiasKaminsky:
Ready to release

tobiasKaminsky:
(sorry was mixing up PRs)
As this is "purely" UI, I wait until you sorted this out :)

JorisBodin:
@jancborchardt I completely agree with you.
But it's annoyed Kilian, is to do an unnecessary PR. Because the mockup has been validated. And it's not animation. And the list of changes is written and validated here: https://github.com/nextcloud/android/issues/5207#issuecomment-617270183

@tobiasKaminsky you can close this PR...

tobiasKaminsky:
Sorry to hear :-(
Indeed wasted dev time is nothing we want…

jancborchardt:
Sorry about the miscommunication @Shagequi @JorisBodin, I should have more thoroughly reviewed the mockup. I was pretty amazed at the initiative on your side and didn’t want to already get into a deep discussion up front, cause usually we have enough discussions still on the pull requests.

I do understand though (and actually like it more as well) that you prefer to validate the mockups and then work based on that. To that effect I hope the additional feedback I wrote for the mockups helps clear up any other issues. :)

PR #6035, opened 2020-05-09
Add etag column to capabilities table

kesselb:
- [x] Requires https://github.com/nextcloud/android-library/pull/434
- [x] Reset `androidLibraryVersion` to `master-SNAPSHOT`

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/operations/GetCapabilitiesOperation.java
+++ b/src/main/java/com/owncloud/android/operations/GetCapabilitiesOperation.java
@@ -19,6 +19,7 @@
  */
 package com.owncloud.android.operations;
 
+import com.owncloud.android.datamodel.FileDataStorageManager;
 import com.owncloud.android.lib.common.OwnCloudClient;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult;
 import com.owncloud.android.lib.resources.status.GetCapabilitiesRemoteOperation;
@@ -32,10 +33,11 @@ public class GetCapabilitiesOperation extends SyncOperation {
 
     @Override
     protected RemoteOperationResult run(OwnCloudClient client) {
-        OCCapability currentCapability = null;
+        final FileDataStorageManager storageManager = getStorageManager();
 
-        if (getStorageManager() != null && getStorageManager().getAccount() != null) {
-            currentCapability = getStorageManager().getCapability(getStorageManager().getAccount().name);
+        OCCapability currentCapability = null;
+        if (storageManager.getAccount() != null) {
+            currentCapability = storageManager.getCapability(storageManager.getAccount().name);
         }
 
         RemoteOperationResult result = new GetCapabilitiesRemoteOperation(currentCapability).execute(client);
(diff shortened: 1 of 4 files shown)


kesselb:
image

Seems to work :tada:

src/main/java/com/owncloud/android/operations/GetCapabilitiesOperation.java
```diff
@@ -32,8 +32,13 @@
 
     @Override
     protected RemoteOperationResult run(OwnCloudClient client) {
-        GetCapabilitiesRemoteOperation getCapabilities = new GetCapabilitiesRemoteOperation();
-        RemoteOperationResult result = getCapabilities.execute(client);
+        OCCapability currentCapability = null;
+
+        if (getStorageManager() != null && getStorageManager().getAccount() != null) {
```
ezaquarii, line 37: 1. there should be 1 statement per line
2. save the storage manager into a `final` so we don't need to call the getter over and over again; bonus: value returned by the getter can become null between invocations, defying safety `null` check.
3. same for account

src/main/java/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -2193,6 +2194,24 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
...
+
+                    upgraded = true;
+                    db.setTransactionSuccessful();
+                } finally {
+                    db.endTransaction();
+                }
+            }
+
+            if (!upgraded) {
+                Log_OC.i(SQL, String.format(Locale.ENGLISH, UPGRADE_VERSION_MSG, oldVersion, newVersion));
```
ezaquarii, line 2213: Ok, so what happens if the upgrade fails? It looks like we completely ignore that fac and just happily continue like all is fine. This is pretty much undefined behavior.

kesselb: That seems to be enough for the 56 upgrades before :)

ezaquarii: I'm not a fan of idea that it's ok make bad things worse.

If he upgrade fails, we should be notified about it properly, not via some urelated crash. If we can't or don't want to do anything, I'd throw exception. This will trigger an error message pop-up that - with a bit of luck - will land in our issues page with a spot-on error message.

Hopefully, if we set a good example, this bad behaviour will stop here.

ezaquarii: One thing that is nasty about database migrations is that if failed, the user is left with app that is disfunctional, potentially leading to a data loss.

If we throw a well defined exception here, we can catch that in `ShowErrorActivity` and propose a workaround of downgrading the app to previous version.

@tobiasKaminsky Worth pursuing from UX point of view?

tobiasKaminsky: Let us move this away from this PR.
Both of you are right: it is a nasty way, but we are doing it since the beginning of the app.

Let us discus it in a new issue and then we can change it afterwards.

tobiasKaminsky:
Is this ready to review? @kesselb

kesselb:
Yes but we need the library update first :)

tobiasKaminsky:
> Yes but we need the library update first :)

library PR merged :-)

tobiasKaminsky:
Thanks for your help!
Once CI is done, we can merge it!

PR #6042, opened 2020-05-11
New design (UI) - Favorite and "downloaded" icons alignment

Shagequi:
Hi, 

Always based on the following mockups : https://github.com/nextcloud/android/issues/5207#issuecomment-617270183. 

The favorite icon is now aligned on the top right, and the downloaded icon on the bottom right of the thumbnail. 
The "Floating" effect is normal for folders, because the folder icon doesn't take all the thumbnail reserved/available space.

Here is the alignment proof : 
Capture d’écran 2020-05-11 à 10 32 27

And here is the difference with "now":  

|Before|After|
|---|---|
|Capture d’écran 2020-05-11 à 10 11 16|Capture d’écran 2020-05-11 à 10 12 50|

Signed-off-by: Kilian Périsset <kilian.perisset@infomaniak.com>

### Testing 

### DIFF
--- a/src/main/res/layout/list_item.xml
+++ b/src/main/res/layout/list_item.xml
@@ -22,65 +22,73 @@
     android:layout_width="match_parent"
     android:layout_height="@dimen/standard_list_item_size"
     android:background="@drawable/list_selector"
+    android:baselineAligned="false"
     android:descendantFocusability="blocksDescendants"
     android:foreground="?android:attr/selectableItemBackground"
-    android:baselineAligned="false"
     android:orientation="horizontal">
 
-    <RelativeLayout
+    <androidx.constraintlayout.widget.ConstraintLayout
         android:layout_width="@dimen/standard_list_item_size"
         android:layout_height="@dimen/standard_list_item_size"
-        android:paddingStart="@dimen/zero"
-        android:paddingTop="@dimen/standard_padding"
-        android:paddingEnd="@dimen/standard_quarter_padding"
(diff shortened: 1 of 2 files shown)


src/main/res/layout/list_item.xml
```diff
@@ -22,65 +22,73 @@
...
-            android:layout_alignParentTop="true"
-            android:layout_alignParentEnd="true"
-            android:layout_marginEnd="@dimen/standard_quarter_margin"
-            android:contentDescription="@string/favorite"
-            android:src="@drawable/favorite" />
+        </RelativeLayout>
 
         <ImageView
             android:id="@+id/localFileIndicator"
             android:layout_width="@dimen/list_item_local_file_indicator_layout_width"
```
tobiasKaminsky, line 70: Here you use a fixed with, but then use wrap_content for height.
This could, in theory, lead to a too high image, or?
Maybe it is better to re-use the old "@dimen/list_item_local_file_indicator_layout_height" ?

Shagequi: In theory, it's not necessary to define a fixed height because the ratio defined by the SVG will always be respected. In addition, this SVG is square-shaped.

But, but, to keep the code uniformity, I'm going to set the height with a dimen ;)

Shagequi: Note that I mistook the name of the commit, it's definitely the height that I edited :-)

Issue #5866, opened 2020-04-14
Intents are not type safe


tobiasKaminsky:
As we had this recently: https://github.com/nextcloud/android/pull/5857

To prevent this, my idea is to have a static "launch" function in ever needed activity, that handles creating the intent correctly.
In this case it could be:
```
public static function launch/createIntent(File file, Account account) {
Intent i = new Intent(fileActivity, ConflictsResolveActivity.class);
i.putExtra(ConflictsResolveActivity.EXTRA_FILE, file);
i.putExtra(ConflictsResolveActivity.EXTRA_ACCOUNT, account);

return i;
}
```

What do you think? @ezaquarii @AndyScherzinger

ezaquarii:
@tobiasKaminsky Yes, this is analogous to `Fragment.newInstance()` approach and works well.

AndyScherzinger:
Yes, sound good to me too 👍

Issue #4205, opened 2019-07-06
Sticky broadcast giving security exception in Android for work in Android P


rumbleball80:
### Actual behaviour
- I installed Nextcloud from Play Store in an Android for Work environment. After login and getting access to filesystem the app crashes
- every retry to open the app ends directly in an crash 

### Expected behaviour
- It should launch the application correctly
 
### Steps to reproduce
1. Configure Android for Work
2. Install Nextcloud android app and configure it
3. Launch it one time. 

### Environment data
Android version: 9 patch version 05/05/2019

Device model: HUAWEI P20 lite

Stock or customized system: stock

Nextcloud app version: 3.6.2

Nextcloud server version: 15.0.9

#### Web server error log
Nothing

#### Nextcloud log (data/nextcloud.log)

> 07-05 21:28:46.172 18341 18341 W InputMethodManager: startInputReason = 1
> 07-05 21:28:46.177 18341 18341 D FileDisplayActivity: Received broadcast com.owncloud.android.syncadapter.FileSyncAdapter.EVENT_FULL_SYNC_START
> 07-05 21:28:46.179  1138  3469 W ActivityManager: Permission Denial: unbroadcastIntent() from pid=18341, uid=1110168 requires android.permission.BROADCAST_STICKY
> 07-05 21:28:46.179  1138  3469 W System.err: java.lang.SecurityException: Permission Denial: unbroadcastIntent() from pid=18341, uid=1110168 requires android.permission.BROADCAST_STICKY
> 07-05 21:28:46.181 18341 18341 D AndroidRuntime: Shutting down VM
> 07-05 21:28:46.182 18341 18341 E AndroidRuntime: FATAL EXCEPTION: main
> 07-05 21:28:46.182 18341 18341 E AndroidRuntime: Process: com.nextcloud.client, PID: 18341
> 07-05 21:28:46.182 18341 18341 E AndroidRuntime: java.lang.RuntimeException: Error receiving broadcast Intent { act=com.owncloud.android.syncadapter.FileSyncAdapter.EVENT_FULL_SYNC_START flg=0x10 pkg=com.nextcloud.client (has extras) } in com.owncloud.android.ui.activity.FileDisplayActivity$SyncBroadcastReceiver@4b91581

### Workaround

after searching for "android.permission.BROADCAST_STICKY" I found this:
[...]

PR #6048, opened 2020-05-12
New download manager

ezaquarii:
New download service is a complete rewrite of a legacy `FileDownloader` service.

### Sync code uses old solution - this PR does not replace it as it would be too much work and risk to swallow.

Small comparison of old vs new downloader design:

| Old | New |
| --- | --- |
| Using broadcast intents for state notifications | Plain listeners |
| UI must track broadcast history | No UI state |
| State transferred via type unsafe `Intent`s | Immutable, strongly typed data model |
| No test coverage | Full test coverage |
| Multi-user state management | Every user has private downloader instance |
| Single threaded downloads | Multi-threaded downloads |
| No single source of truth | Single source of truth regarding all downlaods |
| Downloader async logic depends on Android API | Async logic uses generic concurrency API |

### `LiveData<T>` - not used in the service anymore

`LiveData<T>` is an observable value of type `T` that integrates with Android life-cycle mess. It's a relatively new API introduced by Google in `androidx` that is meant to cut the chores of lifecycle management. It works pretty well and the only issue I have with it is that unit testing is a bit more involving.

_Update: I removed dependency on `LiveData<T>` - it was a bad idea o have it there._

### `LocalConnection<S>` and `LocalBinder<S>`

`LocalConnection` and `LocalBinder` are generic implementations of platform `ServiceConnection` and `Binder` that everybody re-implements over and over again in the same form. It also provies a good dumping ground for typed service interaction API in general, so we can address https://github.com/nextcloud/android/issues/5866 for services this way. We already used this approach in [PlayerServiceConnection.kt](https://github.com/nextcloud/android/blob/9876c49001495dcc64b59f699d3f4bd7295f87eb/src/main/java/com/nextcloud/client/media/PlayerServiceConnection.kt) (extracted from legacy Java inline implementation).

Apart from that, we need a connection object as a facade for `Service`, because `Service` instance is created asynchonously and might (will) not be available when UI wants to interact with it.

### `Downloader` interface

I managed to extract public downloader API and calls can be nicely delegated through all layers -> connector -> binder -> user downloader processor. All layers use the same interface.

### `DownloaderConnection`

This component expsoses very vanilla API to the underlying Android `Service` mess, hiding all asynchronous issues behind a boring interface.

Updates are delivered using vanilla listener API (`addListener()`/`removeListener()`, so every
developer should be familiar with them. API solves problematic missed updates issue caused by asynchronous start/binding races, so no more any need for messaging bus pattern to deliver updates reliably.

If we'd like to expose `LiveData<T>` interface to the UI, this is the place to do it.

### `DownloaderService`

This is a proposed replacement for `FileDownloader` service. Apart from cleaner code, it will
keep per-user `Downloader` instance and automatically terminate when all downloads are finished.

### `AsyncRunner`

`AsyncRunner` is a replacement for `AsyncTask` that uses generic Java concurrency API and does not require any subclassing (tasks are just lambdas or functions so can be tested easily, contrary to `AsyncTask` being virtually untestable piece of global state). Also, `AsyncTask` API is officially deprecated by Google in favour of `java.util.concurrent' APIs.

This PR extends the API by adding progress and task cancelation, as those will be needed to provide adequate UX.

### MockK

`MockK` is a `Mockito`, but in Kotlin. I introduced it because I find `Mockito` unable to mock Kotlin classes on Dalvik (it works in `test`, but not in `androidTest`). There is some weird issue with bytecode generator and I'm still investigating it. `MockK` has some nice properties, like support for co-routines, but the community behind it is smaller. `Mockito` is more mature, but the android support seems to still be rather poorly maintained. I do not commit myself to use it - I needed something to move forward with unit tests.

### DexOpener

Mocking final classes Android below P (API <= 27) is not possible, which makes testing very troublesome. Android 28+ has out-of-the-box support for mocking, but our CI runs tests on older versions.

DexOpener is installed by instrumentation test runner and removes final attribute, enabling mocking of Kotlin classes.

### `ContactListFragment` - proof of concept use

Contacts imports has been migrated to a new API. To test it:

1) backup your contacts
2) Settings->Show hidden files
3) navigate to contacts backup dir and delete backup locally
4) go to Settings->Back up contacts->Choose date

Contacts will be downloaded using proposed downloader service.

### `MockUser`

Sadly, using mockito to mock `User` interface will blow up when actually trying to parcel it into Intent (mock does no contain proper `Parcelable` implementation, obviously).

This litle helper provides a true implementation mock to be used in connected tests.

### `EtmDownloaderFragment` and test mode

ETM page to test the downloader. Any download request can be scheduled with a `test` flag `true` - this triggers "fake" download task that just `Thread.sleep()` and pushes some progress. One can use it for logic testing or UI work.

### `NotificationsManager`

Downloader service must be switched to foreground and that requires a notifcation. I ported some notification code that is smeared in `FileDownloader` and placed it behind an interface.

Notifications requires some more work, as the UX is not great - it contains too much redundant text and not enough interesting information. Also, new downloader architecture allows us potentially to show multiple downloads, download transfer rate, number of pending files, etc, etc... We can do more.

### DIFF
--- a/src/main/java/com/nextcloud/client/files/downloader/DownloadTask.kt
+++ b/src/main/java/com/nextcloud/client/files/downloader/DownloadTask.kt
@@ -1,0 +1,100 @@
+/*
+ * Nextcloud Android client application
+ *
+ * @author Chris Narkiewicz
+ * Copyright (C) 2020 Chris Narkiewicz <hello@ezaquarii.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.nextcloud.client.files.downloader
+
+import android.content.ContentResolver
+import android.content.Context
+import com.nextcloud.client.core.IsCancelled
+import com.owncloud.android.datamodel.FileDataStorageManager
+import com.owncloud.android.datamodel.OCFile
+import com.owncloud.android.lib.common.OwnCloudClient
+import com.owncloud.android.operations.DownloadFileOperation
+import com.owncloud.android.utils.MimeTypeUtil
+import java.io.File
+
+/**
+ * This runnable object encapsulates file download logic. It has been extracted to wrap
+ * network operation and storage manager interactions, as those pose testing challenges
+ * that cannot be addressed due to large number of dependencies.
+ *
+ * This design can be regarded as intermediary refactoring step.
+ */
+class DownloadTask(
+    val context: Context,
+    val contentResolver: ContentResolver,
+    val clientProvider: () -> OwnCloudClient
+) {
+
+    data class Result(val file: OCFile, val success: Boolean)
+
+    /**
+     * This class is a helper factory to to keep static dependencies
+     * injection out of the downloader instance.
+     *
+     * @param context Context
+     * @param clientProvider Provide client - this must be called on background thread
+     * @param contentResolver content resovler used to access file storage
+     */
+    class Factory(
+        private val context: Context,
+        private val clientProvider: () -> OwnCloudClient,
+        private val contentResolver: ContentResolver
+    ) {
+        fun create(): DownloadTask {
+            return DownloadTask(context, contentResolver, clientProvider)
+        }
+    }
+
+    fun download(request: Request, progress: (Int) -> Unit, isCancelled: IsCancelled): Result {
+        val op = DownloadFileOperation(request.user.toPlatformAccount(), request.file, context)
+        val client = clientProvider.invoke()
+        val result = op.execute(client)
+        if (result.isSuccess) {
+            val storageManager = FileDataStorageManager(
+                request.user.toPlatformAccount(),
+                contentResolver
+            )
+            val file = saveDownloadedFile(op, storageManager)
+            return Result(file, true)
+        } else {
+            return Result(request.file, false)
+        }
+    }
+
+    private fun saveDownloadedFile(op: DownloadFileOperation, storageManager: FileDataStorageManager): OCFile {
+        val file = storageManager.getFileById(op.getFile().getFileId()) as OCFile
+        val syncDate = System.currentTimeMillis()
+        file.lastSyncDateForProperties = syncDate
+        file.lastSyncDateForData = syncDate
+        file.isUpdateThumbnailNeeded = true
+        file.modificationTimestamp = op.getModificationTimestamp()
+        file.modificationTimestampAtLastSyncForData = op.getModificationTimestamp()
+        file.etag = op.getEtag()
+        file.mimeType = op.getMimeType()
+        file.storagePath = op.getSavePath()
+        file.fileLength = File(op.getSavePath()).length()
+        file.remoteId = op.getFile().getRemoteId()
+        storageManager.saveFile(file)
+        if (MimeTypeUtil.isMedia(op.getMimeType())) {
+            FileDataStorageManager.triggerMediaScan(file.storagePath)
+        }
+        return file
+    }
+}
--- a/src/main/java/com/owncloud/android/ui/fragment/contactsbackup/ContactListFragment.java
+++ b/src/main/java/com/owncloud/android/ui/fragment/contactsbackup/ContactListFragment.java
@@ -4,6 +4,7 @@
  * @author Tobias Kaminsky
  * Copyright (C) 2017 Tobias Kaminsky
  * Copyright (C) 2017 Nextcloud GmbH.
+ * Copyright (C) 2020 Chris Narkiewicz <hello@ezaquarii.com>
  * <p>
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Affero General Public License as published by
@@ -22,11 +23,9 @@
 package com.owncloud.android.ui.fragment.contactsbackup;
 
 import android.Manifest;
-import android.content.BroadcastReceiver;
+import android.app.Activity;
 import android.content.Context;
 import android.content.DialogInterface;
-import android.content.Intent;
-import android.content.IntentFilter;
 import android.database.Cursor;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
@@ -58,12 +57,15 @@ import com.google.android.material.snackbar.Snackbar;
 import com.nextcloud.client.account.User;
 import com.nextcloud.client.account.UserAccountManager;
 import com.nextcloud.client.di.Injectable;
+import com.nextcloud.client.files.downloader.Download;
+import com.nextcloud.client.files.downloader.DownloadState;
+import com.nextcloud.client.files.downloader.DownloaderConnection;
+import com.nextcloud.client.files.downloader.Request;
 import com.nextcloud.client.jobs.BackgroundJobManager;
 import com.nextcloud.client.network.ClientFactory;
 import com.owncloud.android.R;
 import com.owncloud.android.datamodel.FileDataStorageManager;
 import com.owncloud.android.datamodel.OCFile;
-import com.owncloud.android.files.services.FileDownloader;
 import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.ui.TextDrawable;
 import com.owncloud.android.ui.activity.ContactsPreferenceActivity;
@@ -101,6 +103,7 @@ import butterknife.ButterKnife;
 import ezvcard.Ezvcard;
 import ezvcard.VCard;
 import ezvcard.property.Photo;
+import kotlin.Unit;
 
 import static com.owncloud.android.ui.fragment.contactsbackup.ContactListFragment.getDisplayName;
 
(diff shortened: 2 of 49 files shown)


ezaquarii:
@tobiasKaminsky @AndyScherzinger Please take a look and let me know what you think.

AndyScherzinger:
I don't quite understand all the new concepts/structures like live data but the code/structure look good to me 👍

ezaquarii:
> I don't quite understand all the new concepts/structures like live data but the code/structure look good to me +1

@AndyScherzinger 

### `LiveData<T>`

`LiveData<T>` is an observable value of type `T` that integrates with Android life-cycle mess. It's a relatively new API introduced by Google in `androidx` that is meant to cut the hores of lifecycle management. It works pretty well and the only issue I have with it is that unit testing is a bit more involving.

### `LocalConnection<S>` and `LocalBinder<S>`

`LocalConnection` and `LocalBinder` are generic implementations of platform `ServiceConnection` and `Binder` that everybody re-implements over and over again in the same form. It also provies a good dumping ground for typed service interaction API in general, so we can address https://github.com/nextcloud/android/issues/5866 for services this way. We already used this approach in [PlayerServiceConnection.kt](https://github.com/nextcloud/android/blob/9876c49001495dcc64b59f699d3f4bd7295f87eb/src/main/java/com/nextcloud/client/media/PlayerServiceConnection.kt) (extracted from legacy Java inline implementation).

Apart from that, we need a connection object as a facade for `Service`, because `Service` instance is created asynchonously and might (will) not be available when UI wants to interact with it.

### `AsyncRunner`

`AsyncRunner` is a replacement for `AsyncTask` that uses generic Java concurrency API and does not require any subclassing (tasks are just lambdas or functions so can be tested easily, contrary to `AsyncTask` being virtually untestable piece of global state). Also, `AsyncTask` API is officially deprecated by Google in favour of `java.util.concurrent' APIs.

This PR extends the API by adding progress and task cancelation, as those will be needed to provide adequate UX.

### MockK

`MockK` is a `Mockito`, but in Kotlin. I introduced it because I find `Mockito` unable to mock Kotlin classes on Dalvik (it works in `test`, but not in `androidTest`). There is some weird issue with code bytecode generator and I'm still investigating it. `MockK` has some nice properties, like support for co-routines, but the community behind it is smaller. `Mockito` is more mature, but the android support seems to still be rather poorly maintained.

ezaquarii:
@AndyScherzinger @tobiasKaminsky Partial rewrite to eliminate some issues I found when trying to use it in contacts import. I have updated the changes description as well: https://github.com/nextcloud/android/pull/6048#issuecomment-627540662

ezaquarii:
@tobiasKaminsky @AndyScherzinger size of the diff surprised me, but actually ~60% of it are tests. Remaining 40% is also blown a bit by the KDoc/JavaDoc so no worries. This should not be hard to review.

ezaquarii:
@tobiasKaminsky @AndyScherzinger ready for review

ezaquarii:
Getting rid of broadcast-based progress updates is required to solve issues similar to this one: https://github.com/nextcloud/android/issues/4205

src/main/java/com/nextcloud/client/files/downloader/DownloadTask.kt
```diff
@@ -0,0 +1,103 @@
...
+        fun create(): DownloadTask {
+            return DownloadTask(context, contentResolver, clientProvider)
+        }
+    }
+
+    fun download(request: Request, progress: (Int) -> Unit, isCancelled: IsCancelled): Boolean {
+        val op = DownloadFileOperation(
+            request.user.toPlatformAccount(),
+            request.file,
+            OCFileListFragment.DOWNLOAD_SEND,
```
tobiasKaminsky, line 68: This and the next 2 lines is only needed in one special case (FDA:2255), but it should not be hardcoded.

src/main/java/com/owncloud/android/ui/fragment/contactsbackup/ContactListFragment.java
```diff
@@ -497,19 +503,16 @@ public int hashCode() {
...
-                ocFile = storageManager.getFileByPath(downloadedRemotePath);
-                loadContactsTask.execute();
-            }
+    private Unit onDownloadUpdate(Download download) {
+        final Activity activity = getActivity();
+        if (download.getState() == DownloadState.COMPLETED && activity != null) {
+            String downloadedRemotePath = download.getRequest().getFile().getRemotePath();
+            FileDataStorageManager storageManager = new FileDataStorageManager(user.toPlatformAccount(),
+                                                                               activity.getContentResolver());
+            ocFile = storageManager.getFileByPath(downloadedRemotePath);
```
tobiasKaminsky, line 512: With the new approach, is this roundtrip via FDSM needed, as we already have the OCFile?

tobiasKaminsky:
This is just amazing! This is a giant step forward!
Thanks Chris for all your work :heart:  :heart:  :heart:  :heart:

src/main/java/com/owncloud/android/ui/fragment/contactsbackup/ContactListFragment.java
ezaquarii: `OCFile` in download status is part of the `Request`, so it's not updated.

`DownloadStatus` does not carry result `OCFile`, but this can be changed.

Shagequi:
@ezaquarii @tobiasKaminsky It seems to be great ! Quick question, which approach would you advise about UI/UX for notifications ?

ezaquarii:
@Shagequi I'm happy to update the download notification, but it requires some finishing touch in downloader infrastructure first.

Nevertheless I see no blocker for UX design in the meantime.

src/main/java/com/owncloud/android/ui/fragment/contactsbackup/ContactListFragment.java
ezaquarii: @tobiasKaminsky Downloaded `OCFile` is now exposed in the download status.

src/main/java/com/nextcloud/client/files/downloader/DownloadTask.kt
ezaquarii: The thing is that `DownloadFileOperations` does not use those values for anything.
It seems to me that it just carries them around as part `FileDownloader` state.

Since my downloader implementation does not support sending downloaded files anywhere at this moment, we don't need those values.

I'm also pretty sure that we won't use them in the future, as this state simply doesn't belong here. If or when we implement sending files to activity, I'm sure it will be structured differently.

At first glance, information about post-download actions should be carried by `Request` and actioned in `DownloaderService` when downloader signals completion or failure.

However, it's pretty pointless to put those hardcoded values here, I'll add an overloaded constructor for this case.

ezaquarii: @tobiasKaminsky Let me know if you'd like to migrate more use cases to the downloader as part of this PR - in such case, we'll need to add this functionality. If this is not the case, I think we can skip it. I'm confident that the design is readu to support this feature with minimal effort.

ezaquarii:
It looks like mocks are not initialized when tests are run in AVD.

ezaquarii:
@tobiasKaminsky Test failure on emulator has been resolved. Mocking doesn't work on API 27 and below due to Kotlin classes being `final` by default.

tobiasKaminsky:
I will have a deeper look into it next week.
If an user is downloading a very big file, this might be killled in background, or?
And we do not (yet) have a way to resume it, or?

ezaquarii:
> I will have a deeper look into it next week.
> If an user is downloading a very big file, this might be killled in background, or?

Any long-runnning operation can be terminated by the OS without notice, including download. The new implementation - same as the old one - uses foreground service, so that the OS will try hard to keep it alive, but it is not guaranteed.

> And we do not (yet) have a way to resume it, or?

In case of an individual file, the behavior should be the same as in current `FileDownloader`, because it is based on the same  `DownloadFileOperation` component.

If we talk about the download queue persistance, this is not implemented by the downloader and I believe that existing `FileDownloader` doesn't implement such functionality either. Queue persistance is done using database and some sort of restart helper brings up the downloader service back to life after app restart, but I'm not familiar with details.

ezaquarii:
@tobiasKaminsky I have noticed that dependabot caused multiple sorce reformatting commits in recent time.

This messes up rebases, as I had to resolve conflicts caused by meaningless formatting changes.
More importantly tho, it pollited history of changes with completely meaningless changes.

Can we reconsider it's use?

AndyScherzinger:
@ezaquarii you might refer to this one in particular https://github.com/nextcloud/android/pull/6180/commits while the reformatting haven't come from the bot itself but the code changes to make the lib upgrade(s) work.

ezaquarii:
@AndyScherzinger @tobiasKaminsky Ah, ok, I was confused by the merger ID (dependabot). I have no issue with concious manual changes at all.

AndyScherzinger:
pinging @tobiasKaminsky

tobiasKaminsky:
@ezaquarii 
> ViewModel classes should never point to Views or non-application Contexts.

This is for https://github.com/nextcloud/android/blob/651801a98302e9def64230d956fa38515dd4a87e/src/main/java/com/nextcloud/client/etm/EtmViewModel.kt#L49

From my understing it would help to move downloaderConnection to EtmDownloaderFragment?

ezaquarii:
> @ezaquarii
> 
> > ViewModel classes should never point to Views or non-application Contexts.

And it does not, AFAIK. Context injected by our DI is actually `Application` context in all cases.

> This is for
> 
> https://github.com/nextcloud/android/blob/651801a98302e9def64230d956fa38515dd4a87e/src/main/java/com/nextcloud/client/etm/EtmViewModel.kt#L49
> 
> From my understing it would help to move downloaderConnection to EtmDownloaderFragment?

Then we'd move non-GUI component into GUI layer, which goes against VM purpose. Although I have to admint that the `Context` is not entirely desirable here, as ideally we'd inject the connection. It's the current user that makes things a bit inconvenient.

IMO not great but not terrible either.

ezaquarii:
@tobiasKaminsky rebased on top of master

tobiasKaminsky:
@ezaquarii unit tests with TestEtmViewModel fail…

ezaquarii:
@tobiasKaminsky test should be ok now

PR #6054, opened 2020-05-12
Update folder icons

AndyScherzinger:
as a follow-up to #5966, replace all folder icon variants with a material like version that is based on the material design folder icon.


tobiasKaminsky:
Awesome!

PR #6057, opened 2020-05-12
Remove spot bug type DM_BOXED_PRIMITIVE_FOR_COMPARE

tobiasKaminsky:
Tried to remove it.
Then found out that Long.compare(…) is API 19+.

[ ] Let us see if this is recognized

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/utils/FileSortOrderByDate.java
+++ b/src/main/java/com/owncloud/android/utils/FileSortOrderByDate.java
@@ -45,7 +45,7 @@ public class FileSortOrderByDate extends FileSortOrder {
         final int multiplier = isAscending ? 1 : -1;
 
         Collections.sort(files, (o1, o2) ->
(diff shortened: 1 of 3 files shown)


tobiasKaminsky:
@ezaquarii 
I tested this with API17 and it works, but
https://developer.android.com/reference/java/lang/Long?hl=en#compare(long,%20long)
shows that it was only added in API19.

Also Lint does not complain, but I do not understand why this works…?

ezaquarii:
> I tested this with API17 and it works, but
> https://developer.android.com/reference/java/lang/Long?hl=en#compare(long,%20long)
> shows that it was only added in API19.

@tobiasKaminsky 
Runtime might be lacking this API, but the [d8](https://developer.android.com/studio/command-line/d8) might be inlining this trivial call, so it works by accident, because it compares numbers in place and does not call any API at all. One would need to look at the DEX bytecode to be sure.

PR #6065, opened 2020-05-13
Fix selected item background color

sirambd:
Replace blue background color when an item has been selected

nextcloud-gray-select


src/main/res/layout/uploader_list_item_layout.xml
```diff
@@ -19,7 +19,9 @@
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
     android:layout_height="@dimen/standard_list_item_size"
-    android:background="@drawable/list_selector"
+    android:focusable="true"
+    android:clickable="true"
+    android:background="?android:attr/selectableItemBackground"
```
JorisBodin, line 24: Why did you have to add that? There's no effect in this view, I think.

sirambd: I don't know which view it is, but if no effect in this view, we can delete it

JorisBodin: I think is Auto upload :)

sirambd: Okay, i'll go to delete this background and events 😁

sirambd: It's fixed on https://github.com/nextcloud/android/pull/6065/commits/30bc33307e92cfe3f51efd8d28ee2fc8545254ec

tobiasKaminsky:
Code wise :+1: 
@jancborchardt for design :-)

PR #6067, opened 2020-05-13
New design (UI) - Home Appbar

JorisBodin:
Hello,

A big part of the new design is finished.
I would add an animation, but I prefer to propose PR now.

|Light|Dark|
|---|---|
|light|dark|


JorisBodin:
"Lint increased!"
Is fake positive.
Because "Possible overdraw", if remove `android:background="@color/bg_default"` in AppBarLayout, the AppBarLayout take "primaryColor" of theme. But primaryColor is blue ...

AndyScherzinger:
> "Lint increased!"
Is fake positive.

@JorisBodin can you add an exclusion for this one? 

Code looks good 👍
Will give it a test drive 🚀

JorisBodin:
> @JorisBodin can you add an exclusion for this one?

@AndyScherzinger How do I add it?

AndyScherzinger:
@JorisBrodin I'd have to check, else just raise the score in the lint file in the scripts, analysis folder :)

szaimen:
I think the white border around the search field looks strange.
This is how gmail is doing it:

JorisBodin:
@AndyScherzinger you can change this for me? 🙏

JorisBodin:
> Nice to see everything coming together!
> 
> I think the white border around the search field in dark mode looks strange.
> This is how google is doing it in the gmail app:
> 

@szaimen How do you enable dark mode in gmail? May be only available on android 10
There is no border either in light with gmail

AndyScherzinger:
@JorisBodin I guess the border is rather a glow effect and if you zoom into the image you also see that the dark version has a dark glow effect.

JorisBodin:
> @JorisBodin I guess the border is rather a glow effect and if you zoom into the image you also see that the dark version has a dark glow effect.

@AndyScherzinger There's a higher elevation. I'm trying to...

JorisBodin:
It's a different style. I prefer
Screenshot_1589443530
Screenshot_1589443522

AndyScherzinger:
> It's a different style. I prefer

Looks great @JorisBodin !

Only thing that seems of (also saw this on another view) is that the PDF icon should be razer-sharp (since it is  a xml-drawable) but for some reason it isn't ?!

JorisBodin:
> > It's a different style. I prefer
> 
> Looks great @JorisBodin !
> 
> Only thing that seems of (also saw this on another view) is that the PDF icon should be razor-sharp (since it is a xml-drawable) but for some reason it isn't ?!

All files icons are blurred. Even on stable

JorisBodin:
And if we want copy google...

|Light|Dark|
|---|---|
|Screenshot_1589444150|Screenshot_1589444140|

src/main/res/animator/appbar_elevation_off.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tobiasKaminsky, line 1: All new files need to have a license header, gpl3+?

tobiasKaminsky:
All new files need to have a license header, gpl3+?

Otherwise this is awesome!

tobiasKaminsky:
One thing, which we can do in a later PR:
when going back to main view (FDA), it shows shortly a solid blue round circle and then loads the avatar.
This should be changed to the default avatar icon…

tobiasKaminsky:
@JorisBodin if you go into a subfolder then I see the old toolbar. 
Will this be fixed with an upcoming PR?

szaimen:
> @JorisBodin if you go into a subfolder then I see the old toolbar. 
> Will this be fixed with an upcoming PR?

The same is happening if you swipe to refresh...

JorisBodin:
> All new files need to have a license header, gpl3+?

@tobiasKaminsky  It's done

> One thing, which we can do in a later PR:
> when going back to main view (FDA), it shows shortly a solid blue round circle and then loads the avatar.
> This should be changed to the default avatar icon…

Yes i see it. Bug i reused the same methode in drawer. With glide it'd be so much easier.



> @JorisBodin if you go into a subfolder then I see the old toolbar.
> Will this be fixed with an upcoming PR?

It's normal behavior. The new appbar is only at the root. So, it would have been easier to implement everywhere.

> > Will this be fixed with an upcoming PR?
> 
> The same is happening if you swipe to refresh...

@szaimen Yes? ... 
And where? How?...

szaimen:
> @szaimen Yes? ... 
> And where? How?...

When you are in the root directory:

JorisBodin:
> > @szaimen Yes? ...
> > And where? How?...
> 
> When you are in the root directory:
> Screen_Recording_20200514-142133_1_1_1

You don't have the latest version...
The bar isn't identical to google.

szaimen:
> > When you are in the root directory:
> 
> You don't have the latest version...
> The bar isn't identical to google.
> 
> You can test with last commit plz?

I would like to but I need to wait for a newer apk from the nextcloud-bot.

szaimen:
> > @JorisBodin if you go into a subfolder then I see the old toolbar.
> > Will this be fixed with an upcoming PR?
> 
> The same is happening if you swipe to refresh...

@JorisBodin That doesn't happen anymore 👍

PR #6105, opened 2020-05-19
New design (UI) - User info activity

JorisBodin:
I've also fix the navigationbar color in all activities.

|Light|Dark|
|---|---|
|Screenshot_1589900575|Screenshot_1589898565|


src/main/java/com/owncloud/android/ui/activity/UserInfoActivity.java
```diff
@@ -151,7 +150,7 @@ public void onCreate(Bundle savedInstanceState) {
         if (actionBar != null) {
             actionBar.setDisplayHomeAsUpEnabled(true);
             actionBar.setDisplayShowHomeEnabled(true);
-            actionBar.setTitle("");
+            ThemeUtils.tintBackButton(actionBar, this);
```
AndyScherzinger, line 153: shouldn't we be able to set the backbutton via the theme since we don't have server-depending theming for the actionbar anymore?

JorisBodin: @AndyScherzinger It's true
But there's a lot of modification. You need a dedicated PR refactor for that.

PR #6107, opened 2020-05-19
Fix upload and wrote tests

tobiasKaminsky:
This is only in dev/master, so no backport needed.
:+1:  for using dev myself and suffering from it ;-)
 
### Testing 

### DIFF
--- a/src/androidTest/java/com/owncloud/android/UploadIT.java
+++ b/src/androidTest/java/com/owncloud/android/UploadIT.java
@@ -27,12 +27,16 @@ import com.nextcloud.client.device.PowerManagementService;
 import com.nextcloud.client.network.Connectivity;
 import com.nextcloud.client.network.ConnectivityService;
 import com.owncloud.android.datamodel.UploadsStorageManager;
-import com.owncloud.android.datamodel.OCFile;
 import com.owncloud.android.db.OCUpload;
+import com.owncloud.android.files.services.FileUploader;
+import com.owncloud.android.lib.common.operations.RemoteOperationResult;
+import com.owncloud.android.operations.RefreshFolderOperation;
 import com.owncloud.android.operations.RemoveFileOperation;
+import com.owncloud.android.operations.UploadFileOperation;
 import com.owncloud.android.utils.FileStorageUtils;
 
 import org.jetbrains.annotations.NotNull;
+import org.junit.After;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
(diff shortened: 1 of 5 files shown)


src/androidTest/java/com/owncloud/android/UploadIT.java
```diff
@@ -183,4 +184,151 @@ public void testUploadInNonExistingFolder() {
...
+        newUpload.setRemoteFolderToBeCreated();
+        newUpload.addRenameUploadListener(() -> {
+            // dummy
+        });
+
+        RemoteOperationResult result = newUpload.execute(client, getStorageManager());
+        assertTrue(result.toString(), result.isSuccess());
+
+        // cleanup
+        new RemoveFileOperation("/testUpload/", false, account, false, targetContext).execute(client, getStorageManager());
```
ezaquarii, line 262: what if that fails? can the test be repeated again?

src/androidTest/java/com/owncloud/android/UploadIT.java
```diff
@@ -183,4 +184,151 @@ public void testUploadInNonExistingFolder() {
...
+            public BatteryStatus getBattery() {
+                return new BatteryStatus(true, 100);
+            }
+        };
+
+        OCUpload ocUpload = new OCUpload(FileStorageUtils.getSavePath(account.name) + "/empty.txt",
+                                         "/testUpload/charging.txt", account.name);
+        ocUpload.setWhileChargingOnly(true);
+
+        UploadFileOperation newUpload = new UploadFileOperation(
```
ezaquarii, line 240: This setup seems to be repetive. We could do it in:

```
@Before
void setUp() { ... }
```

ezaquarii:
```
com.owncloud.android.util.EncryptionTestIT > testCryptFileWithoutMetadata[android-27(AVD) - 8.1.0] FAILED
--
(...) Instrumentation process crashed.
```

src/androidTest/java/com/owncloud/android/UploadIT.java
tobiasKaminsky: If this fails something is seriously broken and then I think it is ok that this (or any following) test fails.

src/androidTest/java/com/owncloud/android/UploadIT.java
tobiasKaminsky: It is, and for unmodified uploads I have "testUpload(…)".
But the 4 new added tests have slightly modifications:
- use special connectivtyServiceMock / powerManagementServiceMock
- use different values for onWifiOnly/whileChargingOnly

So for better readability I would keep it this way.

PR #6172, opened 2020-05-28
fix search related items in drawer

tobiasKaminsky:
On latest master search related items were gone.

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/utils/DrawerMenuUtil.java
+++ b/src/main/java/com/owncloud/android/utils/DrawerMenuUtil.java
@@ -31,6 +31,7 @@ import com.owncloud.android.R;
 import com.owncloud.android.lib.resources.status.OCCapability;
 import com.owncloud.android.lib.resources.status.OwnCloudVersion;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 /**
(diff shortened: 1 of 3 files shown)


src/main/java/com/owncloud/android/utils/DrawerMenuUtil.java
```diff
@@ -42,31 +42,26 @@ private DrawerMenuUtil() {
...
-        } else if (!user.isAnonymous()) {
-            filterMenuItems(menu, R.id.nav_recently_added, R.id.nav_recently_modified, R.id.nav_videos);
+        if (!resources.getBoolean(R.bool.videos_enabled)) {
+            menu.removeItem(R.id.nav_videos);
         }
     }
 
     public static void filterTrashbinMenuItem(Menu menu, User user, @Nullable OCCapability capability) {
-        if (!user.isAnonymous() &&
+        if (!user.isAnonymous() ||
```
AndyScherzinger, line 64: sure?

AndyScherzinger: wouldn't it have to be 
```suggestion
        if (user.isAnonymous() ||
```

As in remove menu item in case user _is_ anonymous?

tobiasKaminsky: I guess we can just remove this.
It was previously needed as we had the check based on account, as we needed to check if the server (which is identified via account) is capable of search.

But as nowadays every supported NC server has search support, this is always true and thus we can remove also this check.
(If a user is anonymous it means that the app is not logged in, so it should not show this screen at all…)

I will do it on Tuesday.

ezaquarii: anonymous user is the old `account == null`, which in turn was ususally a retrofitted workaround for a NPE bug

PR #6244, opened 2020-06-08
black nav bar on image preview

### LINKED ISSUE #6242: preview image: white navigation bar

tobiasKaminsky:
Fix #6242 

### Testing 


AndyScherzinger:
@tobiasKaminsky afaik the navigationbar color can be defined v21+ so why is this qadded to a v27 style?

jancborchardt:
It's black, but also half transparent in the image viewer just like the header bar, right? From the code "translucent" it looks like that is the case, can you confirm?

src/main/res/values-v21/styles.xml
```diff
@@ -75,4 +75,9 @@
         <item name="android:windowBackground">@drawable/launch_screen</item>
         <item name="android:textColorHint">@color/secondary_text_color</item>
     </style>
+
+    <style name="Theme.ownCloud.Overlay" parent="Theme.ownCloud.OverlayBase">
```
tobiasKaminsky, line 79: So on <v21 what is then used? Parent of Overlay? (never did this before :see_no_evil: )

tobiasKaminsky:
> It's black, but also half transparent in the image viewer just like the header bar, right? From the code "translucent" it looks like that is the case, can you confirm?

You can try it yourself by installing the QA build :-)

AndyScherzinger:
> You can try it yourself by installing the QA build :-)

which then leads to the test result that the bar isn't translucent... but black :/

tobiasKaminsky:
I changed black to transparent, but then it is white…


When toggling fullscreen it becomes transparent:
 

@AndyScherzinger if you have time to fiddle around with this, it would be awesome!

AndyScherzinger:
> if you have time to fiddle around with this, it would be awesome!

Not sure if I find the time @tobiasKaminsky so I'd vote for merging as is and opening a follow-up issue since this (partial) fix is still an improvement 👍

Issue #6247, opened 2020-06-08
Make use of View Binding


stefan-niedermann:
As an intermediate step to #3146 i propose to use [View Binding](https://developer.android.com/topic/libraries/view-binding) (this is **not** Data Binding).

- It will save you *lots* of code
- It is recommended
- It provides compile time null safety and type checks
- It can be used partially to the current used `findViewById()`-method, no big bang required
- It matches the current technology stack and will be an easy first step to a cleaner architecture

I think Data Binding is still interesting, but it requires huge efforts to make it work nicely everywhere, while View Binding is the "little brother" - better than the current state but way easier to switch compared to Data Binding.

I'd like to hear your opinions @tobiasKaminsky @AndyScherzinger @ezaquarii 

If everyone agrees i could create a initial Pull Request to add support in a few cases to demonstrate it on a concrete sample.

AndyScherzinger:
@stefan-niedermann afaik we already decided to go for view binding replace the now deprecated Butterknife usage 😉 

cc @tobiasKaminsky @ezaquarii

AndyScherzinger:
@stefan-niedermann I also think @tobiasKaminsky already implemented this in 1-2 cases in the app :)

tobiasKaminsky:
Yes, we want to do this.
But it is some sort of time tradeoff, so at least I want to do it only for new code or code that I touch.
Changing existing code without enhancing anything, I do not have time for this.

But if you are up to this, feel free to do it :+1: 

As this is discussed, I will close it.

ezaquarii:
@stefan-niedermann
CC @tobiasKaminsky 

The value of migrating existing code to view bindings is IMO limited, as the view bindings solve a trivial problem. It buys minimal amount of code reduction for a price of extra dependency on non-trivial compiler tooling that can be poorly maintained, buggy or just abandoned by it's maintainers. 
Plus one needs to understand this piece of magic to work with it. Value `null` safety is a value-added proposition, but this class of bugs is not very pressing concern in that segment of code and are trivial to spot as well.

The platform's API `findViewById(int)` is simple, well understood (one doesn't sweat over a trivial getter) and doesn't require extra compiler toolchain to work.

[...]

PR #6249, opened 2020-06-08
#6247 View Binding - ShareFileFragment

stefan-niedermann:
This is the first of a few PR i would like to provide for #6247.

It
- replaces `findViewById(R.id.xyz)` calls with `binding.xyz`
- replaces `getView()` calls with `binding.getRoot()` calls
- replaces `getActivity()` calls with `requireActivity()` calls

### DIFF
--- a/src/main/java/com/owncloud/android/ui/fragment/ShareFileFragment.java
+++ b/src/main/java/com/owncloud/android/ui/fragment/ShareFileFragment.java
@@ -31,17 +31,12 @@ import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
-import android.widget.Button;
 import android.widget.CompoundButton;
-import android.widget.ImageView;
-import android.widget.LinearLayout;
-import android.widget.ListView;
-import android.widget.ScrollView;
-import android.widget.TextView;
 
 import com.google.android.material.button.MaterialButton;
 import com.google.android.material.snackbar.Snackbar;
 import com.owncloud.android.R;
+import com.owncloud.android.databinding.ShareFileLayoutBinding;
 import com.owncloud.android.datamodel.OCFile;
 import com.owncloud.android.datamodel.ThumbnailsCacheManager;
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/ui/fragment/ShareFileFragment.java
```diff
@@ -56,6 +51,8 @@
 import com.owncloud.android.utils.MimeTypeUtil;
 import com.owncloud.android.utils.ThemeUtils;
 
+import org.jetbrains.annotations.NotNull;
```
AndyScherzinger, line 54: sure we need a jetbrains annotation?

stefan-niedermann: Will have a look, got a lint error that a overiding method didn't add a `@NonNull` annotation

AndyScherzinger: Sweet, while `@NonNull` and `@NotNull` are 2 different annotations :D

ezaquarii:
@stefan-niedermann Could you please provide a roadmap for this refactoring? You mentioned that this is first of a few PRs.

The reason I'm asking is that there are some things that I believe could be addressed, but maybe you already plan to implement them as followup PR?

I'd like ot better understand the context of your changes in this PR.

stefan-niedermann:
I will provide PR for files which i changed - with a sane PR size. So this was a big file -> 1 PR, if i have a few smaller files, i will add a few files to 1 PR.

I don't want to make one large PR with all changes because this will lead to many merge conflicts and will be terrible to review, therefore step by step.

stefan-niedermann:
> The reason I'm asking is that there are some things that I believe could be addressed

I explicitely wanted to only do View Binding for easier reviews. If i see small linter issues or so, i will fix them also.

But i didn't plan to change anything about the logic, because i don't know it yet and the chances are high that i break something or we get a regression.

tobiasKaminsky:
@stefan-niedermann once it is read to review please set label "2. to review" :)

PR #6284, opened 2020-06-12
Fix appbar photos view when enter in background

JorisBodin:
Actually if we're on the photo view. Exit and return to the application. We had the wrong appbar


tobiasKaminsky:
Ah, but it not closing the app, but suspending it.

tobiasKaminsky:
Rebased and ready to merge

Issue #6277, opened 2020-06-12
Nextcloud App crash on clicking .vcf


psyca:
### Steps to reproduce
1. Create a Backup of your (smartphone) contacts via App
2. Open the hidden .contact folder and click on the created .vcf
3. App crashes

### Expected behaviour
- Tell us what should happen
Open the file, download it or something else but no crash

### Actual behaviour
- Tell us what happens
The Nextcloud App crash

### Environment data
Android version: 9.0

Device model: Samsung Galaxy S8

Stock or customized system: Stock

Nextcloud app version: 3.12.0

Nextcloud server version: 19.0

### Logs
************ CAUSE OF ERROR ************

java.lang.RuntimeException: Unable to start activity ComponentInfo{com.nextcloud.client/com.owncloud.android.ui.activity.ContactsPreferenceActivity}: java.lang.ClassCastException: com.nextcloud.client.account.RegisteredUser cannot be cast to org.parceler.ParcelWrapper
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3114)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3257)
	at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:78)
	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:108)
	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:68)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1948)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loop(Looper.java:214)
	at android.app.ActivityThread.main(ActivityThread.java:7050)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:965)
Caused by: java.lang.ClassCastException: com.nextcloud.client.account.RegisteredUser cannot be cast to org.parceler.ParcelWrapper
	at org.parceler.Parcels.unwrap(Parcels.java:91)
	at com.owncloud.android.ui.activity.ContactsPreferenceActivity.onCreate(ContactsPreferenceActivity.java:96)
	at android.app.Activity.performCreate(Activity.java:7327)
	at android.app.Activity.performCreate(Activity.java:7318)
[...]

PR #6289, opened 2020-06-13
Migrate ContactsPreferenceActivity startup to type-safe API

ezaquarii:
This change addresses #6277 bug in more comprehensible way.
All calls to Context.startActivity(...) starthing
ContactsPreferenceActivity are replaced with type-safe
calls to static ContactsPreferenceActivity.startActivity*
sets of APIs.

No client code should be calling `Intent(..., ContactsPreferenceActivity.class)`
constructor anymore.

See #5866 for discussion about the pattern used in this PR.

- [x] Manual test only: follow #6277 bug reproduction scenario
- [x] Fix `ContactsBackupFragment` us of sidebar extra  - default values are different
- [x] Move `ContactsBackupFragment` to type-safe API as well

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/ContactsPreferenceActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/ContactsPreferenceActivity.java
@@ -22,6 +22,7 @@
  */
 package com.owncloud.android.ui.activity;
 
+import android.content.Context;
 import android.content.Intent;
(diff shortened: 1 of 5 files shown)


AndyScherzinger:
Works fine for me and code looks good too 👍

@tobiasKaminsky why do we launch the contact backup screen when clicking on a vcf file? (out of scope for this PR!)

PR #6301, opened 2020-06-16
Activity mock

tobiasKaminsky:
### Testing 


src/androidTest/java/com/owncloud/android/LocalActivitiesRepository.java
```diff
@@ -0,0 +1,133 @@
...
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+
+import androidx.annotation.NonNull;
+
+import static com.owncloud.android.ui.activity.ContactsPreferenceActivity.TAG;
+
+public class LocalActivitiesRepository implements ActivitiesRepository {
+    String json = "{\n" +
```
ezaquarii, line 49: Kotlin allows to use `"""` to denote raw strings that do not need any escaping. It's really handy for stuff like that and regexps.

src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java
```diff
@@ -119,12 +119,16 @@ public ActivityListAdapter(
         this.storageManager = storageManager;
         this.capability = capability;
         this.clientFactory = clientFactory;
+        try {
+            this.client = clientFactory.createNextcloudClient(currentAccountProvider.getUser());
```
ezaquarii, line 123: Wouldn't that cause `NPE` somewhere down the line?

src/androidTest/java/com/owncloud/android/TestApp.kt
```diff
@@ -0,0 +1,33 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.owncloud.android
+
+
+class TestApp : MainApp() {
```
ezaquarii, line 26: Documentation for this would be helpful, as it apparently offers some specialized, limited functionality regarding full application instance. I'm not sure what capabilities are provided here and what are the design-assumed limitations.

src/main/java/com/owncloud/android/utils/ScreenshotWithServerTest.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotate any screenshot test with this so it is run only when updating/testing screenshots
+ */
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface ScreenshotWithServerTest {
```
ezaquarii, line 36: Why is it called _with server_?

src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -240,6 +238,14 @@ protected void attachBaseContext(Context base) {
         }
     }
 
+    @VisibleForTesting
+    public void setupDagger() {
```
ezaquarii, line 242: We can make it `protected`.

tobiasKaminsky:
> If I understand it correctly, we have a specialized app to conduct screenshot tests, right?
> And we want to swap network I/O layer to eliminate reliance on live server?

Yes.



> Swapping big compinents for tests is the right approach to do it. Ideally, we'd like to swap 1 component to achieve it, but with network ops spread between independent `*Operation` classes, doing it at HTTP client level seems to be a reasonable choice.

I have two approaches…
- This one uses dagger as mocks one specific activitiesRepository.
- Then there is https://github.com/nextcloud/android/pull/6299 which tries to use one generic interceptor to mock any server request.
- another approach would be to have a lightweight (and probably stateless) http server which serves static responses.

My intention was to have a reproducable state, and ideally no external components (e.g. docker) needed (this is true for option 1/2).
Another benefit is that it is currently all handled in code, so there is no context switching for testing.

tobiasKaminsky:
> However, this will be really tedious and fragile to mock, as mock responses would have to track server changes. Of course this could be worked out as well with suitable refactoring, but the question about cost to benefit ratio must be answered first.

True.
My "ideal" plan would look like this:
- in library repo we test against stable/master
- in files app we test against mocked server for most of the operations

So if there are api changes on server they would first be captured by library.

But as you said, I am really uncertain about time effort with this interceptor approach…
Maybe it is first a good start to have "only" #6301 (mocking some parts of the app for testing).

Issue #6334, opened 2020-06-22
Nextcloud locally deletes synced folder without reason


stephanritscher:
- ### Steps to reproduce
1. Sync a sub folder "/folder/sub" to your mobile.
2. Make changes to the folder remotely.
3. Use nextcloud app including visiting the synced folder and using Android file provider to open/edit files in this folder.
4. Repeat similar actions with bad/no connectivity.
(I didn't figure out what exactly triggered the issue, but encountered it repeatedly.)

### Expected behaviour
- All synced files remain on mobile device (in the original or an updated variant)

### Actual behaviour
- From time to time, all the synced files from the subfolder get deleted locally without apparent reason.
- From adb logcat and webserver log, I was able to see, that OfflineSyncJob was sending a PROPFIND for the path "/sub" instead of "/folder/sub".

### Environment data
Nextcloud app version:
3.13.0 Alpha 1
Nextcloud server version:
18.0.4

Thovi98:
I've got this same issue a few times. The last one was this week.
I could'nt take logs unforrunatly...

Android 12L PixelExperience custom rom and the files were on sdcard

joshtrichards:
Fixed in #6335

PR #6335, opened 2020-06-22
Fix improper deleting synced folder by OfflineSyncWork

stephanritscher:
The CheckEtagRemoteOperation (PROPFIND operation) must use the remote path (e.g. folder/sub) instead of the encrypted file name (which is only the last part of the file name, e.g. "sub").
Admittedly, I don't use encryption, so I hope someone with more knowledge of the code or a setup with encryption can review/test this.
This solves #6334 as far as I can tell from testing with my own instance.


AndyScherzinger:
Thanks @stephanritscher for taking the time to analyse and proposing a fix. Looping in @tobiasKaminsky for encryption knowledge 🙏

tobiasKaminsky:
Thanks for finding and fixing the bug :tada:

PR #6364, opened 2020-06-25
Fix NPE in OCFileListFragment when OCFile is null

ezaquarii:
1. Fixed NPE
2. Audited all use of nullable field in all called APIs
3. Updated called APIs nullability documentation

- [x] No test as this is a speculative fix for a tricky edge case + audit changes

### DIFF
--- a/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
+++ b/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
@@ -1300,7 +1300,7 @@ public class OCFileListFragment extends ExtendedListFragment implements
         }
 
         // FAB
-        setFabEnabled(mFile.canWrite());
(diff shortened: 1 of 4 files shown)


src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1330,10 +1330,10 @@ private void setGridSwitchButton() {
      * Determines if user set folder to grid or list view. If folder is not set itself,
      * it finds a parent that is set (at least root is set).
      *
-     * @param folder Folder to check.
+     * @param folder Folder to check or null for root folder
```
tobiasKaminsky, line 1333: Is this really true?
Root folder should also just be a OCFile? 🤔

tobiasKaminsky: Created issue #6388

PR #6358, opened 2020-06-25
add multiple share links

tobiasKaminsky:
@AndyScherzinger as dicussed, would be great if you can polish it a bit :heart: 

Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

### Testing 
Writing tests is very important. Please try to write some tests for your PR. 
If you need help, please do not hesitate to ask in this PR for help.

[unit tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#unit-tests)
[instrumented tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#instrumented-tests)
[UI tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#ui-tests)

- [ ] Tests written, or not not needed

PR #6332, opened 2020-06-22
Enh/dark theme optimizations

AndyScherzinger:
Resolves #6323

edge case improvements (dark/light):
* active drawer menu item contrast
* primary button coloring

**OPEN**
* [x] light-theme+white for `MaterialButton`s
* [x] dark-theme+white for `MaterialButton`s
* [x] `FAB` coloring black/white extreme scenarios for dark/light themes
* [x] probably any `MaterialButton` (maybe add a theming method for buttons handling all edge cases and securing consistent theming, to be done for standard _and_ borderless buttons)
* [x] replace use of `elementColor(...)` with `primaryColor(...)` and drop method

cc @dan0xii for collaboration

### Testing 
Writing tests is very important. Please try to write some tests for your PR. 
If you need help, please do not hesitate to ask in this PR for help.

[unit tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#unit-tests)
[instrumented tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#instrumented-tests)
[UI tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#ui-tests)

- [x] Tests written, or not not needed

PR #6400, opened 2020-07-01
Add theming logic for share link icons

### LINKED ISSUE #6399: Adapt new theming logic for multiple share link icons
As a follow-up for #6332 _and_ #6358 the link icon/background should be themed like the fab/primary-action-buttons now available via the #6332 theming logic.

AndyScherzinger:
Resolve #6399

As a follow-up for #6332 and #6358 the link icon/background should be themed like the fab/primary-action-buttons now available via the #6332 theming logic.

### DIFF
--- a/src/main/java/com/owncloud/android/utils/ThemeUtils.java
+++ b/src/main/java/com/owncloud/android/utils/ThemeUtils.java
@@ -688,15 +688,9 @@ public final class ThemeUtils {
     public static void colorFloatingActionButton(FloatingActionButton button, @DrawableRes int drawable,
                                                  Context context) {
         int primaryColor = ThemeUtils.primaryColor(null, true, false, context);
-        colorFloatingActionButton(button, context, primaryColor);
 
-        if (Color.BLACK == primaryColor) {
-            button.setImageDrawable(ThemeUtils.tintDrawable(drawable, Color.WHITE));
-        } else if (Color.WHITE == primaryColor) {
-            button.setImageDrawable(ThemeUtils.tintDrawable(drawable, Color.BLACK));
-        } else {
-            button.setImageDrawable(ThemeUtils.tintDrawable(drawable, ThemeUtils.fontColor(context, false)));
-        }
+        colorFloatingActionButton(button, context, primaryColor);
+        button.setImageDrawable(ThemeUtils.tintDrawable(drawable, getColorForPrimary(primaryColor, context)));
     }
 
     public static void colorFloatingActionButton(FloatingActionButton button, Context context) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/owncloud/android/utils/ThemeUtils.java
```diff
@@ -712,6 +706,28 @@ public static void colorFloatingActionButton(FloatingActionButton button, int ba
...
+        imageView.getDrawable().mutate().setColorFilter(getColorForPrimary(primaryColor), PorterDuff.Mode.SRC_IN);
+    }
+
+    /**
+     * returns a primary color matching color for texts/icons on top of a primary-colored element (like buttons).
+     *
+     * @param primaryColor the primary color
+     */
+    private getColorForPrimary(int primaryColor) {
+        if (Color.BLACK == primaryColor) {
```
tobiasKaminsky, line 722: Is this obeying dark mode?

src/main/java/com/owncloud/android/utils/ThemeUtils.java
```diff
@@ -712,6 +706,28 @@ public static void colorFloatingActionButton(FloatingActionButton button, int ba
...
+            return Color.WHITE;
+        } else if (Color.WHITE == primaryColor) {
+            return Color.BLACK;
+        } else {
+            return ThemeUtils.fontColor(context, false);
```
tobiasKaminsky, line 727: Isn't this also a good idea to have this computed on server? font-color-light and font-color dark?

AndyScherzinger: Don't know, but @nickvergessen might be able to shed some light on the font-color calculations. I'd hope that with a fallback-ing background color working for dark/light the text-color of a button doesn't have any contrast issues anymore. So we would rather have to discuss if primary button texts would have to be white/black or if we just use white in both theme variants.

So also pinging @jancborchardt about this. ATM we don't have text-color variant from the server (also with element-color reference)

src/main/java/com/owncloud/android/utils/ThemeUtils.java
AndyScherzinger: Not at this very moment, we just do this based on the primary color for edge cases and else use the server's text-color value. See also my other comment.

src/main/java/com/owncloud/android/utils/ThemeUtils.java
tobiasKaminsky: Fine for me to have this in next PR :+1:

PR #6430, opened 2020-07-07
Theming Refactoring

AndyScherzinger:
**WiP**

The genral idea/motivation behind this PR is to cleanup and restructure the theming implementation to:
* generally separate/extract color manipulation and color access
* separate color calculations for the theme (primary/dark/accent) including fallbacks
* separate UI element theming implementations per UI element (like primary button, border-less button, preference category, editText, etc.)

Looping in @dan0xii @JorisBodin @Shagequi @sirambd so you are aware of this in-progress refactoring, also looping in @ezaquarii since it might involve moving from account to user while I'd probably postpone that to a later PR (while I tend to write the UI element theming using kotlin, so looking forward to your feedback when this is in a reviewable state)

### DIFF
--- a/src/main/java/com/nextcloud/ui/theming/ColorsUtils.java
+++ b/src/main/java/com/nextcloud/ui/theming/ColorsUtils.java
@@ -258,34 +258,34 @@ public final class ColorsUtils {
     }
 
     /**
-     * Tests if light color is set
-     *
-     * @param color the color
-     * @return true if primaryColor is lighter than MAX_LIGHTNESS
-     */
-    public static boolean lightTheme(int color) {
-        float[] hsl = ColorsUtils.colorToHSL(color);
-
-        return hsl[INDEX_LUMINATION] >= MAX_LIGHTNESS;
-    }
-
-    /**
      * Tests if dark color is set
      *
      * @return true if dark theme -> e.g.use light font color, darker accent color
      */
     public static boolean darkColor(Context context) {
-        int primaryColor = ColorsUtils.elementColor(context);
-        float[] hsl = ColorsUtils.colorToHSL(primaryColor);
+        int elementColor = ColorsUtils.elementColor(context);
+        float[] hsl = ColorsUtils.colorToHSL(elementColor);
 
         return hsl[INDEX_LUMINATION] <= 0.55;
     }
 
-    private static float[] colorToHSL(int color) {
+    public static float[] colorToHSL(int color) {
         float[] hsl = new float[3];
         ColorUtils.RGBToHSL(Color.red(color), Color.green(color), Color.blue(color), hsl);
 
         return hsl;
+    }
+
+    /**
+     * Tests if color is light
+     *
+     * @param color the color
+     * @return true if color is lighter than MAX_LIGHTNESS
+     */
+    public static boolean isLightColor(int color) {
+        float[] hsl = colorToHSL(color);
+
+        return hsl[INDEX_LUMINATION] >= MAX_LIGHTNESS;
     }
 
     private static boolean isDarkModeActive(Context context) {
--- a/src/main/java/com/nextcloud/ui/theming/ColorsUtils.java
+++ b/src/main/java/com/nextcloud/ui/theming/ColorsUtils.java
@@ -1,0 +1,318 @@
+/*
+ * Nextcloud Android client application
+ *
+ * @author Tobias Kaminsky
+ * @author Andy Scherzinger
+ * Copyright (C) 2017 Tobias Kaminsky
+ * Copyright (C) 2017 Nextcloud GmbH
+ * Copyright (C) 2018 Andy Scherzinger
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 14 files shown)


ezaquarii:
@AndyScherzinger any chance we could ditch static helpers and seal it behind interface? It would be nice if we could switch between themes just by swapping the implementation instead of if-ing around with light/dark helpers.

Just 2 cents.

AndyScherzinger:
> any chance we could ditch static helpers and seal it behind interface? It would be nice if we could switch between themes just by swapping the implementation instead of if-ing around with light/dark helpers.

If you can elaborate a bit more on how I can achieve this, then sure thing.

What we need to have are:
* primary/primaryDark/accent colors (dark/accent being variants of primary and need to be calculated)
* dark/light variants of the above
* making sure to not end up with low contrast values (again depending on dark/light)

ATM the server provides us with server and element color while "server" is the exact color value and element is contrast aware for light. With server version 19.0.1 we can retrieve element light/dark colors which are calculated in a way that takes care of contrast.

So one thing is colors where I am unsure if/how we can hide that behind an interface (the implementation would have to "know" the clients dark/light theme state and use the "right" colors (besides element color we also need to check the font color that depends on the element color making sure contrast is high enough (black/white font color).

The other thing is the theming of UI elements (mainly tinting of background+icons and setting background/text colors) where I would have thought that static, coloring helpers would be the easiest way to go since I don't see a benefit in things like creating custom on-the-fly theming UI elements.

@ezaquarii can you give me some hints/pointers on how to hide this behind an interface (or several)? ❤️

ezaquarii:
@AndyScherzinger 

I did some more research and it looks to me like the bulk of functionality is to provide colors.

One thing that is annoying about current API is the constant necessity to provide `Context` and `Account` (which should be `User`). There is also quite some `null` workarounds for notorious, historical issues with uninitialized `Account` value (we temporarily changed it for more explicit `Optional<User>` and we plan to make entire `User` API `null` safe eventually).

I was thinking about something like this:

```
interface Theme {
    /**
     * Traverse view tree and apply theme-specific cusomizations for view elements.
     * not sure if this is possible, but it would be nice.
     */
    void apply(View rootView);

    /**
     * Obtain primary color for the currently active user.
     */
    int primaryColor()

    /**
     * Obtain primary color for a specific user.
     */
    int primaryColor(@NonNull User user)

    /**
    * Apply foo customization to edit text using theme of currently active user.
    */
    void applyFoo(EditText et);
    ...
}
```

```
/// uam will provide currently active user and the API is aleardy null-safe, so no checks are needed
/// also, since this is instance with state, caching of any computed values can be done at this level
class ActiveUserTheme(private val context: Context, private val uam: UserAccountManager) : Theme {
    ...
}
```

DI can inject `Theme` into fragments / activities ensuring null-safety.

More or less something in that direction. We could save a bit of hassle with `Context` and `Account` or other dependencies, such as settings. There is quite some null checks and ifology aroung those.

However, this might not be worth the effort, if we don't plan to support dynamic theme switching, so maybe it's not that good idea cost/benefit wise. I'll leave it as food for thoughts.

AndyScherzinger:
> I'll leave it as food for thoughts. 

Liking the food for thought, so further comments and thoughts

> One thing that is annoying about current API is the constant necessity to provide Context and Account (which should be User). There is also quite some null workarounds for notorious, historical issues with uninitialized Account value (we temporarily changed it for more explicit Optional<User> and we plan to make entire User API null safe eventually).

totally agree that we should move on to `User` instead of `Account`. It is not just the account that might be a problem since we have had "context==null" crashed via the GPlay dev console in the past, so this is like you said notorious hardening of the code which I still think is a good idea (to some extend) since "theming/coloring" should never crash the application.

> I was thinking about something like this:

I like the idea/approach. Some comments/additional thoughts.

There are different areas I want to tackle here where some of them should be simple, static util methods (more on that below).
What I want to achieve in the long run is a proper separation of different tasks being mixed up here.

Different tasks to cover:
* **contrast aware color correction and calculation** (basically color calculations checking for luminosity etc.) - This is something that is purely based on colors and colors would just be handed over to these functions (thus can/should be static and is purely util functions)
* **dark/light color retrieval/storage** this is basically working with the server-side color infos we get via the capabilities api utilizing the above mentioned re-calculations based on contrast/luminosity/etc. This is where all the context magic happens since we might not get server values (theming disabled on the server) and need to have a fallback (client-side Android xml values) which might fail if context is null (fallback to hardcoded color values)
* **theme color calculations and offering** the server color is just that a color information with some dark/light contrast color values. The server doesn't know "material design" and it's shades (primary, primaryDark, accent). So these need to be calculated in a dark/light background-color aware way
* **theming UI elements** The actual UI elements coloring logic that uses the theme color values offering different ways for different scenarios (e.g. a "borderless button" needs to be themed differently than one with a "primary action / secondary action button").

> However, this might not be worth the effort, if we don't plan to support dynamic theme switching, so maybe it's not that good idea cost/benefit wise.

We don't have dynamic theme switching and the rootView traversal sounds awesome but I also agree that this might not be an efficient way and has its problems regarding theming a single UI element in different ways in different context so "manually" or explicitly doing this might be a good trade off variant of implementing this.

So here is my idea on how to move forward in a evolutionary way with respect to your feedback/input

1. restructure code to support different tasks (see above) to have concerns separated (I am aware that this will have some sunk cost on my side)
2. improve color calculations / harden color retrieval (user/context fallbacks) (I am aware that this will have some sunk cost on my side)
3. improve theming helper API (remove any direct color access from UI code and switch to theme only "business" methods) (this way we remove any UI code dependency to getting the right colors and would have a centralized theming point to harmonize UI theming).
4. move theming implementation into a "injectable" component like mentioned by @ezaquarii 

I am aware that some changes I'd have to do would have do be re-done/thrown away but it would
* keep PR on a mid-sized level
* provide insights throughout the evolutionary steps as for what is needed for theming, how to calculate colors the right way and how to slice things up properly

What do you think @ezaquarii @tobiasKaminsky ?

tobiasKaminsky:
I like the idea.
We then can also create some UI tests with "hardcoded" values, if theming is injected.
Ping me, if you need help for this :)

AndyScherzinger:
successor: #8014

Issue #6360, opened 2020-06-25
Follow up on "hide on scrolling"


tobiasKaminsky:
I have tested the PR and have found two issues:
* [x] #6569 The part below the header bar looks strange in dark mode: different grey tone?
* [ ] The FAB button seems to need a long touch time to get hidden/shown. If I scroll fast (and thus touch short to trigger the scrolling), I can trigger a state where the FAB button doesn't get hidden/shown while scrolling.
* [x] The FAB button and the sorting/grid/list action bar are visible in the Text preview screen (and possible others as well)
* [ ] The empty list elements aren't vertically centered anymore (might be to to the parent element PR, can't say for sure).
* [x] header is scrolled down:
 - open app
 - scroll down, and up so header is gone
 - go into subfolder / detail view
 - header is gone (as this state is remembered)

_Originally posted by @szaimen in https://github.com/nextcloud/android/pull/6262#issuecomment-648956174_

**Screenshots**
![com nextcloud client FileDisplayActivityIT_open](https://user-images.githubusercontent.com/1315170/85758720-ddfe1f00-b710-11ea-95bd-9f6a3deae43d.png)
![com owncloud android ui preview PreviewTextFileFragmentTest_displayJavaSnippetFile](https://user-images.githubusercontent.com/1315170/85758724-de96b580-b710-11ea-9c2d-64f5e2ab49f2.png)

AndyScherzinger:
@tobiasKaminsky I added further issues I found to the list

tobiasKaminsky:
>  The FAB button seems to need a long touch time to get hidden/shown. If I scroll fast (and thus touch short to trigger the scrolling), I can trigger a state where the FAB button doesn't get hidden/shown while scrolling.

Is this still valid?

szaimen:
> > The FAB button seems to need a long touch time to get hidden/shown. If I scroll fast (and thus touch short to trigger the scrolling), I can trigger a state where the FAB button doesn't get hidden/shown while scrolling.
> 
> Is this still valid?

On 3.13.0 it is, yes 👍

tobiasKaminsky:
I cannot reproduce it.
If it is still valid, please re-open with a video.

szaimen:
Here is a screenrecording:
[Screen_Recording_20200924-101039.zip](https://github.com/nextcloud/android/files/5274470/Screen_Recording_20200924-101039.zip)
(You have to touch really short during scrolling to trigger this but it definitely still happens)

[...]

PR #6441, opened 2020-07-09
fix wrong show of sort/list group and FAB in details or text preview

### LINKED ISSUE #6437: Preview View shows Sorting/grid/list element and FAB while it shouldn't
See screenshot, beware that the same applies for the details view

cc @tobiasKaminsky @JorisBodin @Shagequi 

💡 I guess it is due to #6262 ... but that is just a guess

**->** IMHO we should not release 3.13.0 without having this matter fixed

tobiasKaminsky:
Fix #6437

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -1046,6 +1046,8 @@ public class FileDisplayActivity extends FileActivity
             }
             if (listOfFiles != null) {  // should never be null, indeed
                 setFile(listOfFiles.getCurrentFile());
+                listOfFiles.setFabVisible(true);
+                showSortListGroup(true);
             }
             cleanSecondFragment();
         }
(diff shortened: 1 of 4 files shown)


AndyScherzinger:
> I'll create an issue for that if I don't find an existing related one. Or maybe should I add it to #6360? Or both?

Hi @eneiluj, in this case best would be to create a separate one since nested scroll views were there probably before so that problem just got worse but has somewhat existed before :/ Thanks ❤️

AndyScherzinger:
@tobiasKaminsky partially fixed the issue. While the fab/sorting isn't shown anymore the details view is scrolled-down to the activity/share tabs and you need to scroll-up to see the toolbar etc. again. Even in case the tabs are more or less empty.

AndyScherzinger:
further issue:  #6448

tobiasKaminsky:
> details view is scrolled-down at the beginning

This is independent from this PR, or?
I guess Android remembers the last "scroll state".

tobiasKaminsky:
> details view is scrolled-down at the beginning

I will track this in a new issue.

PR #6454, opened 2020-07-11
Remove dead class SignatureVerification

ezaquarii:

- [x] No tests are needed. If it compiles - it works.


ezaquarii:
Nope, my `grep` lied to me and it is actually alive.

Issue #6477, opened 2020-07-15
Display thumbnails for picture files in Activity


pilsnerbeer:
### Is your feature request related to a problem? Please describe.
Thumbnails should be rendered for pictures (even videos, if possible) in Activity to help recognition. I barely name my pics (see screenshot) so it does help with navigation. Furthermore, thumbnails should be clickable (issue exists).

### Describe the solution you'd like
Replace grey-colored shape (see pic) with actual thumbnail of the photo.

### Additional context
![Screenshot_20200715-150303_1 (1)](https://user-images.githubusercontent.com/36133540/87555316-93d7e000-c6b5-11ea-8443-cea53d4eefb4.jpg)

AndyScherzinger:
Thanks for reporting this @pilsnerbeer, labeled as bug now since this is a regression and has worked before.

@tobiasKaminsky tested against Nc19 in my case. Didn't have the time to debug.

tobiasKaminsky:
That is really strange.
I found out that the content of preview image is correctly retrieved, but then it is not shown…
As we did not changed anything on Glide's side I am currently a bit out of ideas…

AndyScherzinger:
> As we did not changed anything on Glide's side I am currently a bit out of ideas…

@tobiasKaminsky Glide shows the placeholders which would also be the case when glide gets an error, so maybe all these thumbnail calls simply fail?

AndyScherzinger:
shipping with next feature release

PR #6519, opened 2020-07-23
Enhancing activity list

tobiasKaminsky:
While debugging #6477  I enhanced it…

- show correct placeholder
- click on preview open file

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java
+++ b/src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java
@@ -234,7 +234,10 @@ public class ActivityListAdapter extends RecyclerView.Adapter<RecyclerView.ViewH
                     for (PreviewObject previewObject : activity.getPreviews()) {
                         if (!isDetailView || MimeTypeUtil.isImageOrVideo(previewObject.getMimeType()) ||
                             MimeTypeUtil.isVideo(previewObject.getMimeType())) {
-                            ImageView imageView = createThumbnailNew(previewObject);
+                            ImageView imageView = createThumbnailNew(previewObject,
+                                                                     activity
+                                                                         .getRichSubjectElement()
+                                                                         .getRichObjectList());
                             activityViewHolder.list.addView(imageView);
                         }
                     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java
```diff
@@ -256,23 +259,34 @@ public void onBindViewHolder(@NonNull RecyclerView.ViewHolder holder, int positi
         }
     }
 
-    private ImageView createThumbnailNew(PreviewObject previewObject) {
+    private ImageView createThumbnailNew(PreviewObject previewObject, ArrayList<RichObject> richObjectList) {
```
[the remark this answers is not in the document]
AndyScherzinger: @tobiasKaminsky should use `List` instead

src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java
```diff
@@ -256,23 +259,34 @@ public void onBindViewHolder(@NonNull RecyclerView.ViewHolder holder, int positi
...
         LinearLayout.LayoutParams params = new LinearLayout.LayoutParams(px, px);
         params.setMargins(10, 10, 10, 10);
         ImageView imageView = new ImageView(context);
         imageView.setLayoutParams(params);
 
+        for (RichObject object : richObjectList) {
+            if (Integer.parseInt(object.getId()) == previewObject.getFileId()) {
+                imageView.setOnClickListener(v -> activityListInterface.onActivityClicked(object));
+                break;
```
AndyScherzinger, line 271: why the `break`?

AndyScherzinger:
should be backported imho 👍

src/main/java/com/owncloud/android/ui/adapter/ActivityListAdapter.java
tobiasKaminsky: https://github.com/nextcloud/activity/blob/master/docs/endpoint-v2.md#example

We are using the PreviewObject, which does not contain the remote path.
So we cannot handle click operation with only this info. (to open a file we currently need the remote path. We could of course also write a function that retrieves first the remote path for a given file from server and use this then).

As this info is already in subject_rich  list, we search for matching info. As there is no direct relation between those (not via index), we need to use fileID and compare it with id.
As there can only be one match, we can do a break, if we found this.

createThumbnailNew(…) creates then one ImageView with corresponding click listener.
For next preview object creatThumbnailNew(…) will be called again and thus do the same.

PR #6533, opened 2020-07-24
add TestActivity to launch fragments in it

tobiasKaminsky:
- screenshot scripts also supports kotlin now
- fix wrong class cast exception message

- [x] add 0b image to it, as test for #6772

This allows us to launch any fragment without our FDA overhead, so that we can test UI without server or delay.

@ezaquarii only downside is that I needed to put TestActivity within regular app.
Do you might have an idea how to prevent this?

### Testing 


ezaquarii:
@tobiasKaminsky Sadly I have nothing on top of my head.

tobiasKaminsky:
> @ezaquarii only downside is that I needed to put TestActivity within regular app.
> Do you might have an idea how to prevent this?

I found a way to only add this in "debug" buildFlavor, so that we do not expose this during "release". :+1:

AndyScherzinger (approved):

PR #6537, opened 2020-07-24
Trashbin: show empty message only if empty

tobiasKaminsky:
Fix #6272 

Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

  

### Testing 
Writing tests is very important. Please try to write some tests for your PR. 
If you need help, please do not hesitate to ask in this PR for help.

[unit tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#unit-tests)
[instrumented tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#instrumented-tests)
[UI tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#ui-tests)

- [ ] Tests written, or not not needed
- [ ] change to ViewBindings
- [ ] write UI tests

PR #6556, opened 2020-07-27
TrashbinActivity: viewbindings tests

tobiasKaminsky:
- moved to ViewBindings
- add UI test for error/view

Follow up of #6537 

### Testing 


tobiasKaminsky:
@ezaquarii not sure if this is the best/easiest approach, but it works.…

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
```diff
@@ -301,29 +263,32 @@ public void removeAllFiles() {
...
+            binding.emptyList.emptyListViewText.setText(message);
+            binding.emptyList.emptyListViewText.setVisibility(View.VISIBLE);
+            binding.emptyList.emptyListIcon.setVisibility(View.VISIBLE);
+            binding.emptyList.emptyListView.setVisibility(View.VISIBLE);
+            binding.emptyList.emptyListProgress.setVisibility(View.GONE);
         }
     }
+
+    @VisibleForTesting
+    public void setTrashbinPresenter(TrashbinPresenter presenter) {
```
AndyScherzinger, line 291: shouldn't the parameter be named trashbinPresenter  and the assignment be done using this? (just for naming convention purposes of course)

ezaquarii:
@tobiasKaminsky This is quite annoying issue with `Activity` design that does not provide clean access to the constructor and enforces 2-stage initialization in most of the components.

However, instead of exposing this public setter, you can make the member package-private (which is one notch above `private`). If the test resides in the same package, you can just assign the presenter to the activity. The same can be done with a setter.

BTW, this is what code-generated dependency injection does, both in direct field injection and setter-based injection.

PR #6572, opened 2020-07-29
Add dark screenshots

tobiasKaminsky:

### Testing 


tobiasKaminsky:
https://github.com/nextcloud/android/pull/6572/files?short_path=6da5b02#diff-6da5b024a398f120874671da2be95d73
image


This seems to be wrong, Settings needs to be in white font.

tobiasKaminsky:
Cause is that settings uses "appFont", which is dynamically changed by Android.
So if Android is set to light, but in settings it is changed to dark theme, this will look wrong.

AndyScherzinger:
> Cause is that settings uses "appFont", which is dynamically changed by Android.
So if Android is set to light, but in settings it is changed to dark theme, this will look wrong.

Nice catch @tobiasKaminsky 👍

AndyScherzinger:
@tobiasKaminsky while it is wrong for the light screenshots too: https://www.kaminsky.me/nc-dev/android-integrationTests/15208-Screenshot-master-dark-blue/images/recorded/com.nextcloud.client.SettingsActivityIT_open.png

tobiasKaminsky:
Hm. It seems that the way we tint our Settings action bar just does not work.
I tried to tint it with Color.RED, but it is not visible at all…🤷‍♂️

tobiasKaminsky:
> Hm. It seems that the way we tint our Settings action bar just does not work.
> I tried to tint it with Color.RED, but it is not visible at all…man_shrugging

This only happens during tests, e.g. when launching via IntentsTestRule…

tobiasKaminsky:
I saw that preferences should be replaced with androidx.preferences: https://developer.android.com/guide/topics/ui/settings

Is it is then only a Fragment, hopefully tinting title is more easy.
For now I would keep it this way, as it is only within tests.

AndyScherzinger (approved):

PR #6600, opened 2020-07-31
l10n: Add a period in the text string

Valdnet:
It is a description that is a complete sentence that should end with a period.


AndyScherzinger:
Thanks @Valdnet ❤️

PR #6619, opened 2020-08-03
Rename Download into Transfer

ezaquarii:
Extending Download into more generic Transfer that
can support both downloads and uploads.

Upload functionality is not provided in this PR - it's
only a rename.

### Testing 
- [x] Tests updated

### DIFF
--- a/src/main/java/com/nextcloud/client/files/downloader/Direction.kt
+++ b/src/main/java/com/nextcloud/client/files/downloader/Direction.kt
@@ -1,0 +1,25 @@
+/*
+ * Nextcloud Android client application
+ *
+ * @author Chris Narkiewicz
+ * Copyright (C) 2020 Chris Narkiewicz <hello@ezaquarii.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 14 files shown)


ezaquarii:
@tobiasKaminsky I'm working on an uploader component and realized that downloader logic is mostly reusable. Instead of providing parallel copy, I'd rather change some concepts to support bi-directional operations.

This PR is an attempt to de-noise the actual Uploader PR from the massive rename change.

There should be no functional changes in this PR.

src/main/java/com/nextcloud/client/files/downloader/Direction.kt
```diff
@@ -0,0 +1,6 @@
+package com.nextcloud.client.files.downloader
```
AndyScherzinger, line 1: lacks a license header

AndyScherzinger:
@ezaquarii looks good to me, only https://github.com/nextcloud/android/pull/6619#discussion_r464319980 Direction should have a license header

src/main/java/com/nextcloud/client/files/downloader/Direction.kt
ezaquarii: fixed

Issue #6518, opened 2020-07-23
 3.13.0 RC1 crashes, when I choose an E2EE-encrypted folder for auto upload


spackmat:
I tried to test auto upload with an E2EE encrypted folder in 3.13.0 RC1, but the app crashes when I select an encrypted folder. When I choose an unencrypted folder, it works as expected. I get the error log below.

Maybe related to #6417, maybe not.

### Steps to reproduce
1. Have a working E2EE folder in 3.13.0 RC1
2. Enable auto upload for a folder and select that encrypted folder
3. Oberve the app crashing

### Expected behaviour
- Choosing an encrypted folder for auto upload should work as before (old E2EE preview) or as with an unencrypted folder.

### Actual behaviour
- The app crashes every time I choose an E2EE encrypted folder as target for auto upload.

### Can you reproduce this problem on https://try.nextcloud.com?
- Didn't test as I don't expect that to have the fresh E2EE beta stuff installed.

### Environment data
Android version: 10

Device model: Xiaomi Mi Mix 3

Stock or customized system: Stock

Nextcloud app version: 3.13.0 RC1

Nextcloud server version: 19.0.1

Reverse proxy: No

#### Nextcloud log (data/nextcloud.log)
Crash report log:
```
************ CAUSE OF ERROR ************

java.lang.NullPointerException: Attempt to invoke virtual method 'void com.google.android.material.floatingactionbutton.FloatingActionButton.setOnClickListener(android.view.View$OnClickListener)' on a null object reference
	at com.owncloud.android.ui.fragment.OCFileListFragment.registerFabListener(OCFileListFragment.java:422)
	at com.owncloud.android.ui.fragment.OCFileListFragment.updateLayout(OCFileListFragment.java:1302)
	at com.owncloud.android.ui.fragment.OCFileListFragment.listDirectory(OCFileListFragment.java:1272)
	at com.owncloud.android.ui.fragment.OCFileListFragment.listDirectory(OCFileListFragment.java:1214)
	at com.owncloud.android.ui.fragment.OCFileListFragment.onItemClicked(OCFileListFragment.java:951)
	at com.owncloud.android.ui.adapter.OCFileListAdapter.lambda$onBindViewHolder$1$OCFileListAdapter(OCFileListAdapter.java:379)
	at com.owncloud.android.ui.adapter.-$$Lambda$OCFileListAdapter$pouWgEOOHT1aDFHaHM0Ifm3xWtY.onClick(Unknown Source:4)
[...]

PR #6623, opened 2020-08-03
OCFileListFragment can be used witout FDA, and this crashed

tobiasKaminsky:
Fix #6518 

Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

### Testing 
Writing tests is very important. Please try to write some tests for your PR. 
If you need help, please do not hesitate to ask in this PR for help.

[unit tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#unit-tests)
[instrumented tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#instrumented-tests)
[UI tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#ui-tests)

- [ ] Tests written, or not not needed

PR #6625, opened 2020-08-03
add scrollable header interface

AndyScherzinger:
fixing https://github.com/nextcloud/android/issues/6518#issuecomment-667064758

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -49,6 +49,7 @@ import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewTreeObserver;
 
+import com.google.android.material.appbar.AppBarLayout;
 import com.google.android.material.snackbar.Snackbar;
 import com.nextcloud.client.account.User;
 import com.nextcloud.client.appinfo.AppInfo;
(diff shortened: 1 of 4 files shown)


tobiasKaminsky:
This is a really great idea, which I had nearly the same.
As we have a quick fix (#6623), we can do this even better for next version.

Currently OCFileListFragment can be used in FDA and others (e.g. SyncedFolderActivity).
This leads to problems as only FDA has FAB, but the others not.

So we should also extract those to the interface, and allow OCFileListFragment to call these methods without any try/catch/casting.

FDA or SyncedFolderActivity then decide what to do, e.g. on register FAB, only, FDA is doing it.

tobiasKaminsky:
While thinking about this again…
I am not sure if my idea the right solution.

Let us go one step back:
- we have ExtendedList
- as child of this OCFileList and LocalFileList
- now we have a "read only" OCFileList (used by e.g. SyncedFolderActivity, FolderPickerActivity and others), and a "full" OCFileList (used by FDA).

This way we would separate this better, or?
FullOCFileList -> calls registerFAB() as it is known via interface to be allowed to do this
ReadOnlyOCFileList -> does not even call it, also e.g. does not have code to handle overflowMenu, etc.

AndyScherzinger:
> This way we would separate this better, or?

Really tricky since we are talking about "lists" of files that might have different "features", a folder sync remote folder picker might not have a overflow menu but might want to have a FAB for folder creation? So maybe we rather need a base file list fragment and have separate fragment child implementations that add the designed features via a composite rather then via inheritance?

PR #6655, opened 2020-08-05
Rename Downloader to TransferManager

ezaquarii:
As the Downloader architecture can support any type of
asynchronous operations so we can easily support file uploads.

Rename Downloader to TransferManager to reflect this.
This is just a rename, without upload implementation.
Upload feature shall follow in separate PRs.

- [x] No functional change - existing tests should still pass

### DIFF
--- a/src/androidTest/java/com/nextcloud/client/files/downloader/TransferManagerConnectionTest.kt
+++ b/src/androidTest/java/com/nextcloud/client/files/downloader/TransferManagerConnectionTest.kt
@@ -34,9 +34,9 @@ import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Test
 
-class DownloaderConnectionTest {
+class TransferManagerConnectionTest {
 
-    lateinit var connection: DownloaderConnection
+    lateinit var connection: TransferManagerConnection
 
     @MockK
     lateinit var context: Context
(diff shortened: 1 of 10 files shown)


ezaquarii:
@tobiasKaminsky @AndyScherzinger This is another foundation change before I can start implementing the uploader.
After analysing the code I concluded that there is no need to implement second uploader component as the current
downloader design can happily support both types of operations.

Hence, we need to change its name.

There are 2 questions we need to answer before we move forward:
Currently, the application has 2 separate components `FileUploader` and `FileDownloader`. Both can service 1 running request.
That means that we can upload 1 file and download 1 file simultaneously.

We have 2 options:
1. we can use 1 transfer manager component to service all tasks (upload or download); that means if we have a long queue of downloads and user wants to upload something, the upload will be placed after downloads
2. we can use 2 transfer manager components, allowing independent upload and download queues - this will complicate any UI that needs to have 1 consistent view of running operations (2 manager, 2 states must be merged

@tobiasKaminsky @AndyScherzinger what does iOS app do? what will be our choice?

AndyScherzinger:
@ezaquarii @tobiasKaminsky I don't know how the iOS app handles it but would propose some kind of a priority value for queue items

1. manually triggered uploads/downloads
2. "synced" files (formerly downloaded/favorited files)
3. auto upload

The idea is that manually triggered actions are top priority since the user is actively doing something, sync comes next since we should keep files in sync of course, auto upload is last since we do this based on certain connection characteristics if set and also have a certain amount of time before we actually upload (within 15 minutes).

PR #6699, opened 2020-08-14
Handle back button in richdocuments

eszkadev:
This patch allows to go back in menus using back button in Collabora Online instead of closing the file (only when menu is opened).

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/RichDocumentsEditorWebView.java
+++ b/src/main/java/com/owncloud/android/ui/activity/RichDocumentsEditorWebView.java
@@ -251,6 +251,17 @@ public class RichDocumentsEditorWebView extends EditorWebView {
         startActivity(intent);
     }
 
+    @Override
+    public boolean onKeyDown(int keyCode, KeyEvent event) {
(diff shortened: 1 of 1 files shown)


kendy:
Szymon: I haven't try, but given that it seems that ultimately RichDocumentsEditorWebView is an Activity, shouldn't it rather be like:

@Override
public void onBackPressed() {
    if (webview.canGoBackOrForward(-2)) {
        webview.goBack();
        return;
    }
    super.onBackPressed();
}

instead of the onKeyPressed?

Tobi: Can you please review?

tobiasKaminsky:
I tested this and "webview.canGoBackOrForward(-2)" is always false for me.
I tried this:
- open up odt
- directly click back

- open up odt
- open menu
- click back

- open up odt
- open menu
- go into submenu
- click back



> instead of the onKeyPressed?

Documentation says that onKeyDown is only called if no other is handling key events, and it seems that RD/Collabora are handling all key presses, e.g. abc, on their own,but only let back button pass.

kendy:
Tobias: Against which server were you trying this - can you please write that to me privately? Szymon had a patch that was needed on the Collabora Online side (particularly fbfc1f4d2c5fb9e0383cfb681efe7bbe82211f03 "Handle back button on mobile" - if you want to check yourself in the Help -> About, in the online repo).

PR #6725, opened 2020-08-17
Improve performance of DocumentsProvider#isChildDocument()

grote:
The old solution iterates through all parents of the child until it finds the given parent or the storage root. This has been show to be very expensive in empirical tests.

Therefore, this commit introduces a more efficient solution that simply compares the file paths of child and given parent. It also ensures that parent and child belong to the same account.

Reviewers need to take extra care that this change does not introduce security issues by claiming a document is a child of a parent when it is really not. Maybe also check if the correct path is compared as there seem to be many: local/remote/encrypted/decrypted

### Testing 

I'd be happy to provide instrumentation tests against the `DocumentsProvider` implementation. However, it seems that there's no framework for that in place, yet. So this would need to be discussed before.

The code change is rather small and it should be evident, that if the accounts match, the following is true:

* `/a/b/c` is a child of `/a/b` and `/a`
* `/a/b/c` is *not* a child `/a/x`

### DIFF
--- a/src/main/java/com/owncloud/android/operations/DownloadFileOperation.java
+++ b/src/main/java/com/owncloud/android/operations/DownloadFileOperation.java
@@ -171,7 +171,7 @@ public class DownloadFileOperation extends RemoteOperation {
             modificationTimestamp = downloadOperation.getModificationTimestamp();
             etag = downloadOperation.getEtag();
             newFile = new File(getSavePath());
-            if (!newFile.getParentFile().mkdirs()) {
+            if (!newFile.getParentFile().exists() && !newFile.getParentFile().mkdirs()) {
                 Log_OC.e(TAG, "Unable to create parent folder " + newFile.getParentFile().getAbsolutePath());
             }
 
--- a/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
+++ b/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
@@ -618,15 +618,29 @@ public class DocumentsStorageProvider extends DocumentsProvider {
         Log.d(TAG, "isChildDocument(), parent=" + parentDocumentId + ", id=" + documentId);
 
         try {
-            Document currentDocument = toDocument(documentId);
+            // get file for parent document
             Document parentDocument = toDocument(parentDocumentId);
-
-            while (!ROOT_PATH.equals(currentDocument.getRemotePath())) {
-                currentDocument = currentDocument.getParent();
-                if (parentDocument.getFile().equals(currentDocument.getFile())) {
-                    return true;
-                }
+            OCFile parentFile = parentDocument.getFile();
+            if (parentFile == null) {
+                throw new FileNotFoundException("No parent file with ID " + parentDocumentId);
             }
+            // get file for child candidate document
+            Document currentDocument = toDocument(documentId);
+            OCFile childFile = currentDocument.getFile();
+            if (childFile == null) {
+                throw new FileNotFoundException("No child file with ID " + documentId);
+            }
+
+            String parentPath = parentFile.getDecryptedRemotePath();
+            String childPath = childFile.getDecryptedRemotePath();
+
+            // The alternative is to go up the folder hierarchy from currentDocument with getParent()
+            // until we arrive at parentDocument or the storage root.
+            // However, especially for long paths this is expensive and can take substantial time.
+            // The solution below uses paths and is faster by a factor of 2-10 depending on the nesting level of child.
+            // So far, the same document with its unique ID can never be in two places at once.
+            // If this assumption ever changes, this code would need to be adapted.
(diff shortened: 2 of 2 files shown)


src/main/java/com/owncloud/android/operations/DownloadFileOperation.java
```diff
@@ -172,7 +172,8 @@ protected RemoteOperationResult run(OwnCloudClient client) {
             etag = downloadOperation.getEtag();
             newFile = new File(getSavePath());
             if (!newFile.getParentFile().mkdirs()) {
-                Log_OC.e(TAG, "Unable to create parent folder " + newFile.getParentFile().getAbsolutePath());
+                // File#mkdirs() also returns false when the directory already existed, so this not a hard error
+                Log_OC.d(TAG, "Unable to create parent folder " + newFile.getParentFile().getAbsolutePath());
```
grote, line 176: This is an unrelated change I am hoping to sneak in here. Seeing this "error" in the logs is quite distracting when debugging things.

If you'd want to keep this as an error level log message, we should add a check if that file already exists.

tobiasKaminsky:
> I'd be happy to provide instrumentation tests against the `DocumentsProvider` implementation. However, it seems that there's no framework for that in place, yet. So this would need to be discussed before.

I fear we cannot use our own app to test against DocumentsProvider, right? So we would need to have an entirely different app that is testing it?

grote:
> I fear we cannot use our own app to test against DocumentsProvider, right? So we would need to have an entirely different app that is testing it?

That's how I am testing at the moment. However, I haven't tried using the same app. It might work. The bigger issue might be to automatize getting the folder permission with ACTION_OPEN_DOCUMENT_TREE, so the tests can run in the CI. I am currently manually getting this permission, take a persistent permission on it and then run my instrumentation tests.

Also, do you already have a framework that provides instrumentation tests with a logged in Nextcloud account that is set up against a test server? This would also be needed I guess.

grote:
From https://medium.com/androiddevelopers/building-a-documentsprovider-f7f2fb38e86a

> remember **a document can be in multiple places in your hierarchy** so this is checking ‘downward’ as it were from parent to potential descendant (child, grandchild, etc).
> 
> Ideally, this request shouldn’t rely on the network as it **can be called frequently and in rapid succession** so if you want to support this use case, make sure you can handle this request locally.

So I don't know if in Nextcloud one document with the same ID can be in several places at once. My assumption is that it can only be in one place.

tobiasKaminsky:
> That's how I am testing at the moment. However, I haven't tried using the same app. It might work. The bigger issue might be to automatize getting the folder permission with ACTION_OPEN_DOCUMENT_TREE, so the tests can run in the CI. I am currently manually getting this permission, take a persistent permission on it and then run my instrumentation tests.

This should be possible via 
@Rule public GrantPermissionRule mRuntimePermissionRule =        GrantPermissionRule.grant(Intent.ACTION_OPEN_DOCUMENT_TREE);



> Also, do you already have a framework that provides instrumentation tests with a logged in Nextcloud account that is set up against a test server? This would also be needed I guess.

Yes, there are lots of them, e.g. UploadIT, this uploading a file to a server while CI.
You can try it and if you somehow fail/need info, we can work together on a separate PR?

tobiasKaminsky:
> > 
> 
> So I don't know if in Nextcloud one document with the same ID can be in several places at once. My assumption is that it can only be in one place.

It could somehow, if you e.g. have two shares from different users but targeting the same folder.
This is then two different files on Android with different db ids.
But as the android user will then approach it via different folder hierarchies, I think we are safe here.

grote:
> This should be possible via @rule public GrantPermissionRule mRuntimePermissionRule =  GrantPermissionRule.grant(Intent.ACTION_OPEN_DOCUMENT_TREE);

Hmm.. are you sure? I know this only for granting runtime permissions, but `Intent.ACTION_OPEN_DOCUMENT_TREE` is not a permission, but an intent used to launch the system's file app which allows the user to grant you permission to a SAF folder.

This could help: https://stackoverflow.com/questions/58491688/how-to-create-a-content-uri-for-intending-action-open-document-tree

>  You can try it and if you somehow fail/need info, we can work together on a separate PR?

Great, I can also open a dedicated ticket to discuss these tests further.

>  But as the android user will then approach it via different folder hierarchies, I think we are safe here.

Graet. Just to be sure: Does `toDocument(documentId)` then give you the file in the other hierarchy?

src/main/java/com/owncloud/android/operations/DownloadFileOperation.java
tobiasKaminsky: Can you change to !newFile.getParentFile().exists() &&!newFile.getParentFile().mkdirs()

then we only try to create the folder if it does not exist and thus the error is correct, if it fails.

tobiasKaminsky:
Please fix codacy warnings :-)

grote:
All warnings fixed.

tobiasKaminsky (approved):

PR #6748, opened 2020-08-19
no sticky broadcast

### LINKED ISSUE #6654: Replace sticky broadcasts with and event-mechanism or else
**Workaround**
Installing app into the main profile also (just installing is enough)

**Issue description**
As reported in several tickets the app crashes with a security exception in case the app is used in a non-primary profile which might be a secondary profile or a Android 4 Work profile. This is due to the usage of (deprecated) sticky broadcasts, see for example #4205

> Google 
Sticky broadcasts should not be used. They provide no security (anyone can access them), no protection (anyone can modify them), and many other problems. The recommended pattern is to use a non-sticky broadcast to report that something has changed, with another mechanism for apps to retrieve the current value whenever desired.

It seems the solution is to move to events e.g. via an Eventbus, see https://stackoverflow.com/a/46836794 with GreenRobot EventBus being part of our application already _OR_ see  #6048 by @ezaquarii to move to complete rewrites for these areas of the code base.

What do you think @tobiasKaminsky @ezaquarii ? Since the sticky broadcasts have been deprecated with API21 we will liekly get into more trouble in the future while I can't say if @ezaquarii's solution can be applied in all situations where we use sticky*.

Closing #4205 in favor of this issue that rather discusses the solution and can be used as a central issue.

**Reports**
#4205 #4229 #4476 #4553 #4565 #4573 #4615 #4627 #4661 #4896 #4931 
#5059 #5131 #5309 #5366 #5514 #5709 #5831 #5907 
#6545 #6649

tobiasKaminsky:
Fix #6654 

- use local broadcast system instead
- no need to remove sticky broadcasts then

### Testing 


AndyScherzinger (approved):

src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -691,6 +692,6 @@ private void sendLocalBroadcast(String event, String dirRemotePath, RemoteOperat
         intent.putExtra(FileSyncAdapter.EXTRA_RESULT, dataHolderItemId);
 
         intent.setPackage(mContext.getPackageName());
-        mContext.sendStickyBroadcast(intent);
+        LocalBroadcastManager.getInstance(mContext.getApplicationContext()).sendBroadcast(intent);
```
tobiasKaminsky, line 695: @ezaquarii how could I use Dagger here?
Only via constructor, or?

ezaquarii: Yeah, dagger works only in main components. Everything else is constructor based injection or setter if 2 stage init is required (I think we have 1 case of the latter).

tobiasKaminsky: Do you think it is worth the effort, or can we keep it this way?
After all it is a singleton, so use of Dagger is not that much needed, but mainly only for centralizing the initiation?

tobiasKaminsky: Let us keep it now this way to get it in faster.
If we need DI for testing, we can change it anyways.

PR #6757, opened 2020-08-20
Remove big spinner

### LINKED ISSUE #6271: 2 loader in same time
Hello @jancborchardt

Do you have any idea what that should be like?
|On device|On first folder download|
|---|---|
|image|image|

### LINKED ISSUE #6684: [3.13.0 RC4] Dual loading animation
### Steps to reproduce
1. Launch the app.
2. Quickly select "Photos" from the menu while the loading wheel is still visible for "All files".

### Expected behaviour
- **Only** one loading wheel.

### Actual behaviour
- Double load wheel for "Photos" while "All Files" is still loading.

obraz

**Double charging also appears when we enter "On device" or "Shared" and refresh by swiping down.**

### Environment data
Android version: 10

Nextcloud app version: 3.13.0 RC4 and Dev 20200809

tobiasKaminsky:
Fix #6684 
Fix #6271

### Testing 

UI Tests:

- [x] UI Test for Contacts List shimmer


tobiasKaminsky:
I "stupidly" removed all occurrences of big spinner.
Now there are also references on e.g. PreviewImageFragment, where we might keep it?

AndyScherzinger (approved):

tobiasKaminsky:
Let us get this in, so we can see where it might be needed :+1:

AndyScherzinger (approved):

AndyScherzinger:
@tobiasKaminsky looking at the failing screenshot tests I guess we will be missing a loading indicator in quite some places, no?

tobiasKaminsky:
As discussed with @jancborchardt we remove the big spinner, but show instead a skeleton row, with shimmer, like for thumbnails.

AndyScherzinger:
@tobiasKaminsky I added a shimmer loading screen for activities via https://github.com/nextcloud/android/pull/6757/commits/8343cb39320c7eaceb13cb02590ca2f50180b797

szaimen:
Hey, I've tested this and it looks much nicer but the small spinner is now on top of the **Loading...** word which looks a bit odd to me...
Screenshot_20201016-000735_Nextcloud QA

AndyScherzinger:
@jancborchardt any idea for a skeleton view for the Android app when loading a full-screen image? ATM we have an all empty screen with a horizontally and vertically, spinning circle but with moving to use skeletons for loading screen animation in all places (within this PR) we might also want to change this for the preview screen but @tobiasKaminsky and me don't have a nice idea how such a screen could look like.

jancborchardt:
@AndyScherzinger that’s indeed a good question! What could be some good options if they are possible:
- If we can preload the image measurements: Showing a container with the measurements of that image, pulsing a bit like the skeleton screens
- If we can get a dominant color or some colors of the image, coloring that container in a color of the image (I think Instagram does this)
- Try to load lower-resolution images first and in the background automatically replace them with higher-res ones as they are received by the client (even if they are veeery very low-res like 5*5px or so, just load and blur them, by which we basically get the previous suggestion)
- Other ideas? :)

tobiasKaminsky:
If we have a thumbnail, we show this, as far as I know.
But what if we do not have it…?

jancborchardt:
If we don’t have anything of any sort which refers to the image (not even the dimensions), then we should show the file type icon, if possible also in the shimmering way that the skeleton screen also shows.

tobiasKaminsky:
@AndyScherzinger is this ready? :-)

AndyScherzinger:
> @AndyScherzinger is this ready? :-)

not yet, the larger parts are done but still some quirks to iron out.

AndyScherzinger:
I've taken care of activities screen as well as auto upload while working on notifications screen atm with some polishing work to be done for preview image loading. At least file details tabs aren't touched yet and I have to check if there are other screens that need spinners as well

AndyScherzinger:
@tobiasKaminsky fancy taking a look at the preview image loading? the shimmer loading icon is only shown for downloaded images, else it is never visible while I moved it to the top layer. Could this be related to some async image fetching glide magic?

AndyScherzinger:
@tobiasKaminsky I vote to merge this PR since it basically fixes the issue at hand. The problem with the loading animation and the image preview is a regression (just checked and the spinner isn't visible at the moment either...) so this PR isn't making anything worse :)
Will update the latest screenshots as soon as they get published.

AndyScherzinger:
@tobiasKaminsky it seems unit tests fail due to the changed loading animation (shimmer instead of loading state)

tobiasKaminsky:
> @tobiasKaminsky it seems unit tests fail due to the changed loading animation (shimmer instead of loading state)

https://www.kaminsky.me/nc-dev/android-integrationTests/6757-Screenshot-Blue-Light-23-53/#comnextcloudclientFileDisplayActivityScreenshotIT_drawer


Test class: com.nextcloud.client.FileDisplayActivityScreenshotIT
Test name: open
---> this should also have skeleton, or?

Test class: com.owncloud.android.ui.fragment.FileDetailFragmentStaticServerIT
Test name: showFileDetailActivitiesFragment
---> the left "single" short line looks a bit strange…

Test class: com.owncloud.android.ui.trashbin.TrashbinActivityIT
Test name: loading
---> the very right small skeleton is the overflow icon, I guess. Is there a change to remove this and the "restore" button skeleton? I think that the skeleton then looks a bit cleaner.
Also the thumbnail is somehow cut off? It should have rounded corners on every 4, and not only top left, or?

tobiasKaminsky:
I'll take care of tests now.

src/main/java/com/owncloud/android/ui/activities/ActivitiesPresenter.java
```diff
@@ -48,7 +48,11 @@
 
     @Override
     public void loadActivities(int lastGiven) {
-        activitiesView.setProgressIndicatorState(true);
+        if (UNDEFINED == lastGiven) {
+            activitiesView.showLoadingMessage();
```
tobiasKaminsky, line 52: showLoadingMessage() is then set to View.GONE.
Can you directly set this to "gone" in xml, and we can skip this then?

src/main/java/com/owncloud/android/ui/activities/ActivitiesPresenter.java
```diff
@@ -48,7 +48,11 @@
...
+        } else {
+            activitiesView.setProgressIndicatorState(true);
```
tobiasKaminsky, line 54: setProgressIndicatorState sets "isLoadingActive".
This is needed for scroll listener, so we need setProgressIndicatorState to be called.

src/main/java/com/owncloud/android/ui/activities/ActivitiesPresenter.java
AndyScherzinger: No, the empty list view is part of the included layout so this would then impact any screen that includes the empty list file (!)

src/main/java/com/owncloud/android/ui/activities/ActivitiesPresenter.java
AndyScherzinger: How would this need to change? setting isLoading is fine while setRefreshing() shall not be called here (else you have a mini spinner _and_ a shimmer element!

AndyScherzinger:
>Test class: com.nextcloud.client.FileDisplayActivityScreenshotIT
>Test name: open
>---> this should also have skeleton, or?

I think we decided against this for FDA but could add that (later). I am unsure here since navigation through folder structure would lead to a shimmering list on all there folders initially (first opening). So this might be a bit "hectic"?

>Test class: com.owncloud.android.ui.fragment.FileDetailFragmentStaticServerIT
>Test name: showFileDetailActivitiesFragment
>---> the left "single" short line looks a bit strange…

I removed it now, it was replacing the left-aligned date header (today/

>Test class: com.owncloud.android.ui.trashbin.TrashbinActivityIT
>Test name: loading
>---> the very right small skeleton is the overflow icon, I guess. Is there a change to remove this and the "restore" button >skeleton? I think that the skeleton then looks a bit cleaner.
>Also the thumbnail is somehow cut off? It should have rounded corners on every 4, and not only top left, or?

removed the icon shimmers
Thumbnail isn't cutoff (in reality) just in the screenshots, some other screenshots too :/

tobiasKaminsky:
> I think we decided against this for FDA but could add that (later). I am unsure here since navigation through folder structure would lead to a shimmering list on all there folders initially (first opening). So this might be a bit "hectic"?

Fine, let us discuss this in a new issue.

tobiasKaminsky:
> Thumbnail isn't cutoff (in reality) just in the screenshots, some other screenshots too :/

This might be due to different screen resolutions…

AndyScherzinger:
@tobiasKaminsky this is finally ready to go

tobiasKaminsky:
Thanks for all the great work, @AndyScherzinger :tada:  :tada:  :tada:  :tada:

Issue #6796, opened 2020-08-24
Video playback controls not functioning in 3.13 


seanthegeek:
### Steps to reproduce
1. Start streaming a video (e.g. .mp4 file
2.  Attempt to use the playback control buttons

### Expected behaviour
- Playback controls should function
- Seek bar should appear
- Changing rotation or switching apps should not break the UI or return to the start of the video

### Actual behaviour
- Pause/rewind/fast-forward buttons and the video itself do not respond to taps
- Seek bar is missing
- Changing rotation or switching apps causes file browser UI elements to appear on top of the video UI, and also causes the video to seek to the start

### Can you reproduce this problem on https://try.nextcloud.com?
Not a server issue

### Environment data
Android version: 10

Device model: Pixel 3 XL

Stock or customized system: Stock

Nextcloud app version: 3.13

Nextcloud server version: Snap

Reverse proxy: Nginx

ezaquarii:
This seems to be duplicate of #6796

agilob:
@ezaquarii this is #6796 :D

Issue #6782, opened 2020-08-23
Tapping music (media) player notification should open the player


pilsnerbeer:
# Note: Items 1 & 2 are addressed; leaving open to track enhancement idea in item 3 and updated issue title to reflect that idea

<details>
<summary>PROBLEM 1 - music controls not working</summary>

### ~~Steps to reproduce~~
1. ~~Open music file~~
2. ~~try to pause it~~

~~Screenshot: music controls not responsive.~~

![Screenshot_3](https://user-images.githubusercontent.com/36133540/90974808-ce436100-e52e-11ea-8dee-bcf5bd7b9d33.png)
</details>

<details>
<summary>PROBLEM 2 - exiting & re-opening music player renders weird layout</summary>

### ~~Steps to reproduce~~
~~1. Open music file~~
~~2. hit HOME to minimize it~~
~~3. open it again from multitasking view~~

~~A non-responsive layout is rendered with no music controls, FAB and some elements from main view:~~

![Screenshot_4](https://user-images.githubusercontent.com/36133540/90974827-0e0a4880-e52f-11ea-84f7-1182f67955eb.png)
</details>

PROBLEM 3 - tapping music player notification produces no response
### Steps to reproduce
1. play music
2. hit HOME to return to homescreen
3. tap nextcloud music player notification

Nothing happens. I believe the player should open.

________

Android version: 10

Device model: OP5T

Nextcloud app version: 3.13.0

Nextcloud server version: 19

ezaquarii:
@tobiasKaminsky PR #6799 provides only partial fix for first reported problem.

Second problem - controls messed up after screen switch - is much more sinister.

Preview works by adding `PreviewMediaFragment` to the stack, on top of `OCFileListFragment`. The problem is that both fragments are attached to the activity and both depend on "global" activity state with different expectations.

When screen is updated (resumed from background or rotated), `FileDisplayActivity.onResume()` retrieves `OCFileListFragment` and calls `OCFileListFragment.listDirectory()` that reconfigures `fab_main` button and `sort_list_button_group` via
`OCFileListFragment.updateLayout()`.

This happens when `PreviewMediaFragment` is on top of the fragment backstack.

Issue was introduced in 38c28332e266fa3f4324868b36963fc0d3176700 where FAB manipulation was added to `OCFileListFragment`. 

I see 2 possible fixes for this problem:
[...]

PR #6799, opened 2020-08-24
Fix unresponsive preview player controls

ezaquarii:
Media controls view were covered by a ScrollView
stealing all touch events.

This fixes immediate problem with unresponsive media controls
reported in #6796 and #6782.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/preview/PreviewMediaFragment.java
+++ b/src/main/java/com/owncloud/android/ui/preview/PreviewMediaFragment.java
@@ -49,6 +49,7 @@ import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.ProgressBar;
 import android.widget.RelativeLayout;
+import android.widget.ScrollView;
 import android.widget.TextView;
 import android.widget.VideoView;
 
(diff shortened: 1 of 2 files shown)


ezaquarii:
Encryption IT fails. https://www.kaminsky.me/nc-dev/android-integrationTests/15553-IT-master/

tobiasKaminsky:
@ezaquarii 
thanks for finding the cause of this problem!
I could not find it!
As you pointed now to the root problem, it is also solvable by changing ScrollView to LinearLayout in file_preview.xml.
Then there is no code change needed.

What do you think?

ezaquarii:
I did now know why scrollview was needed there and assumed that the scrolling behavior is needed for some reason (landscape maybe?).

This should work as well as layouts are not capturing touch events.

AndyScherzinger:
Sounds good. In case we rely on scroll view for landscape then we should rather change the layout for landscape than rely on scrolling 👍

ezaquarii:
I'll try `LinearLayout` approach.

ezaquarii:
Fix moved to https://github.com/nextcloud/android/pull/6805

PR #6806, opened 2020-08-25
Improve DocumentsProvider and add tests

### LINKED ISSUE #6753: Make DocumentsProvider operations synchronous
This ticket started from #6726 which has more background information and reasoning why this is a good idea. (Seafile [does it like this as well](https://github.com/haiwen/seadroid/blob/5a0710c1d8fc4a57c26dfbc9df201b96c903bbd6/app/src/main/java/com/seafile/seadroid2/provider/SeafileProvider.java).)

I looked into making the entire `DocumentsProvider` do its operations synchronously on the I/O thread that its methods are already called on. It actually already does this for most operations (move/rename/create/delete) and making the rest synchronous is quite straightforward and would improve the code as well. I'd be happy to provide a PR for that.

The only thing I am not sure about is the use of two `SynchronizeFileOperation`s in [`DocumentsStorageProvider`](https://github.com/nextcloud/android/blob/7dd42f0bac0ad336934a244b762e097072ce9de3/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java). 

The first one [happens](https://github.com/nextcloud/android/blob/7dd42f0bac0ad336934a244b762e097072ce9de3/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java#L226) when a file to be opened is available locally already. It is probably a good idea to sync this before, when the caller wants to read from the file. If you just want to write into the file, it might not be needed. However, syncing it can mean that async operations like uploads are triggered (and it can reveal conflicts we can't deal with here).
It would be easier if we can assume the local file to up-to-date, but I don't know if there's methods to check for that and how Nextcloud propagates server changes to its local cache.

The second use of `SynchronizeFileOperation` happens when a caller is done writing to a file and closed it. In this case, we probably don't need to do a sync, but can do a `UploadFileRemoteOperation` (like in `createFile()`) right away, right? Or maybe better use a `UploadFileOperation` so we get chunking for files that are two big.

@tobiasKaminsky do you want me to provide a PR for this and do you have thoughts about making the remaining operations synchronous, especially the two cases above?

@jmue since you contributed most of the to be modified code I would also be interested in your opinion.

grote:
This PR improves the implementation of the `DocumentsProvider` and adds first instrumentation tests.

### Testing 

Closes #6753


grote:
@jmue maybe you also want to have a look!

src/androidTest/java/com/owncloud/android/providers/DocumentsProviderUtils.kt
```diff
@@ -0,0 +1,198 @@
...
+import java.io.IOException
+import java.io.InputStream
+import kotlin.coroutines.resume
+
+// Uploads can sometimes take a bit of time, so 15sec is still considered recent enough
+private const val RECENT_MILLISECONDS = 15_000
+
+object DocumentsProviderUtils {
+
+    internal fun DocumentFile.getOCFile(storageManager: FileDataStorageManager): OCFile? {
```
tobiasKaminsky, line 42: Is there a reason to do this?
Other than that it is now in Kotlin? :)

tobiasKaminsky (approved):

jmue:
I tested your changes with the internal android files app using the included archive functionality and some other SAF tools. I encountered no new errors, so I want to give a :thumbsup:. Besides that I covered two (previously existing) issues and filled the bug reports here https://github.com/nextcloud/android/issues/6883 and here https://github.com/nextcloud/android/issues/6884

grote:
Thanks @jmue for testing and filing those tickets! The first one I noticed as well, but decided not to tackle it within this PR.

PR #6822, opened 2020-08-26
Reduce extend of FileActivity, use DrawerActivity instead

tobiasKaminsky:
This cleans up unused dependencies, which might help us for later refactoring.

### Testing 


AndyScherzinger:
Not sure about upload files but user info only has a back navigation in the toolbar so it wouldn't need the drawer (loaded)

tobiasKaminsky:
> Not sure about upload files but user info only has a back navigation in the toolbar so it wouldn't need the drawer (loaded)

Unfortunately it does require DrawerActivity for account switching, which we can change in another PR.

PR #6871, opened 2020-09-01
fix inplace scrollview in sharing fragment

jmue:
This patch removes the embedded scrollview as well as the size constraint of the two unrelated views.

The embedded scrollview leads to the problem that further shares are (sometimes completely) hidden. The constraint between the two views creates empty space that is not necessary.

Screenshot_1598971890
Screenshot_1598971590

### DIFF
--- a/src/main/res/layout/file_details_sharing_fragment.xml
+++ b/src/main/res/layout/file_details_sharing_fragment.xml
@@ -101,8 +101,7 @@
         <androidx.recyclerview.widget.RecyclerView
             android:id="@+id/publicShareList"
             android:layout_width="match_parent"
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
> This patch removes the embedded scrollview

Is there something missing? I do only see changes related to RecyclerView, but not ScrollView?

jmue:
Sorry, I my explanations are misleading. There is no explicit scrollview instance which I changed. The recyclerview spans up a "embedded scrollview experience" as the recyclerview content area is smaller than the content it holds. This is due to the size ratio between the two recyclerviews. The layouter wants to keep them equal no matter how much content they hold.

PR #6902, opened 2020-09-07
prefer IPv6 if available fixes #3193

### LINKED ISSUE #3193: Default to IPv6 when both AAAA and A record is returned by resolver
**Is your feature request related to a problem? Please describe.**
Normal home internet connection will only be given 1 IPv4 address and use NAT. If there is another service using TCP/443, you will not get another TCP/443 for other service. Since there are no NAT for IPv6, you can open TCP/443 to the Internet on multiple machine without needing to use other port number for port forwarding.

**Describe the solution you'd like**
Use IPv6 by default when DNS resolver return both IPv6 and IPv4 addresses. Fallback to IPv4 if IPv6 connection does not works.

**Additional context**
Follow the Internet Standard! 
Nextcloud desktop sync client seems to use IPv6 by default, why not do the same for android app?

200Puls:
I was a bit unsure where to put the property, so I decided to put it in the main setup where it should work before any other network related things.

Let me know if this is ok for you.

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -246,6 +246,8 @@ public class MainApp extends MultiDexApplication implements HasAndroidInjector {
         setAppTheme(preferences.getDarkThemeMode());
         super.onCreate();
 
+        System.setProperty("java.net.preferIPv6Addresses", "true");
+	
         insertConscrypt();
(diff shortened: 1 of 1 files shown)


200Puls:
This did not work as expected. See: https://github.com/nextcloud/android-library/pull/503

PR #6917, opened 2020-09-09
DocumentsStorageProvider: prevent opening stale file contents

### LINKED ISSUE #6883: DocumentStorageProvider opens stale file contents
If a file is locally available on the client and the server changes the content in the meantime, the document storage provider provides the available content while an internal content update is started at the same time.

Steps to reproduce:
1. Create a text file with the content "Content"
2. Open this file via the SAF and then close it afterwards
3. Change the contents of the file on the server to "New Contents".
4. Open the file again via SAF. Notice the content of "Contents".

You will also see Nextcloud notifications while downloading as you open the file the second time.

The problem here is with the SynchronizeFileOperation, which detects the conflict and calls the FileDownloader intent internally (which itself shows the notifications we want to remove).

To solve this we have to remove this SynchronizeFileOperation here as well, detect the conflict ourselves and start the file download as if no file is down.
This solution requires copying some code from [SynchronizeFileOperation](https://github.com/nextcloud/android/blob/b580352ccc3c5f5baaa1f710e1a0cbe798e95d75/src/main/java/com/owncloud/android/operations/SynchronizeFileOperation.java#L194-L222) to our storage provider.

jmue:
Replace SynchronizeFileOperation by ReadFileRemoteOperation to prevent opening stale file contents
fix #6883

Signed-off-by: Jens Mueller <tschenser@gmx.de>

### DIFF
--- a/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
+++ b/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
@@ -196,9 +196,7 @@ public class DocumentsStorageProvider extends DocumentsProvider {
         OCFile ocFile = document.getFile();
         Account account = document.getAccount();
 
-        int accessMode = ParcelFileDescriptor.parseMode(mode);
-
-        boolean needsDownload = (accessMode != ParcelFileDescriptor.MODE_WRITE_ONLY) && (!ocFile.isDown() || hasServerChange(ocFile, account));
+        boolean needsDownload = !ocFile.isDown() || hasServerChange(document);
         if (needsDownload) {
             if (ocFile.getLocalModificationTimestamp() > ocFile.getLastSyncDateForData()) {
                 // TODO show a conflict notification with a pending intent that shows a ConflictResolveDialog
@@ -216,6 +214,7 @@ public class DocumentsStorageProvider extends DocumentsProvider {
 
         File file = new File(ocFile.getStoragePath());
 
+        int accessMode = ParcelFileDescriptor.parseMode(mode);
         if (accessMode != ParcelFileDescriptor.MODE_READ_ONLY) {
             // The calling thread is not guaranteed to have a Looper, so we can't block it with the OnCloseListener.
             // Thus, we are unable to do a synchronous upload and have to start an asynchronous one.
@@ -249,10 +248,11 @@ public class DocumentsStorageProvider extends DocumentsProvider {
         }
     }
 
-    private boolean hasServerChange(OCFile ocFile, Account account) throws FileNotFoundException {
+    private boolean hasServerChange(Document document) throws FileNotFoundException {
         Context context = getNonNullContext();
+        OCFile ocFile = document.getFile();
         RemoteOperationResult result = new CheckEtagRemoteOperation(ocFile.getRemotePath(), ocFile.getEtag())
-            .execute(account, context);
+            .execute(document.getAccount(), context);
         switch (result.getCode()) {
             case ETAG_CHANGED:
                 return true;
--- a/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
+++ b/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
@@ -64,7 +64,9 @@ import com.owncloud.android.lib.common.OwnCloudClientManagerFactory;
 import com.owncloud.android.lib.common.accounts.AccountUtils.AccountNotFoundException;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult;
 import com.owncloud.android.lib.common.utils.Log_OC;
+import com.owncloud.android.lib.resources.files.ReadFileRemoteOperation;
 import com.owncloud.android.lib.resources.files.UploadFileRemoteOperation;
+import com.owncloud.android.lib.resources.files.model.RemoteFile;
 import com.owncloud.android.operations.CopyFileOperation;
 import com.owncloud.android.operations.CreateFolderOperation;
 import com.owncloud.android.operations.DownloadFileOperation;
@@ -187,7 +189,7 @@ public class DocumentsStorageProvider extends DocumentsProvider {
     @SuppressLint("LongLogTag")
     @Override
     public ParcelFileDescriptor openDocument(String documentId, String mode, CancellationSignal cancellationSignal)
-            throws FileNotFoundException {
+        throws FileNotFoundException {
         Log.d(TAG, "openDocument(), id=" + documentId);
 
         Document document = toDocument(documentId);
@@ -196,27 +198,32 @@ public class DocumentsStorageProvider extends DocumentsProvider {
 
         OCFile ocFile = document.getFile();
         Account account = document.getAccount();
-        final User user = accountManager.getUser(account.name).orElseThrow(RuntimeException::new); // should exist
 
-        if (ocFile.isDown()) {
-            RemoteOperationResult result;
-            try {
-                result = new SynchronizeFileOperation(ocFile, null, user, true, context)
-                    .execute(document.getClient(), document.getStorageManager());
-            } catch (Exception e) {
-                throw getFileNotFoundExceptionWithCause("Error synchronizing file: " + ocFile.getFileName(), e);
-            }
-            if (result.getCode() == RemoteOperationResult.ResultCode.SYNC_CONFLICT) {
-                // TODO show a conflict notification with a pending intent that shows a ConflictResolveDialog
-                Log_OC.w(TAG, "Conflict found: " + result);
-            } else if (!result.isSuccess()) {
+        int accessMode = ParcelFileDescriptor.parseMode(mode);
+
+        boolean needsDownload = (accessMode != ParcelFileDescriptor.MODE_WRITE_ONLY);
+        if (needsDownload && ocFile.isDown()) {
+            RemoteOperationResult result = new ReadFileRemoteOperation(ocFile.getRemotePath())
+                .execute(document.getClient());
+            if (result.isSuccess()) {
+                OCFile serverFile = FileStorageUtils.fillOCFile((RemoteFile) result.getData().get(0));
+                boolean serverChanged = !serverFile.getEtag().equals(ocFile.getEtag());
+                boolean localChanged = ocFile.getLocalModificationTimestamp() > ocFile.getLastSyncDateForData();
+                if (localChanged && serverChanged) {
+                    // TODO show a conflict notification with a pending intent that shows a ConflictResolveDialog
+                    Log_OC.w(TAG, "Conflict found: " + result);
+                    // open local version for now
+                    needsDownload = false;
+                } else {
+                    needsDownload = serverChanged;
+                }
+            } else if (result.getCode() == RemoteOperationResult.ResultCode.FILE_NOT_FOUND) {
                 Log_OC.e(TAG, result.toString());
                 throw new FileNotFoundException("Error synchronizing file: " + ocFile.getFileName());
             }
-            // TODO test if this needed here
-            // block thread until file is saved
-            FileStorageUtils.checkIfFileFinishedSaving(ocFile);
-        } else {
+        }
+
+        if (needsDownload) {
             DownloadFileOperation downloadFileOperation = new DownloadFileOperation(account, ocFile, context);
             RemoteOperationResult result = downloadFileOperation.execute(document.getClient());
             if (!result.isSuccess()) {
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
Maybe I am wrong, but 
- download a file
- change it on server -> outdated on device
- open this file via DSP
- with current system: file gets synced, user sees new content
- with this approach: still content is there, as ReadFileRemoteOperation only updates metadata, but not content?

jmue:
>     * with this approach: still content is there, as ReadFileRemoteOperation only updates metadata, but not content?

ReadFileRemoteOperation is only used for checking if a file download is still needed, if so [this codepath](https://github.com/nextcloud/android/blob/cae0e77267989d5177ede1624c0a217e70888658/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java#L226-L234) is taken.

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
```diff
@@ -196,27 +198,32 @@ public ParcelFileDescriptor openDocument(String documentId, String mode, Cancell
...
+                boolean localChanged = ocFile.getLocalModificationTimestamp() > ocFile.getLastSyncDateForData();
+                if (localChanged && serverChanged) {
+                    // TODO show a conflict notification with a pending intent that shows a ConflictResolveDialog
+                    Log_OC.w(TAG, "Conflict found: " + result);
+                    // open local version for now
+                    needsDownload = false;
+                } else {
+                    needsDownload = serverChanged;
+                }
+            } else if (result.getCode() == RemoteOperationResult.ResultCode.FILE_NOT_FOUND) {
```
grote, line 220: What if the result was not a success, but had a different error than this one? Are we just silently ignoring it?

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
```diff
@@ -196,27 +198,32 @@ public ParcelFileDescriptor openDocument(String documentId, String mode, Cancell
...
-            if (result.getCode() == RemoteOperationResult.ResultCode.SYNC_CONFLICT) {
-                // TODO show a conflict notification with a pending intent that shows a ConflictResolveDialog
-                Log_OC.w(TAG, "Conflict found: " + result);
-            } else if (!result.isSuccess()) {
+        int accessMode = ParcelFileDescriptor.parseMode(mode);
+
+        boolean needsDownload = accessMode != ParcelFileDescriptor.MODE_WRITE_ONLY;
+        if (needsDownload && ocFile.isDown()) {
+            RemoteOperationResult result = new ReadFileRemoteOperation(ocFile.getRemotePath())
+                .execute(document.getClient());
```
grote, line 207: Do we always need to do that? Is there another way to find out if the local file can be considered up-to-date?

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
```diff
@@ -196,27 +198,32 @@ public ParcelFileDescriptor openDocument(String documentId, String mode, Cancell
...
                 throw new FileNotFoundException("Error synchronizing file: " + ocFile.getFileName());
             }
-            // TODO test if this needed here
-            // block thread until file is saved
-            FileStorageUtils.checkIfFileFinishedSaving(ocFile);
-        } else {
+        }
+
+        if (needsDownload) {
             DownloadFileOperation downloadFileOperation = new DownloadFileOperation(account, ocFile, context);
```
grote, line 227: This is a logic change. Before, we were always downloading the file when it was *not* down (available locally I guess). Now, we only do this when want to read the file which sounds plausible. I am just flagging this, because it changes the behavior.

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
```diff
@@ -196,27 +198,32 @@ public ParcelFileDescriptor openDocument(String documentId, String mode, Cancell
...
+                } else {
+                    needsDownload = serverChanged;
+                }
+            } else if (result.getCode() == RemoteOperationResult.ResultCode.FILE_NOT_FOUND) {
                 Log_OC.e(TAG, result.toString());
                 throw new FileNotFoundException("Error synchronizing file: " + ocFile.getFileName());
             }
-            // TODO test if this needed here
-            // block thread until file is saved
-            FileStorageUtils.checkIfFileFinishedSaving(ocFile);
```
grote, line 218: Did you test/check/verify that this is not needed anymore when using the `ReadFileRemoteOperation` and `FileStorageUtils.fillOCFile()`?

grote (commented):
Thanks @jmue for picking this up! I left some questions, but am not an expert on the Nextcloud bits.

Did you run the integration tests I had added? (The CI is flaky unfortunately)
Would maybe be nice to add another test to them that shows the improved behavior, aka not serving stale files anymore?

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
tobiasKaminsky: We could use https://github.com/nextcloud/android-library/blob/master/src/main/java/com/owncloud/android/lib/resources/files/CheckEtagRemoteOperation.java

which only checks eTag, so this should be a lot faster.

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
tobiasKaminsky: We then should exit with any valid return code, 404?

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
tobiasKaminsky: This is in openDocument() function.
When / how is this called if not hen the user/app wants to read it?

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
```diff
@@ -196,27 +198,32 @@ public ParcelFileDescriptor openDocument(String documentId, String mode, Cancell
...
-            } catch (Exception e) {
-                throw getFileNotFoundExceptionWithCause("Error synchronizing file: " + ocFile.getFileName(), e);
-            }
-            if (result.getCode() == RemoteOperationResult.ResultCode.SYNC_CONFLICT) {
-                // TODO show a conflict notification with a pending intent that shows a ConflictResolveDialog
-                Log_OC.w(TAG, "Conflict found: " + result);
-            } else if (!result.isSuccess()) {
+        int accessMode = ParcelFileDescriptor.parseMode(mode);
+
+        boolean needsDownload = accessMode != ParcelFileDescriptor.MODE_WRITE_ONLY;
```
tobiasKaminsky, line 204: If this is open with write_only and file is downloaded, then it should also be checked for changed content/eTag, or?

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
jmue: I've changed that.

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
jmue: From my perspective this is a leftover of previous SFO use. The idea was to spin up a thread, let the SFO synchronize the file contents (asynchronously) und wait for cmpletion with that lines. Unfortunately that didn't work.

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
jmue: > 
> 
> This is in openDocument() function.
> When / how is this called if not hen the user/app wants to read it?

openDocument() is also used to write to a file.

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
jmue: > Did you run the integration tests I had added? (The CI is flaky unfortunately)
Yes

> Would maybe be nice to add another test to them that shows the improved behavior, aka not serving stale files anymore?
Done

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
```diff
@@ -268,6 +249,22 @@ public ParcelFileDescriptor openDocument(String documentId, String mode, Cancell
         }
     }
 
+    private boolean hasServerChange(OCFile ocFile, Account account) throws FileNotFoundException {
```
grote, line 252: We could also pass in the document here, since it has OCFile and Account.

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
tobiasKaminsky: Yes, this is why I ask…We need to make sure that we do not overwrite new content of server.

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
jmue: Changed.

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
jmue: Yes you're right here. We have always to sync the file metadata here, even if we open in write only mode. Otherwise we will get conflicts on upload if the server etag has changed. So I changed that for now to the old logic (download file content even if we open write only).

jmue: For efficiency reanson we can think of doing a ReadFileRemoteOperation when we're within the needsDownload condition and are opening write only. So:

```
if (needsDownload) {
  if (writeOnly) {
    <ReadFileRemoteOperation + storeMetadata>
  } else {
   <DownloadFileOperation  + storeMetadata + storeContents>
  }
}

```

But that would give us more clutter.

tobiasKaminsky: It is not only about metadata, but also about content.
Imagine that you have 
- a file with "123"
- download it to mobile
- change it on server to "1234"
- open it with writeOnly
- without download you would write 5 to 123, which then results in data loss
- as soon as etag has changed, we must download/sync (maybe even with conflict handling) the server content, so that adding 5 will then be the correct "12345".

jmue: Correct, I wasn't aware of opening write+append, I had only write+truncate in my mind. But as downloading in any case is done now, there is no further source code change necessary.

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
```diff
@@ -263,6 +248,23 @@ public ParcelFileDescriptor openDocument(String documentId, String mode, Cancell
...
+            .execute(document.getAccount(), context);
+        switch (result.getCode()) {
+            case ETAG_CHANGED:
+                return true;
+            case ETAG_UNCHANGED:
+                return false;
+            case FILE_NOT_FOUND:
+            default:
+                Log_OC.e(TAG, result.toString());
+                throw new FileNotFoundException("Error synchronizing file: " + ocFile.getFileName());
```
grote, line 264: What happens if we read/write a file again that was not yet uploaded to the server. Does this return `FILE_NOT_FOUND` here and thus throws?

Do all other tests from `DocumentsStorageProviderIT` still pass?

jmue: Good catch, so we want to return `false `for `FILE_NOT_FOUND`? This will solve that case.

But suppose we have:
- client opens for write, upload file to server
- server deletes file
- client opens for write again
What is our expectation here?

Yes, all other test pass on my side, same for the integration bot.

tobiasKaminsky: in this case the client should retrieve file on server and sees that it not available and thus return file not found.

grote: Do you mean throw a `FileNotFoundException` and thus fail the write operation? As uploads are done async one after the other, it is probably a common scenario when using the DocumentsProvider programmatically, that the file hasn't been uploaded, yet, but is written to or read from again.

I was asking about passing tests, because this happens here as well: https://github.com/nextcloud/android/blob/master/src/androidTest/java/com/owncloud/android/providers/DocumentsStorageProviderIT.kt#L135

But thinking about this again, creating and writing/reading are two separate operations, and creating happens synchronously, so we should not run into the case where `FILE_NOT_FOUND` is `false`.

grote:
I have tested this branch (rebased on current master HEAD) with seedvault and production data. Not only did I not find any new issues (with this particular usage pattern), but also get less connection errors.

src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
jmue: So no further change is necessary here.

jmue:
@tobiasKaminsky Are there any open demands for code changes left? I don't know how I should further proceed here. From my perspective the code behavior at the reviewed sections is the same as on the master branch. Please guide me what is needed from my side here.

tobiasKaminsky:
I just overlooked it.
Thanks for bringing this up again.
@grote thanks for testing this with seedvault :+1:

tobiasKaminsky (approved):

PR #6997, opened 2020-09-23
Re-use the existing foreground notification for FileUploader progress notifications

### LINKED ISSUE #4998: Suppress Up/Download Notifications when accessed through DocumentsProvider

When Nextcloud's storage is accessed through Android's storage access framework programmatically, lots of files might be uploaded. This causes a flood of upload and download notifications popping up all the time.

Insofar this is related to #261 which would already help

If notifications could be suppressed for access that happens via the DocumentsProvider this would be perfect as apps accessing it usually provide their own progress reporting.

Maybe there could also be other mechanisms to suppress the notifications such as the presence of a special file in a parent folder like `.nomedia`.

An alternative could be to only show a single notification for the up/downloads like described in #261.

grote:
Closes #4998

### DIFF
--- a/src/main/java/com/owncloud/android/files/services/FileUploader.java
+++ b/src/main/java/com/owncloud/android/files/services/FileUploader.java
@@ -255,7 +255,7 @@ public class FileUploader extends Service
      */
     private void resurrection() {
         // remove stucked notification
-        mNotificationManager.cancel(R.string.uploader_upload_in_progress_ticker);
+        mNotificationManager.cancel(FOREGROUND_SERVICE_ID);
     }
 
     /**
@@ -269,6 +269,9 @@ public class FileUploader extends Service
         mServiceHandler = null;
         mServiceLooper.quit();
         mServiceLooper = null;
+        if (mNotificationManager != null) {
+            mNotificationManager.cancel(FOREGROUND_SERVICE_ID);
+        }
         mNotificationManager = null;
 
         // remove AccountsUpdatedListener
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -240,24 +240,13 @@ public void onCreate() {
...
     }
 
-    /**
-     * Service clean-up when restarted after being killed
-     */
-    private void resurrection() {
-        // remove stucked notification
-        mNotificationManager.cancel(R.string.uploader_upload_in_progress_ticker);
-    }
-
```
grote, line 260: This does not seem to be necessary anymore as we now cancel the notification when the service is destroyed.

Was this maybe a relic from before using a foreground service or did you still see that the service was killed without calling `onDestroy()`?

src/main/java/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -748,8 +740,6 @@ private void notifyUploadResult(UploadFileOperation upload, RemoteOperationResul
             mNotificationManager = (NotificationManager) getSystemService(NOTIFICATION_SERVICE);
         }
 
-        mNotificationManager.cancel(R.string.uploader_upload_in_progress_ticker);
-
```
grote, line 752: First, I was replacing the progress notification with the general foreground service notification here (`mNotification`), but that doesn't seem to be necessary as we either cancel it when the service gets destroyed or replace it with a new notification when starting the next upload.

src/main/java/com/owncloud/android/files/services/FileUploader.java
tobiasKaminsky: I think this is to prevent edge cases like 
- upload ongoing, notification is shown
- app is killed
- notification is still shown

At least I remember something like that. Maybe it is already better handled by newer Android versions…

grote: Probably safer to leave it in then. I re-added it back.

tobiasKaminsky (approved):

PR #7050, opened 2020-10-05
Ensure that upload notification gets cancelled

grote:
This is a follow up to #6997.

Normally, the notification gets cancelled when the service is destroyed. But that does not always happen in practice, so we also cancel it after every result.


AndyScherzinger (approved):

grote:
@tobiasKaminsky after using this branch, I haven't noticed any stuck notifications form FileUploader anymore.

PR #7110, opened 2020-10-19
DocumentsStorageProviderIT: testServerChangedFileContent fails

tobiasKaminsky:
@jmue @grote 

This test fails now with:
> ComparisonFailure: expected:<[new] content> but was:<[initial] content>

So it seems that we do not refresh doc storage after updating.
I am unsure when this changed and what the impact of this is?

### Testing 

### DIFF
--- a/src/androidTest/java/com/owncloud/android/providers/DocumentsStorageProviderIT.kt
+++ b/src/androidTest/java/com/owncloud/android/providers/DocumentsStorageProviderIT.kt
@@ -202,6 +202,7 @@ class DocumentsStorageProviderIT : AbstractOnServerIT() {
         putMethod.releaseConnection() // let the connection available for other methods
 
         // read back content bytes
-        assertReadEquals(content2, contentResolver.openInputStream(file1.uri))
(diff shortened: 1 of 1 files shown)


jmue:
I's hard to say if this is a serious issue. Maybe this is a timing one. As the upload is performed asynchronous, it is possible that the file is created and the upload is queued on the nextcloud client. The server change could be faster in this scenario and the nextcloud upload is executed afterwards. So there will be 'initial content' at the check. Currently I don't see how to perform the steps in intended order. Maybe do some sleep after upload?

grote:
A sleep can cause flaky tests and there's enough of those already ;)

Maybe if the diagnosis is correct, make a while(notUploaded) sleep; loop instead and periodically check if the file was uploaded, if there's no other way to get informed about completed uploaded?

jmue:
> loop instead and periodically check if the file was uploaded

What about the eTag, we do have a valid eTag after file creation and could check for changed value after the content upload is queued.

jmue:
The test ran successful on my side. Nevertheless I added a check to verify that the requested upload was executed. See my pr #7117

AlvaroBrey:
Outdated; test works properly now, both locally and in CI.

Issue #7116, opened 2020-10-20
After setting up account using QR code, app ends up in broken state


Derkades:
### Steps to reproduce
1. Launch app for the first time
2. Log in using QR code

After logging in, I end up in this view: https://user-images.githubusercontent.com/15892014/96621248-b1e25b80-1308-11eb-861d-54789d8359e1.png

The hamburger menu does nothing. When I re-open the app I am at the standard file list and everything works fine.

### Can you reproduce this problem on https://try.nextcloud.com?
Did not try, server does not seem relevant. If you need me to try please let me know

### Environment data
Android version: 11

Device model: Pixel 3a

Stock or customized system: stock

Nextcloud app version: Nextcloud QA 16210

Nextcloud server version: 19.0.4

Reverse proxy: yes, nginx

#### Web server error log
```
N/A
```

#### Nextcloud log (data/nextcloud.log)
```
I was not able to find it
```
**NOTE:** Be super sure to remove sensitive data like passwords, note that everybody can look here! You can use the Issue Template application to prefill some of the required information: https://apps.nextcloud.com/apps/issuetemplate

AndyScherzinger:
@tobiasKaminsky I can reproduce this with master too

PR #7124, opened 2020-10-21
use custom SSO value for authTokenType -> to detect if account is added by SSO

tobiasKaminsky:
Fix: https://github.com/nextcloud/android/issues/7116
SSO: https://github.com/nextcloud/Android-SingleSignOn/pull/270

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -707,7 +707,8 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
         if (extras == null) {
             return false;
         } else {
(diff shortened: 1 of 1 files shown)


David-Development:
@tobiasKaminsky Doesn't this break backwards compatibility as older SSO libraries don't report that value yet?

tobiasKaminsky:
> @tobiasKaminsky Doesn't this break backwards compatibility as older SSO libraries don't report that value yet?

It is used as "onlyAdd" value, which if is true finish AuthenticatorActivity after adding an new account.
As previously it was checked wrongly, it did something strange on Files app, but worked on 3rd party libs.
So yes, it will break it, but which only means that 3rd party libs without a new SSO release have to manually close Files app after adding a new account.

tl;dr: it is safe to merge

AndyScherzinger (approved):

PR #7135, opened 2020-10-24
l10n: Cloud name with a capital letter

Valdnet:
### Testing 

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -473,7 +473,7 @@
     <string name="share_with_title">Share with…</string>
     <string name="share_via_link_unset_password">Unset</string>
 
-    <string name="share_search">Name, federated cloud ID or email address…</string>
+    <string name="share_search">Name, Federated Cloud ID or email address…</string>
 
     <string name="share_group_clarification">%1$s (group)</string>
     <string name="share_remote_clarification">%1$s (remote)</string>
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -473,7 +473,7 @@
     <string name="share_with_title">Share with…</string>
     <string name="share_via_link_unset_password">Unset</string>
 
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
cc also @jancborchardt - the server (at least latest Nc20 stable doesn't use a capitalized letter)

src/main/res/values/strings.xml
```diff
@@ -473,7 +473,7 @@
     <string name="share_with_title">Share with…</string>
     <string name="share_via_link_unset_password">Unset</string>
 
-    <string name="share_search">Name, federated cloud ID or email address…</string>
+    <string name="share_search">Name, Federated Cloud ID or email address…</string>
```
tobiasKaminsky, line 476: Not sure if this is correct, as it not a name, but a description, like "green house" vs. "the White House".
It basically means that one can type an identifier of an user who is not on the same server, but on another/remote Netcloud server…

Valdnet: > Not sure if this is correct, as it not a name, but a description

If it's not a name then it should be regulated in other applications where "federated cloud" is taken as the name and capitalized.

Valdnet:
I added a space before the ellipsis.
@tobiasKaminsky I checked the Federation Cloud information and in my opinion it should be capitalized as the name. I am waiting for a decision on this to improve spelling in all Nextcloud applications.

src/main/res/values/strings.xml
Valdnet: @tobiasKaminsky See: https://github.com/nextcloud/server/pull/23671#pullrequestreview-518553984

AndyScherzinger: Thanks for taking care of the clarification 🎉

AndyScherzinger (approved):

tobiasKaminsky:
/backport to stable-3.13
(otherwise Transifex will fail)

PR #7164, opened 2020-10-29
Test also for connected networks

### LINKED ISSUE #7129: For automated uploads, when Ethernet is connected, and Wi-Fi is required, "Waiting for free Wi-Fi" is displayed
### Steps to reproduce
1. Set up some Automated uploads in Nextcloud Android app, and require Wi-Fi
2. Connect an Ethernet adapter over USB, and establish a connection
3. Have some content to upload

### Expected behaviour
- The upload progresses over the cable connection

### Actual behaviour
- There is a "Waiting for free Wi-Fi" sign (rough translation from Czech "Čekání na bezplatnou Wi-Fi").

image-upload

My Samsung disconnects the Wi-Fi as soon as the cable is connected.

I know that, technically, this behavior is correct - with anything but Wi-Fi, the upload refuses to progress further. But in my POV, Ethernet should be treated the same as Wi-Fi, if that is possible.

### Environment data
Android version: 10

Device model: SM-J610FN (Samsung Galaxy J6+)

Nextcloud app version: 3.13.1

Nextcloud server version: 19.0.4

tobiasKaminsky:
Fix #7129 

TODO
- [x] let user test if it helps
- [ ] think about wording

### Testing 

### DIFF
--- a/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
+++ b/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
@@ -111,16 +111,17 @@ class ConnectivityServiceImpl implements ConnectivityService {
         if (networkInfo != null) {
             boolean isConnected = networkInfo.isConnectedOrConnecting();
             boolean isMetered = ConnectivityManagerCompat.isActiveNetworkMetered(platformConnectivityManager);
-            boolean isWifi = networkInfo.getType() == ConnectivityManager.TYPE_WIFI || isAnyOtherNetworkWifi();
+            boolean isWifi = networkInfo.getType() == ConnectivityManager.TYPE_WIFI || hasNonCellularConnectivity();
             return new Connectivity(isConnected, isMetered, isWifi, null);
         } else {
             return Connectivity.DISCONNECTED;
         }
     }
 
-    private boolean isAnyOtherNetworkWifi() {
+    private boolean hasNonCellularConnectivity() {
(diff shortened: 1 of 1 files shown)


PaulosV:
Tested on Samsung Galaxy J6+, this PR has fixed the issue.

AndyScherzinger (approved):

src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
```diff
@@ -143,7 +143,8 @@ public Connectivity getConnectivity() {
 
     private boolean isAnyOtherNetworkWifi() {
```
AndyScherzinger, line 144: maybe
```suggestion
    private boolean hasNonCellularConnectivity() {
```

tobiasKaminsky:
image

@AndyScherzinger do you have an idea about wording?
"Only upload on unmetered Wi-Fi and ethernet is a bit too long and complicated and technical…

AndyScherzinger:
Maybe we just call it unmetered connection? We can't use network since cellular is also a network

tobiasKaminsky:
> Maybe we just call it unmetered connection? We can't use network since cellular is also a network

Hm. But connection is also cellular…and this is what it is not about.
So this is too generic, or?

Maybe we can just leave it like this? As it is rather a corner case……

AndyScherzinger:
> Maybe we can just leave it like this?

Perfectly fine with me, yes. 
It seems to be an edge case and if it "just works" then we likely won't get any tickets that is also works on ethernet.

PR #7308, opened 2020-11-12
Get last modified time from SAF files put it on the uploaded file

### LINKED ISSUE #5660: Get last modified time from SAF files?
I would like the last modified time on the server to also be set correctly when uploading files from other apps.

According to [this stackoverflow posting](https://stackoverflow.com/questions/35744654/storage-access-framework-set-last-modified-date-of-local-documentfile) (reading part), it might be possible to use `leakedContentResolver.query(currentUi, null, null, null, null)` and `cursor.getLong(cursor.getColumnIndexOrThrow(DocumentsContract.Document.COLUMN_LAST_MODIFIED))` to get the last modified time of a "Document" in SAF.

One could possibly use this in `src/main/java/com/owncloud/android/ui/asynctasks/CopyAndUploadContentUrisTask.java`, method `doInBackground()` and then either set the mtime of the temporary file or give that time to the final upload function.

Could this work? I am not an Android Developer, I have just researched a bit. And maybe this is all complete nonsense?

* If it is possible to get the last modified time from a "file from an app", please try to set it on the server during upload.
* If that's not possible, then possibly document this in a place that search engines can

hochwasser:
Get the modified time from SAF files and set it on the cache file.
The uploaded then sets the correct modified time ob the server.
This fixes #5660

### Testing 
If a test is needed I need help.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/asynctasks/CopyAndUploadContentUrisTask.java
+++ b/src/main/java/com/owncloud/android/ui/asynctasks/CopyAndUploadContentUrisTask.java
@@ -24,8 +24,10 @@ package com.owncloud.android.ui.asynctasks;
 import android.accounts.Account;
 import android.content.ContentResolver;
 import android.content.Context;
+import android.database.Cursor;
 import android.net.Uri;
 import android.os.AsyncTask;
+import android.provider.DocumentsContract;
 import android.widget.Toast;
 
 import com.owncloud.android.R;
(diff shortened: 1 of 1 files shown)


ChristianTacke (approved):
Looks fine to me.

src/main/java/com/owncloud/android/ui/asynctasks/CopyAndUploadContentUrisTask.java
```diff
@@ -143,6 +145,19 @@ protected ResultCode doInBackground(Object[] params) {
                 currentUri = uris[i];
                 currentRemotePath = remotePaths[i];
 
+                long lastModified = 0;
+                try (Cursor cursor = leakedContentResolver.query(currentUri,
+                                                                 null,
+                                                                 null,
+                                                                 null,
+                                                                 null)) {
+                    if (cursor.moveToFirst()) {
```
tobiasKaminsky, line 154:
```suggestion
                    if (cursor != null && cursor.moveToFirst()) {
```

As cursor can be null.

tobiasKaminsky (approved):

AndyScherzinger (approved):

AndyScherzinger:
(maybe rather 3.14.1 than 3.14.0 - depending on if we might need an RC3 for 3.14.0)

PR #7340, opened 2020-11-17
Directly start foreground service

### LINKED ISSUE #2010: Context.startForegroundService() did not then call Service.startForeground()
Highest number of reports for ANRs since 2.0.0 in the Dev console - Android 8 only - >5K reports in 60 days

Reason I found google-ing: https://stackoverflow.com/questions/44425584/context-startforegroundservice-did-not-then-call-service-startforeground

> From Google's docs on Android 8.0 behavior changes:
> The system allows apps to call Context.startForegroundService() even while the app is in the background. However, the app must call that service's startForeground() method within five seconds after the service is created.

--> https://developer.android.com/about/versions/oreo/android-8.0-changes.html

android.app.RemoteServiceException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{fb59be8 u0 com.nextcloud.client/com.owncloud.android.services.OperationsService}
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1745)
at android.os.Looper.loop(Looper.java:193)
at android.app.ActivityThread.main(ActivityThread.java:6718)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)

ID: com.nextcloud.client
Version: 30080190
Build flavor: generic

Brand: lge
Device: joan
Model: LG-H932
Id: PQ3A.190605.003
Product: joan_tmo_us

SDK: 28
Release: 9
Incremental: eng.markus.20190610.202830

### LINKED ISSUE #6905: Crash due to PlayerService
### Steps to reproduce
1. delete a file
2. see 
```
2020-09-08 07:20:21.303 com.nextcloud.client E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.nextcloud.client, PID: 10750
    android.app.RemoteServiceException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{a08f83f u0 com.nextcloud.client/.media.PlayerService}
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1945)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
```

@ezaquarii this happens on latest dev, and it seems that it is from your foreground fix:
Upon file removed event, FileDisplayActivity onRemoveFileOperationFinish is called and this calls tryStopPlaying.

PR was #6798

tobiasKaminsky:
Fix #2010 
Fix #6905

This is not done at all, but it prevents from crashing on my test device.
@ezaquarii for some reason the startForeground in PlayerService is not called, when starting a video.
Thus it get killed by Android after a specific time.

### Testing 

### DIFF
--- a/src/main/java/com/nextcloud/client/media/PlayerService.kt
+++ b/src/main/java/com/nextcloud/client/media/PlayerService.kt
@@ -35,7 +35,6 @@ import com.owncloud.android.datamodel.OCFile
 import com.owncloud.android.ui.notifications.NotificationUtils
 import com.owncloud.android.utils.ThemeUtils
 import dagger.android.AndroidInjection
-import java.lang.IllegalArgumentException
 import javax.inject.Inject
 
 class PlayerService : Service() {
--- a/src/main/java/com/nextcloud/client/media/PlayerService.kt
+++ b/src/main/java/com/nextcloud/client/media/PlayerService.kt
@@ -35,6 +35,7 @@ import com.owncloud.android.datamodel.OCFile
 import com.owncloud.android.ui.notifications.NotificationUtils
 import com.owncloud.android.utils.ThemeUtils
 import dagger.android.AndroidInjection
+import java.util.Locale
 import javax.inject.Inject
 
 class PlayerService : Service() {
@@ -46,6 +47,7 @@ class PlayerService : Service() {
         const val EXTRA_START_POSITION_MS = "START_POSITION_MS"
         const val ACTION_PLAY = "PLAY"
         const val ACTION_STOP = "STOP"
+        const val ACTION_TOGGLE = "TOGGLE"
         const val ACTION_STOP_FILE = "STOP_FILE"
     }
 
@@ -73,7 +75,7 @@ class PlayerService : Service() {
         }
 
         override fun onStop() {
-            stopForeground(true)
+            stopServiceAndRemoveNotification(null)
         }
 
         override fun onError(error: PlayerError) {
@@ -96,10 +98,20 @@ class PlayerService : Service() {
         player = Player(applicationContext, clientFactory, playerListener, audioManager)
         notificationBuilder = NotificationCompat.Builder(this)
         notificationBuilder.color = ThemeUtils.primaryColor(this)
+
         val stop = Intent(this, PlayerService::class.java)
         stop.action = ACTION_STOP
         val pendingStop = PendingIntent.getService(this, 0, stop, 0)
-        notificationBuilder.addAction(0, "STOP", pendingStop)
+        notificationBuilder.addAction(0, getString(R.string.player_stop).toUpperCase(Locale.getDefault()), pendingStop)
+
+        val toggle = Intent(this, PlayerService::class.java)
+        toggle.action = ACTION_TOGGLE
+        val pendingToggle = PendingIntent.getService(this, 0, toggle, 0)
+        notificationBuilder.addAction(
+            0,
+            getString(R.string.player_toggle).toUpperCase(Locale.getDefault()),
+            pendingToggle
+        )
     }
 
     override fun onBind(intent: Intent?): IBinder? {
@@ -111,8 +123,17 @@ class PlayerService : Service() {
             ACTION_PLAY -> onActionPlay(intent)
             ACTION_STOP -> onActionStop()
             ACTION_STOP_FILE -> onActionStopFile(intent.extras)
+            ACTION_TOGGLE -> onActionToggle()
         }
         return START_NOT_STICKY
+    }
+
+    private fun onActionToggle() {
+        if (player.isPlaying) {
+            player.pause()
+        } else {
+            player.start()
+        }
     }
 
     private fun onActionPlay(intent: Intent) {
@@ -125,12 +146,13 @@ class PlayerService : Service() {
     }
 
     private fun onActionStop() {
-        player.stop()
+        stopServiceAndRemoveNotification(null)
     }
 
     private fun onActionStopFile(args: Bundle?) {
         val file: OCFile = args?.getParcelable(EXTRA_FILE) ?: throw IllegalArgumentException("Missing file argument")
-        player.stop(file)
+
+        stopServiceAndRemoveNotification(file)
     }
 
     private fun startForeground(currentFile: OCFile) {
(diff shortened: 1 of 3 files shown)


ezaquarii:
@tobiasKaminsky this is very strange edge case as video player does not use PlayerService at all - it depends on a media player in VideoView.

It sounds to me like the player service might be receiving unintended events from GUI.

InfamousUser:
Tried it, it does not seem to crash, but I think crashing for me is caused by autoupload, which doesn't seem to want to start on this app. I did turn off doze for it.

InfamousUser:
Ok, autoupload started working, it took a while to initialize and well...

The problem is fixed. The problem being the app crash, now the app doesn't crash but instead the whole app or Android system freezes and/or ANR dialogue is shown. This fixed one of the symptoms apparently. Looks like the underlying problem was just exacerbating the crash case, I did not know there was an underlying problem in the first place.

Okay, so the problem for me is app freezing.
Also the ANR dialogue is different now, there's only 2 options whereas there were 3 before.

P.S. Can't even open the app anymore - it stops responding on the initial loading screen.

src/main/java/com/nextcloud/client/media/PlayerService.kt
```diff
@@ -101,6 +101,13 @@ class PlayerService : Service() {
         stop.action = ACTION_STOP
         val pendingStop = PendingIntent.getService(this, 0, stop, 0)
         notificationBuilder.addAction(0, "STOP", pendingStop)
+
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.O) {
```
ezaquarii, line 105: This is nasty. But if we need this workaround, I'd put it in `onCreate()`.

tobiasKaminsky:
> @tobiasKaminsky this is very strange edge case as video player does not use PlayerService at all - it depends on a media player in VideoView.
> 
> It sounds to me like the player service might be receiving unintended events from GUI.

I was confused by that first, too.
When you click on an image it opens PreviewImageFragment
When you click on an audio/video it opens PreviewMediaFragment

PreviewMediaFragment then can do both: play audio or video.
If audio plays and you click "back", then audio continues to play in background (via PlayerService).
I guess that this is then the reason why PlayerService is started.

When you then click on video in PreviewMediaFragment it opens up PreviewVideoActivity to display video in fullscreen…

src/main/java/com/nextcloud/client/media/PlayerService.kt
tobiasKaminsky: True, this is not meant to be the final version.
I just did a quick PR for testing, so that user can give feedback.

tobiasKaminsky:
Reason why PlayerService is started is:
https://github.com/nextcloud/android/blob/77751da6b8cc59dec2bf59941c407adc62aa862c/src/main/java/com/owncloud/android/ui/preview/PreviewMediaFragment.java#L326-L329

Which is somehow needed:
- play audio
- go back -> audio continues in background
- start a video

tobiasKaminsky:
@InfamousUser can you test it again, please?

tobiasKaminsky:
@ezaquarii I found the cause and hopefully a proper fix. Can you review it? :-)

InfamousUser:
> @InfamousUser can you test it again, please?

> Ok, autoupload started working, it took a while to initialize and well...
> 
> The problem is fixed. The problem being the app crash, now the app doesn't crash but instead the whole app or Android system freezes and/or ANR dialogue is shown. This fixed one of the symptoms apparently. Looks like the underlying problem was just exacerbating the crash case, I did not know there was an underlying problem in the first place.
> 
> Okay, so the problem for me is app freezing.
> Also the ANR dialogue is different now, there's only 2 options whereas there were 3 before.
> 
> P.S. Can't even open the app anymore - it stops responding on the initial loading screen.

Probably same as above, I can't test because of the original problem. The app is frozen solid on startup for 80 seconds (loading screen), then displays frozen main interface.

tobiasKaminsky:
> Probably same as above, I can't test because of the original problem. The app is frozen solid on startup for 80 seconds (loading screen), then displays frozen main interface.
> 
> It also locks the Android interface half the times for up to half a minute, that shouldn't be possible, probably a bug in Android.

Can you create another issue for this? 
Please also try to get logcat.

InfamousUser:
Not sure whether you wanted the log here or in the new issue, but here it is here just in case. Also not sure this is what you mean by logcat, never used it before, this is what I got by finding an app called that.

[Nextcloud.txt](https://github.com/nextcloud/android/files/5567844/Nextcloud.txt)

InfamousUser:
New issue: https://github.com/nextcloud/android/issues/7378

src/main/java/com/owncloud/android/ui/preview/PreviewMediaFragment.java
```diff
@@ -324,7 +324,9 @@ public void onStart() {
                 mMultiListContainer.setVisibility(View.GONE);
                 mPreviewContainer.setVisibility(View.VISIBLE);
             } else if (MimeTypeUtil.isVideo(file)) {
-                stopAudio();
+                if (mMediaPlayerServiceConnection.isConnected() && mMediaPlayerServiceConnection.isPlaying()) {
+                    stopAudio();
+                }
```
ezaquarii, line 329: This looks like a solution for this case.

However, because `stopAudio()` results in `Intent` being sent to a service and thus it starts the service, the call is simply not safe to use. I think we should internally call `Service.stopSelf()` if there is nothing to play as well.

ezaquarii (approved):
This looks like a solution.

src/main/java/com/owncloud/android/ui/preview/PreviewMediaFragment.java
tobiasKaminsky: In PreviewMediaFragment onStart() I tried to bind to an existing player service.
Reason is:
- start a long audio file
- start a video
- audio stops, video sound is played, but audio notification is still there (and cannot be resumed)

So my quick idea was to bind to audio service when starting video and stop audio service.
But this does not work.

tobiasKaminsky: Fixed with adding a "play/pause" button.

tobiasKaminsky:
:exclamation: Rebased :exclamation:

PR #7370, opened 2020-11-19
Modify passcode setting layout alignment

Kuuuna98:
Signed-off-by: Kuuuna98 <yunaghgh@naver.com>

### Testing 
1. Click Settings on the menu.
2. Set the passcode in the APP passcode.
3. A screen will appear to set passcode.

As I changed the passcode entry error message, the central alignment seems to have changed to left alignment.
Is it intended that the title message and cancel button changed to the left alignment when setting the passcode?
If not intended, change back to central alignment and PR!

### DIFF
--- a/src/main/res/layout/passcodelock.xml
+++ b/src/main/res/layout/passcodelock.xml
@@ -34,6 +34,7 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:gravity="center_horizontal"
+            android:layout_gravity="center_horizontal"
             android:text="@string/pass_code_enter_pass_code"
             android:textColor="@color/text_color"
             android:textSize="@dimen/two_line_primary_text_size" />
@@ -42,6 +43,7 @@
             android:id="@+id/explanation"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
+            android:layout_gravity="center_horizontal"
             android:gravity="center_horizontal"
             android:text="@string/pass_code_configure_your_pass_code_explanation"
             android:textAppearance="@android:style/TextAppearance.Small"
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
@AndyScherzinger @nextcloud/designers what do you think?
Before -> after
imageimage

If I recall correct, it is more common to have everything left/start aligned and not centered.

AndyScherzinger:
Tricky, in general the layouting should be rtl/ltr for larger content in this case I think centered would be okay-ish while I would even say we should probably layout it like a dialog, so header left aligned, pin would be fine centered I guess and cancel should be right aligned. Even more it be nice to have an all dark-grey/black background and have the elements in a white card-view with round corners. This way it would mimic a dialog with a non-semi-transparent background (in contrast to a standard dialog that doesn't need to "protect" the content").

jancborchardt:
Looks nice, just small points:
- Agree with @AndyScherzinger on that it would look better as a dialog instead of full-screen. But I would indeed align the cancel button centered too
- The cancel button should not be styled as "primary" action since it’s not the primary action (that one is getting triggered automatically on filling all 4 fields)
- Super detail: Instead of asterisks, can we have dots as placeholders? Like in other pin fields, looks simpler. :) Guess would be this character: `•`

AndyScherzinger:
I agree with @jancborchardt and will move the remarks to a new issue, rebase this one so it can be merged :) (in case I am allowed to via the PR permissions.. we'll see)

AndyScherzinger:
❗ rebased, waiting for the CI and will push updated screenshot tests afterwards and then merge.

Thanks for your work @Kuuuna98 ❤️

AndyScherzinger (approved):

ma12-co:
I'm late to this but I have a remark too: In cases like this shouldn't buttons be located at the bottom of the screen, where your thumb is??

AndyScherzinger:
TL;DR it should all be just fine (in the end, after another PR)

Not really late to the party as a) change has already been merged b) it is an ongoing overhaul.

* When putting in the PIN a keyboard would be displayed so the distance wouldn't be that large
* we discussed to change this to mimic a dialog which would then put the cancel button to the bottom/right (according to Material Design guidelines)

PR #7372, opened 2020-11-19
Closed the app when press back button from App passcode screen

### LINKED ISSUE #7343: Response when press back button from App password screen
### Is your feature request related to a problem? 
After set app password, I can't exit the app by pressing the Back button on the password screen.
I thought there was an error in the app because there was no response even when I pressed the back button.

Android Naxtcloud app v. 3.13.1.
Android: 8
Device: Samsung Galaxy S7

Like the home screen of the app, it would be nice to leave the app by pressing the back button on the password screen. 
Or, when you press the Back button on the password screen, I would like to see a message telling you to press the Home button or the Menu button.

Kuuuna98:
Signed-off-by: Kuuuna98 <yunaghgh@naver.com>

### Testing 

1. Set APP passcode
2. Press the back button on the passcode screen that you see when you run the app.

Result: Leave the app.

Fix #7343 Issue implemented in code. Please check.😊

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java
@@ -23,6 +23,7 @@
 package com.owncloud.android.ui.activity;
 
 import android.content.Intent;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.SystemClock;
 import android.text.Editable;
@@ -365,9 +366,9 @@ public class PassCodeActivity extends AppCompatActivity implements Injectable {
     @Override
     public boolean onKeyDown(int keyCode, KeyEvent event){
         if (keyCode == KeyEvent.KEYCODE_BACK && event.getRepeatCount()== 0){
-            if(ACTION_CHECK.equals(getIntent().getAction())){
+            if(ACTION_CHECK.equals(getIntent().getAction()) && Build.VERSION.SDK_INT>=21){
                 moveTaskToBack(true);
-                finish();
+                finishAndRemoveTask();
             }else if (ACTION_REQUEST_WITH_RESULT.equals(getIntent().getAction()) ||
                 ACTION_CHECK_WITH_RESULT.equals(getIntent().getAction())) {
                 finish();
--- a/src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java
@@ -365,10 +365,13 @@ public class PassCodeActivity extends AppCompatActivity implements Injectable {
     @Override
     public boolean onKeyDown(int keyCode, KeyEvent event){
         if (keyCode == KeyEvent.KEYCODE_BACK && event.getRepeatCount()== 0){
-            if (ACTION_REQUEST_WITH_RESULT.equals(getIntent().getAction()) ||
-                    ACTION_CHECK_WITH_RESULT.equals(getIntent().getAction())) {
+            if(ACTION_CHECK.equals(getIntent().getAction())){
+                moveTaskToBack(true);
+                finishAndRemoveTask();
+            }else if (ACTION_REQUEST_WITH_RESULT.equals(getIntent().getAction()) ||
+                ACTION_CHECK_WITH_RESULT.equals(getIntent().getAction())) {
                 finish();
-            }   // else, do nothing, but report that the key was consumed to stay alive
+            }// else, do nothing, but report that the key was consumed to stay alive
             return true;
         }
         return super.onKeyDown(keyCode, event);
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java
```diff
@@ -365,10 +365,13 @@ protected void clearBoxes(){
     @Override
     public boolean onKeyDown(int keyCode, KeyEvent event){
         if (keyCode == KeyEvent.KEYCODE_BACK && event.getRepeatCount()== 0){
-            if (ACTION_REQUEST_WITH_RESULT.equals(getIntent().getAction()) ||
-                    ACTION_CHECK_WITH_RESULT.equals(getIntent().getAction())) {
+            if(ACTION_CHECK.equals(getIntent().getAction())){
+                moveTaskToBack(true);
+                finishAndRemoveTask();
```
tobiasKaminsky, line 370: This is only available in API >= 21, but not on 19, which we still support.
If there is no fitting function that also works on 19, I suggest to have a version check, so that on <21 it simply does not work, but works with >=21.

Kuuuna98:
I didn't think about it! 
@tobiasKaminsky , thank you for testing and letting me know. :)
It has been modified from finishAndRemoveTask() to finish() in consideration of API  < 21.

src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java
AndyScherzinger: @Kuuuna98 could you rather put in a version check and support 21+ and <21 because according to our Android version support guideline we will drop 4.4 with the next feature release. So since you are targeting master this way would could then easily find this part in the code.

cc @tobiasKaminsky, I already out this in the 3.15. Overview issue.

src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java
```diff
@@ -367,7 +367,7 @@ public boolean onKeyDown(int keyCode, KeyEvent event){
         if (keyCode == KeyEvent.KEYCODE_BACK && event.getRepeatCount()== 0){
             if(ACTION_CHECK.equals(getIntent().getAction())){
                 moveTaskToBack(true);
-                finishAndRemoveTask();
+                finish();
```
tobiasKaminsky, line 370: Have you tried this?
I think this will just close the PassCodeActivity, but not the app and thus lets the user allow to go into the app without entering correct pin code.

tobiasKaminsky (changes requested):

src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java
Kuuuna98: I thought so, but it wasn't the result.
As a result, the app closed well.
Can you see the main screen as a result from other versions of Android?
If so, I will modify it to check the version in the first code.

Below is a test video.
https://drive.google.com/file/d/1e7D9w7HeiOg74xRNwr2AzndOLJAfEVR8/view?usp=sharing

Android version: 5
Device model: Samsung Galaxy S9

AndyScherzinger: @Kuuuna98 you can/should also try the "wack-up" scenario:

* open the app
* put in PIN
* navigate a bit through the app 
* switch to another app
* bring the app back to the front (not having killed the app in between!)

This way you do have a filled "back-stack" within the app and not just a single activity, so `finish()` should just kill the last one and with this change you should have access to the app (which is what @tobiasKaminsky suggested will happen with this change).

Other than that, again: It is fine to implement this for API >= 21 only which will be the minimum Android version anyways for the next release.

Kuuuna98:
Thank you for your opinion, @tobiasKaminsky @AndyScherzinger 
I modified the code to check the API version in the first code.

src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java
```diff
@@ -365,10 +366,13 @@ protected void clearBoxes(){
     @Override
     public boolean onKeyDown(int keyCode, KeyEvent event){
         if (keyCode == KeyEvent.KEYCODE_BACK && event.getRepeatCount()== 0){
-            if (ACTION_REQUEST_WITH_RESULT.equals(getIntent().getAction()) ||
-                    ACTION_CHECK_WITH_RESULT.equals(getIntent().getAction())) {
+            if(ACTION_CHECK.equals(getIntent().getAction()) && Build.VERSION.SDK_INT>=21){
```
tobiasKaminsky, line 369: Please change this to proper name (Lollipop).

tobiasKaminsky:
One last thing, other than that I tested it and is is :+1:

tobiasKaminsky (approved):

AndyScherzinger (approved):

PR #7513, opened 2020-11-26
Drop Android 4.4 Support

AndyScherzinger:

### Testing 


AndyScherzinger:
@axellebot see  #7178 and https://github.com/nextcloud/android/issues/7178#issuecomment-731376006
TL;DR: we drop the oldest supported Android version if it is EOL by Google _and_ <0,5% adoption rate (and 4.4. is EOL _and_ <0,5 thus we drop it). The technical reason is to simplify code since we can drop backward compatible extra implementation that eat dev effort but don't have a large enough audience.

AndyScherzinger:
You basically answered it yourself to some extend ;)

Maintainability is an aspect since 
* you have to maintained a (slightly) larger code base
* you need to test your app against all these Android versions (your code path/execution is different due to all these if/else part so root cause analysis, not that optimal)
* code complexity grows with each supported Android version as soon as you want to use OS features/APIs of newer versions

If you look at all the changes simply dropping 4.4. isn't just some lines of code (read styles and testing) since we already have to jump hoops to support 4.4 e.g. in the screenshot test run we need to bump the level to 21 so the screenshot test lib works we are even blocked with some libraries beyond Android 5 but see this as a trade-off since upgrading a lib isn't worth it if that means dropping support for too many users (security issues would be an exception that needs to be discussed if we could work around such an issue in our own code).

build.gradle
```diff
@@ -98,9 +98,6 @@ for (TaskExecutionRequest tr : getGradle().getStartParameter().getTaskRequests()
             apply from: 'gplay.gradle'
             System.console().println("Applying gplay.gradle")
         }
-        if (arg.contains("lint")) {
-            testMinSdk = 19
```
tobiasKaminsky, line 102: Please also remove all other testMinSdk occurrences

AndyScherzinger: DONE: https://github.com/nextcloud/android/pull/7513/commits/22e730651a153ffe0020188a59942064d445f669

AndyScherzinger:
@tobiasKaminsky ready :)

tobiasKaminsky:
Merci 💙

PR #7538, opened 2020-11-28
Make PIN activity mimic a dialog

bguneys:
Resolves #7455 

- [x] Tests written, or not needed

Screenshot_20201128-185031_Nextcloud

PR #7566, opened 2020-11-30
Pin dialog

AndyScherzinger:
Demo PR for #7538 to run the screenshot tests.

@tobiasKaminsky see PR above, screenshot test won't work for PR from forked repositories so we should skip these checks in case of forked repos.


tobiasKaminsky:
> @tobiasKaminsky see PR above, screenshot test won't work for PR from forked repositories so we should skip these checks in case of forked repos.

As soon as you opened this PR, also in #7538 the screenshot tests are running, or?

AndyScherzinger:
@tobiasKaminsky as discussed yes/no, they run but fail due to missing credentials: https://github.com/nextcloud/android/runs/1472935820?check_suite_focus=true

Issue #7633, opened 2020-12-08
Invisible "Cancel" button


Valdnet:
### Steps to reproduce

1. Launch the app.
2. Go to app settings.
3. Select data locations from general settings.

### Expected behaviour
- Visible `Cancel` button.

### Actual behaviour
- Invisible `Cancel` button.

![obraz](https://user-images.githubusercontent.com/47037905/101515882-220f7400-397f-11eb-9dda-ff49dc5e63ff.png)

![obraz](https://user-images.githubusercontent.com/47037905/101515967-394e6180-397f-11eb-8325-876177a5200e.png)

### Environment data
Android version: 10

Nextcloud app version: 3.14.1 and Dev 20201208

Nextcloud server version: 20.0.2

AndyScherzinger:
true for all other addition selection dialogs within the settings screen (they make use of the fallback theme)

PR #7664, opened 2020-12-14
Fix issue #7633

czlucius:
Fix for issue: #7633 
Add ThemedListPreference, extending ListPreference, to theme dialog created by ListPreference.
Add style for dialog, setting default button text color in fallback dialog theme to ?android:attr/textColorPrimary

### Testing 

How to do a test? I am new to contributing to open source, and I noticed that this issue is labelled with "good-first-issue". Apologies if this PR is not well written/there are problems with this commit.

### DIFF
--- a/src/main/res/values/styles.xml
+++ b/src/main/res/values/styles.xml
@@ -86,7 +86,7 @@
 
     <style name="FallbackTheming.Dialog.ButtonStyle"
         parent="Widget.MaterialComponents.Button.TextButton.Dialog">
-        <item name="android:textColor">?android:attr/textColorPrimary</item>
+        <item name="android:textColor">@color/text_color</item>
     </style>
 
 	<!-- separate action bar style for activities without an action bar -->
--- a/src/main/res/values/styles.xml
+++ b/src/main/res/values/styles.xml
@@ -86,7 +86,7 @@
 
     <style name="FallbackTheming.Dialog.ButtonStyle"
         parent="Widget.MaterialComponents.Button.TextButton.Dialog">
-        <item name="android:textColor">#000000</item>
+        <item name="android:textColor">?android:attr/textColorPrimary</item>
     </style>
 
 	<!-- separate action bar style for activities without an action bar -->
--- a/src/main/res/values/styles.xml
+++ b/src/main/res/values/styles.xml
@@ -80,7 +80,14 @@
         <item name="windowNoTitle">false</item>
         <item name="android:windowBackground">@color/bg_default</item>
         <item name="android:textAllCaps">false</item>
+
+        <item name="android:buttonBarButtonStyle">@style/FallbackTheming.Dialog.ButtonStyle</item>
 	</style>
+
+    <style name="FallbackTheming.Dialog.ButtonStyle"
+        parent="Widget.MaterialComponents.Button.TextButton.Dialog">
+        <item name="android:textColor">#000000</item>
+    </style>
 
 	<!-- separate action bar style for activities without an action bar -->
     <style name="BaseTheme.ownCloud.Toolbar" parent="Theme.MaterialComponents.DayNight.NoActionBar">


AndyScherzinger:
@czlucius could you post a screenshot please? ❤️ Looking at the code shouldn't it be enough to have a custom `alertDialogTheme` added to the fallback theme that takes care of the styling? (just by looking at https://stackoverflow.com/questions/51049516/how-to-change-the-style-of-a-listpreference-popup-dialog)

czlucius:
> @czlucius could you post a screenshot please?  Looking at the code shouldn't it be enough to have a custom `alertDialogTheme` added to the fallback theme that takes care of the styling? (just by looking at https://stackoverflow.com/questions/51049516/how-to-change-the-style-of-a-listpreference-popup-dialog)

Yes. I did not look into this option. I shall modify the code.
Thanks

czlucius:
By the way, `owncloud.AlertDialog` is already the `alertDialogTheme`.

AndyScherzinger:
That one is used to theme the dialogs other than the ones launched via the settings screen (where we have full control on programmatically theming the dialog on-the-fly based on server-side colors. Since that doesn't work nicely for settings when introduced so calls "FallbackThemes" so you might create a grey-ish fallback dialog theme.

AndyScherzinger:
@czlucius any reason for closing the PR? 😢

czlucius:
Sorry, I had resetted my previous commit, so I had to force push my local repo to GitHub.

czlucius:
I think GitHub automatically closed it when I had reset my previous commit

czlucius:
> That one is used to theme the dialogs other than the ones launched via the settings screen (where we have full control on programmatically theming the dialog on-the-fly based on server-side colors. Since that doesn't work nicely for settings when introduced so calls "FallbackThemes" so you might create a grey-ish fallback dialog theme.

Ok. But why does the button colors not change when I modify `FallbackTheming.Dialog` and add a button style? I had to modify `owncloud.AlertDialog` for it to work.

AndyScherzinger:
> Ok. But why does the button colors not change when I modify FallbackTheming.Dialog and add a button style? I had to modify owncloud.AlertDialog for it to work.

I think you need to add the dialog theme to the FallBackTheme

czlucius:
I updated the fallback theme(`FallbackThemingThemeBase`) to include `FallbackTheming.Dialog` as the alert dialog theme, and the cancel button still stays invisible.

Right now, I have found 2 solutions:
- Use a custom ListPreference to launch alert dialog with custom theme
- Update `owncloud.AlertDialog`'s `alertDialogTheme` to `FallbackTheming.Dialog`.

AndyScherzinger:
Wouldn't the later change all other dialogs within the app too? Without having had a look and giving it a try (I am away from my laptop) I'd say the first solution would be save(er) 👍

AndyScherzinger:
Forgot to mention: thanks for all your effort and patience discussing the solution. This is highly appreciated ❤️

czlucius:
> Wouldn't the later change all other dialogs within the app too? Without having had a look and giving it a try (I am away from my laptop) I'd say the first solution would be save(er)

You may want to see the `styles.xml` here: https://pastebin.com/QzS8ipY4

czlucius:
> Forgot to mention: thanks for all your effort and patience discussing the solution. This is highly appreciated ❤️

Thanks

czlucius:
image
I noticed `SettingsActivity` uses `Theme.ownCloud` as the theme instead of the fallback theme. `Theme.ownCloud` has an alert dialog attribute of `ownCloud.Dialog`.

czlucius:
Perhaps that is why `SettingsActivity`'s dialogs follow `ownCloud.Dialog` instead of `FallbackTheming.Dialog`?

AndyScherzinger:
Yeah, that might be the root cause, could you check if it "works" if you switch it to the fallback theme?

czlucius:
> I updated the fallback theme(`FallbackThemingThemeBase`) to include `FallbackTheming.Dialog` as the alert dialog theme, and the cancel button still stays invisible.
> 
> Right now, I have found 2 solutions:
> 
> * Use a custom ListPreference to launch alert dialog with custom theme
> * Update `owncloud.AlertDialog`'s `alertDialogTheme` to `FallbackTheming.Dialog`.

Which solution would be the best?

czlucius:
> Yeah, that might be the root cause, could you check if it "works" if you switch it to the fallback theme?

I have tried switching the theme, but the button stays invisible.

czlucius:
I think the custom ListPreference is still the best option.

AndyScherzinger:
@czlucius then let's go with your solution 👍

czlucius:
Should the list preference accept a custom style, or should the style be hard-coded into the file?

czlucius:
> > Ok. But why does the button colors not change when I modify FallbackTheming.Dialog and add a button style? I had to modify owncloud.AlertDialog for it to work.
> 
> I think you need to add the dialog theme to the FallBackTheme

Sorry. I made a mistake in the styles.xml and used `buttonBarButtonStyle` instead of `android:buttonBarButtonStyle`. I have tested it and it turns out that the dialog's buttons can be styled, just by adding `android:buttonBarButtonStyle` to the `FallbackTheming.Dialog` theme. I think the custom `ListPreference` won't be needed then, as this is a simpler solution.

czlucius:
I have pushed the updates to this PR, you may take a look at the commits made.
Right now, there is a slight problem with using `?android:attr/textColorPrimary`, because if the theme in the app settings was set to always dark, the text color would still follow the system(system light/dark mode) and not remain white color in Android's light mode, when the app theme is set to dark.

AndyScherzinger (approved):
@czlucius with your latst commit it works like a charm 👍 `@color/text_color` is doing it's job 🎉

czlucius:
> @czlucius with your latest commit it works like a charm 👍 `@color/text_color` is doing it's job 🎉 

Thanks

czlucius:
By the way, while testing, I keep on receiving crashes with this exception while opening the app

```

2020-12-17 02:55:46.969 31240-32376/com.nextcloud.client E/AndroidRuntime: FATAL EXCEPTION: Thread-14
    Process: com.nextcloud.client, PID: 31240
    java.lang.NoClassDefFoundError: Failed resolution of: Lokio/Buffer;
        at okhttp3.ResponseBody.create(ResponseBody.java:210)
        at okhttp3.internal.Util.<clinit>(Util.java:62)
        at okhttp3.Credentials.basic(Credentials.java:30)
        at com.owncloud.android.lib.common.OwnCloudClientFactory.createNextcloudClient(OwnCloudClientFactory.java:217)
        at com.owncloud.android.operations.RefreshFolderOperation.updatePredefinedStatus(RefreshFolderOperation.java:330)
        at com.owncloud.android.operations.RefreshFolderOperation.updateCapabilities(RefreshFolderOperation.java:304)
        at com.owncloud.android.operations.RefreshFolderOperation.updateOCVersion(RefreshFolderOperation.java:276)
        at com.owncloud.android.operations.RefreshFolderOperation.run(RefreshFolderOperation.java:229)
        at com.owncloud.android.lib.common.operations.RemoteOperation.run(RemoteOperation.java:360)
        at java.lang.Thread.run(Thread.java:919)
     Caused by: java.lang.ClassNotFoundException: Didn't find class "okio.Buffer" on path: DexPathList[[zip file "/data/app/com.nextcloud.client-f8CDsZTvR0Kck4sHwWuMLw==/base.apk"],nativeLibraryDirectories=[/data/app/com.nextcloud.client-f8CDsZTvR0Kck4sHwWuMLw==/lib/arm64, /data/app/com.nextcloud.client-f8CDsZTvR0Kck4sHwWuMLw==/base.apk!/lib/arm64-v8a, /system/lib64, /product/lib64]]
        at dalvik.system.BaseDexClassLoader.findClass(BaseDexClassLoader.java:196)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:379)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:312)
        at okhttp3.ResponseBody.create(ResponseBody.java:210) 
        at okhttp3.internal.Util.<clinit>(Util.java:62) 
        at okhttp3.Credentials.basic(Credentials.java:30) 
        at com.owncloud.android.lib.common.OwnCloudClientFactory.createNextcloudClient(OwnCloudClientFactory.java:217) 
        at com.owncloud.android.operations.RefreshFolderOperation.updatePredefinedStatus(RefreshFolderOperation.java:330) 
        at com.owncloud.android.operations.RefreshFolderOperation.updateCapabilities(RefreshFolderOperation.java:304) 
        at com.owncloud.android.operations.RefreshFolderOperation.updateOCVersion(RefreshFolderOperation.java:276) 
        at com.owncloud.android.operations.RefreshFolderOperation.run(RefreshFolderOperation.java:229) 
        at com.owncloud.android.lib.common.operations.RemoteOperation.run(RemoteOperation.java:360) 
        at java.lang.Thread.run(Thread.java:919) 
```

I had to turn off the Wi-Fi so that this error will not happen


Steps to reproduce:

- Turn on Wi-Fi or mobile data, connected to the Internet 
- Open the app, signed in to an account
- Pull to refresh/Tap the switch accounts button


Device info:
Samsung Galaxy Tab S6 Lite(SM-P610)
Android 10(API 29)

PR #7767, opened 2021-01-01
Basic kotlin migrations

Doomsdayrs:
I avoided changing major aspects of the code, Focusing on what was simple enough to change without causing cascading issues.

I also noticed some code conventions that I threw TODOs and Deprecated at because they should be on their way out.

I also added some KTX extensions so that writing for kotlin becomes easier for the next people to work on this


Doomsdayrs:
Fun part is that I couldn't even see half these errors because android studio kept running out of memory

Doomsdayrs:
Just confirmed that the build failures for GPlay are not of what I changed, but something that existed on master before hand

AndyScherzinger:
@Doomsdayrs not sure what the issue is with gplay you mentioned but your latest commit works: https://github.com/nextcloud/android/actions/runs/456650526 😁

Doomsdayrs:
btw, Should I append the files I changed with my own (c)?

Doomsdayrs:
@AndyScherzinger Strange, For me it was throwing issues with the Firebase classes

AndyScherzinger:
> btw, Should I append the files I changed with my own (c)?

Feel free to do so, yes. If it is too much work feel free to skip the ones where you just converted the file automatically without any manual changes afterwards. On all others you should while we don't enforce it.

src/main/java/com/nextcloud/client/di/AppModule.kt
```diff
@@ -0,0 +1,181 @@
...
+
+    @Provides
+    fun arbitraryDataProvider(context: Context): ArbitraryDataProvider = ArbitraryDataProvider(context.contentResolver)
+
+    @Provides
+    fun activitiesServiceApi(accountManager: UserAccountManager?): ActivitiesServiceApi =
+        ActivitiesServiceApiImpl(accountManager)
+
+    @Provides
+    fun activitiesRepository(api: ActivitiesServiceApi?): ActivitiesRepository = RemoteActivitiesRepository(api!!)
```
ezaquarii, line 100: I don't think having a nullable types in DI is ok.

src/main/java/com/owncloud/android/authentication/LoginUrlInfo.kt
```diff
@@ -0,0 +1,33 @@
...
+ * License along with this program.  If not, see <http:></http:>//www.gnu.org/licenses/>.
+ */
+package com.owncloud.android.authentication
+
+/**
+ * Data object holding the login url fields.
+ */
+data class LoginUrlInfo(
+    @JvmField
+    var serverAddress: String? = null,
```
ezaquarii, line 28: Why nullable?

ezaquarii (changes requested):
It looks like an automatic code transpilation as a lot of types are nullable and this is not OK.

src/main/java/com/owncloud/android/authentication/LoginUrlInfo.kt
Doomsdayrs: Java fields are all nullable by default, and the way the data classes are used is via inputting the data POST constructor. So I simply left as is.

src/main/java/com/nextcloud/client/di/AppModule.kt
Doomsdayrs: I agree, Did cause some issues for me. I'll remove them

src/main/java/com/owncloud/android/authentication/LoginUrlInfo.kt
Doomsdayrs: I can modify how the code uses the data object, if you would like to remove the defaults

Doomsdayrs:
> It looks like an automatic code transpilation as a lot of types are nullable and this is not OK. All nullable types should be changed. We can permit some exceptions if they are justified, but I can clearly see that they are not. Especially in DI wiring.

The major issue is, as i mentioned in a comment that Java is a null unsafe language, so everything not labelled `@NotNull` is nullable, meaning the Kotlin transpilation will see it as nullable. This is not a Kotlin issue, it is a Java & lack of annotation / documentation issue


I can easily fix it up to be of standard, I'm simply removing boilerplate code that might be redundant with Kotlin syntax.

ezaquarii:
> The major issue is, as i mentioned in a comment that Java is a null unsafe language, so everything not labelled `@NotNull` is nullable

The language does not support this, but the best-practice dictates that everything that is not `Optional<T>` or explicitly annotated with `@Nullable` is treated as not nullable. I.e. `@NotNull` annotation can be regarded as a "cultural" default.

Since the above mentioned is only an advised practice and not followed consistently, transpiler cannot make assumptions and must resolve all ambiguous types as `T?`. This produces low quality, but compilable code.

> I'm simply removing boilerplate code that might be redundant with Kotlin syntax. 

As a result, this PR is removing git blame history (used for bug analysis), creating confusing interfaces (something that takes `T?` just to blow up with `!!` cast) and possibly conflicts with other branches where features are being worked on.

I'd argue that the relatively trivial boilerplate is less problematic than that and - as a result - I don't see a value added proposition here.

ezaquarii:
> Don't think I'm doing a nuke from orbit, I'm reviewing each changed file and bringing it up to spec.

It's 1,000 lines of code for review and testing. We had much larger

Doomsdayrs:
> > The major issue is, as i mentioned in a comment that Java is a null unsafe language, so everything not labelled `@NotNull` is nullable
> 
> The language does not support this, but the best-practice dictates that everything that is not `Optional<T>` or explicitly annotated with `@Nullable` is treated as not nullable. I.e. `@NotNull` annotation can be regarded as a "cultural" default.
> 
> Since the above mentioned is only an advised practice and not followed consistently, transpiler cannot make assumptions and must resolve all ambiguous types as `T?`. This produces low quality, but compilable code.
> 
> > I'm simply removing boilerplate code that might be redundant with Kotlin syntax.
> 
> As a result, this PR is removing git blame history (used for bug analysis), creating confusing interfaces (something that takes `T?` just to blow up with `!!` cast) and possibly conflicts with other branches where features are being worked on.
> 
> I'd argue that the relatively trivial boilerplate is less problematic than that and - as a result - I don't see a value added proposition here.

I can see the issue with git blame, mostly due to the transpiler having some weird behavior with deleting files, Easy enough to fix, One moment.

Doomsdayrs:
@ezaquarii You made some good points, Instead of doing a bulk request Like this I'll make some smaller PRs over time

PR #7894, opened 2021-01-24
Auto upload can compare local and remote files.

tomaszduda23:
An user is asked what to do only if those are different.

Signed-off-by: Tomasz <tomaszduda23@gmail.com>

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
+++ b/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
@@ -53,6 +53,7 @@ import com.owncloud.android.lib.common.operations.RemoteOperationResult.ResultCo
 import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.lib.resources.e2ee.UnlockFileRemoteOperation;
 import com.owncloud.android.lib.resources.files.ChunkedFileUploadRemoteOperation;
+import com.owncloud.android.lib.resources.files.DownloadFileRemoteOperation;
 import com.owncloud.android.lib.resources.files.ExistenceCheckRemoteOperation;
 import com.owncloud.android.lib.resources.files.ReadFileRemoteOperation;
 import com.owncloud.android.lib.resources.files.UploadFileRemoteOperation;
@@ -66,6 +67,7 @@ import com.owncloud.android.utils.UriUtils;
 
 import org.apache.commons.httpclient.HttpStatus;
 import org.apache.commons.httpclient.methods.RequestEntity;
+import org.apache.commons.io.FileUtils;
 import org.lukhnos.nnio.file.Files;
 import org.lukhnos.nnio.file.Paths;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -1119,6 +1113,57 @@ private String getNewAvailableRemotePath(OwnCloudClient client, String remotePat
...
+        } catch (IOException e) {
+            Log_OC.d(TAG, temporalPath + " unable to remove");
+            return false;
+        }
+
+        if (mCancellationRequested.get()) {
+            throw new OperationCancelledException();
+        }
+
+        mDownloadOperation = new DownloadFileRemoteOperation(mFile.getRemotePath(),
```
tobiasKaminsky, line 1143: This means that you download the entire file, or?
What is the purpose of it?

tomaszduda23: The file is downloaded only if a file with the same name and the same size exists on a remote server. In that way the app can decided if the file was uploaded already or it is new. If the file is new notification about conflict is displayed.

It would be better to use some kind of hash for that but it seems that it is impossible with current nextcloud API.

Let me know if you interested in the idea. Than I will fix a few things in this patch.

tobiasKaminsky: Downloading the file is unfortunately something we should not do.
Imagine you have 50 conflicts, or conflicts with very big files (>2Gb).

Let us ask at server repo to get such an endpoint/info directly:
https://github.com/nextcloud/server/issues/25949

In general I really like the idea, but we always need to be aware of such edge cases.
I would vote for closing this PR for now, and see if / what is discussed on server issue and then work further on this afterwards?

tomaszduda23: Now during conflict default behavior is "Ask me every time". I added another option "Ask me if files are different". Downloading file just to compare the content is not the best idea. Still it seems to be better than asking an user regarding each file (I have over 7000 photos) after resetting (db) the app.

I though about adding hash to server side but there was already a few attempts and all of them were rejected. E.g. https://github.com/nextcloud/server/pull/22495

For now downloading and comparing files during conflict seems to be the best option. It could be easily adjusted later on when/if server supports hash API.

BTW even if there are many big files most likely it won't use more data transfer than current options. Rename new version, Overwrite remote version - those will re-upload so transfer will be wasted anyway, Skip uploading - only this save the data transfer. Also for now there is no way for user to now if local and remote files are the same or not.

tobiasKaminsky:
Downloading any file is nothing that is scaling or should be done, for data reason and also UX reason, as the user has to wait for the download.

I wonder if filesize and date might be enough.

@nextcloud/desktop how do you compare such files, to know if they are really the same?

er-vin:
> @nextcloud/desktop how do you compare such files, to know if they are really the same?

Currently it's basically done using a mix of size, mtime and etag.

PR #7999, opened 2021-02-16
[WIP] Internal provider

LukasReschke:


LukasReschke:
I don't think this one is still valuable after so many years! Sorry! 🙈

PR #8005, opened 2021-02-17
Template Dialog: template needs to be selected

### LINKED ISSUE #5144: Creating files from templates is missing an action button
When creating files from templates e.g. with Collabora the view seems to be confusing to quite some people as there is no primary action button present.

image

I would propose the following enhancements to that:
- Add a "Create file" primary action in the bottom
- If only one template is present (empty) hide the template selector

cc @nextcloud/designers

### LINKED ISSUE #7991: Improve Usability of Create-New-Document Dialog
I'm always frustrated when [...]
... I'm creating new text documents, enter the title and then hit the button in the bottom right of the dialog to confirm.
Because that's where my thumb is trained to find the confirm/ok/yes button!

Furthermore, figuring out that you need to click on the template to actually create the file (and finish the dialog) is somewhat unintuitive.

1. Move the *Cancel* button (negative action) to the left, introduce a *Create* button (positive action) on the right.
2. Introduce a "selected" state for the possible templates. At most one can be selected at the time.

I'm not sure on the 2nd point, whether this is a good suggestion. I haven't seen any cases where there actually ever is a choice between 2 or more templates. Thus I'm not 100% sure what the current functionality/flow here is.

Keep it like this? Dunno.

### Additional context

cc @nextcloud/designers

tobiasKaminsky:
Fix #7991
Fix #5144
Closes #6317

~~@AndyScherzinger this is not beautiful at all, do you have time to enhance this? 💙~~

 

 

### Testing 


stefan-niedermann:
Do the error snackbars disappear immediately when entering a name or clicking on a template?

AndyScherzinger:
@tobiasKaminsky DONE

device-2021-02-17-142812

Note that I added a helper text, this way you could change the dialogs behavior:
* grey-out and disable the create button until both infos have been set

jancborchardt:
Looks very nice! Just 2 questions:
- Are we getting the template previews from the server? Would look nicer than if it's all the filetype icon :)
- Is it possible to prefill the filename input with e.g. the template name, so then it would be "Meeting notes.md" with the cursor before the dot?

AndyScherzinger:
> * Are we getting the template previews from the server? Would look nicer than if it's all the filetype icon :)

Yes they are fetched from the server

> * Is it possible to prefill the filename input with e.g. the template name, so then it would be "Meeting notes.md" with the cursor before the dot?

Great idea: Implemented with the latest commit. Prefilling is done in case: 
* filename is empty
* filename equals "."+ the template's extension (so no filename provided)

AndyScherzinger (approved):

szaimen:
Hey, I've just tested the PR and it would be great if the template could be selected by default if only one is available => makes the creation of a new file much faster ;)

AndyScherzinger:
> Hey, I've just tested the PR and it would be great if the template could be selected by default if only one is available => makes the creation of a new file much faster ;)

Done with latest commit (just wait for the new QA build...) and also: Hides the template chooser if only one template exists, see also #5144

szaimen:
@AndyScherzinger thanks for directly implementing my review :)
Just not showing the template if only one template is available is a good idea but in this case, the header maybe should be changed to `Create new file`, than showing `Choose template`. Otherwise if that is too complicate or too much work, it could show the only template and select it by default. 

Also I can sometimes reproduce an issue where the create button is grey e.g. by triggering creating a new file, locking my phone, unlocking it again and looking at the popup again -> the create button is grey now. In my feeling there are more ways to reproduce this issue...

AndyScherzinger:
@szaimen  my pleasure and thanks for testing. There will be 2 more QA builds in the next minutes which should address the title handling as well as the button 👍

szaimen:
@AndyScherzinger really appreciated, thanks!
I am not sure if it is a translation issue but it says now "Neu erstellen Textdokument" bzw. "Neu erstellen Dokument" etc. in the header which sounds a bit strange to me.

Also, but this should most likely get addressed with another PR: I am not able to create a new file from template when beeing in a folder that is mounted via the external storage app:
Screenshot_20210218-003141_Nextcloud QA
To reproduce: 
1. Enable the external storage app on your server
2. Create a local external storage mount with the app and mount it to a user (and make it writeabel)
3. Open the android app, navigate into the external storage folder and try to create a file from template there -> error creating file from template

(But I need to say that this second issue is present since a while now and not new in this PR. I just thought it could make sense to show it since this PR addresses this part of the app)

AndyScherzinger:
@szaimen yes,that is a translation issue already being discussed in another issue (same text as the bottom sheet!).

Same for the mount issue. Best to raise a new issue and ping @tobiasKaminsky about it.

And again, thanks for your testing efforts, highly appreciate it ❤️

PR #8011, opened 2021-02-18
Meaningfull project name

hannesa2:
What happens, when everyone has the name android ?
image

This will change this.

image

or

image

No test is needed

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,3 +1,3 @@
-rootProject.name = 'nextCloud'
+rootProject.name = 'Nextcloud'
 
 include ':'
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,2 +1,3 @@
+rootProject.name = 'nextCloud'
+
 include ':'
-//include ':nextcloud-android-library'


settings.gradle
```diff
@@ -1,2 +1,3 @@
+rootProject.name = 'nextCloud'
```
AndyScherzinger, line 1:
```suggestion
rootProject.name = 'Nextcloud'
```

...so it fits the official naming schema 👍

AndyScherzinger:
Nice catch @hannesa2, just a minor comment about the naming schema 👍

settings.gradle
hannesa2: Now it is

AndyScherzinger (approved):

settings.gradle
```diff
@@ -1,2 +1,3 @@
+rootProject.name = 'Nextcloud'
+
 include ':'
-//include ':nextcloud-android-library'
```
tobiasKaminsky, line 2: Can you keep this line in, as this is "needed" if one wants to include library directly, as explained here:
https://github.com/nextcloud/android/blob/master/SETUP.md#5-how-to

tobiasKaminsky:
Nice! 
Thanks @hannesa2 

Can you do a signed commit? (https://github.com/nextcloud/android/pull/8011/checks?check_run_id=1926214183)

hannesa2:
> Can you do a signed commit? (https://github.com/nextcloud/android/pull/8011/checks?check_run_id=1926214183)

This makes is impossible to apply suggested changes from github web-ui

settings.gradle
hannesa2: In my point of view, this is a misleading comment, next to the `SETUP.md`, which nobody will read. 
If you really want to archive this, I suggest to make two flavors 
* `buildWithLibrary`
* `buildWithCode`

tobiasKaminsky:
> This makes is impossible to apply suggested changes from github web-ui

Totally true. This is a missing thing on Github's side.

PR #8034, opened 2021-02-23
Harmonize screenshot test with a matrix

hannesa2:
Like the title describe

PR #8045, opened 2021-02-24
Proposal to change max line length

hannesa2:
When I edit files, IDE change automatic to wrong line break, eg

image

**This cause errors** and wasn't the first time, eg https://github.com/nextcloud/android/pull/8034#discussion_r581230044

Btw, code format setting was already checked-in, but ignored, so any changes were not commited

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,6 +13,9 @@ max_line_length=120
 indent_style=space
 indent_size=4
 
+[*.yml]
+max_line_length=150
+
 charset=utf-8
 
 # Trimming is good for consistency
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,8 @@ tests/proguard-project.txt
 
 # Android Studio and Gradle specific entries
 .gradle
-.idea
+.idea/*
+!.idea/codeStyles/
 build
 /gradle.properties
 .attach_pid*


AndyScherzinger:
I'd rather keep the 120 in general and use 150 for yaml-only. @tobiasKaminsky If I remeber it correctly we decided to go with 120 since that is the max width in Github displaying files, no?

tobiasKaminsky:
> I'd rather keep the 120 in general and use 150 for yaml-only.

If this is possible, yes.

I also experienced that Android Studio sometimes hard wrap lines in yaml files.
This obviously should not be the case, like it is also not the case in java / kt files.
But I failed to find why and where…

hannesa2:
> If this is possible, yes.

I guess not. In ` .idea/codeStyles/Project.xml` there is just one line length value

But have a look https://github.com/nextcloud/android/commit/83e16cb536b6b2588742287a84fb6312d85b398c
Here the length is/was no problem at all

AndyScherzinger:
> I guess not. In .idea/codeStyles/Project.xml there is just one line length value

I haven't tested it but would hope this to be true:
2021-02-24 11_59_50-Settings

So setting it in the .editorconfig file should be enough 🤔

.editorconfig
```diff
@@ -8,7 +8,7 @@
 root=true
 
 [*]
-max_line_length=120
+max_line_length=150
```
AndyScherzinger, line 11: shouldn't 150 be put into its own section like
[*.yml]
max_line_length=150

(just a guess!)

hannesa2: I applied it. 
No clue, if it works, but we can give it a try

tobiasKaminsky: I tested it locally and it seems to work partly after a restart: line wrap is not done automatically anymore after 120 chars.
(neither it is done after 150 chars)

AndyScherzinger:
@tobiasKaminsky so merge?

AndyScherzinger (approved):

PR #8048, opened 2021-02-24
Harmonize check CI jobs

hannesa2:
To be honest, I would go with `./gradlew check` but I've the feat that `lint` runs into tons of issues.
Maybe `test` will fail to. 
That's why let's go step by step


tobiasKaminsky:
> To be honest, I would go with `./gradlew check` but I've the feat that `lint` runs into tons of issues.

My idea was to have it split up, so that it can run in parallel, whereas "gradlew check" would execute it one after the other, or?

tobiasKaminsky (approved):

hannesa2:
> My idea was to have it split up, so that it can run in parallel, whereas "gradlew check" would execute it one after the other, or?

You are right. This and previous way run it in parallel. `check` would do it serial. But I guess overall runtime will be not increased, because most time consuming Espresso check is slower anyway. Maybe I'm wrong.

PR #8064, opened 2021-02-25
Run screenshot tests only if label is present

AndyScherzinger:

### Testing 


tobiasKaminsky:
I am not that sure about this.
Having tests only executed on special situations might delay finding those bugs.

I cannot find it, but there is an abstract that discourages to do this.
It basically boils down to "if tests are that expensive that you do not want to them always, you have to rethink test strategy".

So in this case I would be fine with:
- always run screenshot tests with blue / light mode, so that we are aware of any UI changes (e.g. we change something on sharing API and suddenly some shares are read-only).
- run entire screenshot suite 
  - on a regular basis, e.g. once a day, show this summary as a persistent link on contributing.md
  - when label is set

I think this is a good trade off?

AndyScherzinger:
@tobiasKaminsky yes, that sounds good to me too! 👍

tobiasKaminsky:
> Edit: While I have no idea how to execute only parts of the matrix...

We can have a job that test if it should be executed.
And all other jobs depend on it via "needs":
> Identifies any job that must complete successfully before this
job will run. It can be a string or array of strings. If a job fails,
all jobs that need it are skipped unless the jobs use a
conditional statement that causes the job to continue.

Issue #8076, opened 2021-02-27
Misdisplaying view in landscape mode (starting from landscape)


Mehdi-git:
### Expected behavior
App should run normally and without any misdisplaying from landscape mode just like starting app from portrait mode.
Correct view: 
![1](https://user-images.githubusercontent.com/52636050/109387521-1ba5ee00-7917-11eb-9b22-e4fa56331908.png)

### Actual behavior
When app runs from landscape mode, view becomes small like picture below:
![2](https://user-images.githubusercontent.com/52636050/109387530-2496bf80-7917-11eb-8ade-31b93d4ab920.png)

and after returning back to portrait again, view appears like picture below: 
![3](https://user-images.githubusercontent.com/52636050/109387592-81927580-7917-11eb-98d1-aeb9b3e3c53c.png)

### Environment data
Android version: 10

Device model: Aphone_6s

Nextcloud app version: 3.16.0 Alpha1

Li-An:
Hello, I meet the same problem 

Environment data
Android version: 7.0

Device model: Samsung Tab S2

Nextcloud app version: 3.5.0

tobiasKaminsky:
To reproduce: 
- rotate device (even emulator) to landscape mode
- start the app

PR #8124, opened 2021-03-07
Fixed Misdisplaying view in landscape mode.

Mehdi-git:
According to issue " Misdisplaying view in landscape mode (starting from landscape) #8076",  I changed  the value of the field "android:layout_width" in files.xml, from @dimen/zero" to "match_parent".
Now it works correctly in either landscape or portrait mode.


tobiasKaminsky:
👍 nice finding.
This whole left/right is from old tablet mode, but we aim to get rid of it.

PR #8290, opened 2021-04-16
Optimize writing out files to Nextcloud via SAF

grote:
This is optimizing write operations to Nextcloud made via the Storage Access Framework. If we are not reading, but only writing to a file, we skip downloading it before writing to it.
Then after having uploaded it, we remove the local copy to save storage space on the local device (important when uploading large amounts of data). If one of the uploaded files is ever accessed, it gets downloaded on demand from Nextcloud reliably.


tobiasKaminsky (approved):

tobiasKaminsky:
@dependabot-bot rebase

PR #8385, opened 2021-05-11
Fix ConnectivityService.isInternetWalled() check when content length is -1

### LINKED ISSUE #8384: Upload on WiFi stopped working
### Steps to reproduce
0. use OpenBSD 6.9 httpd
1. connect to wifi
2. try uploading a file

### Expected behaviour
- Hm...

### Actual behaviour
- Connection error

- No, as it uses different transfer encoding

If the server reply uses `Transfer-Encoding: chunked`, `OkHttp` returns a reply with content length of -1.
This is obviously causing `ConnectivityService.isInternetWalled()` to return true:

https://github.com/nextcloud/android/blob/247d422b3748f0c2b6ea270bdf3613dba0abe6cc/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java#L77

When chunked encoding is used (as happens with FastCGI), content length is simply not available.

ezaquarii:
When server returns no Content-Length, the check was failing due to
expectation of 0. Change the condition to account for no content-length
situation and updated test.

Fixes #8384

### DIFF
--- a/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
+++ b/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
@@ -74,7 +74,8 @@ class ConnectivityServiceImpl implements ConnectivityService {
 
             int status = get.execute(client);
 
-            boolean result = !(status == HttpStatus.SC_NO_CONTENT && get.getResponseContentLength() == 0);
+            // Content-Length is not available when using chunked transfer encoding, so check for -1 as well
+            boolean result = !(status == HttpStatus.SC_NO_CONTENT && get.getResponseContentLength() <= 0);
 
             get.releaseConnection();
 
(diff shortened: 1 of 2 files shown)


ezaquarii:

Upload on certain types of HTTP server could be broken. I think this is a high priority fix that should be backported ASAP.
There is no risk of stability degradation as this is very simple ifology fix.

AndyScherzinger (approved):

ezaquarii:
This is definitely not me:
```
e: /drone/src/src/androidTest/java/com/owncloud/android/ui/fragment/FileDetailFragmentStaticServerIT.kt: (87, 32): None of the following functions can be called with the arguments supplied: 
public constructor PreviewObject() defined in com.owncloud.android.lib.resources.activities.models.PreviewObject
public constructor PreviewObject(fileId: Int, source: String!, link: String!, mimeTypeIcon: Boolean!, mimeType: String!, view: String!, filename: String!) defined in com.owncloud.android.lib.resources.activities.models.PreviewObject
> Task :compileGplayDebugAndroidTestKotlin FAILED
```

maanloper:
Any idea of a possible ETA in the Playstore for this PR? Very interested in seeing if this fixes the problem...

tobiasKaminsky:
Thanks for reporting and fixing :tada:

PR #8390, opened 2021-05-12
Gradle 6.9

hannesa2:
* It comes with Apple Silicon build machine support. 
* Additionally it's a precondition for #8386

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,5 @@
 import com.github.spotbugs.snom.SpotBugsTask
+import org.gradle.internal.jvm.Jvm
 
 // Gradle build file
 //
@@ -37,6 +38,8 @@ apply plugin: 'com.hiya.jacoco-android'
 apply plugin: 'com.github.spotbugs'
 apply plugin: 'io.gitlab.arturbosch.detekt'
 apply plugin: 'shot'
+
+println "Gradle uses Java ${Jvm.current()}"
 
 configurations {
     ktlint
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.9-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists


hannesa2:
Screenshot test sucks

AndyScherzinger:
> Screenshot test sucks

feel free to ignore ;)

looking at https://github.com/nextcloud/talk-android/pull/1230/files default would actually be 6.7.1 +  removed android legacy plugin, so maybe update to that first and see if everything still works. An issue with the actual CI would be that the docker image runs on java8 instead of java11 which is the new default shipping with Android Studio 4.2 (and leads to quite some code generation problems on the talk client...)


Other than that...
> It comes with Apple Silicon build machine support.

So you are running Android Studio on a Apple Silicon maschine (yet)? I thought that might be an issue regarding java, no (at least it is not supported as in Android Studio not officially supporting BigSur, etc.)?

AndyScherzinger:
Other than that I am (mostly) afk this week so might take a while for me or @tobiasKaminsky to have a deeper look into the PR

hannesa2:
> looking at https://github.com/nextcloud/talk-android/pull/1230/files default would actually be 6.7.1 + removed android legacy plugin, so maybe update to that first and see if everything still works. An issue with the actual CI would be that the docker image runs on java8 instead of java11 which is the new default shipping with Android Studio 4.2 (and leads to quite some code generation problems on the talk client...)

At least all my other repos runs well with it, on Java 8 and 11 as well. But who knows ...

> So you are running Android Studio on a Apple Silicon maschine (yet)? I thought that might be an issue regarding java, no (at least it is not supported as in Android Studio not officially supporting BigSur, etc.)?

Almost everything working !

image

It must be api30
image

Then you have a fast Emulator, much faster build (even with x86 Java, I use arm64 Java azul-11.0.11), with better audio-in/out support. But Chrome is not working

hannesa2:
> looking at https://github.com/nextcloud/talk-android/pull/1230/files default would actually be 6.7.1 + removed android legacy plugin, so maybe update to that first and see if everything still works. An issue with the actual CI would be that the docker image runs on java8 instead of java11 which is the new default shipping with Android Studio 4.2 (and leads to quite some code generation problems on the talk client...)

Ok, I do a intermediate step #8396
If you have fear about Java 8 and 11, we could build both in CI, but this needs more runners and time. I would avoid this

AndyScherzinger:
Well, as for Java I'd go for 11 since Android Studio 4.2  ships 11 so anyone checking out the repo and opening the code base would have to change the settings for a locally installed Java 8. So 11 is just fine. I only know there are code generator issues with Java 11 on the talk client (parceler + requery).

hannesa2:
I know it's a pain 
image

Does someone investigated the logcat ? 
But there is no attached 
image

AndyScherzinger:
> Does someone investigated the logcat ?
But there is no attached

@tobiasKaminsky might knowe how to retrieve this info. Unfortunately I don't know how. Nevertheless I'd ignore this for this PR (since it is not introduced by this one) while I can't trigger another CI run (which failed due to a rate limit with the docker hub 😢). Else tested this fresh checkout with Android 4.2.1 on Win10 and it works fine for me. So whenever somebody with Drone CI permissions triggers the CI again I'd be fine with merging :) - @nickvergessen could you trigger this one again when we are fine with the docker hub threshold again?

hannesa2:
Normally I did it that way https://github.com/hannesa2/LiveEdgeDetection/blob/master/.github/workflows/Android-CI-Espresso.yml#L36-L53

But changing the Github action (which provide logact out of the box) is here not that easy because of the screenshot compare mechanism

AndyScherzinger (approved):

PR #8396, opened 2021-05-13
Android Studio 4.2

hannesa2:
Please see the commit messages


hannesa2:
To handle the Java 8  and Java 11 fear https://github.com/nextcloud/android/pull/8390#issuecomment-839847684
I added commit `Build CI with Java 8 and 11` but I recommend to remove it later to avoid too much runners and time for CI


Btw, screenshot test still sucks independent if it's Gradle 6.9 or not #8390

AndyScherzinger:
Looks good 👍 will check CI result and then merge (also I cherry picked the dependabot update for AGP=4.2.1).

Will likely do a follow-up PR as suggested by you @hannesa2 that bumps all GH action builds to Java11 while the Drone CI will be another task since that one still runs on a java 8 docker image...

PR #8405, opened 2021-05-14
Allow 3rd party apps to open the trashbin for a specific account

stefan-niedermann:
# General

This PR allows any third party Android apps to launch the `TrashbinActivity`. The `TrashbinActivity` also allows to choose a target account (because it might be different to the currently selected account) via an Intent argument named `Intent.EXTRA_USER`.

# Specific

The [Notes Android app](https://github.com/stefan-niedermann/nextcloud-notes/) has a menu item called "Trashbin". This currently opens Nextcloud in a web browser to allow restoring of deleted notes. It is not planned to implement a complete trashbin for the Notes app. For a better integration [it shall launch the Android apps trashbin in the future](https://github.com/stefan-niedermann/nextcloud-notes/pull/1217) if possible and use the web browser only as a fallback.

I need some guidance for tests: Are they necessary for this change? (Might make sense to ensure that the argument names aren't changed by accident ...?) How can one test an activity? I don't have experience with testing UIs on Android.

cc: @tobiasKaminsky and @AndyScherzinger as proposed in the community chat.

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
+++ b/src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
@@ -37,6 +37,7 @@ import com.nextcloud.client.account.User;
 import com.nextcloud.client.di.Injectable;
 import com.nextcloud.client.network.ClientFactory;
 import com.nextcloud.client.preferences.AppPreferences;
+import com.nextcloud.java.util.Optional;
 import com.owncloud.android.R;
 import com.owncloud.android.databinding.TrashbinActivityBinding;
 import com.owncloud.android.lib.resources.trashbin.model.TrashbinFile;
@@ -85,17 +86,21 @@ public class TrashbinActivity extends DrawerActivity implements
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        final User user = accountProvider.getUser();
+        final User currentUser = getUser().orElse(accountProvider.getUser());
         final String targetAccount = getIntent().getStringExtra(ARG_TARGET_ACCOUNT_NAME);
-        if (!user.nameEquals(targetAccount)) {
-            if (!getUserAccountManager().setCurrentOwnCloudAccount(targetAccount)) {
+        if (targetAccount != null && !currentUser.nameEquals(targetAccount)) {
+            final Optional<User> targetUser = getUserAccountManager().getUser(targetAccount);
+            if (targetUser.isPresent()) {
+                setUser(targetUser.get());
+            } else {
                 Toast.makeText(this, R.string.associated_account_not_found, Toast.LENGTH_LONG).show();
                 finish();
                 return;
             }
         }
 
-        final RemoteTrashbinRepository trashRepository = new RemoteTrashbinRepository(user, clientFactory);
+        final RemoteTrashbinRepository trashRepository =
+            new RemoteTrashbinRepository(getUser().orElse(accountProvider.getUser()), clientFactory);
         trashbinPresenter = new TrashbinPresenter(trashRepository, this);
 
         binding = TrashbinActivityBinding.inflate(getLayoutInflater());
@@ -130,7 +135,7 @@ public class TrashbinActivity extends DrawerActivity implements
             getStorageManager(),
             preferences,
             this,
-            getUserAccountManager().getUser()
+            getUser().orElse(accountProvider.getUser())
         );
         recyclerView.setAdapter(trashbinListAdapter);
         recyclerView.setHasFixedSize(true);
--- a/src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
+++ b/src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
@@ -285,6 +285,17 @@ public class TrashbinActivity extends DrawerActivity implements
         binding.list.setVisibility(View.GONE);
     }
 
+    @VisibleForTesting
+    public void showUser() {
+        binding.loadingContent.setVisibility(View.GONE);
+        binding.list.setVisibility(View.VISIBLE);
+        binding.swipeContainingList.setRefreshing(false);
+
+        binding.emptyList.emptyListViewText.setText(getUser().get().getAccountName());
+        binding.emptyList.emptyListViewText.setVisibility(View.VISIBLE);
+        binding.emptyList.emptyListView.setVisibility(View.VISIBLE);
+    }
+
     @Override
     public void showError(int message) {
         if (active) {
--- a/src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
+++ b/src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
@@ -23,6 +23,7 @@
  */
 package com.owncloud.android.ui.trashbin;
 
+import android.content.Intent;
 import android.os.Bundle;
 import android.view.Menu;
 import android.view.MenuItem;
@@ -69,8 +70,6 @@ public class TrashbinActivity extends DrawerActivity implements
     TrashbinContract.View,
     Injectable {
 
-    private static final String ARG_TARGET_ACCOUNT_NAME = "TARGET_ACCOUNT_NAME";
-
     public static final int EMPTY_LIST_COUNT = 1;
     @Inject AppPreferences preferences;
     @Inject CurrentAccountProvider accountProvider;
@@ -87,7 +86,7 @@ public class TrashbinActivity extends DrawerActivity implements
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         final User currentUser = getUser().orElse(accountProvider.getUser());
-        final String targetAccount = getIntent().getStringExtra(ARG_TARGET_ACCOUNT_NAME);
+        final String targetAccount = getIntent().getStringExtra(Intent.EXTRA_USER);
         if (targetAccount != null && !currentUser.nameEquals(targetAccount)) {
             final Optional<User> targetUser = getUserAccountManager().getUser(targetAccount);
             if (targetUser.isPresent()) {
--- a/src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
+++ b/src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
@@ -24,6 +24,7 @@
 package com.owncloud.android.ui.trashbin;
 
 import android.os.Bundle;
+import android.util.Log;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
@@ -67,11 +68,15 @@ public class TrashbinActivity extends DrawerActivity implements
     TrashbinContract.View,
     Injectable {
 
+    private static final String TAG = TrashbinActivity.class.getSimpleName();
+    private static final String ARG_TARGET_ACCOUNT_NAME = "TARGET_ACCOUNT_NAME";
+
     public static final int EMPTY_LIST_COUNT = 1;
     @Inject AppPreferences preferences;
     @Inject CurrentAccountProvider accountProvider;
     @Inject ClientFactory clientFactory;
     private TrashbinListAdapter trashbinListAdapter;
+    private User user;
 
     @VisibleForTesting
     TrashbinPresenter trashbinPresenter;
@@ -82,7 +87,17 @@ public class TrashbinActivity extends DrawerActivity implements
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        final User user = accountProvider.getUser();
+        user = accountProvider.getUser();
+        final String targetAccount = getIntent().getStringExtra(ARG_TARGET_ACCOUNT_NAME);
+        if (!user.getAccountName().equals(targetAccount)) {
+            final boolean accountSwitchSuccessful = getUserAccountManager().setCurrentOwnCloudAccount(targetAccount);
+            if (accountSwitchSuccessful) {
+                user = getUserAccountManager().getUser(targetAccount).orElse(user);
+            } else {
+                Log.w(TAG, "Tried to switch current account to \"" + targetAccount + "\", but could not be found.");
+            }
+        }
+
         final RemoteTrashbinRepository trashRepository = new RemoteTrashbinRepository(user, clientFactory);
         trashbinPresenter = new TrashbinPresenter(trashRepository, this);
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
```diff
@@ -118,7 +133,7 @@ private void setupContent() {
             getStorageManager(),
             preferences,
             this,
-            getUserAccountManager().getUser()
+            user
```
stefan-niedermann, line 136: :information_source: This change is needed because if the account gets changed, it gets written asynchronously into the `SharedPreferences`, causing a race condition if it gets read again if it has not yet completely be written.

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
```diff
@@ -24,6 +24,7 @@
 package com.owncloud.android.ui.trashbin;
 
 import android.os.Bundle;
+import android.util.Log;
```
ezaquarii, line 27: The app uses its own logger, not android platform log API.

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
ezaquarii: 1. `SharedPreferences` writes to cache are synchronous, dump to filesystem might not be (it depends on stuff), but it should not matter - reads are from cache, which is vanilla `Map`.

2. Nobody should use `SharedPreferences` anywhere in the app; we use `AppPreferences` instead and it offers reliable read/write API.

3. Nobody should read user from shared preferences or account manager; we have our own `UserAccountManager` abstraction that provides reliable API.

I stripped the API of any weirdness so if you still find any edge cases, you found a bug, and I'll fix it.

If you genuinely expect the value to change unexpectedly, it would be better to pass a delegate instead of a value. Delegate can always deliver current value. The interface for this is `CurrentAccountProvider`.

Sadly, there is no reactive API for account switching, as all activities that do it handle it be rebooting itself.

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
```diff
@@ -82,7 +87,17 @@
...
         super.onCreate(savedInstanceState);
-        final User user = accountProvider.getUser();
+        user = accountProvider.getUser();
+        final String targetAccount = getIntent().getStringExtra(ARG_TARGET_ACCOUNT_NAME);
+        if (!user.getAccountName().equals(targetAccount)) {
+            final boolean accountSwitchSuccessful = getUserAccountManager().setCurrentOwnCloudAccount(targetAccount);
+            if (accountSwitchSuccessful) {
+                user = getUserAccountManager().getUser(targetAccount).orElse(user);
+            } else {
+                Log.w(TAG, "Tried to switch current account to \"" + targetAccount + "\", but could not be found.");
```
ezaquarii, line 97: This should be logged via `Logger` API, although you are simply ignoring the error here, so the user will land in an unexpected place.

I think this is not good. Some sort of error pop up could be better maybe? Toast and quit?

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
```diff
@@ -82,7 +87,17 @@
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        final User user = accountProvider.getUser();
+        user = accountProvider.getUser();
+        final String targetAccount = getIntent().getStringExtra(ARG_TARGET_ACCOUNT_NAME);
+        if (!user.getAccountName().equals(targetAccount)) {
+            final boolean accountSwitchSuccessful = getUserAccountManager().setCurrentOwnCloudAccount(targetAccount);
+            if (accountSwitchSuccessful) {
+                user = getUserAccountManager().getUser(targetAccount).orElse(user);
```
ezaquarii, line 95: This approach switches the currently active user for the entire application.

Current user for the activity is stored in `SessionMixin` and that API should be used to perform user switch in activities.

One of the challenges of previous (lack of) design was that everybody tried to do their own current user accounting. Although I removed some, the amount of duplication and  edge cases during switch is still concerning.

I'd suggest trying `SessionMixin` and if that doesn't work for you - we can extend it. It's not finished, I admit, but fixing the `SessionMixin` sounds IMO better as rolling yet another independent enterprise.

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
```diff
@@ -82,7 +87,17 @@
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        final User user = accountProvider.getUser();
+        user = accountProvider.getUser();
+        final String targetAccount = getIntent().getStringExtra(ARG_TARGET_ACCOUNT_NAME);
+        if (!user.getAccountName().equals(targetAccount)) {
```
ezaquarii, line 92: This won't work account name can have different case.

`User.nameEquals(name: CharSequence?): Boolean` API does the right thing and is guaranteed to be correct in the future.

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
ezaquarii: BTW, this PR also breaks user accounting, as the `SessionMixin` will hold old user (and it will be used by some existing activity API), while trash bin related UI will use a new one. I'm sure we can call it "undefined behaviour".

ezaquarii (commented):
I found some major issues with this approach.

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
stefan-niedermann: Logging replaced with

> Toast and quit

as [proposed by you](https://github.com/nextcloud/android/pull/8405#discussion_r633008025).

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
stefan-niedermann: Logging replaced with

> Toast and quit

as proposed

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
stefan-niedermann: :heavy_check_mark: Thanks for the hint, i wasn't ware of this method

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
stefan-niedermann: `UserAccountManagerImpl` currenlty [*does* work with `SharedPreferences`](https://github.com/nextcloud/android/blob/master/src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java#L284-L286) internally. Using `SharedPreferences` is nothing that i introduced, i only noticed it while digging through the code and thought it might be relevant. Also i do not read anyhting from shared preferences or account manager, I only use `UserAccountManager` as you suggested - just have a look at the diff :wink: 

However, i was not aware that an `apply()` and a following read will deterministically return the new value.

I believe you and changed this line back to `getUserAccountManager().getUser()` :shrug: .

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
stefan-niedermann: Well, i was also not aware about `SessionMixin` and i am unsure about how to use it. Here my guess:
1. Inject it as member variable `@Inject SessionMixin sessionMixin;`
2. Get the target user via `getUserAccountManager().getUser(...)` and resolve the `Optional`
3. Use `sessionMixin.setUser(...)` in favor of `getUserAccountManager().setCurrentOwnCloudAccount(...)`
    - I noticed that `SessionMixin` [falls silently back to `swapToDefaultAccount`](https://github.com/nextcloud/android/blob/master/src/main/java/com/nextcloud/client/mixins/SessionMixin.kt#L64) if the account couldn't be found. This is exactly the behavior [you criticized on my PR](https://github.com/nextcloud/android/pull/8405#discussion_r633008025), isn't it?
4. I am unsure how `SessionMixin` works in combination with `UserAccountManager` and `CurrentAccountProvider`:
    - The current user is needed on various places, before my PR it is one time requested from `CurrentAccountProvider` and another time from `UserAccountManager`. 
    - Do i have to fetch the wanted user always from `SessionMixin` in this class now?
    - Do i have to set the `SessionMixin` user in *any* case explicitly or will it use the last used account per default now?

ezaquarii: If you are confused, don't worry. This aspect of the app is challenging, which is a charitable understatement.

> Inject it as member variable @Inject SessionMixin sessionMixin;

No need to inject it - it's already part of `BaseActivity`. Since mixins are lifecycle aware, they must be properly wired to the activity callbacks and that is done in `MixinsRegistry` for you.

There is a package readme there, which should explain more.

> Get the target user via getUserAccountManager().getUser(...) and resolve the Optional

You could use account manager to validate user provided in arguments and refuse to service the request if it is invalid. That makes sense.

> Use sessionMixin.setUser(...) in favor of getUserAccountManager().setCurrentOwnCloudAccount(...)

A bit of a background first. What `SessionMixin` does now was initially scattered across multiple inheritance levels and lifecycle methods and was specific to files display activity. To put it bluntly - it was a complete mess with bugs near impossible to trace and `SessionMixin` emerged during bugfixing. Logic is now aggregated and hooked into predictable lifecycle, but it is still focused on files UX. Your case might need some extra work.

Now, from what I see in the code, `SessionMixin` is exposed by `BaseActivity.setUser(User)` API. You could take account name from `Intent` extras, get the appropriate `User` from the manager and - if the user is valid - use this to switch activity. It is likely that `SessionMixin` does not offer enough functionality to satisfy your needs, but **this is the place to add it**.

> I am unsure how SessionMixin works in combination with UserAccountManager and CurrentAccountProvider: 

TL;DR:
1. `CurrentAccountProvider` is a subset of `UserAccountManager` interface (they are the same thing)
2. `SessionMixin` is a UI layer code that uses `UserAccountManager` to do UI-related things in UI-specific way

`UserAccountManager` implements `CurrentAccountProvider` interface, so you can treat them as one. The `*Provider` interface is a temporary tool helping us to break circular dependencies caused by global `AccountManager` and global `Account` objects used all over the place in the past. We replaced those calls with this interface as a temporary measure, so we could obtain an explicit dependency graph and calm down static code analyser randomly detecting circular dependencies on every unrelated code change. It is a separate interface, because those places - while depending on full account manager - were using just this tiny subset of it and using limited `CurrentAccountProvider` made further refactoring much easier.

> Do i have to fetch the wanted user always from SessionMixin in this class now?

No, `SessionMixin` is supposed to keep state specific to the currently displayed UI. It replaced dozens of scattered, inconsistently updated `Account` copies. It is related to what you see on the screen and only that.

To search for an existing user, then use `UserAccountManager`. 

Using a frontend analogy, `UserAccountManager` is a database and `SessionMixin` is a view model using that database.

If you want to switch your `Activity` to a new user - do it via `SessionMixin.setUser(User)` API.  This API is exposed via `BaseActivity.setUser(User)`.

Account switching is a complete mess. Architecturally the app never supported it and still handle it by switching active account in the `AccountManager` while also restarting the whole thing, because there is no reactive mechanism to reconfigure components depending on current user. We have reduced the problem a bit, but we are far from a proper solution. I'm not sure if you can switch user in Trashbin without rebooting the activity either. We'll find out.

> Do i have to set the SessionMixin user in any case explicitly or will it use the last used account per default now?

1. It takes the current active user from `UserAccountManager` and sets it in manager as well (this means that the change is global - **this is a legacy issue!**).
2. It maintains a single cached copy of the current user in the UI layer, so we don't need to update it in multiple places. This is the same value you'll read from `UserAccountManager`, but you don't have to get it and do all the `Optional` and `null` check dancing that plagued legacy code.
3. Additionally, it updates some UI specific things. Files storage depends on active user, but there are likely othre things that we don't support correctly yet.
4. Handles lifecycle weirdness.

stefan-niedermann: Phew, i have to admit, that's quite complicated :smile: . I now use `setUser()` / `getUser()` from the `BaseActivity` (which uses `SessionMixin` internally as far as i can see). I *expected* the user only to change during this activity, but it looks like the current user account is still switched permanently (which doesn't feel wrong for me, but is just unexpected after your explanation).

Looking forward to another review :slightly_smiling_face:

tobiasKaminsky (approved):

tobiasKaminsky:
Ping me once this is done, then I can try to add some tests :)

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
ezaquarii: > I expected the user only to change during this activity, but it looks like the current user account is still switched permanently

When you switch via `BaseActivity`, it switches the account manager as well. This is unfortunate, but will not leave the UI weirdly inconsistent, for example after screen rotation which wipes out your local state.

We *could* make the change local, but now as I refreshed my memory on the implementation details, it feels like a rabbit hole. Not worth pursuing, IMO.

LGTM. Let's test it.

ezaquarii (approved):
ok

src/main/java/com/owncloud/android/ui/trashbin/TrashbinActivity.java
stefan-niedermann: > it switches the account manager as well

Yeap, [i observed this](https://github.com/nextcloud/android/pull/8405#discussion_r633318474), too.

Thank you for your review, always a pleasure :slightly_smiling_face: I am happy to provide another PR to change this behavior once the account management works as you plan it to work :+1:

stefan-niedermann:
ping @tobiasKaminsky we are ready from the implementation side.

If someone wants to manually test this PR in combination with the Notes app:
1. Download, unzip install & setup [this `.apk` file](https://github.com/stefan-niedermann/nextcloud-notes/suites/2767893944/artifacts/61305388) from the [`open-trashbin-in-files-app`](https://github.com/stefan-niedermann/nextcloud-notes/pull/1217) branch of the Notes app
3. Open left sidebar and tap on `Deleted notes`
4. Expected behavior: Nextcloud apps trashbin account should be opened with the current Notes account, regardless which account was selected previously

tobiasKaminsky:
Works 👍

stefan-niedermann:
Anything left that needs to be done before we can merge this @tobiasKaminsky ? Do we need to investigate the failing screenshot tests or so?

stefan-niedermann:
Awesome, thanks 🙂 I just saw this will be released with `3.17.0` - can you name the exact version code this release will get? (I am a little confused, thought it will be `3017000`, but there seem to be more digits according to the `build.gradle`...)

AndyScherzinger:
```30170090``` is the version code you are looking for :)

PR #8408, opened 2021-05-14
Fixes spanish translation

CarlosRuiz-globalqss:

Hi, this pull request contains fixes for spanish translation files.

Most of the fixes are typos (f.e. achivo -> archivo, Obten -> Obtén)

Additionally in the file values-es-rCO/strings.xml there is a change for the term "Liga" with "Enlace" - I'm from Colombia and "Liga" is a term used there for something different, maybe in other countries like Spain is used.

Regards,

Carlos Ruiz

________________________________

Hola, este pull request contiene arreglos para los archivos de traducción al español.

La mayoría de arreglos son errores de tipeo  (p.e. achivo -> archivo, Obten -> Obtén)

Adicionalmente en el archivo values-es-rCO/strings.xml hay un cambio del término "Liga" por "Enlace" - soy de Colombia y "Liga" es un término usado para algo diferente, quizás en otros países como España lo utilicen.

Saludos,

Carlos Ruiz


AndyScherzinger:
Hi @CarlosRuiz-globalqss and welcome to Nextcloud. First of all thank you for your efforts in improving the translations. Unfortunately the translations are handled via https://www.transifex.com/nextcloud/nextcloud/android/ (where you can register and edit the translations) and are being pulled every night, so merging your pull request with its changes gets overridden every night but if you change them on transifex then they will find their way into this reporsitory the night after :)

CarlosRuiz-globalqss:
Hi @AndyScherzinger,

I registered in transifex and tried to change one of the typos.

Unfortunately it seems too much work to learn the tool, change all the errors in all the languages.

So, it would be great if the spanish maintainer (if there is one) can be notified about these changes:
https://github.com/nextcloud/android/pull/8408/files

IMHO it's very bad image to have so many typos in spanish translation, it looks unprofessional and unpolished.

I'm not planning to use nextcloud in the future, my intention was just to contribute a bit as I was testing the application.

Regards,

Carlos Ruiz

AndyScherzinger:
Hi @CarlosRuiz-globalqss,

thanks for the update! @rakekniven do you know Spanish maintainers for the translation and could point them here?

rakekniven:
Pinged on person at Transifex.

AndyScherzinger:
Just spend the time to import the files on transifex (which is probably the fastest way to incorporate the changes). Other than that @rakekniven what do you think in general about the specializations. We do have a quite okay translation state for Spanish while all es_* suffer from specific translations or in other words is it worth the effort or should we just drop them like we ignore informal German and only use formal German?

Also lopping @tobiasKaminsky just to be sure we are all aligned when deciding on language support (and yes I am aware there there a translation specific possible in all the Spanish translation variants referenced here but I am unsure it is worth the effort if the maintenance isn't given.

CarlosRuiz-globalqss:
Hi @AndyScherzinger - an opinion as an outsider of the project, but a native spanish speaker.

The differences between the different spanish speaking countries are very subtle, unless you have strong maintainers in every country I would advice to keep just a single translation.  All countries will understand these minor differences without problem.

I see in the code 17 versions of spanish:
```
$ wc -l values-es*/strings.xml | sort -k 1rn
  9332 total
   855 values-es/strings.xml
   837 values-es-rAR/strings.xml
   615 values-es-rMX/strings.xml
   564 values-es-rDO/strings.xml
   563 values-es-rEC/strings.xml
   551 values-es-rCO/strings.xml
   547 values-es-rCL/strings.xml
   547 values-es-rCR/strings.xml
   547 values-es-rGT/strings.xml
   547 values-es-rSV/strings.xml
   452 values-es-rHN/strings.xml
   452 values-es-rPY/strings.xml
   451 values-es-rNI/strings.xml
   451 values-es-rPA/strings.xml
   451 values-es-rPE/strings.xml
   451 values-es-rPR/strings.xml
   451 values-es-rUY/strings.xml
```
Some of those are incomplete.  Seems like the most complete is values-es/

The differences I talk are about usage of words, for example, in Spain they call "Fichero" what we call "Archivo" (File), but all countries can understand the meaning independently of the word used.
"Liga" vs "Enlace"  (Link)

Regards,

Carlos Ruiz

AndyScherzinger:
Thanks for the insight @CarlosRuiz-globalqss - Android's works with a simple fallback mechanism: If a string isn't present in 'es_*' then it'll use the one found in 'es'. So it works with the incomplete translations while like you experienced yourself: isn't obvious but rather confusing for translators and doesn't seem to provide a valuable benefit if there are no translators up for the task 👍

CarlosRuiz-globalqss:
Thanks, nice to learn that  :-)

If that's the case then it would be easier for maintenance that specific country translations just contain the variations, instead of the whole file.  I see the same strings with the same typos replicated in all files.

AndyScherzinger:
That's probably because a single person added them to all the different variants I guess.

rakekniven:
> If that's the case then it would be easier for maintenance that specific country translations just contain the variations, instead of the whole file.

Unfortunately this is not possible with Transifex.

> That's probably because a single person added them to all the different variants I guess.

No, that's the same for all languages. Each string is available in each language. 
There is no way to exclude single strings for a language. 

The strings with the many spanish languages has a simple background. I big customer paid to translate the whole language into all spanish languages. That project has been finished in 2018 IIRC. Now volunteers are missing to keep it maintained and completed.

castillo92:
I always thought that there were too many Spanish languages to translate but as it was not in my hand to maintain all of them because they are too many for one person, I focused on the Spanish of Spain, and that is why I have believed that the best thing to do was to keep the es_ES variation up to date.

It is a complex problem... If we simplify it to only two variations: es_ES and es_LAT I am sure that many fellow Latinos will feel offended, because for them there are some words that depending on the Latin American country, means a simple thing or something vulgar.

What do you think @CarlosRuiz-globalqss? Don't you also think the same?

@rakekniven @AndyScherzinger in es_ES locale there is a couple of good persons who are translating a bit every month, I am not be alone.

castillo92:
> "Liga" vs "Enlace" (Link)
> Carlos Ruiz

Here is a funny example of how a bad translation can mislead or cause a bad impression 😆 

"Liga" means thouse women's stockings that are very tight and transparent.... <women's tights>

Do you like to share a  witch 
Are you sure you want to share that `<women's tights>` link containing that file?

CarlosRuiz-globalqss:
:-)   Yes, there are more neutral words, I think "Enlace" is probably also used in Spain, so it could be used for both.

es_ES and es_LAT is a good idea, if there are maintainers
in case is hard to find maintainer, I think is better to have just one "es" translation of good quality - than two with poor quality.

castillo92:
> 
> 
> :-) Yes, there are more neutral words, I think "Enlace" is probably also used in Spain, so it could be used for both.
> 
> es_ES and es_LAT is a good idea, if there are maintainers
> in case is hard to find maintainer, I think is better to have just one "es" translation of good quality - than two with poor quality.

If were possible to copy the entire es_ES project into a new project es_LAT could be the beginning of that line of work with the advantage of not starting from the beginning.

Thats because thinking in the long term I believe that people from Latin America will prefer a language more similar to their own.

AndyScherzinger:
@castillo92 for Spanish translations there is no es_LAT on Android :/

castillo92:
> /

Hello Andy,

As far as I know, there is no such language in any project or in any section.

I only meant to suggest that it would be a good idea to differentiate the Spanish of Spain from the variation of Spanish spoken in Latin America.

In essence and to be precise, we all speak "Castilian", but there are words that differ from one dialect to another.

AndyScherzinger:
Thanks for the feedback @castillo92 but I guess that there is no todo then, tight? Android will fallback to es_ES in case es_* is active on a user's device. The translation missing will just fallback to the value present in es_ES and if there is no translation then it'll fallback to the default-default which is English. If it is easier as you commented above to simply override es_* with the latest es_ES version we have at the moment then I can do that if you want me to. But I think that will override any effort already spent on es_*

Issue #5368, opened 2020-02-01
Cannot open .eml with email app


mauricebauer:
### Steps to reproduce
1. Save an email to nextcloud (in my case from thunderbird) with file-extension .eml
2. Open the current nextcloud Android App (V3.10.0)
3. Open the .eml file

### Expected behaviour
- The popup with apps to open the file with contains email programs (such as GMail)

### Actual behaviour
- Only text viewers and other apps are shown, but no email app
- When opening a .eml file in the Google Drive app there is the option to open it with Gmail (so the file-type acceptance from the Gmail app is there)

### Environment data
Android version: Android 9 (MIUI Global 11.0.3) and tested on Android 7.0

Device model: Xiaomi Redmi Note 8 Pro

Stock or customized system: Stock from Xiaomi Global

Nextcloud app version: 3.10.0

Nextcloud server version: 18.0.0

![Screenshot_2020-02-01-11-06-37-418_android](https://user-images.githubusercontent.com/34477772/73590684-89e66a00-44e5-11ea-99e6-59a9d9ce6567.jpg)

AndyScherzinger:
mime type detection needs to be extended to map `.eml` to `message/rfc822` (while I can't say if this helps siince I don't know for which mime types the described apps would listen).

AndyScherzinger:
basically see https://github.com/nextcloud/android/pull/4434/commits/bff6395e420db0ee60386e1a0595a81786aca0be for a hint on how such a code change would have to look like and where it would be located

waleedsarwar86:
@AndyScherzinger I would like to work on this issue.

AndyScherzinger:
@waleedsarwar86  That be great 👍 and is highly appreciated!

daniboomberger:
@AndyScherzinger Is this Issue still available? I would like contribute to my first big Open Source Project and would love to help here. Have a nice day!!

AndyScherzinger:
Hi @daniboomberger yes, this issue is still available so feel free to pick it up 👍

PR #8416, opened 2021-05-15
.eml extension added to MIME

amirrezamirzaei:
added .eml extension as discussed on [#5368](https://github.com/nextcloud/android/issues/5368)


AndyScherzinger:
@amirrezamirzaei thanks for the PR, did you test if this triggers mail clients to open the file? Just asking because in the original issue I researched that message/rfc822 would be the mime type while latest Android 10/11 is opening the file even without the changes.

amirrezamirzaei:
sorry for the late response I only tested it for android 11 and it was working as expected.

PR #8513, opened 2021-06-02
Update strings.xml

berlin2123:
Improve text of battery_optimization:

您的设备可能启用了电池优化。仅当您从中排除（允许后台运行）此应用时，自动上传才能正常工作。
Your device may have battery optimization enabled. Auto upload will only work if you exclude this app from it (allowing it to run in the background).

去设置
Go to Settings

无法直接开启电池设置界面，请在系统设置中手动调整。
Unable to start battery settings directly. Please adjust manually in settings.

### DIFF
--- a/src/main/res/values-zh-rCN/strings.xml
+++ b/src/main/res/values-zh-rCN/strings.xml
@@ -80,9 +80,9 @@
     <string name="avatar">头像</string>
     <string name="away">离开</string>
     <string name="battery_optimization_close">关闭</string>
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@berlin2123 thanks for the PR, unfortunately we can'*t merge it since all translations are handled through https://www.transifex.com/nextcloud/nextcloud/android/ and automatically pulled here every night. But if you change them on transifex then they'll be here the next day :)

PR #8544, opened 2021-06-08
Changed wording regardings battery optimisation

rakekniven:
Reported at Transifex.

https://www.transifex.com/nextcloud/nextcloud/translate/#de_DE/$/160743806?q=issue%3Aopen

Signed-off-by: rakekniven <2069590+rakekniven@users.noreply.github.com>

### Testing 

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -795,7 +795,7 @@
     <string name="enter_filename">Please enter a filename</string>
     <string name="error_retrieving_templates">Error retrieving templates</string>
     <string name="battery_optimization_title">Battery optimization</string>
-    <string name="battery_optimization_message">Your device may have battery optimization enabled. Auto upload will only work if you exclude this app from it (allowing it to run in the background).</string>
+    <string name="battery_optimization_message">Your device may have battery optimization enabled. AutoUpload will only work properly if you exclude this app from it (allowing it to run in the background).</string>
     <string name="battery_optimization_disable">Go to Settings</string>
     <string name="battery_optimization_close">Close</string>
     <string name="battery_optimization_no_setting">Unable to start battery settings directly. Please adjust manually in settings.</string>
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -795,8 +795,8 @@
     <string name="enter_filename">Please enter a filename</string>
     <string name="error_retrieving_templates">Error retrieving templates</string>
     <string name="battery_optimization_title">Battery optimization</string>
-    <string name="battery_optimization_message">Your device may have battery optimization enabled. AutoUpload works only properly if you exclude this app from it.</string>
-    <string name="battery_optimization_disable">Disable</string>
+    <string name="battery_optimization_message">Your device may have battery optimization enabled. Auto upload will only work if you exclude this app from it (allowing it to run in the background).</string>
+    <string name="battery_optimization_disable">Go to Settings</string>
     <string name="battery_optimization_close">Close</string>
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):

berlin2123:
This should be better:
Your device may have battery optimization enabled. AutoUpload will only work properly if you exclude this app from it (allowing it to run in the background).

src/main/res/values/strings.xml
```diff
@@ -795,8 +795,8 @@
     <string name="enter_filename">Please enter a filename</string>
     <string name="error_retrieving_templates">Error retrieving templates</string>
     <string name="battery_optimization_title">Battery optimization</string>
-    <string name="battery_optimization_message">Your device may have battery optimization enabled. AutoUpload works only properly if you exclude this app from it.</string>
-    <string name="battery_optimization_disable">Disable</string>
+    <string name="battery_optimization_message">Your device may have battery optimization enabled. AutoUpload will only work properly if you exclude this app from it (allowing it to run in the background).</string>
+    <string name="battery_optimization_disable">Go to Settings</string>
```
tobiasKaminsky, line 799: Starting with Android 6 (M) it should be possible to directly disable battery optimization, thus "disable" should be correct for this.
Only < Android 6 "go to settings" is correct.

berlin2123:
The original reason: difficult to choose which key should be clicked

1) The original two buttons "Close" and "Disable", which have similar meaning. It was difficult to decide which key should be clicked.
2) "Close" = "Close battery optimization" or "Close this pop-up window"? The first understanding will lead to wrong clicks.
3) "Disable" = "Disable battery optimization" or "Exclude this app from battery optimization" or "Prohibit this app from running in the background in battery optimization"?

I Recommend changing "Disable" to "Exclude" for battery_optimization_disable. As the "Exclude" is one key word in the announcement message, It should be easier to choose between "Close" and "Exclude".

src/main/res/values/strings.xml
berlin2123: The original reason: difficult to decide which key should be clicked
1. The original two buttons "Close" and "Disable", which have similar meaning. 
2. "Close" = "Close battery optimization" or "Close this pop-up window"? The first understanding will lead to wrong clicks.
3. "Disable" = "Disable battery optimization" or "Exclude this app from battery optimization" or "Prohibit this app from running in the background in battery optimization"?

Besides "Go to Settings", recommend changing "Disable" to "Exclude" for battery_optimization_disable. As the "Exclude" is one key word in the announcement message, It should be easier to choose between "Close" and "Exclude".

berlin2123:
May anyone tell me the difference between "Close a function" and "Disable a function"? 
Do both mean not to let that function work? 

Thus, this "Disable" should be changed.  

Use "Exclude" or "Go to Settings"?   Let's vote.

rakekniven:
Hey Android users. I am off the building for this one.

rakekniven:
Please merge

AndyScherzinger:
> Please merge

@rakekniven see https://github.com/nextcloud/android/pull/8544#pullrequestreview-699551026 - we obviously won't merge!

rakekniven:
> > Please merge
> 
> @rakekniven see [#8544 (review)](https://github.com/nextcloud/android/pull/8544#pullrequestreview-699551026) - we obviously won't merge!

Thanks for pointing me to this. Forgot the history.
So we better close here?

AndyScherzinger:
As for the PR, I think so. Fixing it in the way @tobiasKaminsky pointed out there would need to be a code change first to support both Android implementations by version codes to end up with 2 strings for translation.

berlin2123:
I suggest changing "Disable" to "Exclude", so there is only one string.

PR #8632, opened 2021-06-27
Fix sync contact npe

ezaquarii:
Another fragment lifecycle bug.


AndyScherzinger (approved):

AndyScherzinger:
Looks good. Die Lint errors come from transifex I guess. At least the zero-quantity translations are there

tobiasKaminsky:
> Looks good. Die Lint errors come from transifex I guess. At least the zero-quantity translations are there

Solution is in https://github.com/nextcloud/android/pull/8634

PR #8643, opened 2021-06-30
Use localID from webdav

### LINKED ISSUE #9826: Android app fails to play media files (music) from bind-mounted folder
I cannot play music files with my nextcloud android app. I can navigate to the files and select them, but I only receive the error "Could not read the media file". The server log files show a 404 which is strange, since the files are clearly there and can actually be played without an issue using the web UI. 

The folder holding the files is mounted (mount --bind) into the user's data files section. File permissions are like those files created by nextcloud itself: user and group ownership for account "www-data", the files are group writeable and readable for everyone. The server system is a standard Ubuntu system. 

### Steps to reproduce
1. have some media files mounted into your nextcloud files section (ogg or mp3) (mount --bind)
2. scan the file system to import those files (sudo -u www-data php -f public/occ files:scan arkascha)
3. open the nextcloud app on an android device
4. navigate to the media files and select one

### Expected behaviour
The file should get played. Actually the UI clearly suggests that it is trying to play the file, there is even a play button. 

### Actual behaviour
The file is not played, I receive an error toast "Could not read the media file". 

I cannot reproduce the same situation (mount --bind) on that system, so I cannot check this. Sorry. 

### Environment data
Android version:
LineageOS 18.1 (Android 11)

Device model: 
Oneplus 7 Pro

Stock or customized system:
Customized: LineageOS

Nextcloud app version:
Version 3.19.0, installed from google

Nextcloud server version:
Nextcloud version 23.1

Reverse proxy:
---

#### Web server error log
```
31.17.208.40 - - [13/Feb/2022:21:32:38 +0100] "POST /ocs/v2.php/apps/dav/api/v1/direct?format=json HTTP/1.1" 404 1456 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.19.0"
31.17.208.40 - - [13/Feb/2022:21:32:34 +0100] "GET /index.php/apps/files/api/v1/thumbnail/256/256/Multimedia/music/All%20Saints/Black%20Coffee.mp3 HTTP/1.1" 404 822 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.19.0"
31.17.208.40 - - [13/Feb/2022:21:32:51 +0100] "POST /ocs/v2.php/apps/dav/api/v1/direct?format=json HTTP/1.1" 404 1456 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.19.0"
```

```
- - -
```

### LINKED ISSUE #10261: impossible to watch video on stream
If I try to open video stored on my nextcloud instance, th app tell me :
"Internal stream not possible" and I have this error :

### Cause of error
```java
Exception in thread "main" java.lang.NullPointerException
    at com.owncloud.android.ui.preview.PreviewVideoActivity.onStart(PreviewVideoActivity.kt:225)
    at android.app.Instrumentation.callActivityOnStart(Instrumentation.java:1435)
    at android.app.Activity.performStart(Activity.java:8018)
    at android.app.ActivityThread.handleStartActivity(ActivityThread.java:3457)
    at android.app.servertransaction.TransactionExecutor.performLifecycleSequence(TransactionExecutor.java:221)
    at android.app.servertransaction.TransactionExecutor.cycleToPath(TransactionExecutor.java:201)
    at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:173)
    at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066)
    at android.os.Looper.loop(Looper.java:223)
    at android.app.ActivityThread.main(ActivityThread.java:7664)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)

```

### App information
* ID: `com.nextcloud.client`
* Version: `30200190`
* Build flavor: `gplay`

### Device information
* Brand: `OnePlus`
* Device: `OnePlus7Pro`
* Model: `GM1917`
* Id: `RQ3A.211001.001`
* Product: `OnePlus7Pro`

### Firmware
* SDK: `30`
* Release: `11`
* Incremental: `173fcd6e23`

With an older nextcloud app version (3.19.1) : no problem.
I think i didn't find this problem in existent issues

tobiasKaminsky:
Needs https://github.com/nextcloud/android-library/pull/651
After: 
- [x] https://github.com/nextcloud/android/pull/11098

@juliushaertl
~~we need to 
- do not use localID (which might make sense in terms of federation)
- or expose unique server id via capabilities
- or have server return localID via webdav

EDIT: using localId via webdav is the solution!

E.g. my instance has 14 chars for unique id @juliushaertl 
~~To handle this properly we need uniqe id exposed via registered capability, if this is ok for you.~~

Fixes #9826
Fixes #10261

### DIFF
--- a/app/src/androidTest/java/com/owncloud/android/datamodel/FileDataStorageManagerIT.java
+++ b/app/src/androidTest/java/com/owncloud/android/datamodel/FileDataStorageManagerIT.java
@@ -369,4 +369,5 @@ abstract public class FileDataStorageManagerIT extends AbstractOnServerIT {
 
         assertEquals(capability.getUserStatus(), newCapability.getUserStatus());
     }
+
 }
--- a/app/src/androidTest/java/com/owncloud/android/datamodel/FileDataStorageManagerIT.java
+++ b/app/src/androidTest/java/com/owncloud/android/datamodel/FileDataStorageManagerIT.java
@@ -344,7 +344,7 @@ abstract public class FileDataStorageManagerIT extends AbstractOnServerIT {
                                               targetContext).execute(client).isSuccess());
 
         OCFile newFile = new OCFile("/1/1/1.txt");
-        newFile.setRemoteId("123");
+        newFile.setRemoteId("12345678");
 
         sut.saveNewFile(newFile);
     }
--- a/app/src/main/java/com/nextcloud/client/database/NextcloudDatabase.kt
+++ b/app/src/main/java/com/nextcloud/client/database/NextcloudDatabase.kt
@@ -58,7 +58,8 @@ import com.owncloud.android.db.ProviderMeta
     ],
     version = ProviderMeta.DB_VERSION,
     autoMigrations = [
-        AutoMigration(from = 65, to = 66)
+        AutoMigration(from = 65, to = 66),
+        AutoMigration(from = 66, to = 67)
     ],
     exportSchema = true
 )
--- a/app/src/main/java/com/owncloud/android/datamodel/OCFile.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/OCFile.java
@@ -45,6 +45,7 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 import androidx.core.content.FileProvider;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import third_parties.daveKoeller.AlphanumComparator;
 
 public class OCFile implements Parcelable, Comparable<OCFile>, ServerFileInterface {
@@ -83,6 +84,7 @@ public class OCFile implements Parcelable, Comparable<OCFile>, ServerFileInterfa
     private String etagOnServer;
     private boolean sharedViaLink;
     private String permissions;
+    private long localId; // unique fileId for the file within the instance
     private String remoteId; // The fileid namespaced by the instance fileId, globally unique
     private boolean updateThumbnailNeeded;
     private boolean downloading;
@@ -166,6 +168,7 @@ public class OCFile implements Parcelable, Comparable<OCFile>, ServerFileInterfa
         etagOnServer = source.readString();
         sharedViaLink = source.readInt() == 1;
         permissions = source.readString();
+        localId = source.readLong();
         remoteId = source.readString();
         updateThumbnailNeeded = source.readInt() == 1;
         downloading = source.readInt() == 1;
@@ -208,6 +211,7 @@ public class OCFile implements Parcelable, Comparable<OCFile>, ServerFileInterfa
         dest.writeString(etagOnServer);
         dest.writeInt(sharedViaLink ? 1 : 0);
         dest.writeString(permissions);
+        dest.writeLong(localId);
         dest.writeString(remoteId);
         dest.writeInt(updateThumbnailNeeded ? 1 : 0);
         dest.writeInt(downloading ? 1 : 0);
@@ -284,8 +288,7 @@ public class OCFile implements Parcelable, Comparable<OCFile>, ServerFileInterfa
     }
 
     /**
-     * Can be used to check, whether or not this file exists in the database
-     * already
+     * Can be used to check, whether or not this file exists in the database already
      *
      * @return true, if the file exists in the database
      */
@@ -486,6 +489,7 @@ public class OCFile implements Parcelable, Comparable<OCFile>, ServerFileInterfa
         etagOnServer = null;
         sharedViaLink = false;
         permissions = null;
+        localId = -1;
         remoteId = null;
         updateThumbnailNeeded = false;
         downloading = false;
@@ -594,17 +598,14 @@ public class OCFile implements Parcelable, Comparable<OCFile>, ServerFileInterfa
         return !TextUtils.isEmpty(getFileName()) && getFileName().charAt(0) == '.';
     }
 
-    /**
-     * The unique fileId for the file within the instance
-     *
-     * @return file fileId, unique within the instance
-     */
-    @Nullable
-    public String getLocalId() {
-        if (getRemoteId() != null) {
-            return getRemoteId().substring(0, 8).replaceAll("^0*", "");
+    @SuppressFBWarnings("STT")
+    public long getLocalId() {
+        if (localId > 0) {
+            return localId;
+        } else if (remoteId != null) {
+            return Long.parseLong(remoteId.substring(0, 8).replaceAll("^0*", ""));
         } else {
-            return null;
+            return -1;
         }
     }
 
@@ -770,6 +771,10 @@ public class OCFile implements Parcelable, Comparable<OCFile>, ServerFileInterfa
 
     public void setFileId(long fileId) {
         this.fileId = fileId;
+    }
+
+    public void setLocalId(long localId) {
+        this.localId = localId;
     }
 
     public void setParentId(long parentId) {
--- a/app/schemas/com.nextcloud.client.database.NextcloudDatabase/67.json
+++ b/app/schemas/com.nextcloud.client.database.NextcloudDatabase/67.json
@@ -1,0 +1,1130 @@
+{
+    "formatVersion": 1,
+    "database": {
+        "version": 67,
+        "identityHash": "be969fd72e75dd6116b4fa746bcf5b6a",
+        "entities": [
+            {
+                "tableName": "arbitrary_data",
+                "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`_id` INTEGER PRIMARY KEY AUTOINCREMENT, `cloud_id` TEXT, `key` TEXT, `value` TEXT)",
+                "fields": [
+                    {
+                        "fieldPath": "id",
+                        "columnName": "_id",
+                        "affinity": "INTEGER",
+                        "notNull": false
+                    },
+                    {
+                        "fieldPath": "cloudId",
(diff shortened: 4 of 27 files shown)


src/main/java/com/owncloud/android/datamodel/OCFile.java
```diff
@@ -552,7 +552,8 @@ public boolean isHidden() {
      * @return file fileId, unique within the instance
      */
     public String getLocalId() {
-        return getRemoteId().substring(0, 8).replaceAll("^0*", "");
+        // fileId, instanceId. instanceId is 12 chars long
+        return getRemoteId().substring(0, getRemoteId().length() - 12).replaceAll("^0*", "");
```
tobiasKaminsky, line 556: We cannot rely on 12 long instance id!

juliushaertl:
> EDIT: using localId via webdav is the solution!

yep sounds like the better approach :+1:

app/src/main/java/com/nextcloud/client/database/NextcloudDatabase.kt
```diff
@@ -57,7 +57,7 @@ import com.owncloud.android.db.ProviderMeta
     ],
     version = ProviderMeta.DB_VERSION,
     autoMigrations = [
-        AutoMigration(from = 65, to = 66)
+        AutoMigration(from = 65, to = 67)
```
tobiasKaminsky, line 60: @AlvaroBrey is this needed?
I also changed it in https://github.com/nextcloud/android/pull/8643/files#diff-ffa8f9b74f7af219ebf0927bb94e6c8d03e2b34eb9e39b1287edb1b65532a72fR38

tobiasKaminsky: If needed, can we combine this? Otherwise I fear we will forget it at some point.

app/schemas/com.nextcloud.client.database.NextcloudDatabase/66.json
```diff
@@ -2,7 +2,7 @@
     "formatVersion": 1,
```
AlvaroBrey, line 2: Please undo changes to this file. We must _never_ change already existing schemas when changing database, that's why the new one is created. Otherwise migrations will break.

app/src/main/java/com/nextcloud/client/database/NextcloudDatabase.kt
AlvaroBrey:
```suggestion
        AutoMigration(from = 65, to = 66),
        AutoMigration(from = 66, to = 67),
```

See https://developer.android.com/training/data-storage/room/migrating-db-versions#automated. Your original change would not allow migration from 66 to 67, only from 65.

> If needed, can we combine this? Otherwise I fear we will forget it at some point.

No, we can't combine this. The migration specs need to reference static DB versions as they must be immutable in time, not a constant that will be changed later.

app/schemas/com.nextcloud.client.database.NextcloudDatabase/66.json
AlvaroBrey: To prevent this happening, you should bump the DB version _before_ making changes to the entities.

app/src/androidTest/java/com/owncloud/android/datamodel/FileDataStorageManagerIT.java
```diff
@@ -369,4 +369,11 @@ public void testOCCapability() {
 
         assertEquals(capability.getUserStatus(), newCapability.getUserStatus());
     }
+
+    @Test
+    public void saveFolder() {
+
+    }
```
AlvaroBrey, line 376: Empty test?

app/src/main/java/com/owncloud/android/datamodel/OCFile.java
```diff
@@ -283,9 +286,13 @@ public String getRemotePath() {
         }
     }
 
+    @Override
+    public String getImageKey() {
+        return remoteId;
+    }
```
AlvaroBrey, line 292: This feels out of place, the `OCFile` shouldn't have image-specific things in my opinion. I'd just use `getRemoteId` where it is needed.

AlvaroBrey (changes requested):

app/schemas/com.nextcloud.client.database.NextcloudDatabase/66.json
tobiasKaminsky: I never intentionally changed there something.
Must be done by Room…

I only changed FileEntity.

tobiasKaminsky: > To prevent this happening, you should bump the DB version before making changes to the entities.

That might be the keypoint! I changed Entity, started app, Room complained and then I adjusted DB version.

app/src/main/java/com/owncloud/android/datamodel/OCFile.java
tobiasKaminsky: This is such an old PR, that I even do not know why I did this back then.
And you are right, this feels totally off!

app/schemas/com.nextcloud.client.database.NextcloudDatabase/66.json
AlvaroBrey: Yep, the schemes get autogenerated by Room when building. In any case, just undo the changes in 66.json and it will be good to go.

app/src/main/java/com/owncloud/android/datamodel/OCFile.java
tobiasKaminsky: I know remember.
We have TrashbinFile, FileVersion and OCFile, all extending form ServerFileInterface.
imageKey was then used to abstract from actual detail on ThumbnailCacheManager.

But I reverted it.

AlvaroBrey (approved):

tigermick:
hi,
I have a problem here : https://github.com/nextcloud/android/issues/10261
and your Nextcloud QA release here : https://www.kaminsky.me/nc-dev/android-artifacts/8643.apk fix my problem

AlvaroBrey:
@tobiasKaminsky is this available in all supported server versions?

Btw the new attribute in the DB will likely conflict with #11098 , so I'd rather merge that one first and solve conflicts here. Otherwise we risk losing the new attribute.

AlvaroBrey:
@tobiasKaminsky can you see what's happening with the failing test?

tobiasKaminsky:
> @tobiasKaminsky is this available in all supported server versions?

Yes, I tested it with NC16, our oldest supported version, and it returns:
```
<oc:id>00000002oc8wojm72e0q</oc:id>
<oc:fileid>2</oc:fileid>
```

tobiasKaminsky:
> @tobiasKaminsky can you see what's happening with the failing test?

Strange, I do not understand why it fails with this change.
It should have been failing either all the time, or not at all.

I will continue to check.

app/src/main/java/com/owncloud/android/datamodel/OCFile.java
```diff
@@ -594,17 +598,14 @@ public boolean isHidden() {
...
-    @Nullable
-    public String getLocalId() {
-        if (getRemoteId() != null) {
-            return getRemoteId().substring(0, 8).replaceAll("^0*", "");
+    @SuppressFBWarnings("STT")
+    public long getLocalId() {
+        if (localId > 0) {
+            return localId;
+        } else if (remoteId != null) {
+            return Long.parseLong(remoteId.substring(0, 8).replaceAll("^0*", ""));
```
AlvaroBrey, line 606: Are there servers with a remoteID shorter than 8 characters? If so we should do `substring(0, Math.min(remoteId.length(), 8)` instead of `substring(0.8)`.

tobiasKaminsky: I think 8 is minimum, but there might be some with more than 8.
But as this is the old behaviour, it should not break any existing installation.

AlvaroBrey: Then I don't understand why some tests failed in this pr due to a too short remoteId, but didn't fail before :\

AlvaroBrey (approved):

app/src/main/java/com/owncloud/android/datamodel/OCFile.java
tobiasKaminsky: Me neither. Which one is failing? Then I can try to debug it locally.

tobiasKaminsky: https://github.com/nextcloud/android/pull/8643/commits/e1d564026f7c479b61d4404083126fbee71fbbdb#diff-940d35bacfe8905a4b0b8019f3b78e40fcb366e8b3f313dc350ecbdf671f8385R249

Here it is because of:
It tries to save file in FileDataStorageManager:460
-> and saves localId, which is then as fallback tries to substring(0,8).
In this case (on master) we just used remoteId, as is. And thus it never failed.

AlvaroBrey:
```
/drone/src/app/src/androidTest/java/com/nextcloud/client/EndToEndRandomIT.java:656: error: cannot find symbol
--
712 | assertTrue(new DeletePrivateKeyOperation().execute(client).isSuccess());

```

tobiasKaminsky:
Library merged, so let us merge this asap.

PR #8032, opened 2021-02-23
[do not merge] Debug only: debug failing tests on emulator

tobiasKaminsky:
Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

### Testing 
Writing tests is very important. Please try to write some tests for your PR. 
If you need help, please do not hesitate to ask in this PR for help.

[unit tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#unit-tests)
[instrumented tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#instrumented-tests)
[UI tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#ui-tests)

- [ ] Tests written, or not not needed

PR #8791, opened 2021-07-30
Fix connected test crashes

alvarobrey:
This is a fix for the "Process crashed" message that was crashing the `connectedTest`s in my device and emulator.
This should additionally fix crashes in Drone and in screenshot tests.

Summary of the changes:
- In tests where `Looper.prepare()` was used, made sure to quit the looper after the tests. Otherwise the thread would remain active indefinitely and the process would get killed as an ANR.
- In `DialogFragmentIT` tests, force set account tokens. This works around a race condition (I think) where the account token was peeked _before_ it was saved in the AccountManager cache.
- Changed `server.com` and `server` to `nextcloud.localhost` everywhere. `server.com` is a real domain with DNS records, which gives a timeout when trying to connect. This causes crashes and also dramatically slows down tests. An invalid domain such as `nextcloud.localhost` should fail to resolve immediately, preventing all of that. Note: `nextcloud.localhost` is just the first thing I thought of. Better suggestions welcome!

Related: https://github.com/nextcloud/android/pull/8032, https://github.com/nextcloud/android/pull/8059

### DIFF
--- a/src/androidTest/java/com/nextcloud/client/SettingsActivityIT.java
+++ b/src/androidTest/java/com/nextcloud/client/SettingsActivityIT.java
@@ -86,6 +86,7 @@ public class SettingsActivityIT extends AbstractIT {
         SettingsActivity sut = activityRule.launchActivity(null);
         sut.handleMnemonicRequest(intent);
 
+        Looper.myLooper().quitSafely();
         assertTrue(true); // if we reach this, everything is ok
     }
 }
--- a/src/androidTest/java/com/nextcloud/client/SyncedFoldersActivityIT.java
+++ b/src/androidTest/java/com/nextcloud/client/SyncedFoldersActivityIT.java
@@ -71,7 +71,7 @@ public class SyncedFoldersActivityIT extends AbstractIT {
                                                                    false,
                                                                    false,
                                                                    true,
-                                                                   "test@https://server.com",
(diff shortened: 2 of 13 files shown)


alvarobrey:
I see there's still some crashes in CI. I'll try to work that out.

AndyScherzinger:
Thanks a lot @alvarobrey for tackling the automated tests / automated QA ❤️ This is really great!!! 🎉

alvarobrey:
Just confirmed that after the current patches,  the remaining crashes in the screenshot tests are fixed by using an invalid domain instead of server.com (see [my fork](https://github.com/alvarobrey/android/pull/2)), as was suspected in https://github.com/nextcloud/android/pull/8032.

I'll close https://github.com/nextcloud/android/pull/8792 and consolidate it into this PR, while also trying to make some changes so that the Drone integration tests work as expected.

alvarobrey:
Seems like everything ran to completion! :partying_face:  (except the QA build but I can't do anything about that)

AndyScherzinger:
> Seems like everything ran to completion! 🥳 (except the QA build but I can't do anything about that)

No worries QA build only works for PRs from _within_ the repository due the the missing secrets on forks 👍

tobiasKaminsky:
Wow!
Many thanks for this PR!

tobiasKaminsky (approved):

PR #8792, opened 2021-07-30
Use invalid domain for server in integration tests

alvarobrey:
Changed it elsewhere (layout files, unit tests) for consistency too.

This speeds up integration tests significantly, as the domain resolution fails immediately, in contrast to the timeout for server.com, which can take a while.

In my machine and emulator, this takes almost 3 minutes off of the connected tests runtime (from ~9:30 to ~6:40).

As a bonus, we also avoid making actual network requests to server.com!

The domain I've chosen is just the first idea I had for a fake domain. Better suggestions are highly appreciated!

### DIFF
--- a/src/androidTest/java/com/nextcloud/client/SyncedFoldersActivityIT.java
+++ b/src/androidTest/java/com/nextcloud/client/SyncedFoldersActivityIT.java
@@ -71,7 +71,7 @@ public class SyncedFoldersActivityIT extends AbstractIT {
                                                                    false,
                                                                    false,
                                                                    true,
-                                                                   "test@https://server.com",
+                                                                   "test@https://nextcloud.nodomain",
                                                                    0,
                                                                    0,
                                                                    true,
--- a/src/androidTest/java/com/nextcloud/client/account/RegisteredUserTest.kt
+++ b/src/androidTest/java/com/nextcloud/client/account/RegisteredUserTest.kt
@@ -38,7 +38,7 @@ class RegisteredUserTest {
 
     private companion object {
         fun buildTestUser(accountName: String): RegisteredUser {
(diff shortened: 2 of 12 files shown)


AndyScherzinger:
Hi @alvarobrey,

@tobiasKaminsky and I are AFK for the weekend, but you might also check out #8059. "server.com" is somethign we try/tried to bind as a spawned server running end-to-end tests with a real Nextcloud server on Drone our container based CI at https://drone.nextcloud.com/nextcloud/android

alvarobrey:
> Hi @alvarobrey,
> 
> @tobiasKaminsky and I are AFK for the weekend, but you might also check out #8059. "server.com" is something we try/tried to bind as a spawned server running end-to-end tests with a real Nextcloud server on Drone our container based CI at https://drone.nextcloud.com/nextcloud/android

I see. What was the reason behind using a real domain for this? AFAIK an invalid domain should work fine as long as it resolves locally, which you were already setting up for server.com anyway.

In any case, I did't change `gradle.properties`, so tests using the `TEST_SERVER_URL` variable should be running the same as before I think.

alvarobrey:
Superseded by https://github.com/nextcloud/android/pull/8791 since this seems to have influence on whether the tests crash or not.

AndyScherzinger:
> I see, what was the reason behind using a real domain for this? AFAIK an invalid domain should work fine as long as it resolves locally, which you were already setting up for server.com anyway.

You are absolutely right 👍 

> In any case, I did't change gradle.properties, so tests using the TEST_SERVER_URL variable should be running the same as before, I think.

I guess so, yes, while @tobiasKaminsky knows best

> Thanks!

Happy to have you around ❤️

PR #8828, opened 2021-08-08
Only perform deep connection test on an unmeted connection

hrk:
Using the isMetered property of the Connection object allows us to avoid performing the deep connection check on metered WIFI connections.

Depending on the rationale of the initial implementation of the deep check, the condition could be simplified to c.isConnected() and !c.isMetered().

### DIFF
--- a/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
+++ b/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
@@ -61,7 +61,7 @@ class ConnectivityServiceImpl implements ConnectivityService {
     @Override
     public boolean isInternetWalled() {
         Connectivity c = getConnectivity();
-        if (c.isConnected() && c.isWifi()) {
+        if (c.isConnected() && c.isWifi() & !c.isMetered()) {
 
             Server server = accountManager.getUser().getServer();
             String baseServerAddress = server.getUri().toString();


src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
```diff
@@ -61,7 +61,7 @@ public GetMethod invoke(String url) {
     @Override
     public boolean isInternetWalled() {
         Connectivity c = getConnectivity();
-        if (c.isConnected() && c.isWifi()) {
+        if (c.isConnected() && c.isWifi() && !c.isMetered()) {
```
tobiasKaminsky, line 64: If we then have a metered connection this will be false and thus it will go to else branch:
` return !c.isConnected() ` --> false or true (depending on isConnected()

as the the function should indicate if an internet connection is metered/walled, it then might return the wrong.
What about adding before this:
```
if (c.isMetered()) {
return true;
}
```
Then the whole deep check is only done if Andoid is not aware that connection is metered.

tobiasKaminsky (changes requested):
See inline comment

src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
hrk: I respectfully disagree, because I don't think that isInternetWalled() should inherently return true on metered connections. A metered connection is not walled. In fact, the current UI for auto-uploads allows the upload on metered connections too.

ezaquarii: @tobiasKaminsky 
The logic is that this check is performed only on Wi-Fi, and I believe it is done so under the assumption that Wi-Fi is free/unmetered.

For anything else that might not be free, it takes the `else` branch and does a simple heuristic.

I think that using metered status here makes sense.

ezaquarii (changes requested):
@hrk This change is missing the unit test.

tobiasKaminsky:
> @hrk This change is missing the unit test.

Can you create a test? If not, please ping me and we can do it.

PR #8831, opened 2021-08-08
Fix: boxed primitive for compare

hrk:
[Broken Window Theory] As per current SpotBugs report, replace all usages
of boxed Long.compareTo(Long) with Long.compare(long, long)

Note: the justification for the SuppressFBWarning annotation in FileStorage appears to be obsolete (has no current application) and, so, I removed the annotation.


tobiasKaminsky:
Thank you very much and sorry for the long delay!

Issue #7905, opened 2021-01-28
Cannot get rid of file upload conflict notification


jdaviescoates:
### Steps to reproduce
1. Get a notification like this

![Screenshot_20210128-155456](https://user-images.githubusercontent.com/2113743/106163554-369f0a00-6181-11eb-9a69-8a6a2b7b4080.png)

2. Click it and end up at a dialog like this:

![Screenshot_20210128-120617](https://user-images.githubusercontent.com/2113743/106163634-4c143400-6181-11eb-9d32-7a3a9be20a00.png)

3. Whatever I do (choose original file, choose the new file, or cancel) doesn't make any difference, within a minutes or so (or even less) I get the exact same notification about the exact same conflict again. Again, and again, and again

### Expected behaviour
- Choose which file to keep, then not get the notification again.

### Actual behaviour
- I keep getting the same notification about the same file conflict over and over and over again.

### Environment data
Android version:

8.1.0

Device model: 

Wileyfox Swift 2 Plus

Stock or customized system:

Stock

Nextcloud app version:

3.14.3

Nextcloud server version:

20.0.6

Reverse proxy:

?

dremerb:
Can confirm this behavior. For me this happened after restoring a backup and re-syncing - I have set all folders to "Skip uploading" (not sure about the correct name, German system here) and the last conflict remains in the notifications.
Even after restarting the device, the sync starts again and the notification reappears.

jdaviescoates:
I finally managed to get rid of the incessant notifications but selecting both files.

Klemens2:
I have the same problem very often too.

tobiasKaminsky:
What kind of file is this? Does it come via auto upload?

Klemens2:
File types are: pictures (JPG) and videos (MP4).

 

Yes, it’s a matter of auto-upload (camera-upload).

 

Von: Tobias Kaminsky <notifications@github.com> 
Gesendet: Donnerstag, 25. Februar 2021 08:50
An: nextcloud/android <android@noreply.github.com>
Cc: Klemens2 <klemens.heuboeck@gmx.at>; Comment <comment@noreply.github.com>
Betreff: Re: [nextcloud/android] Cannot get rid of file upload conflict notification (#7905)

 

What kind of file is this? Does it come via auto upload?

dremerb:
When trying to get it working again, I encountered the following:
Deleted all of the existing files in Nextcloud. NC client on Android still had sync activated.
[...]

PR #8923, opened 2021-09-01
Fix sync conflict behavior if user chooses to keep server version

BurningKarl:
Dear Nextcloud Android app developers,

I've recently noticed a bug and decided to fix it myself. Similarly to what is described in #7905 the sync conflict notifications reappeared after a while if I chose to keep the "already existing file" (i.e. the server version), but did not when I chose to keep the "new file" (i.e. the local version). While searching for the reason for this in the code, I noticed that when the user selects to keep the server version (`ConflictsResolveActivity.Decision.KEEP_SERVER`) a new download is triggered to overwrite the local file with the server version. The conflict should then be marked as resolved if the download finishes successfully inside of FileDownloader.java (line 672). Unfortunately, the sync conflict is never deleted because `FileDownloader.notifyDownloadResult` stops early if the download was successful to avoid generating a notification for successful downloads.

I considered only fixing the bug inside of `FileDownloader.notifyDownloadResult` but I feel like overwriting the local file with the server version is already unexpected behavior for most users (at least it was for me). The sync conflict dialog only asks which version of the file the user wants to keep _on the server_ and does not indicate any consequences for the local file. Similarly, the existing tests in `FileUploaderIT.kt` identify the "keep server" option with `NameCollisionPolicy.CANCEL` which just cancels the upload if a name collision occurs and does not trigger any download.

To underline my point consider this scenario: You download a file to your phone's Download directory and make some changes. You upload it to the appropriate Nextcloud folder but because the old version has the same name there is a sync conflict. You decide to stop the upload, rename the local file, and upload it again. So, you click on the sync conflict notification and choose to "keep the already existing file" in the popup, rename the local file and upload it again. All of your actions are reasonable, but it turns out that now you have the same file twice on your sever and all your local changes are gone. That can't be right.

My PR therefore changes the sync conflict behavior such that choosing `KEEP_SERVER` marks the conflict as resolved (i.e. deletes the upload) without triggering any download. Additionally I refactored and simplified `FileDownloader.notifyDownloadResult`.

### Testing 

Ideally, there would be tests to confirm that choosing any valid option in the sync conflict resolution dialog marks the sync conflict as resolved, but I'm not knowledgeable enough to know whether that's feasible or useful and how to do it.

### DIFF
--- a/src/main/java/com/owncloud/android/files/services/FileDownloader.java
+++ b/src/main/java/com/owncloud/android/files/services/FileDownloader.java
@@ -56,7 +56,6 @@ import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.lib.resources.files.FileUtils;
 import com.owncloud.android.operations.DownloadFileOperation;
 import com.owncloud.android.providers.DocumentsStorageProvider;
-import com.owncloud.android.ui.activity.ConflictsResolveActivity;
 import com.owncloud.android.ui.activity.FileActivity;
 import com.owncloud.android.ui.activity.FileDisplayActivity;
 import com.owncloud.android.ui.dialog.SendShareDialog;
@@ -116,8 +115,6 @@ public class FileDownloader extends Service
     private int mLastPercent;
 
     private Notification mNotification;
-
-    private long conflictUploadId;
 
     public boolean mStartedDownload = false;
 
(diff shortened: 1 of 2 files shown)


AlvaroBrey (commented):
I'm currently  leaning towards #9036, as it introduces less behaviour changes while also improving the workflow a bit.

The main reason I want to keep behaviour changes to a minimum is because this activity is currently agnostic to the fact that a upload may be manually triggered or automatically (from synced folders), and I don't think those conflicts should be treated the same.

For example, your change in `KEEP_SERVER` makes sense for manual uploads but not for autoupload, as in the latter case the conflict would just pop up again in the next sync cycle. Of course some warnings could be added to make the user understand that "keep server" will overwrite the existing file, but that's beside the point.

Therefore, I'm going to go for #9036 until we do a proper overhaul of the conflicts and how autoupload relates to it. Sorry for the late review.

BurningKarl:
Hi @AlvaroBrey, thanks for reviewing this PR. I understand your reasoning to keep the changes to a minimum until a proper overhaul is done. I didn't think about the auto-upload case to be honest and I can see how my changes would break things there. I still think the conflict overhaul is desperately needed, and should address the scenario, where one just wants to cancel the upload, and the naming issues brought up by @wknds.

Last thought: If you are going to split up conflict resolution into the normal upload and auto-upload case, it does not make sense to associate "keep server version" with overwriting the local file in either case.

Issue #5265, opened 2020-01-21
Offline files on other apps without opening Nextcloud app


ktuulos:
On Dropbox, it is possible to mark certain files to be available offline. Those files can then be opened on other applications, when Dropbox is offline, without opening the Dropbox application. Those files are also updated automatically whenever Dropbox gets online.

On Nextcloud, when a file is marked as available offline, it can not be opened from other application. It can be opened only through Nextcloud application. This makes it very difficult for opening those files, compared using on e.g. desktop, where a file can be opened just by navigating to file and opening it.

One great use case of these is that on Dropbox, it is possible to create shortcuts to certain files, which are available offline, and place those shortcuts to home screen. Then, when these shortcuts are clicked, the application is directly opened without opening the Dropbox application. Similar behaviour would be needed for Nextcloud also.

tobiasKaminsky:
You can open any downloaded file via a file browser, or if the other app supports it via document storage system (like an open).

So e.g. if you have keepass marked as available offline, you can use keepass2android to browse for it. There is no need to open Nextcloud for it.

ktuulos:
OK, now I realized, what the problem is. Nextcloud browser gives such path, which can not be opened outside of Nextcloud browser.

When a file is opened directly from Nextcloud browser, the 3rd party application will get file name as follows:
content://org.nextcloud.files/external_files/emulated/0/nextcloud/....
This path can not be opened directly from 3rd party application.

When the application navigates itself to the file, using system navigator, the path will be as follows:
content://com.android.externalstorage.documents/document/primary¤3Anextcould%2F .... etc

So, the Nextcloud browser should be modified to give the permanent path, not a temporary path pointing to Nextcloud browser's cache.

pstorch:
I use https://github.com/Kunzisoft/KeePassDX which uses the Nextcloud App as Document Provider. My keepass file has this green checkmark, which indicates it is available offline, but KeePassDX can't open the document URI if I'm offline.

t-m-w:
This issue has been a major annoyance for me ever since I started using KeePassDX instead of Keepass2Android, which has native Nextcloud support.

Oddly, this problem was not present in [3.10.0 (January 17, 2020)](https://github.com/nextcloud/android/releases?after=dev-20200128), which was the latest version available when this GitHub issue was opened. However, it exists today, and it has been around since November, affecting all versions 3.14.0 and up.

### (Re-?)Introduction of Issue
[...]

PR #8925, opened 2021-09-02
Fix opening downloaded files from Documents Provider when offline (#5265)

t-m-w:
This is my attempt at a fix for #5265. You have my full consent to use this PR however you see fit, including making any adjustments -- no need to wait for me to make your desired changes, since in my view this is a high-priority bug that makes Nextcloud for Android nearly unusable, at least for me. I don't necessarily need any credit whatsoever if you want to re-do this your own way. I would just love to have this fixed one way or another without needing to maintain a fork or something.

The code for popping up a toast that the file is not sync'd is copied from an earlier version of DocumentsStorageProvider.java, and that's what the added import lines are about.


newhinton:
A short test didn't show any issues, great work!

newhinton:
@t-m-w i think it could be an improvement to show the error only when appropriate, eg. when internet-connection is available AND the sync failed. Otherwise the error is expected and may be annoying.

newhinton:
>  I guess in some cases the user may not be aware the server is unavailable and could want to know...?

This is a case that needs to be handled with a toast, but what i mean if there is no connection to the internet from the device, eg. no wifi and no mobile data. In those cases, it is clear that the device needs to get a local version, and i think we dont need to show a message. When internet is available (or a connection to a local network WITHOUT access) the normal behaviour (showing a toast) is appropriate. 


Regarding the screenshots: I think it's normal, and you dont have to do anything as long as you didn't cause it :D and from what you touched, you didnt cause it ;)

t-m-w:
Come to think of it, the Nextcloud app itself pops up the sync failure message (in a different, in-app style) and waits for a fixed, unskippable amount of time before proceeding to open the file -- far more annoying, to be honest. Perhaps that should be addressed separately from this patch, unless you think both should be adjusted at once. I am not sure it makes sense for the documents provider to take a different approach from the app.

In any case...

> When internet is available (or a connection to a local network WITHOUT access) the normal behaviour (showing a toast) is appropriate.

Do you have any suggestions for how to detect this? It looks as though the simple [ConnectivityManager function calls for checking network state are deprecated](https://developer.android.com/reference/android/net/NetworkInfo#isConnected()) and Android now wants developers to use callbacks instead. Does Nextcloud already track this somewhere else? A search for the term NetworkCallback found nothing in this repo or android-library. Maybe if the DownloadFileOperation returned something more useful we could just use that. It may have access to more information simply by being the functionality that attempts the download in the first place.

AlvaroBrey (approved):

AndyScherzinger (approved):

Issue #8585, opened 2021-06-15
Crashes related to unregistered Fido USB receiver


Valdnet:
### Steps to reproduce

1. Enter the application settings.
2. Change the theme.

```
********** CAUSE OF ERROR **********

java.lang.RuntimeException: Unable to pause activity {com.nextcloud.android.beta/com.owncloud.android.ui.activity.FileDisplayActivity}: java.lang.IllegalArgumentException: Receiver not registered: de.cotech.hw.internal.transport.usb.UsbConnectionDispatcher$1@a17fb6c

	at android.app.ActivityThread.performPauseActivityIfNeeded(ActivityThread.java:5127)

	at android.app.ActivityThread.performPauseActivity(ActivityThread.java:5076)

	at android.app.ActivityThread.handlePauseActivity(ActivityThread.java:5016)

	at android.app.servertransaction.TransactionExecutor.performLifecycleSequence(TransactionExecutor.java:242)

	at android.app.servertransaction.TransactionExecutor.cycleToPath(TransactionExecutor.java:215)

	at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:187)

	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:105)

	at android.app.ClientTransactionHandler.executeTransaction(ClientTransactionHandler.java:57)

	at android.app.ActivityThread.handleRelaunchActivityLocally(ActivityThread.java:5889)

	at android.app.ActivityThread.access$4500(ActivityThread.java:296)

	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2397)

	at android.os.Handler.dispatchMessage(Handler.java:107)

	at android.os.Looper.loop(Looper.java:213)

	at android.app.ActivityThread.main(ActivityThread.java:8178)

	at java.lang.reflect.Method.invoke(Native Method)

	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:513)

	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1101)

Caused by: java.lang.IllegalArgumentException: Receiver not registered: de.cotech.hw.internal.transport.usb.UsbConnectionDispatcher$1@a17fb6c

	at android.app.LoadedApk.forgetReceiverDispatcher(LoadedApk.java:1507)

	at android.app.ContextImpl.unregisterReceiver(ContextImpl.java:1647)

[...]

PR #8933, opened 2021-09-02
Revert "Bump hwsecurity-fido from 4.1.0 to 4.4.0"

AlvaroBrey:
This reverts commit 0d7d40e1d3afae4bf1822a265667f85b2e182137.

This is a fix for crashes reported in #8585 and several related issues, all caused by an unregistered receiver when trying to pause Activities.

@dschuermann If you can, please look at why this might be happening,

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -69,7 +69,7 @@ ext {
     byteBuddyVersion = "1.11.15"
     espressoVersion = "3.4.0"
     workRuntime = "2.5.0"
-    fidoVersion = "4.4.0"
+    fidoVersion = "4.1.0"
 
     ciBuild = System.getenv("CI") == "true"
 


Valdnet (approved):

build.gradle
```diff
@@ -69,7 +69,6 @@ ext {
     byteBuddyVersion = "1.11.15"
     espressoVersion = "3.4.0"
     workRuntime = "2.5.0"
-    fidoVersion = "4.4.0"
```
tobiasKaminsky, line 72: Having this version pattern is quite good to make sure that the libs are on the same version.
Otherwise we would get two different version bumps by depend a bot, and if both libs are relying on each other, they might then fail.
So better just switch it here :)

tobiasKaminsky (changes requested):

build.gradle
AlvaroBrey: Sure, I just reverted the original commit. Will change it back to the previous pattern and just bump down the number.

tobiasKaminsky (approved):

PR #8936, opened 2021-09-03
Minor script improvements

AlvaroBrey:
This should fix github comments after screenshot tests. Also some minor improvements for the androidScreenshotTest script.


scripts/uploadReport.sh
```diff
@@ -1,15 +1,21 @@
 #!/usr/bin/env bash
 
+curl_gh() {
```
tobiasKaminsky, line 3: That's a nice way 👍

AlvaroBrey: Seems like it didn't work though. Stand by, I'll check what's happening

tobiasKaminsky:
Still does not work, as screenShotTest.yml passes GITHUB_TOKEN as $7, but token in uploadReport.sh is 8.
I think $7 GIT_USERNAME is not used anymore, so this can be deleted, making $7 GIT_TOKEN.
When you change this, please check where uploadReport.sh is also used and adapt it there.

AlvaroBrey:
> Still does not work, as screenShotTest.yml passes GITHUB_TOKEN as $7, but token in uploadReport.sh is 8.
> I think $7 GIT_USERNAME is not used anymore, so this can be deleted, making $7 GIT_TOKEN.
> When you change this, please check where uploadReport.sh is also used and adapt it there.

Seems to be working now!

.github/workflows/screenShotTest.yml
```diff
@@ -23,8 +23,6 @@ jobs:
                     mkdir -p $HOME/.gradle
                     echo "org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryError" > $HOME/.gradle/gradle.properties
                     ./gradlew assembleGplayDebug
-            -   name: Delete old comments
```
tobiasKaminsky, line 26: Why do you remove this?
Idea behind this is then if you have a failing screenshot, GitHub posts a comment, then you fix/restart it, tests passes and removes the old comment, so that we always only see valid failing comments

AlvaroBrey: See https://github.com/nextcloud/android/pull/8936/commits/131373f066e077622a5c806ba6d10f56a45fcfd0 and https://github.com/nextcloud/android/blob/master/scripts/uploadReport.sh#L3-L14.

`uploadReport.sh` is already doing this deletion, so we don't need it in the `.yml`.

Additionally, I think `deleteOldComments.sh` was broken anyway, since it's [declaring `GITHUB_USER`](https://github.com/nextcloud/android/blob/master/scripts/deleteOldComments.sh#L13) but [using `GITHUB_TOKEN` instead](https://github.com/nextcloud/android/blob/master/scripts/deleteOldComments.sh#L19).

If you think we still need the separate script/step, I can restore it and fix it, no problem :)

tobiasKaminsky: Oh boy…I wrote this mess and now I am the one who does not understand it.
Thanks for pointing this out to me!

PR #8934, opened 2021-09-03
Replace Facebook lib to get screenshots again

tobiasKaminsky:
Previously screenshot test ran fine (no error), but also no screenshot was taken.

Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

### Testing 
Writing tests is very important. Please try to write some tests for your PR. 
If you need help, please do not hesitate to ask in this PR for help.

[unit tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#unit-tests)
[instrumented tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#instrumented-tests)
[UI tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#ui-tests)

- [x] Tests written, or not not needed

PR #8944, opened 2021-09-06
dependabot: Pin facebook screenshot version

AlvaroBrey:
As outlined in https://github.com/nextcloud/android/pull/8934, upgrading this dependency
breaks screenshot tests (0 tests are detected by Shot), so let's pin it back for now

The actual "final" fix should be having Shot take the screenshots directly, rather than
manually taking them with the facebook lib

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -89,3 +89,6 @@ updates:
                     versions:
                         - ">= 1.2.17.a"
                         - "< 1.2.18"
+                -   dependency-name: com.facebook.testing.screenshot:core
+                    versions:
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Can you check if this works for you? https://github.com/nextcloud/android/pull/8934
Then we hopefully do not need this.
If so, please close it :-)

AlvaroBrey:
> Can you check if this works for you? #8934
> Then we hopefully do not need this.
> If so, please close it :-)

Sure, this was just to prevent Dependabot from opening PRs for new versions

PR #8961, opened 2021-09-09
scripts: uploadReport: Fix deleteOldComments function

AlvaroBrey:
It didn't work because the full output was redirected to /dev/null, so old comments
were never found

### DIFF
--- a/scripts/uploadReport.sh
+++ b/scripts/uploadReport.sh
@@ -9,7 +9,7 @@ curl_gh() {
 deleteOldComments() {
     # delete all old comments, matching this type
     echo "Deleting old comments for $BRANCH_TYPE"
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@AlvaroBrey I would expect to see the lint/spotbugs report (even if the build turned green) but it is missing, so I'd guess something is still wrong 😿

AlvaroBrey:
> @AlvaroBrey I would expect to see the lint/spotbugs report (even if the build turned green) but it is missing, so I'd guess something is still wrong crying_cat_face

You're right, it seems the analysis scripts need some updates to work with the new github auth schema. However this PR is just about fixing comment deletion for screenshot tests so we don't end up with too many comments

AndyScherzinger:
@AlvaroBrey while https://github.com/nextcloud/android/pull/8965 opened 3 hours ago works just fine 👀

AndyScherzinger (approved):

AndyScherzinger:
But yeah, feel free to merge @AlvaroBrey 👍

AlvaroBrey:
No, you're right, uploadReport is probably wrongly deleting the messages somewhere. I'll come back to fix these scripts in time, but for now I'll just close this and roll back some changes in the snapshot scripts so deletion works.

PR #9044, opened 2021-09-30
Target SDK 30

### LINKED ISSUE #8913: Bump to SDK 30
- bump to target sdk 30
- check scoped storage

AlvaroBrey:
Closes #8913 


AlvaroBrey:
@tobiasKaminsky Several integration tests fail on SDK 30 with this exception:

```
java.io.IOException: Operation not permitted
	at java.io.UnixFileSystem.createFileExclusively0(Native Method)
	at java.io.UnixFileSystem.createFileExclusively(UnixFileSystem.java:317)
	at java.io.File.createNewFile(File.java:1008)
	at com.owncloud.android.AbstractIT.createFile(AbstractIT.java:251)
	at com.owncloud.android.AbstractIT.getDummyFile(AbstractIT.java:240)
```
The file that the test attempts to create is `File file = new File(FileStorageUtils.getTemporalPath(account.name) + File.separator + name);`, so it looks like we don't have permission to write in the temporal path. The real path ends up being
`/storage/emulated/0/Android/media/com.nextcloud.client/nextcloud/tmp/test@https%3A%2F%2Fnextcloud.localhost/.dotfile.ext`.

I haven't been able to reproduce this crash via smoke testing manually.

tobiasKaminsky:
> `/storage/emulated/0/Android/media/com.nextcloud.client/nextcloud/tmp/test@https%3A%2F%2Fnextcloud.localhost/.dotfile.ext`.

I think this is "just" a test problem, as this tries to use public tmp dir.

FileStorageUtils:getTemporalPath should be removed entirely, and we should instead use getInternalTemporalPath, I think.
As this is always read/writable by the app, as far is I understand it.

AlvaroBrey:
@tobiasKaminsky this is ready for review now. Tests are passing on SDK 30 and I haven't seen any  major crashes while smoke testing.

I suggest to take a look at the list of commits so that the changeset is a bit more understandable. In the end a lot of seemingly unrelated changes were needed to deal with scoped storage and other SDK 30 changes.

src/main/java/com/owncloud/android/utils/PermissionUtil.kt
```diff
@@ -0,0 +1,95 @@
+package com.owncloud.android.utils
```
AndyScherzinger, line 1: missing license header

AndyScherzinger (approved):
Looks good code-wise, just a single missing license header 👍

src/main/java/com/owncloud/android/utils/PermissionUtil.kt
AlvaroBrey: Fixed in https://github.com/nextcloud/android/pull/9044/commits/5e0073391210983d058544bea7830357f2ca718e

PR #9072, opened 2021-10-07
Deprecate FileDataStorageManager constructors depending on legacy Account

ezaquarii:
This commit deprecates constructors using legacy platform `Account` type
and wraps them using new API accepting `User` model.

All calls to `FileDataStorageManager` constructors that were trivially
portable without are updated as well. Places where passing new type
requires more risky work were left out for now.

`FileDataStorageManager` still uses `Account` internally - modifying this
requires getting rid of all legacy constructors first.


ezaquarii:
@tobiasKaminsky @AlvaroBrey Tests require a bit more work in fixtures and few places are left for now until I have time to tackle them one by one (those are more risky). Converted calls were using `User.toPlatformAccount()` so there should be no risk of regressions here.

tobiasKaminsky:
Restarted CI.

PR #9114, opened 2021-10-20
Clean up notification channels

strugee:
The commits mostly describe the work here, but the general idea is to make the notification channels more understandable. "Notification channel", for example, is both redundant (the user is already looking at the notification channels screen) and unnecessarily technical ("notification channels" is a developer term; the UI refers to this as "notification categories").

There is almost certainly something here that gets l10n wrong - it's unclear to me in the face of https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#translations whether strings should be deleted when they're no longer needed. In this case though it seems fine since it seems like this was never actually used?

### Testing 

It's unclear to me whether this kind of change should be tested? If it should be there will probably already be tests and CI should tell me I suppose. I _have_ manually tested this.

### DIFF
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -500,9 +500,7 @@ public class MainApp extends MultiDexApplication implements HasAndroidInjector {
                               R.string.notification_channel_file_sync_name,
                               R.string.notification_channel_file_sync_description, context);
 
-                createChannel(notificationManager, NotificationUtils.NOTIFICATION_CHANNEL_FILE_OBSERVER,
-                              R.string.notification_channel_file_observer_name, R.string
-                                  .notification_channel_file_observer_description, context);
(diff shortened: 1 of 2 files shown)


strugee:
Both of the new lint warnings are unused resources - this is due to the policy of not changing l10n strings and instead creating new values. So these warnings are spurious.

strugee:
I glanced through the screenshot test - looks like all the failures are unrelated to this PR? Is the test suite broken?

AlvaroBrey:
> I glanced through the screenshot test - looks like all the failures are unrelated to this PR? Is the test suite broken?

Yes, screenshot tests are currently broken, don't worry about them

tobiasKaminsky:
@strugee is there a way to re-order channels:
image

would be totally great to have download and upload next to each other.

tobiasKaminsky:
> Both of the new lint warnings are unused resources - this is due to the policy of not changing l10n strings and instead creating new values. So these warnings are spurious.

As those old ones are not used, you can remove them.
New ones are just a good idea so that not suddenly the meaning/context changes.

PR #9117, opened 2021-10-20
WIP: try to run integration tests in github actions

AlvaroBrey:
Current situation:
- We need to bring up a Nextcloud server instance to test again
- And we also need to use an Android emulator

The big problem with this:
- Github Actions `macos` machines have hardware virtualization (which we need for emulator) but no Docker
- On the other hand, the `ubuntu` machines have Docker, but no hardware virtualization

The initial tests to install docker on the `macos` machines have been not really that successful. Installing docker takes a _long_ time, and then there's no clear way to connect the emulator to the dockerized server.

The only reasonable alternative might be to bring up the Nextcloud server instance in a way that doesn't require Docker.


AlvaroBrey:
Abandoned due to technical limitations on GH actions (described in the first post)

PR #9152, opened 2021-10-26
Categorize notification channels into groups

strugee:
Implements notification channels as discussed in https://github.com/nextcloud/android/pull/9114#issuecomment-948909856.

However, according to https://material.io/design/platform-guidance/android-notifications.html#settings (search for "Grouping channels") we aren't really supposed to do this because we don't have >10 notification channels. I wonder if maybe we should just merge Uploads/Downloads/File sync into one notification channel?

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -27,6 +27,7 @@ import android.app.Activity;
 import android.app.ActivityManager;
 import android.app.Application;
 import android.app.NotificationChannel;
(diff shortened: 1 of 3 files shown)


tobiasKaminsky:
> Yeah, I think grouping may be overkill here. Merging those into a single "File transfers" channel makes sense to me, what do you think @tobiasKaminsky ?

Merging is good.
Maybe we want to separate "normal" messages from error messages, so that an user can hide all upload/download messages, but will only see messages if transfer fails, as this is then something they need to react on.

PR #9186, opened 2021-11-01
Enable StrictMode during development

tobiasKaminsky:
Ref: https://developer.android.com/reference/android/os/StrictMode

Let us check this from time to time to see what we can improve, if there are too many false positives, etc.

### DIFF
--- a/src/main/java/com/owncloud/android/MainApp.java
+++ b/src/main/java/com/owncloud/android/MainApp.java
@@ -244,6 +244,8 @@ public class MainApp extends MultiDexApplication implements HasAndroidInjector {
     @SuppressFBWarnings("ST")
     @Override
     public void onCreate() {
+        enableStrictMode();
+
         setAppTheme(preferences.getDarkThemeMode());
         super.onCreate();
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -244,6 +244,20 @@ protected void attachBaseContext(Context base) {
...
+                                           .detectDiskReads()
+                                           .detectDiskWrites()
+                                           .detectAll()
+                                           .penaltyLog()
+                                           .build());
+            StrictMode.setVmPolicy(new StrictMode.VmPolicy.Builder()
+                                       .detectLeakedSqlLiteObjects()
+                                       .detectLeakedClosableObjects()
+                                       .penaltyLog()
+                                       .build());
```
AlvaroBrey, line 258: Can you extract this into a separate method (`enableStrictMode()` for example)? `onCreate` is already too big I think. Otherwise looks good.

AlvaroBrey (approved):

AlvaroBrey (approved):

PR #9188, opened 2021-11-01
Use LeakCanary in debug mode

tobiasKaminsky:
See: https://square.github.io/leakcanary/fundamentals/

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -338,6 +338,8 @@ dependencies {
         }
     }
 
+    debugImplementation 'com.squareup.leakcanary:leakcanary-android:2.7'
+
     // dependencies for local unit tests
     testImplementation 'junit:junit:4.13.2'
     testImplementation "org.mockito:mockito-core:$mockitoVersion"
--- a/build.gradle
+++ b/build.gradle
@@ -338,7 +338,9 @@ dependencies {
         }
     }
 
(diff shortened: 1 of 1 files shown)


AlvaroBrey (approved):

AlvaroBrey:
Nice, but tests are failing for some reason. I seem to remember that LeakCanary freezes the app to take memory snapshots, so that may mess with our CI tests. Let's see if they keep failing after the rebase.

AlvaroBrey:
Apparently LeakCanary is disabled by default  only in tests that use androidx `InstrumentationTestRunner`: https://github.com/square/leakcanary/pull/1436/files
For all other tests, we could probably disable it manually in `@Before` with something like this: https://github.com/square/leakcanary/pull/1436/files#diff-5021ef7717778dad8ab1d9befeeb28d40c2435286b6d9e9f22db2f1542c363bbR98

tobiasKaminsky:
Can we have this "implementation" based on whether we are on CI or not? 
We already have ciBuild in build.grade

AlvaroBrey:
> Can we have this "implementation" based on whether we are on CI or not? We already have ciBuild in build.grade

That seems to have worked. @tobiasKaminsky feel free to merge if it looks right

AlvaroBrey (approved):

PR #9195, opened 2021-11-03
Add What's New view to warn users of changes in storage after sdk 30

AlvaroBrey:
<details>
<summary>Click for screenshot</summary>

Screenshot_1635954417

</details>


tobiasKaminsky:
Nice.
Maybe we want to add a second page, which 
- explicitly suggests user to check their upload folder
- informs that also other apps cannot directly see/use synced files from NC app

src/main/java/com/nextcloud/client/onboarding/WhatsNewActivity.java
```diff
@@ -113,7 +113,7 @@ protected void onCreate(Bundle savedInstanceState) {
         if (showWebView) {
             tv.setText(R.string.app_name);
         } else {
-            tv.setText(String.format(getString(R.string.whats_new_title), appInfo.getFormattedVersionCode()));
+            tv.setText(String.format(getString(R.string.whats_new_title), BuildConfig.VERSION_NAME));
```
tobiasKaminsky, line 116: I think you can revert this, as appInfo provides the same info, but has the advantage that we could easily mock this for screenshot tests.

AlvaroBrey: http://github.com/nextcloud/android/blob/be43a629fda8ce37b2592410cfe0dbffa2fc6763/src/main/java/com/nextcloud/client/appinfo/AppInfoImpl.java#L32-L32
The old version just shows the version code as a string, so `30190000` instead of `3.19.0 Alpha 1`. I will instead add a `getVersionName()` to `AppInfo`.

PR #9209, opened 2021-11-05
build.gradle: Refactor spotbugs setup

AlvaroBrey:
- Don't duplicate tasks. Use the ones that spotbugs provides and link them to their sourcesets
 - Add dependencies to compileSources to the spotbugs tasks
 - Update spotbugs plugin

This results in a much higher amount of issues. They were probably not being properly scanned earlier.


AndyScherzinger:
The report doesn't show the scanned classes anymore, can we get that back, at least to see what is now scanned additionally?

AlvaroBrey:
> The report doesn't show the scanned classes anymore, can we get that back, at least to see what is now scanned additionally?

Done in https://github.com/nextcloud/android/pull/9209/commits/9eb2fcc6c1aae48528e24df2ffe098bddb57f33f

tobiasKaminsky:
> This results in a much higher amount of issues. They were probably not being properly scanned earlier.

Oh my dear… ;)
Can you update spotbugs results, so that this passes.
And then we can slowly again try to lower the count…

tobiasKaminsky (approved):

tobiasKaminsky (approved):

PR #9210, opened 2021-11-05
Compile and target SDK 31

AlvaroBrey:
Lint is already complaining!

No huge breaking changes in this version.


AlvaroBrey:
Let's take another stab at this later

PR #9228, opened 2021-11-08
Ditch scoped storage in favor of MANAGE_EXTERNAL_STORAGE

AlvaroBrey:
We _are_ a file-managing app, after all.

This permission matches our needs more than scoped storage does. 


AndyScherzinger:
yeah, according to the docs like you said we should be allowed to make use of it.

tobiasKaminsky (approved):

PR #9248, opened 2021-11-11
Unified search: fix createNextcloudClient called from main thread

### LINKED ISSUE #9244: [3.18.0 RC5] Trapped in an app unlock loop after executing a search
### Actual behavior
If you search for a specific file on your Nextcloud server and you've activated the device app pin unlock function, the Nextcloud app is instantly locked and if you try to unlock it, the dialog appears again and again. (This problem didn't happen in RC4)

### Expected behavior
If you search a specific file on your Nextcloud server, the search should be executed and the result should be displayed afterwards.

### Steps to reproduce
1. Open the Nextcloud app and unlock it with the device pin unlock function or your fingerprint etc.
2. Select the search bar on top of the app, enter the search term and execute the search.
-> The Nextcloud app is instantly been locked. If you unlock it, you're trapped in an unlock loop. The only way to open the app again is, to kill it first in Settings > Apps.

### Environment data
Android version: 11.0
Device model: Samsung Galaxy S20
Stock or customized system: Stock ROM
Nextcloud app version: 3.18.0 RC5
Nextcloud server version: 21.0.5

AlvaroBrey:
Fixes #9244

### DIFF
--- a/src/main/java/com/owncloud/android/ui/unifiedsearch/UnifiedSearchRemoteRepository.kt
+++ b/src/main/java/com/owncloud/android/ui/unifiedsearch/UnifiedSearchRemoteRepository.kt
@@ -24,8 +24,12 @@ package com.owncloud.android.ui.unifiedsearch
 import com.nextcloud.client.account.CurrentAccountProvider
 import com.nextcloud.client.core.AsyncRunner
 import com.nextcloud.client.network.ClientFactory
+import com.nextcloud.common.NextcloudClient
 import com.owncloud.android.lib.common.SearchProviders
 import com.owncloud.android.lib.common.utils.Log_OC
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 
 class UnifiedSearchRemoteRepository(
     private val clientFactory: ClientFactory,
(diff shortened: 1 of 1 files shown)


src/main/java/com/owncloud/android/ui/unifiedsearch/UnifiedSearchRemoteRepository.kt
```diff
@@ -35,6 +39,11 @@ class UnifiedSearchRemoteRepository(
 
     private var providers: SearchProviders? = null
 
+    private fun runAsyncWithNcClient(cb: (client: NextcloudClient) -> Unit) = CoroutineScope(Dispatchers.IO).launch {
+        val client = clientFactory.createNextcloudClient(currentAccountProvider.user)
+        cb(client)
```
tobiasKaminsky, line 44: Is there a better name for cb?

tobiasKaminsky (approved):

src/main/java/com/owncloud/android/ui/unifiedsearch/UnifiedSearchRemoteRepository.kt
AlvaroBrey: Sure, I'll just name it `callback`

AndyScherzinger (approved):

PR #9249, opened 2021-11-11
Bump compileSDK to 31

AlvaroBrey:
Required by many new library versions already


AndyScherzinger (approved):
👍 fine with me, sdk31 does contain a lot of changes like enforced notification layout (much smaller) so that would be one thing that needs to be checked after the bump for the targetSdk (I know, not bumped within this PR :)

Issue #7722, opened 2020-12-23
Save the rename of a file's


KwonEunKyung:
### Is your feature request related to a problem? Please describe.
I know that “dot-file” can be viewed by activating “Show hidden files” in the settings.

However, I think there is a problem that a file cannot be easily found when a user does not know this fact or accidentally saves a file name as space.

### Describe the solution you'd like
Therefore, I think the user’s intended “dot-file” name should be at least 1 character long except for dot+extension. That is, I think it would be better to prevent users from saving their files rename as spaces.

**NOTE:** Be super sure to remove sensitive data like passwords, note that everybody can look here! You can use the Issue Template application to prefill some of the required information: https://apps.nextcloud.com/apps/issuetemplate

tobiasKaminsky:
Can you give us an example, I fear that I fail to understand what should be changed.

PR #9287, opened 2021-11-17
File rename input warning message

liam-p-23:
Implementation of a feature to address the concern raised in #7722

### Testing 

### DIFF
--- a/src/main/java/com/owncloud/android/ui/dialog/RenameFileDialogFragment.java
+++ b/src/main/java/com/owncloud/android/ui/dialog/RenameFileDialogFragment.java
@@ -109,8 +109,9 @@ public class RenameFileDialogFragment
         ThemeTextInputUtils.colorTextInput(binding.userInputContainer,
                                            binding.userInput,
                                            ThemeColorUtils.primaryColor(getActivity()));
-
-        binding.userInput.setSelection(0, getSelectionEndIndex(currentName));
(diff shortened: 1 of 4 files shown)


AlvaroBrey (changes requested):
This is a nice idea but it would be better to use the existing `TextInputLayout`'s  `setError` method for this, instead of creating a new element in the dialog's layout. That way the UI would be more consistent and we don't have to deal with margins, paddings etc.

Of course this is a warning and not an error, so [a custom error `TextAppearance`](https://stackoverflow.com/questions/33709066/how-to-set-textinputlayout-error-message-colour)  should be used so that the text isn't red. This can be either the normal text color or a new color to signify a warning (material yellow?)

AlvaroBrey:
@liam-p-23 mostly looks good now, but please rebase or merge master into your branch to fix CI issues (a bug was introduced in master and fixed earlier today).

Additionally, check https://github.com/nextcloud/android/pull/9287/checks?check_run_id=4255707196 in order to comply with DCO

AlvaroBrey (changes requested):
Almost good to go, this looks great!  I've tested it locally and using `@color/text_color` looks a bit out of place. Please use `@color/primary` instead. Sorry for the confusion.

AlvaroBrey (approved):

AlvaroBrey:
Thank you @liam-p-23 !

PR #9333, opened 2021-11-20
Migate OCUpload to User

ezaquarii:

### DIFF
--- a/src/main/java/com/owncloud/android/files/services/FileUploader.java
+++ b/src/main/java/com/owncloud/android/files/services/FileUploader.java
@@ -468,7 +468,7 @@ public class FileUploader extends Service
             return;
         }
 
-        OCUpload ocUpload = new OCUpload(file, user.toPlatformAccount());
+        OCUpload ocUpload = new OCUpload(file, user);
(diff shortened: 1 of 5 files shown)


src/main/java/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -1070,61 +1070,42 @@ public static void retryUpload(@NonNull Context context, @NonNull Account accoun
...
-     *                     uploads due to any result will be retried.
      */
     public static void retryFailedUploads(
         @NonNull final Context context,
-        @Nullable final Account account,
         @NonNull final UploadsStorageManager uploadsStorageManager,
         @NonNull final ConnectivityService connectivityService,
         @NonNull final UserAccountManager accountManager,
-        @NonNull final PowerManagementService powerManagementService,
-        @Nullable final UploadResult uploadResult
```
ezaquarii, line 1087: All removed arguments were `null` anyway. As a result, 2 levels of `if`ology could be removed.

AlvaroBrey (approved):

Issue #9373, opened 2021-11-26
Empty placeholder flashes while loading


AlvaroBrey:
### Steps to reproduce
1. Open "Shared" or "Favorites" screen
2. Pull to refresh

(also happens during initial load) More screens may be affected.

### Expected behaviour
- Loading message appears

### Actual behaviour
- "No files here" appears for a short moment, before loading message does

### Environment data
Android version: 11 (sdk 30)

Device model: emulator

Stock or customized system: stock

Nextcloud app version: 3.18.0

Nextcloud server version: 22

Reverse proxy: no

PR #9432, opened 2021-12-02
if no shared files are on server, show correct empty message

tobiasKaminsky:
TODO:
~- [ ] upon first showing shared files, for a short "empty list is" shown, but instead "loading…" should be shown~ Fixed by #9972 

Actions Performed
1. Open the  app
2. Open the hamburger menu
3. Tap on Shared

Expected Result
The files are displayed immediately or a file load icon/message is displayed.

Actual Result
The message that there are no files here is displayed for a few seconds, then the files are displayed.


AlvaroBrey (changes requested):
This shows the correct empty message when there are no files, but does not fix #9373: the empty files placeholder is still shown while loading, instead of the `Loading...` message.

AlvaroBrey (approved):

PR #9435, opened 2021-12-02
OcFileListFragment: throttle overflow menu clicks

### LINKED ISSUE #9319: File menus overlapping because multiple can be opened at the same time
### Steps to reproduce
1. Tap on two files' More icons (three dots)

### Expected behaviour
Only the last one's menu should be opened. Even if the clicks are at the same time, the events should be separate.

### Actual behaviour
Both menus open and overlap each other

### Environment data
Android version: 11

Device model:  Xiaomi Mi 9T Pro

Stock or customized system: LineageOS-based

Nextcloud app version: 3.18.0

Nextcloud server version: 22

Reverse proxy: no

AlvaroBrey:
Fixes #9319

### DIFF
--- a/src/main/java/com/owncloud/android/utils/Throttler.kt
+++ b/src/main/java/com/owncloud/android/utils/Throttler.kt
@@ -20,10 +20,8 @@
  */
 package com.owncloud.android.utils
 
-import android.util.ArrayMap
 import com.nextcloud.client.core.Clock
 
-// TODO write tests
 /**
  * Simple throttler that just discards new calls until interval has passed.
  *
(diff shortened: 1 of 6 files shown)


src/main/java/com/owncloud/android/utils/Throttler.kt
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.owncloud.android.utils
+
+import android.util.ArrayMap
+import com.nextcloud.client.core.Clock
+
+// TODO write tests
```
tobiasKaminsky, line 26: Please remove, write some or note it down as a task for later :)

tobiasKaminsky (approved):

src/main/java/com/owncloud/android/utils/Throttler.kt
AlvaroBrey: https://github.com/nextcloud/android/pull/9435/commits/e138bb7175e15b71a8340c297e9c9070fcc28019

PR #9478, opened 2021-12-07
Empty list is not centered and in landscape cut off

tobiasKaminsky:
Therefore as a quick win, moved all up, so that user does not think it should be centered.
This also then works in landscape.

Real solution is to compute height of toolbar and then programmatically (depending on orientation) center it.


AlvaroBrey (approved):
Nice solution!

PR #9480, opened 2021-12-07
On video player hide next button, as it is always disabled

tobiasKaminsky:


AlvaroBrey (approved):
Looks a bit weird without it, but if it doesn't do anything then it makes sense to remove it.

PR #9482, opened 2021-12-08
Fullscreen toggle

tobiasKaminsky:
Now we have a real toggle fullscreen button in preview and fullscreen.

TODO
- [ ] check different sizes (as 143 is hardcoded)

 


AlvaroBrey (approved):
Works also on tablet

PR #9528, opened 2021-12-10
Huawei flavor

tobiasKaminsky:

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/CommunityActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/CommunityActivity.java
@@ -39,7 +39,7 @@ import com.owncloud.android.utils.theme.ThemeColorUtils;
  */
 public class CommunityActivity extends DrawerActivity {
 
-    private CommunityLayoutBinding binding;
+    protected CommunityLayoutBinding binding;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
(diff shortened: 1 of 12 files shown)


src/main/java/com/owncloud/android/ui/activity/CommunityActivity.java
```diff
@@ -109,6 +111,10 @@ private void setupContent() {
 
         binding.communityBetaApk.setOnClickListener(
             l -> DisplayUtils.startLinkIntent(this, R.string.beta_apk_link));
+
+        if (BuildConfig.FLAVOR.equals("huawei")) {
```
AlvaroBrey, line 115: I think it would be better to define "huawei" as a constant somewhere.

src/main/java/com/owncloud/android/ui/activity/NotificationsActivity.java
```diff
@@ -132,6 +133,10 @@ protected void onCreate(Bundle savedInstanceState) {
     }
 
     private void setupPushWarning() {
+        if (BuildConfig.FLAVOR.equals("huawei")) {
```
AlvaroBrey, line 136: I think it would be better to define "huawei" as a constant somewhere.

AlvaroBrey (approved):
Looks good!

src/main/java/com/owncloud/android/ui/activity/NotificationsActivity.java
tobiasKaminsky: Right. Do you have an idea where? 
setup.xml/strings.xml?
As a constant in some helper class?

I think this fits nowhere…🤷

tobiasKaminsky:
I hope this is now a better approach, what do you think?

AlvaroBrey (approved):
All good now!

PR #9535, opened 2021-12-12
When stopping PlayerService, do not use foreground service api

### LINKED ISSUE #9491: PlayerService crash (again)

android.app.RemoteServiceException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{1953b9 u0 com.nextcloud.client/.media.PlayerService}
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2081)
	at android.os.Looper.loop(Looper.java:236)
	at android.app.ActivityThread.main(ActivityThread.java:7864)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:620)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1011)

ID: com.nextcloud.client
Version: 30180090
Build flavor: gplay

Brand: Redmi
Device: galahad
Model: M2004J19C
Id: RP1A.200720.011
Product: galahad_eea

SDK: 30
Release: 11
Incremental: V12.5.3.0.RJCEUXM

ezaquarii:
Sending async message to the service using foreground API
requires setting foreground flag. This won't happen if
the service is being stopped, so we must use startService()
API instead.

This API will throw IllegalStateException when app is banned
from running a service due to power management constraints.

Since stopping a service under such condition is not needed
anyway (it is not running), we can safely catch and ignore it.

This happens when app is woken up after long inactivity and
UI is recreated from cache.

Fixes #9491

### DIFF
--- a/src/main/java/com/nextcloud/client/media/PlayerService.kt
+++ b/src/main/java/com/nextcloud/client/media/PlayerService.kt
@@ -152,7 +152,6 @@ class PlayerService : Service() {
 
     private fun onActionStopFile(args: Bundle?) {
         val file: OCFile = args?.getParcelable(EXTRA_FILE) ?: throw IllegalArgumentException("Missing file argument")
-
         stopServiceAndRemoveNotification(file)
     }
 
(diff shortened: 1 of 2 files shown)


AlvaroBrey (approved):

AlvaroBrey:
Thanks @ezaquarii , seems like a reasonable workaround!

PR #9551, opened 2021-12-15
fixing some warnings

AbdulrhmanSayedAli:


AlvaroBrey (changes requested):
* Undo the changes in all files unrelated to what the PR intends (restore .idea files, gradle wrapper
* Explain what the intention is

PR #9582, opened 2021-12-17
Player: don't try to pause a non-running media player

AlvaroBrey:
Fixes an issue triggered by pausing, and then minimizing and coming back to the app.

Right now state is still lost (file returns to the beginning) but as a stopgap it's OK.
In the future, the state machine and how it interacts with the fragment should be reworked.

### DIFF
--- a/src/main/java/com/nextcloud/client/media/Player.kt
+++ b/src/main/java/com/nextcloud/client/media/Player.kt
@@ -152,8 +152,10 @@ internal class Player(
 
         override fun onPausePlayback() {
             trace("onPausePlayback()")
-            mediaPlayer?.pause()
-            listener?.onPause()
+            if (mediaPlayer?.isPlaying == true) {
+                mediaPlayer?.pause()
+                listener?.onPause()
+            }
         }
 
         override fun onRequestFocus() {


ezaquarii:
The state machine should prevent pause when stopped. What happened?

AlvaroBrey:
> The state machine should prevent pause when stopped. What happened?

Yep, when examining the state machine I thought so too, but it doesn't and I can't figure out why. Steps to reproduce:

- Play an audio file.
- Pause with the pause button. (state -> `PAUSED`)
- Leave the app with recents or home button. (state -> `STOPPED`)
- Return to the app. The `onPause` method is called again (and the player is left in an unusable state)

As I mentioned, this is more of a stopgap to prevent this edge case bug. If you can figure out why the state machine is not working properly I'll be glad to take that patch instead.

ezaquarii:
Thanks for the repro steps. I'll take a look at it so please don't merge it yet. 

As a design rule, this should be handled by the state machine specifically to avoid messing with if-ology around other, random parts of the code.

ezaquarii:
@AlvaroBrey 

I think I know what's happening.

PlayerStateMachine

ezaquarii (approved):
Looks like it does what it should do.

ezaquarii:
@AlvaroBrey Small note about the commit message.

I think that both the state machine and service are capable of pausing/resuming the playback when UI is hidden (service is started in foreground), but the UI requests full stop when minimized. When it is shown again, it requests new playback - including file download.

I see 2 avenues:
1. inject start position when resuming activity (marginal improvement)
2. pause instead of stopping

The latter one could be hindered by the fragment/activity lifecycle mess. Player generally needs new UI.

AlvaroBrey:
Thank you for the feedback and insights @ezaquarii . I will take a look at options to not lose track of the player state.

PR #9590, opened 2021-12-20
DocumentsStorageProviderIT: add timeout to testServerChangedFileContent()

AlvaroBrey:
This test sometimes hangs forever, resulting in (for example) CI builds being killed.
This way, only this test fails and the rest of the suite continues.


tobiasKaminsky:
Wow. Cool, I did not know that this is possible.

tobiasKaminsky (approved):

PR #9677, opened 2022-01-10
Fix sending multiple files

AlvaroBrey:
Right now, any attempt to send multiple files from context menu results in an empty BottomSheetFragment. This fixes that.


tobiasKaminsky:
Can you add some tests? As this is a core functionality we should make sure that it does not break (again) in future.

PR #9685, opened 2022-01-12
Test library version for ipv4 fallback

AlvaroBrey:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -64,7 +64,7 @@ ext {
 	daggerVersion = "2.40.5"
     markwonVersion =  "4.6.2"
     prismVersion = "2.0.0"
-    androidLibraryVersion = "b1e4bdc5faee33ab2307e8989815e0e39a6de5b5"
+    androidLibraryVersion = "082905cb3410d4e1221bf6f3ed5f37a341ea029c"
     mockitoVersion = "4.2.0"
     mockkVersion = "1.12.2"
     powermockVersion = "2.0.9"
--- a/build.gradle
+++ b/build.gradle
@@ -64,7 +64,7 @@ ext {
 	daggerVersion = "2.40.5"
     markwonVersion =  "4.6.2"
     prismVersion = "2.0.0"
-    androidLibraryVersion = "master-SNAPSHOT"
+    androidLibraryVersion = "495605efd2fd0aff9d057ddef84015ad194d1b0e"
     mockitoVersion = "4.2.0"
     mockkVersion = "1.12.2"
     powermockVersion = "2.0.9"
--- a/build.gradle
+++ b/build.gradle
@@ -64,7 +64,7 @@ ext {
 	daggerVersion = "2.40.5"
     markwonVersion =  "4.6.2"
     prismVersion = "2.0.0"
-    androidLibraryVersion = "81279a0928889b47ac170c570e4c44ed478d6897"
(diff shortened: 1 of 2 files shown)


AlvaroBrey:
> I just scanned the QR code with my smartphone and downloaded the APK. However, when I tried to open it with "packet installer" I just got "Error parsing the packet".
> 
> What can be the problem? I'm not familiar with manual installation of an APK. Am I missing something?
> 

APK works fine for me, problem must be on your side. Just click the downloaded APK from the Downloads folder and let the system install it, rather than using third party apps. If the problem persists try downloading it again.

mburnicki:
> > I just scanned the QR code with my smartphone and downloaded the APK. However, when I tried to open it with "packet installer" I just got "Error parsing the packet".
> > What can be the problem? I'm not familiar with manual installation of an APK. Am I missing something?
> 
> APK works fine for me, problem must be on your side.

Hmm.

> Just click the downloaded APK from the Downloads folder and let the system install it, rather than using third party apps.

That's exactly what I did. I'm not using 3rd party apps, the APK is opened with the installer that is provided by Android/Oxygen. My original language is German, and when I click on the downloaded APK, the proposed action is "Mit Paketinstallation öffnen". I don't know how this is called in a different language.

> If the problem persists try downloading it again.

Just did it and it worked. The second download has a slightly different size than the first one.

Anyway, when I start the -dev version and try to connect to the Nextcloud server on my local intranet by scanning the barcode provided by my web browser on my computer, it says: "Could not find the host" (translated from German).

If I run the host command from my Linux PC, I get:

$ host abcdefghij.burnicki.net
abcdefghij.burnicki.net is an alias for xyz.myfritz.net.
xyz.myfritz.net has address 77.22.aaa.bbb
xyz.myfritz.net has IPv6 address 2a02:a::b:...

Exactly the same if I run the command in a termux session on my smartphone.

AlvaroBrey:
@mburnicki please try the [newer APK](https://github.com/nextcloud/android/pull/9685#issuecomment-1011009774)

mburnicki:
> @mburnicki please try the [newer APK](https://github.com/nextcloud/android/pull/9685#issuecomment-1011009774)

@AlvaroBrey Have uninstalled the previous version, and downloaded and installed the new one. Same problem, unfortunately.

If you want, I can give you the real DNS name for my Nextcloud server, so you can try yourself if your modified app can reach it, and don't need to setup your own test installation first.

However, I would prefer to send the info directly to you, and not publicly share it here at github.

AlvaroBrey:
> > @mburnicki please try the [newer APK](https://github.com/nextcloud/android/pull/9685#issuecomment-1011009774)
> 
> @AlvaroBrey Have uninstalled the previous version, and downloaded and installed the new one. Same problem, unfortunately.
> 
> If you want, I can give you the real DNS name for my Nextcloud server, so you can try yourself if your modified app can reach it, and don't need to setup your own test installation first.
> 
> However, I would prefer to send the info directly to you, and not publicly share it here at github.
> 
> Thanks for your effort, by the way.

Thanks @mburnicki, please send the info to alvaro.brey@nextcloud.com, although I suspect it will also depend on my local setup.

mburnicki:
> APK file: https://www.kaminsky.me/nc-dev/android-artifacts/9685.apk qrcode To test this change/fix you can simply download above APK file and install and test it in parallel to your existing Nextcloud app.

Works for me!
Thanks a bunch!

AlvaroBrey:
Superseded by #9719

PR #9692, opened 2022-01-13
Screenshot tests: Improve Activities activity tests

AlvaroBrey:
Force-close snackbar to not have the timing break tests


src/androidTest/java/com/nextcloud/client/ActivitiesActivityIT.kt
```diff
@@ -51,9 +50,9 @@ class ActivitiesActivityIT : AbstractIT() {
         shortSleep()
         Espresso.onView(withId(R.id.drawer_layout)).perform(DrawerActions.open())
         sut.runOnUiThread {
-            sut.binding.emptyList.emptyListView.visibility = View.VISIBLE
-            sut.binding.list.visibility = View.INVISIBLE
+            sut.dismissSnackbar()
         }
+        shortSleep()
```
tobiasKaminsky, line 55: shortSleep might cause flaky tests, if emulator is too slow.
#internet recommends therefore waitForIdleSync().
Can you test if this also works?

AlvaroBrey: Added `waitForIdleSync()` after the `shortSleep` calls, but if I try to remove the `shortSleep` I get this error randomly:

```
        java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Class java.lang.Object.getClass()' on a null object reference
        at com.facebook.testing.screenshot.layouthierarchy.LayoutHierarchyDumper.dumpHierarchy(LayoutHierarchyDumper.java:114)

```

tobiasKaminsky: Strange, so for some reasons it does not wait long enough, maybe as it is on a different thread?

AlvaroBrey: No idea :\

tobiasKaminsky: Ok. Then let us keep it this way.

tobiasKaminsky (approved):

PR #8217, opened 2021-03-28
Add scan document feature

thelittlefireman:
Signed-off-by: thelittlefireman <thelittlefireman@users.noreply.github.com>

APK size about 36Mb
based on [https://github.com/meganz/AndroidDocumentScanner](https://github.com/meganz/AndroidDocumentScanner)

### Testing 
Writing tests is very important. Please try to write some tests for your PR. 
If you need help, please do not hesitate to ask in this PR for help.

[unit tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#unit-tests)
[instrumented tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#instrumented-tests)
[UI tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#ui-tests)

- [ ] Tests written, or not not needed

PR #9717, opened 2022-01-19
Scan doc

tobiasKaminsky:
Successor of https://github.com/nextcloud/android/pull/8217
Uses https://github.com/zynkware/Document-Scanning-Android-SDK


build.gradle
```diff
@@ -131,6 +131,10 @@ android {
         // adapt structure from Eclipse to Gradle/Android Studio expectations;
         // see http://tools.android.com/tech-docs/new-build-system/user-guide#TOC-Configuring-the-Structure
 
+        ndk {
+            abiFilters 'arm64-v8a', 'x86_64', 'armeabi-v7a', 'x86'
```
tobiasKaminsky, line 135: @AlvaroBrey 
https://en.wikipedia.org/wiki/Android_version_history#HWREQ

I therefore would go only with arm:

```suggestion
            abiFilters 'arm64-v8a',  'armeabi-v7a'
```

which is then:
28mb master
43mb without any modification
34mb with minification/shrink/proguard

As you have more experience in minification, etc. do you think it is worth it?

AlvaroBrey: I have two concerns:

a) Minification is generally a good idea but we need to do extensive testing to ensure we don't break anything when enabling it.

b) Removing architectures for this library is fine, but wouldn't this filter also apply to other NDK libraries? We also use `droidsonroids.gif` and `conscrypt`, which have NDK components too.

AlvaroBrey: By the way there's also a way to generate per-abi APKs: https://developer.android.com/studio/build/configure-apk-splits although it would require updating our fastlane setup and I don't know if FDroid and Huawei store support it.

tobiasKaminsky: You are right, removing it entirely kills login.
I try to find a way to remove it only for this lib.

tobiasKaminsky: Hm, found that x86 is used way often nowadays, so I think best is indeed to go with split builds and disable it for fdroid/huawei.
What do you think? 🙈

tobiasKaminsky: 28      master
35      gplay-arm64-v8a-release-30200000.apk
29      gplay-armeabi-v7a-release-30200000.apk
54      gplay-x86_64-release-30200000.apk
45      gplay-x86-release-30200000.apk

AlvaroBrey: > Hm, found that x86 is used way often nowadays, so I think best is indeed to go with split builds and disable it for fdroid/huawei. What do you think? see_no_evil

I think that's OK. We should also upload the single APK to Github (even if we upload the splits too) for releases, to make it easier for users. 

I also see that there's no clear way to only enable abi split for a particular flavor, but that's something we can probably script around with gradle.

My biggest concern is how this will affect the fastlane process.

P.S.: We probably still should enable minification

tobiasKaminsky: Then let us merge it.
Small follow up issue here: https://github.com/nextcloud/android/issues/9848

PR #9718, opened 2022-01-19
Fix minor issues with recently modified view

AlvaroBrey:
- Fix titlebar not being set correctly
- Fix loading not working on first open
- Fix loading not working when coming back from share


AlvaroBrey:
Still needs tests

tobiasKaminsky (approved):

PR #9740, opened 2022-01-24
Test for recent files

AlvaroBrey:
@tobiasKaminsky I reverted the merge and reopened this PR because the screenshot tests won't work, as we have no server on screenshot tests.


tobiasKaminsky:
Right…
Then I will split this up:
- IT with server, which verifies it not via screenshot test, but with assert
- screenshot test to check only UI (will check if this is needed at all, as we might already have other search screenshots)

PR #9761, opened 2022-01-26
Improve freezes in local files screens when folder contains many files

AlvaroBrey:
For this:
 - Load folders asynchronously with loading animation, to avoid ANR and give feedback to the user
 - Do some optimizations to file sorting

While I was here: convert FileSortOrder and children to Kotlin.


tobiasKaminsky (approved):

tobiasKaminsky:
Awesome work :tada:

PR #9770, opened 2022-01-27
i18n: Change to a capital letter

Valdnet:

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -875,7 +875,7 @@
     <string name="logs_status_not_filtered">Logs: %1$d kB, no filter</string>
     <string name="logs_menu_search">Search logs</string>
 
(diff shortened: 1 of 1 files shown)


AlvaroBrey:
Does this trigger some kind of updated notice in Transifex? So that other translators are aware and fix this too

Valdnet:
@AlvaroBrey Translators receive information about an untranslated text string. They will have to translate it again, but they can use the translation suggestion 😉.

AndyScherzinger:
Important part for dev is that changes in existing strings need to be backported to all `stable-*` branches, else the transifex automation won't work and translations don't get pushed to this repo anymore. (This is just an info, I already labeled it for the backports, of course the merged would still need to be done after the bot created the backports)

AlvaroBrey (approved):

AlvaroBrey:
Thanks @Valdnet !

PR #9791, opened 2022-02-01
FileActivity: fix crash on share operations when FileDetailFragment is second fragment

AlvaroBrey:
When a file is downloaded, FileDetailFragment is opened as SECOND_FRAGMENT instead of LIST_OF_FILES, which breaks this logic.

### DIFF
--- a/src/main/java/com/owncloud/android/ui/activity/FileActivity.java
+++ b/src/main/java/com/owncloud/android/ui/activity/FileActivity.java
@@ -39,7 +39,6 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.os.IBinder;
 import android.text.TextUtils;
-import android.widget.Toast;
 
 import com.google.android.material.snackbar.Snackbar;
 import com.nextcloud.client.account.User;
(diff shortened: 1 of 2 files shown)


tobiasKaminsky (approved):

tobiasKaminsky:
Our goal is to remove second fragment, and always use only one (list_of_files) which we then can rename.

Therefore real fix is to use in FileDisplayActivity:startDownloadForPreview() -> setLeftFragment() instead of setSecondFragment().

After this we can check where else this is used and remove the other occurrences 👍

AlvaroBrey:
> Our goal is to remove second fragment, and always use only one (list_of_files) which we then can rename.
> 
> Therefore real fix is to use in FileDisplayActivity:startDownloadForPreview() -> setLeftFragment() instead of setSecondFragment().
> 
> After this we can check where else this is used and remove the other occurrences +1

I initially tried to do this, but it is very entangled with FileDisplayActivity's logic and couldn't get it working after some time. If you prefer we do that, I can try again no problem

tobiasKaminsky:
Yes, please try it again. And if you have troubles with it, push it as a draft and I can have a look :-)

AlvaroBrey:
> Yes, please try it again. And if you have troubles with it, push it as a draft and I can have a look :-)

I've made the changes in https://github.com/nextcloud/android/pull/9791/commits/6e79cfefc6f1fca14125623605807df33393cb11 and didn't find any crashes or bugs while I was testing it. Please also test it as I am not 100% sure I hit all of the use cases for FileDetailsFragment.

Note: after doing this, secondFragment appears unused from my understanding.

tobiasKaminsky (approved):

PR #9798, opened 2022-02-03
[WIP] Move app source to app subproject for modern gradle standards

AlvaroBrey:
Android Studio has started to fail for some stuff with that old structure (for example connected tests don't work).
It's time to do this.

**NOTE: rebase or update sources in app before merging!** Otherwise we may lose changes


AlvaroBrey:
Superseded by #9960

PR #9806, opened 2022-02-07
Fallback pdf viewer

### LINKED ISSUE #9158: Internal PDF viewer
Add a PDF viewer, so that user do not have to leave app/install another one.
There is a PDF engine from Google.

Howto: https://proandroiddev.com/rendering-pdfs-on-android-the-easy-way-c05635b2c3a8

AlvaroBrey:
Closes #9158 

- [x] pdf preview fragment
- [x] tests
- [x] working for undownloaded files too
- [x] don't open if another PDF app is available
- [x] extra screen to zoom in on pages


szaimen:
> * don't open if another PDF app is available

Not sure about that one. Would be cool if it was possible to use the build-in viewer in that case, too :)

AlvaroBrey:
> > * don't open if another PDF app is available
> 
> Not sure about that one. Would be cool if it was possible to use the build-in viewer in that case, too :)

That may be possible at a later stage. This PDF viewer will not be amazing due to platform limitations (pages are rendered as a bitmap, so no text selection; maybe no way to zoom in either); it's intended to be a fallback in case the user doesn't have any PDF reader installed, but a third-party will be almost always a better experience.

szaimen:
> That may be possible at a later stage. This PDF viewer will not be amazing due to platform limitations (pages are rendered as a bitmap, so no text selection; maybe no way to zoom in either); it's intended to be a fallback in case the user doesn't have any PDF reader installed, but a third-party will be almost always a better experience.

okay, good to know, thanks!

src/androidTest/java/com/owncloud/android/ui/preview/PreviewBitmapScreenshotTest.kt
```diff
@@ -0,0 +1,64 @@
...
+package com.owncloud.android.ui.preview
+
+import android.content.Intent
+import androidx.test.espresso.intent.rule.IntentsTestRule
+import com.owncloud.android.AbstractIT
+import com.owncloud.android.utils.ScreenshotTest
+import org.junit.Rule
+import org.junit.Test
+
+class PreviewBitmapScreenshotTest : AbstractIT() {
```
tobiasKaminsky, line 37: I call all tests on Android emulator/device "IT", and unit tests "Test".

src/androidTest/java/com/owncloud/android/ui/preview/pdf/PreviewPdfFragmentScreenshotTest.kt
```diff
@@ -0,0 +1,74 @@
...
+import androidx.lifecycle.Lifecycle
+import androidx.test.espresso.intent.rule.IntentsTestRule
+import com.nextcloud.client.TestActivity
+import com.owncloud.android.AbstractIT
+import com.owncloud.android.datamodel.OCFile
+import com.owncloud.android.utils.ScreenshotTest
+import org.junit.Rule
+import org.junit.Test
+
+class PreviewPdfFragmentScreenshotTest : AbstractIT() {
```
tobiasKaminsky, line 39: Same here :)
```suggestion
class PreviewPdfFragmentScreenshotIT : AbstractIT() {
```

src/main/java/com/owncloud/android/ui/preview/pdf/PreviewPdfViewModel.kt
```diff
@@ -0,0 +1,93 @@
+/*
```
tobiasKaminsky, line 1: This is the wrong license, seems also to be the case for other new files.

src/main/java/com/owncloud/android/ui/preview/pdf/PreviewPdfViewModel.kt
```diff
@@ -0,0 +1,93 @@
...
+    fun onClickPage(page: Bitmap) {
+        val file = File.createTempFile("pdf_page", ".bmp", MainApp.getAppContext().cacheDir)
+        val outStream = FileOutputStream(file)
+        page.compress(Bitmap.CompressFormat.PNG, 0, outStream)
+        outStream.close()
+
+        _previewImagePath.value = file.path
+    }
+
+    fun onZoomTipShown() {
```
tobiasKaminsky, line 89: Maybe we want to show it 3 times, not only one time, so that people get more used to it?

tobiasKaminsky:
Awesome work!

The only thing I miss is a scrollbar, so that user can quickly scroll to end of pdf.

PR #9832, opened 2022-02-14
Grote/6830 no http response exception

tobiasKaminsky:
@grote I found this stale branch without any PR, so I created one.
Can you tell us what the status is? :-)


grote:
This is from 2 years ago and I can't really remember. I though this had been merged long ago, especially the changes to `src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java` getting rid of those extra threads and busy waiting.

PR #9853, opened 2022-02-17
Use split apks

### LINKED ISSUE #9848: Split builds
Ref: https://github.com/nextcloud/android/pull/9717#discussion_r806578925

- [ ] add split build support for APK
- [ ] still build one APK for Github Releases
- [ ] create QA version 
  -  without scan library ~~or~~
  - ~~with only ARM support, but then remove scan button~~

tobiasKaminsky:
closes #9848 

- [x] rebase and fix after merge of #9717

### DIFF
--- a/scripts/removeScan.sh
+++ b/scripts/removeScan.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+#
+# Nextcloud Android client application
+#
+# @author Tobias Kaminsky
+# Copyright (C) 2022 Tobias Kaminsky
+# Copyright (C) 2022 Nextcloud GmbH
+#  
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#  
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#  
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+#
+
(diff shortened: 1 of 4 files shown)


scripts/removeScan.sh
```diff
@@ -0,0 +1,30 @@
...
+#
+
+sed -i '/Document-Scanning/d' build.gradle
+rm src/main/java/com/owncloud/android/ui/activity/AppScanActivity.kt
+
+sed -i '/AppScanActivity/d' src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
+sed -i '/DocumentScanner/d' src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
+sed -i '/public void scanFromCamera(Activity activity, int requestCode) {/','/^    }/d' src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
+sed -i 's/scanFromCamera/uploadFromCamera/' src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
+sed -i '/android:id="@+id\/menu_scan_doc_upload\"/a android:visibility="gone"' src/main/res/layout/file_list_actions_bottom_sheet_fragment.xml
```
AlvaroBrey, line 30: Maybe it's better to generate a patch, keep it in the repo, and apply it here. It would be more resilient, I think.

tobiasKaminsky: Isn't a patch then line sensitive and we have to update it "all" the time?

AlvaroBrey: _should_ be context-sensitive rather than line sensitive, and we would have to update it if outside changes to these classes break it; but that's what we want

tobiasKaminsky:
@AlvaroBrey not needed, as we decided to use bundle for gplay, right? :)

AlvaroBrey:
> @AlvaroBrey not needed, as we decided to use bundle for gplay, right? :)

Correct, although we haven't reduced the universal APK size for QA. If we want to do that it's better to do it separately. Closing this.

PR #9865, opened 2022-02-21
Cache capabilities

### LINKED ISSUE #9854: Outsource ThemeUtils
Currently we any ThemeUtils needs to access capabilities and thus database, which sometimes results in >5 queries per activity.

I can imagine that a global state, provided via dagger (with `@singleton`) will reduce this and might enhance performance.

tobiasKaminsky:
Fix #9854 


tobiasKaminsky:
@AlvaroBrey we also have SessionMixin, which has capabilities bound to the session.
It now feels that we do it twice…

But it helps:
adb logcat -d | grep getCapability | wc is down from 316 to 18 🎉
(counts are not calls, as logcat is very verbos)

AlvaroBrey (commented):
Looks good, but it would be better to make CapabilityUtils and ThemeColorUtils injectable singletons, for better testability, rather than use static methods.

Additionally, I see there's still a lot of usages of `FileDataStorageManager.getCapability(String accountName)` throughout the codebase. If the intention is not to address those in this same PR, we should probably mark it as `@Deprecated` with a comment to use `CapabilityUtils` instead.

AlvaroBrey:
> we also have SessionMixin, which has capabilities bound to the session.
> It now feels that we do it twice…

We can change `SessionMixin` so that it calls `CapabilityUtils` instead of storing the capabilities itself:

```
    val capabilities: OCCapability?
        get() = getUser()
            .map { CapabilityUtils.getCapability(it, activity) }
            .orElse(null)
```

AlvaroBrey:
@tobiasKaminsky is this ready for review?

app/src/main/java/com/owncloud/android/utils/theme/CapabilityUtils.java
```diff
@@ -41,10 +48,43 @@ public static OCCapability getCapability(Context context) {
...
         if (user != null) {
-            FileDataStorageManager storageManager = new FileDataStorageManager(user, context.getContentResolver());
-            return storageManager.getCapability(user.getAccountName());
+            return getCapability(user, context);
         } else {
             return new OCCapability();
         }
     }
+
+    public static OCCapability getCapability(Account acc, Context context) {
```
AlvaroBrey, line 57:
```suggestion
    /**
     * @deprecated use {@link #getCapability(User, Context)} instead
     */
    @Deprecated
    public static OCCapability getCapability(Account acc, Context context) {
```

AlvaroBrey (approved):
Looks good! Minor comment about deprecation

Issue #9222, opened 2021-11-08
Remove jetifier by upgrading dependencies


tobiasKaminsky:
- use https://github.com/dipien/bye-bye-jetifier
- see currently following dependencies as a problem to disable jetifier
  - [ ] parceler
  - [ ] dexopener
  - [ ] mockito-android
  - [ ] eventbus
  - [ ] sectioned-recyclerview
  - [ ] glide
  - [ ] findsecbugs-plugin (maybe false alarm)

PR #9866, opened 2022-02-21
Migrate from Parceler to kotlin-parcelize

Unpublished:
Needs https://github.com/nextcloud/android-library/pull/824

For #9222 (removes Parceler as dep)


src/main/java/com/owncloud/android/datamodel/Template.kt
```diff
@@ -0,0 +1,50 @@
...
+@Parcelize
+data class Template(
+    val id: Int,
+    val name: String,
+    val thumbnailLink: String,
+    val type: Type,
+    val extension: String
+) : Parcelable
+
+enum class Type {
```
AlvaroBrey, line 40: Please put this enum class inside the data class so that it's still adressable as `com.owncloud.android.datamodel.Template.Type` rather than just `com.owncloud.android.datamodel.Type`.

src/androidTest/java/com/owncloud/android/ui/activity/ManageAccountsActivityIT.java
```diff
@@ -68,7 +70,7 @@ public void userInfoDetail() {
                                          "Address 123, Berlin",
                                          "https://www.nextcloud.com",
                                          "https://twitter.com/Nextclouders",
-                                         new Quota(),
+                                         new Quota(QUOTA_LIMIT_INFO_NOT_AVAILABLE),
```
AlvaroBrey, line 73: I would rather add an empty constructor to `Quota` with all zeroes, to keep the previous value untouched.

src/main/java/com/owncloud/android/files/FetchTemplateOperation.java
```diff
@@ -81,8 +82,8 @@ protected RemoteOperationResult run(OwnCloudClient client) {
                     templateArray.add(new Template(templateObject.getInt("id"),
                                                    templateObject.getString("name"),
                                                    templateObject.optString("preview"),
-                                                   Template.parse(templateObject.getString("type")
-                                                                             .toUpperCase(Locale.ROOT)),
+                                                   Type.Companion.parse(templateObject.getString("type")
```
AlvaroBrey, line 85: Can you annotate  `parse` as `@JvmStatic` to avoid having to use `Companion`?

src/main/java/com/owncloud/android/datamodel/Template.kt
```diff
@@ -0,0 +1,50 @@
...
+    DOCUMENT, SPREADSHEET, PRESENTATION, UNKNOWN;
+
+    companion object {
+        fun parse(name: String) = try {
+            valueOf(name.uppercase(Locale.getDefault()))
```
AlvaroBrey, line 45: Use `Locale.ROOT` or just don't pass any argument to `uppercase`. `Locale.getDefault()` depends on device languages, which we don't want for non-user-facing strings like these.

src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1741,13 +1741,11 @@ public void exitSelectionMode() {
...
-            (!TextUtils.isEmpty(event.getSearchQuery()) ||
-                event.searchType == SearchRemoteOperation.SearchType.SHARED_SEARCH ||
-                event.searchType == SearchRemoteOperation.SearchType.SHARED_FILTER ||
-                event.searchType == SearchRemoteOperation.SearchType.FAVORITE_SEARCH ||
-                event.searchType == SearchRemoteOperation.SearchType.RECENTLY_MODIFIED_SEARCH);
+        if (event == null) {
+            return false;
+        }
+        SearchRemoteOperation.SearchType searchType = event.getSearchType();
+        return !TextUtils.isEmpty(event.getSearchQuery()) || searchType == SearchRemoteOperation.SearchType.SHARED_SEARCH || searchType == SearchRemoteOperation.SearchType.SHARED_FILTER || searchType == SearchRemoteOperation.SearchType.FAVORITE_SEARCH || searchType == SearchRemoteOperation.SearchType.RECENTLY_MODIFIED_SEARCH;
```
AlvaroBrey, line 1748: Please break this into several lines (one per condition) for readability.

AlvaroBrey (changes requested):
Thank you @Unpublished! This looks really good, but there's some things that could be improved. Please look at the individual comments.

Unpublished:
Thanks for your review @AlvaroBrey, all improvements should be applied now.

AlvaroBrey (approved):

AlvaroBrey:
Thank you @Unpublished!

P.S: In the off chance you may be interested, [we're hiring](https://nextcloud.com/jobs/#androiddev) for this app.

PR #9870, opened 2022-02-22
Get rid of disk read on main thread

tobiasKaminsky:
With enabled strictMode we saw multiple warnings as access to DiskLRUCache and ArbitraryDataProvider was done in main thread.

### DIFF
--- a/src/main/java/com/owncloud/android/utils/DisplayUtils.java
+++ b/src/main/java/com/owncloud/android/utils/DisplayUtils.java
@@ -94,6 +94,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


src/main/java/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -494,16 +495,22 @@ public static void setAvatar(@NonNull User user,
...
+        new Thread(() -> {
+            ArbitraryDataProvider arbitraryDataProvider = new ArbitraryDataProvider(context.getContentResolver());
+
+            String eTag = arbitraryDataProvider.getValue(userId + "@" + serverName, ThumbnailsCacheManager.AVATAR);
+            String avatarKey = "a_" + userId + "_" + serverName + "_" + eTag;
+
+            bitmap.set(ThumbnailsCacheManager.getBitmapFromDiskCache(avatarKey));
+        }).start();
+
+        Drawable avatar = BitmapUtils.bitmapToCircularBitmapDrawable(resources, bitmap.get());
```
AlvaroBrey, line 513: This does not look like an adequate threading model. I've never worked with `AtomicReference` so it may do some magic, but here's what I understand is happening here:

1. `bitmap` reference is created
2. a `Thread` is started in parallel that does stuff with that reference
3. in the main thread, the reference is read without waiting for the new `Thread` at all

The result of this is that there's a read-write race condition between the two threads. What should be happening instead:

1. All the work in `setAvatar` is offloaded to a new thread (probably best to use `Executors.newSingleThreadExecutor`instead). The executor or thread has a `Handler` or some other way to access the main thread.
2. When the async work is finished, the thread uses that `Handler` to run whatever needs to be ran in the main thread.

AlvaroBrey (changes requested):
Unless I understand something wrong, the threading model does not look correct. Check the comment in `DisplayUtils` for details, but the same applies to the adapter.

PR #9873, opened 2022-02-22
Make external storage permission optional

### LINKED ISSUE #9343: Allow app to function partially without All Files access on Android 10+
On https://github.com/nextcloud/android/issues/9327, users are upset that All Files access is now required on Android 10+ devices.  We should be able to make the app work without this access, although with limited functionality.

## User perspectives

We have the clashing needs of

* Users who want to upload files that normal apps can't access (such as other apps' data. a common one is Whatsapp backups). These users need All Files permission.
* Users who don't want to grant Nextcloud access to their data. These users need the app to function without storage access.
* Finally, less savvy users may just want to upload photos and videos (media collections). This should still work with scoped storage (only READ_EXTERNAL_STORAGE).

### Option A
Implement logic to detect that the permission has been denied and not ask for it again on app startup. **BUT**, we should always ask for the permission again if the user attempts to do something that requires it (setup auto-upload, select files for upload...)

With this option, uploads will be almost completely non-functional without the permission. Browsing server content and downloading/sharing files should still work.
   
### Option B
* Use READ_EXTERNAL_STORAGE so that scoped storage provides access to at least some of the data on external storage
* Allow users to request this permission the first time they try to upload files or setup autoupload (or other access-sensitive tasks)
* Add a shortcut in Settings/Extra to request the  All Files permission later, if the user so desires.

With this option, users would get a choice to let Nextcloud access only media files or all files. It gives the users more choice but is also a bit more confusing (and harder to implement).

This option could also be implemented AFTER option A (first let the app work without uploads, and later get the app to work with the reduced, media-only access).

## Technical context:

Since November, all apps must abide by the new storage rules in Android 10+. These rules mostly consist of:

* The old "access external storage read-write"  permission is no longer available.
* The old "READ_EXTERNAL_STORAGE" permission now only allows access to media collections (mostly photos and videos).
* There is a new "All files" permission, that requires approval by Google and allows similar access to what we had before. This is what we're currently using.
   * Additionally we *require* this permission for the app to work now. It isn't optional, as some stuff was broken without it.
* The new "All files" permission grant UX is scarier for the user. They get taken to a system dialog where they have to manually flip a switch and read that Nextcloud will have access to all data. This is an intentional privacy-aware change on Google's part.
* Downloads to our private storage should still function without All Files access.

### LINKED ISSUE #9845: No system activity for All Files permission

android.content.ActivityNotFoundException: No Activity found to handle Intent { act=android.settings.MANAGE_APP_ALL_FILES_ACCESS_PERMISSION dat=package:com.nextcloud.android.beta }
	at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:2258)
	at android.app.Instrumentation.execStartActivity(Instrumentation.java:1914)
	at android.app.Activity.startActivityForResult(Activity.java:5326)
	at androidx.activity.ComponentActivity.startActivityForResult(ComponentActivity.java:588)
	at android.app.Activity.startActivityForResult(Activity.java:5284)
	at androidx.activity.ComponentActivity.startActivityForResult(ComponentActivity.java:574)
	at com.owncloud.android.utils.PermissionUtil.requestManageFilesPermission$lambda-1(PermissionUtil.kt:134)
	at com.owncloud.android.utils.PermissionUtil.$r8$lambda$FE9iUchxoTWOaaSGRZ8L-ymr578(Unknown Source:0)
	at com.owncloud.android.utils.PermissionUtil$$ExternalSyntheticLambda0.onClick(Unknown Source:2)
	at androidx.appcompat.app.AlertController$ButtonHandler.handleMessage(AlertController.java:167)
	at android.os.Looper.loop(Looper.java:233)
	at android.app.ActivityThread.main(ActivityThread.java:8068)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:631)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:978)

ID: com.nextcloud.android.beta
Version: 20220208
Build flavor: versionDev

Brand: OnePlus
Device: OnePlus6T
Model: ONEPLUS A6013
Id: RKQ1.201217.002
Product: OnePlus6T

SDK: 30
Release: 11
Incremental: 2111252327

### Steps to reproduce
1. 
2. 
3. 

### Expected behaviour
- Tell us what should happen

### Actual behaviour
- Tell us what happens

- Please create a test demo account and see if this still happens there.
- If yes, please open up a bug report
- If not, please verify server setup and ask for help on forum

### Environment data
Android version:

Device model: 

Stock or customized system:

Nextcloud app version:

Nextcloud server version:

Reverse proxy:

#### Web server error log
```
```

```
```

AlvaroBrey:
Fixes #9343
Fixes #9845

- [x] Don't force permissions on app launch
- [x] Force permissions for upload activity
- [x] Force permissions for synced folders activity
- [x] Allow choosing between All Files and just READ_EXTERNAL_STORAGE in SDK 30+
    - [x] Allow disabling All Files completely (~~through `setup.xml`?~~ Just remove the permission from the manifest, code detects that)
    - [x] Handle cases where there's no way to obtain All Files (#9845 ). Probably just assume that All Files is not available...
    - [x] Properly designed choice dialog

### DIFF
--- a/app/src/androidTest/java/com/nextcloud/client/GrantStoragePermissionRule.kt
+++ b/app/src/androidTest/java/com/nextcloud/client/GrantStoragePermissionRule.kt
@@ -21,10 +21,10 @@
 
 package com.nextcloud.client
 
+import android.Manifest
 import android.os.Build
 import androidx.test.platform.app.InstrumentationRegistry
 import androidx.test.rule.GrantPermissionRule
-import com.owncloud.android.utils.PermissionUtil
 import org.junit.rules.TestRule
 import org.junit.runner.Description
 import org.junit.runners.model.Statement
@@ -35,8 +35,7 @@ class GrantStoragePermissionRule private constructor() {
         @JvmStatic
         fun grant(): TestRule = when {
             Build.VERSION.SDK_INT < Build.VERSION_CODES.R -> GrantPermissionRule.grant(
-                PermissionUtil
-                    .getExternalStoragePermission()
+                Manifest.permission.WRITE_EXTERNAL_STORAGE
             )
             else -> GrantManageExternalStoragePermissionRule()
         }
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -35,6 +35,8 @@
     <uses-permission
         android:name="android.permission.MANAGE_EXTERNAL_STORAGE"
         tools:ignore="ScopedStorage" />
+    <uses-permission
+        android:name="android.permission.READ_EXTERNAL_STORAGE" />
 
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.VIBRATE" />
--- a/app/src/main/java/com/nextcloud/client/preferences/AppPreferences.java
+++ b/app/src/main/java/com/nextcloud/client/preferences/AppPreferences.java
@@ -373,4 +373,8 @@ public interface AppPreferences {
     void setPdfZoomTipShownCount(int count);
 
     int getPdfZoomTipShownCount();
+
+    boolean isStoragePermissionRequested();
+
+    void setStoragePermissionRequested(boolean value);
 }
--- a/app/src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
+++ b/app/src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
@@ -97,6 +97,8 @@ public final class AppPreferencesImpl implements AppPreferences {
 
     private static final String PREF__PDF_ZOOM_TIP_SHOWN = "pdf_zoom_tip_shown";
 
+    private static final String PREF__STORAGE_PERMISSION_REQUESTED = "storage_permission_requested";
+
     private final Context context;
     private final SharedPreferences preferences;
     private final CurrentAccountProvider currentAccountProvider;
@@ -694,6 +696,16 @@ public final class AppPreferencesImpl implements AppPreferences {
     @Override
     public int getPdfZoomTipShownCount() {
         return preferences.getInt(PREF__PDF_ZOOM_TIP_SHOWN, 0);
+    }
+
+    @Override
+    public boolean isStoragePermissionRequested() {
+        return preferences.getBoolean(PREF__STORAGE_PERMISSION_REQUESTED, false);
+    }
+
+    @Override
+    public void setStoragePermissionRequested(boolean value) {
+        preferences.edit().putBoolean(PREF__STORAGE_PERMISSION_REQUESTED, value).apply();
     }
 
     @VisibleForTesting
--- a/app/src/main/java/com/owncloud/android/datamodel/MediaProvider.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/MediaProvider.java
@@ -28,11 +28,8 @@ import android.net.Uri;
 import android.provider.MediaStore;
 import android.util.Log;
 
-import com.google.android.material.snackbar.Snackbar;
 import com.owncloud.android.MainApp;
-import com.owncloud.android.R;
 import com.owncloud.android.utils.PermissionUtil;
-import com.owncloud.android.utils.theme.ThemeSnackbarUtils;
 
 import java.io.File;
 import java.util.ArrayList;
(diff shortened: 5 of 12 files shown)


AlvaroBrey:
@nextcloud/designers I'd like to request feedback on the dialog design. If you have a better idea of how to present these choices, please let me know.

This is the dialog shown at app startup, at which point the permission is optional. The text will change in cases where permissions are absolutely required (file upload, autosync setup) to reflect that ("Permissions needed" or something like that).

<details>
<summary> Click to view screenshot </summary>

screen

</details>


Extra food for thought: we may want to skip this optional, first-time permission request; after all it will be requested when actually needed. This makes app usage a bit more seamless first time, BUT autoupload folder detection won't  work unless users grant the permissions first.

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
```diff
@@ -159,10 +160,6 @@ protected void onCreate(Bundle savedInstanceState) {
         // TODO: The content loading should be done asynchronously
         setupContent();
 
-        if (ThemeUtils.themingEnabled(this)) {
-            setTheme(R.style.FallbackThemingTheme);
-        }
-
```
tobiasKaminsky, line 165: Why is this removed? Seems unrelated to scope of PR?

src/main/java/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -635,20 +629,24 @@ public void onClick(View v) {
...
+                    new CheckAvailableSpaceTask(this, mFileListFragment.getCheckedFilePaths())
+                        .execute(mBehaviourSpinner.getSelectedItemPosition() == 0);
                 }
-                setResult(RESULT_OK, data);
-
-                finish();
             } else {
-                new CheckAvailableSpaceTask(this, mFileListFragment.getCheckedFilePaths())
-                    .execute(mBehaviourSpinner.getSelectedItemPosition() == 0);
+                PermissionUtil.requestExternalStoragePermission(this, true);
```
tobiasKaminsky, line 649:
```suggestion
                requestPermissions()
```

src/main/java/com/owncloud/android/utils/PermissionUtil.kt
```diff
@@ -76,70 +81,141 @@ object PermissionUtil {
...
+        )
+        else -> checkSelfPermission(context, Manifest.permission.WRITE_EXTERNAL_STORAGE)
     }
 
     /**
-     * Request relevant external storage permission depending on SDK.
+     * Request relevant external storage permission depending on SDK, if needed.
+     *
+     * Activities should implement [Activity.onRequestPermissionsResult]
+     * and handle the [PERMISSIONS_EXTERNAL_STORAGE] code, as well ass [Activity.onActivityResult]
```
tobiasKaminsky, line 106: *as

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
AlvaroBrey: Oh, I meant to ask you about this but I didn't remember. This breaks the styling of the buttons in the permissions dialog (all show up as black/grey instead of theme color), but I don't see anything else broken when I remove this block.

Can you clarify what it is used for?

src/main/java/com/owncloud/android/utils/PermissionUtil.kt
AlvaroBrey: LOL I will fix this

jancborchardt:
@AlvaroBrey text looks good, just 2 questions:
- It says "permissions to access **external** storage" – is it enough to just say "permissions to access storage", or is that permission there automatically?
- The dialog has no border radius, as per guidelines there’s a bit of border-radius to soften it (would also look better with our general design): https://material.io/components/dialogs#usage

AlvaroBrey:
>     * It says "permissions to access **external** storage" – is it enough to just say "permissions to access storage", or is that permission there automatically?

Good point, that's a technicality and can be removed.

>     * The dialog has no border radius, as per guidelines there’s a bit of border-radius to soften it (would also look better with our general design): https://material.io/components/dialogs#usage

I will try to see if that can be changed, but it's like that in all other dialogs in the app too :\

src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.java
AlvaroBrey: @tobiasKaminsky ping

tobiasKaminsky: I think this is for the toggle buttons, when enabling/disabling things in preference dialog:
image

AlvaroBrey:
@tobiasKaminsky ready for re-review

tobiasKaminsky (approved):

PR #9876, opened 2022-02-23
Move all Theme classes to DI

tobiasKaminsky:


tobiasKaminsky:
Seems that Github actions do not work when in draft mode…

AlvaroBrey:
> Seems that Github actions do not work when in draft mode…

I think it's because of the Drone commit (which includes `[skip ci]`). We could probably remove that from the drone commit message

AlvaroBrey:
> > Seems that Github actions do not work when in draft mode…
> 
> I think it's because of the Drone commit (which includes `[skip ci]`). We could probably remove that from the drone commit message

note: analysis commits are currently not working due to changes in the actions/checkout@v3, I'm looking into it

AlvaroBrey:
> > Seems that Github actions do not work when in draft mode…
> 
> I think it's because of the Drone commit (which includes `[skip ci]`). We could probably remove that from the drone commit message

Ugh, commits using the default Github Actions token (like we do in the analysis workflow) [don't trigger github actions](https://github.com/ad-m/github-push-action/issues/32). I'll change that to use a PAT from nextcloud-bot.

AlvaroBrey (changes requested):
Nice work! Some changes are needed though.

Aside from the comments below, both detekt and spotbugs need some attention.

tobiasKaminsky:
This now needs DI in TestActivity and I fail to do so.
Let us do this during company week.

AlvaroBrey:
> This now needs DI in TestActivity and I fail to do so. Let us do this during company week.

Fixed the crash in BottomSheetDialog in https://github.com/nextcloud/android/pull/9876/commits/d047907afce7bb74945f9fa7f692cbe5fe2756ca, but I don't know why CI is not running

AlvaroBrey:
> This now needs DI in TestActivity and I fail to do so. Let us do this during company week.

Fixed dialog crashes. Let's see how it goes

AlvaroBrey:
> > This now needs DI in TestActivity and I fail to do so. Let us do this during company week.
> 
> Fixed dialog crashes. Let's see how it plays out

@tobiasKaminsky this fixed the tests. All that is remaining is fixing detekt/lint/spotbugs

AlvaroBrey (approved):

PR #8805, opened 2021-08-04
Add stacktrace directly to github issue

tobiasKaminsky:
Question is now what to do
- enhance error reporting process to make more clear that user has to copy stacktrace manually
  - results maybe in better issues
- report only stacktrace, without template
  - which might cause more work on our side

Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

### Testing 
Writing tests is very important. Please try to write some tests for your PR. 
If you need help, please do not hesitate to ask in this PR for help.

[unit tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#unit-tests)
[instrumented tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#instrumented-tests)
[UI tests](https://github.com/nextcloud/android/blob/master/CONTRIBUTING.md#ui-tests)

- [x] Tests written, or not not needed

PR #9880, opened 2022-02-23
Fix LoggerImpl to also log throwable

Unpublished:


AlvaroBrey (approved):

AlvaroBrey:
We should also log the stacktrace to the file log (the `enqueue` call), but for that I'd better wait until #8805 is merged so that we have the stacktrace recursive formatter available.

Thanks @Unpublished !

PR #9884, opened 2022-02-24
Correct handling of creation timestamp

tobiasKaminsky:
Lib is https://github.com/nextcloud/android-library/pull/827

### DIFF
--- a/src/main/java/com/owncloud/android/utils/FileUtil.java
+++ b/src/main/java/com/owncloud/android/utils/FileUtil.java
@@ -20,9 +20,17 @@
 
 package com.owncloud.android.utils;
 
+import android.os.Build;
 import android.text.TextUtils;
 
+import com.owncloud.android.lib.common.utils.Log_OC;
+import com.owncloud.android.lib.resources.files.UploadFileRemoteOperation;
+
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/owncloud/android/utils/FileUtil.java
```diff
@@ -52,4 +60,20 @@ String getFilenameFromPathString(@Nullable String filePath) {
...
+        try {
+            return Files.readAttributes(file.toPath(), BasicFileAttributes.class)
+                .creationTime()
+                .to(TimeUnit.SECONDS);
+        } catch (IOException e) {
+            Log_OC.e(UploadFileRemoteOperation.class.getSimpleName(),
+                     "Failed to read creation timestamp for file: " + file.getName());
+            return -1L;
+        }
+    }
```
AlvaroBrey, line 78: What do `1L` and `-1L` mean for this?

AlvaroBrey (approved):

src/main/java/com/owncloud/android/utils/FileUtil.java
tobiasKaminsky: Uh. Both need to be -1L, which is undefined.
Maybe it should be null instead, to make it more visible?

AlvaroBrey: `null` is more understandable in my opinion. If you go with `-1L`, then please make a constant with a name like `TIMESTAMP_UNDEFINED` or something

tobiasKaminsky: Updated 👍

PR #9893, opened 2022-02-25
remoteId not needed when deleting a file

tobiasKaminsky:
This was an old leftover where remoteId was needed to remove cached thumbnail, but as this is LRU cache, this unneeded deletion was dropped.

Thus we also do not need the special test-only OCFile(String name, String remoteID).


AlvaroBrey (approved):
Nice!

PR #9894, opened 2022-02-25
Remove not needed toString() call

tobiasKaminsky:


src/main/java/com/owncloud/android/db/OCUpload.java
```diff
@@ -238,7 +238,7 @@ public String toFormattedString() {
             return localPath + " status:" + getUploadStatus() + " result:" +
                     (getLastResult() == null ? "null" : getLastResult().getValue());
         } catch (NullPointerException e) {
-            Log_OC.d(TAG, "Exception " + e.toString());
```
AlvaroBrey, line 241: Better use `Log_OC.d(TAG, "Exception", e)` instead of concatenating the exception

AlvaroBrey (approved):

src/main/java/com/owncloud/android/db/OCUpload.java
```diff
@@ -238,7 +238,7 @@ public String toFormattedString() {
...
+            Log_OC.d(TAG, "Exception " + e);
```
AlvaroBrey, line 241:
```suggestion
            Log_OC.d(TAG, "Exception", e);
```

AlvaroBrey (approved):

PR #9898, opened 2022-02-25
FileDisplayActivity: finally get rid of secondFragment

AlvaroBrey:
After https://github.com/nextcloud/android/pull/9791, secondFragment is finally unused.

Some smoke testing on my part found that behaviour is the same as before this change, but please
test some more so we can be as sure as possible.


tobiasKaminsky:
Wow. Super cool to have this finally away :-)

tobiasKaminsky (approved):

PR #9916, opened 2022-03-03
Chore: Update codeStyles/Project.xml

AlvaroBrey:
It's constantly being auto-changed by AS since latest update.

Do we even need to have this in our git repo? 


tobiasKaminsky:
> Do we even need to have this in our git repo?

We wanted to have this for margin (120) and other stuff.
Normally this only changes when switching Android Studio versions…

AlvaroBrey:
> > Do we even need to have this in our git repo?
> 
> We wanted to have this for margin (120) and other stuff. Normally this only changes when switching Android Studio versions…

That should probably be handled in `.editorconfig` instead. However, that's for another PR. Can we merge this?

tobiasKaminsky (approved):

PR #9921, opened 2022-03-04
Fix #9842: Run network operations in async task

### LINKED ISSUE #9842: Exception on sharing files stored on nextcloud instance
### Steps to reproduce
1. Share some files stored on nextcloud (either from within an application or file manager)

### Expected behaviour
- Well, a file shared

### Actual behaviour
- Crashes/Files not shared

- Please create a test demo account and see if this still happens there.
- If yes, please open up a bug report
- If not, please verify server setup and ask for help on forum

### Environment data
Android version: 10, lineageos, no Gapps

Device model: 
Google PixelXL

Nextcloud app version:
3.19.0 (F-Droid)

Nextcloud server version:
22.2.3

### Logs

[log.txt](https://github.com/nextcloud/android/files/8066935/log.txt)

* Since this did not happen with all applications I´ve opened https://github.com/vector-im/element-android/issues/5236 for element-android.
* I could not reproduce this with `Conversations` xmpp client.

akallabeth:
This pull request fixes running network tasks on UI thread.
Some applications invoke the `DocumentProvider` from `UI` thread. Although that is not really good practice it should not lead to the `DocumentProvider` failing to download files.

### DIFF
--- a/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
+++ b/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
@@ -86,9 +86,8 @@ import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.FutureTask;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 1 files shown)


AlvaroBrey (changes requested):
Theoretically, the `DocumentProvider` shouldn't even be making network requests in its API methods, but that's not the point of this PR I guess.

Either way, I don't think its a good idea to make this threading change in the `DownloadFileOperation` itself. `RemoteOperation`s (like that one) are also scheduled in background thread in other parts of the app, so this would be redundant.

Please instead do the threading changes in `DocumentProvider.openDocument`.

akallabeth:
@AlvaroBrey Ok, done.

AlvaroBrey (approved):

akallabeth:
@AlvaroBrey If I´m not mistaken you can remove my commit as yours completely replaces it.

AlvaroBrey:
> @AlvaroBrey If I´m not mistaken you can remove my commit as yours completely replaces it.

That would not  count your contribution, and there's people who care about that :)  I've instead signed off your commit myself. Thank you!

Issue #6855, opened 2020-08-30
Parallel downloads thumbnails pictures / Concurrent server requests


kfirfer:
The android app making calls to the server synchronized when fetching thumbnails, one thumbnail at a time

In the web version its parallel, sending multiple API calls at once

I think it will improve performance significant in the android app if he will be parallel thumbnails fetch's as-well 

App version: 3.13.0
Android version: 10

tobiasKaminsky:
Totally true.
I am not sure what holds it back, but I fear that our underlying http library is somehow preventing this.

tobiasKaminsky:
I started 50 threads with some remote operation, all were started in the same second:
```
1598858210.881 com.nextcloud.client D/HttpClient: 203063136 start
1598858210.886 com.nextcloud.client D/HttpClient: 240626713 start
1598858210.888 com.nextcloud.client D/HttpClient: 220903103 start
1598858210.904 com.nextcloud.client D/HttpClient: 263102698 start
1598858210.904 com.nextcloud.client D/HttpClient: 254669013 start
1598858210.905 com.nextcloud.client D/HttpClient: 22937563 start
1598858210.906 com.nextcloud.client D/HttpClient: 123525752 start
1598858210.906 com.nextcloud.client D/HttpClient: 233424209 start
1598858210.906 com.nextcloud.client D/HttpClient: 145931959 start
1598858210.907 com.nextcloud.client D/HttpClient: 72212918 start
1598858210.907 com.nextcloud.client D/HttpClient: 46212644 start
1598858210.907 com.nextcloud.client D/HttpClient: 162817421 start
1598858210.910 com.nextcloud.client D/HttpClient: 13525826 start
1598858210.910 com.nextcloud.client D/HttpClient: 19566419 start
1598858210.911 com.nextcloud.client D/HttpClient: 66879632 start
1598858210.911 com.nextcloud.client D/HttpClient: 115569038 start
1598858210.912 com.nextcloud.client D/HttpClient: 17747337 start
1598858210.912 com.nextcloud.client D/HttpClient: 222612911 start
1598858210.913 com.nextcloud.client D/HttpClient: 22091196 start
1598858210.913 com.nextcloud.client D/HttpClient: 260946245 start
1598858210.913 com.nextcloud.client D/HttpClient: 237574298 start
1598858210.914 com.nextcloud.client D/HttpClient: 236623307 start
1598858210.915 com.nextcloud.client D/HttpClient: 190162113 start
[...]

PR #9922, opened 2022-03-04
SyncedFoldersActivity: asynchronous loading

### LINKED ISSUE #8863: App restarts when opening the AutoUpload Settings
### Steps to reproduce
1. Go to Settings, Auto Upload -> Works as expected
2. Enable Sync for a Folder with ~200 Files (upload of existing files even disabled)
3. Go back to Main Screen
4. Go to Settings, Auto Upload again -> Activity opens noticeably slower (~ 2 sec)
5. Enable Sync for a Folder with ~400 Files
6. Go back to Main Screen
7. Go to Setting, Auto Upload again -> App freezes for ~5 sec, then is restarted

### Expected behaviour
- Auto Upload Settings can be opened and modified

### Actual behaviour
- App automatically restarts without any Error Log (probably runs into ANR, though nothing is reported by the system)
- There is no way to recover/modify the Auto Upload Settings once broken
  - Only solution is to remove the Account and relogin
- Additional Information:
  - Marking Folders as hidden from the AutoUpload List also noticeably prolongs the time the Activity loads
  - I switched from a Redmi to a Realme Device recently, issue does not occur on old phone
    - Old Phone is running Nextcloud 3.13.1 and new one 3.16.1
    - going to try with 3.13.1 on new phone later on

- Not tested yet (does not look like a Server Side issue)

### Environment data
Android version:
- Android 11 (Security Update from 2021-06-05)

Device model: 
- realme 8 (RMX3085)

Stock or customized system:
- Stock system
  - realm UI 2.0

Nextcloud app version:
- 3.16.1
- also tested with 3.17.0 RC3, same behaviour

Nextcloud server version:
 - 20.0.4

Reverse proxy:

#### Web server error log
```
```

```
```
No error log is shown, App just restarts. Is there a way to retrieve the App log anyway?

AlvaroBrey:
- Converted to Kotlin first.
- Load synced folders asynchronously
- Load thumbnails in parallel

See commit list for details

Fixes #8863
Also related to #6855 (need to fix it in more places)


detekt.yml
```diff
@@ -1,5 +1,5 @@
 build:
-  maxIssues: 10
+  maxIssues: 11
```
tobiasKaminsky, line 2: Nice trick ;-)

tobiasKaminsky (approved):

detekt.yml
AlvaroBrey: This is because of `mergeFolderData`, which raises a  `TooComplexMethod` issue. It was already too complex before porting to Kotlin, but I didn't want to refactor at the same time of these changes, but I also didn't like suppressing it, because we _should_ fix it at some point.

Let me know if you would approach it in a different way.

tobiasKaminsky (approved):

PR #9932, opened 2022-03-08
Fix analysis workflow checkout on PRs

AlvaroBrey:
Broken since upgrade to checkout@v3


AlvaroBrey:
For reference, fixed by https://github.com/nextcloud/android/commit/421a672e1d26f23523b32c09bd7e2ce3be5ff15a

PR #9933, opened 2022-03-08
Fix CI on analysis commits

AlvaroBrey:
To auto-cancel previous Github Actions on new pushes, check this: https://stackoverflow.com/a/67939898


AlvaroBrey:
Well, after some changes here and on the workflow file (on master) I finally have gotten the analysis commits to run with the proper token, so they run the expected Github Actions now. 

However, now the drone run isn't executed for the analysis commits. Is drone configured somehow so that it ignores commits from nextcloud-android-bot, or something like that? @tobiasKaminsky

tobiasKaminsky:
Nice one 👍👍👍
Can you rebase to get rid of testing commits and then we can merge it  :)

AlvaroBrey:
Note to self: replicate changes in library when this is merged

tobiasKaminsky (approved):

PR #9959, opened 2022-03-11
Fix sharing from media view

tobiasKaminsky:
- open media view
- select an image
- select send/share
- share to an user

This now works, but it is too small, and I have no clue where this size is set:
image
@AlvaroBrey do you have an idea?

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/fragment/FileDetailsSharingProcessFragment.kt
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/FileDetailsSharingProcessFragment.kt
@@ -162,6 +162,7 @@ class FileDetailsSharingProcessFragment : Fragment(), ExpirationDatePickerDialog
 
     private fun showShareProcessFirst() {
         binding.shareProcessGroupOne.visibility = View.VISIBLE
+        binding.shareProcessEditShareLink.visibility = View.VISIBLE
         binding.shareProcessGroupTwo.visibility = View.GONE
 
         // set up UI for modifying share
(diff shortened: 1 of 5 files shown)


AlvaroBrey:
>  @AlvaroBrey do you have an idea?

Layout for the details sharing fragment is http://github.com/nextcloud/android/blob/5da6fd235f037e52cd5537e87b901e7135514281/src/main/res/layout/file_details_sharing_process_fragment.xml#L28-L28, but I don't think it's prepared for showing in a dialog.

tobiasKaminsky:
> but I don't think it's prepared for showing in a dialog.

What do you mean by this? I know that dialogs are not fullscreen, but in above screenshot it is so tiny, that there is still half of height possible, but not shown…

AlvaroBrey:
> > but I don't think it's prepared for showing in a dialog.
> 
> What do you mean by this? I know that dialogs are not fullscreen, but in above screenshot it is so tiny, that there is still half of height possible, but not shown…

I mean that the layout for that fragment has `    android:layout_height="match_parent"`, and the `ScrollView` inside has `        android:layout_height="0dp"` and is sized by constraints. I don't know how that works inside a `Dialog`, which has variable height per se.

app/src/main/java/com/owncloud/android/ui/fragment/FileDetailsSharingProcessFragment.kt
```diff
@@ -309,6 +309,7 @@ class FileDetailsSharingProcessFragment : Fragment(), ExpirationDatePickerDialog
      */
     private fun showShareProcessSecond() {
         binding.shareProcessGroupOne.visibility = View.GONE
+        binding.shareProcessEditShareLink.visibility = View.GONE
```
AlvaroBrey, line 312: Restoring the visibility is needed in `showShareProcessFirst`. Otherwise the text will remain hidden if the user advances to the note screen and then presses Cancel.

AlvaroBrey (changes requested):
Nice! I've pushed https://github.com/nextcloud/android/pull/9959/commits/f6298f79044703c0cbd59cb31d63e1c48dab2712 to improve the minHeight approach a bit.

Additionally, please look at the comment below.

AlvaroBrey (approved):

PR #9969, opened 2022-03-15
build: Specify NDK version (optional install)

AlvaroBrey:
If this + CMake is installed, native libraries are correctly stripped, and then APK size is reduced.
If not installed, everything still works as we don't actually need to compile any native stuff.

If `ndkVersion` is not specified, then the default version for Android Gradle Plugin is expected,
and stripping will not work even if other NDK version is installed. So it's better to use a
specific version, to prevent it from stopping working in the future after an AGP update.

To benefit from this change, run: `sdkmanager 'ndk;21.4.7075529' 'cmake;3.18.1'` or install them from AS SDK manager.

Note: NDK and CMake seem to already be installed in GH Actions' `ubuntu-latest` image, so no changes are expected there (QA build). However I've also explicitly installed them in the QA workflow, in case that changes in the future.


tobiasKaminsky:
https://github.com/nextcloud/android/blob/master/scripts/buildDev#L12

We should it also add to buildDev, as this script builds the (daily) apk:
sdkmanager 'ndk;21.4.7075529' 'cmake;3.18.1'

Do you have in idea how to centralize the version numbers?

AlvaroBrey:
> https://github.com/nextcloud/android/blob/master/scripts/buildDev#L12
> 
> We should it also add to buildDev, as this script builds the (daily) apk: sdkmanager 'ndk;21.4.7075529' 'cmake;3.18.1'
> 
> Do you have in idea how to centralize the version numbers?

Yeah, I think we can through gradle.properties. I'll make a PoC shortly

AlvaroBrey:
> https://github.com/nextcloud/android/blob/master/scripts/buildDev#L12
> 
> We should it also add to buildDev, as this script builds the (daily) apk: sdkmanager 'ndk;21.4.7075529' 'cmake;3.18.1'
> 
> Do you have in idea how to centralize the version numbers?

Done in ec98c5d for `qa.yml` and `build.gradle`. I didn't do it for `buildDev`, as I don't know the location of `sdkmanager` in the environment that `buildDev` is executed on.

tobiasKaminsky:
> > 
> 
> Done in [ec98c5d](https://github.com/nextcloud/android/commit/ec98c5db526b6fa1274c79cc9504ed78fb155c75) for `qa.yml` and `build.gradle`. I didn't do it for `buildDev`, as I don't know the location of `sdkmanager` in the environment that `buildDev` is executed on.

Done. It is in $path :-)

tobiasKaminsky (approved):

PR #9972, opened 2022-03-15
Fix search loading animation and display

### LINKED ISSUE #9373: Empty placeholder flashes while loading

AlvaroBrey:
Fixes:
 - When switching search type (shared, favorites, etc), loading animation doesn't show 
 - When switching search type the empty message is displayed instead of "Loading..." (Fixes #9373)


app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1510,6 +1510,87 @@ public void onMessageEvent(final SearchEvent event) {
         handleSearchEvent(event);
     }
 
+    private static class SearchAsyncTask extends AsyncTask<Void, Void, Boolean> {
```
tobiasKaminsky, line 1513: Maybe we want to extract this class, so that Fragment file is not that big?

tobiasKaminsky (approved):

app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
AlvaroBrey: Done in  https://github.com/nextcloud/android/pull/9972/commits/769816978a1342d1c54b4cdb63e07e2e489d7c67

PR #9984, opened 2022-03-18
workflows/qa: Skip NDK install if PR is not from this repo

AlvaroBrey:
We already do this for the rest of steps, I forgot to add it when I added the NDK install

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -20,6 +20,7 @@ jobs:
                     distribution: "temurin"
                     java-version: 11
             -   name: Install NDK and cmake
+                if: ${{ steps.check-secrets.outputs.ok == 'true' }}
                 run: |
                     source ndk.env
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Why is this needed? Forked repo would still benefit from this to have a reduced QA size?

AlvaroBrey:
> Why is this needed? Forked repo would still benefit from this to have a reduced QA size?

QA already is skipped for forks, due to missing secrets. See https://github.com/nextcloud/android/pull/8861

AlvaroBrey:
> > Why is this needed? Forked repo would still benefit from this to have a reduced QA size?
> 
> QA already is skipped for forks, due to missing secrets. See #8861

Let me add: if we do want QA builds for forks, it can technically be done (using `pull_request_target` instead of `pull_request`). Let me know what you think.

tobiasKaminsky:
> if we do want QA builds for forks

This would be great, as community do great PR and this way they can be tested more easily.

AlvaroBrey:
> > if we do want QA builds for forks
> 
> This would be great, as community do great PR and this way they can be tested more easily.

OK, I will do this instead in a new PR

PR #9985, opened 2022-03-18
Fastscroll for all lists

### LINKED ISSUE #6297: add fast scrollbar like in Google Photos
I'm always frustrated when I have to scroll in folders with 500+ files. It's slow if I have to the middle or to the 3/4 of the data. Google has a good solution for this in their Photos app.

As Google has implemented it in Web and App: there's are timeline scrollbar at the right. With that scrollbar you can navigate ultra fast through your files. If you've added 500+ files on one day it's still pretty fast to scroll through.
grafik

No timeline - but a scrollbar which is faster and always visible.

tobiasKaminsky:
Fix #6297 


AndyScherzinger:
I thought latest recycler implementation of android can now do this more or less out of the box, not sure if it also supports the sectioned part though.

tobiasKaminsky:
@jancborchardt @AlvaroBrey can you test this and give me feedback?

AlvaroBrey:
> @jancborchardt @AlvaroBrey can you test this and give me feedback?

Some comments:
1. I find  it hard to trigger the fast scroll, the trigger area seems to be too narrow. I need to drag my finger on the very very edge of the screen for the fast scroll to appear.  This is made worse because you can only start the scroll from where the scrollbar currently is, but it's hidden by default and autohides.
2. There is also no indication that the fast scroll exists, without knowing it's there already. We should perhaps make the scrollbar visible at all times, not only when dragging. This should also "fix" number 1 somewhat.
3. The performance of fast scrolling is pretty choppy, it doesn't feel smooth at all. We may be able to improve this by making `getSectionName` more efficient, but it might just be an issue with how we fetch and process data.
5. For alphabetic sort specifically, I would capitalize all letters in indicator. It feels weird to see the capitalization change throughout the scroll, and doesn't really add that much value.
6. The size indicator doesn't really seem useful to me. It's different for each element, in most cases, and we can already see the size in each item. 

Additionally, there is (I think) something weird with the UX here, which is that our sorting orders don't actually correspond with what we say they do. When sorting A-Z we're not _actually_ sorting A-Z. We're sorting favorites, then folders, then files, and each of those subsections is A-Z.  This means that if we have favorites, folders and files, the indicator will go from A-Z three times during the entire scrolling. It's not terrible but it's definitely weird.

tobiasKaminsky:
Fastscroll is now always enabled, fixing 1 and 2.
3. not sure, but even returning a static string "A" feels choppy, although a bit less.
4. done
5. True, but would feel weird if there is no info when sorted by size

Basic requirement was to have it in GalleryFragment, but I thought it would be nice on all lists.
Maybe I was too optimistic.
If Jan also does not like it, then I can limit it to GalleryFragment.

tobiasKaminsky:
@jancborchardt @nimishavijay 
please see https://github.com/nextcloud/android/pull/9985#issuecomment-1091829006

jancborchardt:
Just for the record what I said to @tobiasKaminsky in our call:
We should go for the default way where the Fastscroll is not enabled all the time. Reasons:
- Scrolling long lists is not something that is done all the time
- Most Android apps (except nerdy ones) don’t show a scroll bar all the time)
- It’s not a "feature" we want or need to advertise. It should come in handy when needed (scrolling a bunch) and then you can use it.
- If the common way that this is done in Android apps changes, we can adjust of course. :)

tobiasKaminsky:
@AlvaroBrey I merged this already to move on, please check after your vacation and ping me for any problems found in code review :)

AndyScherzinger:
@tobiasKaminsky backport to 3.20 branch then?

PR #10006, opened 2022-03-22
SETUP: update local library linking instructions with composite builds

AlvaroBrey:
Reference: https://docs.gradle.org/current/userguide/composite_builds.html

Not only is this the "official" way to do this, but also:
 - It's much simpler (changes only one file)
 - No need for symlinks
 - Android Studio support seems to finally be perfect


tobiasKaminsky (approved):

tobiasKaminsky:
Wow!
This is super cool and super helpful!

PR #10008, opened 2022-03-22
Staggered grid view

tobiasKaminsky:

Metadata might not be always available:
If no metadata: 
 - show square placeholder 
 - download preview, show as cropped image --> no flickering view
 - store metadata from preview in OCFile
 - next time use metadata

- [ ] use https://github.com/nextcloud-gmbh/server/issues/263#issuecomment-1090556686

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -38,7 +38,6 @@ import android.media.MediaMetadataRetriever;
 import android.media.ThumbnailUtils;
 import android.net.Uri;
 import android.os.AsyncTask;
-import android.provider.MediaStore;
 import android.text.TextUtils;
 import android.view.Display;
 import android.view.View;
@@ -546,11 +545,7 @@ public final class ThumbnailsCacheManager {
                         tagId = String.valueOf(((TrashbinFile) mFile).getRemoteId());
                     }
                     if (String.valueOf(imageView.getTag()).equals(tagId)) {
-                        if (gridViewEnabled) {
-                            BitmapUtils.setRoundedBitmapForGridMode(bitmap, imageView);
-                        } else {
-                            BitmapUtils.setRoundedBitmap(bitmap, imageView);
-                        }
+                        imageView.setImageBitmap(bitmap);
                     }
                 }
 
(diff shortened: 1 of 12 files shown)


tobiasKaminsky:
@AlvaroBrey @AndyScherzinger @jancborchardt can you test this please?
Go to media, let it fetch thumbnails and then check if you are satisfied with the result.
(without metadata info, this is only way to test staggered view)

I am really unsure about this…

I also tested it with 2 columns.

AlvaroBrey:
> @AlvaroBrey @AndyScherzinger @jancborchardt can you test this please? Go to media, let it fetch thumbnails and then check if you are satisfied with the result. (without metadata info, this is only way to test staggered view)
> 
> I am really unsure about this…
> 
> I also tested it with 2 columns.
> 
> Edit: we can also have a short call, and I show it to you.

There's some weird behaviours when scrolling up and down (with gaps), I guess this would be solved if we had the metadata and the placeholder could be sized appropriately.

Either way, I'm afraid I don't really like this. This looks like it would be good for showcasing something (like in a portfolio) but I find it harder to find specific pictures, which is the intented purpose of this section. Additionally it will be weird when having pictures grouped by year, etc.

tobiasKaminsky:
> There's some weird behaviours when scrolling up and down (with gaps), I guess this would be solved if we had the metadata and the placeholder could be sized appropriately.

Yes, this should be fixed with metadata.
It also does not happen when all thumbnails are loaded.



> Either way, I'm afraid I don't really like this. This looks like it would be good for showcasing something (like in a portfolio) but I find it harder to find specific pictures, which is the intented purpose of this section. 

That is also what I am afraid of. And the reason I ask for all your feedback.

> Additionally it will be weird when having pictures grouped by year, etc.

Then there will be "just" date separator spanning across all three columns.

tobiasKaminsky:
Do you think this is better? https://github.com/bidrohi/asymmetrical-gallery-example

AlvaroBrey:
> Do you think this is better? https://github.com/bidrohi/asymmetrical-gallery-example

Yep, having constant height definitely makes it easier to parse through visually. Let's see what the others think about both approaches.

jancborchardt:
We should definitely go for the horizontal sorting approach (agreeing with @AlvaroBrey) instead of vertical. However, we need to make sure that the photos are _not_ cropped, and thus the height of rows can and should not be constant, but would be based on the photos within.

Example of the algorithm:
- If in phone portrait mode, we should in most cases show exactly 2 photos per row, be it 1 landscape 1 portrait or 2 landscapes
- The exception is when there are 2 portraits, if the next one is a portrait too, we can do 3 portraits next to each other.
- These photos are then put next to each other (plus the whitespace inbetween) and sized so they are the same height, then treated as a row group
- The width of that group is sized to the viewport
→ That should be it

Of course when we switch to phone landscape, this needs to reflow. There the default can probably be upped by 1 each, so 3 photos per row, or 4 if it’s all portraits. (Have to test if that looks better, or the portrait default also works for landscape.)
Also note there is inside whitespace between photos and rows, but no whitespace on the outside towards the screen edge.



**Some references:**
- [Amazon Photos](https://play.google.com/store/apps/details?id=com.amazon.clouddrive.photos) does the view well, with varying row height:
   image
- This is also how [Flickr](https://flickr.com/search/?text=nextcloud) does it – no cropping of pictures, row heights are variable.
- Unsure about [Google Photos](https://play.google.com/store/apps/details?id=com.google.android.apps.photos) (my phone is broken atm so can not test) – judging from the screenshots they use grid view on phone and staggered grid on tablet, but not sure if they use different heights



**For the record, I was under the impression that this would be a separate view, not replacing the current view. Especially considering Google Photos and iOS Photos at least on smartphone uses grid view with cropping by default (just like we do in Nextcloud Photos on web), I’m hesitant to have us fully replace it** @AndyScherzinger @karlitschek

AndyScherzinger:
Can't say about replacing since having multiple entering points to. Look at photos also seems strange.
As for the view, I also did propose the asymmetrical grid and the variable line height might be a next step.
Also since you mentioned it @jancborchardt the Google photos app does it a bit different, so they do the square grid 3x3 basically while mix in images spanning 2x2 - on mobile. For tablet size they seem to then switch to the asymmetrical grid layouting. Since we don't have specific layouts for mobile and tablet we would need to decide for one.

So if the existing view and its layout stay how and where would you put the new one then?

jancborchardt:
> Also since you mentioned it @jancborchardt the Google photos app does it a bit different, so they do the square grid 3x3 basically while mix in images spanning 2x2 - on mobile.

Yes – iOS Photos also does this, but I’d say let’s fully ignore that for now.

> For tablet size they seem to then switch to the asymmetrical grid layouting. Since we don't have specific layouts for mobile and tablet we would need to decide for one.

Yes, Tobi mentioned that we don’t have distinction – so let’s ignore that for now as well.


> So if the existing view and its layout stay how and where would you put the new one then?

Already communicated with @tobiasKaminsky as well: In the interest of time I’m fine with the new tile layout replacing the old grid layout in the media tab. (Especially cause we have grid view in the Files view still.) However we should still keep in mind that Google Photos and iOS Photos does it a certain way and we want to compete with them → but something for the future.

AndyScherzinger:
Sounds good to me 👍

AlvaroBrey:
> So if the existing view and its layout stay how and where would you put the new one then?

I would add a "layout change"  icon button like we have for list/grid in the files list, although in this case it would change square grid/staggered grid.

Issue #9996, opened 2022-03-21
Use Hilt for Dagger configuration


AlvaroBrey:
Reference: https://developer.android.com/training/dependency-injection/hilt-android

This will ease most of the Dagger handling, in a standard way and without having to manage custom classes (most of our `di` package will become unnecessary as it's handled by Hilt).

ezaquarii:
@AlvaroBrey can you loop me in for the reviews? I'm curious how it will pan out.

AlvaroBrey:
@ezaquarii sure, although this is more of a reminder to look into this in the future, than a concrete plan :)

PR #10035, opened 2022-03-31
Hilt PoC

AlvaroBrey:
Rel #9996 

Signed-off-by: Álvaro Brey Vilas <alvaro.brey@nextcloud.com>


AlvaroBrey:
@ezaquarii you might be interested in this. Sorry for the huge single commit! Didn't want to deal with partial migrations.

Apart from what's described in #9996, the only notable change is that `ContentProvider`s and subclasses of `PreferenceActivity` have custom `EntryPoint`s, as they lack out-of-the-box support by Hilt.

ezaquarii:
No worries, I'll take a look later today.

app/src/main/java/com/owncloud/android/providers/FileContentProvider.java
```diff
@@ -118,19 +119,25 @@
...
+        HashMap<String, String> tempMap = new HashMap<>();
         for (String projection : ProviderTableMeta.FILE_ALL_COLUMNS) {
             tempMap.put(projection, projection);
         }
         FILE_PROJECTION_MAP = Collections.unmodifiableMap(tempMap);
     }
 
 
-    @Inject protected Clock clock;
+    protected Clock clock;
```
ezaquarii, line 130: Protected scope is probably not needed here. We can do with a stricter default  package level.

app/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
```diff
@@ -322,15 +330,19 @@ private void saveDownloadedFile(FileDataStorageManager storageManager, DownloadF
...
-
+        inject();
         // initiate storage manager collection, because we need to serve document(tree)s
         // with persist permissions
         initiateStorageMap();
 
         return true;
     }
 
+    private void inject() {
```
ezaquarii, line 341: Yuck!

app/src/main/java/com/owncloud/android/ui/activity/ThemedPreferenceActivity.java
```diff
@@ -38,7 +41,13 @@
     private boolean themeChangePending;
     private boolean paused;
 
-    @Inject AppPreferences preferences;
+    AppPreferences preferences;
+
+    @EntryPoint
+    @InstallIn(SingletonComponent.class)
+    public interface ThemedPreferenceEntrypoint {
```
ezaquarii, line 48: Why `@AndroidEntryPoing` cannot be used? It's `Activity` after all.

ezaquarii (commented):
Looks like the conversion is straight forward. The only hideous kludge is this `inject()` helper that pulls stuff from some global singleton, but this was a problem from day 1 as some components do not offer good lifecycle API. I'd consider pulling stuff from `Application` in such case.

app/src/main/java/com/owncloud/android/ui/activity/ThemedPreferenceActivity.java
AlvaroBrey: Hilt [only supports](https://dagger.dev/hilt/android-entry-point#android-types) activities extending from `ComponentActivity`, which all `AppCompatActivity's do. However `PreferenceActivity` extends directly from `Activity` and as such is not supported, so we have to resort to `EntryPoint`s.

(side note: `PreferenceActivity` is also deprecated)

app/src/main/java/com/owncloud/android/providers/FileContentProvider.java
AlvaroBrey: Yeah, for `ContentProvider`s and `PreferenceActivity` I just left the fields as they were, but with the `EntryPoint` they can all be private.

app/src/main/java/com/owncloud/android/providers/DocumentsStorageProvider.java
AlvaroBrey: Yeah, I don't particularly like it either. For Kotlin classes I could probably at least setup the "injected" fields with a getter that pulls directly from the `EntryPoint`, but no such luxuries are afforded in Java.

I still think this is better than pulling stuff directly from the Application instance, because it doesn't pollute the Application class further, and gives a clear view of the needs of the injected class. Also should be easier to change for tests. However if you have any idea of how to make this cleaner for the few classes that need it, I'm up for it.

AlvaroBrey:
This was, as the title indicate, just a PoC. I'll try to do the changes in a more organized manner in the future.

Issue #9992, opened 2022-03-21
Group photos as a timeline


tobiasKaminsky:
Fix https://github.com/nextcloud/files-clients/issues/20

PR #10050, opened 2022-04-06
Group photos as timeline by month

tobiasKaminsky:

When merging https://github.com/nextcloud/android/issues/9992, we need in GalleryAdapter.kt:
- [x] change  
``` 
override fun getSectionName(position: Int): String {
        return DisplayUtils.getMonthYear(files[getRelativePosition(position).section()].date, context)
    }
```

- [x] remove ocFileListFragmentInterface.isGalleryFragment() and instead use sortOrderType directly in both getSectionName() functions.


app/src/main/java/com/owncloud/android/ui/adapter/GalleryAdapter.kt
```diff
@@ -0,0 +1,233 @@
...
+    override fun isCheckedFile(file: OCFile): Boolean {
+        return ocFileListDelegate.isCheckedFile(file)
+    }
+
+    override fun getCheckedItems(): Set<OCFile> {
+        return ocFileListDelegate.checkedItems
+    }
+
+    override fun removeCheckedFile(file: OCFile) {
+        ocFileListDelegate.removeCheckedFile(file)
```
AlvaroBrey, line 209: Not entirely sure if this is possible to do with the current interfaces, but you can avoid this delegation boilerplate by doing this:

1. Have `OCFileListDelegate` implement `CommonOCFileListAdapterInterface`, or a specific interface for the delegate if not possible.
2. Pass an `OCFileListDelegate` instance to `GalleryAdapter`'s constructor and use `by` to delegate all methods in the interface to the delegate automatically:

In short:

```kotlin
public class OCFileListDelegate implements OCFileListDelegateInterface
...
class GalleryAdapter(delegate: OCFileListDelegate,...): SectionedRecycler..., OCFileListDelegateInterface by delegate {
``` 
This way, all the methods in the delegate interface are automatically delegated. More info here: https://kotlinlang.org/docs/delegation.html

Additionally, if you don't want to build a delegate instance outside the Adapter, just make the constructor private, and add a static `newInstance` method that takes care of building the delegate and calling the private constructor.

app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -1188,271 +884,71 @@ protected void publishResults(CharSequence constraint, Filter.FilterResults resu
...
-        }
-
-        @Override
-        public ImageView getShared() {
-            return binding.sharedIcon;
+    @NonNull
+    @Override
+    public String getSectionName(int position) {
+        // sort
+        Enum<FileSortOrder.SortType> sortOrderType;
```
AlvaroBrey, line 902: This can just be a `FileSortOrder.SortType`, no need for the `Enum` wrapper

app/src/main/java/com/owncloud/android/ui/asynctasks/GallerySearchTask.java
```diff
@@ -100,7 +100,7 @@ public GallerySearchTask(GalleryFragment photoFragment,
                 boolean emptySearch = parseMedia(startDate, endDate, result.getData());
                 long lastTimeStamp = findLastTimestamp(result.getData());
 
-                photoFragment.getAdapter().showAllGalleryItems(storageManager);
+                photoFragment.getGalleryAdapter().showAllGalleryItems(storageManager);
```
AlvaroBrey, line 103: Better to call a method in the `Fragment` that then calls the `Adapter`, rather than expose the `Adapter`to the `AsyncTask`

app/src/main/java/com/owncloud/android/ui/dialog/ConflictsResolveDialog.java
```diff
@@ -194,17 +194,17 @@ public Dialog onCreateDialog(Bundle savedInstanceState) {
...
-                                       false,
-                                       getContext());
+        OCFileListDelegate.setThumbnail(existingFile,
+                                        binding.existingThumbnail,
+                                        user,
+                                        new FileDataStorageManager(user,
+                                                                   requireContext().getContentResolver()),
+                                        asyncTasks,
+                                        false,
+                                        getContext());
```
AlvaroBrey, line 207: This is a clear indication that this method should live somewhere else, not in any `OCFileList*` class. Maybe `DisplayUtils` for now?

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
```diff
@@ -229,4 +265,13 @@ private void loadMoreWhenEndReached(@NonNull RecyclerView recyclerView, int dy)
             }
         }
     }
+
+    @Override
+    public boolean isGalleryFragment() {
+        return true;
+    }
```
AlvaroBrey, line 272: The `isGalleryFragment` method breaks abstraction.  Since it's only used in the `getSectionName` method in order to get the sort order type, I'd replace it for a `getSortOrderType` method.

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
```diff
@@ -229,4 +265,13 @@ private void loadMoreWhenEndReached(@NonNull RecyclerView recyclerView, int dy)
...
+
+    public GalleryAdapter getGalleryAdapter() {
+        return mAdapter;
+    }
```
AlvaroBrey, line 276: See my comment on `GallerySearchTask`

AlvaroBrey (commented):
Great work! The approach generally looks good, although there are a few minor things that can be improved.

Additionally, there's a bunch of new small classes and interfaces that can be easily converted to Kotlin, but that can be done later if it's too much work.

app/src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -231,8 +231,9 @@ protected void attachBaseContext(Context base) {
         // we don't want to handle crashes occurring inside crash reporter activity/process;
         // let the platform deal with those
         final boolean isCrashReportingProcess = getAppProcessName().endsWith(":crash");
+        final boolean useExceptionHandler = !appInfo.isDebugBuild();
 
-        if (!isCrashReportingProcess) {
+        if (!isCrashReportingProcess && useExceptionHandler) {
```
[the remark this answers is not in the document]
tobiasKaminsky: This will be a separate PR. I just added it to have crashes again in logcat.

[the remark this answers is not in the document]
tobiasKaminsky: PR: https://github.com/nextcloud/android/pull/10059

app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
tobiasKaminsky: Argh. This was just a test to check if fast scroll works. It should never been in this pr, but only in #9985 
Let me revert it (and take your comment to the other PR)

tobiasKaminsky: Reverted

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
tobiasKaminsky: This is removed in this PR and will be handled better in #9985 after merging this.

AndyScherzinger (approved):
I agree, looks very nice design wise 💙

app/src/main/java/com/owncloud/android/ui/adapter/GalleryAdapter.kt
tobiasKaminsky: Let us try this a bit later, I somehow fail 🙈

AlvaroBrey (approved):

PR #10052, opened 2022-04-06
File locking

### LINKED ISSUE #10012: Implement file locking
Ref: https://github.com/nextcloud/files-clients/issues/2

AlvaroBrey:
Closes #10012
Needs https://github.com/nextcloud/android-library/pull/851

- [x] Show lock status
    - [x] Lock status on file list
    - ~~[] Lock status on details view~~ will show up as activity
    - [x] Lock details popup (owner, time, type, TTL) 
- [x] Allow manual lock/unlock
    - [x] Only allow unlock if lock owner is user and lock is manual type 
- [x] Prevent write operations on locked files
    - [x]  Prevent delete
    - [x] Prevent move (allow copy)


app/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -917,6 +918,12 @@ public void toggleEncryption(OCFile file, boolean shouldBeEncrypted) {
         }
     }
 
+    public void toggleFileLock(OCFile file, boolean shouldBeLocked) {
+        if (file.isLocked() != shouldBeLocked) {
+            EventBus.getDefault().post(new FileLockEvent(file.getRemotePath(), shouldBeLocked));
```
AlvaroBrey, line 923: Note for reviewers: I am unsure about this approach. I thought about using OperationsService, but it does not support NextcloudClient I think.

app/src/main/java/com/owncloud/android/ui/events/FileLockEvent.kt
```diff
@@ -0,0 +1,22 @@
+/**
+ * Nextcloud Android client application
+ *
+ * @author Álvaro
+ * Copyright (C) 2017 Mario Danic
```
tobiasKaminsky, line 5: 2022 Alvaro?

app/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
tobiasKaminsky: Fine for now, as we use EventBus also in lots of other places.
Once we feel uncomfortable with it, we can find a better solution and change it all in one rush.

app/src/main/res/drawable/ic_locked_dots_small.xml
```diff
@@ -0,0 +1,13 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
tobiasKaminsky, line 1: License?

tobiasKaminsky (approved):

PR #10079, opened 2022-04-14
Tile view

tobiasKaminsky:
Sucessor of #10008 


app/src/main/java/com/owncloud/android/datamodel/ImageDimension.kt
```diff
@@ -0,0 +1,26 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.owncloud.android.datamodel
+
+import java.io.Serializable
+
+data class ImageDimension(var width: Float = -1f, var height: Float = -1f) : Serializable
```
AlvaroBrey, line 26: Any reason why this needs to be `Serializable`? if so, it needs a serialVersionUuid

app/src/main/java/com/owncloud/android/datamodel/OCFile.java
```diff
@@ -948,4 +952,13 @@ public String getLockToken() {
     public void setLockToken(@Nullable String lockToken) {
         this.lockToken = lockToken;
     }
+
+    public void setImageDimension(@NonNull ImageDimension imageDimension) {
```
AlvaroBrey, line 956: getter is nullable but setter is non null

app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -252,6 +253,163 @@ public static Bitmap getBitmapFromDiskCache(String key) {
...
+        private OCFile getFile() {
+            return file;
+        }
+
+        private String getImageKey() {
+            return imageKey;
+        }
+    }
+
+    public static class GalleryImageGenerationTask extends AsyncTask<Object, Void, Bitmap> {
```
AlvaroBrey, line 274: We should not be writing new `AsyncTask`s at this point, they are very deprecated and lead to too many problems. Use coroutines or an Executor instead, if possible

AlvaroBrey (commented):
The approach generally looks good, but it needs a lot of polish. Current implementation does weird things with changing sizes when refreshing, crashes when scrolling down, and shows the wrong aspect ratio for some images.

app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -509,6 +509,7 @@ private ContentValues createContentValuesForFile(OCFile file) {
         cv.put(ProviderTableMeta.FILE_LOCK_TIMEOUT, file.getLockTimeout());
         cv.put(ProviderTableMeta.FILE_LOCK_TOKEN, file.getLockToken());
         cv.put(ProviderTableMeta.FILE_MODIFIED, file.getModificationTimestamp());
+        cv.put(ProviderTableMeta.FILE_METADATA_SIZE, new Gson().toJson(file.getImageDimension()));
```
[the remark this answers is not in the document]
tobiasKaminsky: True. But we are doing it already for other values.
Let us keep it for now, and once we have a better plan, we can move it.

app/src/main/java/com/owncloud/android/datamodel/ImageDimension.kt
tobiasKaminsky: Indeed not needed.

app/src/main/java/com/owncloud/android/datamodel/OCFile.java
tobiasKaminsky: Can be null.

app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
tobiasKaminsky: With both I do not have real experience.
Do you have an example within our app?

AlvaroBrey: Executors: http://github.com/nextcloud/android/blob/5daf12fd65cdf1a27a5b4f00186ead6f3ae325b3/app/src/main/java/com/owncloud/android/ui/preview/PreviewMediaFragment.java#L321-L321

Coroutines: http://github.com/nextcloud/android/blob/5daf12fd65cdf1a27a5b4f00186ead6f3ae325b3/app/src/main/java/com/owncloud/android/ui/fragment/SharedListFragment.kt#L108-L108

tobiasKaminsky: I tend to postpone this, until we/I resolved all bugs…

AlvaroBrey (changes requested):

AlvaroBrey:
Also, fast scroller needs to be removed from GalleryFragment

app/src/main/java/com/owncloud/android/ui/adapter/GalleryAdapter.kt
```diff
@@ -245,8 +257,27 @@ class GalleryAdapter(
...
+                }
+            }
+        }
+
+        // month, row
+        return if (item == null || row == null) {
+            getAbsolutePosition(0, 0)
+        } else {
+            getAbsolutePosition(item, row)
+        }
```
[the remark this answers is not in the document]
tobiasKaminsky: Awesome!

app/src/main/java/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -829,13 +839,109 @@ static public void openSortingOrderDialogFragment(FragmentManager supportFragmen
...
+        if (context == null) {
+            context = MainApp.getAppContext();
+        }
+        df = new SimpleDateFormat(pattern, context.getResources().getConfiguration().locale);
         df.setTimeZone(TimeZone.getTimeZone(TimeZone.getDefault().getID()));
 
         return df.format(timestamp);
     }
 
+    public static void setGalleryImage(OCFile file,
```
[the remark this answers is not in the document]
tobiasKaminsky: This will still have two calls to static DisplayUtils, but it is way better.

AlvaroBrey:
Rebased and fixed after m3 merge ✨

Issue #9598, opened 2021-12-21
Constantly pinging index.php/204 but not auto uploading


Mazvy:
Clients have enabled auto upload on their devices, the apps are constantly pinging `index.php/204` in bursts but not auto uploading anything. 

Sometimes the clients do start to upload, but more often then not it requires them to open the app and only then do the files start to transfer over.

Seeing as this pinging is constant and periodic, I assume the background process is working and not put to sleep by Android, however auto uploads are more often than not - not getting through.

    xx.x.x.xxx - - [21/Dec/2021:17:51:16 +0200] "GET /index.php/204 HTTP/1.1" 204 6640 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.18.0"
    xx.x.x.xxx - - [21/Dec/2021:17:51:17 +0200] "GET /index.php/204 HTTP/1.1" 204 6660 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.18.0"
    xx.x.x.xxx - - [21/Dec/2021:17:54:12 +0200] "GET /index.php/204 HTTP/1.1" 204 6676 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.18.0"
    xx.x.x.xxx - - [21/Dec/2021:17:54:12 +0200] "GET /index.php/204 HTTP/1.1" 204 6672 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.18.0"
    xx.x.x.xxx - - [21/Dec/2021:17:56:26 +0200] "GET /index.php/204 HTTP/1.1" 204 6654 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.18.0"
    xx.x.x.xxx - - [21/Dec/2021:17:56:26 +0200] "GET /index.php/204 HTTP/1.1" 204 6648 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.18.0"
    xx.x.x.xxx - - [21/Dec/2021:17:57:49 +0200] "GET /index.php/204 HTTP/1.1" 204 6676 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.18.0"
    xx.x.x.xxx - - [21/Dec/2021:17:57:50 +0200] "GET /index.php/204 HTTP/1.1" 204 6650 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.18.0"
[...]

PR #10087, opened 2022-04-16
Do not ping the Server for checking the Connection

obel1x:

This is for fixing the Issue #9598, which leads to high battery drainage and unneccessary communication. The method which pinged, is called multiple times in a row by the source, which causes many pings to the server even if in powersave-mode. I removed the pings completely, as i did not see any reason to ping the server at all and fixed powersaving- handling.

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/OfflineSyncWork.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/OfflineSyncWork.kt
@@ -57,17 +57,19 @@ class OfflineSyncWork constructor(
 
     override fun doWork(): Result {
         val wakeLock: WakeLock? = null
-        if (!powerManagementService.isPowerSavingEnabled && !connectivityService.isInternetWalled) {
-            val users = userAccountManager.allUsers
-            for (user in users) {
-                val storageManager = FileDataStorageManager(user, contentResolver)
-                val ocRoot = storageManager.getFileByPath(OCFile.ROOT_PATH)
-                if (ocRoot.storagePath == null) {
-                    break
+        if (!powerManagementService.isPowerSavingEnabled) {
+            if (!connectivityService.isInternetWalled) { // Put in a new CodeBlock to minimise impact while powersaving
+                val users = userAccountManager.allUsers
+                for (user in users) {
+                    val storageManager = FileDataStorageManager(user, contentResolver)
+                    val ocRoot = storageManager.getFileByPath(OCFile.ROOT_PATH)
+                    if (ocRoot.storagePath == null) {
+                        break
+                    }
+                    recursive(File(ocRoot.storagePath), storageManager, user)
                 }
-                recursive(File(ocRoot.storagePath), storageManager, user)
+                wakeLock?.release()
             }
-            wakeLock?.release()
         }
         return Result.success()
     }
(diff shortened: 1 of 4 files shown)


obel1x:
for me this looks fine - there are warnings about the possibility to combine the if-blocks. This may not be done, to make powersaving effective and is intentionally done otherwise. Can those warning be supressed?

One other thing: is it really good in OfflineSyncWork dowork() to initiate wakeLock before the if statement, but to relaese it only inside? That means not release if the statement evaluates to false... I would expect it to either initiated inside and released inside or both outside the statement...

obel1x:
> The point of pinging the server is to check if there is connection to the server. We can't trust the device connectivity status because there may be a captive portal, or a local DNS issue, or something of the sort, that would still allow connectivity to other sites but not to the server. There is no working around this.

There is not need to check the connectivity like this. If you want to upload or download, then do it. If there is no connectivity, fetch that error and react to it the right way. This ping massively costs bandwidth, battery and serverload and is unnecessary at all - there is no benefit in knowing if the other side can be pinged or not.

After testing my version for some time, battery drain is gone, logs are clean and everything still works. There are many other downsides in that app, that may be removed - but at least that massive downside has gone.

> cc @tobiasKaminsky can you add some context on OfflineSyncWork? I'm not entirely familiar with its purpose.
> 
> > for me this looks fine - there are warnings about the possibility to combine the if-blocks. This may not be done, to make powersaving effective and is intentionally done otherwise. Can those warning be supressed?
> 
> Those warnings are correct, there is no benefit to separating the `if`s. In Java (and Kotlin) if you have a condition like `a() && b()`, and `a()` is `false`, then `b()` is not evaluated at all. This is known as a logical short circuit.

don't know if those "if"s are really working always like this. If you can guaranty this, you may change it. Mind, that i have changed the way that evaluations were done. I wouldn't be sure about this working like it was before the right way.

> 
> > One other thing: is it really good in OfflineSyncWork dowork() to initiate wakeLock before the if statement, but to relaese it only inside? That means not release if the statement evaluates to false... I would expect it to either initiated inside and released inside or both outside the statement...
> 
> As far as I can see, that wakelock is unused and uninitialized. Seems to be a leftover from some previous code.

jep, so remove it here or in another pull?

Over all, there are only some warnings, but problem is solved.
Please merge this or do some other stuff to eleminate the battery drain and pings.

ezaquarii:
> don't know if those "if"s are really working always like this. 

They are, unless you do some JVM tuning, which is not applicable on Android anyway.

If this is going to land, I believe you should also update `ConnectivityServiceTest`.

ezaquarii:
@AlvaroBrey if we'd like to retain this feature of captive portal checks, we could cache the result of the check and this way avoid expensive re-checking every time we initiate sync. Cache could be invalidated by network status change.

ezaquarii:
>  As far as I can see, that wakelock is unused and uninitialized. Seems to be a leftover from some previous code.

Yes, this was ported from legacy Java code and converted to Kotlin. Tests were added to check that behaviour did not change during conversion.

None of the above mean that the code makes any sense. :)

AlvaroBrey:
> @AlvaroBrey if we'd like to retain this feature of captive portal checks, we could cache the result of the check and this way avoid expensive re-checking every time we initiate sync. Cache could be invalidated by network status change.

Yup, we already do a similar thing with the DNS cache.

Additionally I think in the case of OfflineSyncWork we can skip the check entirely, and just fail silently if `CheckEtagOperation` can't connect; it would be basically the same result as currently.

obel1x:
Should be fine with #10448

PR #10095, opened 2022-04-18
Drone: force-enable apps on master server

AlvaroBrey:
When server master version has been updated, but `appinfo.xml` on apps has not been updated yet, the apps can't be installed so our CI with master server fails always.

Better solution: build our CI test images separately, instead of configuring the server for every build. This way we would have working master images (even if slightly older) for our CI, until we can build newer images.

Signed-off-by: Álvaro Brey Vilas <alvaro.brey@nextcloud.com>

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -102,11 +102,11 @@ services:
       - su www-data -c "php /var/www/html/occ group:adduser users user1"
       - su www-data -c "php /var/www/html/occ group:adduser users user2"
       - su www-data -c "git clone -b master https://github.com/nextcloud/activity.git /var/www/html/apps/activity/"
-      - su www-data -c "php /var/www/html/occ app:enable activity"
+      - su www-data -c "php /var/www/html/occ app:enable -f activity"
       - su www-data -c "git clone -b master https://github.com/nextcloud/text.git /var/www/html/apps/text/"
-      - su www-data -c "php /var/www/html/occ app:enable text"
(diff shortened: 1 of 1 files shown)


AlvaroBrey:
I've decided to close this because:
1. we should move to prebuilt images as detailed in the original post, I believe that is a better approach
2. merging this PR might mean that we get flaky tests due to mismatched apps. the server build failing is clearer.

tobiasKaminsky:
> Better solution: build our CI test images separately, instead of configuring the server for every build. This way we would have working master images (even if slightly older) for our CI, until we can build newer images.

This cannot work with latest master versions, as those need to be checked out  / updated after starting image.
Or is there a way to do things directly after image start?

AlvaroBrey:
> > Better solution: build our CI test images separately, instead of configuring the server for every build. This way we would have working master images (even if slightly older) for our CI, until we can build newer images.
> 
> This cannot work with latest master versions, as those need to be checked out / updated after starting image. Or is there a way to do things directly after image start?
My preference would be to automatically update the master image daily (nightly?). That way we would have almost-up-to-date image for CI, but if the build fails one night, at least it won't break our CI while we fix it.

tobiasKaminsky:
Okay :) Go for it.
We then need a proper detection if master image is really updated every night, so that we do not have a silent failing update and thus test against an outdated master.

tobiasKaminsky:
Plus point for this approach is also speed up, as not every CI needs to update all code 👏👏👏

PR #10117, opened 2022-04-27
Screenshot tests reliability improvements

AlvaroBrey:


AlvaroBrey:
I've restarted the screenshot workflow several times and it seems like reliability is greatly improved by slowing down these 4 tests a bit.

tobiasKaminsky (approved):

tobiasKaminsky:
While sleep is not the best approach, it is a good tradeoff for having more reliable tests 👏👏👏

AlvaroBrey:
> While sleep is not the best approach, it is a good tradeoff for having more reliable tests 

I agree, however `waitForIdleSync` does not seem to be enough for these cases :\

PR #10121, opened 2022-04-27
Media view filter options

surinder-tsys:
Media view with filter options:
1. User can select specific media folder to see the medias inside it.
2. User can filter if he wants to see only images.
3. User can filter if he wants to see only videos.

Media_Options

media_image_seleted

media_video_selected


AlvaroBrey:
Ping @jancborchardt for design review

As it stands, this UX looks like it's a 1:1 copy of the iOS application, but I can see some small issues with it:

1. The "check" mark when  something is hidden doesn't add any value in my opinion, since the icon and the text of the action already change to reflect state. It also is confusing because it looks like the option is "selected", not "active", because of the checkmark; this is not a common interaction in Android.
    * I would probably just remove the  checkmark
2. The show/hide state for images and videos is stored in appPreferences so it is global. This means that if the user sets filters, leaves the gallery and later comes back, the filters will still be applied. _But_ there will be no indication that the view is being filtered, unless the user opens the three dot menu.
    * I would make this state non-persistent, i.e. resets if you leave gallery and come back
3. I think `Select the "Media" folder` is not a great text for what it does (showing only media in specific directories), but I can't think of a clearly better alternative. To me, it makes it look like "Media" is an actual folder, and this option will change it somehow.
    * Maybe just `Select folder` is better?  

@surinder-tsys Thanks for this PR! The changes above are just for design discussion, no need to make them yet. We'll do a code review soon.

surinder-tsys:
Hi @AlvaroBrey I have updated the PR with the mentioned changes. 
Please have a look.
default_state

both_selected

videos_only

single_selection

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
```diff
@@ -221,6 +271,76 @@ public void searchCompleted(boolean emptySearch, long lastTimeStamp) {
...
+    private String setDefaultRemotePath() {
+        if (remoteFilePath == null) {
+            setRemoteFilePath(remotePath);
+        }
+        return remotePath;
+    }
+
+    private void setRemoteFilePath(String remotePath) {
+        remoteFilePath = new OCFile(remotePath);
+        remoteFilePath.setFolder();
```
AlvaroBrey, line 292: `setFolder` is only for testing purposes and you shouldn't need it. You are starting at `/`and only letting the user select folders, so it should always be a folder.

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
```diff
@@ -221,6 +271,76 @@ public void searchCompleted(boolean emptySearch, long lastTimeStamp) {
...
+                                        themeColorUtils.appBarPrimaryFontColor(requireContext()));
+        }
+
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+
+        // Handle item selection
+        if (item.getItemId() == R.id.action_three_dot_icon) {
```
AlvaroBrey, line 314: Combine these two `if` statements with `&&`

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
```diff
@@ -221,6 +271,76 @@ public void searchCompleted(boolean emptySearch, long lastTimeStamp) {
...
+                galleryFragmentBottomSheetDialog.show();
+                return true;
+            }
+        }
+        return super.onOptionsItemSelected(item);
+    }
+
+    @Override
+    public void onActivityResult(int requestCode, int resultCode, @Nullable Intent data) {
+        if (requestCode == SELECT_LOCATION_REQUEST_CODE) {
```
AlvaroBrey, line 325: Combine these two `if` statements with `&&`

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
```diff
@@ -221,6 +271,76 @@ public void searchCompleted(boolean emptySearch, long lastTimeStamp) {
...
+        super.onActivityResult(requestCode, resultCode, data);
+    }
+
+    private void searchAndDisplayAfterChangingFolder() {
+        mAdapter.resetAdapter();
+        mediaObject.clear();
+        runGallerySearchTask();
+    }
+
+    private void runGallerySearchTask() {
```
AlvaroBrey, line 343: Use this function also on line 229 (searchAndDisplay)

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
```diff
@@ -259,19 +379,55 @@ private void loadMoreWhenEndReached(@NonNull RecyclerView recyclerView, int dy)
...
-        mAdapter.showAllGalleryItems(mContainerActivity.getStorageManager());
+        mAdapter.showAllGalleryItems(mContainerActivity.getStorageManager(), remoteFilePath.getRemotePath(),
+                                     mediaObject, galleryFragmentBottomSheetDialog.isHideVideos(),
+                                     galleryFragmentBottomSheetDialog.isHideImages(),
+                                     imageList, videoList, this);
+
+        updateSubtitle(galleryFragmentBottomSheetDialog.isHideVideos(), galleryFragmentBottomSheetDialog.isHideImages());
+    }
+
+    private void updateSubtitle(boolean isHideVideos, boolean isHidePhotos) {
```
AlvaroBrey, line 420: Use a state enum rather than two booleans, as already commented on other classes.

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragmentBottomSheetDialog.java
```diff
@@ -0,0 +1,134 @@
...
+import com.owncloud.android.ui.activity.FileActivity;
+
+import androidx.core.content.ContextCompat;
+
+public class GalleryFragmentBottomSheetDialog extends BottomSheetDialog {
+
+    //media view states
+    private static final int MEDIA_STATE_DEFAULT = 1; //when both photos and videos selected
+    private static final int MEDIA_STATE_PHOTOS_ONLY = 2;
+    private static final int MEDIA_STATE_VIDEOS_ONLY = 3;
```
AlvaroBrey, line 42: Better to use an enum rather than constantes

app/src/main/java/com/owncloud/android/utils/theme/ThemeMenuUtils.java
```diff
@@ -43,4 +49,17 @@ public void tintMenuItemText(MenuItem item, int color) {
...
         item.setTitle(newItemTitle);
     }
+
+    /**
+     * tinting menu item color
+     *
+     * @param item    the menu item object
+     * @param color   the color wanted as a color resource
+     */
+    public void tintMenuIcon(@NonNull MenuItem item, int color) {
```
AlvaroBrey, line 59: Nice!

app/src/main/res/drawable/ic_tick.xml
```diff
@@ -0,0 +1,12 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
AlvaroBrey, line 1: Needs license header

app/src/main/res/drawable/ic_video_camera.xml
```diff
@@ -0,0 +1,12 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
AlvaroBrey, line 1: Needs license header

AlvaroBrey (changes requested):
Thanks! Some things to fix: 

* Please format your code (`./gradlew ktlintFormat`)

* There's a bug with the toolbar subtitle; once it gets set in `GalleryFragment`, it keeps showing on the rest of views (if you change to Favorites from Media for example). To prevent this, please add a call to clear the subtitle at  `FileDisplayActivity.setLeftFragment`.

* Please take a look at the comments below for other things to fix

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
```diff
@@ -54,10 +70,25 @@ public class GalleryFragment extends OCFileListFragment {
...
+
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         searchFragment = true;
+
+        setHasOptionsMenu(true);
+
+        if (galleryFragmentBottomSheetDialog == null) {
+            FileActivity activity = (FileActivity) getActivity();
```
AlvaroBrey, line 88:
```suggestion
```
Unused

app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -528,6 +530,10 @@ private void setLeftFragment(Fragment fragment) {
             searchView.post(() -> searchView.setQuery(searchQuery, true));
         }
         setDrawerIndicatorEnabled(false);
+
+        //clear the subtitle while going from
+        updateToolbarSubtitle("");
```
AlvaroBrey, line 535: The proper way to clear the subtitle is to call `actionBar.setSubtitle(null)`, not an empty string. Additionally the comment is incomplete.

AlvaroBrey (changes requested):
Great! Just two more minor things from my part

AlvaroBrey (approved):
Thank you @surinder-tsys !

@tobiasKaminsky please review

app/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
```diff
@@ -128,6 +132,14 @@ protected void onCreate(Bundle savedInstanceState) {
                     mSearchOnlyFolders = true;
                     mDoNotEnterEncryptedFolder = true;
                     break;
+                case CHOOSE_LOCATION:
+                    caption = getResources().getText(R.string.choose_location).toString();
+                    mSearchOnlyFolders = true;
+                    mDoNotEnterEncryptedFolder = true;
+                    mChooseBtn.setText(getResources().getString(R.string.common_select));
```
tobiasKaminsky, line 139: That's unusual to put it here.
If we want to have this, then it should also be for move/copy operation.

In general this is uncommon to have, I think.

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
```diff
@@ -56,10 +72,23 @@ public class GalleryFragment extends OCFileListFragment {
     private int limit = 300;
     private GalleryAdapter mAdapter;
 
+    private static final int SELECT_LOCATION_REQUEST_CODE = 212;
+    private OCFile remoteFile = new OCFile("/");//default root path
+    private List<OCFile> mediaObject;
```
tobiasKaminsky, line 77: This List should be in Adapter, as a Fragment should not contain any real data.
We have similar approach in OcFileListAdapter, with mFiles and mFilesAll.

But actually in this case, as all items are in storageManager.allGalleryItems, I am unsure if two lists are needed, or if it not enough to retrieve subset directly from storageManager

tobiasKaminsky (changes requested):
Please see my comments

app/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.java
surinder-tsys: @tobiasKaminsky  Can you elaborate more on this? What is your expectation?

tobiasKaminsky: With uncommon I mean to have a "✔" on a button.
See https://material.io/components/buttons#behavior, this is used in only rare cases, but not for a "normal" confirm dialog.

surinder-tsys: Removed the tick icon from button

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
```diff
@@ -56,10 +72,23 @@ public class GalleryFragment extends OCFileListFragment {
     private int limit = 300;
     private GalleryAdapter mAdapter;
 
+    private static final int SELECT_LOCATION_REQUEST_CODE = 212;
+    private OCFile remoteFile = new OCFile("/");//default root path
```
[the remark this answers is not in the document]
surinder-tsys: Accessed remote file path from storage manager.

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
surinder-tsys: Removed the mediaObject list from GalleryFragment.

tobiasKaminsky:
@surinder-tsys this looks good now, but it seems that you did a mistake while rebasing, as there are now way too many commits?
Can you recheck this?

surinder-tsys:
> @surinder-tsys this looks good now, but it seems that you did a mistake while rebasing, as there are now way too many commits? Can you recheck this?

@tobiasKaminsky I have reduced the commits now. But still there are duplicate commits which caused due to rebasing I suppose.
I tried but couldn't find anything to get rid of them. If you have any idea please let me know.

tobiasKaminsky:
Alvaro re-based it and create a new PR here: #10484

Issue #10126, opened 2022-04-29
StrictMode warning & potential ANR problem with fixStoragePath()


manycore-com:
### Steps to reproduce
MainApp.fixStoragePath() checks SD card (if present) which can be potentially slow.
I propose we put this initialization in a thread. This will improve app start and decrease likelihood of ANR.
I have prepared a patch and will make a PR.

### Expected behaviour
This is an improvement, not a bug. App works great without it.

### Environment data
Android version:11

Device model: 
moto g10

Stock or customized system:
stock

Nextcloud app version:
/master

Nextcloud server version:

Reverse proxy:

PR #10127, opened 2022-04-29
ANR fix for fixStoragePath() Issue #10126

manycore-com:
Signed-off-by: manycore-com <marcus@manycore.io>

### DIFF
--- a/app/src/main/java/com/owncloud/android/MainApp.java
+++ b/app/src/main/java/com/owncloud/android/MainApp.java
@@ -90,6 +90,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.inject.Inject;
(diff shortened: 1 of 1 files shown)


AlvaroBrey:
The problem with this is that solving the issue creates another potential issue: the entire point of this patch is to not block `onCreate` while `fixStoragePaths` is running, but that also means that `onCreate` may potentially continue executing, and even some Activities may start, without the path being fixed.

Additionally, this method _should_ only be called once since it sets a preference when it's done.

Ideally this sort of stuff wouldn't be done on Application.onCreate, but we would have a more streamlined bootup sequence, with some sort of early-boot Activity that could show a loading animation while these kinds of things are done in background. However, that is not the case, so I think we'll have to deal with StrictMode violations for now.

@tobiasKaminsky what's your opinion on this?

AlvaroBrey:
@manycore-com as per the above, I'm closing this PR. However, thank you a lot for your interest and initiative!

PR #10168, opened 2022-05-04
Make SAF root IDs stable through server version updates

pelzvieh:
Bug #9906 seems to be triggered by a version change of the nextcloud server the client is connected to. It appears the most logical solution to make nextcloud object IDs stable against server version changes. These object IDs are derived from the User object using account/anonymous, and the referenced Server object. Providing a hashCode function in Server that ignores the server's version therefore provides stability of object IDs. This seems safe to do, as an URI is already designed to be an identificator.

### DIFF
--- a/app/src/main/java/com/nextcloud/client/utils/HashUtil.kt
+++ b/app/src/main/java/com/nextcloud/client/utils/HashUtil.kt
@@ -1,0 +1,42 @@
+/*
+ *  Nextcloud Android Library is available under MIT license
+ *
+ *  @author Álvaro Brey Vilas
+ *  Copyright (C) 2022 Álvaro Brey Vilas
+ *  Copyright (C) 2022 Nextcloud GmbH
+ *
+ *  Permission is hereby granted, free of charge, to any person obtaining a copy
+ *  of this software and associated documentation files (the "Software"), to deal
+ *  in the Software without restriction, including without limitation the rights
+ *  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ *  copies of the Software, and to permit persons to whom the Software is
+ *  furnished to do so, subject to the following conditions:
+ *
+ *  The above copyright notice and this permission notice shall be included in
+ *  all copies or substantial portions of the Software.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ *  EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ *  MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ *  NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ *  BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ *  ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ *  CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ *  THE SOFTWARE.
+ */
+
+package com.nextcloud.client.utils
+
+import org.apache.commons.codec.binary.Hex
+import java.security.MessageDigest
+
+object HashUtil {
+    private const val ALGORITHM_MD5 = "MD5"
+
+    @JvmStatic
+    fun md5Hash(input: String): String {
+        val digest = MessageDigest.getInstance(ALGORITHM_MD5)
+            .digest(input.toByteArray())
+        return String(Hex.encodeHex(digest))
+    }
+}
--- a/app/src/androidTest/java/com/owncloud/android/providers/DocumentsStorageProviderIT.kt
+++ b/app/src/androidTest/java/com/owncloud/android/providers/DocumentsStorageProviderIT.kt
@@ -38,7 +38,7 @@ class DocumentsStorageProviderIT : AbstractOnServerIT() {
     private val authority = context.getString(R.string.document_provider_authority)
 
     private val rootFileId = storageManager.getFileByEncryptedRemotePath(ROOT_PATH).fileId
-    private val documentId = "${user.hashCode()}${DOCUMENTID_SEPARATOR}$rootFileId"
+    private val documentId = "${DocumentsStorageProvider.rootIdForUser(user)}${DOCUMENTID_SEPARATOR}$rootFileId"
     private val uri = DocumentsContract.buildTreeDocumentUri(authority, documentId)
     private val rootDir get() = DocumentFile.fromTreeUri(context, uri)!!
 
--- a/app/src/main/java/com/nextcloud/client/account/Server.kt
+++ b/app/src/main/java/com/nextcloud/client/account/Server.kt
@@ -24,7 +24,6 @@ import android.os.Parcel
 import android.os.Parcelable
 import com.owncloud.android.lib.resources.status.OwnCloudVersion
 import java.net.URI
-import java.util.Objects
 
 /**
  * This object provides all information necessary to interact
@@ -43,10 +42,6 @@ data class Server(val uri: URI, val version: OwnCloudVersion) : Parcelable {
         writeSerializable(uri)
         writeParcelable(version, 0)
     }
-    
-    override fun hashCode() = 
-        // a server is identified by its uri, the version is an attribute of this one server
-        Objects.hash(uri)
 
     companion object {
         @JvmField
--- a/app/src/main/java/com/nextcloud/client/account/Server.kt
+++ b/app/src/main/java/com/nextcloud/client/account/Server.kt
@@ -24,6 +24,7 @@ import android.os.Parcel
 import android.os.Parcelable
 import com.owncloud.android.lib.resources.status.OwnCloudVersion
 import java.net.URI
+import java.util.Objects
 
 /**
  * This object provides all information necessary to interact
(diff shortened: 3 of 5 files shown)


AlvaroBrey (changes requested):
When overriding `hashCode`,  `equals` must be overridden as well, otherwise bad things happen. 

`equals` is also easy to do incorrectly (need to check referential equality, and class) so I suggest you just use `Alt + Insert` -> generate Equals and hashcode in Android Studio for this.

Additionally you don't need to import `Objects`,  `uri.hashCode()` works here.

AlvaroBrey:
cc @ezaquarii , you probably have a better understanding of what this change may/may not break

pelzvieh:
> When overriding `hashCode`, `equals` must be overridden as well, otherwise bad things happen.

While I agree that overriding equals makes a lot of sense, the strict requirement is the other way round: if I override equals, I must make sure that equal objects yield identical hash codes. So in a nutshell: if I override equals, I need to override hashCode.

The other way round is no strict requirement, as there's never a guarantee that unequal objects have different hash codes, code has to deal with hash collisions, only performance may suffer (shouldn't be an issue with Server objects). (*)

On the other hand, overriding would introduce a concept of equality, that currently is not required, but simply two instances of Server would be handled unequal. Changing this requires overview on how these objects are used, if the are compared anywhere, put into HashSets or the like, where introducing previously absent equality might break existing logic. That's why I did not yet dare to do this.

> `equals` is also easy to do incorrectly (need to check referential equality, and class) so I suggest you just use `Alt + Insert` -> generate Equals and hashcode in Android Studio for this.

Fully agree here. I'll be happy to do so as soon as an expert on the software states it is safe to do, and I'm glad you'll have an eye on the result ;-)

> Additionally you don't need to import `Objects`, `uri.hashCode()` works here.

Just to be sure: uri can never be null?


(*) PS: By the way, this could be a flaw (strictly a bug? but with very low probability to show up) of the concept to use hash codes as identifiers for the Android storage framework: how do you deal with hash collisions of User objects?

AlvaroBrey:
> While I agree that overriding equals makes a lot of sense, the strict requirement is the other way round: if I override equals, I must make sure that equal objects yield identical hash codes. So in a nutshell: if I override equals, I need to override hashCode.
> 
> The other way round is no strict requirement, as there's never a guarantee that unequal objects have different hash codes, code has to deal with hash collisions, only performance may suffer (shouldn't be an issue with Server objects). (*)
> 
> On the other hand, overriding would introduce a concept of equality, that currently is not required, but simply two instances of Server would be handled unequal. Changing this requires overview on how these objects are used, if the are compared anywhere, put into HashSets or the like, where introducing previously absent equality might break existing logic. That's why I did not yet dare to do this.

Fair enough, but I'd rather figure this out before merging than run into a bug 6 months from now because of `equals` and `hashCode` behaving differently than expected.

> > `equals` is also easy to do incorrectly (need to check referential equality, and class) so I suggest you just use `Alt + Insert` -> generate Equals and hashcode in Android Studio for this.
> 
> Fully agree here. I'll be happy to do so as soon as an expert on the software states it is safe to do, and I'm glad you'll have an eye on the result ;-)

Let's wait for @ezaquarii to shed some light on this.

> Just to be sure: uri can never be null?

Correct, Kotlin is null-safe and `uri` is declared as non-null.

> (*) PS: By the way, this could be a flaw (strictly a bug? but with very low probability to show up) of the concept to use hash codes as identifiers for the Android storage framework: how do you deal with hash collisions of User objects?

Yes, this makes sense and I  was planning to look into this. We should probably use `[accountname]@[server uri]` or something like that for the storage roots map instead.

ezaquarii:
> Just to be sure: uri can never be null?

It's non-nullable, so it cannot.

> (*) PS: By the way, this could be a flaw (strictly a bug? but with very low probability to show up) of the concept to use hash codes as identifiers for the Android storage framework: 

Yes, this is a design bug and should be fixed.

> how do you deal with hash collisions of User objects?

We don't use hashes for non-volatile data, but `accountName`, which is stable. If you spotted `hashCode` being persisted anywhere in the app code, please raise a bug. Using `hashCode` for volatile, runtime purposes (`HashMap`, `HashSet`, etc) is ok, because it is stable in the context of a running process.

Now, about the PR.

I don't like this fix.  Kotlin compiler generates correct `hashCode` and `equals` methods for data classes and customizing them to accommodate buggy code does not look like a right direction to me. Although the proposed `hashCode` implementation does not violate the contract, it weakens the hash quality and puts the burden of future maintenance on humans.

First, the value of `hashCode` is not guaranteed to be stable between process runs. Oracle JVM itself has few `hashCode` implementation that can be selected at startup with a `-XX:hashCode=n` switch. And they switched default one after Java8. Android can change it's `hashCode` with next OS upgrade, if they wish so. Yikes!

In simple terms, persisting `hashCode` is a design bug. If it works, it's a happy accident.

I think this should be fixed in the storage framework. It seems that the storage manager has a stable link to a particular server instance, which can be unambiguously identified by `Server.uri`. Since `uri` cannot change, I think we should switch the manager to derive stable id from `uri`. Can it be `uri.hashCode()`? It will work... most of the time, until it doesn't again, because `hashCode` implementation changes with next upgrade of OS or app.

I'd just use crypto-hash of the URI string to obtain a stable ID for storage manager and use that instead of `hashCode()`. I'd also hunt down any persistent storage of `hashCode` and start planning for a migration.

ezaquarii:
@AlvaroBrey @pelzvieh 

> Just to be sure: uri can never be null?

It's non-nullable, so it cannot.

> (*) PS: By the way, this could be a flaw (strictly a bug? but with very low probability to show up) of the concept to use hash codes as identifiers for the Android storage framework: 

Yes, this is a design bug and should be fixed.

> how do you deal with hash collisions of User objects?

We don't use hashes for non-volatile data, but `accountName`, which is stable. If you spotted `hashCode` being persisted anywhere in the app code, please raise a bug. Using `hashCode` for volatile, runtime purposes (`HashMap`, `HashSet`, etc) is ok, because it is stable in the context of a running process.

Now, about the PR.

I don't like this fix.  Kotlin compiler generates correct `hashCode` and `equals` methods for data classes and customizing them to accommodate buggy code does not look like a right direction to me. Although the proposed `hashCode` implementation does not violate the contract, it weakens the hash quality and puts the burden of future maintenance on humans.

First, the value of `hashCode` is not guaranteed to be stable between process runs. Oracle JVM itself has few `hashCode` implementation that can be selected at startup with a `-XX:hashCode=n` switch. And they switched default one after Java8. Android can change it's `hashCode` with next OS upgrade, if they wish so. Yikes!

In simple terms, persisting `hashCode` is a design bug. If it works, it's a happy accident.

I think this should be fixed in the storage framework. It seems that the storage manager has a stable link to a particular server instance, which can be unambiguously identified by `Server.uri`. Since `uri` cannot change, I think we should switch the manager to derive stable id from `uri`. Can it be `uri.hashCode()`? It will work... most of the time, until it doesn't again, because `hashCode` implementation changes with next upgrade of OS or app.

I'd just use crypto-hash of the URI string to obtain a stable ID for storage manager and use that instead of `hashCode()`. I'd also hunt down any persistent storage of `hashCode` and start planning for a migration.

ezaquarii:
If we decide that going forward with a proper fix is too much for now, converting the manager to use `server.uri.hashCode()` will apply @pelzvieh solution without compromising `Server.hashCode`.

pelzvieh:
> If we decide that going forward with a proper fix is too much for now, converting the manager to use `server.uri.hashCode()` will apply @pelzvieh solution without compromising `Server.hashCode`.

Caution, I would think it is necessary for the manager to create a unique ID from server URI _and_ login. In order to support being logged in to one server with multiple accounts.

pelzvieh:
> Also, the value of `hashCode` is not guaranteed to be stable between process runs. Oracle JVM itself has few `hashCode` implementations that can be selected at startup with a `-XX:hashCode=n` switch. They switched default one after Java8 as well. I wonder how many enterprise clunkers stopped running on JVM upgrade back then 🙃. Android can change it's `hashCode` with next OS upgrade, if they wish so. Yikes!

Is there any reason to use a hash at all? An Android root id is a String. To me it seems logical to use something like user@uri as root id. The implementation of DocumentsStorageProvider does not allow '/' to be part of the root id, but this could be coped with by url encoding user@uri.

AlvaroBrey:
> Is there any reason to use a hash at all? An Android root id is a String. To me it seems logical to use something like user@uri as root id. The implementation of DocumentsStorageProvider does not allow '/' to be part of the root id, but this could be coped with by url encoding user@uri.

The only drawback I can see is that `user@server` can be of arbitrary length, but I can't find any documentation on documentId max lengths, so I'd say we're "safe".

ezaquarii:
> The only drawback I can see is that `user@server` can be of arbitrary length

@AlvaroBrey hash it

pelzvieh:
> Looks fine. We need to do some testing to see what happens to existing app installations when this is merged.

The update itself should do less harm than does a server version change right now: the app is necessarily re-started, so no stale root ids are registered with Android's storage manager. The root ids _do_ change of course which _will_ break any persisted document IDs (like e.g. the last used database of Keepass). The same damage as now occurs with each server version update.

AlvaroBrey (approved):

app/src/main/java/com/nextcloud/client/utils/HashUtil.kt
```diff
@@ -0,0 +1,42 @@
...
+package com.nextcloud.client.utils
+
+import org.apache.commons.codec.binary.Hex
+import java.security.MessageDigest
+
+object HashUtil {
+    private const val ALGORITHM_MD5 = "MD5"
+
+    @JvmStatic
+    fun md5Hash(input: String): String {
```
tobiasKaminsky, line 37: We have some other md5 occurrences already in code.
Let us consolidate them after merge of this.

tobiasKaminsky (approved):

AlvaroBrey:
@pelzvieh Thanks for looking into this and contributing!

PR #10323, opened 2022-06-02
Disable sharing on e2e

tobiasKaminsky:

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/FileDetailTabAdapter.java
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/FileDetailTabAdapter.java
@@ -70,6 +70,10 @@ public class FileDetailTabAdapter extends FragmentStatePagerAdapter {
 
     @Override
     public int getCount() {
-        return 2;
+        if (file.isEncrypted()) {
+            return 1;
+        } else {
+            return 2;
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/owncloud/android/ui/adapter/FileDetailTabAdapter.java
```diff
@@ -70,6 +70,10 @@ public FileDetailActivitiesFragment getFileDetailActivitiesFragment() {
 
     @Override
     public int getCount() {
-        return 2;
+        if (file.isEncrypted()) {
+            return 1;
+        } else {
+            return 2;
+        }
```
AlvaroBrey, line 77: Add a comment here please, otherwise these just look like magic numbers. Something like "Sharing not allowed for encrypted files"

scripts/uploadReport.sh
```diff
@@ -54,7 +54,7 @@ fi
 if [ -e $FOLDER ]; then
     upload $FOLDER
 else
-    deleteOldComments
+    scripts/deleteOldComments.sh "$BRANCH" "$TYPE" "$PR"
```
AlvaroBrey, line 57: ?

app/src/main/java/com/owncloud/android/ui/adapter/FileDetailTabAdapter.java
tobiasKaminsky: Done

AlvaroBrey (approved):

PR #10358, opened 2022-06-09
Upload logcat on failing master-IT builds

AlvaroBrey:

### DIFF
--- a/scripts/runCombinedTest.sh
+++ b/scripts/runCombinedTest.sh
@@ -5,6 +5,17 @@ LOG_USERNAME=$2
 LOG_PASSWORD=$3
 DRONE_BUILD_NUMBER=$4
 
+function upload_logcat() {
+    set -x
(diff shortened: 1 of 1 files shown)


scripts/runCombinedTest.sh
```diff
@@ -5,6 +5,17 @@ LOG_USERNAME=$2
 LOG_PASSWORD=$3
 DRONE_BUILD_NUMBER=$4
 
+function upload_logcat() {
+    set -x
```
tobiasKaminsky, line 9: This will show all executed commands, also with potential app token.
Do we need this?

AlvaroBrey: No, leftover from testing, sorry. Fixed.

(as an aside, the secrets should always be masked in the output, regardless of how we print them, so they wouldn't leak from this)

tobiasKaminsky (approved):

PR #10367, opened 2022-06-10
move md5 function to tests, as md5 should never be used is production

tobiasKaminsky:


AlvaroBrey:
Note that md5 is also used in DocumentsStorageProvider (for root storage point ID generation) and BitmapUtils (for color generation). As these are not encryption or security usages, it should be fine.

PR #10377, opened 2022-06-13
Groupfolders cannot be encrypted

tobiasKaminsky:
added test showing folder types


app/src/androidTest/java/com/owncloud/android/ui/fragment/OCFileListFragmentStaticServerIT.kt
```diff
@@ -80,6 +81,8 @@ class OCFileListFragmentStaticServerIT : AbstractIT() {
     /**
      * Use same values as {@link FileDetailSharingFragmentIT listSharesFileAllShareTypes }
      */
+    @Test
+    @ScreenshotTest
```
AlvaroBrey, line 85: This test was disabled because it fails, IIRC. Check the failing screenshots in this PR.

scripts/androidScreenshotTest
```diff
@@ -61,6 +61,7 @@ done < <(adb devices | cut -f1)
 
 if [ "$emulatorIsRunning" == false ] ; then
     "$(command -v emulator)" -writable-system -avd uiComparison -no-snapshot -gpu swiftshader_indirect -no-audio -skin 500x833 &
+    sleep 20
```
AlvaroBrey, line 64: leftover from testing?

AlvaroBrey (changes requested):

scripts/androidScreenshotTest
tobiasKaminsky: This is is when the emulator is yet running. So it starts the emulator, waits a bit and then starts testing. 
Otherwise it might be that tests start too soon and thus they fail…

app/src/androidTest/java/com/owncloud/android/ui/fragment/OCFileListFragmentStaticServerIT.kt
tobiasKaminsky: This was due to modification date not stored correctly in db, I fixed this in this PR.

AlvaroBrey: I think the user avatar generation is broken as well

Issue #7887, opened 2021-01-23
crash on "hidden" unwanted screen


fungos:
### Steps to reproduce
1. Open NextCloud sync (without any account set)
2. In the connection screen, press Android back button two times - the timing between both presses is hard, try to sync with the left-to-right sliding animation, pressing the back button for the second time near the end of the animation.
3. If successful, a white screen will appear, looking like this:
![20210123_084627](https://user-images.githubusercontent.com/151544/105579965-90ee3400-5d57-11eb-91a0-dfc53c9ffa38.jpg)

4. In the above screen, if you press the + button, a crash occurs:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'com.owncloud.android.lib.resources.status.CapabilityBooleanType com.owncloud.android.lib.resources.status.OCCapability.getRichDocuments()' on a null object reference
	at com.owncloud.android.ui.fragment.OCFileListBottomSheetDialog.onCreate(OCFileListBottomSheetDialog.java:128)
	at android.app.Dialog.dispatchOnCreate(Dialog.java:494)
	at android.app.Dialog.show(Dialog.java:342)
	at com.owncloud.android.ui.fragment.OCFileListFragment.lambda$registerFabListener$2$OCFileListFragment(OCFileListFragment.java:442)
	at com.owncloud.android.ui.fragment.-$$Lambda$OCFileListFragment$aYkoWPoxzAOahKL5AVjT06NhcnM.onClick(Unknown Source:4)
	at android.view.View.performClick(View.java:6897)
	at android.view.View$PerformClick.run(View.java:26104)
	at android.os.Handler.handleCallback(Handler.java:789)
	at android.os.Handler.dispatchMessage(Handler.java:98)
	at android.os.Looper.loop(Looper.java:164)
	at android.app.ActivityThread.main(ActivityThread.java:6944)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:327)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1374)
```

************ APP INFORMATION ************
ID: com.nextcloud.client
Version: 30140190
Build flavor: generic

************ DEVICE INFORMATION ************
Brand: samsung
Device: heroltebmc
Model: SM-G930W8
Id: R16NW
Product: heroltebmc

************ FIRMWARE ************
SDK: 26
Release: 8.0.0
Incremental: G930W8VLS7CTB2

[...]

PR #10384, opened 2022-06-13
Check account in onResume fixing #7887

olebittner:
Hi all,

I was able reproduce bug #7887.
I implemented a check in the onResume function to restart the accountCreation, smiliar to the onRestart. Since onResume is more frequently called than onRestart, I simplified the check to just check the currentAccount variable. Also, I do not see how the Account should get deleted between pause and resume, opposing to between restarts...

I'm not sure about the test cases, I was able to test the case when currentAccount is null. But I was not able to test the other case without mocking the mixin itself, and I feel like it somehow defeats the purpose if I mock the class I want to test.

PS:  this is my first contribution to an oss project, hope I'm doing this the right way 😅 I just browsed through the approved bugs for something small, I could try myself on.

### DIFF
--- a/app/src/main/java/com/nextcloud/client/mixins/SessionMixin.kt
+++ b/app/src/main/java/com/nextcloud/client/mixins/SessionMixin.kt
@@ -133,4 +133,11 @@ class SessionMixin constructor(
         val account = accountManager.currentAccount
         setAccount(account)
     }
+
+    override fun onResume() {
+        super.onResume()
+        if (currentAccount == null) {
+            swapToDefaultAccount()
+        }
+    }
 }
--- a/app/src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
@@ -89,7 +89,9 @@ public abstract class BaseActivity extends AppCompatActivity implements Injectab
     @Override
     protected void onResume() {
         super.onResume();
-        mixinRegistry.onResume();
+        if (enableAccountHandling) {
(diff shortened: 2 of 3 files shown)


AlvaroBrey (approved):

AlvaroBrey:
I've verified that this does fix the issue. Thank you for your contribution @olebittner! 

As it stands, this points to a deeper problem on how first start is handled. Instead of launching `FileDisplayActivity` and having the mixin detect that there's no session, we should launch a separate `SplashActivity` that then chooses whether to launch `FDA` or `AuthenticatorActivity`. This is, however, outside of the scope of this PR. cc @tobiasKaminsky

AlvaroBrey:
@olebittner please fix the `ktlint` errors and we'll merge this. Just running `./gradlew ktlintFormat` should do it

olebittner:
no idea, whats going on with the tests now... They run locally and I just added two newlines.
The test that is failing is:
```
com.nextcloud.client.logger.LoggerTest > queue overflow warning is logged FAILED
    org.mockito.exceptions.verification.junit.ArgumentsAreDifferent at LoggerTest.kt:263
```

AlvaroBrey:
> no idea, whats going on with the tests now... They run locally and I just added two newlines. The test that is failing is:
> 
> ```
> com.nextcloud.client.logger.LoggerTest > queue overflow warning is logged FAILED
>     org.mockito.exceptions.verification.junit.ArgumentsAreDifferent at LoggerTest.kt:263
> ```

Don't worry, we have some flaky tests unfortunately. Thanks for your contribution!

PR #10392, opened 2022-06-14
on favorite/recent search: update encrypted files

tobiasKaminsky:
- create encrypted folder
- add a file in it
- favorite this file
- clear cache
- do favorite search
--> see that file is correctly shown with decrypted file name


AlvaroBrey (approved):
This is going to be fairly slow with loads of files. At some point we should optimize `parseVirtuals`, move it out of the adapter, and make it background.

Issue #10338, opened 2022-06-04
app crash on browsing folder with a lot of photos


vladbejenaru:
### Cause of error
```java
Exception in thread "main" java.util.concurrent.RejectedExecutionException: Task android.os.AsyncTask$3@fb5b3d8 rejected from java.util.concurrent.ThreadPoolExecutor@e399331[Running, pool size = 17, active threads = 17, queued tasks = 128, completed tasks = 26]
    at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2078)
    at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:843)
    at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1389)
    at android.os.AsyncTask.executeOnExecutor(Unknown Source:34)
    at com.owncloud.android.utils.DisplayUtils.setThumbnail(DisplayUtils.java:952)
    at com.owncloud.android.ui.adapter.OCFileListDelegate.bindGridViewHolder(OCFileListDelegate.kt:92)
    at com.owncloud.android.ui.adapter.OCFileListAdapter.onBindViewHolder(OCFileListAdapter.java:366)
    at androidx.recyclerview.widget.RecyclerView$Adapter.onBindViewHolder(RecyclerView.java:7254)
    at androidx.recyclerview.widget.RecyclerView$Adapter.bindViewHolder(RecyclerView.java:7337)
    at androidx.recyclerview.widget.RecyclerView$Recycler.tryBindViewHolderByDeadline(RecyclerView.java:6194)
    at androidx.recyclerview.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline(RecyclerView.java:6460)
    at androidx.recyclerview.widget.GapWorker.prefetchPositionWithDeadline(GapWorker.java:288)
    at androidx.recyclerview.widget.GapWorker.flushTaskWithDeadline(GapWorker.java:345)
    at androidx.recyclerview.widget.GapWorker.flushTasksWithDeadline(GapWorker.java:361)
    at androidx.recyclerview.widget.GapWorker.prefetch(GapWorker.java:368)
    at androidx.recyclerview.widget.GapWorker.run(GapWorker.java:399)
    at android.os.Handler.handleCallback(Unknown Source:2)
    at android.os.Handler.dispatchMessage(Unknown Source:4)
[...]

PR #10394, opened 2022-06-15
Catch RejectedExecutionException

tobiasKaminsky:
Needs feedback from reporting customer.


AlvaroBrey:
#10338

AlvaroBrey:
Additionally you may want to use a separate ThreadPoolExecutor rather than the default one, so that we have a bit more leeway. In the end, we should get rid of AsyncTasks and do it differently, but this is OK for now.

Issue #9328, opened 2021-11-20
Regression: app creates invalid folder when moving files to app folder


svenb1234:
### Steps to reproduce
1. Update android app to 3.18
2. Let auto upload upload and move files to app folder
3. 

### Expected behaviour
- pictures should be moved to the app's folder (android/media.... At least before the update)
- pictures should show up in Google's photos app.

### Actual behaviour
- The app creates a folder which shows up as invalid
- photos are not found/shown by the photos app

### Environment data
Android version: 12

Device model: pixel 4a

Stock or customized system: stock

Nextcloud app version: 3.18.0

Nextcloud server version: 21.0.6

Reverse proxy: no

![Screenshot_20211120-125904](https://user-images.githubusercontent.com/46424536/142725710-b5fc14f1-4ad4-450e-af67-ad8bd74e6c15.png)

Bob-le-pirate:
Dear @svenb1234, you are not alone. I am facing the same issue.
Downgrading to previous version 3.17.1 solves it temporary.

simonspa:
Same here, my gallery app also shows empty thumbnails for pictures the Nextcloud app has moved. 

![Screenshot_20211125-190338](https://user-images.githubusercontent.com/1677436/143486240-be18d51c-18dc-432c-b7da-33e2d6c90170.png)

ksarnelli:
Same issue here.  It also happens when downloading files (only tested with images) from cloud to local via the app.  The empty folders created at the bottom of the "(invalid)" directory tree are causing the real files to not be found by other picture apps.

redoranges:
I'm seeing the same thing

Android version: 11

Device model: Motorola one 5G ace

Stock or customized system: stock

Nextcloud app version: 3.18.0

Nextcloud server version: 21.0.3

Reverse proxy: yes

simonspa:
@AlvaroBrey is there a way we can help to debug this and find the root cause? I'd love to have that fixed - maybe even in 3.18.1 :innocent: since my default use case (auto-upload photos and move to Nextcloud app folder) is terribly broken - only empty images are displayed by gallery apps...

svenb1234:
IMHO the fix should not only fix it for future uploads but also fix it for those uploads 3.18 messed up. I.e. it should move the pictures to the correct folders and remove the invalid folder.

simonspa:
[...]

PR #10400, opened 2022-06-16
Migrations: add migration to remove (invalid) media folder created by incorrect scan

AlvaroBrey:
This folder was created in version 3.18.1  - 3.20.2 (inclusive) when moving media files to private storage.

This is caused by MediaScanner not being able to access the private storage, and instead of crashing it creates
the (invalid) folder, with folders named the same as the original pictures.

Note: this will only work if the app has been granted storage permission, and the fake images won't disappear from
gallery apps / mediaStore until next reboot or media scan.

More info: #9328

### DIFF
--- a/app/src/main/java/com/nextcloud/client/migrations/Migrations.kt
+++ b/app/src/main/java/com/nextcloud/client/migrations/Migrations.kt
@@ -19,25 +19,35 @@
  */
 package com.nextcloud.client.migrations
 
+import android.content.Context
 import android.os.Build
+import android.os.Environment
+import android.provider.MediaStore
 import androidx.work.WorkManager
 import com.nextcloud.client.account.UserAccountManager
 import com.nextcloud.client.jobs.BackgroundJobManager
 import com.nextcloud.client.logger.Logger
 import com.owncloud.android.datamodel.ArbitraryDataProvider
 import com.owncloud.android.ui.activity.ContactsPreferenceActivity
+import com.owncloud.android.utils.PermissionUtil
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.Paths
 import javax.inject.Inject
+import kotlin.streams.toList
 
 /**
  * This class collects all migration steps and provides API to supply those
  * steps to [MigrationsManager] for execution.
  */
+@Suppress("LongParameterList")
 class Migrations @Inject constructor(
     private val logger: Logger,
     private val userAccountManager: UserAccountManager,
     private val workManager: WorkManager,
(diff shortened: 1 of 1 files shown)


AlvaroBrey:
@tobiasKaminsky please give me your thoughts on whether we should do this or not. I have my doubts, reasoning at https://github.com/nextcloud/android/issues/9328#issuecomment-1157430516

@ezaquarii Can you give me some feedback as to whether this makes sense as a Migration, and whether it should be mandatory or not? I'm not that familiar with the `Migrations` model

ezaquarii:
@AlvaroBrey Application of migrations API looks valid to me in this case, as you want to perform one-time change on application upgrade. 

One thing that stands out for me is the need of permissions. I guess that the user must have provided this permission prior to creation of the bad media folder, because migration will be applied as no-op otherwise (i.e. marked as applied but no change done). I'm not sure if this is intended behaviour.

If the folder deletion is to be decided by the user, perhaps some other mechanism should be used.

ezaquarii:
@AlvaroBrey If you'd like to re-try a migation after failure, this can be done using optional migrations. Check out this test case: https://github.com/nextcloud/android/blob/fbfa7fa746a509be165219d960030cf4a0967eea/app/src/androidTest/java/com/nextcloud/client/migrations/MigrationsManagerTest.kt#L262

AlvaroBrey:
> One thing that stands out for me is the need of permissions. I guess that the user must have provided this permission prior to creation of the bad media folder, because migration will be applied as no-op otherwise (i.e. marked as applied but no change done). I'm not sure if this is intended behaviour.

The invalid folder is only created when uploading a file with "move to Nextcloud folder" selected as an after-upload behaviour; this is only possible with external storage permissions.

The only way this would not work is if the users actively remove the permission after uploading things and before upgrading to the version that contains this migration, but that seems narrow enough that we don't need to handle it.

AlvaroBrey:
@tobiasKaminsky opinion on this?

BEEDELLROKEJULIANLOCKHART:
@AlvaroBrey, potentially send a notification if the storage permission is not provided, and state why? Most software that is for Android that I know of does this.

AlvaroBrey:
> @AlvaroBrey, potentially send a notification if the storage permission is not provided, and state why? Most software that is for Android that I know of does this.

Not a good idea, as we need the permission to even check the invalid folder exists; so we would be showing that notification to every person without storage permission, even if they don't have the invalid folder.  And as stated above, it's much more likely that people with the invalid folder already have the permission, as it can't be created without it.

PR #10414, opened 2022-06-20
Test daily screenshot test on github actions

tobiasKaminsky:

### DIFF
--- a/.github/workflows/dailyScreenshotTest.yml
+++ b/.github/workflows/dailyScreenshotTest.yml
@@ -1,0 +1,86 @@
+name: "Screenshot Test"
+
+on:
+    push:
+    pull_request:
+    schedule:
+        -   cron: '0 0 * * *'  # every day at midnight
+        branches: [ master, stable-* ]
+
+jobs:
+    screenshot:
+        runs-on: macOS-latest
+        strategy:
+            fail-fast: false
+            matrix:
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Let's see if this runs also when only in this branch.

.github/workflows/dailyScreenshotTest.yml
```diff
@@ -0,0 +1,86 @@
+name: "Screenshot Test"
+
+on:
+    push:
+    pull_request:
+    schedule:
+        -   cron: '0 0 * * *'  # every day at midnight
+        branches: [ master, stable-* ]
```
AlvaroBrey, line 8:
```suggestion
on:
    push:
        branches: [ master, stable-* ]
    pull_request:
        branches: [ master, stable-* ]
    schedule:
        -   cron: '0 0 * * *'  # every day at midnight
```

AlvaroBrey (changes requested):
YAML is invalid, see suggestion below.

Additionally, I assume the plan is to remove the `push` and  `pull_request` events before merging and leave only the cron right? We'll need to remove the uploadReport step and other comment stuff if applicable.

tobiasKaminsky:
> I assume the plan is to remove the `push` and `pull_request` events before merging and leave only the cron right?

Yes, this is only for testing.

PR #10431, opened 2022-06-23
Implemented google in-app review.

surinder-tsys:
To provide rating and review option inside the app we are using [Google Play In-App Review](https://developer.android.com/guide/playcore/in-app-review) SDK to do so. 

We are showing the in-app review flow in the following conditions:
a.	Minimum 10 app restarts should be there
b.	Maximum 15 times a year

Inkedin-app-review

There are some conditions from google play which can possibly not show the review dialog:
a.	User has already done the review may be from app or from directly in play store.
b.	Once review dialog is shown it will show again too frequently.
c.	Primary account should be selected in play store.

References:
a.	https://developer.android.com/guide/playcore/in-app-review/test#troubleshooting
b.	https://stackoverflow.com/a/63342266

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -348,6 +348,9 @@ dependencies {
 
     // upon each update first test: new registration, receive push
     gplayImplementation "com.google.firebase:firebase-messaging:23.0.5"
+
+    // Play In-App Review
+    implementation 'com.google.android.play:review-ktx:2.0.0'
 }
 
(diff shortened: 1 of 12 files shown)


surinder-tsys:
No longer valid. Will be tracked here: https://github.com/nextcloud/android/pull/11541 after new merging process.
Closing the PR.

PR #10436, opened 2022-06-24
Remove unused script 'getBranchName.sh'

timkrueger:
If somewhere needed maybe better use

  git branch --show-current

since Git 2.22 (Q2 2019).

Signed-off-by: Tim Krüger <t@timkrueger.me>


tobiasKaminsky:
We need the base name, and this is only available via this script.

PR #10441, opened 2022-06-27
Remove bytebuddy library dependency

AlvaroBrey:
Only used to generate random strings for tests and creates too many Dependabot PRs.
Let's just use Kotlin for that.


AlvaroBrey:
Note to self: this kind of test util classes is a good candidate for nextcloud-android-common (in a separate test library)

tobiasKaminsky (approved):

PR #10443, opened 2022-06-27
File export / download

tobiasKaminsky:


app/src/main/java/com/nextcloud/client/jobs/FilesDownloadWork.kt
```diff
@@ -0,0 +1,198 @@
...
+import com.owncloud.android.operations.DownloadType
+import com.owncloud.android.ui.dialog.SendShareDialog
+import com.owncloud.android.ui.fragment.OCFileListFragment
+import com.owncloud.android.ui.notifications.NotificationUtils
+import com.owncloud.android.utils.FileExportUtils
+import com.owncloud.android.utils.FileStorageUtils
+import com.owncloud.android.utils.theme.ThemeColorUtils
+import java.security.SecureRandom
+
+class FilesDownloadWork(
```
AlvaroBrey, line 53: Should it be `FilesExportWork` instead?

app/src/main/java/com/owncloud/android/utils/FileExportUtils.kt
```diff
@@ -0,0 +1,156 @@
...
+class FileExportUtils {
+    companion object {
+        const val INITIAL_RENAME_COUNT = 2
+    }
+
+    @Throws(RuntimeException::class)
+    fun exportFile(
+        fileName: String,
+        mimeType: String,
+        contentResolver: ContentResolver,
```
AlvaroBrey, line 47: If possible I'd prefer to see the `contentResolver` injected in `FileExportUtils` via constructor, and then `FileExportUtils` injected wherever needed, not sure how hard that is.

app/src/main/java/com/owncloud/android/utils/FileExportUtils.kt
```diff
@@ -0,0 +1,156 @@
...
+                } while (uri == null)
+            }
+
+            try {
+                val inputStream = if (ocFile != null) {
+                    contentResolver.openInputStream(ocFile.storageUri)
+                } else if (file != null) {
+                    FileInputStream(file)
+                } else {
+                    throw RuntimeException("ocFile and file both may not be null")
```
AlvaroBrey, line 77: Throw an `IllegalStateException` instead, for example.

app/src/main/java/com/owncloud/android/utils/FileExportUtils.kt
```diff
@@ -0,0 +1,156 @@
...
+                }
+
+                val pfd = contentResolver.openFileDescriptor(uri, "w")
+                val outputStream = FileOutputStream(pfd!!.fileDescriptor)
+                IOUtils.copy(inputStream, outputStream)
+            } catch (e: IOException) {
+                e.printStackTrace()
```
AlvaroBrey, line 84: Use `Log_OC`, do not print stack trace. Same for other occurrences in this file

app/src/main/java/com/owncloud/android/utils/FileExportUtils.kt
```diff
@@ -0,0 +1,156 @@
...
+
+                inputStream.use { fis ->
+                    outputStream.use { os ->
+                        val buffer = ByteArray(1024)
+                        var len: Int
+                        while (fis!!.read(buffer).also { len = it } != -1) {
+                            os.write(buffer, 0, len)
+                        }
+                    }
+                }
```
AlvaroBrey, line 127: Why the two different ways of copying an `InputStream` to an `OutputStream` in the two if branches? Should be the same (`IOUtils.copy` for both, for example)

app/src/main/java/com/owncloud/android/utils/FileExportUtils.kt
```diff
@@ -0,0 +1,156 @@
...
+            } catch (e: FileNotFoundException) {
+                e.printStackTrace()
+            } catch (e: IOException) {
+                e.printStackTrace()
+            } finally {
+                outputStream?.close()
+                inputStream?.close()
```
AlvaroBrey, line 134: Already using `use`, which will autoclose the resources; no need for this `finally`. 

Additionally, the streams in the other if branch are never closed. Same `use` or try-with-resources should be  in place.

app/src/main/java/com/owncloud/android/utils/FileExportUtils.kt
```diff
@@ -0,0 +1,156 @@
...
+        if (extPos >= 0) {
+            extension = name.substring(extPos + 1)
+            nameWithoutExtension = name.substring(0, extPos)
+        }
+
+        return if (extPos >= 0) {
+            "$nameWithoutExtension$suffix.$extension"
+        } else {
+            name + suffix
+        }
```
AlvaroBrey, line 154:
```suggestion
        val extPos = name.lastIndexOf('.')
        val suffix = " ($count)"

        return if (extPos >= 0) {
            val extension = name.substring(extPos + 1)
            val nameWithoutExtension = name.substring(0, extPos)
            "$nameWithoutExtension$suffix.$extension"
        } else {
            name + suffix
        }
```
Can be simplified and made immutable by merging the `if`s

app/src/main/res/values/strings.xml
```diff
@@ -1014,4 +1018,10 @@
     <string name="locked_by">Locked by %1$s</string>
     <string name="locked_by_app">Locked by %1$s app</string>
     <string name="lock_expiration_info">Expires: %1$s</string>
+    <string name="filedetails_export">Export</string>
+    <string name="locate_folder">Locate folder</string>
+    <string name="successful_download">Downloaded %1s</string>
+    <string name="download_failed">Failed to download %1s</string>
+    <string name="download_partially_failed">Downloaded %1s, skipped %2s due to error</string>
+    <string name="export_start">%1s will be downloaded. See notification for details.</string>
```
AlvaroBrey, line 1026: Mixed language: using "export" and "download" to mean the same thing. I would use "export" to not confuse users because of the current Download function, which just downloads to internal storage.

AlvaroBrey (changes requested):
Generally looks fine but needs some polish, especially `FileExportUtils`.

Additionally this PR should be squashed or rebased when merged, as many WIP commits are in this branch

app/src/main/java/com/owncloud/android/utils/FileExportUtils.kt
tobiasKaminsky: Android 7 does not work with IOUtils, so I had to do it in a different way.
Basically like in UploadFileOperation -> copy()

app/src/main/res/values/strings.xml
tobiasKaminsky: in iOS Files it is called:
download -> set as available offline (which fits also in our case)
export -> save to photo gallery (for images, as iOS cannot export other file types)

With this in mind, I mixed it up… :S 

For now, to not confuse users, we could use
download -> set as available offline
export -> export

Then I can also rename all new functions / classes to "export"?

app/src/main/java/com/owncloud/android/utils/FileExportUtils.kt
AlvaroBrey: Then change the copy in the other branch (it uses `IOUtils`)

app/src/main/res/values/strings.xml
AlvaroBrey: I would keep "Download" as it is right now, and just rename export stuff to "Export"; just be careful that no strings related to exporting files say "download".

I think it might make sense to change "download" to "set as available offline" in the future, but that's a different PR.

tobiasKaminsky:
> Additionally this PR should be squashed or rebased when merged, as many WIP commits are in this branch

Of course, I do this at the end, or sometimes before creating PR, but I will make sure to not commit it in this state.

app/src/main/java/com/owncloud/android/utils/FileExportUtils.kt
tobiasKaminsky: I failed doing it 😇
We can do it later, I think.

app/src/main/java/com/nextcloud/client/jobs/BackgroundJobFactory.kt
```diff
@@ -100,11 +100,25 @@ class BackgroundJobFactory @Inject constructor(
                 AccountRemovalWork::class -> createAccountRemovalWork(context, workerParameters)
                 CalendarBackupWork::class -> createCalendarBackupWork(context, workerParameters)
                 CalendarImportWork::class -> createCalendarImportWork(context, workerParameters)
+                FilesExportWork::class -> createFilesDownloadWork(context, workerParameters)
                 else -> null // caller falls back to default factory
             }
         }
     }
 
+    private fun createFilesDownloadWork(
```
AlvaroBrey, line 109: should be `createFilesExportWork`

app/src/main/java/com/nextcloud/client/jobs/FilesExportWork.kt
```diff
@@ -0,0 +1,173 @@
...
+    private val themeColorUtils: ThemeColorUtils,
+    params: WorkerParameters
+) : Worker(appContext, params) {
+    companion object {
+        const val FILES_TO_DOWNLOAD = "files_to_download"
+        private const val NUMERIC_NOTIFICATION_ID = "NUMERIC_NOTIFICATION_ID"
+    }
+
+    override fun doWork(): Result {
+        val fileIDs = inputData.getLongArray(FILES_TO_DOWNLOAD) ?: LongArray(0)
```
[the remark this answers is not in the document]
tobiasKaminsky: I cannot really imagine how this can happen, but logging is better than nothing.

app/src/main/java/com/nextcloud/client/jobs/FilesExportWork.kt
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    private fun showSuccessNotification(successfulExports: Int) {
+        val files = appContext.resources.getQuantityString(R.plurals.files, successfulExports, successfulExports)
+        showNotification(
+            appContext.getString(
+                R.string.export_successful,
+                files
+            )
+        )
```
AlvaroBrey, line 147: This way of doing plurals doesn't work for many languages such as Spanish. The whole string (`R.string.export_successful`) needs to be pluralized, as the verb tense changes as well (`exportado` for 1,  `exportados` for more than 1).

AlvaroBrey (changes requested):

AlvaroBrey (approved):

PR #10447, opened 2022-06-28
Fix QA upload

tobiasKaminsky:


AlvaroBrey (changes requested):
`GIT_TOKEN` is our manually added bot token. `GITHUB_TOKEN` is the automatic token from GH actions, which should be preferred everywhere.

If something's not working with `GITHUB_TOKEN` then it's likely that the permissions are not right or the environment variable is wrong.

AlvaroBrey (approved):

PR #10448, opened 2022-06-28
Avoid checking for walled connection until necessary

AlvaroBrey:
Both when syncing offline files and when retrying failed uploads, don't check for walled internet until the moment it's needed, thus avoiding network calls in cases where it's not needed (missing files, or other conditions not met)

This should reduce load on server, data consumption, and battery drain in such cases.

This is a successor to #10087 with a slightly different approach, thanks @obel1x for your findings and ideas.

In the future we might consider removing this check completely as it is a source of problems, and just handle the errors when trying to do operations in a walled network.

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/OfflineSyncWork.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/OfflineSyncWork.kt
@@ -57,7 +57,7 @@ class OfflineSyncWork constructor(
 
     override fun doWork(): Result {
         val wakeLock: WakeLock? = null
-        if (!powerManagementService.isPowerSavingEnabled && !connectivityService.isInternetWalled) {
+        if (!powerManagementService.isPowerSavingEnabled) {
             val users = userAccountManager.allUsers
             for (user in users) {
                 val storageManager = FileDataStorageManager(user, contentResolver)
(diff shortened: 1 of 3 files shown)


AlvaroBrey:
> Does your pull consider these deleted files to not cause pings any more?

Yes, FILE_NOT_FOUND is handled separately and will not cause a ping.

AlvaroBrey:
> > Does your pull consider these deleted files to not cause pings any more?
> 
> Yes, FILE_NOT_FOUND is handled separately and will not cause a ping.

Actually, not that sure, because this PR only changes things offline sync, not for autoupload; that should be changed separately

AlvaroBrey:
> > > Does your pull consider these deleted files to not cause pings any more?
> > 
> > 
> > Yes, FILE_NOT_FOUND is handled separately and will not cause a ping.
> 
> Actually, not that sure, because this PR only changes things offline sync, not for autoupload; that should be changed separately

Handled in https://github.com/nextcloud/android/pull/10448/commits/c9be796e1c90dfad64c64d378d7921c790e716fa. Will reword this PR to make this explicit.

obel1x:
tested it: made a picture in offline-mode. picture folder is set to upload only if wlan is free/available. Got connection error shown as message and in upload queue.
Maybe something not working with that setting now? imo it should not try to upload and not giving errors.

AlvaroBrey:
> tested it: made a picture in offline-mode. picture folder is set to upload only if wlan is free/available. Got connection error shown as message and in upload queue. Maybe something not working with that setting now? imo it should not try to upload and not giving errors.

That is also the behaviour in master, so it's not something introduced in this PR

tobiasKaminsky (approved):

PR #10474, opened 2022-07-04
New screenshot for com.owncloud.android.ui.dialog.SetupEncryptionDialogFragmentIT_error

tobiasKaminsky:


tobiasKaminsky:
Seems to happen randomly, so this is not a proper fix :S

PR #10490, opened 2022-07-07
add broadcastReceiver when switching from media -> on device

tobiasKaminsky:


AlvaroBrey (approved):
the receiver should likely be in the fragment instead, but that's for another PR

PR #10504, opened 2022-07-11
Add @RunWith(AndroidJUnit4.class) to test classes

tobiasKaminsky:


AlvaroBrey:
Upon reading some more docs, it looks like this is not necessarily needed:
```
// @RunWith is required only if you use a mix of JUnit3 and JUnit4.
```
Source: https://developer.android.com/training/testing/instrumented-tests#create-instrumented

AlvaroBrey:
This also breaks screenshot tests as they're no longer using the ScreenshotTestRunner

tobiasKaminsky:
> Upon reading some more docs, it looks like this is not necessarily needed:

Hm. okay :)
Then our missing/not complete logcat is caused by something else…

PR #10522, opened 2022-07-13
Don't allow focus on empty view and restore mis-updated screenshot

AlvaroBrey:
`NestedScrollView` requests focus, unlike the previous `LinearLayout`. Also it doesn't respect the `focusable=true` attribute due to interactions with `Recyclerview`, so it's necessary to explicitly set it in code.

Follow up to #10497


tobiasKaminsky:
Do we then have to set this to true again at some point?

AlvaroBrey:
> Do we then have to set this to true again at some point?

No, AFAIK we don't need the empty view to be focusable (and it wasn't before, when it was just a listview)

tobiasKaminsky (approved):

PR #10528, opened 2022-07-15
Use issue form for bug reports

### LINKED ISSUE #9782: Transform issue templates into forms
See https://github.com/nextcloud/server/pull/30778 for an idea of what it would look like. Better usability for reporters, and more standardized info for us.

As discussed in the above PR, special care should be taken as to which fields we make mandatory. If they're too many, we'll discourage reporting, which we don't want.

Another thing to consider is how this would interfere with auto-reporting from the app.

AlvaroBrey:
This enforces filling certain fields so that we don't end up with more empty issues

Heavily inspired by the templates and config in https://github.com/nextcloud/server

Fixes #9782

Check https://github.com/AlvaroBrey/issue-form-test/issues/new?assignees=&labels=bug&template=bug_report.yml for a demo, feel free to submit issues to see the results.


tobiasKaminsky:
Do we need "reverse proxy" question? Is this causing that many problems lately?
Apart from "search on repo", we could use the same for Android Talk and thus put it in android-common?

tobiasKaminsky (approved):

PR #10560, opened 2022-07-26
Fix upload conflict in encrypted folder

tobiasKaminsky:


AlvaroBrey:
@tobiasKaminsky I'm not sure how to resolve the conflict in UploadFileOperation

app/src/main/java/com/owncloud/android/ui/activity/ConflictsResolveActivity.java
```diff
@@ -141,6 +142,9 @@ protected void onCreate(Bundle savedInstanceState) {
                     uploadsStorageManager.removeUpload(upload);
                     break;
                 case KEEP_SERVER: // Download
+                    if (newFile.isEncrypted()) {
+                        return;
```
surinder-tsys, line 146: Instead of return it should be break so that finish() function on line number 161 has to execute else the activity won't finish at all.

PR #10636, opened 2022-08-17
Persist contacts/calendar backup preferences

AlvaroBrey:
Needs #10609

- [x] Make the backup worker respect these settings
- [x] Fix hiding/unhiding "backup now" button

### DIFF
--- a/app/src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
+++ b/app/src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
@@ -24,7 +24,6 @@ package com.nextcloud.client.preferences;
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.SharedPreferences;
-import android.net.Uri;
 
 import com.nextcloud.client.account.CurrentAccountProvider;
 import com.nextcloud.client.account.User;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
```diff
@@ -708,6 +716,30 @@ public void setStoragePermissionRequested(boolean value) {
         preferences.edit().putBoolean(PREF__STORAGE_PERMISSION_REQUESTED, value).apply();
     }
 
+    @Override
+    public boolean isContactsBackupEnabled(final User user) {
+        final String key = getUserScopedKey(user, PREF__CONTACTS_BACKUP_ENABLED);
```
tobiasKaminsky, line 721: For this we wave ArbitraryDataProvider, which stores values based on key and user/account.
Might be better?

AlvaroBrey: `ArbitraryDataProvider` writes/reads from database, which is inefficient and also a StrictMode violation if called on main thread. It also lets its clients use random keys instead of well-defined methods like AppPreferences, which is IMO not great.
I don't really know if using AppPreferences for this is the best idea (it would probably be a better idea to introduce a new thing, something like `UserPreferences` with a factory).

That being said, I'm okay with using `ArbitraryDataProvider` for this case if you think it's better, for the sake of not splitting user-scoped data somewhere else.

AlvaroBrey: In the interest of not introducing a new design for this in an unrelated PR, i've changed it to use ArbitraryDataProvider in https://github.com/nextcloud/android/pull/10636/commits/68c7eddf46b8d174f8956256a85642e9f9523868.

That being said I still believe that a database is not a good place for user preferences and we should do something about that at some point.

tobiasKaminsky: Totally agree. 
DataStore sounds like a way to centralize all those different things in a sane way.

tobiasKaminsky (approved):

PR #10644, opened 2022-08-18
⚗️ PoC for Material 3️⃣ theming 🎨

### LINKED ISSUE #10641: 🎨 Check if `targetSdk=31` is needed for Material 3️⃣
check if `targetSdk=31` is needed to run Material 3 due to potential theming requirements with `com.google.android.material:material:1.6.1` which is already in place but likely not in use from a theming perspective.

AndyScherzinger:
Resolves #10641

### DIFF
--- a/app/src/main/res/layout/file_details_fragment.xml
+++ b/app/src/main/res/layout/file_details_fragment.xml
@@ -185,7 +185,6 @@
         android:background="@color/bg_default"
         app:tabGravity="center"
         app:tabMode="fixed"
-        app:tabTextAppearance="@style/AppTabTextAppearance"
         app:tabTextColor="@color/text_color"
         app:tabInlineLabel="true" />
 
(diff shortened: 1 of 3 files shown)


AndyScherzinger:
@AlvaroBrey I tested if there might be crashes when using material 3 from the 1.6.x lib which needs sdk=31 while we have that for compile but not for target. PR shows this seems to work just fine. 
So I think this box can be ticked and the PR can be closed and the changes to be moved to a (larger) theming PR because currently the server color still works but many things aren't m3 themed yet so no matter the server color, dialogs have the surface blue color due to the xml defined primary color (this is handled in our theming implementation from talk already, so all good.

Aside from this I am happy to lend a hand if you want to collaborate on a feature branch for the theming of files or wanna have a chat about it.

AlvaroBrey:
We bumped the targetSDK to 31 in https://github.com/nextcloud/android/pull/10545 anyway, so yeah, that issue can be closed

AndyScherzinger:
@AlvaroBrey true, while this PR would also give you the first step to adapt the general app theme to Material3 ;)

PR #10688, opened 2022-08-28
Avoid full folder refresh during onBrowseUp

starypatyk:
I have noticed that during going up in the folder hierarchy the folder gets refreshed longer than expected.

It turns out that during execution of **com.owncloud.android.ui.fragment.OCFileListFragment.onBrowseUp**, a full refresh is requested (with **ignoreETag** set to **true**).

Is this the expected behaviour?

If not - this PR is an attempt to fix it, but I am not sure if this is a correct way (in current code value of **ignoreETag** is lost between **ExtendedListFragment.onRefresh** and **FileDisplayActivity.onRefresh**).

Thoughts?


tobiasKaminsky:
I think this is fine.
Thanks for finding this!

tobiasKaminsky (approved):

AlvaroBrey:
Thanks @starypatyk !

Issue #10629, opened 2022-08-16
Add dashboard widgets to mobile


tobiasKaminsky:
Ref: https://github.com/nextcloud/files-clients/issues/17

PR #10719, opened 2022-09-07
Dashboard

tobiasKaminsky:
Fix https://github.com/nextcloud/android/issues/10629

Requires https://github.com/nextcloud/android-library/pull/972

TODO
- [x] if returned entry size is 14, then show "load more" button, else not
- [x] use new background/style from MD3 -> postponed


nickvergessen:
> if returned entry size is 14, then show "load more" button, else not

So should the dashboards register the button or not?
If the dashboards do, you shouldn't be able to identify/distinguish it from any other button, so I think the registration condition should be checked by the app rather than the clients?

nickvergessen:
Problem | Question
---|---
Screenshot_2022-09-19-13-07-49-434_com nextcloud android qa | The icons you retrieve from the server are supposed to always be black I guess. So you can invert them for bright mode? But I see user status is always returning a white icon, so I guess it is broken there in bright mode?

nickvergessen:
More issues:

1. The widgets icon is rounded (cutting of the tick in the bottom right making Talk look like a circle)
2. The widget item icons are not rounded (avatars should be round)
3. If the widget item is a "small" it should be used in a suitable size (the icons are 16px iirc on the server)

Screenshot_2022-09-19-14-52-58-881_com miui home

nickvergessen:
Okay the top bar icon is fixed by clicking the refresh button, but the avatars stay "square" and small

app/src/main/java/com/nextcloud/client/widget/DashboardWidgetProvider.kt
```diff
@@ -0,0 +1,194 @@
...
+
+        for (appWidgetId in appWidgetIds) {
+            appPreferences.deleteWidget(appWidgetId)
+        }
+    }
+
+    companion object {
+        const val OPEN_INTENT = "open"
+
+        @SuppressLint("UnspecifiedImmutableFlag")
```
AlvaroBrey, line 105: Instead of suppressing the lint, set the correct mutability in the PendingIntents. I assume they can all just be `PendingIntent.FLAG_IMMUTABLE | PendingIntent.FLAG_UPDATE_CURRENT`

app/src/main/java/com/nextcloud/client/widget/DashboardWidgetProvider.kt
```diff
@@ -0,0 +1,194 @@
...
+        internal fun updateAppWidget(
```
AlvaroBrey, line 106: As outlined in another comment, I think this method does not belong here as a static function, and instead belongs in a separate class with a lot of these parameters already injected into it.

app/src/main/java/com/owncloud/android/ui/adapter/DashboardWidgetListAdapter.kt
```diff
@@ -0,0 +1,72 @@
...
+import androidx.recyclerview.widget.RecyclerView
+import com.nextcloud.android.lib.resources.dashboard.DashboardWidget
+import com.nextcloud.client.account.UserAccountManager
+import com.nextcloud.client.network.ClientFactory
+import com.nextcloud.client.widget.DashboardWidgetConfigurationInterface
+import com.owncloud.android.databinding.WidgetListItemBinding
+import com.owncloud.android.utils.theme.ThemeDrawableUtils
+
+class DashboardWidgetListAdapter(
+    val themeDrawableUtils: ThemeDrawableUtils,
```
AlvaroBrey, line 38: This will have to be changed when M3 work is merged, but we'll talk about that when that happens

app/src/main/res/values/strings.xml
```diff
@@ -1041,4 +1041,15 @@
     <string name="file_already_exists">Filename already exists</string>
     <string name="filedetails_export">Export</string>
     <string name="locate_folder">Locate folder</string>
+    <string name="appwidget_text">EXAMPLE</string>
+    <string name="add_widget">Add widget</string>
+    <string name="app_widget_description">Shows all widgets from dashboard</string>
```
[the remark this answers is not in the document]
tobiasKaminsky: True

app/src/main/res/values/strings.xml
```diff
@@ -1041,4 +1041,15 @@
...
+    <string name="icon_of_dashboard_widget">Icon of dashboard widget</string>
+    <string name="refresh_content">Refresh content</string>
+    <string name="icon_of_widget_entry">Icon of widget entry</string>
+    <string name="choose_widget">Choose widget</string>
+    <string name="reload">Reload</string>
+    <string name="change_account">Change account</string>
+    <string name="widgets_not_available"><![CDATA[Widgets are only available on >= NC25]]></string>
```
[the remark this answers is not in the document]
tobiasKaminsky: No idea. This was done by AS. I'll remove it.

app/src/androidTest/java/com/nextcloud/ui/BitmapIT.kt
```diff
@@ -0,0 +1,150 @@
...
+import com.owncloud.android.R
+import com.owncloud.android.utils.BitmapUtils
+import com.owncloud.android.utils.DisplayUtils
+import com.owncloud.android.utils.ScreenshotTest
+import org.junit.Rule
+import org.junit.Test
+
+class BitmapIT : AbstractOnServerIT() {
+    @get:Rule
+    val testActivityRule = IntentsTestRule(TestActivity::class.java, true, false)
```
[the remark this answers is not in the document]
tobiasKaminsky: I noted it down to do it in one step for files and library, so that then there is no old code left, which I could accidentally copy.

app/src/main/java/com/owncloud/android/ui/adapter/DashboardWidgetListAdapter.kt
tobiasKaminsky: Correct, let us merge this and then deal with M3 afterwards.
Or the other way around :)

app/src/main/java/com/nextcloud/client/network/ClientFactoryImpl.java
```diff
@@ -37,20 +37,20 @@
 
 import java.io.IOException;
 
-class ClientFactoryImpl implements ClientFactory {
+public class ClientFactoryImpl implements ClientFactory {
```
AlvaroBrey, line 40:
```suggestion
class ClientFactoryImpl implements ClientFactory {
```
Not needed anymore since injection works on the WidgetProvider. Same for the constructor, but GH won't let me add a suggestion there.

app/src/main/java/com/nextcloud/client/widget/DashboardWidgetProvider.kt
```diff
@@ -0,0 +1,89 @@
...
+    lateinit var appPreferences: AppPreferences
+
+    @Inject
+    lateinit var clientFactory: ClientFactory
+
+    @Inject
+    lateinit var accountProvider: CurrentAccountProvider
+
+    @Inject
+    lateinit var widgetUpdater: DashboardWidgetUpdater
```
AlvaroBrey, line 49:
```suggestion
import android.appwidget.AppWidgetManager
import android.appwidget.AppWidgetProvider
import android.content.Context
import android.content.Intent
import com.nextcloud.client.preferences.AppPreferences
import dagger.android.AndroidInjection
import javax.inject.Inject

/**
 * Manages widgets
 */
class DashboardWidgetProvider : AppWidgetProvider() {
    @Inject
    lateinit var appPreferences: AppPreferences

    @Inject
    lateinit var widgetUpdater: DashboardWidgetUpdater
```
Unused fields

app/src/main/res/values/strings.xml
```diff
@@ -1041,4 +1041,15 @@
     <string name="file_already_exists">Filename already exists</string>
     <string name="filedetails_export">Export</string>
     <string name="locate_folder">Locate folder</string>
+    <string name="app_widget_description">Shows one widget from dashboard</string>
+    <string name="icon_of_dashboard_widget">Icon of dashboard widget</string>
+    <string name="refresh_content">Refresh content</string>
+    <string name="icon_of_widget_entry">Icon of widget entry</string>
+    <string name="choose_widget">Choose widget</string>
+    <string name="reload">Reload</string>
+    <string name="widgets_not_available">Widgets are only available on >= NC25</string>
```
AlvaroBrey, line 1050:
```suggestion
    <string name="widgets_not_available">Widgets are only available on Nextcloud 25 or later</string>
```
Rewording

AlvaroBrey (changes requested):
Some suggestions. Also some conversations from the previous review are unresolved still.

app/src/main/java/com/nextcloud/client/network/ClientFactoryImpl.java
tobiasKaminsky: This is needed by BitmapIT.

AlvaroBrey: Then add @VisibleForTesting

AlvaroBrey (approved):
Need to fix detekt, but appart from that looks good!

tobiasKaminsky:
⚠ I rebased

AlvaroBrey (approved):

PR #10723, opened 2022-09-08
Brander: white primary color

tobiasKaminsky:
As you know we allow any color during branding.
I now tested white without enforced domain and to make server-url input screen reasonable nice, I had to change it like this.

My "problem" is now, that I had to change two other files, instead of simply relying on setup.xml.
Do you have any idea how to do this better?

We might need to have a better differentiation between
- primary color as background, can be any color
- primary color used as background, but text is written onto it (e.g. primary button, login screen)

In later case white can still be primary color, but then the text color must be based on primary color.

Whatever we come up with, we then shall test it with 
- NC blue
- white
- black
- random color
(logic how current colors are computed are on brander: https://github.com/nextcloud-gmbh/brander/blob/19824f733dc6d1534d521ae228ae12650eba21f9/src/Brander/Worker/AndroidJobHandler.php#L445)

🤷 @AlvaroBrey @AndyScherzinger 

### DIFF
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -22,7 +22,7 @@
     <color name="list_item_lastmod_and_filesize_text">@color/secondary_text_color</color>
     <color name="black">#000000</color>
     <color name="white">#ffffff</color>
-    <color name="white_helper_text">#B3FFFFFF</color>
+    <color name="white_helper_text">#000</color>
     <color name="text_color">#333333</color>
     <color name="drawer_text_color">@color/secondary_text_color</color>
     <color name="text_color_inverse">#ffffff</color>
--- a/app/src/main/res/values/setup.xml
+++ b/app/src/main/res/values/setup.xml
@@ -22,7 +22,7 @@
 
     <!-- URLs and flags related -->
     <bool name="show_server_url_input">true</bool>
-    <bool name="show_provider_or_own_installation">true</bool>
+    <bool name="show_provider_or_own_installation">false</bool>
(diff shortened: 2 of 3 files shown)


AlvaroBrey:
> We might need to have a better differentiation between
> 
>     * primary color as background, can be any color
> 
>     * primary color used as background, but text is written onto it (e.g. primary button, login screen)

This can be solved by material 3 on runtime ([`Scheme`](http://github.com/nextcloud/android-common/blob/05edc813295567d24ea468ac1b32c3ab1e950dfb/material-color-utilities/src/main/java/scheme/Scheme.java#L28-L28) already calculates `onPrimary` and other color variations for this kind of stuff)

We already calculate a `Scheme` from server color, but we can also just calculate a "default" scheme from `colorPrimary` and use that for the pre-login screens.

AndyScherzinger:
What Álvaro said, I also expect this to be solved with m3 in place 👍

tobiasKaminsky:
Awesome news! 
Then let us keep this open as a reminder.

AlvaroBrey:
The last time we discussed this via material color utils, we decided not to do it because the color scheme generation would alter the primary color and thus the login background color would not be the same as the normal primary color.

Since then, MCU  has published a [dynamic color scheme](https://github.com/material-foundation/material-color-utilities/blob/main/java/scheme/SchemeFidelity.java) that claims to allow the primary color to remain relatively unchanged in `primaryContainer`, but it doesn't seem to be compatible with the normal `Scheme` yet.

PR #10737, opened 2022-09-15
Upload in background

tobiasKaminsky:
- [ ] Progress in Uploads activity

### DIFF
--- a/app/src/androidTest/java/com/owncloud/android/files/services/FileUploaderIT.kt
+++ b/app/src/androidTest/java/com/owncloud/android/files/services/FileUploaderIT.kt
@@ -42,7 +42,7 @@ import junit.framework.Assert.assertTrue
 import org.junit.Before
 import org.junit.Test
 
-class FileUploaderIT : AbstractOnServerIT() {
+abstract class FileUploaderIT : AbstractOnServerIT() {
     var uploadsStorageManager: UploadsStorageManager? = null
 
     val connectivityServiceMock: ConnectivityService = object : ConnectivityService {
@@ -74,7 +74,7 @@ class FileUploaderIT : AbstractOnServerIT() {
     // disabled, flaky test
     // @Test
     // fun testKeepLocalAndOverwriteRemote() {
-    //     val file = getDummyFile("/chunkedFile.txt")
+    //     val file = getDummyFile("chunkedFile.txt")
     //     val ocUpload = OCUpload(file.absolutePath, "/testFile.txt", account.name)
     //
     //     assertTrue(
@@ -101,7 +101,7 @@ class FileUploaderIT : AbstractOnServerIT() {
     //
     //     assertEquals(file.length(), (result.data[0] as RemoteFile).length)
     //
-    //     val ocUpload2 = OCUpload(getDummyFile("/empty.txt").absolutePath, "/testFile.txt", account.name)
+    //     val ocUpload2 = OCUpload(getDummyFile("empty.txt").absolutePath, "/testFile.txt", account.name)
     //
     //     assertTrue(
     //         UploadFileOperation(
@@ -132,7 +132,7 @@ class FileUploaderIT : AbstractOnServerIT() {
      */
     @Test
     fun testKeepLocalAndOverwriteRemoteStatic() {
-        val file = getDummyFile("/chunkedFile.txt")
+        val file = getDummyFile("chunkedFile.txt")
 
         FileUploader.uploadNewFile(
             targetContext,
@@ -156,7 +156,7 @@ class FileUploaderIT : AbstractOnServerIT() {
         assertEquals(file.length(), (result.data[0] as RemoteFile).length)
 
         val ocFile2 = OCFile("/testFile.txt")
-        ocFile2.storagePath = getDummyFile("/empty.txt").absolutePath
+        ocFile2.storagePath = getDummyFile("empty.txt").absolutePath
 
         FileUploader.uploadUpdateFile(
             targetContext,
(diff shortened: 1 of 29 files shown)


app/src/main/java/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -1118,9 +1092,13 @@ public static String getUploadFinishMessage() {
     }
 
 
+    private static boolean useFilesUploadWorker(Context context) {
```
tobiasKaminsky, line 1095: @AlvaroBrey this leads to all tests run only with new approach (in FileUploaderIT).
Do you have an idea how we can have those tests run on both scenarios, old and new upload mechanism?

AlvaroBrey: Something like this:
```java
FileUploader.java
---
private static boolean forceLegacyUploads = false;
---
@VisibleForTesting
public static void setForceLegacyUploads(final Boolean value){
    this.forceLegacyUploads = value;
}
---
private static boolean useFilesUploadWorker(Context context) {
     if (forceLegacyUploads) return false;
     return Build.VERSION.SDK_INT >= Build.VERSION_CODES.S || context.getResources().getBoolean(R.bool.is_beta);
}
```

And just call `forceLegacyUploads(true)` before tests where you want to use the legacy service.  If this was not all static I could come up with cleaner methods, but this is the best I can think of right now.

tobiasKaminsky: Then I would have to run all the tests twice, but just add this before.
Similar to FileDataStorageManagerIT

Thanks!

tobiasKaminsky:
Let's see what CI says :)

tobiasKaminsky:
@AlvaroBrey spotlessKotlinCheck fails, but auto format in AS is fine with it.

app/build.gradle
```diff
@@ -82,7 +82,7 @@ android {
 
     defaultConfig {
         minSdkVersion 23
-        targetSdkVersion 30
+        targetSdkVersion 31
```
AlvaroBrey, line 85: Should be a separate PR I think.

app/src/androidTest/java/com/owncloud/android/files/services/FileUploaderIT.kt
```diff
@@ -42,7 +42,7 @@ import junit.framework.Assert.assertTrue
 import org.junit.Before
 import org.junit.Test
 
-class FileUploaderIT : AbstractOnServerIT() {
+abstract class FileUploaderIT : AbstractOnServerIT() {
```
AlvaroBrey, line 45: comment: For reference, this could instead be done by using the Parameterized runner instead of creating two new classes.

app/src/main/java/com/nextcloud/client/di/ComponentsModule.java
```diff
@@ -119,6 +119,7 @@
 import com.owncloud.android.ui.preview.PreviewVideoActivity;
 import com.owncloud.android.ui.preview.pdf.PreviewPdfFragment;
 import com.owncloud.android.ui.trashbin.TrashbinActivity;
+import com.owncloud.android.utils.FilesUploadHelper;
```
AlvaroBrey, line 122:
```suggestion
```
Unused import

app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManager.kt
```diff
@@ -138,6 +139,8 @@ interface BackgroundJobManager {
 
     fun startNotificationJob(subject: String, signature: String)
     fun startAccountRemovalJob(accountName: String, remoteWipe: Boolean)
+    fun startFilesUploadJob(user: User)
+    fun getFileUploads(user: User): LiveData<MutableList<WorkInfo>>
```
AlvaroBrey, line 143: This should expose `JobInfo`, not `WorkInfo`. Check [fromWorkInfo](https://github.com/nextcloud/android/blob/e2ef6c41125569d67843dc22db9bfabe98d41563/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt#L131-L131) for an explanation. You will need to use `Transformations.map` to map from a `List<WorkInfo>` to a `List<JobInfo>`

app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
```diff
@@ -446,6 +447,17 @@ internal class BackgroundJobManagerImpl(
...
     }
 
+    override fun startFilesUploadJob(user: User) {
+        val request = oneTimeRequestBuilder(FilesUploadWorker::class, JOB_FILES_UPLOAD, user)
+            .build()
+
+        workManager.enqueue(request)
+    }
+
+    override fun getFileUploads(user: User): LiveData<MutableList<WorkInfo>> {
```
AlvaroBrey, line 457: Should expose `LiveData<List<JobInfo>>` instead. Check the comment in `BackgroundJobManager` interface.

app/src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -272,7 +272,7 @@ public static AppComponent getAppComponent() {
     @SuppressFBWarnings("ST")
     @Override
     public void onCreate() {
-        enableStrictMode();
+        // enableStrictMode();
```
AlvaroBrey, line 275:
```suggestion
        enableStrictMode();
```
Leftover.

AlvaroBrey (changes requested):

app/build.gradle
tobiasKaminsky: Done in https://github.com/nextcloud/android/pull/10797

app/src/androidTest/java/com/owncloud/android/files/services/FileUploaderIT.kt
tobiasKaminsky: Great idea. I will do it later.

AlvaroBrey:
Rebased and fixed for m3 ✨

AlvaroBrey (approved):

PR #10754, opened 2022-09-22
see favorite in shared view

tobiasKaminsky:
allow to favorite/unfavorite it

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
@@ -217,9 +217,9 @@ public class OCFileListAdapter extends RecyclerView.Adapter<RecyclerView.ViewHol
         return position;
     }
 
-    public void setFavoriteAttributeForItemID(String fileId, boolean favorite, boolean removeFromList) {
+    public void setFavoriteAttributeForItemID(String remotePath, boolean favorite, boolean removeFromList) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -217,9 +217,9 @@ public int getItemPosition(@NonNull OCFile file) {
         return position;
     }
 
-    public void setFavoriteAttributeForItemID(String fileId, boolean favorite, boolean removeFromList) {
```
AlvaroBrey, line 220: Method should be renamed to `setFavoriteAttributeForRemotePath`

app/src/main/java/com/owncloud/android/ui/adapter/OCShareToOCFileConverter.kt
```diff
@@ -55,8 +55,8 @@ object OCShareToOCFileConverter {
             isPreviewAvailable = firstShare.isHasPreview
             mimeType = firstShare.mimetype
             note = firstShare.note
-            fileId = firstShare.fileSource
-            remoteId = firstShare.remoteId.toString()
+            remoteId = firstShare.fileSource.toString()
```
AlvaroBrey, line 58: This is a bit confusing. Why are we setting `remoteId` from `OCShare.fileSource` and not from `OCShare.remoteId`? It may make sense to add a comment to explain why

PR #10762, opened 2022-09-23
Rotate images

surinder-tsys:
Rotate image functionality added to image files.

**How it works:**
1. Open any image in preview mode.
2. Click on 3 dot menu at top right and then click on Rotate menu option. Rotate Menu Option
3. On clicking Rotate menu option it will rotate the image 90° clockwise.
4. Now the rotated image will be uploaded to cloud under 2 case:
       4.1. When user rotates the screen i.e. configuration changes.
       4.2. When user comes back from the preview screen. 
Uploading rotate image notification
5. We are updating the image thumbnail in cache to give seamless experience to user and also updating the new rotated image file with current one.

**Note: Rotate functionality only applicable to JPG and PNG files.**

### DIFF
--- a/app/src/main/java/com/nextcloud/client/files/downloader/UploadTask.kt
+++ b/app/src/main/java/com/nextcloud/client/files/downloader/UploadTask.kt
@@ -2,7 +2,9 @@
  * Nextcloud Android client application
  *
  * @author Chris Narkiewicz
+ * @author TSI-mc
  * Copyright (C) 2021 Chris Narkiewicz <hello@ezaquarii.com>
+ * Copyright (C) 2022 TSI-mc
  *
  * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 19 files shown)


tobiasKaminsky:
Sorry for coming back this late.

I think the condition when a rotated images is saved is to unclear for the user.
I would rather use a persistent snackbar, which then directly (in background) saves the file.

surinder-tsys:
@tobiasKaminsky There can be condition when user can rotate multiple images at a time and upload all the images at once.
So for this case also do we have to show Snackbar with **Images has been modified** or we have to handle it differently.

And just a favor is that possible to merge this branch with any of your temporary branch so that we can fix the issue reported?

surinder-tsys:
Closing this PR as new PR has created for Rotate during new merging process and here is the link: https://github.com/nextcloud/android/pull/11490

Issue #10915, opened 2022-10-24
Server color inconsistency on NC 25 on first login


AlvaroBrey:
App will show server color on first screen after login, but user background color afterwards.

Caused by https://github.com/nextcloud/android/commit/2bf3cbe473096af823ee26113f38c834db81b42c, which was implemented without knowing the behaviour change in NC25 regarding server color capabilities.

_Originally posted by @AlvaroBrey in https://github.com/nextcloud/android/issues/10638#issuecomment-1285070831_

Issue #10638, opened 2022-08-18
🎨 Material 3️⃣


AndyScherzinger:
Based on the work for talk Android in https://github.com/nextcloud/talk-android/pull/2084

- [x] nextcloud/android#10641
- [x] nextcloud/talk-android#2312
- [x] nextcloud/android#10640

AlvaroBrey:
We have a slight problem with colors in Nextcloud 25, which I noticed this week. It looks like the server color capability now depends on each user's background picture. But this is a problem because:

- The app will look inconsistent (actual server color during and immediately after login, then user color next times)
- The background image is never actually shown in the app
- This will be a problem for branded clients, where we want brand color to be consistent all the time.

For these reasons I believe we should have a way to retrieve the true "server color" independently of user background, and always use that for the mobile apps.

Thoughts @AndyScherzinger @jancborchardt @tobiasKaminsky ?

cc @timkrueger @mahibi

jancborchardt:
> The app will look inconsistent (actual server color during and immediately after login, then user color next times)

How about we do this:
- Use our blue color for during log in process
- As soon as the account is connected we fetch the person’s color, or fall back to server color

That way we don’t have a worst-case 3-step color switch. And we really need to make sure this inconsistency does not exist, which we can by optimizing the loading, no?

> The background image is never actually shown in the app

So the splash screen is the 1 thing which we can unfortunately not edit (only in branded clients). But we might still want to use the background e.g. in the header part of the navigation.

> This will be a problem for branded clients, where we want brand color to be consistent all the time.

This actually depends on the customer I’d say. Admins could disable the functionality for people to set their own background. (Not sure if this is already implemented.)

AlvaroBrey:
> Use our blue color for during log in process

We already do this for the initial screen (the one with the Log In / Host your own server buttons), but we cannot control the login screen, as that is just a webview for the server's login flow. So if a user has a background image different from the server login image, it will always look different. For reference, here's an example of the worst-case scenario:

<details>
<summary> Example </summary>

In this example, the admin has set the server color to red, and the user has configured a yellow background image.

| Initial screen | During login | After login |
| --- | --- | --- |
[...]

PR #10925, opened 2022-10-25
Fix login colors in NC25

AlvaroBrey:
- AuthenticatorActivity: fetch user capability after account creation
- AuthenticatorActivity: on NC25, don't tint status/navigation bars

### DIFF
--- a/app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -83,10 +83,12 @@ import com.nextcloud.client.di.Injectable;
 import com.nextcloud.client.onboarding.FirstRunActivity;
 import com.nextcloud.client.onboarding.OnboardingService;
 import com.nextcloud.client.preferences.AppPreferences;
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
> * AuthenticatorActivity: fetch user capability after account creation

Why do we need them? They will be fetched on first loading of root folder. Or is it too late and thus some colors are not shown correctly?
I remember that therefore we fetch capabilities during login?

AlvaroBrey:
> > * AuthenticatorActivity: fetch user capability after account creation
> 
> Why do we need them? They will be fetched on first loading of root folder. Or is it too late and thus some colors are not shown correctly?

It's too late, the FAB and permission dialog will get drawn before the SynchronizeFolderOperation completes.

>  I remember that therefore we fetch capabilities during login?

We fetch the capabilities for the server, but in NC25 the color capability when logged in (calculated from user background) can be different from the one from the server.  See https://github.com/nextcloud/android/issues/10915 and https://github.com/nextcloud/android/issues/10638#issuecomment-1285070831 for context

tobiasKaminsky (approved):

tobiasKaminsky:
Thanks for explanation!

PR #10941, opened 2022-10-26
i18n: Change to a capital letter

Valdnet:


AlvaroBrey:
IIRC, when changing strings that have been translated already they need to be renamed instead of updated. @tobiasKaminsky can confirm

tobiasKaminsky:
> IIRC, when changing strings that have been translated already they need to be renamed instead of updated. @tobiasKaminsky can confirm

Either that, or we have to backport it to latest stable branch.
As it is only those two we can also backport it.

tobiasKaminsky:
Just a note: this is only for screenreader, so capitalization shall not make a difference, but still good to fix it.
(and shame on me for not doing it correct in first place)

tobiasKaminsky (approved):

comradekingu:
Why not "Icon for video overlay" in the name of consistency with the other string?

PR #10968, opened 2022-11-01
Check if e2e is available in DocumentsStorageProvider

tobiasKaminsky:
This does not yet work, as we query ArbitraryStorage, which is not allowed within context of DSP.
Maybe you have an idea @AlvaroBrey 


AlvaroBrey:
> Maybe you have an idea @AlvaroBrey

We need to stop using `ContentProvider`s for internal database access :sweat:. They are not meant for this. Otherwise I don't see how to work around this case.

tobiasKaminsky:
Seems now to directly work.
This is amazing!

Aizmansor:
Dihantar daripada telefon saya-------- Mesej asal --------Daripada: Tobias Kaminsky ***@***.***>Tarikh: Isn, 21 Nov 2022, 8:12 PTGKepada: nextcloud/android ***@***.***>Sk: Subscribed ***@***.***>Tajuk: Re: [nextcloud/android] Check if e2e is available in DocumentsStorageProvider (PR #10968)
Seems now to directly work.
This is amazing!

AlvaroBrey (approved):

Issue #10213, opened 2022-05-09
[Meta] Crashes caused by too much data in SQLite rows


AlvaroBrey:
This issue is intended to be a compilation of crashes caused (apparently) by too much data in SQL. We've patched some of these in the past but it's evident that a more integral solution or rework is needed to stop this happening all the time.

## Row too big to fit in CursorWindow
* https://github.com/nextcloud/android/issues/10199
* https://github.com/nextcloud/android/issues/10150
* https://github.com/nextcloud/android/issues/9693
* https://github.com/nextcloud/android/issues/9384
* many more closed ones

## Couldn't read row... Make sure Cursor is initialized
* https://github.com/nextcloud/android/issues/10193
* https://github.com/nextcloud/android/issues/10192
* https://github.com/nextcloud/android/issues/9948
* https://github.com/nextcloud/android/issues/6103
* many more closed ones

svenb1234:
Is this also a duplicate crash?

### Cause of error
```java
Exception in thread "main" android.database.sqlite.SQLiteBlobTooBigException: Row too big to fit into CursorWindow requiredPos=7538, totalRows=6830
    at android.database.sqlite.SQLiteConnection.nativeExecuteForCursorWindow(Native Method)
    at android.database.sqlite.SQLiteConnection.executeForCursorWindow(SQLiteConnection.java:1003)
    at android.database.sqlite.SQLiteSession.executeForCursorWindow(SQLiteSession.java:838)
    at android.database.sqlite.SQLiteQuery.fillWindow(SQLiteQuery.java:62)
    at android.database.sqlite.SQLiteCursor.fillWindow(SQLiteCursor.java:153)
    at android.database.sqlite.SQLiteCursor.onMove(SQLiteCursor.java:123)
    at android.database.AbstractCursor.moveToPosition(AbstractCursor.java:255)
    at android.database.AbstractCursor.moveToNext(AbstractCursor.java:287)
    at android.database.CursorWrapper.moveToNext(CursorWrapper.java:206)
    at com.owncloud.android.datamodel.FileDataStorageManager.getGalleryItems(FileDataStorageManager.java:2226)
    at com.owncloud.android.datamodel.FileDataStorageManager.getAllGalleryItems(FileDataStorageManager.java:2169)
    at com.owncloud.android.ui.adapter.GalleryAdapter.showAllGalleryItems(GalleryAdapter.kt:149)
[...]

PR #10977, opened 2022-11-02
Setup Room database

AlvaroBrey:
The point of this PR is to set up Room as a DB client, even if we are still using the FileContentProvider as an accessor for now.

After this is merged:
- Room will handle creation and upgrades of database
- The app will keep using the FileContentProvider as a database access point
- The FileContentProvider will stay mostly the same, but will get the database helper from Room instead of creating it.

This is the first step in the recommended way to migrate an app to Room incrementally, when a full migration is not possible immediately.

After this is finalized and merged, it will be possible to write DAOs (and Repositories) for the introduced entities and progressively stop using the ContentProviders as a way to access the DB internally. It should also be possible to use Room's auto-migrations for future DB upgrades.

Resources: 
https://developer.android.com/training/data-storage/room/sqlite-room-migration#incremental
https://medium.com/androiddevelopers/incrementally-migrate-from-sqlite-to-room-66c2f655b377 

- [x] Check projectionMap and setStrict
    -  Not supported directly in SupportSqliteQueryBuilder
    -  Extra validations added in place
- [x] Migrations
   - [x] Migrate LONG, STRING to INTEGER, TEXT
   - [x] Test with 3.22
   - [x] Test with several old app versions
       - [x] 3.21
       - [x] 3.20
       - [x] 3.19
       - [x] 3.18
       - [x] 3.17
       - [x] 3.10 -> crashes because of files_drop column that doesn't exist anymore
             - added new migrations 
       - [x] 3.3.0
       - [x] 2.0.0
       - [x] fall back to destructive migration for older versions?
            - [x] remove migration code for versions under that one?
       - [x]  check what happens when migrating a filelist table with too big rows (#10213) -> migration doesn't trigger crash, so that's fine
- [x] Finalize Room config (export schemas, automigration)
    - Automigration config will be added first time we do a migration
-  ~[ ] TypeConverters for boolean <-> int etc?~ out of scope. let's get room working first


AlvaroBrey:
@tobiasKaminsky @AndyScherzinger this is still in an early stage, but I'd appreciate any input on the approach and your opinion about this in general

AndyScherzinger:
Approach looks good to me and super happy to see this change being kicked off 👍

AlvaroBrey:
@tobiasKaminsky @AndyScherzinger this is now ready for review. I still  have to test migrations from 3.18-3.20, but since it's working from 3.10, I'd say it's a safe bet it will work for newer versions.

That being said, 3.10 is a semi-random point I chose for cutoff, as it's from January 2019, so almost 3 years old. Do you have any other idea of a "safe" point to test back to? Maybe 3.0?

AlvaroBrey:
> That being said, 3.10 is a semi-random point I chose for cutoff, as it's from January 2019, so almost 3 years old. Do you have any other idea of a "safe" point to test back to? Maybe 3.0?

Tested back to 3.3.0, which is the oldest 3.x version I can find an APK for.

AndyScherzinger:
I am only partially on duty this and next week, so it's unlikely that find the time to test drive this or next week 😔

AlvaroBrey:
> > That being said, 3.10 is a semi-random point I chose for cutoff, as it's from January 2019, so almost 3 years old. Do you have any other idea of a "safe" point to test back to? Maybe 3.0?
> 
> Tested back to 3.3.0, which is the oldest 3.x version I can find an APK for.

Finally set the cutoff to 2.0 and dropped earlier migrations. 2.0 is the oldest that doesn't break (upgrade from 1.x to current 3.22 is broken too, not just in this PR).

tobiasKaminsky:
> Finally set the cutoff to 2.0 and dropped earlier migrations. 2.0 is the oldest that doesn't break (upgrade from 1.x to current 3.22 is broken too, not just in this PR).

2.0 is more than enough :clap:

tobiasKaminsky:
Awesome!

tobiasKaminsky (approved):

PR #11001, opened 2022-11-07
Use material 3 style for SSO dialog

AlvaroBrey:
| Before | After |
|  --- | --- |
|  Screenshot_1667822281   |  Screenshot_1667822194  |


AndyScherzinger (commented):
@AlvaroBrey the buttons/background(?) need theming after `.show()`

AlvaroBrey:
> @AlvaroBrey the buttons/background(?) need theming after `.show()`

I'm not sure it makes sense; as this activity may be called asking for permission for an user that's not the currently active user.

stefan-niedermann:
Regarding theming I'd recommend white / black text for thr buttons. The default Material You color is not used anywhere I am aware of...

AlvaroBrey:
> did you check on background theming with `surface`

Done. The only difference is that the background for the dark variant is a little darker (I guess for better contrast with the buttons)

stefan-niedermann (approved):

AndyScherzinger (approved):

PR #11015, opened 2022-11-09
Bump compileSDK to 32

AlvaroBrey:
Once again it's blocking library updates. See https://github.com/nextcloud/android/pull/10841 for example


AndyScherzinger:
I run it on my phone as the production app, no issues so far but also had no recent events that would have triggered a push notification for example.

So seems good to go I'd say

AndyScherzinger (approved):
experiencing no issues so far 👍

Issue #10980, opened 2022-11-03
E2E: better UX


tobiasKaminsky:
![7546bd37-a6cc-4b85-91e8-cfbba43cd786](https://user-images.githubusercontent.com/5836855/198204283-dc91ee9b-0711-48ae-a0ef-28340493b2c6.PNG)

Android:
- add entry in settings
- state 1: setup
- state 2: add device
- state 3: show/remove locally

PR #11031, opened 2022-11-14
E2e ux

tobiasKaminsky:
Fix https://github.com/nextcloud/android/issues/10980
Needs: https://github.com/nextcloud/android-library/pull/1000


app/src/main/java/com/owncloud/android/ui/activity/SetupEncryptionActivity.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.owncloud.android.ui.dialog.SetupEncryptionDialogFragment
+
+class SetupEncryptionActivity : AppCompatActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        val user = intent?.getParcelableExtra("EXTRA_USER") as User?
+
+        if (user == null) {
+            Toast.makeText(this, "Error showing setup encryption dialog!", Toast.LENGTH_LONG).show()
```
AlvaroBrey, line 38: Message should be localized

app/src/main/res/values/strings.xml
```diff
@@ -1053,4 +1053,13 @@
     <string name="no_items">No items</string>
     <string name="check_back_later_or_reload">Check back later or reload.</string>
     <string name="e2e_not_yet_setup">E2E not yet setup</string>
+    <string name="displays_mnemonic">Displays your 12 words passphrase</string>
```
AlvaroBrey, line 1056:
```suggestion
    <string name="displays_mnemonic">Displays your 12 word passphrase</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -1053,4 +1053,13 @@
...
+    <string name="prefs_setup_e2e">Set up end to end encryption</string>
+    <string name="prefs_e2e_active">End to end encryption is set up!</string>
+    <string name="prefs_remove_e2e">Remove encryption locally</string>
+    <string name="remove_e2e">You can remove end to end encryption locally on this client</string>
+    <string name="confirm_removal">Remove local encryption</string>
+    <string name="remove_e2e_message">You can remove end to end encryption locally on this client. The encrypted files will remain on server, but will not be synced to this computer any longer.</string>
+    <string name="setup_e2e">During setup of end to end encryption you will receive random 12 word mnemonic, which you need to add another client. This will only be stored on this computer and can be shown here. Please note it down to a secure place!</string>
+    <string name="comon_copy">Copy</string>
```
AlvaroBrey, line 1064: Typo:  `comon`  needs an extra M

AlvaroBrey (changes requested):
I feel like the text under the preference is too long. I'd rather show a dialog with that text, which when clicking "OK" starts the other e2e setup steps. Additionally, see comments below

tobiasKaminsky:
> I feel like the text under the preference is too long. I'd rather show a dialog with that text, which when clicking "OK" starts the other e2e setup steps.

True, but idea from designer is that user can directly understand what is done, and do not first has to think if clicking on it might directly enable it.

app/src/main/res/values/strings.xml
```diff
@@ -1053,4 +1054,13 @@
...
     <string name="check_back_later_or_reload">Check back later or reload.</string>
     <string name="e2e_not_yet_setup">E2E not yet setup</string>
+    <string name="displays_mnemonic">Displays your 12 word passphrase</string>
+    <string name="prefs_setup_e2e">Set up end to end encryption</string>
+    <string name="prefs_e2e_active">End to end encryption is set up!</string>
+    <string name="prefs_remove_e2e">Remove encryption locally</string>
+    <string name="remove_e2e">You can remove end to end encryption locally on this client</string>
+    <string name="confirm_removal">Remove local encryption</string>
+    <string name="remove_e2e_message">You can remove end to end encryption locally on this client. The encrypted files will remain on server, but will not be synced to this computer any longer.</string>
+    <string name="setup_e2e">During setup of end to end encryption you will receive random 12 word mnemonic, which you need to add another client. This will only be stored on this computer and can be shown here. Please note it down to a secure place!</string>
```
AlvaroBrey, line 1064:
```suggestion
    <string name="setup_e2e">During setup of end to end encryption, you will receive a random 12 word mnemonic, which you will need to open your files on other devices. This will only be stored on this device, and can be shown again in this screen. Please note it down in a secure place!</string>
```
Sorry for not catching these yesterday 😥

AlvaroBrey (approved):

AlvaroBrey:
Screenshot tests need updating

app/src/main/java/com/owncloud/android/ui/activity/SetupEncryptionActivity.kt
```diff
@@ -0,0 +1,46 @@
...
+        super.onCreate(savedInstanceState)
+
+        val user = intent?.getParcelableExtra("EXTRA_USER") as User?
+
+        if (user == null) {
+            Toast.makeText(this, getString(R.string.error_showing_encryption_dialog), Toast.LENGTH_LONG).show()
+            return
+        }
+
+        val setupEncryptionDialogFragment = SetupEncryptionDialogFragment.newInstance(user, -1)
```
AlvaroBrey, line 43: Need to add a listener to close the activity when the dialog is cancelled or dismissed, otherwise user will have a blank screen

AlvaroBrey (changes requested):

app/src/main/java/com/owncloud/android/ui/activity/SetupEncryptionActivity.kt
AlvaroBrey: Done in https://github.com/nextcloud/android/pull/11031/commits/9ba895280749a7991b6fa6ed2bc9ba1984055a95

I've also set the background of the activity to transparent, so that it's more seamless for the user:

<details>

Screenshot_1668761427


</details>

AlvaroBrey (approved):

app/schemas/com.nextcloud.client.database.NextcloudDatabase/65.json
```diff
@@ -1,18 +1,18 @@
 {
```
AlvaroBrey, line 1: Already commited schemas must not be changed! The app was probably built with changed entities but before bumping db version.

- [ ] Revert this file

app/schemas/com.nextcloud.client.database.NextcloudDatabase/66.json
```diff
@@ -0,0 +1,1124 @@
+{
```
AlvaroBrey, line 1: - [ ] Enable automatic upgrades in room

app/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
```diff
@@ -410,6 +421,45 @@ private void setupRecommendPreference(PreferenceCategory preferenceCategoryMore)
         }
     }
 
+    private void setupE2EPreference(PreferenceCategory preferenceCategoryMore) {
+        Preference preference = findPreference("setup_e2e");
+
+        if (preference != null) {
+            if (FileOperationsHelper.isEndToEndEncryptionSetup(this, user) ||
+                !CapabilityUtils.getCapability(this).getEndToEndEncryptionKeysExist().isFalse()) {
```
AlvaroBrey, line 429:
```suggestion
                CapabilityUtils.getCapability(this).getEndToEndEncryptionKeysExist().isTrue()) {
```

Otherwise it will also be hidden for UNKNOWN value.

tobiasKaminsky: I thought that this is what it should be…and I was wrong :S

app/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
```diff
@@ -860,6 +941,13 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
             }
         } else if (requestCode == ACTION_SHOW_MNEMONIC && resultCode == RESULT_OK) {
             handleMnemonicRequest(data);
+        } else if (requestCode == ACTION_E2E && data != null && data.getBooleanExtra(SetupEncryptionDialogFragment.SUCCESS, false)) {
+            PreferenceCategory preferenceCategoryMore = (PreferenceCategory) findPreference("more");
+
+            setupE2EPreference(preferenceCategoryMore);
+            setupE2EKeysExist(preferenceCategoryMore);
+            setupE2EMnemonicPreference(preferenceCategoryMore);
+            removeE2E(preferenceCategoryMore);
```
AlvaroBrey, line 950: In my case, when I input the passphrase and finish the setup, this hides all of the e2e prefs. Not sure what's happening there.

tobiasKaminsky: This is because in first run the preferences are removed as there is no other way.
And then after setup e2e they cannot be shown again.

Right now I see no other option other that the user has to re-enter settings page.

AlvaroBrey: I can see two options:

- Add the preferences instead (new Preference(), and then add to the group)
- Restart the Activity  (new Intent with FLAG_NO_ANIMATION, launch, finish current)

tobiasKaminsky: Nice idea with restarting the activity!

Issue #10783, opened 2022-10-06
Improve performance/usability when navigating between folders


starypatyk:
I have noticed a few issues related to performance/usability when navigating between folders - esp. when a folder contains larger number of files. In my testing I use a folder with about 300 photos.

I plan to create separate issues/PRs for each of the items below and use this issue to track the overall progress.

- [x] Full refresh of a folder is performed during onBrowseUp (back button) #10688
- [x] WebDAV call sent by the Android app to read folder contents executes a few database queries for each file.
  - [x] Reading information about share notes requires 4 DB queries per file. https://github.com/nextcloud/server/pull/34471
  - [x] Reading DAV:displayname property requires 1 query per file. https://github.com/nextcloud/server/pull/34508
- [ ] API call to retrieve information about files shared from a folder requires a few database queries for each file. https://github.com/nextcloud/server/pull/34918
- [x] https://github.com/nextcloud/android/pull/11033
- [ ] When changing folder in the app the file list gets refreshed three times, i.e. OCFileListAdapter::swapDirectory is invoked three times. Two of the calls seem redundant when all the information cached on the device is still valid.
  - [x] OCFileListAdapter::swapDirectory is invoked in response to EVENT_SINGLE_FOLDER_CONTENTS_SYNCED broadcast - even when the ETag check in RefreshFolderOperation determines that the folder has not changed. #11100
  - [ ] OCFileListAdapter::swapDirectory is invoked again in response to EVENT_SINGLE_FOLDER_SHARES_SYNCED broadcast. It seems that the code that retrieves the shares does not try to check, if the data about shares retrieved from the server is the same as cached data. Trying to optimize the process of saving shares in the local DB in #11255.
- [x] Reading file information from the local SQLite database (FileDataStorageManager::getFolderContent) takes much longer than expected. Profiling suggests that most of the time is spent inside createFileInstance method in these operations:
  - [x] deserializing JSON data (~50% of time) ~~Initial attempt to improve this (#11181) abandoned.~~ Simpler approach proposed: #11251.
  - [x] obtaining column indices in cursor.getColumnIndexOrThrow (~30% of time) https://github.com/nextcloud/android/pull/11098
  - [x] https://github.com/nextcloud/android/pull/11227

starypatyk:
Baseline results from my dev environment (331 files in a folder). When entering a folder the app executes the following API calls:

Request | Execution time [ms] | # of DB queries
--- | ---: | ---:
[...]

PR #11033, opened 2022-11-14
File list "jumps" after scrolling when an asynchronous operation finishes

starypatyk:
During investigation of #10783 I have found that after scrolling up, the file list "jumps" when an asynchronous operation finishes.

This seems to be caused by AppBar being expanded during refresh (after completion of the async operations). If the AppBar was hidden earlier by the user scrolling up, this causes a "jump".

@tobiasKaminsky - this is the issue that we discussed briefly during the conference - fix proposed.

See short videos below.

https://user-images.githubusercontent.com/8277636/201758099-ea13bec8-9834-4d95-a123-496cb65f2c6e.mp4

https://user-images.githubusercontent.com/8277636/201758268-f7a6dd09-503b-44cd-b439-4a75bd2c92e9.mp4


AlvaroBrey (approved):

AlvaroBrey:
Thanks @starypatyk !

PR #11008, opened 2022-11-08
Redesign file actions menu as a BottomSheet

AlvaroBrey:
⚠ Don't merge without squashing

Fixes #6239

- [x] Tests written, or not not needed

PR #11037, opened 2022-11-15
Only show "set encrypted" if folder is empty and has no empty files/folder in it

tobiasKaminsky:


AlvaroBrey:
Tests need updating. Also this will conflict with #11008 , but that can be handled after one of them is merged

AlvaroBrey (approved):

PR #11044, opened 2022-11-16
Fixed unknown error while encrypting empty shared folder.

surinder-tsys:
**Precondition:**
User is logged in
Files / Folders available

 
**Steps:**

1. Start the app
2. Create a folder 

3. share the folder
4. set encryption to the shared folder
 

**Actual result:**

unknown error appears as an in app notification

**Expected result:**

error is displayed that shared folders cannot be encrypted

error_screenshot
**"Unknown error" screenshot**

fixed_screenshot
**Correct error message screenshot**

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
+++ b/app/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
@@ -11,7 +11,7 @@
  * Copyright (C) 2015 ownCloud Inc.
  * Copyright (C) 2018 Andy Scherzinger
  * Copyright (C) 2020 Chris Narkiewicz <hello@ezaquarii.com>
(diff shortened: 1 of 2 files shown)


AlvaroBrey (approved):

AlvaroBrey:
Thanks @surinder-tsys !

AlvaroBrey:
We should also hide the `Encrypt` item from the file actions menu, not just show an error, but that's out of the scope of this PR. I'll create a new issue for that.

surinder-tsys:
@AlvaroBrey By hiding "Encrypt" menu options the current PR changes will be invalid right?

tobiasKaminsky:
Superseeded by #11052 as we do not show options if they are not allowed.
This is better UX than showing all options with various reasons why it is not allowed.

PR #10491, opened 2022-07-07
Allow Share Dialogs to create Shortcuts

newhinton:
- [ ] Tests written, or not not needed

Closes #3175 

This is based on #5522. 
Thanks to @wiwie!

This Pull request allows the user to create homescreen-shortcuts via the share dialog.

| Share for Folder         | Share for Files     |
|--------------|-----------|
| grafik | grafik |

PR #11059, opened 2022-11-19
Allow Pin to Home with file actions bottom sheet

### LINKED ISSUE #3175: Add home screen folder/file shortcut widget
Requesting the ability to drop a shortcut on Android home screens that link to specific folders or files. Thanks!

newhinton:

This PR superseeds #10491
It implements the same functionality for the new bottom-toolbar.

Closes #10491
Closes #3175

### DIFF
--- a/app/src/main/java/com/nextcloud/utils/ShortcutUtil.kt
+++ b/app/src/main/java/com/nextcloud/utils/ShortcutUtil.kt
@@ -41,7 +41,6 @@ import com.owncloud.android.ui.activity.FileDisplayActivity
 import com.owncloud.android.utils.MimeTypeUtil
 import com.owncloud.android.utils.theme.ViewThemeUtils
 import javax.inject.Inject
-import kotlin.math.roundToInt
 
 class ShortcutUtil @Inject constructor(private val mContext: Context) {
 
@@ -104,12 +103,8 @@ class ShortcutUtil @Inject constructor(private val mContext: Context) {
     }
 
     private fun bitmapToAdaptiveBitmap(orig: Bitmap): Bitmap {
-        val screenDensity = mContext.resources.displayMetrics.density
-        val adaptiveIconSizeConst = mContext.resources.getInteger(R.integer.adaptive_icon_size)
-        val adaptiveIconPaddingConst = mContext.resources.getInteger(R.integer.adaptive_icon_padding)
-
-        val adaptiveIconSize = (adaptiveIconSizeConst * screenDensity).roundToInt()
-        val adaptiveIconOuterSides = (adaptiveIconPaddingConst * screenDensity).roundToInt()
+        val adaptiveIconSize = mContext.resources.getDimensionPixelSize(R.dimen.adaptive_icon_size)
+        val adaptiveIconOuterSides = mContext.resources.getDimensionPixelSize(R.dimen.adaptive_icon_padding)
         val drawable: Drawable = BitmapDrawable(mContext.resources, orig)
         val bitmap = Bitmap.createBitmap(adaptiveIconSize, adaptiveIconSize, Bitmap.Config.ARGB_8888)
         val canvas = Canvas(bitmap)
--- a/app/src/main/java/com/nextcloud/utils/ShortcutUtil.kt
+++ b/app/src/main/java/com/nextcloud/utils/ShortcutUtil.kt
@@ -25,15 +25,13 @@ import android.app.PendingIntent
 import android.app.PendingIntent.FLAG_IMMUTABLE
 import android.content.Context
 import android.content.Intent
-import android.content.pm.ShortcutInfo
-import android.content.pm.ShortcutManager
 import android.graphics.Bitmap
 import android.graphics.Canvas
 import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.Drawable
-import android.graphics.drawable.Icon
-import android.os.Build
-import androidx.annotation.RequiresApi
+import androidx.core.content.pm.ShortcutInfoCompat
+import androidx.core.content.pm.ShortcutManagerCompat
+import androidx.core.graphics.drawable.IconCompat
 import androidx.core.graphics.drawable.toBitmap
 import com.owncloud.android.R
 import com.owncloud.android.datamodel.OCFile
@@ -52,22 +50,20 @@ class ShortcutUtil @Inject constructor(private val mContext: Context) {
      *
      * @param file The file/folder to which a pinned shortcut should be added to the home screen.
      */
-    @RequiresApi(Build.VERSION_CODES.O)
     fun addShortcutToHomescreen(file: OCFile, viewThemeUtils: ViewThemeUtils) {
-        val shortcutManager = mContext.getSystemService(ShortcutManager::class.java)
-        if (shortcutManager.isRequestPinShortcutSupported) {
+        if (ShortcutManagerCompat.isRequestPinShortcutSupported(mContext)) {
             val intent = Intent(mContext, FileDisplayActivity::class.java)
             intent.action = FileDisplayActivity.OPEN_FILE
             intent.putExtra(FileActivity.EXTRA_FILE, file.remotePath)
             intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
             val shortcutId = "nextcloud_shortcut_" + file.remoteId
-            val icon: Icon
+            val icon: IconCompat
             var thumbnail = ThumbnailsCacheManager.getBitmapFromDiskCache(
                 ThumbnailsCacheManager.PREFIX_THUMBNAIL + file.remoteId
             )
             if (thumbnail != null) {
                 thumbnail = bitmapToAdaptiveBitmap(thumbnail)
-                icon = Icon.createWithAdaptiveBitmap(thumbnail)
+                icon = IconCompat.createWithAdaptiveBitmap(thumbnail)
             } else if (file.isFolder) {
                 val bitmapIcon = MimeTypeUtil.getFolderTypeIcon(
                     file.isSharedWithMe || file.isSharedWithSharee,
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1147,6 +1148,9 @@ public boolean onFileActionChosen(@IdRes final int itemId, Set<OCFile> checkedFi
                 mContainerActivity.getFileOperationsHelper().toggleFileLock(singleFile, true);
             } else if (itemId == R.id.action_unlock_file) {
                 mContainerActivity.getFileOperationsHelper().toggleFileLock(singleFile, false);
+            } else if (itemId == R.id.action_pin_to_homescreen) {
+                new ShortcutUtil(getContext()).addShortcutToHomescreen(singleFile);
```
AlvaroBrey, line 1152: This should be injected instead of created here

app/src/main/java/com/owncloud/android/utils/MimeTypeUtil.java
```diff
@@ -172,6 +172,28 @@ public static Drawable getDefaultFolderIcon(Context context,
...
+            drawableId = R.drawable.folder_encrypted;
+        } else if (WebdavEntry.MountType.EXTERNAL == mountType) {
+            drawableId = R.drawable.folder_external;
+        } else if (WebdavEntry.MountType.GROUP == mountType) {
+            drawableId = R.drawable.folder_group;
+        } else {
+            drawableId = R.drawable.folder;
+        }
+
+        return drawableId;
```
AlvaroBrey, line 193: To avoid duplicating, the existing `getFolderTypeIcon`  method should be refactored so that it calls this one to get the ID. Additionally the  logic in `getFolderTypeIcon` has been changed so that the groupfolder type takes precedence, please fix that.

app/src/main/java/com/owncloud/android/utils/ShortcutUtils.kt
```diff
@@ -0,0 +1,112 @@
...
+import android.content.Context
+import android.graphics.Bitmap
+import android.graphics.Canvas
+import android.graphics.drawable.Drawable
+import android.graphics.drawable.BitmapDrawable
+import android.graphics.drawable.Icon
+import android.os.Build
+import kotlin.math.roundToInt
+
+class ShortcutUtil(val mContext: Context) {
```
AlvaroBrey, line 42:
```suggestion
class ShortcutUtil @Inject constructor (val mContext: Context) {
```
This will allow you to then inject the `ShortcutUtil` in `OCFileListFragment` and wherever else is needed.

Additionally the file should be name to `ShortcutUtil.kt` or the class to `ShortcutUtils`, to have them both match.

app/src/main/java/com/owncloud/android/utils/ShortcutUtils.kt
```diff
@@ -0,0 +1,112 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.owncloud.android.utils
```
AlvaroBrey, line 22: Please move to `com.nextcloud.client.utils` instead

app/src/main/java/com/owncloud/android/utils/ShortcutUtils.kt
```diff
@@ -0,0 +1,112 @@
...
+
+class ShortcutUtil(val mContext: Context) {
+
+    /**
+     * Adds a pinned shortcut to the home screen that points to the passed file/folder.
+     *
+     * @param file The file/folder to which a pinned shortcut should be added to the home screen.
+     */
+    fun addShortcutToHomescreen(file: OCFile) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
AlvaroBrey, line 50: If the functionality only works >= O, then the menu item should be hidden (in FileMenuFilter) when the current SDK is less than that.

app/src/main/java/com/owncloud/android/utils/ShortcutUtils.kt
```diff
@@ -0,0 +1,112 @@
...
+            val shortcutManager = mContext.getSystemService(ShortcutManager::class.java)
+            if (shortcutManager.isRequestPinShortcutSupported) {
```
AlvaroBrey, line 52: Same as above: this should be taken into account in order to hide the menu entry.

app/src/main/java/com/owncloud/android/utils/ShortcutUtils.kt
```diff
@@ -0,0 +1,112 @@
...
+                    successCallback.intentSender
+                )
+            }
+        }
+    }
+
+    private fun bitmapToAdaptiveBitmap(orig: Bitmap): Bitmap {
+        val screenDensity = mContext.resources.displayMetrics.density
+        val adaptiveIconSize = (108 * screenDensity).roundToInt()
+        val adaptiveIconOuterSides = (18 * screenDensity).roundToInt()
```
AlvaroBrey, line 101: What do these icon sizes correspond to?  In any case, they should be extracted to `dimen` resources or at least constants.

app/src/main/res/values/strings.xml
```diff
@@ -1054,4 +1054,6 @@
     <string name="check_back_later_or_reload">Check back later or reload.</string>
     <string name="e2e_not_yet_setup">E2E not yet setup</string>
     <string name="error_file_actions">Error showing file actions</string>
+    <string name="pin_home_text">This will create a shortcut on your Homescreen</string>
+    <string name="pin_home">Pin to Homescreen</string>
```
AlvaroBrey, line 1058:
```suggestion
    <string name="pin_home_text">This will create a shortcut on your Home screen</string>
    <string name="pin_home">Pin to Home screen</string>
```

AlvaroBrey (changes requested):
Hey @newhinton, many thanks! This works as expected, but there's a bunch of things that need to be polished before it can be merged.

You can do this yourself and re-request review. If you feel unsure about how to implement some of my requests, feel free to reach out to me via Talk or email.

If  you feel overwhelmed I can also do some the more complicated things for you. Let me know what you think!

newhinton:
@AlvaroBrey I know this has been mentioned before, but could you ask your team if it is possible to run all the github actions only on selected branches and or tagged branches? 

First, usually not every commit needs every action automatically run wich saves resources, and to be frank at some point the mails it sends out are just spam. It would be great if this would be applied more carefully.


Regarding the changes: i will make them, that's not a problem at all. I will rerequest a review when i am done!

app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
newhinton: What exactly do you mean by injection?

app/src/main/java/com/owncloud/android/utils/MimeTypeUtil.java
newhinton: I think i added this wrongly after pulling code from the old one to the new pr. I will remove the "broken" call completely and adapt the call at the source.

AlvaroBrey:
> @AlvaroBrey I know this has been mentioned before, but could you ask your team if it is possible to run all the github actions only on selected branches and or tagged branches?
> 
> First, usually not every commit needs every action automatically run wich saves resources, and to be frank at some point the mails it sends out are just spam. It would be great if this would be applied more carefully.

We run them on PRs that target master or a stable branch on purpose, so we can verify the changes _before_ merging. You may reduce the number of comments by:

- pushing the commits in batches rather than one at a time. actions will only run for the last of the batch
-  changing your subscription for the PR. with "customize" you can set it so that only mentions and close/merge/reopen events trigger an email
- changing your notification settings for Github as a whole so that you don't get emails for every notification

> 
> Regarding the changes: i will make them, that's not a problem at all. I will rerequest a review when i am done!

Thanks!

app/src/main/java/com/owncloud/android/utils/ShortcutUtils.kt
newhinton: If i remember correctly, those are taken from here:
https://developer.android.com/develop/ui/views/launch/icon_design_adaptive

They are specified by convention, and i could not find constants for them. I will extract them to dimen, and add a comment explaining that.

app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
AlvaroBrey: We use Dagger as a DI framework; the `ShortcutUtil` class should be injected into OCFileListFragment instead of created by it. I can do this if you want

newhinton: That would be great! I have not used Dagger so i wouldn't really know where to start. But i'll watch what you do with it!

AlvaroBrey: Done here: https://github.com/nextcloud/android/pull/11059/commits/8a24ea03a014f70fdc5d0df0d9603e83809c6c5b

AlvaroBrey:
@newhinton I just pushed the DI patch, and a couple of extras. Take a look at  https://github.com/nextcloud/android/pull/11059/commits/52e1e8cef5031a12fc57e0b75081c55d94164fa9 especially, as it will allow shortcuts on SDK < 26 too.

app/src/main/java/com/nextcloud/utils/ShortcutUtil.kt
```diff
@@ -0,0 +1,120 @@
...
+                icon = IconCompat.createWithBitmap(bitmapIcon)
+            } else {
+                icon = IconCompat.createWithResource(
+                    mContext,
+                    MimeTypeUtil.getFileTypeIconId(file.mimeType, file.fileName)
+                )
+            }
+            val pinShortcutInfo = ShortcutInfoCompat.Builder(mContext, shortcutId)
+                .setShortLabel(file.fileName)
+                .setLongLabel("Open " + file.fileName)
```
AlvaroBrey, line 85: This string needs to be extracted to strings.xml in order to be translatable. Use a parameter in the extracted string (`%1$s`) and then call `context.getString(R.string.your_string_id, file.fileName)`

AlvaroBrey (changes requested):

AlvaroBrey (approved):
Great work @newhinton ! Thanks very much. I'll wait for @tobiasKaminsky to review, but from my perspective this is ready to go

PR #11067, opened 2022-11-22
GalleryFragment: fix crash when scrolling slowly until the end of the list

AlvaroBrey:


AlvaroBrey:
> Never experienced it, but makes sense, as then follow up code cannot work.

I can reproduce it reliably with ~10 pictures and scrolling very slowly down.

PR #11073, opened 2022-11-23
Disable StrictMode by default

AlvaroBrey:
StrictMode is not a good fit for the state of our app, since we are way too far from async disk operations or correct threading.
As a result, we find ourselves disabling the strict mode manually every time we actually want to work on something, as otherwise the logspam drowns everything else.

With this PR I propose to disable StrictMode unless a prop is set in the Gradle build, which will also control LeakCanary.
Running `./gradlew installGplayDebug -P perfAnalysis` will install the app with both LeakCanary and StrictMode enabled.
This allows us to analyze performance problems on demand, without making our daily work harder.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -69,6 +69,8 @@ file("$project.rootDir/ndk.env").readLines().each() {
     ndkEnv.put(key, value)
 }
 
+def perfAnalysis = project.hasProperty('perfAnalysis')
+
 android {
 
     compileSdkVersion 32
@@ -85,6 +87,7 @@ android {
         targetSdkVersion 31
 
         buildConfigField 'boolean', 'CI', ciBuild.toString()
+        buildConfigField 'boolean', 'RUNTIME_PERF_ANALYSIS', perfAnalysis.toString()
 
         javaCompileOptions {
             annotationProcessorOptions {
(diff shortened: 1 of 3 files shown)


tobiasKaminsky:
That is a way better approach!
Back then I thought it is good to enable it, so that it bugs us that we get rid of those async.
But indeed it is way harder to do, and meanwhile it distracts us.

app/build.gradle
```diff
@@ -69,6 +69,8 @@ file("$project.rootDir/ndk.env").readLines().each() {
     ndkEnv.put(key, value)
 }
 
+def perfAnalysis = project.hasProperty('perfAnalysis')
```
tobiasKaminsky, line 72: Can you add this to one of our docs?
Especially "-P " might not be known to all devs.

tobiasKaminsky (changes requested):
Please add doc

app/build.gradle
AlvaroBrey: Done!

AlvaroBrey:
> Back then I thought it is good to enable it, so that it bugs us that we get rid of those async.

Future project: get StrictMode warnings from logcat in CI, as another check

tobiasKaminsky (approved):
Awesome idea to combine leakCanary and strictMode with perf flag!

PR #11089, opened 2022-11-24
Video preview: use fullscreen dialog for video instead of new activity

AlvaroBrey:
This allows transfering the playback directly between Player views, thus avoiding creating
a new ExoPlayer, re-loading the stream, having to pass playing status/current position, etc.

In summary, this provides a much more seamless experience when switching to/from fullscreen.

Additionally:
- Always enable buffering animation so it's clear when a video is loading
- Add padding to fullscreen duration numbers so they don't get cut off on rounded screens


tobiasKaminsky:
That is a super idea!
I tested it with two videos
- one from $internet: looks good
- one captured on same emulator: ratio is totally off

AlvaroBrey:
> * one captured on same emulator: ratio is totally off

This looks like a pre-existing bug that just shows here easier. The video that shows the problem is **stored in horizontal but has a rotation metadata**. There was a bug in Android with this, but it was patched in Android 11 (API 30). Reference: https://github.com/google/ExoPlayer/issues/1482 There are other bugs with SurfaceView and metadata rotation, and workarounds with TextureView.

In the master branch and previous releases, this is reproducible by:
- Going to fullscreen and then back
- Opening preview fragment, leaving the app by Home and then coming back from Recents. Video (stable-3.22): 
[ rotation.webm](https://user-images.githubusercontent.com/8677743/203965477-32a2186a-d751-45c1-bbac-45471b9488a2.webm)

It looks like the bug is only triggered when the video is "resumed" on a surface, either a new one or the same one it was already on. 

I can reproduce this in api 29 and lower, but can't reproduce in a higher API level.  I can also only reproduce it with the 29 - Google Play image, not with the 29 - google APIs one. Fix may have been backported?

There may be a possible workaround by using TextureView and applying the rotation manually (https://github.com/google/ExoPlayer/issues/91#issuecomment-130604964). This would have to be confined to android <=29, as TextureView is more battery-consuming than the default SurfaceView.

tobiasKaminsky:
First, great that you figured it out!

> @tobiasKaminsky what do you think we should do here? Attempt the workaround for API <29, or just let it be? It only happens with some <=29 devices, and only with videos with metadata rotation.

Actually I have no idea either. <=29 is quite a lot of users, which might be not a problem, depending how many videos actually use this rotation info.




> There may be a possible (but fairly complicated) workaround by using TextureView and applying the rotation manually. This would have to be confined to android <=29, as TextureView is more battery-consuming than the default SurfaceView.

How long would this roughly take?

Then we could first merge without above workaround, but when to many users suffer by it, we can use this workaround or revert.

Another option would be to have this only for >29 enabled, and once we drop 28 support, we can change it.

AlvaroBrey:
Another workaround idea: in devices with api <= 29, keep using the old approach (new exoplayer rather than transferring the old one). This will invalidate the UX gains for api 29 and under, though.

AlvaroBrey:
> How long would this roughly take?

I have no idea :( I think the workaround in my previous comment may be preferrable

AlvaroBrey:
Rotation bug worked around in https://github.com/nextcloud/android/pull/11089/commits/4fc73f069ad8fc365b17dc0583c1b4b359b459a1

tobiasKaminsky:
Works perfectly fine with "broken" video.

tobiasKaminsky (approved):

PR #11092, opened 2022-11-25
FolderPickerActivity: disable copy/move to the same folder

AlvaroBrey:
It's a no-op anyways, right now, and causes bad UX

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
@@ -74,6 +74,9 @@ open class FolderPickerActivity :
     private var mChooseBtn: MaterialButton? = null
     private var caption: String? = null
 
+    private var mAction: String? = null
+    private var mTargetFilePaths: ArrayList<String>? = null
+
     @Inject
     lateinit var localBroadcastManager: LocalBroadcastManager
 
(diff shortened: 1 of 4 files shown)


tobiasKaminsky:
It works with
- move file in root folder
- see that root folder is not allowed
- go into subfolder
- see that it is allowed

But when I start within a subfolder
- create folder 1
- create empty file.md
- move file
- see that it is allowed in root folder and in folder 1

AlvaroBrey:
> It works with
> 
> * move file in root folder
> * see that root folder is not allowed
> * go into subfolder
> * see that it is allowed
> 
> But when I start within a subfolder
> 
> * create folder 1
> * create empty file.md
> * move file
> * see that it is allowed in root folder and in folder 1

Fixed!

tobiasKaminsky:
Now this works again:
-create /testFolder
- move this folder
- go to /testFolder
--> it is allowed

tobiasKaminsky:
> You mean moving a folder inside itself is allowed?

It is within this PR right now possible, but it should not be allowed.

Similar, it should not be possible to move /TestFolder to /TestFolder/1/ or any other subfolder of its own.

AlvaroBrey:
> > You mean moving a folder inside itself is allowed?
> 
> It is within this PR right now possible, but it should not be allowed.
> 
> Similar, it should not be possible to move /TestFolder to /TestFolder/1/ or any other subfolder of its own.

This was out of the scope of this PR, but fixed anyway in https://github.com/nextcloud/android/pull/11092/commits/6d4a65b45ccf67801dba90d7da1c8856c5968c4f

tobiasKaminsky:
Great one!

tobiasKaminsky (approved):

PR #11098, opened 2022-11-26
Use Room for filelist table reads

starypatyk:
This is an attempt to improve performance of retrieval of information from the `filelist` table of the local SQLite database by using Room infrastructure prepared by @AlvaroBrey in #10977.

As the first step I have converted the `getFolderContent` method only.

Preliminary timings from a low end phone (on a folder with ~330 files) suggest reduction of the execution time of this method from ~**400** ms to ~**200** ms.

I have tried to follow the pattern proposed by @AlvaroBrey - any comments are welcome.

I think that the PR might be merged once all the methods in `FileDataStorageManager` that reference the `filelist` table are converted to Room. Any suggestions otherwise?

**EDIT**

I have noticed a significant increase of issues reported by SpotBugs. This seems to be caused by the fact that the old code somehow automatically converted NULL values in integer columns to zero. Although I am not sure about this behaviour - see: https://stackoverflow.com/a/43702322

I have implemented wrapper functions `nullToZero` to mitigate this, but I am not sure if this is a correct approach.

I have also noticed that the file/directory modification dates are incorrect. Seems that the timestamp columns must be declared as `Long` rather than `Int`. I have changed the `FileEntity` class accordingly - for all columns that looked like a timestamp.

**EDIT 2**

I have also converted the `getFileById` method.


AlvaroBrey (commented):
This is great :blue_heart: ! Thank you very much for getting this started.

The approach generally looks good; the only thing I dislike is the duplication of `createFileInstance` (and later `createContentValues*` if you tackle write methods), but it can't be undone until everything is ported over to Room.

> I think that the PR might be merged once all the methods in FileDataStorageManager that reference the filelist table are converted to Room. Any suggestions otherwise?

Yes, this makes sense. I wouldn't merge it before because of the aforementioned duplication. There is a lot of outstanding technical debt in `FileDataStorageManager` (like, why does this class also handle capabilities and shares?  and media scans!?), but that is for another moment.

However, migrating all of this may not be so easy. If you need any guidance please reach out in our community chat, and I can contribute to this PR as well when I have the time, if you find that OK.

> I have implemented wrapper functions nullToZero to mitigate this, but I am not sure if this is a correct approach.

This looks fine in principle. At a later stage, the fields that are actually booleans may be converted to `Boolean` in the entity and that should make the code much cleaner too. (this is again, out of scope for this PR)

> I have also noticed that the file/directory modification dates are incorrect. Seems that the timestamp columns must be declared as Long rather than Int. I have changed the FileEntity class accordingly - for all columns that looked like a timestamp.

This was an oversight on my part, thanks for noticing. I've looked through the old `createFileInstance` to see which fields were read as `Long` before, and found that `parent` and `content_length` also need to be changed. Pushed a commit for that.


In general, this is the most critical migration to Room, which is why I didn't start directly with it. Extensive testing is needed to verify everything works as expected with existing databases as well as new ones. Please do your best to test everything you can, but I'll also do a lot of smoketesting before merging, when the PR is ready.

Again, thank you very much for attempting this difficult task, and reach out for guidance if needed!

AlvaroBrey:
> I have implemented wrapper functions `nullToZero` to mitigate this, but I am not sure if this is a correct approach.

I've just taken a look at the [default values for OCFile](https://github.com/nextcloud/android/blob/a489619db3362fe1d2c156331946ae8dca717942/app/src/main/java/com/owncloud/android/datamodel/OCFile.java#L471-L471) to verify this. The only exception is `fileId`, which should instead be `-1` when null. However it's the primary key, so it _shouldn't_ be null, even if nullable for legacy reasons.

app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -1046,6 +1023,93 @@ private OCFile createFileInstance(Cursor cursor) {
         return ocFile;
     }
 
+    private int nullToZero(Integer i) {
```
tobiasKaminsky, line 1026: I am not sure if we should do this.
As far as I think this content provider -> room should be only refactoring, but not changing any context.

E.g. setting parentId or fileId to 0, will artificially link some files together, which should not.

I would rather bail out if such null values occur, as they should not happen in first place?

AlvaroBrey: The point is that the previous approach did this automatically (converting nulls to zeroes), since we're setting things to a non-nullable primitive value in OCFile. See my previous comments.

AlvaroBrey: To expand on this, we could leave it so that there are NPEs thrown if the values are null, but _that_ would be changing the previous behaviour.

app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -138,13 +144,12 @@ OCFile getFileByPath(String type, String path) {
 
     public @Nullable
     OCFile getFileById(long id) {
```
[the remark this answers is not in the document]
starypatyk: Agree - committed in https://github.com/nextcloud/android/pull/11098/commits/b04e499deb6f4f8c9c340cc34ccec5fde2791a9f

app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
tobiasKaminsky: Oh. Okay, then let us keep it this way.
Thanks for explaining and reference.

tobiasKaminsky (approved):

starypatyk:
@AlvaroBrey @tobiasKaminsky 

When I submitted this PR I thought that it would be good to merge this PR when all database operations related to the `filelist` table are migrated to Room.

I have underestimated the number of changes required to achieve this. :disappointed: I would now suggest to merge it in (more or less) current state.

I have managed to migrate most of the `query` operations, just enough to remove the old `createFileInstance(cursor)` method. So there is no need to maintain two very similar methods.

I have been able to fix most of the issues reported by various CI jobs - with two exceptions:

- In the [CodeQL job](https://github.com/nextcloud/android/actions/runs/3614866889/jobs/6091593015) I have not been able to find information about the reason of the failure.
- The [continuous-integration/drone/pr](https://drone.nextcloud.com/nextcloud/android/6999/2/3) reports a failure, but when I look at [the report](https://www.kaminsky.me/nc-dev/android-integrationTests/6999-IT-master-20-31/) everything looks fine. I do not understand what happened. :open_mouth: 

I have tried to make sure that the new methods defined on the `FileDao` interface maintain semantics of the old database calls, but would be good, if you could review them carefully.

I have explicitly tested behaviour of the `getFolderContent`, `getGalleryItems` and `moveLocalFile` methods. Regarding other changed methods - I have not been able to find relevant scenarios easily, but I have not noticed any obvious regressions in the app behaviour.

Please suggest next steps.

app/src/main/java/com/nextcloud/client/database/dao/FileDao.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    // TODO - there should be only one entry - does it make sense to add order by to preserve previous behaviour?
+    @Query("SELECT * FROM filelist WHERE media_path = :path AND file_owner = :fileOwner LIMIT 1")
+    fun getFileByLocalPath(path: String, fileOwner: String): FileEntity?
+
+    // TODO - there should be only one entry - does it make sense to add order by to preserve previous behaviour?
+    @Query("SELECT * FROM filelist WHERE remote_id = :remoteId AND file_owner = :fileOwner LIMIT 1")
+    fun getFileByRemoteId(remoteId: String, fileOwner: String): FileEntity?
+
+    @Query("SELECT * FROM filelist WHERE parent = :parentId ORDER BY " + ProviderTableMeta.FILE_DEFAULT_SORT_ORDER)
```
AlvaroBrey, line 51:
```suggestion
    @Query("SELECT * FROM filelist WHERE parent = :parentId ORDER BY ${ProviderTableMeta.FILE_DEFAULT_SORT_ORDER}")
```
Using string templates (and thus keeping a single string) avoids breaking the Android Studio SQL language injection. This allows executing the queries from AS, for example.

AlvaroBrey (commented):
> I have underestimated the number of changes required to achieve this. disappointed I would now suggest to merge it in (more or less) current state.

I have managed to migrate most of the query operations, just enough to remove the old createFileInstance(cursor) method. So there is no need to maintain two very similar methods.

This is fine, thank you very  much for your contribution :) It's great progress already.

I've reviewed the code and have a few minor suggestions I am attaching here. I still have to do some smoketesting before merging

app/src/main/java/com/nextcloud/client/database/dao/FileDao.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.room.Query
+import com.nextcloud.client.database.entity.FileEntity
+import com.owncloud.android.db.ProviderMeta.ProviderTableMeta
+
+@Dao
+interface FileDao {
+    @Query("SELECT * FROM filelist WHERE _id = :id LIMIT 1")
+    fun getFileById(id: Long): FileEntity?
+
+    // TODO - there should be only one entry - does it make sense to add order by to preserve previous behaviour?
```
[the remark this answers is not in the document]
AlvaroBrey: Dang, you're right, I completely missed that.

The way I see it, if we have situation where one of those queries can return more than one value, the database is already in an inconsistent state,  and we should have no expectations of consistent behaviour (these column combinations should have UNIQUE constraints, but that's another problem). So I'm fine with leaving the single-element queries unsorted.

AlvaroBrey:
Smoketesting was successful! After pending comments are adressed and this is rebased, it'll be OK to merge from me. Thanks @starypatyk !

app/src/main/java/com/nextcloud/client/database/dao/FileDao.kt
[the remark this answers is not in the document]
starypatyk: OK - removed TODOs in https://github.com/nextcloud/android/pull/11098/commits/69344bd8fb15ea5a6d631fd2d5ce4966d04efa75

starypatyk:
> Smoketesting was successful! After pending comments are addressed and the branch is rebased, it'll be OK to merge from me. Thanks @starypatyk !

Great! :smile:

I have removed TODOs in `FileDao` (the last unresolved comment) and rebased.

AlvaroBrey:
Ready to merge from my side! The CI failures are unrelated.

@tobiasKaminsky pinging in case you want a last look before merging

AlvaroBrey:
Thank you very  much @starypatyk !

Issue #603, opened 2017-01-29
🔑 Support for client certificates


mlsxlist:
### Actual behaviour
In order to secure Nextcloud on TLS level, it would be good if the app could support client certificates. If the client certificate is not sent on handshake, the server prevents access to Nextcloud logon page. This would provide a second line of defense.

### Expected behaviour
Nextcloud app should support client certificates as other apps like caldav sync and carddav sync already do.

riesnerd:
Hello mlsxlist,

if this is of any use to the case, there is an open issue about this over at the owncloud github site:

https://github.com/owncloud/android/issues/163

Best wishes
riesnerd

tobiasKaminsky:
I would vote this "medium" to get this done when we (@mario or me) have time.
Everything that enhances the security of the app/nextcloud should have priority.
cc @mario ?

riesnerd:
Okay thanks! In some way this also enhances security. ;)

mario:
@tobiasKaminsky let me think about it, and I'll come back to you - we need to focus on fixing current issues first to make the app (more) usable than it is now - thought definitely this is something we want to do in the future.

riesnerd:
Okay thanks in advance!

seanlynch:
I would also very much like to see this, because it protects from potential issues with the login page as well as weak passwords and password guessing attacks.

lucacarangelo:
Any news on this ? This is a very important enhancement !

AndreasMettlen:
@mario let me know if you still need a test setup with client certificates enabled. In the meantime I will try to look into this (I haven't worked on android apps yet, but I can find my way around in php/java/python so I will give it a try)

mario:
@AndreasMettlen yes I do. Send me the required cert + url, server and pass to mario@nextcloud.com :)

AndreasMettlen:
@mario Did you have a chance yet to look into this ?

mbrescia:
Hi,
I really look forward for this enhancement.
In the meantime I've tried to patch the code on Android and it kinda works... even if it's ugly as hell 😈 
I've used the Android KeyChain features to allow the user to select a PKCS12 certificate and then stored the cert alias into the app preferences:

AuthenticatorActivity
```
private void checkOcServer() {
        if (mHostUrlInput != null) {
            uri = mHostUrlInput.getText().toString().trim();
            mOkButton.setEnabled(false);
            showRefreshButton(false);
        } else {
[...]

PR #11099, opened 2022-11-26
Support client certificates

stephanritscher:
This change adds support for configurations where the webserver requests a client certificate, e.g. via nginx configuration options ssl_client_certificate and ssl_verify_client.
The client certificate handling is done via InteractiveKeyManager which prompts for a client certificate from a file or the devices keystore.
    
This change is NOT about client certificate authentication to the nextcloud server instance. The regular authentication mechanisms will be used as soon as the communication on TLS level is established.
    
The change addresses ticket #603, while not being a generic solution IMO.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -225,7 +225,8 @@ dependencies {
     // dependencies for app building
     implementation 'androidx.multidex:multidex:2.0.1'
 //    implementation project('nextcloud-android-library')
-    implementation("com.github.nextcloud:android-library:$androidLibraryVersion") {
+    implementation 'com.github.stephanritscher:InteractiveKeyManager:0.1'
+    implementation ("com.github.stephanritscher:nextcloud-android-library:$androidLibraryVersion") {
         exclude group: 'org.ogce', module: 'xpp3' // unused in Android and brings wrong Junit version
     }
 
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -496,6 +496,11 @@
             android:name=".ui.preview.PreviewBitmapActivity"
             android:exported="false"
             android:theme="@style/Theme.ownCloud.OverlayGrey" />
+
+        <!-- InteractiveKeyManager -->
+        <activity
+            android:name="de.ritscher.ssl.SelectKeyStoreActivity"
+            android:theme="@android:style/Theme.Holo.Light.Dialog.NoActionBar" />
     </application>
 
     <queries>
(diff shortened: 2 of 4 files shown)


AlvaroBrey:
Library PR for reference: https://github.com/nextcloud/android-library/pull/1005

Hello @stephanritscher and thank you for your interest! I don't disagree with this feature in itself, but I'm not sure I am okay with the approach taken. Specifically, my concerns are:

- Relying on a third-party library, which is generally not a great idea for security stuff, and gives us less freedom when maintaining, and less control over UX. We've had some bad experiences with this already (e.g. the fido/u2f library)
- On the android-library side the InteractiveKeyManager is also added, which if I understand correctly will hold its own state, which is why there is no code to pass the certificate to the network client on the app side
- Following the previous one: we also have no control over the state of the certificates.

My initial assesment of this is, in a nutshell, that I don't think this is a bad idea, but this is not a way of implementing it that I would approve.

cc @tobiasKaminsky  @AndyScherzinger

AndyScherzinger:
Not an expert on the topic but I think the talk Android app has support for client-certificates so I think it is worth checking how it is implemented there and I would always root for a single approach used throughout the different clients for maintenance / knowledge reasons.

stephanritscher:
Hi @AlvaroBrey,

thanks for your comments - which I can understand in a way, but I have a slightly different way of looking at this.

My design intention for InteractiveKeyManager was a library which is easy to integrate into any app not supporting client certificates (I previously did this for dav5x which now supports client certificates as well as Munin-for-Android and monitclient) and easy to maintain the integration as patch. This boils down to maintaining certificate state in the library to reduce the API interface and to having a separate UI (which is not conform with the rest of the app). So the main points of your reasoning were decided on deliberately, but from different perspective (which I still consider valid).

My general assumption was to consider my work as private but publish it in case others find it useful (and rebasing/rebuiling the APK once or twice a year didn't hurt me). The PR the was requested of @Torqu3Wr3nch (which shows I wasn't quite wrong that this could be relevant).

Now basically you and @AndyScherzinger request a redevelopment of the feature, which is fine, and maybe I would judge the same in your position. I anticipated parts of it so I hesitated to create the PR until it was requested by a third party. Unfortunately I moved to Syncthing for (private) file sharing and Radicale3 as CardDAV/CalDAV server so I won't invest major efforts here.

Feel free to reuse code of mine, and yes, I agree with the discussion in #603 that this is an important feature (and I don't understand that only one of the nextcloud apps seems to support it).

I wish you all the best,
Stephan

AlvaroBrey:
Hey @stephanritscher, the general consensus internally is aligned with my original message, so I'll close this. Many thanks for your interest and goodwill, however. Cheers!

PR #11100, opened 2022-11-27
Refresh UI only when changes are detected on the server

starypatyk:
Related to #10783.

When changing folder in the app the file list is refreshed - i.e. `OCFileListFragment::listDirectory` and `OCFileListAdapter::swapDirectory` are invoked - after checking for changes on the server (ETag check).

This happens even in the case when there are no changes.

The refresh is executed on the main thread and blocks UI. At the first sight this seems unnecessary (folder contents has been shown already) and this PR is an attempt to avoid the UI refresh.

Any comments? I am not sure about potential negative side effects in other areas of the app.

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
@@ -252,7 +252,7 @@ public class RefreshFolderOperation extends RemoteOperation {
             mStorageManager.saveFile(mLocalFolder);
         }
 
-        if (!mSyncFullAccount) {
+        if (!mSyncFullAccount && mRemoteFolderChanged) {
             sendLocalBroadcast(
                 EVENT_SINGLE_FOLDER_CONTENTS_SYNCED, mLocalFolder.getRemotePath(), result
             );


tobiasKaminsky:
We use it here:
https://github.com/nextcloud/android/blob/f797260f76531487944a86d7fe5a49d27a520dde/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java#L1316

This is needed to handle e.g. a changed/revoked app token.

Refreshing the UI seems to be invoked/triggered at some other place.

tobiasKaminsky:
Reason might be that even though eTag did not changed (no actual file changed), it might be that sharees/favorites have changed.
Thus we need to redraw everything :S

There are plans that we also have an eTagMetadata, which changes in those cases, so that we then if both eTags are the same, can omit refresh.

starypatyk:
@tobiasKaminsky 

> Reason might be that even though eTag did not changed (no actual file changed), it might be that sharees/favorites have changed. Thus we need to redraw everything :S

I understand the flow of events as follows - see the picture below with annotated and coloured logcat output. Events on the main thread are in black, events on the background (refresh folder) thread in blue and indented, messages shortened for clarity. Log taken from application **without** proposed change.

obraz

- The user taps on a folder, and the app immediately displays cached version of the folder (first `listDirectory` call).
- From `onBrowsedTo` the `RefreshFolderOperation` is triggered on the background thread.
- The `RefreshFolderOperation` thread sends a request to the server to check ETag.
- The `RefreshFolderOperation` thread detects that ETag matches - but even in that case sends the `EVENT_SINGLE_FOLDER_CONTENTS_SYNCED` broadcast notification **(this is what I want to change)**.
- The notification is received by the `SyncBroadcastReceiver` https://github.com/nextcloud/android/blob/4a0182ac383255ee3601867f3de52bda7315c1bd/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java#L1261 this causes first refresh of the UI in `fileListFragment.listDirectory` https://github.com/nextcloud/android/blob/4a0182ac383255ee3601867f3de52bda7315c1bd/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java#L1306 **before** checking for type of notification (second `listDirectory` call).
- The `RefreshFolderOperation` thread sends another request to the server to retrieve the shares.
- When response from the server is received, the `RefreshFolderOperation` thread updates the share information in the local database.
- Finally the `RefreshFolderOperation` thread sends another broadcast notification `EVENT_SINGLE_FOLDER_SHARES_SYNCED`.
- This notification is received by the `SyncBroadcastReceiver` and the UI is refreshed **for the second time**  (third `listDirectory` call).

It seems to me that the first UI refresh should happen only when the ETag does not match. If share information is changed on the server, the second refresh will be sufficient.

Do I understand it incorrectly?

tobiasKaminsky:
> It seems to me that the first UI refresh should happen only when the ETag does not match. If share information is changed on the server, the second refresh will be sufficient.
> 
> Do I understand it incorrectly?

Yes, this I overlooked!
You are right.

AlvaroBrey (approved):
Works great, thanks @starypatyk !

AlvaroBrey:
Thanks for all your contributions @starypatyk !

Issue #10725, opened 2022-09-08
IndexOutOfBoundsException in Favorite view having duplicate entries: `EmptyRecyclerView`


myxor:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Open Favorite view in app
2. Scroll down

### Cause of error
```java
Exception in thread "main" java.lang.IndexOutOfBoundsException: Inconsistency detected. Invalid view holder adapter positionOCFileListItemViewHolder{845b55b position=16 id=-1, oldPos=-1, pLpos:-1 no parent} com.owncloud.android.ui.EmptyRecyclerView{4c12421 VFED..... .......D 0,0-1080,2134 #7f0a0286 app:id/list_root}, adapter:com.owncloud.android.ui.adapter.OCFileListAdapter@8466ba, layout:androidx.recyclerview.widget.LinearLayoutManager@d88996b, context:com.owncloud.android.ui.activity.FileDisplayActivity@f567254
    at androidx.recyclerview.widget.RecyclerView$Recycler.validateViewHolderForOffsetPosition(RecyclerView.java:6156)
    at androidx.recyclerview.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline(RecyclerView.java:6339)
    at androidx.recyclerview.widget.GapWorker.prefetchPositionWithDeadline(GapWorker.java:288)
    at androidx.recyclerview.widget.GapWorker.flushTaskWithDeadline(GapWorker.java:345)
    at androidx.recyclerview.widget.GapWorker.flushTasksWithDeadline(GapWorker.java:361)
    at androidx.recyclerview.widget.GapWorker.prefetch(GapWorker.java:368)
    at androidx.recyclerview.widget.GapWorker.run(GapWorker.java:399)
    at android.os.Handler.handleCallback(Handler.java:938)
    at android.os.Handler.dispatchMessage(Handler.java:99)
    at android.os.Looper.loop(Looper.java:223)
    at android.app.ActivityThread.main(ActivityThread.java:7664)
    at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #11112, opened 2022-11-29
Resolved - IndexOutOfBoundsException in Favorite view

ayushbhanu:

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.java
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.java
@@ -149,7 +149,9 @@ public class ExtendedListFragment extends Fragment implements
     private ListFragmentBinding binding;
 
     protected void setRecyclerViewAdapter(RecyclerView.Adapter recyclerViewAdapter) {
+         mRecyclerView.getRecycledViewPool().clear();
         mRecyclerView.setAdapter(recyclerViewAdapter);
+        mRecyclerView.getAdapter().notifyDataSetChanged();
     }
 
     protected RecyclerView getRecyclerView() {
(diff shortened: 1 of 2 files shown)


AlvaroBrey:
I don't think it's a good idea to just swallow exceptions instead of finding their cause and adressing it. In #10725 you mention this is caused by "a bug in recycler view which is still not resolved", do you have any references for that? 

You also say that `RecyclerView Data modified in different thread`, do you have any idea which part of the Nextcloud code is changing RV data in background? That should be easy to fix if it's located.

Additionally, clearing the RV pool is just not a good idea. The RV will handle that more efficiently.

Can you provide a way to reproduce the crash (without this patch)?

ayushbhanu:
> I don't think it's a good idea to just swallow exceptions instead of finding their cause and adressing it. In #10725 you mention this is caused by "a bug in recycler view which is still not resolved", do you have any references for that?
> 
> You also say that `RecyclerView Data modified in different thread`, do you have any idea which part of the Nextcloud code is changing RV data in background? That should be easy to fix if it's located.
> 
> Additionally, clearing the RV pool is just not a good idea. The RV will handle that more efficiently.
> 
> Can you provide a way to reproduce the crash (without this patch)?


For the first questions you can also have a look on below stackoverflow link , its is a popular issue.
link - https://stackoverflow.com/questions/31759171/recyclerview-and-java-lang-indexoutofboundsexception-inconsistency-detected-in?page=1&tab=scoredesc#tab-top

Regarding code for thread i am not sure about it but what I understood is that this crash is caused by RV internal threads and above link helps in catching that.

This crash is occuring sometimes when a user scrolls the screen , there is not an exact scenerio to reproduce it.

AndyScherzinger:
Fixed in https://github.com/nextcloud/android/pull/12357 and https://github.com/nextcloud/android/pull/11519

Issue #9855, opened 2022-02-17
status dialog: not able to set only "icon" / only "clear at"


mahibi:
### Steps to reproduce
1. set only icon or only clear at without to enter status message
2. click set staus

### Expected behaviour
- icon or clear at are set

### Actual behaviour
- icon or clear at not set

### Environment data
Nextcloud app version: 3.19.0

# solution
api seems to set single modification for icon or "clear at" only if status message is not empty (must be at least whitespace)

see android talk commit https://github.com/nextcloud/talk-android/pull/1668/commits/a89c5952fa2747c8877c7882a6106311257722bf

joshtrichards:
If someone wants to pick this up, the now closed linked PR from @gitstart can probably still be adapted.

Samalbert977:
doesn't seem reproducible. Setting only "icon" or "clear status after" without status text works. Tested on Nextcloud v33.0.0 on Samsung Android 16 (non-GPlaystore version).

KartikeyTripathi469:
I’d like to work on an issue. Could you please assign one to me?

PR #11147, opened 2022-12-07
Status dialog: not able to set only "icon" / only "clear at"

gitstart:
Fixes https://github.com/nextcloud/android/issues/9855 

This PR will allow users to set a status with an emoji alone without a status message.

The emoji cannot be empty as the server will return an error if the icon is not valid, see  [here](https://github.com/nextcloud/server/blob/d8d15c3c2b8aee01fa2b534414f49378d2295f4f/apps/user_status/lib/Service/StatusService.php#L327)

### To Test

1. Click on user avatar at the top right corner.
2. Tap on Set Status
3. Select an emoji as the status and do not enter a status message
4. Tap on `Set Status Message`

### Issue
https://www.loom.com/share/7d471c92a8404625beaaa6a263e620bd

### Fix
https://www.loom.com/share/a8673ff71ee1413ea9de9c900e41bfec

### DIFF
--- a/app/src/main/java/com/nextcloud/ui/SetStatusDialogFragment.kt
+++ b/app/src/main/java/com/nextcloud/ui/SetStatusDialogFragment.kt
@@ -217,7 +217,7 @@ class SetStatusDialogFragment :
     private fun updateCurrentStatusViews(it: Status) {
         binding.emoji.setText(it.icon)
         binding.customStatusInput.text?.clear()
-        binding.customStatusInput.setText(it.message)
+        binding.customStatusInput.setText(it.message?.ifBlank {it.message?.trim()})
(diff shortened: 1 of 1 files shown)


AlvaroBrey (changes requested):
Hey @gitstart ! Thanks for your contribution, but it looks like your source branch contains all of its changes in a single merge commit. This disallows rebasing and makes the Github checks we have function incorrectly (for example, this PR passes the DCO check even though it's not signed off).

Can you fix your branch so that it contains a normal commit instead of a merge commit? Cheers!

gitstart:
@AlvaroBrey 
Due to some technical glitch, we encountered this issue. We will push all the PRs again.

Issue #9230, opened 2021-11-09
Use SplashScreen API


tobiasKaminsky:
See: https://developer.android.com/about/versions/12/behavior-changes-all#splash-screen-migration

- [ ] make sure that our system (also branding) works and looks nice

PR #11152, opened 2022-12-08
Use SplashScreen API

gitstart:
Fixes https://github.com/nextcloud/android/issues/9230

### To Test

1. Test that the splash screen is the same as before the refactor 

### Before
[next-before.webm](https://user-images.githubusercontent.com/93866435/205010355-46d49458-3515-463c-a559-b6bef5b86171.webm)
### After
[next-after.webm](https://user-images.githubusercontent.com/93866435/205012620-bf6fea94-c04f-4181-b04c-f2e44e26353c.webm)

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -281,12 +281,18 @@
     </style>
 
     <!-- Launch screen -->
-    <style name="Theme.ownCloud.Launcher">
+    <style name="Theme.AppSplash" parent="Theme.SplashScreen">
+        <item name="windowSplashScreenAnimationDuration">300</item>
+        <item name="android:windowBackground">@drawable/launch_screen</item>
+        <item name="postSplashScreenTheme">@style/Theme.AppSplash</item>
+
+        <!-- Status bar and Nav bar configs -->
         <item name="android:statusBarColor">@color/primary</item>
         <item name="android:navigationBarColor">@color/primary</item>
-        <item name="android:windowBackground">@drawable/launch_screen</item>
+        <item name="android:windowLightStatusBar">true</item>
         <item name="android:textColorHint">@color/secondary_text_color</item>
     </style>
+
 
     <style name="Theme.ownCloud.Widget.ActionBar" parent="@style/Theme.MaterialComponents.Light.DarkActionBar.Bridge">
         <item name="android:background">@color/primary</item>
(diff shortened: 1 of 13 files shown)


tobiasKaminsky:
Wow. Coool!

tobiasKaminsky (approved):

app/src/main/res/values/styles.xml
```diff
@@ -281,13 +281,19 @@
     </style>
 
     <!-- Launch screen -->
-    <style name="Theme.ownCloud.Launcher">
+    <style name="Theme.AppSplash" parent="Theme.SplashScreen">
+        <item name="windowSplashScreenAnimationDuration">300</item>
+        <item name="android:windowBackground">@drawable/launch_screen</item>
```
AlvaroBrey, line 286: This theme is configured incorrectly according to the docs. Instead of setting the window background, you should use `windowSplashScreenAnimationIcon` for the icon (you can use `@drawable/ic_launcher_foreground` for this) and use `windowSplashScreenBackground` for the background.

app/src/main/res/values/styles.xml
```diff
@@ -281,13 +281,19 @@
...
+        <item name="postSplashScreenTheme">@style/Theme.AppSplash</item>
+
+        <!-- Status bar and Nav bar configs -->
         <item name="android:statusBarColor">@color/primary</item>
         <item name="android:navigationBarColor">@color/primary</item>
-        <item name="android:windowBackground">@drawable/launch_screen</item>
+        <item name="android:windowLightStatusBar">true</item>
         <item name="android:textColorHint">@color/secondary_text_color</item>
```
AlvaroBrey, line 293: Not sure if these are needed at all, if using the `windowSplashBackground` properly.

AlvaroBrey (changes requested):
Hi @gitstart and thanks for your interest in contributing!

However this is incorrectly applied, according to the docs: https://developer.android.com/reference/androidx/core/splashscreen/SplashScreen

As it stands, this only leaves the app functioning exactly the same as before, as the splash screen theme is ignored by explicitly setting window background and not correctly using the splash screen properties; and the theme is still manually switched on `FileDisplayActivity`'s `onCreate`

* There is  no call to `installSplashScreen` on `FileDisplayActivity`s `onCreate`.
* Additionally to this we should remove  the explicit `setTheme` in `FileDisplayActivity.java` as well, as the splashscreen lib will handle the theme change for us with `postSplashScreenTheme`.

gitstart:
@AlvaroBrey 
Due to some technical glitch, we encountered this issue. We will push all the PRs again.

PR #11158, opened 2022-12-09
Build unit and androidTest code before committing

tobiasKaminsky:

### DIFF
--- a/scripts/hooks/pre-commit
+++ b/scripts/hooks/pre-commit
@@ -12,3 +12,8 @@ if ! ./gradlew --daemon detekt &>/dev/null; then
     echo >&2 "Detekt failed! See report at file://$(pwd)/app/build/reports/detekt/detekt.html"
     exit 1
 fi
+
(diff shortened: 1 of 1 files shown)


AlvaroBrey (commented):
In my opinion this slows down development too much, and we will end up skipping pre-commit hooks :\

Are we really uploading code with compilation errors so frequently that it's worth it to slow down every commit?

tobiasKaminsky:
> Are we really uploading code with compilation errors so frequently that it's worth it to slow down every commit?

For me it happens "every time" when refactoring stuff.

Is there a way to have this optional?

AlvaroBrey:
> Is there a way to have this optional?

I mean, you could have it depend on an environment variable, and then set that in your shell RC, or manually, for example.

Not sure if there's a better way. I'd at least move it to `pre-push` instead of `pre-commit`, though.

tobiasKaminsky:
Is indeed to much overhead.
As we now also want to have tests on every PR, chances that this happens are even less.

Issue #6248, opened 2020-06-08
Upload through app removes GPS data from images


AykutCevik:
### Steps to reproduce
1. Take a picture with a camera app which writes also GPS data into the picture
2. Ensure GPS data is written into the picture with different apps like file browser or gallery app
3. Upload this picture with the Nextcloud Android app version 3.11.1 via the share menu from gallery app
4. Download picture from Nextcloud instance with a browser and look at the GPS data of the file
5. See that the GPS data is missing or more likely set to zero

### Expected behaviour
- Pictures shouldn't lose their GPS data while uploaded through the app
![grafik](https://user-images.githubusercontent.com/1965251/84023644-be34d000-a988-11ea-90ad-45536866c31f.png)

### Actual behaviour
- Pictures are missing GPS data after upload
![grafik](https://user-images.githubusercontent.com/1965251/84023713-e3c1d980-a988-11ea-851a-5fee30530245.png)

### Can you reproduce this problem on https://try.nextcloud.com?
- Upload through the browser to Nextcloud works fine, just app removes the informations

### Environment data
Android version: 10 (OxygenOS 10.3.3)
Device model: OnePlus 6T
Stock or customized system: Stock
Nextcloud app version: 3.11.1
Nextcloud server version: 18.0.5
Reverse proxy: Nginx

#### Web server error log
```
Nothing worth to mention
```

#### Nextcloud log (data/nextcloud.log)
Only related to "Maps" app while trying to analyze the image since the GPS data is wrong.
```
Nothing worth to mention
```

I faced this issue since I saw that the pictures weren't showing up in the Maps app of Nextcloud. So this is not a new issue. Also I can surely say that the problem started before 20th of April, so also with a prior version. I can share pictures in a PM.

Appreciate any ideas for solving this issue.
Thank you for your great work!

strugee:
>     3\. Upload this picture with the Nextcloud Android app version 3.11.1

How are you uploading the picture? Auto Upload? The "Upload files" button? The "Upload content from other apps" button?

> Stock or customized system: FALSE

Does this mean you have a stock system? (I.e. the version of Android that came with your phone)

> Reverse proxy: FALSE

I believe this question is asking what web server you access Nextcloud through.

AykutCevik:
@strugee updated the issue.

Trying several other uploads, I found out that the issue may not be related to the Nextcloud app itself.

I've uploaded pictures through the share menu from both apps:
* OnePlus stock gallery app
[...]

PR #11161, opened 2022-12-12
potential fix for #6248

tobiasKaminsky:


AlvaroBrey (commented):
AFAIK this permission has to be requested at runtime, it's not automatically granted.

tobiasKaminsky:
> AFAIK this permission has to be requested at runtime, it's not automatically granted.

Seems to work, according to users in https://github.com/nextcloud/android/issues/6248

But here it is said that it "requires explicit user consent in order to gain access to this information":
https://developer.android.com/training/data-storage/shared/media#media-location-permission

joshtrichards:
I'm wondering if the need for this is only arising as a side effect of permitting users to *choose* to permit All Files Access (`MANAGE_EXTERNAL_STORAGE / ACTION_MANAGE_ALL_FILES_ACCESS_PERMISSION`)... or not.

What if we just don't let them choose and make it "You must say yes" to this intent action when you install if you want Nextcloud to function properly.

Storage permissions are a PITA to sort out with all the different Android versions. :cry: Worse if we support different combinations of permissions by default.

PR #11163, opened 2022-12-12
Test: explicit version for codeql action

AlvaroBrey:


AlvaroBrey:
Will be fixed in @nextcloud/android-config as this workflow is synced from there

PR #11227, opened 2023-01-05
Skip check for existing files when reading file data from local DB

starypatyk:
As discussed in #10783 and #11181.

It seems that the code to check for existing files when reading file data from the local DB is no longer needed.

In my "standard" benchmark (~330 files in a directory on a low-powered device), with this change  the average execution time of the `OCFileListFragment::listDirectory` call is reduced from **~320 ms** to **~280 ms** (but the timings are not very consistent).

I have not noticed any negative side effects, but to be honest, I have no idea how to really test it.

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
@@ -882,6 +882,19 @@ public class FileDataStorageManager {
         ocFile.setParentId(nullToZero(fileEntity.getParent()));
         ocFile.setMimeType(fileEntity.getContentType());
         ocFile.setStoragePath(fileEntity.getStoragePath());
+        if (ocFile.getStoragePath() == null && ocFile.isFolder()) {
+            // Apparently storagePath is filled only for regular files - even in the current (Jan 2022) implementation.
+            // Check below is still required for directories.
+            //
+            // try to find existing file and bind it with current account;
+            // with the current update of SynchronizeFolderOperation, this won't be
+            // necessary anymore after a full synchronization of the account
+            File file = new File(FileStorageUtils.getDefaultSavePathFor(user.getAccountName(), ocFile));
+            if (file.exists()) {
+                ocFile.setStoragePath(file.getAbsolutePath());
+                ocFile.setLastSyncDateForData(file.lastModified());
+            }
+        }
         ocFile.setFileLength(nullToZero(fileEntity.getContentLength()));
         ocFile.setCreationTimestamp(nullToZero(fileEntity.getCreation()));
         ocFile.setModificationTimestamp(nullToZero(fileEntity.getModified()));
--- a/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
@@ -882,16 +882,6 @@ public class FileDataStorageManager {
         ocFile.setParentId(nullToZero(fileEntity.getParent()));
         ocFile.setMimeType(fileEntity.getContentType());
         ocFile.setStoragePath(fileEntity.getStoragePath());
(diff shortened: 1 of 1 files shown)


starypatyk:
One of the test failures seems to be directly related to the changes:
[com.owncloud.android.FileIT - testRenameFolder](https://www.kaminsky.me/nc-dev/android-integrationTests/7270-IT-stable-23-21/com.owncloud.android.FileIT.html)

I have tried to repeat the issue in the app (renaming folder manually), but this works fine for me.

I am not sure yet, if the test assumes too much or if the storagePath should be set somewhere to compensate for the removed code.

Other failures in [EndToEndRandomIT](https://www.kaminsky.me/nc-dev/android-integrationTests/7270-IT-stable-23-21/com.nextcloud.client.EndToEndRandomIT.html) seem unrelated - at least I do not see any connection.

On the other hand the failure in [DialogFragmentIT](https://www.kaminsky.me/nc-dev/android-integrationTests/11227-Screenshot-blue-Light-21-35/com.owncloud.android.ui.dialog.DialogFragmentIT.html#testAccountChooserDialog) might be somehow related. It might be caused by a change of the timing between test execution in `DialogFragmentIT::testAccountChooserDialog` and delayed execution in the `FileDisplayActivity::startSyncFolderOperation`.

starypatyk:
It turned out that the `file.exists` check is still required for directories.

When a folder is synced locally `storagePath` is set for files, but not for directories.

With the original change the **On Device** pane stopped showing folders. :-1: 

I had to partially revert the previous commit and leave the check for directories. I hope this will also fix the failing **testRenameFolder** test.

Assuming that in the typical situation there are much more files than folders, this still gives similar performance improvement as the first attempt.

BTW - I have found one reason why I get inconsistent timings. At some point after starting the app, the JIT compiler decides to optimize the `FileDao_Impl.getFolderContent` method:

`Compiler allocated 4MB to compile java.util.List com.nextcloud.client.database.dao.FileDao_Impl.getFolderContent(long)
`

This influences the results significantly. Previously I had not paid attention to this message, so I got skewed results depending on the moment the method got optimized. :disappointed:

AlvaroBrey:
> At some point after starting the app, the JIT compiler decides to optimize the `FileDao_Impl.getFolderContent` method

If this kind of thing interests you, you may want to look into [Baseline profiles](https://developer.android.com/topic/performance/baselineprofiles/overview) so that critical methods like this one get compiled at install time.

starypatyk:
> If this kind of thing interests you, you may want to look into [Baseline profiles](https://developer.android.com/topic/performance/baselineprofiles/overview) so that critical methods like this one get compiled at install time.

Interesting indeed. :wink: For now, there are still a few potential "regular" improvements in the pipeline. Would be good to learn this at some point.

> I had to partially revert the previous commit and leave the check for directories. I hope this will also fix the failing testRenameFolder test.

If I see correctly the **testRenameFolder** test is passing now. But I am struggling to find the results...

starypatyk:
I have repeated the baseline tests and tests with the changes from this PR - this time making sure that in all cases the tests are executed **after** JIT decides to optimize the `FileDao_Impl.getFolderContent` method.

Method | Baseline - avg ms | PR - avg ms
-- | -- | --
OCFileListFragment::listDirectory | 291 | 234
FileDataStorageManager::getFolderContent | 168 | 112

We are gaining **~55 ms** when skipping ~330 `file.exists` calls.

AlvaroBrey:
> When a folder is synced locally `storagePath` is set for files, but not for directories

I wonder why it's computed but not saved. @tobiasKaminsky any insights here?

Other than that, this is :+1:  for me

AlvaroBrey (approved):

AlvaroBrey:
OK to merge this for now, thanks @starypatyk . I still don't know why the path is not stored for folders, but we'll go ahead with the current check.

AlvaroBrey:
Thanks @starypatyk !

PR #11228, opened 2023-01-06
E2E setup is only possible with internet connection

tobiasKaminsky:
Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

- [x] Tests written, or not not needed

Issue #11120, opened 2022-11-30
Upgrade screenshots and test emulators


AlvaroBrey:
Screenshots/test emulators are still runing android 27, which is by now several versions behind mainstream (at this moment android 12 - sdk 31). We should fix that.

PR #11248, opened 2023-01-12
Enable overriding injections in AndroidInjection classes during connected tests

AlvaroBrey:
Add testing utilities to override injections in classes injected with `AndroidInjection`. This includes Activities, Fragments, Services, etc. Other classes should use constructor injection,and as such shouldn't need this.
    
This works by overriding the androidInjector method in the `Application` class (new `TestMainApp` added) and allowing to use custom injectors there.
    
A `InjectionOverrideRule` is included for ease of use, but for more complicated cases, injection can be controlled by accessing `TestMainApp` directly.
    
A sample is included in `InjectionTestActivityTest`.

Intended for its use in #11228

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -96,7 +96,7 @@ android {
         }
 
         // arguments to be passed to functional tests
-        testInstrumentationRunner "com.nextcloud.test.ScreenshotTestRunner"
+        testInstrumentationRunner "com.nextcloud.client.ScreenshotTestRunner"
         testInstrumentationRunnerArgument "TEST_SERVER_URL", "${NC_TEST_SERVER_BASEURL}"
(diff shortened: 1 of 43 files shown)


AlvaroBrey:
Note: there's a bunch of noise in the PR because I also moved a lot of classes to the `com.nextcloud.test` package, in order to not have them mixed up with actual test code.

tobiasKaminsky (approved):

AlvaroBrey:
I'm guessing that last commit (https://github.com/nextcloud/android/pull/11248/commits/07d2a0c91bf219ac12e546321b776759e789754f) also explains what happened at https://github.com/nextcloud/android/pull/11242

DexOpener doesn't actually open anything outside of the package the runner is in (in this case `com.nextcloud.client`), so we can't mock any final class outside that, which is a bit stupid. Considering this, we may want to just skip tests that need that kind of mock when running in Android < 28. For this we should complete #11120  first

PR #11255, opened 2023-01-14
Use Room in FileDataStorageManager::saveSharesInFolder

starypatyk:
As diagnosed in #10783, one of the time consuming operations when navigating between folders is saving new share data in the local DB.

I have decided to convert the process to use the Room infrastructure - hoping to improve performance and get better insight on the actual operations taking place.

This initial implementation performs exactly the same database operations as the original code - at least as far as I can tell (with one small exception described below).

Performance is improved significantly. In my regular benchmark (~330 files in a directory on a low-powered device) the execution time of the `FileDataStorageManager::saveSharesInFolder` call is reduced from **~1600 ms** to **~200 ms**.

Regarding insight - I did learn a lot... :wink:

The whole process is divided in 3 major phases - all wrapped in a single database transaction (although in the original implementation, the first step was executed outside of transaction).

- The first step is resetting share flags (`share_by_link` and `shared_via_users`) for all files in the folder (`resetShareFlagsInFolder` method). This is done using a single `UPDATE` query on the `filelist` table.
- The second step is removing share information for all files in the folder from the `ocshares` table (`removeSharesInFolder` method). This is done by first getting the list of files in the folder and then deleting shares for each file with a separate query. Surprisingly, most of the time is spent here in a `getFolderContent` call, rather than in the actual update queries.
- Finally new share information is written back to the `ocshares` table (`insertShares` method). For each inserted share one of the flags (`share_by_link` or `shared_via_users`) is set on an entry in the `filelist` table. Originally the code to update the flags was buried in the `FileContentProvider` class. :open_mouth: Please also note that the `sharees` field of the `filelist` table **is not touched** during this operation - despite the fact that required share information is available at this point.

I have also noticed a discrepancy regarding the `user_id` field of the `ocshares` table. This is an integer field in the local DB. However, the user ID values retrieved from the server and stored in `OCShare.userId` are alphanumeric (contain the internal user name, **not** any numeric ID). Apparently the original code silently used 0 in this case when conversion to integer failed. Room requires strict conversion, so I have marked this with a TODO comment in the `createShareEntity` method.

Most of the other changes should be easy to follow.

It is probably too early to merge, since this leaves duplication of a few very similar methods in the `FileDataStorageManager` class.

Any comments/suggestions?


AlvaroBrey:
Hey @starypatyk, just wanted to let you know that I haven't had time to review this and will likely not have it until next week. Sorry and thanks!

starypatyk:
@AlvaroBrey 

> Hey @starypatyk, just wanted to let you know that I haven't had time to review this and will likely not have it until next week. Sorry and thanks!

No worries - there is no rush. :wink: I am moving quite slowly with this anyway.

On the other hand, I think it would be nice to have #11100 and #11227 merged - if possible before the next release.

starypatyk:
Hi @AlvaroBrey,

Still no rush... :wink:

I just wanted to mention that I plan to get back to this PR. Before continuing, I have decided to rebase. I did it manually via a sequence of cherry-picks, to make sure that I properly resolve conflicts in the Room migration code.

FYI - I have added a migration to fix an error in the definition of the `user_id` field of the `ocshares` table. However, in the meantime there were other migrations added, so I had to merge them manually.

starypatyk:
Hi @AlvaroBrey @tobiasKaminsky,

I have started this PR with an idea to refactor only the `FileDataStorageManager::saveSharesInFolder` method. However it turned out that the changes can be quite easily propagated to other methods handling local database operations on the `ocshares` table.

I think it would be reasonable to merge it once all methods in `FileDataStorageManager` that **write** data to the `ocshares` table are converted to Room.

I have completed conversion of most such methods. The only one remaining is the `saveSharesDB`.
https://github.com/nextcloud/android/blob/fb395f3dd12d2172290b0e4a8b57dff74154cee4/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java#L1315

I wanted to run the code to check if I have migrated the method correctly, but it seems that this method is **never** invoked during regular use of the application.

I have found that the `saveSharesDB` is invoked only from `GetSharesForFileOperation::run`. The `GetSharesForFileOperation` is referenced in a few places in the application code, but it is **created** only in test code in `GetSharesForFileOperationIT`.

I have two questions:

- Are my findings correct?
- If yes, does it make sense to maintain the test for code that is never invoked from the UI?

I would suggest to remove the `GetSharesForFileOperation` entirely. If you agree, I would remove it in a separate PR and got it merged before this one, as this would make completion of this PR much simpler.

AlvaroBrey:
> Are my findings correct?

From the looks of it yes, `GetSharesForFileOperation` seems unused, and as such I'm in favor of removing it. @tobiasKaminsky opinion?

starypatyk:
Hi @tobiasKaminsky,

I would like to go forward with this PR - as there are quite nice performance improvements pending... :wink:

> Performance is improved significantly. In my regular benchmark (~330 files in a directory on a low-powered device) the execution time of the FileDataStorageManager::saveSharesInFolder call is reduced from ~1600 ms to ~200 ms.

I need your input regarding the `GetSharesForFileOperation` class that apparently is never created in regular application code - only in tests (see comments above for details).

Do you agree to remove the class? If yes - in this PR or in a separate PR?

pachulo:
gently ping @tobiasKaminsky as this really looks like a nice improvement

Issue #8154, opened 2021-03-11
'Processing' indicator for AutoUpload


jeffWelling:
### Is your feature request related to a problem? Please describe.
When you tell the NextCloud app to AutoUpload a camera dir with >3000 images in it, the app needs to process each of the images before it begins uploading but there is no visual indicator or mention that this processing needs to be done. This creates an experience where the user tells the app to upload a directory and from the user's perspective nothing is happening, potentially leading them to conclude the feature is broken. In fact, the app is busy processing the files and will begin uploading them shortly.

### Describe the solution you'd like
A visual indicator or notification to the user that the app is processing the requested uploads and that file transfers will be visible in the Uploads section only when processing is completed.

### Describe alternatives you've considered
Leaving it as is can lead reasonable users to conclude that the feature is broken when it is not.

### Additional context
I had problems with autoupload, and from the Issues page I can see that I'm by no means the only one.
I've done some further testing, and I think I've narrowed down the behavior I'm seeing, and perhaps others are seeing too.

**Background**
When Google announced changes to their google photos offering, I began looking for a new solution. However I've already been using this phone for a couple years now, so the photos and videos on it that I want to back are numerous. There's about 15K items that I want to backup, about 60GB. I suspect the situation of having a large directory to upload from your phone to NextCloud is not uncommon.

My workflow when I was testing out NextCloud was as follows:
- Install Nextcloud on my Android device
- Login to Nextcloud on Android device
- Explore a bit, find the AutoUpload feature
- Configured AutoUpload to upload most of dirs on my phone with files I wanted uploaded (60GB of camera images, whatsapp memes and videos, demo videos from slack, the works) all in one go.
- Waited for something to happen, nothing happened. Went to the Upload section, nothing was happening. Checked webserver logs, they were idle. 
- Went into AutoUpload and un-selected and then re-selected everything. Waited again, nothing happened anywhere that I could see on the phone or webserver. 
- Uninstalled NextCloud from my phone concluding that the one feature I was after - autoupload - was broken
- Pondered some more, chatted with a friend who used it before and tried to give it one more go
- This time I figure'd I'd start with one dir, and just for the hell of it lets start with the dir with the fewest images, the one with 42 images in it.
- Almost right away, I started seeing upload notifications on my phone and webserver logs flying past my screen, it was working!
- Ok, let's try another folder, a little bit larger, 330 images this time. 
- A longer delay than before, almost a minute or two this time, but eventually it began working too! 

[...]

PR #11257, opened 2023-01-15
Auto upload sync with progress notification

claudius-kienle:
When auto uploading folders containing thousands of files, the file system sync takes very long. During this sync operation in `FilesSyncWork` no progress is displayed to the user. 
To improve the user experience, this pull request adds a progress notification to the FilesSyncWorker which will display the current status of the sync operation. 

image

Info: I did not create the locale strings for the new notification yet.


ezaquarii (changes requested):
Although I can agree with the UX change idea, I see the following problems with execution:

1. file scanning work talks to UI via platform API, bypassing our business-specific facade API (`AppNotif...`)
2. despite it's name, it only shows db insertion progress, not the actual sync operation 
3. there is zero test coverage *

I'd advise against merging it, until design is addressed.

\* I could be a bit harsh here, because this change touches legacy code that suffers from design† issues; whatever the end result will be, we should at least explore some options and make informed decision

† This is a footnote to a footnote, which says something about complexity of the whole sync "situation". Some time ago we started an alternative file upload service. It is used for contacts backup for now and nothing else. It also has radically more "observable" design, making progress notifications easy and reliable. I think we never got even close to cleaning up the sync mess, but on the off chance it is on a roadmap, perhaps it's time to consider some move in that direction?

tobiasKaminsky:
I remember a discussion with Jan, where we talked about enhancing progress notifications for upload and there he stated, that ideally the app should "just" work and users not no have to care, as they can trust that everything works.
If not, then we shall show an info.
@jancborchardt is this still true? If so, then this PR would go into "wrong" direction.

hlovdal:
Related to issue #8154.

Issue #10882, opened 2022-10-21
Slow Media Tab View 


AndyXheli:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

Since this was implemented the media tab got really slow and dose not work. Photos take a long time to load, app freezes. 
https://github.com/nextcloud/android/pull/10008

This seems to happen more on users that have a lot of photos my users has 45GB + of photos and videos

### Expected behaviour

Should load images on the snap like the IOS client dose. 

### Actual behaviour

Slow and laggy 

### Android version

12

### Device brand and model

SM-G996U1 

### Stock or custom OS?

Stock

### Nextcloud android app version

3.22.1

### Nextcloud server version

24.0.6

### Using a reverse proxy?

No

omnath0:
It's super slow specially after the last update and crashed everytime on Samsung S20/22.

svenb1234:
And it flickers because it recalculates the view over and over again.

eraser2021999:
I have a local NextCloud instance accesible from the phone via WiFi. I used NextCloud on iOS and usually I selected to see only the photos from the current year. But on Android this is slow and it puzzles me, since they are already synced and saved offline on the device, so the the issue it doesn't seem related to transfer speed, but with the view rendering.

andyxheli:
Yes I agree there has always been an issue with the Android Media tab view. hopefully, Nextcloud will fix this but yes since the new version has been released it use less I stopped using it can't use it at all

andyxheli:
Im still having alot of issues with this even with https://github.com/nextcloud/android/pull/11082 applied to 3.23.0 

Please see attached recording, For me the Media tab is completely unusable. Hopefully there's a higher property on this as the app comply crashes   

https://user-images.githubusercontent.com/59488153/205330178-16b357e0-b2af-4a44-9404-d120c4462ed8.mp4

andyxheli:
Hi @AlvaroBrey just wanted to give you a quick update on this issue seems like it's a little better on loading the media on Version 3.23.1 but still lags and freezes

invario:
[...]

PR #11262, opened 2023-01-15
Slow Media Tab View #10882 improvement

eraser2021999:
Signed-off-by: eraser2021999 bcoticopol@gmail.com

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/asynctasks/GallerySearchTask.java
+++ b/app/src/main/java/com/owncloud/android/ui/asynctasks/GallerySearchTask.java
@@ -143,56 +143,37 @@ public class GallerySearchTask extends AsyncTask<Void, Void, GallerySearchTask.R
 
     private boolean parseMedia(long startDate, long endDate, List<Object> remoteFiles) {
         // retrieve all between startDate and endDate
-        Map<String, OCFile> localFilesMap = RefreshFolderOperation.prefillLocalFilesMap(null,
-                                                                                        storageManager.getGalleryItems(startDate * 1000L,
-                                                                                                                       endDate * 1000L));
-        List<OCFile> filesToAdd = new ArrayList<>();
-        List<OCFile> filesToUpdate = new ArrayList<>();
+        Map<String, OCFile> localFilesMap = RefreshFolderOperation.prefillLocalFilesMap(null, storageManager.getGalleryItems(startDate * 1000L, endDate * 1000L));
+        int filesAdded = 0, filesUpdated = 0, filesDeleted = 0;
 
-        OCFile localFile;
         for (Object file : remoteFiles) {
             OCFile ocFile = FileStorageUtils.fillOCFile((RemoteFile) file);
-
-            localFile = localFilesMap.remove(ocFile.getRemotePath());
+            OCFile localFile = localFilesMap.remove(ocFile.getRemotePath());
 
             if (localFile == null) {
                 // add new file
-                filesToAdd.add(ocFile);
+                filesAdded++;
             } else if (!localFile.getEtag().equals(ocFile.getEtag())) {
                 // update file
                 ocFile.setLastSyncDateForData(System.currentTimeMillis());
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/ui/asynctasks/GallerySearchTask.java
```diff
@@ -143,56 +143,37 @@ private long findLastTimestamp(ArrayList<RemoteFile> remoteFiles) {
...
                 // add new file
-                filesToAdd.add(ocFile);
+                filesAdded++;
             } else if (!localFile.getEtag().equals(ocFile.getEtag())) {
                 // update file
                 ocFile.setLastSyncDateForData(System.currentTimeMillis());
-                filesToUpdate.add(ocFile);
+                filesUpdated++;
             }
+            storageManager.saveFile(ocFile);
```
AlvaroBrey, line 161: This is quite smart. It likely won't reduce execution time, but it will decrease memory usage by not holding the lists.

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
```diff
@@ -260,8 +260,6 @@ public void searchCompleted(boolean emptySearch, long lastTimeStamp) {
         }
 
         startDate = endDate - (daySpan * 24 * 60 * 60);
-
-        runGallerySearchTask();
```
AlvaroBrey, line 264: Gallery seems to load correctly when scrolling even without this. Any insight as to why this was added @tobiasKaminsky ?

AlvaroBrey (commented):
Looks OK, the gallery seems to still be loading properly even with the removed `runGallerySearchTask` call.

szaimen:
Is there a way to test this PR? I'd like to give it a shot since my gallery view has a lot of performance problems on my instance...

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
eraser2021999: Based on what I was able to understand from the code, the  `searchCompleted()` method does some subsequent loading after the initial load. I assume when we first enter the screen we get the first 30 days photos and then the next 30 days and so on. 
My impression why the screen still works without that call is because the `runGallerySearchTask()` is called also in `loadMoreWhenEndReached()` method. So basically when we scroll and reach the end (or almost), there is a new data transfer.

AlvaroBrey: Hmm, so that may mean we're not properly loading data if there's no data in the initial periods.

For example, if we have no photos from the last 30 days, but we have some from earlier, then the recyclerview would be empty on first load, and I'm not sure if that triggers loadMoreWhenEndReached()

eraser2021999: If it's empty, you can pull to refresh. Not sure if it works tho.

However, your remark makes me think of another approach: currently the start/end date is loaded based on the current time. 
```
// private void searchAndDisplay()
startDate = (System.currentTimeMillis() / 1000) - 30 * 24 * 60 * 60;
 endDate = System.currentTimeMillis() / 1000;

runGallerySearchTask();
```
Wouldn't be better to adapt the application to work with the timeframe of the library, and not with an arbitrary timeframe we choose? We could retrieve the most recent date from the media library and compute the 30 days period based on that. 
Or, going more into the rabbit hole, we could just pull an arbitrary amount of items (I believe currently is limited at 300 items) and ignore the date and time all together when we retrieve the data. We can use the timestamps only for displaying purposes then (similar with the interface of nc-photos).

eraser2021999: I did some checking and I discovered that the `runGallerySearchTask()` **IS** required if the timestamps don't match the filtering. I will analyze further and see if I can come up with something better.

PR #11273, opened 2023-01-18
build: Bump compileSdk to 33

AlvaroBrey:
Now required by several libraries (check open dependabot PRs)

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/ExternalSiteWebView.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/ExternalSiteWebView.java
@@ -193,9 +193,8 @@ public class ExternalSiteWebView extends FileActivity {
         webSettings.setDomStorageEnabled(true);
 
         // caching disabled in debug mode
-        if ((getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE) == 0) {
-            webSettings.setAppCacheEnabled(true);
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/owncloud/android/ui/activity/ExternalSiteWebView.java
```diff
@@ -193,9 +193,8 @@ private void setupWebSettings(WebSettings webSettings) {
         webSettings.setDomStorageEnabled(true);
 
         // caching disabled in debug mode
-        if ((getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE) == 0) {
-            webSettings.setAppCacheEnabled(true);
-            webSettings.setAppCachePath(getCacheDir().getPath());
+        if ((getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE) != 0) {
+            getWebView().getSettings().setCacheMode(WebSettings.LOAD_NO_CACHE);
         }
```
AlvaroBrey, line 198: The `appCache` methods were deprecated for a long time, are a no-op on Webview > 95 (according to the docs) and have finally been removed in compileSDK 33.

The webview now caches by default; I have altered this block so that it does again what the comment says, which is disabling the cache in debug mode.

tobiasKaminsky (approved):

PR #11287, opened 2023-01-24
upon upgrade localID may be null, better use default value

tobiasKaminsky:

### DIFF
--- a/app/src/main/java/com/owncloud/android/db/ProviderMeta.java
+++ b/app/src/main/java/com/owncloud/android/db/ProviderMeta.java
@@ -35,7 +35,7 @@ import java.util.List;
  */
 public class ProviderMeta {
     public static final String DB_NAME = "filelist";
-    public static final int DB_VERSION = 67;
+    public static final int DB_VERSION = 68;
 
     private ProviderMeta() {
         // No instance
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/owncloud/android/db/ProviderMeta.java
```diff
@@ -35,7 +35,7 @@
  */
 public class ProviderMeta {
     public static final String DB_NAME = "filelist";
-    public static final int DB_VERSION = 67;
+    public static final int DB_VERSION = 68;
```
AlvaroBrey, line 38: Missing the new json schema, and the `AutoMigration` in `NextcloudDatabase`

AlvaroBrey (changes requested):

app/src/main/java/com/owncloud/android/db/ProviderMeta.java
AlvaroBrey: In fact, this (changing null to not null) cannot be done with an automigration. A manual migration must be used.

AlvaroBrey (approved):

galvalarzen:
No solution?

PR #11201, opened 2022-12-23
Multi-page PDF scanner

AlvaroBrey:
Includes a refactor to make the optional feature a separate module, avoiding code duplication between variants as much as possible.

- [x] Request multiple pages from scanner
- [x] Preview scanned pages
- [x] Upload those pages at the same time
- [x] Generate PDF from images
- [x] Choice between PDF (single doc) or JPEG (multiple pages)
- [x] Tests written, or not not needed
    - Might write more depending on time

Fixes #11150
Fixes #10167

Related: #11021, although not fixed completely for now

PR #11296, opened 2023-01-25
Bumps for 3.24.0 RC1

AlvaroBrey:
I want to get the pdf scanner (#11201) into 3.24 as well, but I guess that should go to RC2 as it's still pending review.

### DIFF
--- a/scripts/analysis/analysis-wrapper.sh
+++ b/scripts/analysis/analysis-wrapper.sh
@@ -7,7 +7,7 @@ BUILD_NUMBER=$4
 PR_NUMBER=$5
 
 
-stableBranch="master"
+stableBranch="stable-3.24"
 repository="android"
 
 ruby scripts/analysis/lint-up.rb
--- a/scripts/analysis/getBranchName.sh
+++ b/scripts/analysis/getBranchName.sh
@@ -5,5 +5,5 @@
 if [ -z $3 ] ; then
     echo "master";
 else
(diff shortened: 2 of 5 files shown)


scripts/analysis/analysis-wrapper.sh
```diff
@@ -7,7 +7,7 @@ BUILD_NUMBER=$4
 PR_NUMBER=$5
 
 
-stableBranch="master"
+stableBranch="stable-3.24"
```
tobiasKaminsky, line 10: This should not be changed on master branch, but on stable-3.24 branch.

scripts/analysis/getBranchName.sh
```diff
@@ -5,5 +5,5 @@
 if [ -z $3 ] ; then
     echo "master";
 else
-    curl 2>/dev/null -u $1:$2 https://api.github.com/repos/nextcloud/android/pulls/$3 | grep \"ref\": | grep -v '"master"' | cut -d"\"" -f4
+    curl 2>/dev/null -u $1:$2 https://api.github.com/repos/nextcloud/android/pulls/$3 | grep \"ref\": | grep -v '"stable-3.24"' | cut -d"\"" -f4
```
tobiasKaminsky, line 8: Same here

tobiasKaminsky (changes requested):
Wrong target branch?

AlvaroBrey:
> Wrong target branch?

Yep, fixed. Good catch!

tobiasKaminsky (approved):

AndyScherzinger (approved):

PR #11301, opened 2023-01-25
Fastfile: fix check for screenshots

AlvaroBrey:
We now have the app icon in the fastlane folder, so this check fails.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -259,7 +259,7 @@ end
 
 desc "check if screenshots exists and exit"
 private_lane :checkIfScreenshotsExist do
-    sh(" if [ -e  metadata/android/*/images ] ; then echo 'Screenshots in fastlane folder exist; aborting!' ; exit 1 ; fi")
+    sh(" if [ $(find metadata/android/*/images -type f -not -name icon.png  | grep -c . ) -gt 0 ] ; then echo 'Screenshots in fastlane folder exist; aborting!' ; exit 1 ; fi")
 end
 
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
This check is/was to prevent me from accidentally uploading old screenshots.
Maybe we want use:
find metadata/android/*/images -type f | grep -vc icon.png

So that only "icon.png" is allowed.

AlvaroBrey:
> This check is/was to prevent me from accidentally uploading old screenshots. Maybe we want use: find metadata/android/*/images -type f | grep -vc icon.png
> 
> So that only "icon.png" is allowed.

Added back and fixed

fastlane/Fastfile
```diff
@@ -258,8 +258,8 @@ private_lane :checkLibrary_Final do
 end
 
 desc "check if screenshots exists and exit"
-private_lane :checkIfScreenshotsExist do
-    sh(" if [ -e  metadata/android/*/images ] ; then echo 'Screenshots in fastlane folder exist; aborting!' ; exit 1 ; fi")
+private lane :checkIfScreenshotsExist do
```
tobiasKaminsky, line 261: is this private lane vs private_lane working?

tobiasKaminsky (approved):

fastlane/Fastfile
AlvaroBrey: Oops, no. That was a typo as I made it public to test it, and then messed up when restoring it. Fixed.

PR #11302, opened 2023-01-25
Drone: update server images to use php 8

AlvaroBrey:
Should fix continuous failures on master

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -29,7 +29,7 @@ steps:
 
 services:
   - name: server
-    image: ghcr.io/nextcloud/continuous-integration-server:latest # also change in updateScreenshots.sh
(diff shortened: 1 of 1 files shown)


tobiasKaminsky (approved):

.drone.yml
```diff
@@ -29,7 +29,7 @@ steps:
 
 services:
   - name: server
-    image: ghcr.io/nextcloud/continuous-integration-server:latest # also change in updateScreenshots.sh
+    image: ghcr.io/nextcloud/continuous-integration-shallow-server:latest # also change in updateScreenshots.sh
```
nickvergessen, line 32: The actual fix would be to fix the docker-ci image so it installs the required php...
Now there is still the broken server image available and you switch to something completely different and then wonder why it fails?

PR #11354, opened 2023-02-06
Fix F-Droid metadata to differentiate dev version more

AlvaroBrey:
- Move per-flavor fastlane metadata to top-level src/ folder
- Add dev icon for F-Droid (dev version only)
- Capitalize name for dev version app


tobiasKaminsky:
In scripts/metadata/generate_metadata.py line 14/15 needs to be adjusted (and tested).

AlvaroBrey:
> In scripts/metadata/generate_metadata.py line 14/15 needs to be adjusted (and tested).

Done! I actually wondered how the metadata was synced from transifex

tobiasKaminsky (approved):

PR #11367, opened 2023-02-09
When scrollbar is shown we need more space to let user click on three dots menu

tobiasKaminsky:
@AlvaroBrey not sure if this is the best way, as it now removes space on right side all the time…


AlvaroBrey (changes requested):
Does not work in my tests, it's still impossible to click the menu with an open scrollbar

AlvaroBrey:
@tobiasKaminsky shall we close this?

PR #11402, opened 2023-02-20
when resuming note directly after creating, url is not working anymore, thus we need a new one

tobiasKaminsky:
@AlvaroBrey no idea how to test this :S


AlvaroBrey:
> @AlvaroBrey no idea how to test this :S
> 


Can you give me steps to reproduce this (manually?) maybe that way I can figure something out

tobiasKaminsky:
Right now I only have a way to reproduce this on Samsung
- create new text file
- wait until it is loaded
- open NC app in popup mode (this is then recreating entire fragment; while minimizing/restoring only resumes it)

PR #11436, opened 2023-03-03
PoC: modifying searchable behaviour at runtime via DI

AlvaroBrey:
@tobiasKaminsky


tobiasKaminsky:
Merged into e2e sharing branch, thus closing.
Many thanks again!

Issue #1964, opened 2018-01-15
Push: show real text instead of generic "new notification"


tobiasKaminsky:
User want to see the real content of notification.
As I understand this was due to a server bug, but can now be fixed for next version.

mario:
Haven't I fixed this already? xD

tobiasKaminsky:
It shows the push subject, but not the content.
Would it also be possible to show the complete content (as many text as fits into the notification?)

tobiasKaminsky:
Working.

skjnldsv:
Is this released?
![Screenshot_20201102-083829](https://user-images.githubusercontent.com/14975046/97848202-06b2a880-1cf1-11eb-83ca-e493ca3bfd72.png)

tobiasKaminsky:
Yes, since ages.
If you get this, it means that for some reason the content cannot be decrypted.
Can you check if there are wrong app tokens in web->settings?

tobiasKaminsky:
After debugging with @skjnldsv it turns out that those strange notifications only happens when the app is not in foreground.

MartinConsultingServicesInc:
I'm having this problem, too.  Running client version 3.14.0 against server 19.0.5.  I think it started after the upgrade from 19.0.4.

scroom:
> I'm having this problem, too. Running client version 3.14.0 against server 19.0.5. I think it started after the upgrade from 19.0.4.

Same here.

tobiasKaminsky:
As far as I remember new_notification is the default value of push proxy, if no message is set.
@nickvergessen do you have an idea why such messages might be sent?

MartinConsultingServicesInc:
I thought it was occurring when NC apps had updates available, but I got 
one of these this morning and I don't have any apps needing updates.

I'm not familiar with the language NC is written in, but it may have 
something similar to the Perl "caller" routine, which allows a routine 
to determine who called it. If you have such a thing, you might consider 
changing the default value to include the caller name or other 
identifier. That would make future occurrences easier to debug.

If I learn anything about the source, I'll update the ticket.

             -Brian M

tobiasKaminsky:
Have you removed account in question on Android or reset app token or something like that?

MartinConsultingServicesInc:
I had not made any change there.  I know that the Nextcloud client still 
allowed me to access my files, make changes, etc. That suggests that the 
account credentials were operating normally, at least in some regards.

Nevertheless, I have now deleted the Android client account from the 
[...]

PR #11437, opened 2023-03-03
Notifications: workaround when failing to fetch complete notification info from server

AlvaroBrey:
This is a workaround for #1964. @tobiasKaminsky I leave it up to discussion whether we should close that issue or not.

In order to prevent the `NEW_NOTIFICATION` text, what I'm doing is detect when that's the title and content of the notification,
and displaying a generic "New notification" instead, which when clicked leads to the Notifications activity.

To test this:

```shell
php occ notification:test-push test # should result in normal test notification
php occ php occ notification:generate -l NEW_NOTIFICATION -- test NEW_NOTIFICATION # should result in a generic "new notification"
```

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/NotificationWork.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/NotificationWork.kt
@@ -33,6 +33,7 @@ import android.graphics.BitmapFactory
 import android.media.RingtoneManager
 import android.text.TextUtils
 import android.util.Base64
+import android.util.Log
 import androidx.core.app.NotificationCompat
 import androidx.core.app.NotificationManagerCompat
 import androidx.work.Worker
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
I am not sure if this approach works.
Last time I digged into this, I had the impression that this Android is receiving the push notification and then cannot pass it to NC app and thus this generic one is shown.

I now just looked again into push proxy, and I can see this:
https://github.com/nextcloud-gmbh/push-notification-proxy/blob/6564e324dec5fe12bcfb203a71d6d13d1d2480e8/notification-handling.go#L273

It seems that for some reason the wrong deviceType is set and thus fcm.Notification (L272) is sent, instead of fcm.Message(L264).

AlvaroBrey:
> I am not sure if this approach works. Last time I digged into this, I had the impression that this Android is receiving the push notification and then cannot pass it to NC app and thus this generic one is shown.

This PR is not intended to fix notification fetching/decryption. It just detects that the string `NEW_NOTIFICATION` is about to be shown in a notification, and changes that for a more normal "New notification" which opens the Notifications screen when clicked.

In other words this is an UX improvement when notification fails, not an attempt to fix the failure.

PR #11446, opened 2023-03-10
Fix types after kotlin port of some library classes

AlvaroBrey:
For https://github.com/nextcloud/android-library/pull/1083


AndyScherzinger (approved):

AlvaroBrey:
Spotbugs is not really increasing as I did not add any nullable types; nullability was just unknown beforehand. So it can just be ignored

AndyScherzinger (approved):

PR #11458, opened 2023-03-17
Replaced fileVisitor in automatic file upload with a more resilient system against file permission problems

Bensikrac:
Replaced external dependency with internal code, since external file walker produced nullpointer exceptions on files where permissions where missing, and therefore auto upload was failing often. (especially /android/data and /android/obb if you wanted to mirror entire phone (/storage/emulated/0))
This does not fix upload not refreshing / registering and other bugs.
This only was my personal problem, but I hope it will be helpful for others.
And I have absolutely no idea how to test this properly. (It works on my phone and in android emulator)


tobiasKaminsky:
Thanks for this contribution!
If you want test this, I think something like
- create some files / folders on sdcard (remember them)
- create syncedFolder
- run your recursive method
- check if those files/folders are in provider

If you need help, you can also allow me write changes to your branch and I am happy to help :)

Bensikrac:
I did write some tests now, which should work, but only for the code I changed, since testing for media folders works differently, which I didn't figure out yet.

Bensikrac:
I am out of ideas, why the test fails on your testing setup, while it flawlessly creates the files (I see them in ADB file explorer) and adds them to the sync list. Could you please tell me, what the gradle task for your online test is called in android studio?

Bensikrac:
I still can't manage to reproduce the error, even when running the command above. I am doing the tests in the android emulator. What settings are you using for the emulator in the automated test? (Android version, API version, Android image, Device?)

PR #11474, opened 2023-03-24
Set minSdkVersion to 24 + remove checks

tobiasKaminsky:


tobiasKaminsky:
Moving the assets breaks my AS.
We seem to have anydpi-v26/ic_launcher.xml and this cannot access drawable/ files, but needs drawable-v26 or lower.
Strangely, it compiles still via CLI.

So either we try to fix it (not sure how) or we revert it.

AndyScherzinger:
@tobiasKaminsky as mentioned build fine for me.

AndyScherzinger (approved):

AndyScherzinger:
@tobiasKaminsky what to do here? I expect this to be a AS caching issue of some sorts and that a fresh clone of the repo would solve the matter for you.

ZetaTom:
Looks good to me; build succeeds and runs fine on my machine.

One thing I've noticed, is that the app icon for the QA build now also reads *Dev* instead of *QA*. (see `app/src/qa/res/drawable/ic_launcher_foreground.xml`)

PR #11498, opened 2023-04-02
Implementation of  granularity for autoupload subfolders

fservida:
Hello,
I was tackling the implementation of autoupload subfolders to give the user choice of which granularity to use, either only the year, year/month or year/month/day

I've done similar work for the iOS app (https://github.com/nextcloud/ios/pull/2305) and it worked well in my tests, even though not yet merged I'm now rewriting it for the last update.

I wanted to implement the same on android, following the same general idea.

**However** I'm stuck even before starting; the first step I wanted to implement was the interface and handling of the setting, but I can't seem to get the radio buttons work ok.
The idea would be to have the month radio button selected by default (as it is current behaviour) and let user choose if needed one of the others.
You can see the changes in logic I made, but I'm not sure I understood all the code flow.
If somebody with knowledge of this part of code could help get past the settings problem, I can then try to edit the rest of the code accordingly.

Also, not that used to code for android, so let open to any suggestions.

- [ ] Tests not yet written for year and day granularity (only edited current for default)

### DIFF
--- a/app/src/androidTest/java/com/nextcloud/client/SyncedFoldersActivityIT.java
+++ b/app/src/androidTest/java/com/nextcloud/client/SyncedFoldersActivityIT.java
@@ -65,6 +65,7 @@ public class SyncedFoldersActivityIT extends AbstractIT {
                                                                    false,
                                                                    false,
                                                                    true,
+                                                                   1,
                                                                    "test@https://nextcloud.localhost",
                                                                    0,
                                                                    0,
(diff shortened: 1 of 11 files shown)


AndyXheli:
@AndyScherzinger Hope your doing well :) I tested this featue under the ios app and it works perfect thank you @fservida anyway you can put some prioty on this maybe someone from the dev can help 🙏🏼

PR #11520, opened 2023-04-13
Camera permission rationale

surinder-tsys:
**Pre condition:**
App camera permission should not enabled

**Step to reproduce:**
1. Launch the  app and login
2. Click on '+' icon and select "Upload from camera", Permission pop will open and click on "Allow". (error message coming).
3. Again run step 2 to verify permission applied or not.(Now not asking any permission and open camera)

Actual result:
After allow permission applied successfully but error message coming to set permission manually.

Expected result:
Error message should not come if permission set successfully

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -462,6 +462,9 @@ public class FileDisplayActivity extends FileActivity
                 if (grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
                     // permission was granted
                     getFileOperationsHelper().uploadFromCamera(this, FileDisplayActivity.REQUEST_CODE__UPLOAD_FROM_CAMERA);
+                } else if (!shouldShowRequestPermissionRationale(permissions[0])) {
+                    // user CHECKED "never ask again"
+                    DisplayUtils.showSnackMessage(this, R.string.camera_permission_rationale);
                 }
                 break;
             default:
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -443,6 +443,9 @@ public void onRequestPermissionsResult(int requestCode, @NonNull String[] permis
                     // permission was granted
                     getFileOperationsHelper()
                         .uploadFromCamera(this, FileDisplayActivity.REQUEST_CODE__UPLOAD_FROM_CAMERA);
+                } else if (!shouldShowRequestPermissionRationale(permissions[0])) {
+                    // user CHECKED "never ask again"
+                    DisplayUtils.showSnackMessage(this, R.string.camera_permission_rationale);
```
tobiasKaminsky, line 448: With which Android did you test this?
With Android 10 I only have "allow" or "deny"

surinder-tsys: Check with Android 9.
There is radio button for "Don't ask again".
Screenshot 2023-04-18 100444

tobiasKaminsky: image

Tested with Android 9 emulator.

Maybe this is vendor specific?

surinder-tsys: Yes that can might be the case because we have found this issue in Huawei device.
The issue will only occur in devices where there is option to deny the permission by checking "Don't ask again" (for lower os versions) or  "Don't allow" (for OS 11+) devices.

In both the cases no error snackbar shown to the user.

For more test you can check on Android 11+ devices and select "Don't allow". And again if you click on camera option then no permission dialog and message will be shown.

alperozturk96:
@tobiasKaminsky @surinder-tsys I tested with Android 14 emulator. After tap the don't allow, user already not seeing any error message.


https://github.com/nextcloud/android/assets/67455295/0d836e45-e4df-4809-81f7-8b4f7de90af2

surinder-tsys:
This PR is no longer applicable. So closing the PR.

PR #11523, opened 2023-04-13
E2EE dialog crash on clicking "Set as Encrypted" multiple times

surinder-tsys:
**Steps to reproduce:**

1. Click on Set as encrypted option to do the encryption
2. Once the dialog shows do back press to remove the dialog
3. Repeat 1st and 2nd step this will lead to crash.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/dialog/SetupEncryptionDialogFragment.java
+++ b/app/src/main/java/com/owncloud/android/ui/dialog/SetupEncryptionDialogFragment.java
@@ -22,6 +22,7 @@ package com.owncloud.android.ui.dialog;
 
 import android.accounts.AccountManager;
 import android.app.Dialog;
+import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.AsyncTask;
@@ -50,6 +51,7 @@ import com.owncloud.android.utils.EncryptionUtils;
 import com.owncloud.android.utils.theme.ViewThemeUtils;
 
 import java.io.IOException;
+import java.lang.ref.WeakReference;
 import java.security.KeyPair;
 import java.security.PrivateKey;
 import java.util.ArrayList;
@@ -129,7 +131,7 @@ public class SetupEncryptionDialogFragment extends DialogFragment implements Inj
             viewThemeUtils.platform.colorTextButtons(positiveButton, neutralButton);
         }
 
-        task = new DownloadKeysAsyncTask();
+        task = new DownloadKeysAsyncTask(requireContext());
         task.execute();
     }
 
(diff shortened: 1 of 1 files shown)


ZetaTom:
> **Steps to reproduce:**
> 
>   1. Click on Set as encrypted option to do the encryption
>   2. Once the dialog shows do back press to remove the dialog
>   3. Repeat 1st and 2nd step this will lead to crash.

Using the current master branch (6bbd629), I'm unfortunately not able to observe this behaviour.

ZetaTom:
@surinder-tsys, could you please provide us with your configuration (device model, manufacturer, Android and app version) and check if you are still able to reproduce this behaviour? There have been some changes (#11694) made to this dialogue since this pull request was opened.

surinder-tsys:
@ZetaTom The crash is still reproducible and the other changes has no effect to it.
Its trickier to reproduce. Its not device specific as the crash is happening due to null Context.

To reproduce it --> On clicking "Set as encrypted" option do the back press(as quickly you can) before the api call finishes. 

Since AsyncTask is used for api call and if the dialog is not visible it will lead to crash in onPostExecute method. So using WeakReference for Context will fix this issue.

Refer screenshot: 
E2ee crash

app/src/main/java/com/owncloud/android/ui/dialog/SetupEncryptionDialogFragment.java
```diff
@@ -303,47 +313,58 @@ protected void onPreExecute() {
 
         @Override
         protected String doInBackground(Void... voids) {
+            Context context = mWeakContext.get();
             // fetch private/public key
             // if available
             //  - store public key
             //  - decrypt private key, store unencrypted private key in database
```
ZetaTom, line 321:
```suggestion
            // fetch private/public key
            // if available
            //  - store public key
            //  - decrypt private key, store unencrypted private key in database

            Context context = mWeakContext.get();
```

ZetaTom (commented):
@surinder-tsys, thank you for confirming the details.

After artificially increasing the network latency, I was able to reliably reproduce this issue and can now confirm that this change fixes the problem and works as expected.

app/src/main/java/com/owncloud/android/ui/dialog/SetupEncryptionDialogFragment.java
AndyScherzinger: Sry @ZetaTom I already merged a second ago. can you create a separate PR with the polishing, I can review and merge that one then. Sorry for the extra work.

ZetaTom: No problem, I've opened a new pull request: #11758

PR #11547, opened 2023-04-20
Configurable calendar backup.

surinder-tsys:
Calendar backup is now configurable with flag from setup.xml.

In MC, calendar backup is not required. 

### DIFF
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -80,6 +80,7 @@
     <string name="away">Abwesend</string>
     <string name="backup_settings">Sicherungseinstellungen</string>
     <string name="backup_title">Kontakte &amp; Kalender-Sicherung</string>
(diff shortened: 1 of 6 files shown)


app/src/main/res/values-de/strings.xml
```diff
@@ -80,6 +80,7 @@
     <string name="away">Abwesend</string>
     <string name="backup_settings">Sicherungseinstellungen</string>
     <string name="backup_title">Kontakte &amp; Kalender-Sicherung</string>
+    <string name="contact_backup_title">Kontakte-Sicherung</string>
```
tobiasKaminsky, line 83: They will be overwritten after next Transifex update.
This can be added via https://app.transifex.com/nextcloud/nextcloud/translate/#de/android/91098483

AndyScherzinger:
Successor is #11589

PR #11563, opened 2023-04-27
Download limit for file link share

surinder-tsys:
Implemented download limit for files where share type is Link.
Showing the download count once the download limit is specified.

Conditions to show download limit:
1. Only applicable for files
2. Only applicable where share type is link.
3. Not applicable for Internal, External or any other shares.

Requires the implementation of: https://github.com/nextcloud/files_downloadlimit

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/download_limit/DeleteShareDownloadLimitRemoteOperation.kt
+++ b/app/src/main/java/com/owncloud/android/operations/download_limit/DeleteShareDownloadLimitRemoteOperation.kt
@@ -1,0 +1,77 @@
+/**
+ * ownCloud Android client application
+ *
+ * @author TSI-mc Copyright (C) 2021 TSI-mc
+ *
+ *
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public
+ * License version 2, as published by the Free Software Foundation.
+ *
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+ * warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ *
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/owncloud/android/operations/download_limit/DeleteShareDownloadLimitRemoteOperation.kt
```diff
@@ -0,0 +1,81 @@
+/*
```
tobiasKaminsky, line 1: Please move those RemoteOperation to android-library.
If possible, also add real end to end tests, else I can help you with them.

tobiasKaminsky:
I asked https://github.com/nextcloud/files_downloadlimit/issues/173 to have a capability, as probably not that many server will have this enabled.

app/src/main/java/com/owncloud/android/operations/download_limit/DeleteShareDownloadLimitRemoteOperation.kt
surinder-tsys: Can you please grant me push access to android-library?

surinder-tsys: @tobiasKaminsky I was writing IT for the download limit but I am facing issue in executing them in android-library.
1. For parsing the test cases are running fine as there is no use of AbstractIT class.
2. But for Operations (Get, Post, Delete) we need AbstractIT to be extended and for this  the test cases are not executing. I tried existing test cases of NC but same for them.

Screenshot 2023-05-03 134704

Is there some configuration is required to execute them?

surinder-tsys: Code moved to android-library: https://github.com/nextcloud/android-library/pull/1123

PR #11588, opened 2023-05-04
Different folder icon for AutoUpload

AndyScherzinger:
We are showing different folder icon for AutoUpload. It will help user to differentiate between the folder which has been selected for AutoUpload.

Successor of #11543

## Screenshot
folder_icon


AndyScherzinger:
@tobiasKaminsky fine by me to merge, I replaced the icon to fit our style

@surinder-tsys just as a side-note since you have been using the "photos-folder" icon which I removed and replaced with a "auto upload named one". The reason is that the auto upload functionality supports different types the media ones but also "custom" defined folders that would upload _any_ file in there not just images or video. Hence we used the upload icon. You can of course use yours, but it will then also be applied to non-media folders.

app/src/main/java/com/owncloud/android/datamodel/SyncedFolderProvider.java
```diff
@@ -405,4 +406,47 @@ private ContentValues createContentValuesFromSyncedFolder(SyncedFolder syncedFol
...
+            ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_REMOTE_PATH + " LIKE ? AND " +
+                ProviderMeta.ProviderTableMeta.SYNCED_FOLDER_ACCOUNT + " =? ",
+            new String[]{"%" + remotePath + "%", user.getAccountName()},
+            null);
+
+        if (cursor != null && cursor.getCount() >= 1) {
+            result = true;
+        } else {
+            if (cursor == null) {
+                Log_OC.e(TAG, "Sync folder db cursor for remote path = " + remotePath + " in NULL.");
```
tobiasKaminsky, line 438: Why is this an error?
It is expected, that not every folder matches?

app/src/main/java/com/owncloud/android/ui/activity/EditorWebView.java
```diff
@@ -56,6 +61,11 @@ public abstract class EditorWebView extends ExternalSiteWebView {
 
     RichdocumentsWebviewBinding binding;
 
+    private SyncedFolderProvider syncedFolderProvider;
```
tobiasKaminsky, line 64: As we use this that often nowadays, this should be injected.

app/src/main/java/com/owncloud/android/datamodel/SyncedFolderProvider.java
surinder-tsys: This method is a copy of **findByLocalPathAndAccount()** this method. Due to which logic is also same.

PR #11593, opened 2023-05-05
Fix crash when changing orientation when bottom sheet dialog in preview image is open

tobiasKaminsky:


AndyScherzinger (approved):
tested successfully

PR #11606, opened 2023-05-10
i18n: Removal from translation

Valdnet:
Reported by user from Transifex.


AndyScherzinger (approved):

AndyScherzinger:
Nice catch and thanks for the quick PR @Valdnet 👍

PR #11644, opened 2023-05-23
Prevent having double entries after rotate

tobiasKaminsky:
To test
- open image 
- open menu
- rotate
- see no duplicated menu entries

### DIFF
--- a/app/src/main/java/com/nextcloud/ui/fileactions/FileActionsBottomSheet.kt
+++ b/app/src/main/java/com/nextcloud/ui/fileactions/FileActionsBottomSheet.kt
@@ -199,6 +199,8 @@ class FileActionsBottomSheet : BottomSheetDialogFragment(), Injectable {
     private fun displayActions(
         actions: List<FileAction>
     ) {
+        binding.fileActionsList.removeAllViews()
+
         actions.forEach { action ->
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nextcloud/ui/fileactions/FileActionsBottomSheet.kt
```diff
@@ -199,6 +199,8 @@ class FileActionsBottomSheet : BottomSheetDialogFragment(), Injectable {
     private fun displayActions(
         actions: List<FileAction>
     ) {
+        binding.fileActionsList.removeAllViews()
+
         actions.forEach { action ->
             val view = inflateActionView(action)
             binding.fileActionsList.addView(view)
```
ZetaTom, line 206: Works as expected.

Another approach could be to only add actions to the list when it is empty:
```suggestion
        if (binding.fileActionsList.isEmpty()) {
            actions.forEach { action ->
                val view = inflateActionView(action)
                binding.fileActionsList.addView(view)
            }
```

tobiasKaminsky: I like this idea!

ZetaTom (approved):

PR #11652, opened 2023-05-24
Typo in deletion dialog: it's content -> its content

jaller94:
If this needs to be edited on Transifex instead, please do so. I don't have an account and don't want to create one.

I just want this fixed and a PR seemed to be more precise than filing an issue.
Please consider any required sign-off, license or permission given to commit this typo fix in any commit of your liking.


AndyScherzinger (approved):

AndyScherzinger (changes requested):
Only the master file is manually maintained any other language like this one is auto synced from transifex to github, so this change is overridden over night, always.

AndyScherzinger:
@Valdnet @rakekniven could you take care of fixing the translation on transifex? Thanks in advance 🙏

AndyScherzinger:
Fixed on transifex, will come in tonight

PR #11689, opened 2023-06-02
Fix actions in folder picker

ZetaTom:
This fixes some minor display inconsistencies concerning the *Choose upload folder* dialogue when uploading a file from an external app. Additionally this change also affects the *Copy to* and *Move to* dialogues.

- removed *Select all*
- show create folder icon instead of single item overflow menu
- adjust theming in *Choose upload folder* dialogue

---


tobiasKaminsky:
Nice!
Can you check if you can write a second screenshot test with two accounts?

ZetaTom:
A second screenshot test for multiple accounts has been added.

AndyScherzinger (approved):

PR #11693, opened 2023-06-05
Notification list customization

surinder-tsys:
This PR contains the following:
1. Notification list item height changed.
2. List item icon color customized.
3. List item with 2 action buttons customized to Primary filled and only Text.

Refer screenshots:
|☀️Initial customization|☀️M3 blue|☀️M3 pink|🌙Initial customization|🌙M3 blue|🌙M3 pink|
|---|---|---|---|---|---|
||Screenshot_20230606_193424|Screenshot_20230606_224528||Screenshot_20230606_193412|Screenshot_20230606_224546|

Predecessor of: https://github.com/nextmcloud/android/pull/66


AndyScherzinger:
@surinder-tsys what is the motivation and idea behind changing the the height of the notification items? As you can see in your own screenshots, this has no effect on your item's rendering. I would suggest we leave it at `wrap_content` and given the padding of `16dp` this creates a spacing of `32dp` between 2 items in the list which should be perfectly fine, no? Can't tell what Jens' opinion is or was on this one.

I'll keep working on this PR since the change doesn't make use of the dynamic theming, hence the different colors of the buttons in your screenshot (purple and blue).

However a short feedback on the padding/item-height would be highly appreciated. Thanks in advance.

AndyScherzinger:
Also while it might seem to be the exact same color but different size, we now have 3 styles in place:
* headline color / size (16sp)
* text color / size (14sp)
* timestamp `#666666` / size (13sp)

@jancborchardt can we optimize this a bit more? In alignment with Jens of course

AndyScherzinger:
Furthermore I changed the text color to secondary text color which is dark/light aware to meet contrast level and is not locked down on `#666666` but uses a different shade for dark theme

AndyScherzinger:
Pushed all my updates and would personally be fine with merging it now. All my changes are reflected in the screenshots at the top and the coloring is now purely based on Material 3 spec, see https://m3.material.io/components/lists/specs for reference.

surinder-tsys:
> @surinder-tsys what is the motivation and idea behind changing the the height of the notification items? As you can see in your own screenshots, this has no effect on your item's rendering. I would suggest we leave it at `wrap_content` and given the padding of `16dp` this creates a spacing of `32dp` between 2 items in the list which should be perfectly fine, no? Can't tell what Jens' opinion is or was on this one.
> 
> I'll keep working on this PR since the change doesn't make use of the dynamic theming, hence the different colors of the buttons in your screenshot (purple and blue).
> 
> However a short feedback on the padding/item-height would be highly appreciated. Thanks in advance.

Agreed.  Thanks for taking care.
For action buttons we still be needing the blue color for MC. that I will handle in our internal PR.

Looks good to me too.

AndyScherzinger:
> For action buttons we still be needing the blue color for MC. that I will handle in our internal PR.

@surinder-tsys If you don't mind me asking, what is the reason or motivation behind having the buttons in blue? Especially regarding MC? Just asking out of curiosity since I would have expected a different color. Also from your original screenshot not all buttons would have been tinted in blue shades. So would it be all buttons to be blue or just some?

surinder-tsys:
@AndyScherzinger We received the design for Blue color only not the theme color(Magenta) and that's the reason we are using it. For detailed,  I guess @Jens will have the answer.

And talking about the other action buttons we won't be using in MC and they will never be visible so that's why we didnt customized them,

tobiasKaminsky:
@AndyScherzinger once we are done here, I'd like to add screenshot tests, so that notification with one, two and more actions are shown

jancborchardt (approved):
Seems good! :)

PR #11697, opened 2023-06-06
Update README.md with downloads widget

mathijs81:
Show how popular nextcloud is at the top of the readme page
Sorry about the previous PR, I somehow copied the wrong app, now it's right (1M+ downloads, and will automatically update when more downloads are reached). Is this interesting for you?


tobiasKaminsky:
@jancborchardt what do you think? Shall we link to a 3rd party for this?

mathijs81:
Nice that you are still taking a look at this! I've updated it to match the tip of current master.

jancborchardt (commented):
> @jancborchardt what do you think? Shall we link to a 3rd party for this?

We also linked to Codafy and others for this so it should be fine. Or is there an "official" source from the stores?

mathijs81:
I don't know about Apple, but we made it at AppBrain because Google doesn't provide anything like this for Google Play.

PR #11746, opened 2023-06-16
Add image editor

### LINKED ISSUE #11608: Add crop/resize functionality to photos
tbd

ZetaTom:
This change introduces an image editor, which enables users to crop, flip and rotate images on the fly. It supports a variety of different input formats, but is only able to produce JPEG, PNG and WEBP files as output due to the limitations of a system library.

ImageEditor

closes #11608

---


surinder-tsys:
@ZetaTom I had one observation with this editor is that the size difference is more whenever we rotate the images.
Here are some findings:- 527 kb to 164 kb, 10 Mb to 7 Mb and 20 Mb to 15 Mb etc..

Is this much size reduction is expected or is there any way that it can be handled?

ZetaTom:
@surinder-tsys Given the current implementation, this reduction in file size is to be expected as the image is re-compressed after each edit (after saving using :white_check_mark:). This is handled by the [Android-Image-Cropper](https://github.com/CanHub/Android-Image-Cropper) Library. To avoid overwriting the original image with a version of inferior quality, we decided that edited images should be saved to a new file.

It would be possible to implement rotation functionality without changing the actual image data by manipulating only the corresponding [EXIF](https://en.wikipedia.org/wiki/Exif) fields associated with the image. This would result in an (almost) identical file and file size, although it would be rotated by the image *viewer* itself.

surinder-tsys:
@ZetaTom Thanks for the explanation. 
So the size reduction can not be avoided using Image-Cropper and this is expected and nothing can be done? Did I understood correctly?

ZetaTom:
@surinder-tsys To my knowledge, we don't have any plans to change this behaviour.

PR #11761, opened 2023-06-21
PreviewMediaFragment: UI/UX improvements

ZetaTom:
This change affects media playback via the PreviewMediaFragment.

- allow media to be downloaded in Preview via the actions menu
- hide the *Pin to Home Screen* action in Preview
- fix exception when returning from external player
- autoplay videos only once
  - e.g. not when returning from an external app
- remember playback position when switching apps or opening *Details* view
- show logo when no audio thumbnail is available
  - used to be the colour as the background

---

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/preview/PreviewMediaFragment.java
+++ b/app/src/main/java/com/owncloud/android/ui/preview/PreviewMediaFragment.java
@@ -30,6 +30,7 @@ import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.Color;
+import android.graphics.drawable.Drawable;
 import android.media.MediaMetadataRetriever;
 import android.net.Uri;
 import android.os.AsyncTask;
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/ui/preview/PreviewMediaFragment.java
```diff
@@ -284,7 +288,15 @@ private void setThumbnailForAudio(OCFile file) {
...
-            binding.imagePreview.setImageResource(R.drawable.logo);
+            setGenericThumbnail();
+        }
+    }
+
+    private void setGenericThumbnail() {
+        Drawable logo = AppCompatResources.getDrawable(requireContext(), R.drawable.logo);
+        if (logo != null) {
+            DrawableCompat.setTint(logo, getResources().getColor(R.color.primary));
+            binding.imagePreview.setImageDrawable(logo);
```
ZetaTom, line 299: This might cause an issue, when the `icon` drawable differs from the default one and is not intended to be coloured.

ZetaTom: Should be fixed as of a767dee94c85e9da434da5e60abc37ffbef5a74b.

AndyScherzinger (approved):

PR #11775, opened 2023-06-26
Correctly update file after sharing

tobiasKaminsky:
Step to reproduce:
1. Open the NextCloud app and login
2. Click on share icon of any file- Image/video/audio/ office file
3. Click on "Link" from its popup
4. Now check the icon

Actual result:
Share-Icon is remain same . When refresh the page then icon gets change as link icon

Expected result:
Icon should get change after link share from its popup

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
@@ -740,6 +740,14 @@ public class OCFileListAdapter extends RecyclerView.Adapter<RecyclerView.ViewHol
         new Handler(Looper.getMainLooper()).post(this::notifyDataSetChanged);
     }
 
+    @SuppressLint("NotifyDataSetChanged")
+    public void replaceFile(OCFile file) {
+        mFiles.remove(file);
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -740,6 +740,14 @@ public void setData(List<Object> objects,
         new Handler(Looper.getMainLooper()).post(this::notifyDataSetChanged);
     }
 
+    @SuppressLint("NotifyDataSetChanged")
+    public void replaceFile(OCFile file) {
+        mFiles.remove(file);
+        mFiles.add(file);
+        mFiles = sortOrder.sortCloudFiles(mFiles);
+        notifyDataSetChanged();
+    }
```
ZetaTom, line 749:
```suggestion
    public void replaceFile(OCFile file) {
        mFiles.set(mFiles.indexOf(file), file);
        notifyItemChanged(file);
    }
```

If I'm not mistaken, the sorting order should remain the same. So it should suffice to replace and update only our selected file/item.

ZetaTom (approved):

PR #11788, opened 2023-07-04
Fix multi-selection in shared tab

ZetaTom:
This change addresses several issues related to multiple selection in the *Shared* tab.

Previously, when using *Select All*, files could be selected multiple times, leading to unexpected consequences. This behaviour should be fixed, along with other minor optimisations.

---


AndyScherzinger (approved):
tested successfully

PR #11817, opened 2023-07-19
Fix empty state for media view

surinder-tsys:
**Steps to reproduce:**
1. Select a folder which contains only videos or images
2. Now filter only videos or images from 3 dot menu
3. The list is not updated with empty state.

Or

1. Select a folder which has some media files
2. Now select any empty folder
3. Still previously selected media shows but not the current folder data (empty state)

**Cause:**
The reason behind this was not setting setHasFooter false for media view as there is no footer used and due to which EmptyRecyclerView always has 1 item count and doesn't hide the recyceler view when list is empty. 


AndyScherzinger (approved):
tested successfully

PR #11818, opened 2023-07-19
Avoid auth redirection when activity restarted

surinder-tsys:
Steps to reproduce:
User should not be logged in
1. If any activity sets **enableAccountHandling** to false.
2.  Take app to background
3. Now bring back app to foreground
4. **OnRestart** from base activity will trigger and will launch **AuthenticatorActivity** which overrides the previous activity for which enableAccountHandling is false.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/BaseActivity.java
@@ -109,7 +109,9 @@ public abstract class BaseActivity extends AppCompatActivity implements Injectab
     protected void onRestart() {
         Log_OC.v(TAG, "onRestart() start");
         super.onRestart();
-        mixinRegistry.onRestart();
+        if (enableAccountHandling) {
+            mixinRegistry.onRestart();
+        }
     }
 
     private void onThemeSettingsModeChanged() {


ZetaTom:
@surinder-tsys Unfortunately, I'm not sure what problem this pull request is supposed to fix. Could you please clarify what behaviour you were expecting and what you were observing instead?

surinder-tsys:
@ZetaTom If user is on **FirstRunActivity** and if the activity gets **restarted (pressing home screen and again opening app from Recents)** that time it will redirects to **AuthenticatorActivity**, because **enableAccountHandling** flag check is not applied in **onRestart** function.

ZetaTom (commented):
I've since been able to construct a scenario where this function is called in the required context. Unfortunately, I can't see any obvious change in the behaviour of the app and haven't been able to reproduce the problem described.

Since `enableAccountHandling` is checked in other lifecycle related functions, such as `onResume()`, I can't see why it shouldn't also be checked in `onRestart()`.

@surinder-tsys, could you check if this change is still up to date?

surinder-tsys:
No longer required as this has been addressed in https://github.com/nextcloud/android/pull/13902.

PR #11821, opened 2023-07-20
Fix folder picker empty state headline text

surinder-tsys:
This PR has the fix for empty state headline text while doing Folder Selection from FolderPickerActivity.


AndyScherzinger (approved):
LGTM 👍

PR #11815, opened 2023-07-18
Avoid multiple API calls on create button click

surinder-tsys:
**Actions Performed**
Logged in
1. Open app
2. Go to Files
3. Tap Plus-button and create a new Collabora Office Document / Spreadsheet / or Presentation
4. Enter some individual name for the fiel that doesn't exist yet
5. Double-tap on the "Create" button (not unrealistic speeed-tapping etc., just two successive touches on the "Create" button, which would easily happen for example if the finger bounces on the screen etc.)

**Expected Result**
There should not be a problem with the Create-button; successive activations of this button should be ignored by the app

**Actual Result**
The "Create" button is responsive to more than one activation (successive finger touches on button), although this action results in a crash every time
Furthermore duplicates of the file are created and saved in the cloud this way, like [filename], [filename] (1), ...

**Crash Logs:**
`Exception in thread "AsyncTask #4" java.lang.RuntimeException: An error occurred while executing doInBackground()
    at android.os.AsyncTask$4.done(AsyncTask.java:415)
    at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:383)
    at java.util.concurrent.FutureTask.setException(FutureTask.java:252)
    at java.util.concurrent.FutureTask.run(FutureTask.java:271)
    at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:305)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
    at java.lang.Thread.run(Thread.java:923)
Caused by: Exception in thread "AsyncTask #4" java.lang.IllegalStateException: Fragment ChooseRichDocumentsTemplateDialogFragment{fa07d9c} (e309cf32-a54c-46b8-96dc-4e03e92ed0a4) not attached to a context.
    at androidx.fragment.app.Fragment.requireContext(Fragment.java:972)
    at com.owncloud.android.ui.dialog.ChooseRichDocumentsTemplateDialogFragment$CreateFileFromTemplateTask.doInBackground(ChooseRichDocumentsTemplateDialogFragment.java:344)
    at com.owncloud.android.ui.dialog.ChooseRichDocumentsTemplateDialogFragment$CreateFileFromTemplateTask.doInBackground(ChooseRichDocumentsTemplateDialogFragment.java:309)
    at android.os.AsyncTask$3.call(AsyncTask.java:394)
[...]

PR #11840, opened 2023-07-26
Improve UX when creating rich document

ZetaTom:
This is an alternative solution to #11815. (see https://github.com/nextcloud/android/pull/11815#issuecomment-1651275217)

Previously it was possible to select *Create* multiple times when creating a rich document (+ → Create document/spreadsheet/presentation). This would either result in multiple files being created or a crash.

This change introduces a *Wait a moment...* dialogue to block the UI while providing a progress indicator.

---


AndyScherzinger (approved):
LGTM 👍

PR #11852, opened 2023-08-01
Extra options for uploading subfolders

### LINKED ISSUE #11642: Auto Upload create subfolders for year only
Nextcloud for IOS has the option to sort uploaded photos by year/month/day. But android only has the one option: year+month. Makes it a little janky when uploading iphone+android photos to the same folder. Some photos are in the year folder and some have a month folder.

Add auto upload option to create subfolders for year only. Similar option to what IOS has would be nice.

Taking the photos out of the month subfolders manually works but not particularly convenient. Or sorting both iphone+android by year+month makes it a little more tidy.

### Additional context
Pics attached of android vs IOS auto upload subfolder settings

Thanks!

IMG_1019
Screenshot_20230522_120055
Screenshot 2023-05-22 120447

dean36963:
resolves #11642 

Adds some extra options for uploading photos into various patterns of subfolders. yyyy or yyyy/MM or yyyy/MM/dd supported.

Currently just adds an extra field for this, and has to be used in conjunction with the "use subfolder" checkbox, the new option is greyed out and disabled when "use subfolder" is not checked.

## Notes

List of considerations to add:

- ✅ Does dropdown make sense UI/UX-wise? (Andy: fine I would say, while other option for upload behavior uses a dialog with a radio button group)
- ✅ Does the database migrations make sense? This is the first time I've done that sort of migration so probably needs extra checks.
- ✅ Unit tests! As there's now more ways of uploading files, should we have some new unit tests for checking new upload location is correct for all new options?
- ✅ Order of imports and line wrapping settings, I'm not sure if this is already covered by the idea settings files in git and android-studio doing this, or do I need to run somethings?
- ✅ Order of the dropdown isn't ideal visually, but I wanted to make sure in case after a migration, a 0 in the database should be the first option, so previous behavior is kept (though this makes an assumption that 0 is the default which might be incorrect).
- ✅ Localisation concerns? There's been a change to `prefs_instant_upload_path_use_subfolders_summary` (but only in one language), due to a change in behaviour, so text might not make sense any more in other languages.
- ✅ detectNewJavaFiles: should all new classes be Kotlin? (ah I may have forgotten copyright headers...)


AndyScherzinger:
@dean36963 would you mind adding some screenshots, so @nextcloud/designers can have a look? Thanks in advance 👍

dean36963:
Current screenshots:

Light theme:

| All Disabled | Subfolder Disabled | Subfolder Enabled | Dropdown | Selected |
|--------------|----------------------|-----------------------|-------------|---------|
| Screenshot_20230802-220051_Nextcloud | Screenshot_20230802-220057_Nextcloud | Screenshot_20230802-220105_Nextcloud | Screenshot_20230802-220111_Nextcloud | Screenshot_20230802-220116_Nextcloud |

Dark theme:

| All Disabled | Subfolder Disabled | Subfolder Enabled | Dropdown | Selected |
|--------------|----------------------|-----------------------|-------------|---------|
| Screenshot_20230802-212634_Nextcloud | Screenshot_20230802-212639_Nextcloud | Screenshot_20230802-212644_Nextcloud | Screenshot_20230802-212650_Nextcloud | Screenshot_20230802-212656_Nextcloud |

On looking again, it looks like each area has a bold looking title text, with lighter text below. Not sure if the spinner/dropdown text should also look lighter too, but then to me it looks like it might be greyed out and not selectable.

jancborchardt (commented):
Looks very nice @dean36963! :) Only one adjustment maybe, to make it a little bit less busy when "Use subfolders" is not checked, the "Subfolder options" could only be shown once that is checked?

dean36963:
| All Disabled | Subfolder Disabled | Subfolder Enabled | Dropdown | Selected |
|-- | -- | -- | -- | --|
| Screenshot_20230803-123830_Nextcloud | Screenshot_20230803-123837_Nextcloud | Screenshot_20230803-123844_Nextcloud | Screenshot_20230803-123848_Nextcloud | Screenshot_20230803-123853_Nextcloud |

app/src/main/java/com/nextcloud/client/database/NextcloudDatabase.kt
```diff
@@ -63,7 +63,8 @@ import com.owncloud.android.db.ProviderMeta
         AutoMigration(from = 65, to = 66),
         AutoMigration(from = 66, to = 67),
         AutoMigration(from = 68, to = 69),
-        AutoMigration(from = 69, to = 70)
+        AutoMigration(from = 69, to = 70),
+        AutoMigration(from = 71, to = 72)
```
AndyScherzinger, line 67: @tobiasKaminsky not needed, right? Since there is an actual migration?

app/src/main/java/com/nextcloud/client/database/entity/SyncedFolderEntity.kt
```diff
@@ -57,5 +57,7 @@ data class SyncedFolderEntity(
     @ColumnInfo(name = ProviderTableMeta.SYNCED_FOLDER_TYPE)
     val type: Int?,
     @ColumnInfo(name = ProviderTableMeta.SYNCED_FOLDER_HIDDEN)
-    val hidden: Int?
+    val hidden: Int?,
+    @ColumnInfo(name = ProviderTableMeta.SYNCED_FOLDER_SUBFOLDER_RULE)
+    val subFolderRule: Int?
```
AndyScherzinger, line 62: this one (new column, so no auto migration, right @tobiasKaminsky ?

AndyScherzinger:
Some minor comments and I'll bee needing some feedback from @tobiasKaminsky regarding the room migration code bits.
@jancborchardt are you fine with the dropdown for the folder options, see https://github.com/nextcloud/android/pull/11852#issuecomment-1664628577 - keeping in mind, the "What if the file already exists?" option uses a dialog with a radio button group.

marcoambrosini (approved):
Looks very nice :) 

A couple of non blocking comments regarding design: 

* I think that the complexity of these settings warrant the use of a **full page view as opposed to a dialog**. It will also work better on smaller mobile screens.
* There's a conceptual mismatch between the title of the page and the switch right next to it. I assume that the switch turns auto-upload on and off but the title says **"preferences for auto upload"**, so it's a little confusing. I think we should change the title to **"Auto upload"** only.

app/src/main/java/com/nextcloud/client/database/NextcloudDatabase.kt
dean36963: I tried reverting this section, and when I moved from a build before my changes, to a build including all my changes (including reverting this - not pushed) I saw that some auto upload folder settings were lost. Keeping this line as is currently committed seemed to keep folder settings as I'd expect when upgrading.

It might be that I'd need to revert this **and** also change another section, but I am unsure what that change would be.

AndyScherzinger: Can't say for sure and would need to check if and which changes have been made for 70,71,72 to know When to set the Auto Migration. 👍

tobiasKaminsky: Logic of Room is, afaik:
- if auto migrate is enabled, then it keeps entire DB, but changes/add new column
- if no auto migrate and no manual migration -> resets entire DB -> all is lost
- 72.json is for a new database (new app installation)
--> now the question is, do we need a manual migration, e.g. to add a specific value to hidden or subFolderRule? If not, then it should work with AutoMigration and 72.json as it is now.

dean36963: > now the question is, do we need a manual migration, e.g. to add a specific value to hidden or subFolderRule?

In this regard, I've currently got it so the previous option for this (YYYY/MM) is the first option in the dropdown and appears as the default value post migration (and just in general for new folders). So I think in it's current state, no default value is needing to be set.

However, if it's more sensible to reorder these options visually (yearly _then_ monthly, _then_ daily for example), it might be required to then do a manual migration?

tobiasKaminsky:
Rebased and good to go :shipit:

AndyScherzinger (approved):

tobiasKaminsky (approved):

AndyScherzinger:
Merged! Nice work @dean36963 🎉 And thanks for building and polishing it 💯

PR #11919, opened 2023-08-28
When deleting image in preview activity, move to the next image

### LINKED ISSUE #11918: When deleting image in preview activity, move to the next image
When going through a list of images to clean it up (removing unwanted images), I use the image preview and swipe left/right until I find an image to delete. When I delete the image (three dot menu -> delete), the preview is closed however. In the file list, I then need to find the image before or after the one I just deleted to continue the image preview.

When deleting image in preview activity, keep the activity open and move to the next image

-

### Additional context
Note: This behavior was also mentioned as one improvement suggestion in #4977

PhilLab:
Instead of closing the preview after the deletion of an image file, the next image is shown. This allows to conveniently go through a list of images and delete unwanted ones, without the flow being interrupted.

Closes #11918

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.java
@@ -232,7 +232,15 @@ public class PreviewImageActivity extends FileActivity implements
         super.onRemoteOperationFinish(operation, result);
 
         if (operation instanceof RemoveFileOperation) {
-            finish();
+            // initialize the pager with the new file list
+            initViewPager(getUser().get());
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
I would say it is worth perusing, also looping in @jancborchardt for feedback from a designer's perspective on the behavior introduced with this change.

ZetaTom (approved):
Works well and feels intuitive to use!

One incredibly minor thing I noticed is that the transition between the *Wait a moment...* dialogue and the next image opening can be quite harsh. It's also possible that my emulator is acting up. Either way, I'd suggest merging this as is and maybe addressing this later.

PR #11922, opened 2023-08-28
E2e sharing

tobiasKaminsky:


gcvl:
The menu item is there but it's not working at all: "Eliminazione non riuscita" (deletion failed).
Also, I erased the file manually but the app does not sync and I can still access the deleted file.

Issue #11836, opened 2023-07-25
README.md from root directory not clearing


jj49er:
The README.md from the root directory is staying when opening a new folder. The new folder README.md is showing on top of the root README.md

Android 12
Nextcloud version 3.25.0

parneet-guraya:
Hi, I solved this issue, Here's what was causing it -->
### Actual Behaviour ->
- `OCFileListFragment` which is the main view when launched, starts with fragment transaction (add) without call to addToBackStack(), It means no reverse transaction can happen thus pressing back exits the app (fragment goes in to onStop()).
- While launching `PreviewTextStringFragment` which opens when clicked upon `Readme.md` (preview at the top) adapter item , with transaction (replace) with addToBackStack(), so `replace` cause the `OCFileListFragment` to Destroy and `PreviewTextStringFragment` launches.
- This line [`else if (leftFragment instanceof PreviewTextStringFragment) {
            createMinFragments(null);
        }`](https://github.com/nextcloud/android/blob/266e27f9ddd5e33fa59d6ee2ee345ccb5549c3ae/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java#L1098) is the main issue first of all it's not calling super.onBackPressed(); which lead to non reversal of replace transaction, so neither the current `PreviewTextStringFragment` destroys nor `OCFileListFragment` gets created.

- What [createMinFragments(Bundle savedInstanceState)](https://github.com/nextcloud/android/blob/266e27f9ddd5e33fa59d6ee2ee345ccb5549c3ae/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java#L500) method doing is creating new `OCFileListFragment` with fragment transaction (add) without addToBackStack(), so no reversals.

- So, now when user click on Preview and it opens up and upon pressing back it does not destroy the Preview Fragment and also create new  `OCFileListFragment`, which shows us the Preview Fragment in the fragment and the new `OCFileListFragment` on top of it that's why we are seeing persistent Readme file in the background.

But, In fact we should be calling onBackPressed() as well as not calling the createMinFragments(null); since it's creating the new `OCFileListFragment` and  because if we call this one at this point (along with onBackPressed()) then we will end up with two `OCFileListFragment` showing since one is from reversal transaction and other one is created by function.

And this is what right now we're seeing is the persistent Preview Fragment (due to not calling onBackPressed) and the function creating a new one on top of it.

Linking the PR for this issue.
Thanks!

[...]

Issue #7197, opened 2020-11-01
Crash on Text file + details view


XueSheng-GIT:
### Steps to reproduce
1. Download textfile
2. Open text file (reading view)
3. Open details view (three dot menu)
4. Crash

### Expected behaviour
- Details view should be opened

### Actual behaviour
- Crash

### Logs

************ CAUSE OF ERROR ************

java.lang.ClassCastException: com.owncloud.android.ui.preview.PreviewTextFileFragment cannot be cast to com.owncloud.android.ui.fragment.OCFileListFragment
	at com.owncloud.android.ui.activity.FileDisplayActivity.getListOfFilesFragment(FileDisplayActivity.java:624)
	at com.owncloud.android.ui.activity.FileDisplayActivity.showDetails(FileDisplayActivity.java:1572)
	at com.owncloud.android.ui.activity.FileDisplayActivity.showDetails(FileDisplayActivity.java:1557)
	at com.owncloud.android.ui.preview.PreviewTextFileFragment.seeDetails(PreviewTextFileFragment.java:352)
	at com.owncloud.android.ui.preview.PreviewTextFileFragment.onOptionsItemSelected(PreviewTextFileFragment.java:322)
	at androidx.fragment.app.Fragment.performOptionsItemSelected(Fragment.java:2831)
	at androidx.fragment.app.FragmentManager.dispatchOptionsItemSelected(FragmentManager.java:2729)
	at androidx.fragment.app.FragmentController.dispatchOptionsItemSelected(FragmentController.java:412)
	at androidx.fragment.app.FragmentActivity.onMenuItemSelected(FragmentActivity.java:389)
	at androidx.appcompat.app.AppCompatActivity.onMenuItemSelected(AppCompatActivity.java:228)
	at androidx.appcompat.view.WindowCallbackWrapper.onMenuItemSelected(WindowCallbackWrapper.java:109)
	at androidx.appcompat.view.WindowCallbackWrapper.onMenuItemSelected(WindowCallbackWrapper.java:109)
	at androidx.appcompat.app.ToolbarActionBar$2.onMenuItemClick(ToolbarActionBar.java:65)
	at androidx.appcompat.widget.Toolbar$1.onMenuItemClick(Toolbar.java:207)
	at androidx.appcompat.widget.ActionMenuView$MenuBuilderCallback.onMenuItemSelected(ActionMenuView.java:779)
[...]

PR #11937, opened 2023-09-03
fix persistent text preview fragment in the background

parneet-guraya:
Solves: #11836 

https://github.com/nextcloud/android/assets/111801812/881516af-24dc-4c55-be45-559e53004b1e

https://github.com/nextcloud/android/assets/111801812/07b6eeaf-c522-49eb-b4b2-9c13cb34e490

### What this PR includes?
- Solve the issue by removing [this](https://github.com/nextcloud/android/blob/266e27f9ddd5e33fa59d6ee2ee345ccb5549c3ae/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java#L500) method which was causing the new fragment launch and end up having two same fragments.
- That method was originally added to solve this [issue](https://github.com/nextcloud/android/issues/7197#issue-734004562) and here's the [PR](https://github.com/nextcloud/android/pull/7229#issue-736096633), But now removing this call does not reproduce that same issue also. You can see below -->

https://github.com/nextcloud/android/assets/111801812/2fcbb85d-56e2-4dde-ae2f-92a0a48967b0

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -1096,8 +1096,6 @@ public class FileDisplayActivity extends FileActivity
                 resetTitleBarAndScrolling();
                 setDrawerAllFiles();
             }
-        } else if (leftFragment instanceof PreviewTextStringFragment) {
-            createMinFragments(null);
         } else if (leftFragment instanceof PreviewPdfFragment) {
             super.onBackPressed();
         } else {


parneet-guraya:
Hi, I solved this issue, Here's what was causing it -->
### Actual Behaviour ->
- `OCFileListFragment` which is the main view when launched, starts with fragment transaction (add) without call to addToBackStack(), It means no reverse transaction can happen thus pressing back exits the app (fragment goes in to onStop()).
- While launching `PreviewTextStringFragment` which opens when clicked upon `Readme.md` (preview at the top) adapter item , with transaction (replace) with addToBackStack(), so `replace` cause the `OCFileListFragment` to Destroy and `PreviewTextStringFragment` launches.
- This line [`else if (leftFragment instanceof PreviewTextStringFragment) {
            createMinFragments(null);
        }`](https://github.com/nextcloud/android/blob/266e27f9ddd5e33fa59d6ee2ee345ccb5549c3ae/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java#L1098) is the main issue first of all it's not calling super.onBackPressed(); which lead to non reversal of replace transaction, so neither the current `PreviewTextStringFragment` destroys nor `OCFileListFragment` gets created.

- What [createMinFragments(Bundle savedInstanceState)](https://github.com/nextcloud/android/blob/266e27f9ddd5e33fa59d6ee2ee345ccb5549c3ae/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java#L500) method doing is creating new `OCFileListFragment` with fragment transaction (add) without addToBackStack(), so no reversals.

- So, now when user click on Preview and it opens up and upon pressing back it does not destroy the Preview Fragment and also create new  `OCFileListFragment`, which shows us the Preview Fragment in the fragment and the new `OCFileListFragment` on top of it that's why we are seeing persistent Readme file in the background.

But, In fact we should be calling onBackPressed() as well as not calling the createMinFragments(null); since it's creating the new `OCFileListFragment` and  because if we call this one at this point (along with onBackPressed()) then we will end up with two `OCFileListFragment` showing since one is from reversal transaction and other one is created by function.

And this is what right now we're seeing is the persistent Preview Fragment (due to not calling onBackPressed) and the function creating a new one on top of it.

Thanks!

ZetaTom (approved):

parneet-guraya:
Closing as this solution is now part of this #11965 PR.

Issue #11963, opened 2023-09-12
Cross icon should not show when the search view field is empty.


parneet-guraya:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

- Open Text Preview.
- Click on search icon.
- Now see cross icon is showing even if there's no text.
- Clicking on it would make things out of sync and weird.

https://github.com/nextcloud/android/assets/111801812/b9af2a87-b245-4639-abe6-60c7aac3d4fd

### Expected behaviour

- Only show cross icon when text field has some text. 
- It should only show when there's some text because it is used for clearing the text

### Actual behaviour

- Showing cross icon even if field is empty.

### Android version

13

### Device brand and model

Oneplus 9RT

### Stock or custom OS?

Stock

### Nextcloud android app version

-

### Nextcloud server version

-

### Using a reverse proxy?

I don't know

Issue #11962, opened 2023-09-12
FAB (Floating Action Button) does not change back when coming back from rich preview.


parneet-guraya:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

- Open the rich preview of any directory
- FAB changes to edit icon.
- press back and see FAB does not change back to it's original icon ( which was plus(+) ) 

### Sample -->
https://github.com/nextcloud/android/assets/111801812/f0266f72-b476-4ccd-956d-6de18a06f71f

See how FAB does not change once rich preview is opened and pressed back.

### Expected behaviour

- FAB should change back to previous icon.

### Actual behaviour

- FAB does not change back if once changed to edit icon (upon going in rich preview)

### Android version

13

### Device brand and model

Oneplus 9RT

### Stock or custom OS?

Stock

### Nextcloud android app version

-

### Nextcloud server version

-

### Using a reverse proxy?

I don't know

Issue #11961, opened 2023-09-12
Unintended toolbar type when going to rich preview of root directory.


parneet-guraya:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

- Open the Nextcloud application.
- Then go to rich preview (that shows if a directory has a `Readme.md` file).
- You can see that the toolbar does not change, it is same as it was showing in the main view.

### Demo -->

https://github.com/nextcloud/android/assets/111801812/3ace4287-954a-4292-aa81-4180d5c7ee73

- Opening the rich preview from root does not change the toolbar
- After opening rich preview of photos directory does have the change.

### Expected behaviour

For the root directory the toolbar does not change as it does when going to other directories or previews.

### Actual behaviour

The toolbar remains as same as it was in the main view when opening rich preview from the root directory.

### Android version

13

### Device brand and model

Oneplus 9RT

### Stock or custom OS?

Stock

### Nextcloud android app version

-

### Nextcloud server version

-

### Using a reverse proxy?

I don't know

PR #11965, opened 2023-09-12
Unintended UI elements showing

### LINKED ISSUE #11700: PreviewText: extraneous ui elements shown
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. set up nextcloud app with account
2. create empty `file.txt` in root directory
3. download file via three dots menu
4. open file (local editor)

### Expected behaviour

Text editor ui replaces previous ui.

### Actual behaviour

Some ui elements from previous screen remain visible, such as the display (List/Card view) and sorting options (*A - Z*). These elements do not perform any visible action.

Additionally, the *Search* icon, visible on the top right-hand corner remains on screen. Inserting any text into the provided search and returning using the back button will result in graphical glitches and eventually a list of search results being displayed. Now, one is unable to return to the main, unfiltered, view, as navigating back leads to the same screen. Further attempts to navigate anywhere result in a broken interface.

### Android version

13

### Device brand and model

OnePlus 6T

### Stock or custom OS?

Custom (explain in "additional information")

849799c

### Nextcloud server version

26.0.1

### Using a reverse proxy?

No

### Android logs

_No response_

### Server error logs

_No response_

### Additional information

Custom OS: LineageOS 20.0

parneet-guraya:
Resolves #11700 

- Fixes issues mentions in the issue.
- `sortListGroup` visibility handled by Stack. So, it does not go out of sync. Here's how it's handling -->
    - If going into a View which needs to have different visibility (either visible or not).
    - Then, we save the current visibility first and then change the visibility according to upcoming view
    - So, when going back we can pop out the visibility and we will have the correct visibility.
    - This way view visibility will never go out of sync (because of some unhandled case) and we will always have the correct visibility applied since we're saving it in a stack and popping out when going back.
    - It is somewhat like how Fragments are in backstack and going back they popes out.
    - Same way we're stacking up (when going in but only when we need to change the visibility) and popping out (handled to pop only for the cases where it needs to popped).

### Before

https://github.com/nextcloud/android/assets/111801812/b1d7b12d-f8b2-4f9a-97fc-34669faaff53

### After 

https://github.com/nextcloud/android/assets/111801812/c5e4caf0-91f0-4ea3-9fdc-ee61ca6560dd

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -144,6 +144,7 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Stack;
 
 import javax.inject.Inject;
 
@@ -233,7 +234,7 @@ public class FileDisplayActivity extends FileActivity
     private PlayerServiceConnection mPlayerConnection;
     private Optional<User> lastDisplayedUser = Optional.empty();
     private int menuItemId = -1;
-
+    private Stack<Boolean> previousSortGroupState;
     @Inject
     AppPreferences preferences;
 
@@ -622,7 +623,7 @@ public class FileDisplayActivity extends FileActivity
 
         //clear the subtitle while navigating to any other screen from Media screen
         clearToolbarSubtitle();
-        showSortListGroup(showSortListGroup);
+        setSortListGroup(sortListGroupVisibility(), showSortListGroup);
         FragmentTransaction transaction = getSupportFragmentManager().beginTransaction();
         transaction.addToBackStack(null);
         transaction.replace(R.id.left_fragment_container, fragment, TAG_LIST_OF_FILES);
@@ -1027,9 +1028,9 @@ public class FileDisplayActivity extends FileActivity
     }
 
     private void popBack() {
-        // pop back
+        // pop back fragment
         resetScrolling(true);
-        showSortListGroup(true);
+        popSortListGroupVisibility();
         super.onBackPressed();
     }
 
@@ -1038,7 +1039,6 @@ public class FileDisplayActivity extends FileActivity
         setFile(listOfFiles.getCurrentFile());
         listOfFiles.setFabVisible(true);
         listOfFiles.registerFabListener();
-        showSortListGroup(true);
         resetTitleBarAndScrolling();
         setDrawerAllFiles();
     }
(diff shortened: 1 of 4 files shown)


parneet-guraya:
Also, here are the issues that I found while I was working on this PR (the issues mentioned are not introduced because of this PR). Issues --> 
- #11961 
- #11962 
- #11963

I will be working on these issues. Other than that have a look on this PR and let me know if any changes required or any suggestions.
Thanks!

parneet-guraya:
Hello team 👋, I'd appreciate it if someone could review my PR soon. Let me know if you need any additional information.

 Thanks! 😄

parneet-guraya:
> The changes made to other files, including refactoring parts of `FileDisplayActivity`, are valid and a necessary effort. However, as we try to keep pull requests as modular as possible, I would suggest moving this to a separate branch if it's not too much trouble.

Hi 👋 @ZetaTom , If I understood correctly you mean to have this commit (where I have did some refactoring) https://github.com/nextcloud/android/pull/11965/commits/5260ec7eb3ccdddcca049c4cb774f6f5ebe55c8c to be in a separate branch? I may be wrong, if that's the could you explain it what changes should be pushed in a separate branch?

ZetaTom:
> Hi 👋 @ZetaTom , If I understood correctly you mean to have this commit (where I have did some refactoring) [5260ec7](https://github.com/nextcloud/android/commit/5260ec7eb3ccdddcca049c4cb774f6f5ebe55c8c) to be in a separate branch? I may be wrong, if that's the case could you explain it what changes should be pushed in a separate branch?

Yes, essentially extract those changes, and any changes not directly related to the issue at hand, into a separate pull request. This will make it easier to review and maintain in the future.

parneet-guraya:
I got rid of the refactoring part. Check it out.

parneet-guraya:
Hi, just following up on the changes I made for this PR. Could you please take a look when you get a chance?

Thanks!

ZetaTom (approved):

alperozturk96 (approved):

parneet-guraya:
Hi @AndyScherzinger 👋, I was wondering is there anything left to do before merging this? Already approved.

AndyScherzinger:
@parneet-guraya all good I would say 👍 Let's merge when the checks have finished 👍

PR #11987, opened 2023-09-24
Fixed the appearence of keyboard while renaming

Miniman45:
**Fixed the appearence of keyboard while renaming**

Here's a description of the problem and how it was resolved

The issue was related to the automatic opening of the soft keyboard when displaying a MaterialAlertDialogBuilder dialog. Despite calling showSoftInput() within the onShow() method of the dialog, the keyboard did not open as expected. Instead, the log displayed the message "Ignoring showSoftInput() as view is not served."

**Root Cause Analysis:**

The root cause of the issue appeared to be a race condition between requesting focus for the EditText view and displaying the soft keyboard. The Material Design components used in the layout might have had specific behaviors that affected the timing of these actions.
Resolution:
To resolve this issue, the following steps were taken:
* 		A Handler was introduced to introduce a slight delay between requesting focus and showing the soft keyboard. This delay allowed for better synchronization and prevented the race condition.
* 		The code was modified to use new Handler().postDelayed() to delay the showSoftInput() call by 100 milliseconds after requesting focus on the EditText.

**Result:**

After implementing these changes, the soft keyboard opened automatically when the dialog was shown, as intended. The issue of the keyboard not opening was successfully resolved.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/dialog/RenameFileDialogFragment.java
+++ b/app/src/main/java/com/owncloud/android/ui/dialog/RenameFileDialogFragment.java
@@ -201,7 +201,7 @@ public class RenameFileDialogFragment
                         InputMethodManager imm = (InputMethodManager) requireActivity().getSystemService(Context.INPUT_METHOD_SERVICE);
                         imm.showSoftInput(binding.userInput, InputMethodManager.SHOW_IMPLICIT);
                     }
-                }, 100); // Delay for 200 milliseconds
+                }, 100); // Delay for 100 milliseconds
             }
         });
         return dialog;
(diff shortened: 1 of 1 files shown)


Miniman45:
@AndyScherzinger @tobiasKaminsky could you please review this PR

ZetaTom (commented):
Firstly, thank you @Miniman45 for fixing this issue and bringing it to my attention once again. Apologies that this took a bit longer than anticipated.
I've spent some time on this issue and found that your solution works as expected. However, it seems that we are already using a similar function to delay the keyboard call in `KeyboardUtils.kt`. As you mentioned, this doesn't work properly. The currently [recommended method](https://developer.android.com/develop/ui/views/touch-and-input/keyboard-input/visibility#ShowReliably) uses `WindowInsetsControllerCompat` to reliably show the keyboard.
If you would like, I could assist in making this change.

Miniman45:
> WindowInsetsControllerCompat

Yeah sure

PR #12003, opened 2023-10-04
Bugfix/update folder icons

alperozturk96:
Instead of showing the icons of the file directly, the needed material icons were overlayed above the folder icon. Because we don't have single folder icon for each state of file. Web application also using overlaying mechanism plus we have more flexibility to apply any icon to folder icon.

- [Done] Tests written


tobiasKaminsky:
@alperozturk96 this looks off on our UI screenshot tests:
https://www.kaminsky.me/nc-dev/android-integrationTests/12003-Screenshot-blue-Light-11-02/

you can create the same emulator via:
avdmanager create avd -n uiComparison -c 100M -k "system-images;android-27;google_apis;x86" --abi "google_apis/x86"

AndyScherzinger (commented):
overlays will need to work in dark/light

alperozturk96:
> overlays will need to work in dark/light

If dark mode activated overlay must be dark or vice versa its need to be white. I just noticed that from web app.

tobiasKaminsky:
- Check spotlessKotlinCheck fails, but you can run:
./gradlew :app:spotlessApply
to have this automatically fixed.
- Codacy fails too, which you can see when you click on details.
- DCO is required, so please sign all your git commits:
In your local branch, run: git rebase HEAD~52 --signoff
Force push your changes to overwrite the branch: git push --force-with-lease origin bugfix/Update-folder-icons

Why analysis fails, I will check.

app/src/androidTest/java/com/nextcloud/client/FileDisplayActivityScreenshotIT.kt
```diff
@@ -49,6 +49,7 @@ class FileDisplayActivityScreenshotIT : AbstractIT() {
         Manifest.permission.WRITE_EXTERNAL_STORAGE
     )
 
+    // FIXME test fails
```
tobiasKaminsky, line 52: Are those really failing?
In latest run I only saw location and activity error.
If they are not failing, please remove those comments.

alperozturk96: I ran those marked test still fails in my local machine.

alperozturk96: The reason why the tests fails is that CI/CD runs on Android 8. My local device is Android 14 so the tests are failing. For example the expected file path is different on newer Android versions.

alperozturk96:
> Given the current commit ([1e4e417](https://github.com/nextcloud/android/commit/1e4e41707749d6de7da3c2318b05b1c973536460)), the app unfortunately crashes when attempting to open the overflow menu of any folder. However, it works fine for regular files. I'm running Android 11.0 (API 30) on a Pixel 2 emulator.
> Stacktrace

Could you check with latest commit?

PR #12022, opened 2023-10-12
test only LoginIT, to have better server output

tobiasKaminsky:


tobiasKaminsky:
Verified: $EVAL is set, allowing eval script in ContentSecurityPolicy.php

tobiasKaminsky:
Turns out to be a problem with API level.

PR #12026, opened 2023-10-12
Feature/combine move and copy in menu

### LINKED ISSUE #2212: Combine Move and Copy in menu
Just like in the server, we should have a single entry "Move or copy".

For that the current move/copy view just needs to be minimally adjusted:

- In the root view, header should have an "x" in the top left to cancel operation
- Header text in root view should be "Move <filename>" (not enough space to include "copy" here)
- Primary blue button on right should say "Move here", secondary grey left button " Copy".

cc @nextcloud/android

alperozturk96:
Resolves #2212 

Move and copy file actions combined into one.


AndyScherzinger:
Looping in @jancborchardt for the designer feedback since you raised it, see https://github.com/nextcloud/android/pull/12026/files#diff-2e25592cb526aee4ed8eb25760a886ca36972b6111f2bfa2ab2595cd0d388d63

joshtrichards:
Associated with (and would close) #2212, no?

jancborchardt (changes requested):
Lools very nice! Only one detail: "Move" should be the primary styled button instead of "Copy". :) Placement is good like this though.

alperozturk96:
@jancborchardt @AndyScherzinger In our web application button colors exactly same. Copy [filled](https://m3.material.io/components/buttons/overview) and Move [tonal](https://m3.material.io/components/buttons/overview).

## Web App
Screenshot 2023-10-20 at 12 01 42

I suggest we must sync on both side so that we can preserve our design language across platforms.

AndyScherzinger (approved):
LGTM

PR #12028, opened 2023-10-12
Fix wrong icon overlay colors on OS theme

alperozturk96:


AndyScherzinger (approved):
Works like a charm @alperozturk96 👍

PR #12032, opened 2023-10-13
Feature/use-material-design-3-for-rename-dialog

alperozturk96:
Before:
rename_dialog_before

After:
rename_dialog_after


AndyScherzinger (approved):
LGTM, works fine

PR #12049, opened 2023-10-16
Use M3 for SyncedFolderPreferencesDialogFragment

alperozturk96:
Before:
before_sync

After:
after


tobiasKaminsky (approved):

tobiasKaminsky:
On first screenshot run size of dialog changed.
Let's check if on second try it is the same.
If so, then we need to update screenshot.

PR #12052, opened 2023-10-17
Use Material Design 3 File Downloader Notification

alperozturk96:
Guideline:
Guideline

We don't need to add buttons for downloading because it's not cancelable operation.

Before:
Screenshot 2023-10-17 at 12 00 37

before

After:

Filename too long to show. We can take maybe first 10 character? @AndyScherzinger @tobiasKaminsky 


alperozturk96:
@nextcloud/designers-community Some file names are too long and hard to show in notifications. What if we do something like this?

e.g file name: image-11.11.2011-tes-scrolls.jpeg

1) image-11.11.20...jpeg download
2) Downloading image-11.11.20...jpeg

tobiasKaminsky:
As we wait for answer by Designers, I set "2. developing".

nimishavijay:
@alperozturk96 nice! We could go with option 2, since file names are (most of the time) not easily understandable anyway :)

alperozturk96:
@tobiasKaminsky Failed due to 

FileMenuFilterIT - filter_unset_encryption(), filter_select_all(), filter_noLockingCapability_lockItemsInvisible()

these test, on my local machine all tests are passing, fyi.

PR #12058, opened 2023-10-18
Use Material Desing 3 For File Actions Bottom Sheet

alperozturk96:
Material Design 3 Bottom Sheet Example:
bottomsheet_hero

Before:
before

After:
Screenshot 2023-10-18 at 14 41 05

STATE_EXPANDED bottom sheet behaviour can't have rounded corners. Official [Source](https://github.com/material-components/material-components-android/pull/437)


app/src/main/java/com/nextcloud/ui/fileactions/FileActionsBottomSheet.kt
```diff
@@ -109,9 +111,24 @@ class FileActionsBottomSheet : BottomSheetDialogFragment(), Injectable {
         bottomSheetDialog.behavior.state = BottomSheetBehavior.STATE_EXPANDED
         bottomSheetDialog.behavior.skipCollapsed = true
 
+        applyTintedRoundedBackground()
+
         return binding.root
     }
 
+    private fun applyTintedRoundedBackground() {
```
AndyScherzinger, line 119: fine for now, mid-term we should move this to the common lib, i.e. https://github.com/nextcloud/android-common/blob/c9df82f2aec4dbd79d0ad2196dc8acdbf3496fc4/ui/src/main/java/com/nextcloud/android/common/ui/theme/utils/AndroidViewThemeUtils.kt

app/src/main/res/layout/file_actions_bottom_sheet.xml
```diff
@@ -87,5 +79,7 @@
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:orientation="vertical" />
+
     </androidx.core.widget.NestedScrollView>
-</LinearLayout>
+
+</LinearLayout>
```
AndyScherzinger, line 85: lacks an empty line at the end

app/src/main/java/com/nextcloud/ui/fileactions/FileActionsBottomSheet.kt
```diff
@@ -109,6 +109,8 @@ class FileActionsBottomSheet : BottomSheetDialogFragment(), Injectable {
         bottomSheetDialog.behavior.state = BottomSheetBehavior.STATE_EXPANDED
         bottomSheetDialog.behavior.skipCollapsed = true
 
+        viewThemeUtils.platform.colorViewBackground(binding.bottomSheet, ColorRole.SURFACE_VARIANT)
```
AndyScherzinger, line 112: surface variant seems to be very high contrast, maybe use the same we use for dialog background?

alperozturk96: Indeed, it appears to be slightly darker. Regrettably, we are not providing a ColorRole for the dialog; instead, we are directly passing the dialog as an argument, and the common library applies the scheme internally.

I believe we may need to introduce an additional ColorRole or function for the bottomSheet within the viewThemeUtils.

AndyScherzinger: Yeah, the color roles are from the library implementation build by Google and they don't ship the color role for bottom sheets yet. But when I checked Google Drive bottom sheets and dialog where using the exact same color, so I expect it to be 2 different design tokens but ultimately the same color.

The reason for passing the dialog directly is that we wanted to have or move to mostly higher-level functions so devs don't need to know and choose color roles but have the common lib always choosing the appropriate ones.

alperozturk96:
@AndyScherzinger @tobiasKaminsky In pipiline these tests are failed but in my local machine all of it passed.

pass

PR #12068, opened 2023-10-19
Bugfix Disable focus change via tap

alperozturk96:
Users are capable of inadvertently changing focus through tapping, which can result in accidental input and lead to confusion for App passcode enter.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/PassCodeActivity.java
@@ -42,6 +42,7 @@ import com.owncloud.android.R;
 import com.owncloud.android.authentication.PassCodeManager;
 import com.owncloud.android.databinding.PasscodelockBinding;
 import com.owncloud.android.lib.common.utils.Log_OC;
+import com.owncloud.android.ui.components.PassCodeEditText;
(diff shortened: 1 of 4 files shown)


AndyScherzinger (approved):

tobiasKaminsky:
Great so far.
One small thing: When I enter last digit, higlighted bar is then again at first one.
After a very short timespan then the app is unlocked.
I think it should remain on 4th position.

tobiasKaminsky (changes requested):
Do not jump to 1st position after entering 4th digit

tobiasKaminsky:
Hm. I just tested it with master and this is the already existing behaviour.
Can you nevertheless check if this is quickly possible?
If not, we can keep it this way and directly merge your PR.

alperozturk96:
@tobiasKaminsky Done

ZetaTom (commented):
In my tests, actual input worked well with the virtual keypad and an external keyboard. However, the backspace key on the virtual keypad didn't work. Pressing the backspace key on the external keyboard resulted in one, or sometimes several, inputs being deleted.

tobiasKaminsky:
I think external keyboard cannot be used for testing with an emulator.

tobiasKaminsky (approved):

PR #12085, opened 2023-10-24
Use Material Design 3 For File Details Sharing Process Fragment

alperozturk96:
Before:
Screenshot 2023-10-24 at 10 25 35

After:
after


app/src/main/java/com/nextcloud/utils/extensions/BundleExtensions.kt
```diff
@@ -0,0 +1,44 @@
...
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.nextcloud.utils.extensions
+
+import android.os.Build
+import android.os.Bundle
+import android.os.Parcelable
+import java.io.Serializable
+
```
alperozturk96, line 28: These function needed for getting argument and part of this PR also will be used for future for other fragments.

AndyScherzinger (approved):

PR #12093, opened 2023-10-25
Bugfix Check Network Connection When File Item Clicked

alperozturk96:


ZetaTom (commented):
Refactoring is a necessary effort, but it is generally considered good practise to keep pull requests as modular as possible. I would suggest moving all changes, not strictly needed to fix the issue at hand, into a separate pull request.

alperozturk96:
@ZetaTom I agree. Single responsibility in PR is important. I will revert those changes (enhanced switch case, requireActivity, requireContext usages)

app/src/main/java/com/nextcloud/client/network/ConnectivityObserver.kt
```diff
@@ -0,0 +1,44 @@
...
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package com.nextcloud.client.network
+
+import android.content.Context
+import android.net.ConnectivityManager
+import android.net.NetworkCapabilities
+
+object ConnectivityObserver {
```
tobiasKaminsky, line 28: We already have ConnectivityService for this.
Please use/combine those.

app/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
```diff
@@ -36,7 +37,7 @@
 import androidx.core.net.ConnectivityManagerCompat;
 import kotlin.jvm.functions.Function1;
 
-class ConnectivityServiceImpl implements ConnectivityService {
+public class ConnectivityServiceImpl implements ConnectivityService {
```
tobiasKaminsky, line 40: This is implemention of an interface.
Please add the new function to its interface.

app/src/main/java/com/owncloud/android/ui/activity/FileActivity.java
```diff
@@ -240,6 +241,12 @@ protected void onCreate(Bundle savedInstanceState) {
         }
     }
 
+    public void checkInternetConnection() {
+        if (ConnectivityServiceImpl.isConnected(this)) {
```
tobiasKaminsky, line 245: When it is via interface you can use "connectivityService", which is injected.

tobiasKaminsky (changes requested):
Use interface

alperozturk96:
@tobiasKaminsky Interface used

ZetaTom (commented):
During my testing, I've found that there are certain cases where the message disappears even though the server is unavailable. For example, if you open a media preview while the server is stopped.

alperozturk96:
@ZetaTom When user don't have internet I am always seeing Server not available message. Could you share video?

ZetaTom:
@alperozturk96, the message disappears if the device *is* online but the ***server** isn't*.

[server_availability.webm](https://github.com/nextcloud/android/assets/70907959/a892ae7d-057a-4e12-9a11-e2caf1d87a54)

alperozturk96:
@ZetaTom I assume you expect to still see the message if our backend not available if yes then my question will be our backend providing heartbeat API? So I can check backend health?

JonasMayerDev (approved):

PR #12108, opened 2023-10-30
Check WebView Version For Login Functionality

alperozturk96:
**Problem:**
Login feature relies on WebView. If Android System WebView version is outdated login functionality is not working. Therefore we need to inform the user to update Android System WebView.

**If web view version outdated:**

https://github.com/nextcloud/android/assets/67455295/ddf8fe8b-9bfe-49f5-bf4a-4ec8e72ff470

@nextcloud/designers Is it suitable in terms of UI/UX?


app/src/main/java/com/owncloud/android/utils/WebViewUtil.kt
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Ideally we should fetch from database, reading actual value
+     * from PlayStore not feasible due to frequently api changes made by
+     * Google
+     *
+     */
+    private fun getMinimumSupportedMajorWebViewVersion(): String {
+        return "118"
```
alperozturk96, line 108: @AndyScherzinger @tobiasKaminsky Can we get that value from backend?

AndyScherzinger: Server doesn't have/know this. So we could for the moment hard-code this on client-side and bump it per major release of the server.

AndyScherzinger (approved):

szaimen (approved):
Nice idea! The design LGTM :)

app/src/main/java/com/owncloud/android/utils/WebViewUtil.kt
```diff
@@ -0,0 +1,110 @@
...
+                context.getString(R.string.webview_version_check_alert_dialog_positive_button_title)
+            ) { _, _ ->
+                redirectToAndroidSystemWebViewStorePage()
+            }
+
+        val dialog = builder.create()
+        dialog.show()
+    }
+
+    private fun redirectToAndroidSystemWebViewStorePage() {
```
ZetaTom, line 62: A very similar function already exists in `DrawerActivity.java`, called `openAppStore()`. Currently its visibility is set to private. Would it be possible to consolidate this functionality?

app/src/main/java/com/owncloud/android/utils/WebViewUtil.kt
```diff
@@ -0,0 +1,110 @@
...
+        val playStoreWebUri = Uri.parse("https://play.google.com/store/apps/details?id=$packageName")
+        val webIntent = Intent(Intent.ACTION_VIEW, playStoreWebUri)
+        context.startActivity(webIntent)
+    }
+
+    private fun getCurrentWebViewMajorVersion(): String? {
+        val pm: PackageManager = context.packageManager
+
+        return try {
+            val pi = pm.getPackageInfo("com.google.android.webview", 0)
```
ZetaTom, line 84: During my tests, this line consistently throws `PackageManager.NameNotFoundException`. This doesn't work as expected on devices running AOSP. ~~Maybe the package name differs on AOSP images?~~ It's called `com.android.webview` on my phone running LineageOS.

I've been able to reproduce this using the following configurations:

- Pixel 2 (Android Studio Emulator) x86_64 API 26 AOSP
- Pixel 2 (Android Studio Emulator) x86_64 API 24 AOSP
- Pixel 2 (Android Studio Emulator) x86_64 API 30 AOSP
- OnePlus 6T API 33 LineageOS 20

In that case, one would also have to consider third party *SystemWebView* implementations, such as [Bromite](https://www.bromite.org/system_web_view), for example.

However, because the calling functions returns `true`, if this function returns `null`, this could possibly be ignored?

Issue #11969, opened 2023-09-15
[Android App crash] ParcelFileDescriptor Transfer Thread Java error


sunnyd24:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Go to sleep. 
2. Wake up to see crash report generated. 

### Expected behaviour

No android app crash. 

### Actual behaviour

1. On starting the phone for the first time in morning, I notice that there is no internet available on WiFi (Connected without internet), and has switched to mobile data.
2. See crash report for nextcloud. 

### Android version

Android 13

### Device brand and model

Samsung S21 Ultra 5G

### Stock or custom OS?

Stock

### Nextcloud android app version

30250090

### Nextcloud server version

27

### Using a reverse proxy?

No

### Cause of error
```java
Exception in thread "ParcelFileDescriptor Transfer Thread" java.lang.NullPointerException
    at java.io.SequenceInputStream.nextStream(SequenceInputStream.java:112)
    at java.io.SequenceInputStream.read(SequenceInputStream.java:211)
    at java.io.InputStream.read(InputStream.java:205)
    at com.nextcloud.android.sso.aidl.ParcelFileDescriptorUtil$TransferThread.run(ParcelFileDescriptorUtil.java:76)

```

### App information
* ID: `com.nextcloud.client`
* Version: `30250090`
* Build flavor: `gplay`

### Device information
* Brand: `samsung`
* Device: `p3s`
* Model: `SM-G998B`
* Id: `TP1A.220624.014`
* Product: `p3sxeea`

### Firmware
* SDK: `33`
* Release: `13`
* Incremental: `G998BXXS9EWH1`

### Additional information

Not sure what Nextcloud was trying to do in the background.

PhieF:
same thing is happening when uploading a file using the Android-SingleSignOn library (on the same object)

PR #12109, opened 2023-10-30
Bugfix rash parcel file descriptor util run

alperozturk96:


app/src/main/java/com/nextcloud/android/sso/aidl/ParcelFileDescriptorUtil.kt
```diff
@@ -0,0 +1,97 @@
...
+        override fun run() {
+            try {
+                val buf = ByteArray(1024)
+                var len: Int
+
+                while (inputStream.read(buf).also { len = it } > 0) {
+                    outputStream.write(buf, 0, len)
+                }
+                outputStream.flush()
+            } catch (t: Throwable) {
```
alperozturk96, line 72: Whole file converted to the kotlin file and catch all throwable rather than IOException. In [bug](https://github.com/nextcloud/android/issues/11969) report app is crashed due to NullPointerException however when I look at the "write, close, read" functions all of it throwing IOException and we are catching in code. Somehow user able to came across with this problem. I try to reproduce same crash but no luck.

PR #12112, opened 2023-10-31
Fix warnings for OCFileListFragment

alperozturk96:


app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1786,38 +1734,8 @@ private void encryptFolder(OCFile folder,
         }
     }
 
-    @Subscribe(threadMode = ThreadMode.BACKGROUND)
-    public void onMessageEvent(FileLockEvent event) {
```
alperozturk96, line 1790: Unused function

AndyScherzinger: Like mentioned in the other PR, I expect the event bus to make use of them, so this should actually be a false-positive, best to check with @tobiasKaminsky

tobiasKaminsky: Yeah, this is still used.
You can check where FileLockEvent is being used:
Within FileOperationsHelper there is: EventBus.getDefault().post(new FileLockEvent(file.getRemotePath(), shouldBeLocked));

This is to send the event, and the other function receives it.

In long term I would like to get rid of it…

tobiasKaminsky (changes requested):
Re-add event bus function.

app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -421,8 +430,7 @@ public void onActivityCreated(Bundle savedInstanceState) {
         setTitle();
 
         FragmentActivity fragmentActivity;
-        if ((fragmentActivity = getActivity()) != null && fragmentActivity instanceof FileDisplayActivity) {
-            FileDisplayActivity fileDisplayActivity = (FileDisplayActivity) fragmentActivity;
+        if ((fragmentActivity = getActivity()) != null && fragmentActivity instanceof FileDisplayActivity fileDisplayActivity) {
```
ZetaTom, line 433:
```suggestion
        if (getActivity() instanceof FileDisplayActivity fileDisplayActivity) {
```

app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1786,6 +1731,7 @@ private void encryptFolder(OCFile folder,
         }
     }
 
+    // Do not delete this function
```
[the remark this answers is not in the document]
alperozturk96: IDE will always show as unused function better to keep that comment

alperozturk96:
Out of date due to conflicts

PR #12115, opened 2023-10-31
fix: typo error in SETUP.md

vivekBoii:
configurtation -> configuration


AndyScherzinger:
Thanks for the fix @vivekBoii

PR #12111, opened 2023-10-31
Build Fix

alperozturk96:
- [x] Tests written, or not not needed

PR #12117, opened 2023-11-01
Fix missing title in ConfirmationDialogFragment

ZetaTom:
This change fixes a crash that would occur when attempting to delete any file. (`Resources.NotFoundException`)

#### Analysis
The `RemoveFilesDialogFragment` extends the `ConfirmationDialogFragment` and thus inherits all methods. Upon instantiation, a an argument bundle is created and set appropriately. This bundle does, however, not include any title for the aforementioned dialogue.

This needs to be considered in the super class, when setting up the ui. A recent change unfortunately removed the [necessary check](https://github.com/nextcloud/android/pull/12111/files#diff-fcd4084665d5f956334103105bf953b83aea08f7e09c2cc3426b8cebf04f7ee9L85).

---


AndyScherzinger (approved):

alperozturk96:
@ZetaTom One line solution, fyi

`builder.setTitle(titleId.takeIf { it != 0 && it != -1 } ?: R.string.dialog_alert_title)`

alperozturk96 (approved):

PR #12142, opened 2023-11-07
Open app internal preview for files in upload section

JonasMayerDev:
In upload tab uploading / failed/ waiting items can be viewed locally with android image view by clicking on file icon/ thumbnail.
Files that are finished with uploading lead you when clicked to the cloud file view and opens the file (as if it was opened via the all files tab) to e.g. share it.

- Is it a problem that files that are uploaded and, via the upload option, are kept at the original location of the client device are downloaded again into the Nextcloud save folder in the device to preview? 
- Do I need tests here? 

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -2557,6 +2557,7 @@ public class FileDisplayActivity extends FileActivity
     private Unit onFileRequestError(Throwable throwable) {
         dismissLoadingDialog();
         DisplayUtils.showSnackMessage(this, getString(R.string.error_retrieving_file));
+        Log_OC.e(TAG, "Requesting file from remote failed!", throwable);
         return null;
     }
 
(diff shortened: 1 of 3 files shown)


alperozturk96 (approved):

ZetaTom (commented):
Works well in my tests. In general, the UI behaves as expected but there are some edge cases, which I've found.

When opening a successfully uploaded image and returning to the main file list view, the actionbar shows the filename and not the default searchbar. This is most likely not related to this pull request and might be fixed by #11965.

Another possibly unrelated issue is that once a file has been uploaded, and subsequently deleted, it is still possible to open the preview from the *Uploads* tab. Doing so will result in a *Retrieving file...* dialogue being displayed. As this is a blocking dialogue, the user is unable to interact with the app and will have to restart it manually.

JonasMayerDev:
I have now added an image viewer that opens in the uploads tab, so one does not have to navigate back to uploads to look at other uploaded images (should now behave like in the media tab).

I also combined the whole logic for showing the file in the UploadListAdapter to have it in one place.

tobiasKaminsky (approved):

Issue #11072, opened 2022-11-23
Centralize navigation


AlvaroBrey:
We're always running into problems related to navigation (like incorrect titles, scroll, sort-list, drawer menu etc) on edge cases.

This stems from the fact that our navigation logic is very split throughout different parts of the app.

To avoid this we should have a central navigation system, whether it is the Navigation Component or something custom, that is used to navigate and can also be listened to, for parts of the app to update their UI depending on current route.

PR #12146, opened 2023-11-09
Bugfix Refresh Gallery After Edit, Delete Image

alperozturk96:

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -479,6 +479,12 @@ public abstract class DrawerActivity extends ToolbarActivity
         unsetAllDrawerMenuItems();
     }
 
+    public void openMediaList() {
+        SearchEvent searchEvent = new SearchEvent("image/%", SearchRemoteOperation.SearchType.PHOTO_SEARCH);
+        MainApp.showOnlyFilesOnDevice(false);
+
+        launchActivityForSearch(searchEvent, R.id.nav_gallery);
+    }
 
     private void onNavigationItemClicked(final MenuItem menuItem) {
         setDrawerMenuItemChecked(menuItem.getItemId());
--- a/app/src/main/java/com/nextcloud/client/editimage/EditImageActivity.kt
+++ b/app/src/main/java/com/nextcloud/client/editimage/EditImageActivity.kt
@@ -32,6 +32,7 @@ import androidx.core.content.ContextCompat
 import androidx.core.graphics.drawable.DrawableCompat
 import com.canhub.cropper.CropImageView
 import com.nextcloud.client.di.Injectable
+import com.nextcloud.utils.extensions.getParcelableArgument
 import com.owncloud.android.R
 import com.owncloud.android.databinding.ActivityEditImageBinding
 import com.owncloud.android.datamodel.OCFile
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -479,6 +479,12 @@ public void onMessageEvent(DummyDrawerEvent event) {
         unsetAllDrawerMenuItems();
     }
 
+    public void openMediaList() {
```
tobiasKaminsky, line 482: This is the same as "startPhotoSearch"?
If so, you can combine them

tobiasKaminsky:
Can you check also spotbugs?

alperozturk96:
@tobiasKaminsky  These are the differences and irrelevant with this PR. I rebased, lets wait and see the new results.

 MF: In class com.owncloud.android.ui.trashbin.TrashbinActivity
In class com.owncloud.android.ui.trashbin.TrashbinActivity
Masking field com.owncloud.android.ui.trashbin.TrashbinActivity.viewThemeUtils
Masked field com.owncloud.android.ui.activity.ToolbarActivity.viewThemeUtils
In TrashbinActivity.java
Field TrashbinActivity.viewThemeUtils masks field in superclass com.owncloud.android.ui.activity.ToolbarActivity


NP: In class com.owncloud.android.ui.trashbin.TrashbinActivity
In class com.owncloud.android.ui.trashbin.TrashbinActivity
In method com.owncloud.android.ui.trashbin.TrashbinActivity.onSortingOrderChosen(FileSortOrder)
Parameter sortOrder
At TrashbinActivity.java:[lines 257-260]
sortOrder must be non-null but is marked as nullable


  UrF: In class com.owncloud.android.datamodel.MediaFolder
In class com.owncloud.android.datamodel.MediaFolder
Field com.owncloud.android.datamodel.MediaFolder.numberOfFiles
At MediaProvider.java:[line 155]
Unread public/protected field: com.owncloud.android.datamodel.MediaFolder.numberOfFiles

   UrF: In class com.owncloud.android.datamodel.MediaFolder
In class com.owncloud.android.datamodel.MediaFolder
Field com.owncloud.android.datamodel.MediaFolder.type
At MediaProvider.java:[line 111]
Unread public/protected field: com.owncloud.android.datamodel.MediaFolder.type


../../src/main/java/com/owncloud/android/ui/adapter/ShareeListAdapter.java:195: It will always be more efficient to use more specific change events if you can. Rely on notifyDataSetChanged as a last resort.

../../src/main/java/com/owncloud/android/ui/adapter/ShareeListAdapter.java:172: It will always be more efficient to use more specific change events if you can. Rely on notifyDataSetChanged as a last resort.

JonasMayerDev (approved):
works for me, fixes the issue, doesn't add new issues in my testing

ZetaTom:
I've noticed that the preview behaves slightly differently compared to the current version on master. Some may display a white border, which they lacked before. Is this intentional @alperozturk96?

`master` | this PR
--- | ---
 | 

Notice the toucan image specifically.

alperozturk96:
@ZetaTom Nothing changed regarding border and frame calculation. GalleryRowHolder.computeShrinkRatio() does the job.

JonasMayerDev:
For me, it does not update when going back via details. So create a new edited image, 3 Dots, Details and then the back arrow still leads to the old behavior. But going back via details is in general really buggy, as I documented here:  [issue](https://github.com/nextcloud/android/issues/12163)

ZetaTom (commented):
In terms of the problem mentioned, this fix works well. The media view is updated accordingly each time the user returns to it.

However, with this change, the media view behaves a bit wired. Consider the following example: a user has scrolled down and opened a picture taken a long time ago. Upon returning to the media view, the list refreshes and the user ends up back at the top. As an aside, on my machine the UI flashes every time the list is refreshed and therefore every time the user returns from the preview.

alperozturk96:
@ZetaTom For getting actual data we need to trigger search event again therefore all list gets refreshed and that flashing, scrolling top of view etc issues appears. 

Root of those problem is navigation. Switching activies and using multiple activities for one simple straightforward nav stack is wrong. I suggest we should fix that kind of issues after this [enhancement](https://github.com/nextcloud/android/issues/11072) otherwise we will double the job. @tobiasKaminsky

PR #12161, opened 2023-11-13
Improve media playback

ZetaTom:
This change improves the user experience when previewing media. Video files and images are now displayed in a combined viewpager by default. This allows users to easily switch between multiple files within the same folder, while extending the functionality of the *Media* tab.

To test this functionality, select any media (video or image) file and swipe to the next/previous file in the folder. Audio files should be excluded and video/audio playback should stop automatically, when switching to another file. The same applies to the *Media* tab, which now combines video files and images into a single view.

Previously, video files would open in a separate fragment, requiring users to exit the preview to view the next file. There were also some inconsistencies, when previewing a set of files containing offline and online media.

---

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
@@ -60,6 +60,7 @@ import com.owncloud.android.lib.resources.shares.ShareeUser;
 import com.owncloud.android.lib.resources.status.CapabilityBooleanType;
 import com.owncloud.android.lib.resources.status.OCCapability;
 import com.owncloud.android.operations.RemoteOperationFailedException;
+import com.owncloud.android.ui.preview.PreviewMediaFragment;
 import com.owncloud.android.utils.FileStorageUtils;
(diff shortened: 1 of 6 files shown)


alperozturk96:
@ZetaTom App crashes when I want to open a picture


https://github.com/nextcloud/android/assets/67455295/1b9e634f-ad62-4c41-86cd-3e2a8213b2ae

alperozturk96 (changes requested):
Crashes at PreviewImageActivity:200

alperozturk96:
Working with latest live photo implementation


https://github.com/nextcloud/android/assets/67455295/d79e6d79-6400-45af-b364-8914c7fbbd1b

ZetaTom:
Since this pull request was opened, the media player has been reworked substantially. Especially how we handle media (excl. images) playback using the `PreviewImageActivity` has changed to accommodate better a better and more up-to-date UI/UX.

The original idea was to use the `PreviewImageActivity` for images and videos since it already has a view pager in place which actually handles both media types. Given the newly reworked media flow, we can no longer simply replace all calls and use `PreviewImageActivity` for everything.

Superseded by #14922.

PR #12162, opened 2023-11-13
Bugfix Search Not Working When Sharing File From Outside of the App

alperozturk96:
### Steps to Reproduce

1. Open native file app of the device
2. Select a file or files
3. Tap share icon from native files app
4. Select Nextcloud App from bottom-sheet
5. Try search

@tobiasKaminsky @ZetaTom In adapter I saw logic for setting thumbnail for Images already exist in user's root directory. However for upload user can only choose folder and user can't select image for upload path. IMO we should only show directories. What do you think about it?

### Before
https://github.com/nextcloud/android/assets/67455295/93a22c24-4e19-4d62-ab13-79d1bd454eab

### After
https://github.com/nextcloud/android/assets/67455295/7828cfeb-951d-41cf-925b-9bd962d4712f


ZetaTom (approved):

JonasMayerDev:
Maybe it could also be a consideration to make the search recursive? It would have the upside of getting to nested folders fast, but the downside of not being so simple. I think I like it the way it is, but maybe it is worth a consideration?

I am also not sure if searching should show files, since there is no point of finding a file when you have to select a folder. The only case I could imagine would be if you want to see if another file that is e.g. related to the file you want to upload is in this folder too... Maybe this is also worth a consideration?

PR #12179, opened 2023-11-16
Fix SSO on applications with shared uid

erfanoabdi:
On original behavior of SSO first Nextcloud main app fetches the sso client package name by using `getCallingActivity` which is accurate and returns back the correct value and proceeds to token generation and so on: https://github.com/nextcloud/android/blob/6103234a21e0b89934a11886aa4b7ecb2b2c6a7e/app/src/main/java/com/owncloud/android/ui/activity/SsoGrantPermissionActivity.java#L93

But afterwards in token validation app is using `Binder.getCallingUid` which is the only way to obtain the caller package name, but in this case some SSO client applications that are using shared uid `android:sharedUserId` fails because `getNameForUid` returns only the first matched results in installed packages

So basically this PR just iterates through all the apps with matched UID and validates the token

More info: https://stackoverflow.com/questions/12918731/how-to-get-application-package-name-or-uid-which-is-trying-to-bind-my-service-fr

_Fix for the following error:_
```
com.nextcloud.android.sso.exceptions.TokenMismatchException: Provided authentication token does not match or package name is not permitted to perform this action
```

### DIFF
--- a/app/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
+++ b/app/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
@@ -457,16 +457,23 @@ public class InputStreamBinder extends IInputStreamService.Stub {
     }
 
     private boolean isValid(NextcloudRequest request) {
-        String callingPackageName = context.getPackageManager().getNameForUid(Binder.getCallingUid());
+        String[] callingPackageNames = context.getPackageManager().getPackagesForUid(Binder.getCallingUid());
 
         SharedPreferences sharedPreferences = context.getSharedPreferences(SSO_SHARED_PREFERENCE,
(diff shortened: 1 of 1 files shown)


alperozturk96 (approved):

ZetaTom (commented):
Thank you for taking the time to fix this. In my testing, this change has worked well.

Please note that we have a check in place to block any commits that don't include a Signed-off-by line in the commit message. Refer to the [appropriate check](https://github.com/nextcloud/android/pull/12179/checks?check_run_id=18743499110) for more information, and a guide on how to sign off commits for this pull request.

erfanoabdi:
Thanks @ZetaTom , added the sign off line

ZetaTom (approved):

tobiasKaminsky (approved):

PR #12182, opened 2023-11-17
Update Styles For Material Design 3

alperozturk96:
### Diff (new style on the left side)
diff

Basically just more rounded corners...


AndyScherzinger (approved):
fine to merge 👍 

There would be some changes needed on top for M3 which are currently hard-wired outside of the style definition via values.xml used in the layouts

AndyScherzinger:
@nimishavijay yes, makes sense and there will be further changes beyond them to move to M3 overall.

PR #12184, opened 2023-11-17
Convert Events to Kotlin

alperozturk96:


app/src/main/java/com/owncloud/android/ui/events/DummyDrawerEvent.kt
```diff
@@ -15,12 +15,11 @@
  * GNU Affero General Public License for more details.
  *
  * You should have received a copy of the GNU Affero General Public License
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ * along with this program.  If not, see <http:></http:>//www.gnu.org/licenses/>.
```
JonasMayerDev, line 18:
```suggestion
 * along with this program.  If not, see <http://www.gnu.org/licenses/>.
```

app/src/main/java/com/owncloud/android/ui/events/EncryptionEvent.kt
```diff
@@ -0,0 +1,30 @@
+/*
+ * Nextcloud Android client application
+ *
+ * @author Tobias Kaminsky
```
JonasMayerDev, line 4: Is Tobias really author?

app/src/main/java/com/owncloud/android/ui/events/FavoriteEvent.kt
```diff
@@ -15,14 +15,11 @@
  * GNU Affero General Public License for more details.
  *
  * You should have received a copy of the GNU Affero General Public License
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ * along with this program.  If not, see <http:></http:>//www.gnu.org/licenses/>.
```
JonasMayerDev, line 18:
```suggestion
 * along with this program.  If not, see <http://www.gnu.org/licenses/>.
```

JonasMayerDev (approved):
Looks good to me. App seems to still work fine.

PR #12191, opened 2023-11-20
scrolling kills test

tobiasKaminsky:


tobiasKaminsky:
Not needed.
It was a server regression.

PR #12192, opened 2023-11-21
Fix Crash on Android 14 When Start Foreground Service

alperozturk96:

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -72,6 +72,9 @@
     <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
     <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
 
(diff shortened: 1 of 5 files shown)


tobiasKaminsky:
When is app crashing? So I can test without and with this PR?

alperozturk96:
@tobiasKaminsky Try to start foreground service on Android 14 emulator. e.g download and open pdf.  If I miss any other foreground service please inform me.

### Video

https://github.com/nextcloud/android/assets/67455295/e5e0e87f-92a0-416b-8080-a23c3aed5dac

JonasMayerDev:
I think fix is not needed in FileUpload service since it is not used when android version is above 10 as far as I know (uses the FilesUploadWorker instead). But it also doesn't add new issues, so I guess it shouldn't be a problem.

JonasMayerDev (approved):

PR #12194, opened 2023-11-22
Bugfix/Make cancelation work again for Android >= 11 via Worker   

JonasMayerDev:
File cancellation was broken for the "new" FilesUploadWorker that's used for uploads in Android >= 11. 
Uses the same cancel function as before and should work like in Android <= 10. 

To test: 
Upload (big) file, go to uploads and press on the "X" for an upload. Now it should work as expected, while before it just did nothing. 

Do I need tests here?


alperozturk96:
@JonasMayerDev When you add or modify some kotlin files you need to run these three command from Android Studio terminal to fix spotlessKotlinCheck

./gradlew detekt -> If something wrong it will show everyting in html file so you check your code and fix it. You can find html file from Android Studio log
./gradlew spotlessCheck
./gradlew spotlessApply

If you create new file add new line at the end of the file always and also you can reformat code from Android Studio for styling. Before format code [set](https://stackoverflow.com/questions/30851617/how-to-set-max-line-length-in-android-studio-code-editor) max line lenght limit. Max line lenght for our projects is 120.

I hope it helps :)

alperozturk96 (changes requested):
Spotbugs increased please check dodgy codes

JonasMayerDev:
> Spotbugs increased please check dodgy codes

I think not related to me, fixed by merging master?

alperozturk96 (approved):
Working tested

ZetaTom (approved):

PR #12199, opened 2023-11-26
Remove GetSharesForFileOperation class used only by tests

starypatyk:
When working on #11255 I have found that the `GetSharesForFileOperation` class is used only by tests. There are a few references to it in the application code, but it is **instantiated** only in `GetSharesForFileOperationIT` test class.

I [discussed this some time ago](https://github.com/nextcloud/android/pull/11255#issuecomment-1427145648), I suggest to remove this class and other related code to unblock progress of #11255.


starypatyk:
The `GetSharesForFileOperation` class is now used by the application code (in the `RemoteShareRepository` class used by the `FileDetailSharingFragment`).

This PR no longer makes sense.

PR #12092, opened 2023-10-25
Bugfix After Selecting File From Search It Doesn't Open

alperozturk96:
Before:
https://github.com/nextcloud/android/assets/67455295/8218891c-880d-40c4-a92c-64feb76f00b0

After:
https://github.com/nextcloud/android/assets/67455295/08b00a98-f6e2-4990-a1e4-542f7441eb26

- [x] Tests written, or not not needed

PR #12209, opened 2023-11-28
Add Files Action to Search

alperozturk96:
**Behaviour:**
https://github.com/nextcloud/android/assets/67455295/d09e263c-4311-4391-8f99-57ed8e46e42f

This feature needed for [this](https://github.com/nextcloud/android/pull/12092) bugfix


app/src/main/res/layout/unified_search_item.xml
```diff
@@ -107,5 +107,19 @@
...
+    <ImageButton
+        android:id="@+id/more"
+        android:visibility="gone"
+        android:background="@color/transparent"
+        app:srcCompat="@drawable/ic_dots_vertical"
+        app:tint="@color/standard_grey"
+        android:layout_gravity="center"
+        android:layout_width="50dp"
+        android:layout_height="50dp"
+        android:contentDescription="open file actions"
```
ZetaTom, line 120: Please replace the hardcoded text with an `R.string` based one. We already have a similar button in `list_item.xml`, called `overflow_menu`, which could be used as a template.

ZetaTom (changes requested):

app/src/main/res/layout/unified_search_item.xml
alperozturk96: Fixed

JonasMayerDev (approved):

tobiasKaminsky:
For me it does not work.
When clicking on "three dots" of a search item, it opens root folder in background.
This looks different in your video, and seems also be working for Jonas.

[Screen_recording_20231130_072915.webm](https://github.com/nextcloud/android/assets/5836855/36a798ab-987c-46e4-8ae9-db75750d86b7)

alperozturk96:
> For me it does not work. When clicking on "three dots" of a search item, it opens root folder in background. This looks different in your video, and seems also be working for Jonas.
> 
> EDIT: after reviewing above video, it is the same. So I think this is a bit wrong, as still the search view should be shown. Imagine that an user accidently clicks on "three dots" for wrong file and then needs to search again.

It's same for selecting file. If user tap wrong file needs to search again. So the question has to be asked here are we going to change all search functionality with better selection and showing files action or do we leave the selection process as it is.

Given the current structure of the project, this solution seems appropriate.

JonasMayerDev (approved):
Works for me :)

PR #12213, opened 2023-11-29
Feature/Add cancel button to upload notification

JonasMayerDev:
Adds a cancel button to the upload notification that previously only showed the progress of the upload and (as positive side effect) makes the notification expandable and show filename and percentage. 


alperozturk96 (approved):
Working.

PR #12228, opened 2023-12-04
Provide Better Feedback For Failed Uploads

alperozturk96:
When upload failed user may want to navigate to Upload List so I added action for that into notification. Before this implementation user can just open conflict resolve dialog or dismiss notification

If user taps the notification goes to upload list
If user taps the resolve conflict FileResolveConflictDialog appears

Screenshot 2023-12-06 at 10 55 35

Sometimes file cannot be found to solve conflict and nothing happens. I added snackbar message to provide better feedback
https://github.com/nextcloud/android/assets/67455295/2f36f36a-377e-4c77-8038-9a9415461462


JonasMayerDev (changes requested):
Functionality of quick access to "FileUploadList" already exists. You should actually be able to simply click on the notification and be redirected. Maybe this is not working in Android 14? The implementation for this is in notifyUploadResult a view lines down from your reimplementation. Maybe get the original implementation to work?

JonasMayerDev (approved):
Works for me! Thanks for the change :)

AndyScherzinger (approved):

PR #12238, opened 2023-12-06
Use ServiceCompat For Foreground Service

alperozturk96:
### Problem
ServiceCompat automatically manages Android version compatibility when initiating foreground services. However, due to the absence of Foreground Service Types in older Android versions, it becomes necessary to explicitly check the API Level within the Service implementation.

### Solution
To address this, we can create a wrapper enum class for our foreground services. This allows us to use a unified approach across all Android versions and leverage ServiceCompat without duplicating the same code block for each foreground service type.


AndyScherzinger:
https://github.com/nextcloud/android/pull/12238/commits/647e0b74ab6f496cc6068f4e7e5dd45174bb627c

`rebase` please, not master merges ;)

app/src/main/java/com/owncloud/android/files/services/FileUploader.java
```diff
@@ -1,27 +1,24 @@
...
+ * <p>
+ * This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public
+ * License version 2, as published by the Free Software Foundation.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+ * warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ * <p>
+ * You should have received a copy of the GNU General Public License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/>.
```
AndyScherzinger, line 21: can you remove this reformatting of the license header please? 💙

alperozturk96: Sure, I don't know how it happened.

Issue #10954, opened 2022-10-28
Upload progress not shown in Uploads screen with new uploader


AlvaroBrey:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Use android 12+ or a beta version to get new upload Worker
2. Upload some files
3. Open the Uploads screen

### Expected behaviour

The progress for the currently uploading file should be shown

### Actual behaviour

All files go directly from "pending upload" to "done" without showing progress at all. The progress is correctly shown in notifications.

### Android version

12+

### Device brand and model

Any

### Stock or custom OS?

Stock

### Nextcloud android app version

3.22.0+

### Nextcloud server version

Any

### Using a reverse proxy?

I don't know

### Additional information

cc @tobiasKaminsky

PR #12246, opened 2023-12-06
Fix: To many old notifications  

JonasMayerDev:
Spawns the upload failed notification with a tag that is unique for every file upload operation to make sure only one error notification (e.g. file conflict) is displayed at a time per file upload. 
Also cancels error notification when upload makes progress or finishes successfully to make sure no old not relevant notifications are displayed.

Fixes issues described in https://github.com/nextcloud/android/issues/11974 where one conflict creates a lot of notifications because of automatic retry in background.

To reproduce the bug:
1. upload file that already exists -> create file conflict
2. retry upload in uploads menu via retry all (or wait till automatic retry in background)
3. You should get a new notification every time the upload is retried
4. Solve conflict
5. The notifications should still exist

- [x] Tests written, or not not needed

PR #12247, opened 2023-12-06
Implement other interface functions regarding file upload for upload worker

JonasMayerDev:
Reimplements the existing interface functions like get upload status or add transfer progress listener for new upload worker.
Fixes: https://github.com/nextcloud/android/issues/10954 and probably a lot of other issues where code used that functions but just did nothing

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/FilesUploadWorker.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/FilesUploadWorker.kt
@@ -56,6 +56,7 @@ import com.owncloud.android.ui.activity.ConflictsResolveActivity
 import com.owncloud.android.ui.activity.UploadListActivity
 import com.owncloud.android.ui.notifications.NotificationUtils
 import com.owncloud.android.utils.ErrorMessageAdapter
+import com.owncloud.android.utils.FilesUploadHelper
 import com.owncloud.android.utils.theme.ViewThemeUtils
 import java.io.File
 import java.security.SecureRandom
@@ -77,7 +78,7 @@ class FilesUploadWorker(
     private val notificationManager: NotificationManager =
         context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
     private val fileUploaderDelegate = FileUploaderDelegate()
-    private var currentUploadFileOperation: UploadFileOperation? = null
+
 
     override fun doWork(): Result {
         val accountName = inputData.getString(ACCOUNT)
@@ -379,7 +380,7 @@ class FilesUploadWorker(
             val text = String.format(context.getString(R.string.uploader_upload_in_progress_content), percent, fileName)
             notificationBuilder.setContentText(text)
             notificationManager.notify(FOREGROUND_SERVICE_ID, notificationBuilder.build())
-            //FilesUploadHelper.onTransferProgress(,progressRate,totalTransferredSoFar,totalToTransfer,fileAbsoluteName)
+            FilesUploadHelper.onTransferProgress(currentUploadFileOperation?.user?.accountName,currentUploadFileOperation?.remotePath,progressRate,totalTransferredSoFar,totalToTransfer,fileAbsoluteName)
         }
         lastPercent = percent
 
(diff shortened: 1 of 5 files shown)


TobiPeterG:
This fixes #11974 for me (and likely related issues)

TobiPeterG:
> @TobiPeterG are you sure you mean this PR? I mean it's totally possible in the end it also should fix some bugs regarding uploading but it's a little bit random that you found this PR? Are you sure you don't mean this PR I mentioned in the Issue:  https://github.com/nextcloud/android/pull/12246

All I can say is: with the original app I got the notification spamming
With this version, no issues appear at all (really, no conflict issue at all)
However, it might be completely unrelated. Maybe something else was merged that fixed it
I'll try the original app again

JonasMayerDev:
@TobiPeterG nice and thanks but just to be completely sure, there is a similar PR from me that is about reducing the notification spamming: https://github.com/nextcloud/android/pull/12246.
This one, where we are right now #12247 could also fix the issue but it would be a little surprising to me :)

TobiPeterG:
> @TobiPeterG nice and thanks but just to be completely sure, there is a similar PR from me that is about reducing the notification spamming: https://github.com/nextcloud/android/pull/12246.
> This one, where we are right now #12247 could also fix the issue but it would be a little surprising to me :)

Well, I don't know what solved it, I just checked the official app again. Still the notification spamming (and errors at all)
With this version, I don't get any issues (though that could mean that the error isn't being shown?)

TobiPeterG:
> @TobiPeterG if really this PR changed it and not https://github.com/nextcloud/android/pull/12246 I think this could maybe fix the root cause (I don't know how exactly).


Hmm the issue also doesn't appear with the other app version.
However, it didn't seem to upload any of the images yet

brentobi:
@JonasMayerDev
A laymans idea as to why this PR could resolve the sync error from https://github.com/nextcloud/android/issues/11974:

Is the following assumption true for your change to [`FieldUploader::isUploading`](https://github.com/nextcloud/android/pull/12247/files#diff-3bd067a5809ebf045c846e5baeb5cd2628c83471ccf76ec39adde1601d650361L1251)?

**Previous behaviour(?)**: Only pending uploads were detected, which were triggered by the same instance of FileUploader. (`mPendingUploads` is private)
**New behaviour(?)**: Pending uploads of all instances of FileUploader are detected because `mUploadsStorageManager` seems to read from DB.

**If that assumption is true**, it would explain what I think causes the sync errors on my device. Key observation was that `FileSyncWork::doWork` was running twice in parallel, triggering a second `FileUploader::uploadNewFile` before the first upload even started.

If I understand this PR correctly the second instance of the FileUploader would now detect a pending upload of the first one because it is read from db instead of a private field. (unless there is a very close-call race condition between checking and scheduling uploads)

I posted my reduced and annotated **log files here** some time ago: https://github.com/nextcloud/android/issues/11974#issuecomment-1765444062

P.S.: Thanks for your work!

Kind regards

JonasMayerDev:
> > @TobiPeterG if really this PR changed it and not https://github.com/nextcloud/android/pull/12246 I think this could maybe fix the root cause (I don't know how exactly).
> 
> 
> Hmm the issue also doesn't appear with the other app version.
> However, it didn't seem to upload any of the images yet

@TobiPeterG did uploading work with this PR or are both versions not uploading?

TobiPeterG:
> > > @TobiPeterG if really this PR changed it and not https://github.com/nextcloud/android/pull/12246 I think this could maybe fix the root cause (I don't know how exactly).
> > 
> > 
> > Hmm the issue also doesn't appear with the other app version.
> > However, it didn't seem to upload any of the images yet
> 
> @TobiPeterG did uploading work with this PR or are both versions not uploading?

Both haven't yet, but I don't know if I'll have to wait a bit longer

JonasMayerDev:
@brentobi currently there are 2 methods of uploading files. For Android i think it was 10 and older we use a Service that runs the upload task in the background. For newer android versions we use a Worker (FileUploadWorker.tk) to do the upload. The changes here implement old functions that worked for Upload Service now also for the upload worker. The variable you mentioned is not used when the Worker is used for upload so i think it's sadly more complicated. What could be the case is that as you said the hopefully now also for the worker working implementation of isUploading fixes the issue? But im not sure if really this PR fixes the issue or master already fixes the Issue or if the issue is even solved at all by any changes we made or if just uploading is broken with this PR xd I think testing is now needed maybe also from people that have the issue... Thanks for looking into it :)

alperozturk96 (approved):

Issue #11337, opened 2023-01-31
UI drawer overlap issue


womeier:
- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. open app

### Expected behaviour

![danach](https://user-images.githubusercontent.com/55190123/215866827-11ae521f-8e6f-45c7-b2b2-49d247cbc5db.png)

### Actual behaviour

![davor](https://user-images.githubusercontent.com/55190123/215866467-97d5b5fb-f097-4cac-925d-d6ebb6c9a275.png)

### Android version

11

### Device brand and model

Pixel XL (virtual device)

### Stock or custom OS?

Stock

### Nextcloud android app version

master branch as of 31.1.23 (051513d)

### Nextcloud server version

not relevant

### Using a reverse proxy?

No

womeier:
I'd like to fix this, but I wasn't able to find the correct color (for the dark theme)

It seems, the drawer background depends on a theme/style and is not "@color/appbar" on my phone.

AlvaroBrey:
As I commented in #11336:

> This should probably be fixed by setting a padding or margin in some of the overlapping elements, rather than a background color; I haven't had time to look at it unfortunately

satvik2131:
What could be a better approach  , to change the background color , or to make the quota bar at the bottom and when we scroll up it will not be shown , please suggest ? @ZetaTom

AndyScherzinger:
Hi at @satvik2131 - see https://github.com/nextcloud/android/issues/11337#issuecomment-1427714404 either via a background color or finding a way for padding/margin to prevent scrolling behind the quota.

satvik2131:
hey , I know this must not be the right place to ask , but I want to discuss some doubts , is there any discord channel or any communication medium ?? 

Question: the question is like if we scroll should the quota bar should hide or have to be shown ??

ZetaTom:
> hey , I know this must not be the right place to ask , but I want to discuss some doubts , is there any discord channel or any communication medium ??
> 
> Question: the question is like if we scroll should the quota bar should hide or have to be shown ??

That is a good question. I think we should leave that decision up to @nextcloud/designers.

[...]

PR #12261, opened 2023-12-10
Fix UI drawer overlapping #11337

satvik2131:
I have added appbar color as backround , suggest me any changes if there is required ? 
Screenshot 2023-12-10 125916


ZetaTom (changes requested):
Firstly, thank you for taking the time to suggest and implement a fix for this problem.

At first glance, the solution seems to be working. Upon closer inspection, I noticed a slight colour mismatch between the background colour of the navbar and `@color/appbar`.

Also, please remove any changes not related to this issue from this pull request, such as those in `PreviewImageFragment`, as this will make the review process easier. Feel free to open a separate pull request for those changes.

satvik2131:
Hi @ZetaTom what should be good changing the background color or making it aligned at the bottom only , so that if we scroll up we won't see it ?

PR #12265, opened 2023-12-11
Bugfix Despite Connection Is Restored Message About Unreachable Server Still Present

alperozturk96:


alperozturk96:
It will be released for 3.27.1

PR #12276, opened 2023-12-12
Fix infinite loading of folders in shared tab

ZetaTom:
This change fixes an issue where users were left with an infinite *Loading...* message, when attempting to open a folder from the *Shared* tab.

### Steps to reproduce
1. open Nextcloud Android client
2. create empty folder in root directory
3. share said folder with another user on the same instance
4. open *Shared* tab
5. tap on the new folder

Previously the *Loading...* message would be displayed without any further feedback.

Supersedes #11401

---


ZetaTom:
Screenshot test failure seems to be unrelated, as they highlight different fragments/activities.

JonasMayerDev (approved):
Fixes Bug for me

Issue #11797, opened 2023-07-10
Crash when copying multi-selection from favourites/shared tab


ZetaTom:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Install and setup new account in Nextcloud app
2. Navigate into *Photos* folder
3. Select all items and add to favourites
4. Switch to *Favourites* tab
5. Long-press on any item; select all
6. Select *Copy* or *Move* from action menu
7. In *Copy to…* dialogue create new folder (e.g. *A*)
8. Confirm move/copy

### Expected behaviour

Selected items are moved or copied to selected folder.

### Actual behaviour

After a suspiciously jumpy loading animation the app crashes. When examining what files have been copied/moved, there is always a subset of the selected files present.

Filename | Present after crash?
--- | ---
`Birdie.jpg` | :x:
`Frog.jpg` | :heavy_check_mark:
`Gorilla.jpg` | :heavy_check_mark:
`Library.jpg` | :heavy_check_mark:
`Nextcloud community.jpg` | :heavy_check_mark:
`Readme.md` | :x:
`Steps.jpg` | :x:
`Toucan.jpg` | :x:
`Vineyard.jpg` | :x:

### Android version

11 (API level 30; AOSP image)

### Device brand and model

Google Pixel 2 Emulator (Android Studio)

### Stock or custom OS?

Stock

### Nextcloud android app version

ba292cfd2bf449235c3e84edbc6d11d2a70cd37e

### Nextcloud server version

27.0.0

### Using a reverse proxy?

No

### Android logs

[log.txt](https://github.com/nextcloud/android/files/12002762/log.txt)

ZetaTom:
Not reproducible with the latest `master` version.

Closing this issue as there have been major changes made to the way the recycler view, its items and the corresponding positions are handled internally.

PR #12283, opened 2023-12-13
Fix Multi Select Crash

alperozturk96:

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
@@ -89,6 +89,7 @@ import java.util.Date;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
+import java.util.UUID;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -289,21 +293,61 @@ public void setEncryptionAttributeForItemID(String fileId, boolean encrypted) {
...
+
+        if (isFilesEmpty) {
+            if (shouldShowHeader()) {
+                return headerId;
+            } else {
+                return footerId;
+            }
+        } else {
+            return mFiles.get(position).getFileId();
         }
```
ZetaTom, line 308:
```suggestion
        if (shouldShowHeader()) {
            if (position == 0) {
                return headerId;
            }

            // skip header
            position--;
        }

        if (position == mFiles.size()) {
            return footerId;
        } if (position < mFiles.size()) {
            return mFiles.get(position).getFileId();
        }

        // fallback
        return RecyclerView.NO_ID;
```

With this patch in place, it no longer crashes on my test device.

tobiasKaminsky (approved):

JonasMayerDev (approved):
Works for me

joshtrichards:
Was this for #11797?

ZetaTom:
> Was this for #11797?

No, this PR seved to fix a more generic issue, that would occur when selecting the last item in a given directory. Previously, I've managed to reproduce the issue using the following steps:

1. set up new Nextcloud user in app using default skeleton
2. open *All files* tab in the app
3. long press on the second to last item in the root directory (e.g. *Reasons to use Nextcloud.pdf*)
4. add the last item (e.g. *Template credits.md*) to the selection

Following these steps would crash the app with the following exception:

```
java.lang.IllegalStateException: Two different ViewHolders have the same stable ID. Stable IDs in your adapter MUST BE unique and SHOULD NOT change.
```

The exception was caused by `OCFileListAdapter:getItemId()` returning incorrect ids in some cases. These were in conflict with `setHasStableIds(true);`, which requires each item in the recycler list view to have a unique and persistent id.

PR #12298, opened 2023-12-15
restore fab icon when going back

parneet-guraya:
Close: #11962

https://github.com/nextcloud/android/assets/111801812/198e0eb5-e4a4-4c6f-87b7-778686c71b52

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -1025,6 +1025,7 @@ public class FileDisplayActivity extends FileActivity implements FileFragment.Co
      */
     private void popBack() {
         // pop back fragment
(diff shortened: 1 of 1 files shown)


alperozturk96 (approved):

ZetaTom (commented):
Thank you for fixing this issue. In my brief tests, your solution works well.

However, there are some cases where the wrong icon is displayed. For example, if you tap on the header to open the preview (pencil icon appears) and then switch to another activity, such as the *Activities* tab; when you return to the *All Files* tab, the pencil icon appears instead of the plus sign.

parneet-guraya:
> Thank you for fixing this issue. In my brief tests, your solution works well.
> 
> However, there are some cases where the wrong icon is displayed. For example, if you tap on the header to open the preview (pencil icon appears) and then switch to another activity, such as the *Activities* tab; when you return to the *All Files* tab, the pencil icon appears instead of the plus sign.

Yes you're right. I will push the solution handling all the cases ASAP.

PR #12303, opened 2023-12-18
Use Only Work Manager For File Upload

alperozturk96:
FileUploader service still exists. This implementation only uses worker. In next PR FileUploader service will be removed.


JonasMayerDev:
@alperozturk96 What do you think about trowing an error in FileUploader.java in the beginning, of onStartCommand() to make absolutely sure the Service is not run under any circumstance? (To make sure no duplicated uploads are created)

JonasMayerDev (approved):
Simple file upload still works for me on: Android 10, 12 (real device), 14

PR #12308, opened 2023-12-20
Use Work Manager For File Download

alperozturk96:
- [x] Tests written, or not not needed

PR #12327, opened 2023-12-25
Remove File Upload Service

alperozturk96:


alperozturk96:
First we need to merge Use Work Manager For File Download [#12308](https://github.com/nextcloud/android/pull/12308) then we can use common logics for this PR (for e.g removing ServiceConnection logic) as well.

JonasMayerDev (approved):
Seems like everything is working as before or better...

tobiasKaminsky (approved):

ZetaTom (commented):
Other than the one thing I've commented on, which might just be a red herring, this works perfectly.

ZetaTom (approved):

PR #12336, opened 2024-01-01
Changing the Documentation from the Toolbar Options

supreet-2003:
A small change in the search bar hint from "Search in NextCloud" to "Explore in NextCloud".
Screenshot 2024-01-01 104738

### DIFF
--- a/app/src/main/res/layout/toolbar_standard.xml
+++ b/app/src/main/res/layout/toolbar_standard.xml
@@ -174,7 +174,7 @@
                     app:layout_constraintLeft_toRightOf="@id/menu_button"
                     app:layout_constraintRight_toLeftOf="@id/switch_account_button"
                     app:layout_constraintTop_toTopOf="parent"
-                    tools:text="Search in Nextcloud" />
+                    tools:text="Explore in Nextcloud" />
 
                 <com.google.android.material.button.MaterialButton
                     android:id="@+id/switch_account_button"


joshtrichards:
Hi @supreet-2003 - Thanks for the PR. Any particular reason you'd prefer this be changed?

supreet-2003:
Hi Josh Richards,

Thank you for getting back to me.

The reason for choosing the word 'explore' is that Search is when people
find your app via a search for a keyword, explorer is when they come across
your app in another way.

In simple words, search word is used when you start finding something from
the scratch whereas Nextcloud already gives 'access' to your files which
just make the keyword 'explore' more of a better keyword.

These small changes might not look like a differentiator but it definately
gives the overall UI a professional look.

I would be looking forward to your feedback.

Supreet Prajapati

https://www.linkedin.com/in/supreet-prajapati-8b5932258?utm_source=share&utm_campaign=share_via&utm_content=profile&utm_medium=android_app

tobiasKaminsky:
tools:text is only used for rendering a preview inside Android Studio.
The real text is set/changed here: 
https://github.com/nextcloud/android/blob/74868302a275d12f2f0653fddb3f09786e3cd232/app/src/main/java/com/owncloud/android/ui/activity/ToolbarActivity.java#L167

@jancborchardt what do you think?

supreet-2003:
Well that's something totally depending on the functionality of the
application. Anyways, Thanks for reaching out to my PR. Looking forward to
more contributions.

tobiasKaminsky:
Thanks for you contributing, even though this time it is not merged.
We are happy if you stay around and help us :-)

If you want to do something new, but are in doubt, please first create an issue, so we can discuss it and then develop it :)

PR #12345, opened 2024-01-08
Make sure upload is only added once to db

JonasMayerDev:
Make sure that upload can not be added twice to db to help with https://github.com/nextcloud/android/issues/11974
Should not harm: Why would user want to add upload twice and not update existing one? 
For Example, if an upload failed and the user re-uploads, with this pr it updates the existing failed upload with "in progress" to try to upload it again instead of adding the same upload again and causing a conflict next time failed uploads are retried. (See Video)
In addition to that, fixes some issues that occur when duplicates of same upload are in upload list since most functions assume they find only one upload (see uploads tab in video on master)  
[Screen_recording_20240108_112327.webm](https://github.com/nextcloud/android/assets/43114340/57f376e9-4f98-4f3b-96f1-ba0260644314)
Master
[Screen_recording_20240108_112525.webm](https://github.com/nextcloud/android/assets/43114340/b4f6973d-7313-4df2-ac17-ad774d342565)
This PR

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
@@ -87,6 +87,18 @@ public class UploadsStorageManager extends Observable {
      * @return upload id, -1 if the insert process fails.
      */
     public long storeUpload(OCUpload ocUpload) {
+        long startTime = System.nanoTime();
+        OCUpload existingUpload = getPendingCurrentOrFailedUpload(ocUpload);
+        if(existingUpload != null){
+            Log_OC.v(TAG, "Will update upload in db since " + ocUpload.getLocalPath() + " already exists as pending, current or failed upload");
+            ocUpload.setUploadId(existingUpload.getUploadId());
+            return updateUpload(ocUpload);
+        }
+
+        long endTime = System.nanoTime();
+
+        long duration = (endTime - startTime);  // compute the elapsed time in nanoseconds
+
         Log_OC.v(TAG, "Inserting " + ocUpload.getLocalPath() + " with status=" + ocUpload.getUploadStatus());
 
         ContentValues cv = getContentValues(ocUpload);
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -100,14 +111,27 @@ public long storeUpload(OCUpload ocUpload) {
...
             notifyObserversNow();
+
             return new_id;
         }
+
     }
 
     public long[] storeUploads(final List<OCUpload> ocUploads) {
         Log_OC.v(TAG, "Inserting " + ocUploads.size() + " uploads");
+        storeUpload(ocUploads.get(0));
```
ZetaTom, line 122: I'm not sure, what the purpose of this line is. Aren't the individual uploads handled by the `for`-loop?

ZetaTom (commented):
During my brief tests, this change worked well.

app/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
JonasMayerDev: You are absolutely right. Added it for testing and forgot to remove it. Sorry and thanks :)

alperozturk96 (approved):

rloutrel:
Thank you 100 times!

It is nice to provide something to fix this. But I am not sure about what you mean with 'if an upload failed' and the consequences of 'with "in progress" to try to upload it again'.
Actually it does not fail, it is just that the work in the backend takes longer than what the application is waiting to validate the upload at the end, like some post upload processing before sending final Ack (it could also append with smaller document and the interrupted connections of smartphone apps). Therefor no new upload should be done to replace it. It would change the problem from 'infinite' upload loop to an 'infinite' replace loop (better for storage, if we exclude hdd/sdd read/write limitations, but still not sufficient).

Instead of that, the existing file and the file to be uploaded should be compared to evaluate if they are equivalent (size+name+md5+sha256 OR something similar OR some download ID OR some other technique).

Issue #12338, opened 2024-01-03
Collabora not loading in app v3.27.0


expert-geeks:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Upgrade Nextcloud android app to version 3.27.0
2. Open document in app

### Expected behaviour

Collabora should start and open the document as usual

### Actual behaviour

App seems to try to open the document, but does not complete and hangs at the opening document screen. App hangs permanently and does not time-out. No error returned. CODE Admin console doesn't show the user attempting to log in.

Downgrade from v3.27.0 to v3.25.0 and the documents open as usual.

### Android version

10

### Device brand and model

OnePlus 5

### Stock or custom OS?

Stock

### Nextcloud android app version

3.27.0

### Nextcloud server version

25.0.13

### Using a reverse proxy?

Yes

### Server error logs
There's nothing in the nginx error log, here are the access logs;

```bash
xx.xx.xx.xx - user [03/Jan/2024:17:14:34 +0000] "PROPFIND /remote.php/dav/files/user/path/ HTTP/1.1" 207 1331 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.27.0"
xx.xx.xx.xx - user [03/Jan/2024:17:14:34 +0000] "GET /ocs/v2.php/apps/files_sharing/api/v1/shares?path=%2FDocuments%2F&reshares=true&subfiles=true HTTP/1.1" 200 2480 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.27.0"
xx.xx.xx.xx - user [03/Jan/2024:17:14:36 +0000] "GET /ocs/v2.php/cloud/user?format=json HTTP/2.0" 200 566 "-" "Mozilla/5.0 (Android) Nextcloud-android/3.27.0"

then nothing...
```

### Additional information

logcat file doesn't seem to contain any information about app usage, there's plenty in there about signal strength etc but nothing that seems related to nextcloud. Is there a recommended 'filter' to use with logcat to generate useful logs ?

[...]

PR #12349, opened 2024-01-09
Fix collabora files not loading

JonasMayerDev:
Allow loadUrl with extra_url = null to create new document viewer.
Opening a collabora file (eg. docx file) before was not possible (infinite loading screen).
Fixes https://github.com/nextcloud/android/issues/12338 


tobiasKaminsky (approved):

connelhooley:
> APK file: https://www.kaminsky.me/nc-dev/android-artifacts/12349.apk qrcode To test this change/fix you can simply download above APK file and install and test it in parallel to your existing Nextcloud app.

This link isn't working for me, is there a rough idea for when the app will be updated in the app store? Or is there anywhere else to get the APK? I'm just setting NextCloud up for the first time so just wanted to see it all working in the app. Thanks for your work fixing this

PR #12408, opened 2024-01-24
Client cert

tobiasKaminsky:
Replaces https://github.com/nextcloud/android/pull/11314


tobiasKaminsky:
@Elv1zz I have opened it here again, to have working CI and rebased it.
Many thanks for your great PR.

Can you test this once again?
I will also test it soon.

After that I will merge it.

Elv1zz:
> @Elv1zz I have opened it here again, to have working CI and rebased it. Many thanks for your great PR.

:)

> Can you test this once again? I will also test it soon.

I tested it for about 24 hours now and everything seems to work fine.

There is one thing, which does not seem to work completely, but it might be a new thing still under development in the app itself: There is a new notification showing the "upload preparation" progress, which is never going beyond about 20%. Either this is not implemented completely, yet, or it does not fully work with client certificates. But in general photo upload in the background still works.
 
> After that I will merge it.

Yay, already merged 🎉  Looking forward to see this feature in the next release.

PR #12337, opened 2024-01-03
Fix auto upload does not start automatically

JonasMayerDev:
[Screen_recording_20240103_123529.webm](https://github.com/nextcloud/android/assets/43114340/2a10329f-61bc-452e-abd2-655eec888cdf)
- Add foreground info (also interesting for user) to prevent android killing the worker. (Only visible if worker is active more than ca. 1s -> Use large folder with many files for testing) 

- Make getFileChecksum a lot faster (previously over 1 h for a folder of 6000 Images now only minutes)
Screenshot_2024-01-02_13-37-46
Previously 7.5s
Screenshot_2024-01-02_13-45-05
Now 15ms
|_ Should be tested (maybe too good to be true)

For me, it also helps with https://github.com/nextcloud/android/issues/12141

- [X] Tests written, or not not needed

Issue #12141, opened 2023-11-07
excessive background battery usage


Blisk:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

My Android phone pushes me all the time notifications for "excessive background battery usage for Nextcloud".
Also some other users on Android phones, they get the same and when they tap that notification it kills Nextcloud.
After that, nobody has the latest pictures on Nextcloud until the app is started again.

### Expected behaviour

I expect the Nextcloud app won't drain the battery so much on Android, so it is on top of the list of excessive background battery usage.
Or somehow it will check for changes in picture folders so it can start when there is some change and start uploading photos on Nextcloud server. Or make it somehow for users to select, how often it checks folders with pictures in minutes.
It is really annoying when someone gets this notification and kills the app and than Nextcloud doesn't work.

### Actual behaviour

Open Nextcloud, than after a few hours get a notification and when the User clicks it, it kills Nextcloud app.

### Android version

12

### Device brand and model

Huawei P30 pro

### Stock or custom OS?

Stock

### Nextcloud android app version

3.26.0

### Nextcloud server version

27.1.1

### Using a reverse proxy?

No

joshtrichards:
> Or somehow it will check for changes in picture folders so it can start when there is some change and start uploading photos on Nextcloud server. 

That's the current behavior from my understanding.

- What are your typical (precise) auto-upload settings?
- Custom or auto-detected folders?
- What is the app actually doing? Are there queued uploads in the Uploads tab? Perhaps a lot of existing media that hasn't been uploaded before?

lckarssen:
This is maybe the same as #11983, which was closed because it might be caused by  #11974.

Issue #11983, opened 2023-09-21
Android app v 3.26.0 drains much more battery than before


lckarssen:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Upgrade Nextcloud Android app to v3.26.0. 
2. Use phone in the same way as before
3. Observe battery emptying more than before. The Nextcloud app is listed at the top of the list of battery consumers. 

### Expected behaviour

The Nextcloud app should be found somewhere at the bottom of the list of top power consumers.

### Actual behaviour

The app is found at the top of the list of battery power consumption. In my case it is now using roughly between 8% and 12% of battery, easily beating Firefox at ~7% and Always on Display at ~ 5%. 

### Android version

13

### Device brand and model

Samsung S22

### Stock or custom OS?

Stock

### Nextcloud android app version

3.26.0

### Nextcloud server version

27.1.0

### Using a reverse proxy?

Yes

joshtrichards:
This seems related to https://github.com/nextcloud/android/issues/11974#issuecomment-1729361029 so let's focus on addressing that. If this matter remains afterward, we can re-open it.

lckarssen:
> This seems related to [#11974 (comment)](https://github.com/nextcloud/android/issues/11974#issuecomment-1729361029) so let's focus on addressing that. If this matter remains afterward, we can re-open it.

The issue has remained, unfortunately. Instead of reopening this issue, I think follow-up in #7663 makes more sense.

PR #12413, opened 2024-01-26
Remove user notification for autoupload edge cases

JonasMayerDev:
## File Conflicts
This PR should finally fix https://github.com/nextcloud/android/issues/11974.
When ever an upload conflict occurs, the app checks if size, creation and modification date for the local file and remote file are the same. If so, the app assumes it is the same file and skips upload and adds the upload to the uploaded successfully tab with a bold text to inform the user the upload is skipped. The user is not directly notified when such a "same file conflict" occurs.
Screenshot_2024-01-25_18-55-35

## Local file removed
The same "silent" behavior is used when a local file is not present anymore to fix error notifications where users would delete a file directly after creating it (e.g. making a photo and deleting it) in auto upload as suggested by @tobiasKaminsky.  


tobiasKaminsky (approved):

app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
```diff
@@ -300,6 +302,16 @@ public void onBindViewHolder(SectionedViewHolder holder, int section, int relati
             case UPLOAD_FAILED -> itemViewHolder.binding.uploadDate.setVisibility(View.GONE);
             case UPLOAD_SUCCEEDED -> itemViewHolder.binding.uploadStatus.setVisibility(View.GONE);
         }
+
+        // show status if same file conflict or local file deleted
+        if (item.getUploadStatus() == UploadStatus.UPLOAD_SUCCEEDED &&
+        item.getLastResult() != UploadResult.UPLOADED){
+            itemViewHolder.binding.uploadStatus.setVisibility(View.VISIBLE);
+            itemViewHolder.binding.uploadStatus.setTypeface(null, Typeface.BOLD);
```
ZetaTom, line 310: A bold message looks a little out of place, in my opinion, as all other error messages are displayed using the normal typeface.

bold

ZetaTom (commented):
During my brief testing everything worked as expected.

lckarssen:
I have been using this QA APK for the last four days, since this PR was mentioned in https://github.com/nextcloud/android/issues/11974#issuecomment-1913137710 (alongside the regular Nextcloud app, v3.27.0). 

Interestingly, I haven't had any conflicts since. From neither app, so maybe the conflict resolution of the QA app also prevents the conflicts from happening in the regular version of the app. 

Moreover, when looking at the battery usage over the past few days (#12141, #11983), I see that the regular app has used 7.1% of battery, compared to 0.6% for the QA version of this PR.

JonasMayerDev:
@lckarssen I think the improved battery consumption has to do with this PR: https://github.com/nextcloud/android/pull/12337 that should also be included in future releases. It's good to hear that the conflicts are gone, but the changes in the QA app should not affect the normal app...

lckarssen:
> @lckarssen I think the improved battery consumption has to do with this PR: #12337 that should also be included in future releases. It's good to hear that the conflicts are gone, but the changes in the QA app should not affect the normal app...

@JonasMayerDev You were right about the fact that the QA app shouldn't affect the "old" app. I recently had a case again where some of the uploads in the old app had conflicts, whereas the QA uploads went fine. So, another :+1: for this PR.

JonasMayerDev:
@lckarssen That's really good to hear! Thank you for testing this PR :)

app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
JonasMayerDev: @ZetaTom, FYI I talked with Tobias, and he also said that it should not be bold but we for now just merged and i change that in a separate PR

tovine:
I've tried this build over the last week and saw this new message "Same file found on remote, skipping upload" several times, but there were also quite a few false duplicates that were _not_ caught.
So there seems to be some work left before it's 100%, but a definite improvement over how it used to be so I think it's a very good start!

Hopefully you'll revisit the idea of using file checksums to get a more robust check in the future, but in the meantime I'd like to say thank you for the work you put in! 🎉

JonasMayerDev:
> I've tried this build over the last week and saw this new message "Same file found on remote, skipping upload" several times, but there were also quite a few false duplicates that were _not_ caught. So there seems to be some work left before it's 100%, but a definite improvement over how it used to be so I think it's a very good start!
> 
> Hopefully you'll revisit the idea of supporting server side file checksums to get a more robust check in the future, but in the meantime I'd like to say thank you for the work you put in! 🎉

@tovine Do you use a multi account setup? If there are still conflicts, I have absolutely no idea where they come from, except if you use a multi account setup. Then I would guess the conflicts come from files that are uploaded to the wrong account and collide with files from the wrong account as documented here https://github.com/nextcloud/android/pull/12516?
If you don't use a multi account setup, I would suggest that we wait until more people tested the latest releases to see if it is still a problem for a large amount of people.

tovine:
No, I don't (at least not in the Nextcloud QA app).
I have however reconfigured the app with media folders to be synced that I've already synced with the app before, and selected to sync existing files as well, so in that sense I guess it's a "worst case scenario".

Yethal:
@AndyScherzinger would it be possible to fit this change in 3.28.0 release somehow? It's a major usability improvement for Android users.

Yethal:
I don't everything has been smooth sailing for the past few days.

PR #12415, opened 2024-01-27
Fix Japanese string typo for name sub_folder_rule_day

Shiritai:
Just a typo but it disturbed me for a long time (both 年 and 念 have the same spelling: nen).
- [x] No test needed


AndyScherzinger:
Hi @Shiritai Thank you very much for the fixed translation for the Japanese language file. I updated it on transifex where the translations are handled and auto-synced to this repository.

In case you want to take a look, you can find it here for the Android app
https://app.transifex.com/nextcloud/nextcloud/translate/#ja_JP/android

Closing the PR, since the next transifex sync will commit the fixed string here 👍

PR #12465, opened 2024-01-30
Deprecate OwncloudClient - Helpers

ZetaTom:
This is one of a series of pull requests which aim to replace all instances of `OwnCloudClient` with `NextcloudClient`. The reason for this change is that the newer `NextcloudClient` uses OkHttp, replacing the outdated Jackrabbit methods.

Specifically, this pull request implements the following:
- helper classes to ease the transition and refactoring
- general changes
- fallback mechanism to convert legacy `OwnCloudClient` to `NextcloudClient` at runtime
- see https://github.com/nextcloud/android-library/pull/1271

---


tobiasKaminsky (approved):

tobiasKaminsky:
Will be in https://github.com/nextcloud/android/pull/12266

PR #12519, opened 2024-02-07
Bugfix for unspecific errors when renaming files from different session 

JonasMayerDev:
### Issue
- Create file A
- Rename file A to B from another session, e.g. web browser 
- Without manual reload on the client side, try renaming, share by link

Error for Renaming: "Server not found"
Issue for share by link: "Server not found" + No reload after error → File A gets not updated

### New behavior
Error for Renaming: "File not found"
Behavior for share by link: "File not found" + Reload after error → File A gets updated to file B

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
+++ b/app/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
@@ -523,6 +523,7 @@ public class FileOperationsHelper {
                 showShareFile(file);
             }
         }
+
         fileActivity.refreshList();
     }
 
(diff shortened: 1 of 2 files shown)


ZetaTom (approved):

app/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -523,6 +523,7 @@ public void getFileWithLink(@NonNull OCFile file, final ViewThemeUtils viewTheme
                 showShareFile(file);
             }
         }
+        fileActivity.refreshList();
```
ZetaTom, line 526:
```suggestion

        fileActivity.refreshList();
```

Issue #383, opened 2016-11-14
Ability to pause pending uploads


tobiasKaminsky:
A user takes 100 photos and right after the device connects to wlan, all uploads are started (one by one). The user wants to pause the upload as the (slow, not metered) wlan traffic needs to be used somewhere else.

I would put it in the upload pane in the overflow menu as as checkbox "enabled/disabled".

cc: @AndyScherzinger @eppfel @jancborchardt

vbauchart:
Another use case : I'm temporary located in an area with very limited bandwith, I'd like to pause synchro

jancborchardt:
We could do it like in Netflix on Android and show a menu with:
- Retry (only show when it errored)
- Pause all (or »Pause« when this is the only upload)
- Cancel upload

augustseptember:
What are the recent plans or progresses on this feature request? I am really in the need for pausing the auto uploads. Unfortunately i cannot code it myself. 
Thanks for your work!

tugcanolgun:
I just wanted to bump this issue. Not having a pause button creates a big problem for me. I have a very limited bandwidth and I would like to be able to pause the uploads.

PatriceVigier:
Me too, I just wanted to bump this issue. Not having a pause button creates a big problem for me. I have a very limited bandwidth and I would like to be able to pause the uploads.

andreapx:
It would be a really useful option. Any news about it?
I would donate if there was a bounty

eka249:
Bumping this issue.
We upload very large files, so it's been a struggle to use Nextcloud.

JamesTiberiusKirk:
Bump, especially since half the time upload doesnt work peoperly. Should be able to at least pause it while the connection is flaky. 
When on holidays with bad connection and taking loads of pictures this really is a must.

If y'all open to it I could look into opening a PR in the future as well.

much-doge:
Bumping this issue once more, its a (i think) simple feature yet a very useful one as there are many use cases, e.g., when using phone internet data plan and the data is running out, I want to still be able to use my internet without worrying about Nextcloud uploading all the recent photos I took.

much-doge:
https://github.com/nextcloud/android/issues/6319 Relevant issue

ConnectToTom:
[...]

PR #12527, opened 2024-02-09
Add global upload pause button

JonasMayerDev:
This PR implements a global pause functionality like described here: https://github.com/nextcloud/android/issues/383
Use case: E.g. In bad Wi-Fi and upload should not happen at this moment to prevent overload.

Added Pause button in uploads tab where previously the "Clear failed uploads" option was. The "Clear failed uploads" is now available in a popup menu in the failed uploads section.

The global pause mode stops all current and future uploads until all uploads are resumed manually. To make sure the user does not forget that uploads are paused, a not dismissible notification is shown that tells the user that uploads are paused. 

[Screen_recording_20240212_110157.webm](https://github.com/nextcloud/android/assets/43114340/9ee08942-f634-49ca-bb60-864c19b8b4fa)

[Screen_recording_20240212_111807.webm](https://github.com/nextcloud/android/assets/43114340/7dc74cf4-ee28-4f20-8323-e674d59457f3)

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploaderIntents.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploaderIntents.kt
@@ -97,10 +97,10 @@ class FileUploaderIntents(private val context: Context) {
         )
     }
 
-    fun notificationStartIntent(operation: UploadFileOperation): PendingIntent {
+    fun notificationStartIntent(operation: UploadFileOperation?): PendingIntent {
         val intent = UploadListActivity.createIntent(
-            operation.file,
-            operation.user,
+            operation?.file,
+            operation?.user,
             Intent.FLAG_ACTIVITY_CLEAR_TOP,
             context
         )
--- a/app/src/main/java/com/nextcloud/client/jobs/upload/UploadNotificationManager.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/upload/UploadNotificationManager.kt
@@ -37,16 +37,12 @@ import com.owncloud.android.utils.theme.ViewThemeUtils
 
 class UploadNotificationManager(private val context: Context, private val viewThemeUtils: ViewThemeUtils) {
     companion object {
-
         private const val ID = 411
     }
 
     private var notification: Notification? = null
-    private var notificationBuilder: NotificationCompat.Builder
-    private val notificationManager = context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
-
-    init {
-        notificationBuilder = NotificationUtils.newNotificationBuilder(context, viewThemeUtils).apply {
+    private var notificationBuilder: NotificationCompat.Builder =
+        NotificationUtils.newNotificationBuilder(context, viewThemeUtils).apply {
             setContentTitle(context.getString(R.string.foreground_service_upload))
             setSmallIcon(R.drawable.notification_icon)
             setLargeIcon(BitmapFactory.decodeResource(context.resources, R.drawable.notification_icon))
@@ -55,13 +51,15 @@ class UploadNotificationManager(private val context: Context, private val viewTh
                 setChannelId(NotificationUtils.NOTIFICATION_CHANNEL_UPLOAD)
             }
         }
+    private val notificationManager = context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
 
+    init {
         notification = notificationBuilder.build()
     }
 
     @Suppress("MagicNumber")
     fun prepareForStart(upload: UploadFileOperation, pendingIntent: PendingIntent, startIntent: PendingIntent) {
-        notificationBuilder = NotificationUtils.newNotificationBuilder(context, viewThemeUtils).apply {
+        notificationBuilder.run {
             setSmallIcon(R.drawable.notification_icon)
             setOngoing(true)
             setTicker(context.getString(R.string.foreground_service_upload))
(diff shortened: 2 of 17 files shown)


app/src/main/java/com/nextcloud/client/jobs/upload/FileUploaderIntents.kt
```diff
@@ -97,10 +97,10 @@ class FileUploaderIntents(private val context: Context) {
         )
     }
 
-    fun notificationStartIntent(operation: UploadFileOperation): PendingIntent {
+    fun notificationStartIntent(operation: UploadFileOperation?): PendingIntent {
         val intent = UploadListActivity.createIntent(
-            operation.file,
-            operation.user,
+            operation?.file,
```
alperozturk96, line 102: Why you are passing optinal operation? Do we need this function for pause uploads?

You can check existance of optional operation if its null just return null PendingIntent.

app/src/main/java/com/nextcloud/client/jobs/upload/UploadNotificationManager.kt
```diff
@@ -192,4 +193,23 @@ class UploadNotificationManager(private val context: Context, private val viewTh
     fun dismissWorkerNotifications() {
         notificationManager.cancel(ID)
     }
+
+    fun notifyPaused(startIntent: PendingIntent) {
+        notificationBuilder = NotificationUtils.newNotificationBuilder(context, viewThemeUtils).apply {
```
alperozturk96, line 198: Why you are recreating notificationBuilder? You can simply change current message or dismiss all previous message and show a new message.

app/src/main/java/com/owncloud/android/ui/activity/UploadListActivity.java
```diff
@@ -262,16 +267,33 @@ protected void onPause() {
     }
 
     @Override
+    @SuppressFBWarnings("RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT")
     public boolean onCreateOptionsMenu(Menu menu) {
         MenuInflater inflater = getMenuInflater();
         inflater.inflate(R.menu.activity_upload_list, menu);
 
+        if (menu.getItem(0).getItemId() == R.id.action_toggle_global_pause){
```
alperozturk96, line 275: You can extract this big if block into the another function. Also if user don't have any upload progress no need to show any icon. Only show paused/resume icon if user have ongoing upload process.

No need to repeat  menu.getItem(0).setIcon, menu.getItem(0).setTitle functions. You can create two variable e.g String uploadStateTitle and Int uploadStateIcon then inside if else block assign correct values then only one time call menu.getItem(0)... functions

app/src/main/java/com/owncloud/android/ui/activity/UploadListActivity.java
```diff
@@ -280,17 +302,21 @@ public boolean onOptionsItemSelected(MenuItem item) {
             } else {
                 openDrawer();
             }
-        } else if (itemId == R.id.action_clear_failed_uploads) {
-            for (OCUpload upload : uploadsStorageManager.getFailedButNotDelayedUploadsForCurrentAccount()){
-                uploadListAdapter.cancelOldErrorNotification(upload);
+        } else if (itemId == R.id.action_toggle_global_pause) {
+            preferences.setGlobalUploadPaused(!preferences.getGlobalUploadPaused());
```
alperozturk96, line 306: Could you move that else if block into the another function?

app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
```diff
@@ -122,29 +122,56 @@ public void onBindHeaderViewHolder(SectionedViewHolder holder, int section, bool
...
-                    powerManagementService)).start();
-                default -> {
+                case FINISHED -> {
+                    uploadsStorageManager.clearSuccessfulUploads();
+                    loadUploadItemsFromDb();
                 }
-                // do nothing
-            }
+                case FAILED -> {
+                    popup = new PopupMenu(MainApp.getAppContext(), headerViewHolder.binding.uploadListAction);
```
alperozturk96, line 149: Same here you can extract PopupMenu into the another function.

alperozturk96 (changes requested):

app/src/main/java/com/nextcloud/client/jobs/upload/UploadNotificationManager.kt
JonasMayerDev: You are right. I now also reuse notificationBuilder in prepareForStart, please check if my changes are ok!

app/src/main/java/com/owncloud/android/ui/activity/UploadListActivity.java
JonasMayerDev: I think it makes sense to show the pause button even if not upload is in progress since the idea is also to prevent future uploads e.g. trough auto upload when for example in bad Wi-Fi. So it is important that the user can at any time deactivate the upload, even for future uploads.

Thanks for the improvement ideas!

alperozturk96 (approved):
Refactored, tested working on Android 14 Emulator

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/ui/activity/UploadListActivity.java
```diff
@@ -280,17 +318,11 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
-            uploadsStorageManager.clearFailedButNotDelayedUploads();
-            uploadListAdapter.loadUploadItemsFromDb();
-        } else {
-            retval = super.onOptionsItemSelected(item);
+        } else if (itemId == R.id.action_toggle_global_pause) {
+            toggleGlobalPause(item);
         }
 
-        return retval;
+        return true;
```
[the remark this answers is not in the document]
JonasMayerDev: Good catch, thank you!

alperozturk96 (approved):

PR #12540, opened 2024-02-12
BugFix Check emptiness of previousSortGroupState

alperozturk96:

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -2486,6 +2486,10 @@ public class FileDisplayActivity extends FileActivity
      * visibility earlier using {@link #setSortListGroup(boolean, boolean)}
      */
     private void popSortListGroupVisibility() {
+        if (previousSortGroupState.isEmpty()) {
+            return;
+        }
+
         boolean popped = previousSortGroupState.pop();
         showSortListGroup(popped);
     }
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -2487,6 +2487,7 @@ public class FileDisplayActivity extends FileActivity
      */
     private void popSortListGroupVisibility() {
         if (previousSortGroupState.isEmpty()) {
+            showSortListGroup(false);
             return;
         }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -2485,6 +2485,10 @@ private void setSortListGroup(boolean currentListGroupVisibility, boolean show)
      * visibility earlier using {@link #setSortListGroup(boolean, boolean)}
      */
     private void popSortListGroupVisibility() {
+        if (previousSortGroupState.isEmpty()) {
+            return;
+        }
+
         boolean popped = previousSortGroupState.pop();
         showSortListGroup(popped);
```
ZetaTom, line 2493:
```suggestion
        if (!previousSortGroupState.isEmpty()) {
            boolean popped = previousSortGroupState.pop();
            showSortListGroup(popped);
        }
```

In my opinion this version is a touch cleaner.

alperozturk96: IMO avoid [nesting code](https://stackoverflow.com/a/268187/16787943) better.

app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -2486,7 +2486,12 @@ private void setSortListGroup(boolean currentListGroupVisibility, boolean show)
      * visibility earlier using {@link #setSortListGroup(boolean, boolean)}
      */
     private void popSortListGroupVisibility() {
-        boolean popped = previousSortGroupState.pop();
-        showSortListGroup(popped);
+        showSortListGroup(false);
```
ZetaTom, line 2489: This would negate purpose of `previousSortGroupState`, which was introduced by #11965. If this choice was deliberate, check that all edge cases are covered and remove this variable entirely.

alperozturk96: Current implementation on master is broken. We should't see SortListGroup in file detail or file sharing tab also we have crash right now. Crash has been fixed and appearance of SortListGroup working as expected (visible in file list not visible in file details, sharing). 

This PR aims to fix crash and appearance of not refactor.

JonasMayerDev: I also agree with Tom, but we should create a new Issue about it and for now merge to fix the crash.

JonasMayerDev (approved):
Quick test, fixes crash

app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
alperozturk96: @JonasMayerDev I agree

PR #12546, opened 2024-02-13
Bugfix Check ClassCastException edge, Nullability of Bundle or Intent

alperozturk96:
In some cases we are getting class cast exception, this PR fixes that also cover nullable Intent and Bundle cases. Unit tests written for both extensions.
 

AndyScherzinger (approved):

1div0:
Vielen Dank.

AndyScherzinger:
Gern geschehen 👍

Issue #1256, opened 2017-08-01
DEV 30/07. Autoupload enhancement. Cannot pause or repeat autouploads.


ruvashev:
Current issues:
1. Cannot pause uploads.
2. Cancelled uploads via "X" button cannot be repeated and photos with videos have to be manually uploaded.

Can we please implement such enhancements? Thanks.

tobiasKaminsky:
Pause is a nice idea, but don't know if this is possible with the newest implementation @mario?
For Cancel: we could add a new section "cancelled", but I am unsure if this is used often...

mario:
@tobiasKaminsky Well it's probably "Cancel" without removing it from the DB?

mario:
Otherwise, the upload operation is the same as before ... so if you could have pause before, you can have it now too :)

tobiasKaminsky:
I understand pause as "pause the complete autoupload for some time", but maybe I am wrong.
Pausing an individual upload is currently also not possible...

mario:
Well, I'd guess it's more about individual one (or pausing all uploads, which would pause all CURRENTLY stored pending auto uploads too).

ruvashev:
Either "Pause" implementation (all together or individual) is good to have in case bandwidth is needed right away for something else. Use Case example: say one travel, took many photos, checked in hotel, got wifi and now upload starts. 

As for the "Cancel" - no preferences, but i guess having cancelled section would help.

Thanks for considerations, guys.

dusscloud:
> 
> 
> Either "Pause" implementation (all together or individual) is good to have in case bandwidth is needed right away for something else. Use Case example: say one travel, took many photos, checked in hotel, got wifi and now upload starts.
> 

I had exactly this issue : in some place/city/country, the wifi signal may be good enough for sending email/message but not for picture transfer, especially when there is a lot taken.
In that case, I would like to pause the autoupload, and resume it whenever I feel the Wifi bandwith is better

joshtrichards:
Pausing is covered in #383. Let's move that discussion there (where there already is an active one). 

I'll keep this issue open for the second suggestion for the time being. Please only submit one bug/enhancement per Issue in the future (it's impossible to coordinate work when an issue covers multiple work items that may not occur at the same time or even be handled by the same person).

PR #12549, opened 2024-02-14
Add resume option for cancelled uploads

JonasMayerDev:
The idea here is to be able to pause and resume uploads by not directly deleting them when a user manually cancels an upload as mentioned in https://github.com/nextcloud/android/issues/1256 for example.
This implements the individual pause mechanic as described in https://github.com/nextcloud/android/issues/383 because a large file (bigger than 10 MB) will remember its progress and behave like paused.

The dedicated cancelled uploads section has 2 options: "Remove all cancelled uploads" and "Resume all cancelled uploads"
The user can also resume individual uploads by clicking on them.

See this video for a demonstration:
[Screen_recording_20240214_043340.webm](https://github.com/nextcloud/android/assets/43114340/f5a2ef87-aaa3-4dae-b276-840f7b5aba53)

  

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
@@ -175,7 +175,7 @@ public class UploadListAdapter extends SectionedRecyclerViewAdapter<SectionedVie
         failedPopup.show();
     }
 
-    public void showCancelledPopupMenu(HeaderViewHolder headerViewHolder) {
+    private void showCancelledPopupMenu(HeaderViewHolder headerViewHolder) {
         PopupMenu popup = new PopupMenu(MainApp.getAppContext(), headerViewHolder.binding.uploadListAction);
         popup.inflate(R.menu.upload_list_cancelled_options);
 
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
```diff
@@ -173,6 +179,34 @@ private void createFailedPopupMenu(HeaderViewHolder headerViewHolder) {
         failedPopup.show();
     }
 
+    public void createCancelledActionsPopupMenu(HeaderViewHolder headerViewHolder) {
```
alperozturk96, line 182: maybe we can rename this function to "showCancelledUploadsPopupMenu". Because its not just creating, its creating and showing it.

app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
```diff
@@ -371,7 +416,9 @@ public void onBindViewHolder(SectionedViewHolder holder, int section, int relati
         itemViewHolder.binding.thumbnail.setImageResource(R.drawable.file);
 
         // click on item
-        if (item.getUploadStatus() == UploadStatus.UPLOAD_FAILED) {
+        if (item.getUploadStatus() == UploadStatus.UPLOAD_FAILED ||
+            item.getUploadStatus() == UploadStatus.UPLOAD_MANUALLY_CANCELLED) {
```
alperozturk96, line 420: UPLOAD_MANUALLY_CANCELLED -> UPLOAD_CANCELLED

alperozturk96 (approved):

PR #12577, opened 2024-02-19
Fix Path Traversal To Internal File ExFiltration

alperozturk96:

How come old implementation, check sensitivity of uri? @tobiasKaminsky  As far as I understood from code we are only accepting files from other apps.

**Test After Change**

https://github.com/nextcloud/android/assets/67455295/0baec58f-f51f-4c20-b514-9da3189d6851

### DIFF
--- a/app/src/androidTest/java/com/owncloud/android/ui/helpers/UriUploaderIT.kt
+++ b/app/src/androidTest/java/com/owncloud/android/ui/helpers/UriUploaderIT.kt
@@ -5,10 +5,13 @@ import androidx.test.core.app.launchActivity
 import com.nextcloud.client.jobs.upload.FileUploadWorker
 import com.nextcloud.test.TestActivity
 import com.owncloud.android.AbstractIT
+import com.owncloud.android.lib.common.utils.Log_OC
 import org.junit.Assert
 import org.junit.Test
 
 class UriUploaderIT : AbstractIT() {
+
+    private val tag = "UriUploaderIT"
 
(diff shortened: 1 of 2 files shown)


ZetaTom (commented):
During my testing, I've noticed that this PR causes some behavioural changes. These changes make the app behave more consistently than before, but they also break select use cases.

Consider the following example: *Copy-by-Share* / *Share-to-Copy*
Some users tend to copy files by first selecting them in the Nextcloud app, and then sharing them to Nextcloud itself again via (sharing icon → Nextcloud).
As a user I'd expect this to just work and the file(s) to be copied to the selected destination.

However, the current implementation (on master), only allows this under certain conditions. One such satisfactory case is met when the user tries to share *images*, (`.png`, `.jpg`, etc.) because they are handled by our image cache (the URI will differ to one from a non-image file). In this case *Copy-by-Share* will work but it will fail for other, arbitrary file types (`.pdf`, `.md`, etc.).

Another thing to consider is sharing via the native Android/Google *Files* app. When sharing a file contained in Nextcloud from the Files app to the actual Nextcloud app, I would again expect the file to just be copied over, which is exactly what happens on master. When applying the changes from this PR this use case is no longer available because the app will reject the URI.

These are some of the design considerations we have to account for when implementing such changes.

@nextcloud/android

tobiasKaminsky (approved):

tobiasKaminsky:
Overlooked comments by Tom.

PR #12583, opened 2024-02-21
Fix Crash When Write Activities From Random Any File

alperozturk96:

Crash was happening due to asynchronous code execution. In some rare scenarios ownCloud client was null. In this PR I check thread completion and ownCloud client nullability.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/fragment/FileDetailActivitiesFragment.java
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/FileDetailActivitiesFragment.java
@@ -341,6 +341,8 @@ public class FileDetailActivitiesFragment extends Fragment implements
                             populateList(activitiesAndVersions, lastGiven == -1);
                         }
                     });
+
+                    isDataFetched = true;
                 } else {
                     Log_OC.d(TAG, result.getLogMessage());
                     // show error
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/ui/fragment/FileDetailActivitiesFragment.java
```diff
@@ -363,6 +364,14 @@ private void fetchAndSetData(int lastGiven) {
         });
 
         t.start();
+
+        try {
+            t.join();
+            isDataFetched = true;
```
tobiasKaminsky, line 370: Shouldn't this be set to true when thread is done?

app/src/main/java/com/owncloud/android/ui/fragment/FileDetailActivitiesFragment.java
```diff
@@ -363,6 +364,14 @@ private void fetchAndSetData(int lastGiven) {
         });
 
         t.start();
+
+        try {
+            t.join();
```
ZetaTom, line 369: As far as I'm aware, `t.join()` will block execution of the current thread, until `t` has finished executing. If this is the case, it would defeat the purpose of threading. Since `t.join()` is called immediately after `t.start()`, the current thread should just hang until `t` has done it's thing.

ZetaTom (approved):
In my tests, everything worked as expected.

If the server is unavailable and one tries to comment on a file/directory, it doesn't work, which is expected. But the button still shows the ripple effect and the user is left without any feedback. I think we should address the lack of feedback at a later point in time.

PR #12591, opened 2024-02-22
Fix MaterialToolbar Crash

alperozturk96:

**How to Test (Updated)**

Since normal login not available for outdated WebView. We can use qr code. Simply we can put one line of code to the AuthenticatorActivity.startQRScan() and disable previous logic.

e.g

```
private void startQRScanner() {
    parseAndLoginFromWebView("nc://login/user:admin&password:admin&server:10.0.2.2:55000");

    // Intent intent = new Intent(this, QrCodeActivity.class);
    // startActivityForResult(intent, REQUEST_CODE_QR_SCAN);
}
```

com.google.android.material.appbar.MaterialToolbar crashes on Android 8 due to misusage of main app theme

**Reproducing the Issue:**

It's difficult to test the app on an emulator because login functionality doesn't work on older Android versions 7 and 8. This is due to our login process relying on a WebView, and older Android versions use WebView versions like 55.x.x.x, which causes compatibility issues.

If you have a real device running Android 7 or 8, you can reproduce the crash because WebView updates automatically via the Play Store.

**Why WebView Can't be Updated on Emulators?**

For Mac OS Apple Silicon users, there's no option to install an Android Emulator with Google Play services for Android versions 7 and 8; only AOSP image versions are available. Consequently, you cannot update WebView with AOSP 7 and 8.

There's a workaround: making all AVD instances writable, then using adb commands to push a new WebView version. You can find a guide [here](https://gist.github.com/ppoffice/9ce9790708eeabbec1281467e25139e4).

Alternative emulators on MacOS still have issues because they don't provide Android 7 and 8 images with Google Services.

For Windows users, testing is also challenging because [Intel HAXM](https://github.com/intel/haxm), which is deprecated, is required for older OS versions. While you can still install Intel HAXM, it doesn't work smoothly on all systems and can be error-prone. Virtualization for the CPU must be enabled, and Windows Hyper-V services must be functioning properly to run Android 7 and 8 Google Play images.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/preview/PreviewMediaActivity.kt
+++ b/app/src/main/java/com/owncloud/android/ui/preview/PreviewMediaActivity.kt
@@ -35,6 +35,7 @@ import android.graphics.drawable.Drawable
 import android.media.MediaMetadataRetriever
 import android.net.Uri
 import android.os.AsyncTask
+import android.os.Build
 import android.os.Bundle
 import android.os.Handler
 import android.os.Looper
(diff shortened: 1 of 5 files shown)


JonasMayerDev (approved):
Fixes the crash in quick test

ZetaTom (approved):
Everything worked as expected during my tests. I did not see any visual differences between master and this branch.

PR #12612, opened 2024-02-26
Assistant Feature

alperozturk96:


app/src/main/res/values-sk-rSK/strings.xml
```diff
@@ -789,7 +789,7 @@
...
... môžete mať kompletne pod kontrolou.\n\nFunkcie:\n * Jednoducho použiteľné moderné rozhranie, hodiace sa k vzhľadu vášho servera* \n Nahrávanie súborov na Nextcloud server\n * Ich sprístupnenie s inými ľudmi\n * Synchronizácia vašich obľúbených súborov a adresárov\n * Vyhľadávanie naprieč všetkými adresármi na serveri\n * Automatické nahrávanie fotiek a videí nasnímaných vašim zariadením\n * Doručovanie notifikácií\n * Podpora viac účtov naraz\n * Zabezpečený prístup k vašim dátam pomocou odtlačku prstu alebo kódom PIN\n * Začlenenie DAVx5 (predtým známy ako DAVdroid) pre jednoduchý prístup ku kalendáru &amp; synchronizácii kontaktov\n\n Akékoľvek problémy prosím hláste na https://github.com/nextcloud/android/issues a o aplikácii môžete diskutovať na https://help.nextcloud.com/c/clients/android\n\n Nepoznáte ešte Nextcloud? Nextcloud je server pre súkromnú synchronizáciu súborov &amp;, zdieľanie a komunikáciu. Je to slobodný softvér a môžete si ho prevádzkovať buď sami alebo si ho prenajímať od nejakej spoločnosti. Týmto spôsobom získate plnú vládu nad svojimi fotkami, údajom v kalendári a kontaktoch, dokumentom a všetkým ostatným.\n\n Viac zistíte na https://nextcloud.com</string>
```
alperozturk96, line 792: & sign causes build error, it needs to be removed from translation

alperozturk96:
@tobiasKaminsky Screenshot tests are irrelevant. We can merge this because other PRs also failing because of the ss tests

tobiasKaminsky:
@alperozturk96 you may never change an existing DB version, in this case 78.
This would break auto migration.
Please thus revert 78 and check that an update from 3.28.0 to this branch works.

app/src/main/res/values-sk-rSK/strings.xml
ZetaTom: Please revert this commit. The issue has since been fixed on Transifex and has been incorporated into master.

alperozturk96:
> @alperozturk96 you may never change an existing DB version, in this case 78. This would break auto migration. Please thus revert 78 and check that an update from 3.28.0 to this branch works.

When I install master app then install this PR I am getting crash due changes in DB. I think we can just bump the version.

_**"java.lang.RuntimeException: Unable to create application com.owncloud.android.MainApp: java.lang.IllegalStateException: Room cannot verify the data integrity. Looks like you've changed schema but forgot to update the version number. You can simply fix this by increasing the version number. Expected identity hash: ec997f271f9045e8483b260f036a168f, found: f26afed3b9b87a3acb578947a26223ac"**_

tobiasKaminsky:
For "all task" this is not true, as there is no button.
Can you rephrase it for this one?

tobiasKaminsky:
According to https://github.com/nextcloud/files-clients/issues/50#issuecomment-1909730092 please

- add NC Assistant in header as first item before Talk "circle"
- move Assistant below notification, show it only if is_branded_client is true (bool in setup.xml)

tobiasKaminsky:
Another small thing:
Can the text input field be at least two lines or even auto growing?
Try to enter a long text. As we have enough space it should not truncate first part.

tobiasKaminsky (approved):

PR #12619, opened 2024-02-29
Fix back button in trash bin

Unpublished:
Pressing back would never actually go back but crashed with a stack exception (`navigateUp` calling itself recursively via `TrashBinActivity.close()` infinitely).

The `OnBackPressedCallback` needs to be disabled, on root path, to allow the back press to be propagated to the activity.


alperozturk96 (approved):
tested working

JonasMayerDev (commented):
PR works for me, but I have no issue on master, can someone describe how to reproduce the issue?

JonasMayerDev:
Yes, you are right, I used the hamburger menu at the top. Thanks for clarification!

PR #12621, opened 2024-02-29
Fix crash when restoring contacts

ZetaTom:
Due to a small error in #12087, `BackupListFragment.newInstance()` would always be called with an array of nulls instead of the expected backup file(s). This would cause the app to crash when attempting to restore any backup.

---


alperozturk96 (approved):
Tested working

PR #12651, opened 2024-03-06
Show Cached Information Instead of Not Showing Quota Information at All When User Not Authorized

alperozturk96:

When a user adds an additional account, then deletes that same account, and subsequently login again with the same account, the quota information cannot be retrieved from the backend due to a 401 unauthorized error. We can display the cached value and update it when the user reopens the app to reflect the new value.


alperozturk96:
Actual fix and PR will be opened

PR #12668, opened 2024-03-12
Bugfix E2E folder file creation

alperozturk96:

### How To Test

1. Create encrypted folder
2. Create another folder inside encrypted folder
3. Upload file inside encrypted folder and sub folder

All uploads must be completed without any issue

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
@@ -508,7 +508,7 @@ public class UploadFileOperation extends SyncOperation {
                 metadata.getMetadata().setMetadataKey(encryptedMetadataKey);
 
                 if (object instanceof DecryptedFolderMetadataFileV1) {
-                    metadata.getMetadata().setChecksum(((DecryptedFolderMetadataFileV1) object).getMetadata().getChecksum());
+                    metadata = (DecryptedFolderMetadataFileV1) object;
                 }
 
                 object = metadata;
@@ -696,6 +696,7 @@ public class UploadFileOperation extends SyncOperation {
                         serializedFolderMetadata = EncryptionUtils.serializeJSON(encryptedFolderMetadata);
                     }
 
+                    // FIXME after first failed upload folder stay locked, uploadMetadata throws UploadException for encrypted folder
                     // upload metadata
                     EncryptionUtils.uploadMetadata(parentFile,
                                                    serializedFolderMetadata,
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -507,8 +509,11 @@ private RemoteOperationResult encryptedUpload(OwnCloudClient client, OCFile pare
                 String encryptedMetadataKey = EncryptionUtils.encryptStringAsymmetric(metadataKey, publicKey);
                 metadata.getMetadata().setMetadataKey(encryptedMetadataKey);
 
+                if (object instanceof DecryptedFolderMetadataFileV1) {
+                    metadata = (DecryptedFolderMetadataFileV1) object;
+                }
+
                 object = metadata;
-                metadataExists = false;
             }
```
ZetaTom, line 517: I'm a bit confused. This expression should be equivalent to the following, shouldn't it?

```java
if (!(object instanceof DecryptedFolderMetadataFileV1)) {
    object = metadata;
}
```

Another thing of note is that if these expressions are indeed equivalent to one another, wouldn't `metadata` be dismissed if `object` is set? If this is the case, the condition should be moved up to line 503, since the `metadata` object wouldn't be used after all.

app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -489,13 +490,14 @@ private RemoteOperationResult encryptedUpload(OwnCloudClient client, OCFile pare
 //                                                   mContext);
 
             Object object = EncryptionUtils.downloadFolderMetadata(parentFile, client, mContext, user);
+            if (object instanceof DecryptedFolderMetadataFileV1 && ((DecryptedFolderMetadataFileV1) object).getMetadata() != null) {
```
ZetaTom, line 493:
```suggestion
            if (object instanceof DecryptedFolderMetadataFileV1 decrypted && decrypted.getMetadata() != null) {
```

ZetaTom (approved):
Other than the minor nitpicks, this worked perfectly in my tests. Well done! :tada:

app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
alperozturk96: Yes I agree. But I don't want to touch it more code is extremely error prone. I added TODO Refactor. I will create another PR and refactor that huge function. Thats why I left it that way

JonasMayerDev (approved):
In a quick test, this solves the issue of subfolders having wrong names and then files not being able to upload there!

PR #12701, opened 2024-03-18
Fix Screenshot Tests

alperozturk96:


app/src/main/java/com/owncloud/android/ui/dialog/SslUntrustedCertDialog.java
```diff
@@ -76,9 +76,7 @@ public static SslUntrustedCertDialog newInstanceForEmptySslError(SslError error,
         if (error == null) {
             throw new IllegalArgumentException("Trying to create instance with parameter error == null");
         }
-        if (handler == null) {
```
alperozturk96, line 79: null control already exists for usage of handler. For screenshot tests no need to pass SslErrorHandler. Mockito throws error for it.

app/src/androidTest/java/com/owncloud/android/ui/trashbin/TrashbinActivityIT.kt
```diff
@@ -60,7 +61,7 @@ class TrashbinActivityIT : AbstractIT() {
             sut.trashbinPresenter = TrashbinPresenter(trashbinRepository, sut)
             sut.runOnUiThread { sut.loadFolder() }
             shortSleep()
-            screenshot(sut)
+            screenshot(sut, UUID.randomUUID().toString())
```
tobiasKaminsky, line 64: I fear a random name will not work, as then the recorded screenshot has always a different name than the original one.
What about giving them a fixed name: "TrashbinActivityIT_normalUser"?

This is what usually is done automatically, but I assume due to usage of scenario this does not work.

AndyScherzinger:
❗ REBASED ❗

alperozturk96:
Out of date due to conflicts

Issue #12596, opened 2024-02-24
Automatic Upload is prepared


LunaticTazz:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

Set up Automatic Upload for in total four folders.
From time to time a notification is present on Android, that the automatic upload is prepared (German version "Automatisches Hochladen wird vorbereitet"). A progress bar is at min, the message disappears after a while and is retriggered from time to time. According to Android it pops up about 80 times a day.
Mobile background data is allowed, but uploads are only allowed on a WiFi network.
In addition to this, mobile data is consumed, but no files seem to be uploading. There's no files in the upload queue, there is nothing arriving on the server.
It is unknown to me, which files caused this behavior, but it's also present with only the photo folder being on automated upload (the other three have been deactivated).

### Expected behaviour

Wait for WiFi to upload the files and be quiet if there is nothing to upload. No usage of mobile data.

### Actual behaviour

Constant notifications, decreased battery life and mobile data consumption.

### Android version

Android 11

### Device brand and model

moto G 5G

### Stock or custom OS?

Stock

### Nextcloud android app version

3.28.0

### Nextcloud server version

26.0.10

### Using a reverse proxy?

No

davidcrds:
I have the same issue on Android 14 (S22+). Auto Upload still works fine, but it keeps trying to do something which drains battery. 

Running Nextcloud AIO currently on Version 27.1.5 on the Server.

kadrim:
I have the same issue. It appeared a few days ago, most probably due to an update of the android app

- samsung note 9
- android 10
- nextcloud server 28.0.2

steel-moose:
Same issue here.

- android 9
- nokia 8 (HMD TA-1004)
- version 3.28.0
- nextcloud snap 27.1.6

Seems to be draining the battery much faster.

regs01:
Also stuck for over almost a day now.
<img width="712" alt="firefox_2024-02-25_19-36-18" src="https://github.com/nextcloud/android/assets/10468217/eb7a9cce-cbe1-462a-9fc3-9bc7b2cd5d69">

[...]

PR #12719, opened 2024-03-21
Try to only process changed files for auto upload

JonasMayerDev:
Initiated by https://github.com/nextcloud/android/issues/12596 the idea was to reduce the times the full file system scan for changed files for auto upload is needed. This PR adds that when ever ContentObserverWorker detects media files (so only images and videos) change, the FilesSyncWorker does not go through every file in the corresponding folders. Now it only checks which sync folder (if any) the file belongs to and only enqueues this file for upload.   

## Test this PR
To test this functionality, one can set up the camera folder as an auto upload folder and take pictures. The upload of the file should start after about at most 15s automatically in the background. 

This behavior is the same as before, but it now should take less time for especially large media folders, and it does not use a foreground worker for the sync worker. 

## Known Issues
Sometimes some changes are not detected when too many changes happen one after another and the upload of those files stops until the app is opened. So sometime files are missing from upload. Though this is not optimal, it seems like this is an issue with the content observer worker. As we not really rely on this functionality (a full file system scan is still scheduled every 15 min) I think this is still better than the functionality before. So this functionality should rather be seen as a bonus that uploads files faster if it works but if not there is still a fallback.    

### DIFF
--- a/app/src/androidTest/java/com/nextcloud/client/jobs/BackgroundJobManagerTest.kt
+++ b/app/src/androidTest/java/com/nextcloud/client/jobs/BackgroundJobManagerTest.kt
@@ -214,7 +214,7 @@ class BackgroundJobManagerTest {
         fun job_is_unique_and_replaces_previous_job() {
             verify(workManager).enqueueUniqueWork(
                 eq(BackgroundJobManagerImpl.JOB_CONTENT_OBSERVER),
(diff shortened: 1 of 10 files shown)


alperozturk96 (approved):
Tested on Android 14 and Android 7.1.1 working.

AndyScherzinger (approved):

PR #12736, opened 2024-03-25
Fix Upload Problems For Encrypted Folder

alperozturk96:


JonasMayerDev (approved):
After quick test, upload still has bugs (on master and in this PR) (will report as new issues) but definitely works better than before.

alperozturk96:
Duplication [of](https://github.com/nextcloud/android/pull/12796)

PR #12752, opened 2024-03-27
Bugfix editing folder description 

JonasMayerDev:
Before editing of folder description using readme.md files would lead to error message.
With this PR editing works.

[Screen_recording_20240327_190105.webm](https://github.com/nextcloud/android/assets/43114340/a7bb1435-e9ec-4bf3-b016-e9f7b2c29b5d)

Known Issue:
Folder README will only be updated when manually refreshing. 

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/preview/PreviewTextStringFragment.java
+++ b/app/src/main/java/com/owncloud/android/ui/preview/PreviewTextStringFragment.java
@@ -52,6 +52,8 @@ public class PreviewTextStringFragment extends PreviewTextFragment {
     @Inject UserAccountManager accountManager;
     @Inject ViewThemeUtils viewThemeUtils;
 
+    private boolean isEditorWebviewLaunched = false;
+
     /**
      * Creates an empty fragment for previews.
      */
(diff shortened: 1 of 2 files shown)


alperozturk96 (changes requested):
You can use LocalBroadcastManager to trigger refresh automatically.

GalleryFragment.#L113 -> sample usage

alperozturk96 (approved):
tested working

JonasMayerDev:
Approved by Alper

app/src/main/java/com/owncloud/android/ui/preview/PreviewTextStringFragment.java
```diff
@@ -107,6 +109,16 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
         return view;
     }
 
+    @Override
+    public void onStart() {
+        if (isEditorWebviewLaunched && containerActivity instanceof FileDisplayActivity fileDisplayActivity) {
+            fileDisplayActivity.getSupportFragmentManager().popBackStack();
```
ZetaTom, line 115: This line causes an issue when the editor is opened from an active search. I've been able to reproduce the issue using the following steps:

1. Open any folder (excl. root directory) containing a `README.md`
2. Select preview at top of folder
3. Open text search using icon on top right
4. Enter some text
5. Enter edit mode using FAB
6. Exit back out of editor

Now one is unable to return to the normal files view.

Given all of this, I'm unsure how severe this issue really is.

ZetaTom (commented):
This changes fixes the issue at hand but also introduces a new undesirable behaviour where the user is unable to return to the normal file view.

app/src/main/java/com/owncloud/android/ui/preview/PreviewTextStringFragment.java
JonasMayerDev: Yes, good catch. I created an Issue https://github.com/nextcloud/android/issues/12805. I work on it when I have time, but right now I would not prioritize it...

PR #12753, opened 2024-03-27
Bugfix change shares database USER_ID type to String

JonasMayerDev:
The database type for the owners userid for shares was int instead of String (which was assumed everywhere, like "getString" etc.). It seemd to have worked well for short userids but for large userids the data will not be correctly saved. Ids like "12320078498127907419284723" would be converted to "1.23200e24" which leads to following misbehavior for example:

[Screen_recording_20240327_220052.webm](https://github.com/nextcloud/android/assets/43114340/6e03843e-70b1-465c-b5c1-87d8efc29bc3)

Now database type is String. To test this fix / to experience the issue, you need to be logged in as a user with a really long userid (=username / loginname) and share a file or folder with another person on the instance.  
 
[Screen_recording_20240327_220139.webm](https://github.com/nextcloud/android/assets/43114340/4bec0b86-fe57-4524-bfb1-ddd9bb220a76)

### DIFF
--- a/app/schemas/com.nextcloud.client.database.NextcloudDatabase/80.json
+++ b/app/schemas/com.nextcloud.client.database.NextcloudDatabase/80.json
@@ -1,0 +1,1203 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 80,
+    "identityHash": "dda37e2f97cecb10da04723f5c472d00",
+    "entities": [
+      {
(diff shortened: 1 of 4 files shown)


ZetaTom (approved):
Works and fixes the issue.

alperozturk96 (approved):
Tested on Android 14 emulator. Didn't crash.

PR #12779, opened 2024-04-01
Fix upload download problems

alperozturk96:

**This PR Fixes Following Problems**

+ Sync icon visibility during download
+ Accessing after download attempt to etag throws null point exception
+ Remove file dialog actions order is wrong for downloaded and normal file
+ File download

**Note**

The Spotbugs count increased due to the reintroduction of old code (with improvements) in this PR.

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/OCFile.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/OCFile.java
@@ -361,15 +361,6 @@ public class OCFile implements Parcelable, Comparable<OCFile>, ServerFileInterfa
         return false;
     }
 
-    public String getFileNameWithoutExtension(String fileName) {
-        int dotIndex = fileName.lastIndexOf('.');
-        if (dotIndex > 0) {
-            return fileName.substring(0, dotIndex);
-        } else {
-            return fileName;
-        }
-    }
-
     /**
      * The path, where the file is stored locally
      *
--- a/app/src/main/java/com/owncloud/android/operations/DownloadFileOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/DownloadFileOperation.java
@@ -273,7 +273,7 @@ public class DownloadFileOperation extends RemoteOperation {
                 }
             }
 
-            if (downloadType == DownloadType.DOWNLOAD) {
+            if (downloadType == DownloadType.DOWNLOAD && !file.isEncrypted()) {
                 moved = tmpFile.renameTo(newFile);
                 newFile.setLastModified(file.getModificationTimestamp());
                 if (!moved) {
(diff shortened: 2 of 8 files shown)


app/src/main/java/com/owncloud/android/datamodel/OCFile.java
```diff
@@ -361,15 +361,6 @@ public boolean existsOnDevice() {
         return false;
     }
 
-    public String getFileNameWithoutExtension(String fileName) {
```
alperozturk96, line 364: Unused function

app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -726,7 +726,7 @@ private RemoteOperationResult encryptedUpload(OwnCloudClient client, OCFile pare
                                                                  metadata,
                                                                  token,
                                                                  client,
-                                                                 metadataExists,
```
alperozturk96, line 729: Metadata is already available for version 2, so there is no need to check the metadataExists value for version 2 because this value is only verified for version 1.

app/src/main/java/com/owncloud/android/operations/DownloadFileOperation.java
```diff
@@ -273,7 +273,13 @@ protected RemoteOperationResult run(OwnCloudClient client) {
                 }
             }
 
-            if (downloadType == DownloadType.EXPORT) {
+            if (downloadType == DownloadType.DOWNLOAD) {
```
alperozturk96, line 276: Code that was removed during end-to-end (e2e) fixes previously caused download issues for e2e. However, after recent improvements, it no longer affects e2e and doesn't disrupt normal file downloads either. Special thanks to @Unpublished for this.

app/src/main/java/com/owncloud/android/operations/DownloadFileOperation.java
```diff
@@ -273,7 +273,13 @@ protected RemoteOperationResult run(OwnCloudClient client) {
                 }
             }
 
-            if (downloadType == DownloadType.EXPORT) {
+            if (downloadType == DownloadType.DOWNLOAD && !file.isEncrypted()) {
```
alperozturk96, line 276: Code that was removed during end-to-end (e2e) fixes previously caused download issues for e2e. However, after recent improvements, it no longer affects e2e and doesn't disrupt normal file downloads either. Special thanks to @Unpublished for this. 

I just added isEncrypted check because if we don't ignore this if block for e2e file try to be opened as encrypted, therefore user will not able to open file.

I tested for v1 and v2 encryption, it's working.

alperozturk96:
Duplication [of](https://github.com/nextcloud/android/pull/12794)

PR #12790, opened 2024-04-02
Add SPDX header

AndyScherzinger:


AndyScherzinger:
@David-Development the 3 SSO related classes where flagged with GPL3 while they should be AGPL3. Are you fine with changing the license to AGPL3? Thanks for your feedback 🙏

David-Development:
@AndyScherzinger yes, changing it to AGPL3 sounds good to me! 🙂 Thank you!

PR #12793, opened 2024-04-02
Fix Crash Show Fragment During FileActivity Lifecycle Change

JonasMayerDev:
**By Alper**

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileActivity.java
@@ -520,17 +520,20 @@ public abstract class FileActivity extends DrawerActivity
      * Show loading dialog
      */
     public void showLoadingDialog(String message) {
-        // grant that only one waiting dialog is shown
-        dismissLoadingDialog();
-        // Construct dialog
-        Fragment frag = getSupportFragmentManager().findFragmentByTag(DIALOG_WAIT_TAG);
-        if (frag == null) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/ui/activity/FileActivity.java
```diff
@@ -532,17 +532,20 @@ protected void updateFileFromDB(){
...
+            dismissLoadingDialog();
+            // Construct dialog
+            Fragment frag = getSupportFragmentManager().findFragmentByTag(DIALOG_WAIT_TAG);
+            if (frag == null) {
+                Log_OC.d(TAG, "show loading dialog");
+                LoadingDialog loading = LoadingDialog.newInstance(message);
+                FragmentManager fm = getSupportFragmentManager();
+                FragmentTransaction ft = fm.beginTransaction();
+                loading.show(ft, DIALOG_WAIT_TAG);
+                fm.executePendingTransactions();
```
ZetaTom, line 547:
```suggestion
            ft.add(loading, DIALOG_WAIT_TAG);
            ft.commitAllowingStateLoss();
```

I think that we should use `ft.commitAllowingStateLoss()`, as we're already using `loading.dismissAllowingStateLoss()` in `dismissLoadingDialog()`. Then we can probably remove the if statement checking `mResumed`.

alperozturk96:
Duplication of https://github.com/nextcloud/android/pull/13037

PR #12794, opened 2024-04-02
Fix upload download problems

JonasMayerDev:
**By Alper**
This PR Fixes Following Problems

Sync icon visibility during download
Accessing after download attempt to etag throws null point exception
Remove file dialog actions order is wrong for downloaded and normal file
File download
Note

The Spotbugs count increased due to the reintroduction of old code (with improvements) in this PR.

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/OfflineSyncWork.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/OfflineSyncWork.kt
@@ -115,8 +115,12 @@ class OfflineSyncWork constructor(
      * @return new etag if changed, `null` otherwise
      */
     private fun checkEtagChanged(folderName: String, storageManager: FileDataStorageManager, user: User): String? {
-        val ocFolder = storageManager.getFileByPath(folderName)
-        Log_OC.d(TAG, folderName + ": currentEtag: " + ocFolder.etag)
+        val ocFolder = storageManager.getFileByPath(folderName) ?: return null
+
+        ocFolder.etag?.let {
+            Log_OC.d(TAG, "$folderName: currentEtag: $it")
+        }
+
         // check for etag change, if false, skip
         val checkEtagOperation = CheckEtagRemoteOperation(
             ocFolder.remotePath,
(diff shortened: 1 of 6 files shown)


tobiasKaminsky:
Can you either suppress spotbugs, or manually increase count, so that CI is green?

app/src/main/java/com/nextcloud/client/jobs/OfflineSyncWork.kt
```diff
@@ -115,8 +115,12 @@ class OfflineSyncWork constructor(
      * @return new etag if changed, `null` otherwise
      */
     private fun checkEtagChanged(folderName: String, storageManager: FileDataStorageManager, user: User): String? {
-        val ocFolder = storageManager.getFileByPath(folderName)
-        Log_OC.d(TAG, folderName + ": currentEtag: " + ocFolder.etag)
+        val ocFolder = storageManager.getFileByPath(folderName) ?: return null
+
+        ocFolder.etag?.let {
+            Log_OC.d(TAG, "$folderName: currentEtag: $it")
+        }
```
ZetaTom, line 122: `Log_OC.d()` should be able to handle `null` values, shouldn't it?

```suggestion
        Log_OC.d(TAG, "$folderName: currentEtag: ${ocFolder.etag}")
```

app/src/main/java/com/nextcloud/client/jobs/download/FileDownloadHelper.kt
```diff
@@ -49,9 +49,9 @@ class FileDownloadHelper {
         val fileStorageManager = FileDataStorageManager(user, MainApp.getAppContext().contentResolver)
         val topParentId = fileStorageManager.getTopParentId(file)
 
-        return if (file.isFolder) {
-            backgroundJobManager.isStartFileDownloadJobScheduled(user, file.fileId) ||
-                backgroundJobManager.isStartFileDownloadJobScheduled(user, topParentId)
+        val isJobScheduled = backgroundJobManager.isStartFileDownloadJobScheduled(user, file.fileId)
+        return isJobScheduled || if (file.isFolder) {
+            backgroundJobManager.isStartFileDownloadJobScheduled(user, topParentId)
```
ZetaTom, line 54: Shouldn't `isJobScheduled` be joined onto line 54 (inside the `if` branch) instead of line 53? Was this change made on purpose?
Otherwise this could change the return value of `isDownloading()`.

ZetaTom (changes requested):
Downloading and uploading worked flawlessly in my tests.

Please address my comment on `FileDownloadHelper`.

tobiasKaminsky (approved):

PR #12807, opened 2024-04-03
Bugfix Parse Deeplink

alperozturk96:

**How to Test**

1. Share file
2. run this command from terminal adb shell am start "share_link"
3. App should automatically open given file

**Known Issues**

1. If Google Chrome installed, chrome catches deep link before app
2. If file doesn't exist in device user can't able to open file


alperozturk96:
Public share link not suppose to supported by app

Issue #9346, opened 2021-11-23
Deep links to Nextcloud not working in android 12


jakobroehrl:
I am afraid this is not working with Android 12 anymore. 
Tested with a Pixel smartphone. 
Removed NC App and installed it again, no change.

AlvaroBrey:
Hi @jakobroehrl. Can you give steps to reproduce and expected result? I am not entirely sure what you're referring to in this issue.

jakobroehrl:
Sure. 
There is a function to recognize nextcloud links and instead of open this in a browser the andoid app will be opened. 
But since Android 12 only the browser opens. 
@szaimen

szaimen:
I can't reproduce that. It works for me on Android 11 with 3.18.0. But it is definitely possible that something changes on Android 12 that breaks the feature.

szaimen:
@jakobroehrl maybe you could add some steps to reproduce so that Alvaro is able to try to reproduce it?

AlvaroBrey:
Most likely related: https://developer.android.com/about/versions/12/behavior-changes-all#web-intent-resolution

If this is behaving as described, we likely will have to implement domain association logic in Android 12 (as in, ask the user to confirm that the domain should be opened with the Nextcloud app). This probably should be done when the account is configured.

Ref:  https://developer.android.com/training/app-links/verify-site-associations#request-user-associate-app-with-domain

jakobroehrl:
> @jakobroehrl maybe you could add some steps to reproduce so that Alvaro is able to try to reproduce it?

Take a Android 12 phone, tip on a deep link, the link will be open in the android app (if not, reinstall the android app)
Take a Android > 12 phone, tip on a deep link, the link will be open in the browser

AndyScherzinger:
cc @schiessle this is from yesterday, so yes issue with Android 12 (his and my phone run Android 12)

InQuize:
Too bad such a fundamental feature does not deserve a uniform experience priority.
Seamless transition between web and native apps is a must for file-sharing tool.

**As a crude workaround**, normal `https://` links open in Android app using `Share` button to `Open Link With...` [Android app](https://play.google.com/store/apps/details?id=com.tasomaniac.openwith)
Alternatively, this 3rd party app may act as a (default) browser itself and redirect to native apps for each domain according to your choice (once / every time).

[...]

PR #12820, opened 2024-04-04
Handle Missing DeepLinks

alperozturk96:

### What this PR does?

It adds auto verify for intent filters
Adds missing intent filters
Calls handleOpenFileViaIntent() function in line 564 for missing case

**Note: If we support only Android 12+ we can delete  path patterns and http scheme**

### **How to Test**

**Note: Works for Only Android 11 and Below**

adb shell am start \
	-W -a android.intent.action.VIEW \
	-d "YOUR_INTERNAL_SHARE_LINK"\

### Why It's Not Working For Android 12+

In Android 12 and beyond, it's not possible to provide generic host data, and you must add an assetlinks.json file to the .well-known directory of every potential host so we can't do that.

### How It's Working For Android 12+

1. Host URL must have  assetlinks.json in .well-known directory with correct sha256_cert_fingerprints. Example -> https://example.cloud.com/.well-known/assetlinks.json

```
[
  {
    "relation": [
      "delegate_permission/common.handle_all_urls"
    ],
    "target": {
      "namespace": "android_app",
      "package_name": "com.nextcloud.client",
      "sha256_cert_fingerprints": [
        ".............."
      ]
    }
  }
]
```

2. In Android Manifest file for FileDisplayActivity intent filter must added. Example ->

```
  <intent-filter android:autoVerify="true">
                <action android:name="android.intent.action.VIEW" />

                <category android:name="android.intent.category.DEFAULT" />
                <category android:name="android.intent.category.BROWSABLE" />

                <data android:scheme="https" />
                <data android:host="example.cloud.com" />
 </intent-filter> 
```

### Useful links

Our previous [topic](https://github.com/nextcloud/android/issues/9346#issuecomment-1326113482)
[Official Documentation](https://developer.android.com/training/app-links/verify-android-applinks)
[Similar Documentation](https://doordash.engineering/2022/01/25/your-deep-links-might-be-broken-web-intents-and-android-12/)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -104,6 +104,7 @@ android {
     }
 
     defaultConfig {
+        applicationId "com.nextcloud.client"
         minSdkVersion 24
         targetSdkVersion 34
         compileSdk 34
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -148,26 +148,93 @@
             <intent-filter>
                 <action android:name="android.intent.action.SEARCH" />
             </intent-filter>
-            <intent-filter android:autoVerify="true">
+
+            <meta-data
+                android:name="android.app.searchable"
+                android:resource="@xml/users_and_groups_searchable" />
+
+            <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
 
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
 
                 <data android:scheme="http" />
+                <data android:host="*" />
+                <data android:pathPattern="/f/..*" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+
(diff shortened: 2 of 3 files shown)


AndyScherzinger:
Looks good, but can't tell if the app I'd in the manifest causes brander issues, hence cc @tobiasKaminsky

AndyScherzinger (approved):

phatair:
Hi, I'm a little bit confused because I would like to open nextcloud links with the nextcloud app on my android 14 device but it always uses the browser for opening the links.

There are a lot of infos on GitHub and in the nextcloud doc but I'm not sure if we can use the deep link feature if we use the official android nextcloud app and a self hosted nextcloud site.

We have to create assetlinks.json on the webserver but for me it looks like that we can do this only, if we develop our own nextcloud app because we need the cert finterprint of the app.

Maybe someone can help me with this.
i also have created a thread in the official forum
https://help.nextcloud.com/t/open-nextcloud-links-with-the-android-nextcloud-app-instead-with-browser/223046

Thanks.

PR #12828, opened 2024-04-05
Bugfix Handle Temp Encrypted File Correctly

alperozturk96:

1. An encrypted file is temporarily generated in the application directory instead of being stored directly on the device.
2. The temporarily encrypted file is automatically deleted under any circumstances during the uploading process.
3. The filename is correctly displayed in the notification.
4. When user clear failed uploads or clear cancelled uploads, temp_encrypted directory will be deleted as well.

**Note**

cacheDir can be used as well for this operation application directory chosen instead cacheDir just in case to cover edge case such as upload very big file can be failed due to cache directory

java.io.tmpdir doesn't have automatically remove mechanism or provide different solution to than File.createTempFile(). It's just a default temporary-file directory. For deletion on when virtual machine killed, **deleteOnExit()** method must be used. And you can use deleteOnExit() with any file created via createTempFile().

In our case, it will not save us from crash during upload scenario as well. 

Please read [documentation](https://docs.oracle.com/javase/8/docs/api/java/io/File.html) before PR Review. @ZetaTom @JonasMayerDev 

**Known Issues and Scope of PR**

Resuming or retrying cancelled uploads done via OCUpload therefore even our deleted temp encrypted files are not affect to the retry or resume scenario because we can get fully encrypted or decrypted path from OCUpload for retry or resume operation.

However during my tests e2e uploads and normal uploads for retry scenario worked but not always so for this PR it's out of scope. Please understand the general purpose of this PR, then review.

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
@@ -272,11 +272,7 @@ public class UploadFileOperation extends SyncOperation {
     }
 
     public String getFileName() {
-        if (mFile == null) {
-            return null;
-        }
-
-        return mFile.getFileName();
+        return (mFile != null) ? mFile.getFileName() : null;
     }
 
     public OCFile getFile() {
--- a/app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
@@ -753,9 +753,12 @@ public class UploadFileOperation extends SyncOperation {
                 result = unlockFolderResult;
             }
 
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -442,6 +442,7 @@ private RemoteOperationResult encryptedUpload(OwnCloudClient client, OCFile pare
         File temporalFile = null;
```
alperozturk96, line 442: Removing usage of temporal file breaks to catch scenario

tobiasKaminsky (approved):

skjnldsv:
Was checking something with the backport bot, nothing to see here, move along 
Awkward Season 4 GIF by The Office

PR #12829, opened 2024-04-05
Update strings.xml

IgorPolyakov:
the lines are confused in meaning


AndyScherzinger:
Nice catch @IgorPolyakov 🥇

PR #12847, opened 2024-04-09
Add visual hint to collapsible headers in uploads tab

JonasMayerDev:
The dividers / headers in the upload tab are collapsible, but I think it is missing a visual hint that it is.

The fact that I only found out about that feature after countless hours in that view underlines my point :)

Screenshot_2024-04-09-11-25-17-06_367f977887f4a5f4712ef04f9b973e21
Current design

Screenshot_2024-04-09-11-18-53-22_06ea656a9ba653ff96746f477f32d0f9
I just quickly added an icon for it

@nextcloud/designers what do you think? Is that icon ok?
 


JonasMayerDev:
Does anyone know if that change in the license header is the right procedure when modifying the material icon? @alperozturk96 @tobiasKaminsky @ZetaTom

nimishavijay (changes requested):
Definitely an improvement, great catch! One small nitpick about alignment: the arrow could be center-aligned with the file icons, and the text ("Aktuell (5)", "Hochgeladen (5)") could be left aligned with the file names :)

alperozturk96 (approved):

marcoambrosini (approved):

AndyScherzinger (approved):

PR #12880, opened 2024-04-15
Assistant Feature UI/UX Fixes

alperozturk96:

**Improvements**

1. Drawer layout accessible via Menu button from Assistant Screen
2. Assistant feature visibility changes according to banner visibility, no more duplication
3. Task output text can be so big and after expand screen is not usable, Task Detail Screen added as bottom sheet, so user can see all output in bottom sheet.

**How to Test**

If client is branded and assistant feature enabled only must be visible in drawer menu

If client is not branded and assistant feature enabled only must be visible in top banner

If client is not branded and assistant feature disabled, top banner shouldn't be visible and assistant feature not be visible

If client is branded and assistant feature disabled, top banner should be visible and assistant feature not be visible

Task detail page should show input and output as expected and scroll should be intuitive with very long outputs. (You can use preview for test)

**Demo**

https://github.com/nextcloud/android/assets/67455295/c7d4ca1b-063b-454f-acae-1d511a0330ab


tobiasKaminsky:
Please re-test with all combination of is_branded_client: true/false  and server with/without assistant.
Check then that entry is shown/hidden correctly.

tobiasKaminsky (approved):

PR #12928, opened 2024-04-24
Increase Cursor Size To Fix SQLiteBlobTooBigException

alperozturk96:

This PR can make this problem less likely to happen.  

Similar [discussion](https://github.com/andpor/react-native-sqlite-storage/issues/364#issuecomment-526423153).
Similar [discussion from our GitHub issue history](https://github.com/nextcloud/android/issues/10213).

### DIFF
--- a/app/src/main/java/com/owncloud/android/MainApp.java
+++ b/app/src/main/java/com/owncloud/android/MainApp.java
@@ -28,6 +28,7 @@ import android.content.pm.ActivityInfo;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.content.res.Resources;
+import android.database.CursorWindow;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.Environment;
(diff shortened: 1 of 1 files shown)


AlvaroBrey:
Unless this crash had a recent spike that needs urgent remediation, I'd advise against this workaround.

- Uses an internal API via reflection, which may stop working with Android updates
- Hides the problem rather than solving it, which may make it blow up even more spectacularly in the future
- And finally, this would make **_every_ cursor in the app allocate a 100mb memory size**. This can quickly cause memory overuse problems.

Generally the approach should be to:
1. Use projections where possible, to not get the entire rows into memory unless absolutely necessary
2. Work towards removing the infinitely-growable fields out of the database somehow. Just avoiding the storage of `rich_workspace` would likely stop _most_ of the crashes.

alperozturk96:
> Unless this crash had a recent spike that needs urgent remediation, I'd advise against this workaround (and even so I'd tune the size down).
> 
> * Uses an internal API via reflection, which may stop working with Android updates
> * Hides the problem rather than solving it, which may make it blow up even more spectacularly in the future
> * And finally, this would make **_every_ cursor in the app allocate a 100mb memory size**. This can quickly cause memory overuse problems.
> 
> Generally the approach should be to:
> 
> 1. Use projections where possible, to not get the entire rows into memory unless absolutely necessary
> 2. Work towards removing the infinitely-growable fields out of the database somehow. Just avoiding the storage of `rich_workspace` would likely stop _most_ of the crashes.

Thanks for the feedback. Your suggestion make more sense to me as well. We can check cursor usages and if possible we can use projections to reduce unnecessary work.

PR #12942, opened 2024-04-26
Bugfix E2E File Deletion For V1

alperozturk96:

**How to Test**

1. Create encrypted folder
2. Upload multiple files inside encrypted folder
3. Delete one file from encrypted folder
4. Refresh folder, file must be deleted
5. Try to add some file again, file conflict notification should't appear

Test it for V1 (Nextcloud 27.1.7) and V2 (v28 and above)

### Note

E2E version must be lower than 2.0


tobiasKaminsky:
Approved, restarted CA as "rename" test failed.

PR #12950, opened 2024-04-30
Give better feedback to the user for UNHANDLED_HTTP_CODE in PreviewMediaActivity

alperozturk96:

**How to Reproduce**

1. Test on Nextcloud 27.1.7 instance
2. Open video file
3. Try to delete from top right inside preview screen


tobiasKaminsky:
This seems to be very specific.
How can we know that this UNHANDLED_HTTP_CODE is really about a locked file?

alperozturk96:
> This seems to be very specific. How can we know that this UNHANDLED_HTTP_CODE is really about a locked file?


You are right. @artonge 


> This seems to be very specific. How can we know that this UNHANDLED_HTTP_CODE is really about a locked file?

Yes you are right. I used httpPhrase rather than code

JonasMayerDev (dismissed):
I would suggest you do the error handling in [this](https://github.com/nextcloud/android/blob/e3d57e7c42efa5dd4b52aeea71ee1dbbd903fe19/app/src/main/java/com/owncloud/android/utils/ErrorMessageAdapter.java#L290) Function to make the code consistent. 

Maybe also test if you can use ResultCode.LOCK_FAILED instead of checking the HttpPhrase?

alperozturk96:
> I would suggest you do the error handling in [this](https://github.com/nextcloud/android/blob/e3d57e7c42efa5dd4b52aeea71ee1dbbd903fe19/app/src/main/java/com/owncloud/android/utils/ErrorMessageAdapter.java#L290) Function to make the code consistent.
> 
> Maybe also test if you can use ResultCode.LOCK_FAILED instead of checking the HttpPhrase?

Didn't aware of existence of this function, cool.

app/src/main/java/com/owncloud/android/operations/common/RemoteOperationHttpPhrase.kt
```diff
@@ -0,0 +1,12 @@
+/*
+ * Nextcloud - Android Client
+ *
+ * SPDX-FileCopyrightText: 2024 Your Name <your@email.com>
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+package com.owncloud.android.operations.common
+
+enum class RemoteOperationHttpPhrase(val value: String) {
```
tobiasKaminsky, line 10: I think this is handled better in Library.
Please check in RemoterOperationResult:254 and followling.
There we have the exception by server, which should be "locked" and then we can a real mCode?

alperozturk96:
Android-library updated and locked HTTP code added in library side and used in client side.

JonasMayerDev:
How can I test? How do I create a locked file to test with?

JonasMayerDev (approved):
Fixes the Issue in quick test, code looks good

AndyScherzinger (approved):

PR #12966, opened 2024-05-02
Fix Image Preview Header

alperozturk96:


JonasMayerDev (approved):
Works for me, Code looks good

PR #13015, opened 2024-05-16
Create individual workers for every Autoupload Folder

JonasMayerDev:
This PR finally gets rid of the foreground worker and the corresponding notification by lowering the chance the worker gets killed.

Internally, the app now starts a new worker for every auto upload folder instead of one worker for all folders.
Advantages are:

- Worker finish quicker -> Lower chance of being killed
- Scanning of Folders can happen in parallel
- Better handling in the future

The best way to test is using logcat and filtering for "File sync" since all important log messages start with that term. Like this, you should be able to observe the behavior of the background worker without messing with it. If you want to check something specific, use the debugger, but be aware that that changes behavior (like execution speed).


alperozturk96 (approved):
Tested. I created 3 folder set auto upload and only 3 worker created for each auto upload and all uploads successfully completed.

PR #13057, opened 2024-05-31
Added deep link support for screens navigation

surinder-tsys:
### Description
This PR adds deep linking support to the Android application, enabling users to navigate directly to specific screens via URLs. 

### Changes
- Added path pattern filters in AndroidManifest.xml
- Implemented URL parsing and navigation logic

### Testing
- Manually tested various deep link scenarios


tobiasKaminsky:
@alperozturk96 when you are testing this, please also check if security (pin / biometric) prevents opening those deep links.

alperozturk96 (approved):
Tested with passcode its working also I hide loading dialog for navigation paths

app/src/main/AndroidManifest.xml
```diff
@@ -160,6 +161,8 @@
                 <data android:pathPattern="/..*/f/..*" />
                 <data android:pathPattern="/..*/..*/f/..*" />
                 <data android:pathPattern="/..*/..*/..*/f/..*" />
+                <!-- path pattern to handle deep link -->
+                <data android:pathPattern="/app/..*" />
```
szaimen, line 165: What about installations in subpaths? Also shouldnt it be `/apps/..*`

alperozturk96: I would say depends on use case scenario. Firstly, deep link will not work above Android 11 due to Google changes. [Explanation](https://github.com/nextcloud/android/pull/12820#issue-2225141822)

And this PR supports this kind of URL “example.com/app/openMedia” it will work, regardless of web app.

Android 11 and below can open related screen from deep link and that deep link may come from any message or email etc.

To provide general naming consistency across all over our apps, we can say "apps/mail", "apps/spreed" can be better naming. 

The real question is, Android app can open deep link for Android 11 and below, but what is going to send those deep links to the user and in what scenario we will send?

PR #13065, opened 2024-06-04
Bugfix Start Foreground Service When App In Foreground

alperozturk96:

**What does this PR?**

Only starts foreground service when app in foreground so we don't came across background crashes anymore
No need to do it for FOREGROUND_SERVICE_MEDIA_PLAYBACK because Media Playback not causing that issue


tobiasKaminsky:
Instead of writing own lifecycle, can't we re-use the Android one?
Like we do here: https://github.com/nextcloud/android/blob/8bb1eaaf2e86623e42175aa7597daabc41e9b0b3/app/src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.kt#L826

alperozturk96:
> Instead of writing own lifecycle, can't we re-use the Android one? Like we do here:
> 
> https://github.com/nextcloud/android/blob/8bb1eaaf2e86623e42175aa7597daabc41e9b0b3/app/src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.kt#L826



> Instead of writing own lifecycle, can't we re-use the Android one? Like we do here:
> 
> https://github.com/nextcloud/android/blob/8bb1eaaf2e86623e42175aa7597daabc41e9b0b3/app/src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.kt#L826



> Instead of writing own lifecycle, can't we re-use the Android one? Like we do here:
> 
> https://github.com/nextcloud/android/blob/8bb1eaaf2e86623e42175aa7597daabc41e9b0b3/app/src/main/java/com/owncloud/android/ui/activity/SyncedFoldersActivity.kt#L826

Then we need to add this dependency implementation "androidx.lifecycle:lifecycle-service:2.8.1" to access lifecycle from service. PR is updated.

AndyScherzinger (approved):

PR #13076, opened 2024-06-06
Bugfix NPE & Convert to Kotlin SSL Untrusted Cert Dialog

alperozturk96:

Previously SSL Untrusted Cert Dialog was java file and a lot of nullable variable used everywhere. I checked our crash reports and a lot of user affected because of NPE. This PR converts java file to Kotlin and safely uses the variables.


app/src/main/java/com/owncloud/android/ui/dialog/SslUntrustedCertDialog.kt
```diff
@@ -0,0 +1,210 @@
...
+ * the certificate from different classes.
+ */
+open class SslUntrustedCertDialog : DialogFragment(), Injectable {
+    @JvmField
+    @Inject
+    var viewThemeUtils: ViewThemeUtils? = null
+
+    protected var binding: SslUntrustedCertLayoutBinding? = null
+    protected var mHandler: SslErrorHandler? = null
+    protected var m509Certificate: X509Certificate? = null
```
tobiasKaminsky, line 53: totally nitpicking here, but while on it, we can remove those "m" prefixes, as we do not use them anymore.

tobiasKaminsky:
How can I test this dialog?
With login v1 I could use my self signed cert on test server, but now with v2?

alperozturk96:
> How can I test this dialog? With login v1 I could use my self signed cert on test server, but now with v2?



> How can I test this dialog? With login v1 I could use my self signed cert on test server, but now with v2?

IDK. What's the difference for certificate between v1 and v2? Can't you create certificate for v2 manually?

tobiasKaminsky:
Right… I was wrong with that one…

tobiasKaminsky (approved):

Issue #13083, opened 2024-06-07
After closing photo, position returns to top of gallery 


coone4life:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

Open app
Click on the three stripes in left upper corner
Click on Media
Scroll down and open a picture
Close the picture 
The app jumps to the begin of the gallery

### Expected behaviour

Not jumping to the begin, stand there where the photo is in the gallery.

### Actual behaviour

The app jumps to the begin of the gallery and I had to scroll down again.

### Android version

14

### Device brand and model

Google Pixel 7 Pro

### Stock or custom OS?

Stock

### Nextcloud android app version

3.29.0

### Nextcloud server version

29.0.2

### Using a reverse proxy?

No

PR #13095, opened 2024-06-11
Bugfix Gallery scroll #13083

satvik2131:

**Demo**
https://github.com/nextcloud/android/assets/59243339/fe5f5118-55f2-4c03-9fd7-db85715944a3


alperozturk96:
Hello

Thank you for your contribution.

I just noticed this PR. I solved the same problem with [another PR](https://github.com/nextcloud/android/pull/13132#pullrequestreview-2118179451), but this solution seems to be better. And I also discovered another bug in the master related to usage of LocalBroadcastManager.

After previewing the image from the media tab, the media screen is not updated when the deletion is performed from preview, we have solved this problem before, but another change has reappeared this problem. Using LocalBroadcastManager is not solving this problem.

I will check it later on.

alperozturk96:
Duplication [of](https://github.com/nextcloud/android/pull/13132)

PR #13108, opened 2024-06-12
mdm only needs to provide server address

tobiasKaminsky:
fallback to normal login


alperozturk96 (approved):
Tested

PR #13145, opened 2024-06-17
Bug-Fix Percentage Value During Download or Upload

alperozturk96:

Most of the time, the percentage value is correct. However, some users have encountered abnormally large integer values at the end of the upload or download progress. **To suppress** this issue until the root cause is identified, `coerceAtMost` and the `OnDataTransferProgressListener` extension used to prevent incorrect usage.

### DIFF
--- a/app/src/main/java/com/nextcloud/utils/extensions/OnDataTransferProgressListenerExtensions.kt
+++ b/app/src/main/java/com/nextcloud/utils/extensions/OnDataTransferProgressListenerExtensions.kt
@@ -1,7 +1,7 @@
 /*
  * Nextcloud - Android Client
  *
- * SPDX-FileCopyrightText: 2024 Your Name <your@email.com>
+ * SPDX-FileCopyrightText: 2024 Alper Ozturk <alper.ozturk@nextcloud.com>
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/nextcloud/utils/extensions/OnDataTransferProgressListenerExtensions.kt
```diff
@@ -0,0 +1,16 @@
+/*
+ * Nextcloud - Android Client
+ *
+ * SPDX-FileCopyrightText: 2024 Your Name <your@email.com>
```
JonasMayerDev, line 4: Add your name :)

JonasMayerDev (changes requested):

app/src/main/java/com/nextcloud/utils/extensions/OnDataTransferProgressListenerExtensions.kt
alperozturk96: I'm always changing Copyright from Android Studio but still not applying it correctly, thanks for the review.

JonasMayerDev (approved):
Could not reproduce bug but works for me and code looks good.

AndyScherzinger (approved):

PR #13149, opened 2024-06-19
Feature Open Calendar Events and Contacts

alperozturk96:
Needs https://github.com/nextcloud/server/pull/46464

**How to Test?**

1. Install [DAVx⁵](https://f-droid.org/en/packages/at.bitfire.davdroid/) and sync via Nextcloud account
2. Search calendar event or contact and give permission to read calendar event and contact, then press event app should redirect to internal app instead of web app.

**How to Test?**

1. Don't give permission then try to open calendar event or contact, app should redirect to the web app.

**Demo**

https://github.com/nextcloud/android/assets/67455295/1cfadfd3-e142-409f-abdd-6d85b3b5ffa5

**Duplicated Contacts**

https://github.com/user-attachments/assets/2d333570-c39d-4567-b188-53252ffb7b50

### DIFF
--- a/app/src/main/java/com/nextcloud/model/SearchResultEntryType.kt
+++ b/app/src/main/java/com/nextcloud/model/SearchResultEntryType.kt
@@ -1,0 +1,25 @@
+/*
+ * Nextcloud - Android Client
+ *
+ * SPDX-FileCopyrightText: 2024 Alper Ozturk <alper.ozturk@nextcloud.com>
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+package com.nextcloud.model
+
+import com.owncloud.android.R
+
+enum class SearchResultEntryType {
+    CalendarEvent, Folder, Note, Contact, Deck, Unknown;
+
+    fun getIconId(): Int? {
+        return when(this) {
+            Folder -> R.drawable.folder
+            Note -> R.drawable.ic_edit
+            Contact -> R.drawable.file_vcard
+            CalendarEvent -> R.drawable.file_calendar
+            Deck -> R.drawable.ic_deck
+            else -> null
+        }
+    }
+}
--- a/app/src/main/java/com/nextcloud/model/SearchResultEntryType.kt
+++ b/app/src/main/java/com/nextcloud/model/SearchResultEntryType.kt
@@ -13,7 +13,7 @@ enum class SearchResultEntryType {
     CalendarEvent, Folder, Note, Contact, Deck, Unknown;
 
     fun getIconId(): Int? {
-        return when(this) {
+        return when (this) {
             Folder -> R.drawable.folder
             Note -> R.drawable.ic_edit
             Contact -> R.drawable.file_vcard
--- a/app/src/main/java/com/nextcloud/model/SearchResultEntryType.kt
+++ b/app/src/main/java/com/nextcloud/model/SearchResultEntryType.kt
@@ -17,14 +17,14 @@ enum class SearchResultEntryType {
     Deck,
     Unknown;
 
-    fun getIconId(): Int? {
+    fun iconId(): Int {
         return when (this) {
             Folder -> R.drawable.folder
             Note -> R.drawable.ic_edit
             Contact -> R.drawable.file_vcard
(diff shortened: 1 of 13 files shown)


AndyScherzinger (approved):
Code lgtm

alperozturk96:
@tobiasKaminsky Can we merge this?

app/src/main/java/com/nextcloud/model/SearchResultEntryType.kt
```diff
@@ -0,0 +1,30 @@
...
+    Unknown;
+
+    fun getIconId(): Int? {
+        return when (this) {
+            Folder -> R.drawable.folder
+            Note -> R.drawable.ic_edit
+            Contact -> R.drawable.file_vcard
+            CalendarEvent -> R.drawable.file_calendar
+            Deck -> R.drawable.ic_deck
+            else -> null
```
tobiasKaminsky, line 27: as fallback maybe 
https://fonts.google.com/icons?selected=Material+Symbols+Outlined:find_in_page:FILL@1;wght@400;GRAD@0;opsz@24&icon.query=search&icon.size=24&icon.color=%235f6368

"find in page".

app/src/main/java/com/nextcloud/utils/extensions/AccountExtensions.kt
```diff
@@ -9,6 +9,12 @@ package com.nextcloud.utils.extensions
 
 import android.accounts.Account
 import android.content.Context
+import com.nextcloud.client.account.UserAccountManager
 import com.owncloud.android.R
+import com.owncloud.android.lib.resources.status.NextcloudVersion
 
 fun Account.isAnonymous(context: Context): Boolean = type.equals(context.getString(R.string.anonymous_account_type))
+
+fun UserAccountManager.isServerVersionNewerOrEqual(version: NextcloudVersion): Boolean {
```
tobiasKaminsky, line 18: Please simply use the method in UnifiedSearchFragment directly.
There is no need to have such an extension.

Furthermore it hides the info how the server version is checked.

app/src/main/java/com/nextcloud/utils/extensions/SearchResultEntryExtensions.kt
```diff
@@ -0,0 +1,36 @@
...
+        SearchResultEntryType.CalendarEvent
+    } else if (icon.startsWith("icon-deck")) {
+        SearchResultEntryType.Deck
+    } else {
+        SearchResultEntryType.Unknown
+    }
+}
+
+@Suppress("MagicNumber")
+fun SearchResultEntry.createdAt(): Long? = attributes["createdAt"]?.toLongOrNull()?.times(1000L)
```
tobiasKaminsky, line 30: All those extension can only work with Calender/Conact.
So maybe put them there? 
Else other devs think that those are general available (for other types).

app/src/main/java/com/owncloud/android/ui/adapter/UnifiedSearchListAdapter.kt
```diff
@@ -73,6 +74,7 @@ class UnifiedSearchListAdapter(
                     false
                 )
                 UnifiedSearchItemViewHolder(
+                    isServerVersionThirtyOrAbove,
```
tobiasKaminsky, line 77: maybe better "supportsOpeningCalendarContactsLocally"?

app/src/main/java/com/nextcloud/utils/extensions/AccountExtensions.kt
alperozturk96: Why there is no need for this extension? I may need to check server version for any other future implementation.

Also, how many different way do we have for server version check? If yes, what kind of problem can it be if the logic is different? If it is, if the result is different, we can say that this is a problem.

Otherwise, I am not in favor of putting shareable logic into fragments, activities.

alperozturk96: https://github.com/nextcloud/android/pull/13149/commits/730af29fe39e44bbe2bbe9838a6d3d8b3c652925

Issue #10389, opened 2022-06-14
no previews generated/shown for local images


svenb1234:
### Steps to reproduce
1. Let the app upload and move images and videos to the app's folder
2. Start the media view in the app

### Expected behaviour
- previews for all pictures and videos should be generated and shown

### Actual behaviour
- only previews for pictures that are not on the phone are generated and shown. Instead the app shows the grey static image/video icon - not the colour changing icon which is shown while the preview is loaded/generated. Opening the pictures works though.

When swiping, the next local pictures are loaded as well. However, local videos are not loaded when opened via swiping. The display just shows "sorry". Opening the local videos directly from the thumbnail view plays them fine.

### Environment data
Android version: 12

Device model: Pixel 6

Stock or customized system: stock

Nextcloud app version: 3.20.3

Nextcloud server version: 23

Tiam-Abderezai:
@nextcloud-android-bot Please grant access / assign issue to me

svenb1234:
If you do not get a video preview that's probably because ffmpeg or similar is not installed. Also, you are not looking at pictures which alre local, i.e. on your phone's nextcloud folder. At least I do not see any green ticks on your screenshot.

![Screenshot_20220617-071805](https://user-images.githubusercontent.com/46424536/174231769-6e4763e1-7271-4693-94ec-28feac48e6ad.png)

Unfortunately I cannot show you any screenshot of the media view (not file view) since it crashes. But there one would see that those pictures which are only on the server get a preview, the local ones not.

As mentioned above, I use the "move to appp folder after upload feature" of the android app.

Tiam-Abderezai:
@svenb1234  Thanks for the screenshot.
> If you do not get a video preview that's probably because ffmpeg or similar is not installed

Ah that's right, our NC org doesn't have ffmpeg installed. The org I'm a member of isn't hosted by me so I can't add it. 

Therefore I'm going to :
1) install a local NC server and join in as a member
2) install ffmpeg on it
3) test by reproducing the error again by following your steps

>  Also, you are not looking at pictures which are local, i.e. on your phone's nextcloud folder. At least I do not see any green ticks on your screenshot

>  But there one would see that those pictures which are only on the server get a preview, the local ones not. As mentioned above, I use the "move to appp folder after upload feature" of the android app.

Ok, I will take these in consideration as well as soon as I start testing and update you soon with my findings.

mlsxlist:
[...]

PR #13158, opened 2024-06-20
fix for media tab not showing thumbnails when image not in cache manager

camillo-positano:

Hello i've had this issue for a long time and wanted to fix it.

The problem i found was that the cache manager never requested the server for a thumbnail because the condition to do so was never true. This caused that all my remote photos thumbnails were the default thumbnail image.

From what i understand if there is not a preview but there is a remote id it should ask the server for a thumbnail.

This never happened because of an && instead of an ||.

I also had to set updateThumbnailNeeded to false when it finished otherwise it entered a loop because it kept passing the validation: if (thumbnail == null || file.isUpdateThumbnailNeeded()).

Regarding test i didn't really understand where and how to test this functionality because, in case of an integration test, i need a server to call.

I would be more than happy to add a test if some kind soul could give me a little help xD <3

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -296,51 +296,58 @@ public final class ThumbnailsCacheManager {
             if (file.getRemoteId() != null || file.isPreviewAvailable()) {
                 // Thumbnail in cache?
                 thumbnail = ThumbnailsCacheManager.getBitmapFromDiskCache(
-                    ThumbnailsCacheManager.PREFIX_RESIZED_IMAGE + file.getRemoteId()
-                                                                         );
+                    ThumbnailsCacheManager.PREFIX_RESIZED_IMAGE + file.getRemoteId());
 
-                if (thumbnail != null && !file.isUpdateThumbnailNeeded()) {
-                    Float size = (float) ThumbnailsCacheManager.getThumbnailDimension();
+                if (thumbnail != null && !file.isUpdateThumbnailNeeded())
+                    return getThumbnailFromCache(thumbnail);
 
-                    // resized dimensions
-                    ImageDimension imageDimension = file.getImageDimension();
-                    if (imageDimension == null ||
-                        imageDimension.getWidth() != size ||
-                        imageDimension.getHeight() != size) {
-                        file.setImageDimension(new ImageDimension(thumbnail.getWidth(), thumbnail.getHeight()));
-                        storageManager.saveFile(file);
-                    }
-
-                    if (MimeTypeUtil.isVideo(file)) {
-                        return ThumbnailsCacheManager.addVideoOverlay(thumbnail, MainApp.getAppContext());
-                    } else {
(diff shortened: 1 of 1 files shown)


alperozturk96 (changes requested):
Hello

Thanks for your contribution 💯 

After this change, I came across something like this. Besides that current implementation on master for ThumbnailsCacheManager also not ideal and good, we will check it, thank you so much 👍 

https://github.com/nextcloud/android/assets/67455295/f1f2b47d-2c8b-4da4-aeb9-e22051638a57

camillo-positano:
Thanks for the test, will verify! 

Could you give me any info of the steps you did to encounter this issue? 

Because on my device with my nextcloud instance that didn't happen. 
Were those images on the device or on the server? 

ty!

camillo-positano:
Hey should be fixed now! could you give it a check?

app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -289,7 +289,7 @@ protected Bitmap doInBackground(Object... params) {
             file = (OCFile) params[0];
```
alperozturk96, line 289: Thanks for the fast response. It's working now like master, but file can be null, because **_Object... params_** maybe not OCFile, maybe empty OCFile array etc.

Could you add those null checks and type checks inside doInBackground function? 

```
            if (params != null && params.length > 0 && params[0] instanceof OCFile downloadedFile) {
                file = downloadedFile;
            } else {
                Log_OC.d(TAG, "Downloaded file is null");
                return null;
            }

            if (file.getRemoteId() == null || !file.isPreviewAvailable()) {
                Log_OC.d(TAG, "File cannot be previewed");
                return null;
            }
```

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
camillo-positano: Sure thing, will do!

camillo-positano: i added the check on params but inverted what you asked to implement a fail fast principle.

then the check you asked to add on the remote id and preview available was already "implemented" so i simply added a log when it enters the else statement.
Because if there is no remote id and the preview is false it will enter that else statement instead of trying to get from the cache or server

camillo-positano: I've also just pushed the extraction in two separate methods!

alperozturk96 (approved):
Tested, working

camillo-positano:
This merge should also close this isse: #10389

AndyXheli:
@camillo-positano do you have an apk so I can insult and test this please ?

camillo-positano:
@AndyXheli if you want to give it a try here is the apk: https://we.tl/t-MCTLpO7EyX

AndyScherzinger:
It will also be part of the next daily build :)

PR #13160, opened 2024-06-20
Remove unnecessary checks and calls from restartFailedUploads

JonasMayerDev:
The main idea of this PR was to fix double-checking upload conditions first in FileSyncHelper and then in FilesUploadHelper.

I noticed that currently these checks are in my opinion useless. The checks should only schedule those uploads that make sense to do at that moment (e.g. if upload is Wi-Fi only and user is not connected to Wi-Fi upload should not be scheduled and display an info message). Those delayed uploads have the state "failed" and last result e.g. "Delayed_for_wifi". This makes sense to me. The problem is that currently the worker not only tries to upload the scheduled Files but also the delayed files every time. 

This makes no sense to me. 
I guess there are 2 options:
1. Remove all the checks before scheduling, schedule all failed uploads, and let the FileUploadWorker decide which to start and which to skip. -> Basically the same behavior as we have currently, but more consistent and less unnecessary double-checks.
2. Make the Worker only try to upload scheduled uploads. This changes the current behavior, since it would only try to upload delayed Uploads when "retryFailedUploads" is called. Currently, they are retried every time the upload worker runs (so e.g. also when uploading a new file etc.). Would be more performant and also more consistent than currently.

@tobiasKaminsky @ZetaTom @alperozturk96 What do you think is the best approach here? Do you think this changed behavior would be ok? 

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/FilesSyncWork.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/FilesSyncWork.kt
@@ -182,7 +182,7 @@ class FilesSyncWork(
         } else {
             // Check every file in synced folder for changes and update
             // filesystemDataProvider database (potentially needs a long time)
-            FilesSyncHelper.insertAllDBEntries(syncedFolder, powerManagementService)
+            FilesSyncHelper.insertAllDBEntriesForSyncedFolder(syncedFolder)
         }
     }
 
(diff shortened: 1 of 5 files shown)


tobiasKaminsky:
> 2\. Make the Worker only try to upload scheduled uploads. This changes the current behavior, since it would only try to upload delayed Uploads when "retryFailedUploads" is called. Currently, they are retried every time the upload worker runs (so e.g. also when uploading a new file etc.). Would be more performant and also more consistent than currently.

I am for this one.
As this is then a bigger change, once it is in master/daily we should check that those failed ones are uploaded at some time (monitor on our own devices)

tobiasKaminsky (approved):

ZetaTom (approved):
Works fine on my machine. Adjusted some variable names and formatting.

PR #13164, opened 2024-06-21
use preview endpoint instead of thumbnail

tobiasKaminsky:
do not use path, but fileid for previews


alperozturk96 (approved):
Tested working

AndyScherzinger:
should we check that the id is not -1 ?

ZetaTom (approved):

PR #13165, opened 2024-06-21
fix dav4jvm checksum

tobiasKaminsky:


tobiasKaminsky:
@AndyScherzinger https://github.com/bitfireAT/dav4jvm/releases/tag/2.2.1
they did not changed their files…
Maybe it is maven specific problem?
Not sure, how to proceed?

Same  with lib: https://github.com/nextcloud/android-library/pull/1483

AndyScherzinger (approved):

AndyScherzinger:
yeah for some reason the release artifacts were gone from jitpack, and I was magically able to re-trigger it by calling the right jitpack URLs which than recreated the artifacts. I expect the artifacts are hence slightly different in terms of checksums because jitpack uses the latest version of the pinned major java version.

PR #13184, opened 2024-06-25
Bugfix Unified Search Load More

alperozturk96:

In certain scenarios even user have more data, "Load More Data" not visible.

e.g. 

User have F1, F2, F3, F4, f 5, f 6, f7, f 8 folders when user make search via "f" user only sees last 4 item without "load more data"

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/unifiedsearch/SearchOnProviderTask.kt
+++ b/app/src/main/java/com/owncloud/android/ui/unifiedsearch/SearchOnProviderTask.kt
@@ -16,7 +16,7 @@ class SearchOnProviderTask(
     private val provider: String,
     private val client: NextcloudClient,
     private val cursor: Int? = null,
-    private val limit: Int = 5
+    private val limit: Int = 6
 ) : () -> SearchOnProviderTask.Result {
     companion object {
         private const val TAG = "SearchOnProviderTask"
--- a/app/src/main/java/com/owncloud/android/ui/unifiedsearch/SearchOnProviderTask.kt
+++ b/app/src/main/java/com/owncloud/android/ui/unifiedsearch/SearchOnProviderTask.kt
@@ -16,7 +16,7 @@ class SearchOnProviderTask(
     private val provider: String,
     private val client: NextcloudClient,
     private val cursor: Int? = null,
-    private val limit: Int = 6
+    private val limit: Int = 5
 ) : () -> SearchOnProviderTask.Result {
     companion object {
         private const val TAG = "SearchOnProviderTask"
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/owncloud/android/ui/unifiedsearch/UnifiedSearchViewModel.kt
```diff
@@ -94,14 +94,6 @@ class UnifiedSearchViewModel(application: Application) : AndroidViewModel(applic
         )
     }
 
-    open fun startLoading(query: String) {
```
alperozturk96, line 97: unused function

app/src/main/java/com/owncloud/android/ui/unifiedsearch/UnifiedSearchViewModel.kt
```diff
@@ -185,10 +177,6 @@ class UnifiedSearchViewModel(application: Application) : AndroidViewModel(applic
         }
     }
 
-    open fun clearError() {
```
alperozturk96, line 188: unused function

app/src/main/java/com/owncloud/android/ui/unifiedsearch/SearchOnProviderTask.kt
```diff
@@ -16,7 +16,7 @@ class SearchOnProviderTask(
     private val provider: String,
     private val client: NextcloudClient,
     private val cursor: Int? = null,
-    private val limit: Int = 5
+    private val limit: Int = 6
```
alperozturk96, line 19: Why is the result of UnifiedSearchRemoteOperation 1 less than the limit we have? If limit is 6 I'm getting 5 results, if 5 I'm getting 4 results etc.

app/src/main/java/com/owncloud/android/ui/unifiedsearch/UnifiedSearchViewModel.kt
```diff
@@ -213,11 +194,13 @@ class UnifiedSearchViewModel(application: Application) : AndroidViewModel(applic
...
             .filter { it.value.results.isNotEmpty() }
             .map { (key, value) ->
+                val hasMoreResults = results.isNotEmpty() && results[key]?.nextCursor() != null
+
                 UnifiedSearchSection(
                     providerID = key,
                     name = value.name()!!,
                     entries = value.results.flatMap { it.entries },
-                    hasMoreResults = !value.isFinished()
+                    hasMoreResults = hasMoreResults
```
AndyScherzinger, line 203: This statement seems not needed...

tobiasKaminsky (approved):

PR #13208, opened 2024-06-28
Crash Fix Internet Connection Check On Item Click

alperozturk96:

It checks mContainerActivity's nullability and type for the checking internet connection


AndyScherzinger (approved):

ralhei:
Hey everyone, thanks for providing fixes for this issue, and if I read it correctly, it should be incorporated into v3.30.0
This is the version I'm currently running, but the problem persists - when ever the screen turns black (however the phone is not locked yet), the playing immediately stops (which I also wouldn't expect and surprises me, but this is a different question), but when I tap on the screen and the app is visible again, it still starts playing from the beginning of the track. Seems like this problem is not completely solved, at least for me.

Issue #13035, opened 2024-05-23
Only pictures are enabled but videos also auto-synced


ThomasKahl:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Enable sync of pictures only
2. Make a video in this directory
3. Let the auto sync function sync the files.

### Expected behaviour

If the video sync function is not enabled, then videos shall not be synced.

### Actual behaviour

If the video sync function is not enabled, then videos are synced together with pictures.

### Android version

14

### Device brand and model

Samsung A54

### Stock or custom OS?

Stock

### Nextcloud android app version

3.29.0

### Nextcloud server version

25.0.13

### Using a reverse proxy?

No

microbearlogist:
Bug still present in 3.29.1

awptechnologies:
Still present in 3.29.2

joshtrichards:
> Bug still present in 3.29.1

That's because the mentioned fix (#13255) is only in >=3.30.0.

PR #13217, opened 2024-07-01
Make filesync worker persistant

JonasMayerDev:
Should Fix https://github.com/nextcloud/android/issues/13035


JonasMayerDev:
The Idea tested in this PR to remember in the DB when, which files were scanned seems to be unfeasible right now. 
The overhead produced when interacting with the DB (adding all files) every time the worker runs adds too much overhead (in my testing, nearly 65% in the best case). Also, trying to interact with the DB in batches did not solve the problem. I will close this PR and try another Idea I had while testing in another PR to keep it organized and documented.

PR #13218, opened 2024-07-01
Make upload handling more efficient

JonasMayerDev:
This PR is a combination of multiple small improvements.
1. Make the cancelAllUploads more efficient and make it not run on the UI Thread
2. Fix a bug where retryAllCancelledUploads duplicates the displayed uploads in the UploadListActivity
3. "Rate limit" retryFailedUploads since it gets called a lot, for example during startup -> Code makes now sure that not multible Threads call retryFailedUploads. During startup and when reloading in UploadListActivity it skips now 2 times -> more efficient


alperozturk96 (approved):
Tested working, statements are correct. retryFailedUploads called from multiple places. cancelFileUploads was running on UI thread and with this PR it fixed.

Issue #13222, opened 2024-07-03
OutOfMemory exception in Nextcloud app when using the nextcloud/Android-SingleSignOn library to upload large files


dipshit-enterprises:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

using the nextcloud/Android-SingleSignOn library:
1. initiate an HTTP PUT request for WebDav
2. use an InputStream to feed the request body
3. have the request body be around 300 MB or more, and quickly accessible

### Expected behaviour

The Nextcloud Files app copies a fixed/limited number of bytes from the request body's buffer from my app over the AIDL interface, making sure the bytes hit my Nextcloud server before reading more so that it doesn't run out of memory.

### Actual behaviour

The Nextcloud Files app instead reads as much of the buffer as possible from my app over the AIDL interface with no limits and without waiting for the bytes to hit my server. It quickly approaches the JVM limit within seconds and crashes, as my app is able to provide the data much faster than it can be uploaded to my server.

### Android version

13

### Device brand and model

samsung SM-G975F

### Stock or custom OS?

Custom (explain in "additional information")

### Nextcloud android app version

3.29.1

### Nextcloud server version

28.0.7

### Using a reverse proxy?

Yes

### Android logs

```
2024-07-02 21:27:34.308 17483-17483 ShowErrorActivity        D  onPause() ending
2024-07-02 21:27:34.907 17483-17483 ShowErrorActivity        D  onStop() ending
2024-07-02 21:27:34.912 17483-17483 ShowErrorActivity        D  onSaveInstanceState(Bundle) starting
2024-07-02 21:27:44.055 15564-15564 AppCompatDelegate        D  Checking for metadata for AppLocalesMetadataHolderService : Service not found
2024-07-02 21:27:44.069 15564-15564 SsoGrantPe...onActivity  D  onCreate(Bundle) starting
2024-07-02 21:27:44.086 15564-15564 Compatibil...geReporter  D  Compat change id reported: 210923482; UID 10213; state: ENABLED
[...]

PR #13232, opened 2024-07-05
Set content length to -1 for InputStream request bodies from the nextcloud/Android-SingleSignOn library

dipshit-enterprises:
This prevents large request bodies submitted through the nextcloud/Android-SingleSignOn library from crashing the app, fixing issue #13222

### Details:
The Apache HTTP Library used will buffer the entire request body if the content length is not defined in the InputStreamRequestEntity. This causes the app to run out of memory for large uploads as it tries to buffer an arbitrarily large byte array. 

Setting the content length to -1 tells the library that the InputStream is of arbitrary size and it should not try to determine the content length itself.

### Other problems:
- A better solution would be allowing the content length to be specified in the NextcloudRequest object
- Uploading large files can take the server a while to finish writing, and there seems to be no way to specify a read timeout in the NextcloudRequest object, leading to a socket timeout that can't be avoided

Unfortunately, due to the usage of object streams, doing any of these would require `performNextcloudRequestAndBodyStreamV3()`. If such a method were to be made, I think it would be wise to try and serialize/de-serialize the NextcloudRequest in a way that's more forwards/backwards compatible.

I don't know how to make the unit tests here since testing requires a server. Here is what I used to test the fix from my own app, if it helps:

```java
    private static class RandomStream extends InputStream {
        private long remaining;
        private final Random random;

        public RandomStream(long length) {
            random = new Random();
            remaining = length;
        }

        @Override
        public int read() {
            return  remaining-- > 0 ? random.nextInt(8) : -1;
        }

        @Override
        public int available() {
            return remaining > 0 ? (int) Math.min(65535, remaining) : -1;
        }
    }

    @Test
    public void testLargeUpload() throws Exception {
        NextcloudAPI nextcloudAPI = getTestNcApi();
        InputStream is = new RandomStream(1000000000); // 1GB
        NextcloudRequest request = new NextcloudRequest.Builder()
                .setMethod("PUT")
                .setUrl("/remote.php/dav/files/user/testing/large%20file%201.bin")
                .setRequestBodyAsStream(is)
                .build();

        Response response = nextcloudAPI.performNetworkRequestV2(request);
        response.getBody().close();
        Log.i(TAG, "Request completed");
        assert is.available() == -1;
    }
```

### DIFF
--- a/app/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
+++ b/app/src/main/java/com/nextcloud/android/sso/InputStreamBinder.java
@@ -237,7 +237,7 @@ public class InputStreamBinder extends IInputStreamService.Stub {
             case "POST":
                 method = new PostMethod(requestUrl);
                 if (requestBodyInputStream != null) {
-                    RequestEntity requestEntity = new InputStreamRequestEntity(requestBodyInputStream);
+                    RequestEntity requestEntity = new InputStreamRequestEntity(requestBodyInputStream, -1);
                     ((PostMethod) method).setRequestEntity(requestEntity);
                 } else if (request.getRequestBody() != null) {
                     StringRequestEntity requestEntity = new StringRequestEntity(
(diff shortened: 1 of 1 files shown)


alperozturk96 (approved):
Thanks for your contribution 💯 

Tested, working.

alperozturk96:
@tobiasKaminsky @AndyScherzinger  Can you merge it? Looks like GitHub actions stuck again.

PR #13257, opened 2024-07-10
Fix auto upload not working when "Also upload existing files" not checked

### LINKED ISSUE #13251: When a folder is set for auto upload with "upload existing files" unchecked, background worker is killed and no auto uploads occurs
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Set a folder to auto upload, with "Also upload existing files" unchecked
2. Observe that auto upload doesn't work, with this entry in logs : `File-sync kill worker since syncedFolder (<id>) is not enabled!`

### Expected behaviour

Auto upload should work

### Actual behaviour

Auto upload worker is killed

### Android version

12

### Device brand and model

Oneplus 7t

### Stock or custom OS?

Stock

20240704 (f-droid)

### Nextcloud server version

29.0.3

### Using a reverse proxy?

Yes

### Android logs

_No response_

### Server error logs

_No response_

### Additional information

I think (based on a quick look to the code, I might be wrong) the issue comes from the check done [here](https://github.com/nextcloud/android/blob/master/app/src/main/java/com/nextcloud/client/jobs/FilesSyncWork.kt#L162-L165):

```kotlin
        if (!setSyncedFolder(syncedFolderID)) {
            Log_OC.d(TAG, "File-sync kill worker since syncedFolder ($syncedFolderID) is not enabled!")
            return true
        }
```

In [`setSyncedFolder`](https://github.com/nextcloud/android/blob/master/app/src/main/java/com/nextcloud/client/jobs/FilesSyncWork.kt#L129-L136):

```kotlin
    private fun setSyncedFolder(syncedFolderID: Long): Boolean {
        val syncedFolderTmp = syncedFolderProvider.getSyncedFolderByID(syncedFolderID)
        if (syncedFolderTmp == null || !syncedFolderTmp.isEnabled || !syncedFolderTmp.isExisting) {
            return false
        }
        syncedFolder = syncedFolderTmp
        return true
    }
```

we can see that if `syncedFolderTmp.isExisting` is false, then false is returned.

But in [`SyncedFolder`](https://github.com/nextcloud/android/blob/master/app/src/main/java/com/owncloud/android/datamodel/SyncedFolder.java#L54), the `existing` property is documented as 

```kotlin
     * @param existing            upload existing files
```

This property does not imply that the folder should not be synced.

I can make a PR if this issue is indeed a bug, but I don't know what the original intention of checking `isExisting` was.

TibboddiT:
Don't exit early from sync background worker for a folder with "Also upload existing files" unchecked.

Fix #13251

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/FilesSyncWork.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/FilesSyncWork.kt
@@ -148,7 +148,7 @@ class FilesSyncWork(
 
     private fun setSyncedFolder(syncedFolderID: Long): Boolean {
         val syncedFolderTmp = syncedFolderProvider.getSyncedFolderByID(syncedFolderID)
-        if (syncedFolderTmp == null || !syncedFolderTmp.isEnabled || !syncedFolderTmp.isExisting) {
+        if (syncedFolderTmp == null || !syncedFolderTmp.isEnabled) {
             return false
         }
         syncedFolder = syncedFolderTmp


JonasMayerDev (approved):

alperozturk96 (approved):

TibboddiT:
Resynced with master.

@tobiasKaminsky can you provide some guidance on what should be done to have this change merged ?

alperozturk96:
@tobiasKaminsky @AndyScherzinger  Can you merge it? Looks like GitHub actions stuck again.

Issue #13078, opened 2024-06-06
Make sync worker run less frequently when auto upload folder requirements are not met 


JonasMayerDev:
The sync background worker is currently running regularly in the background and scanning the file system for changed files. As some people point out (e.g. https://github.com/nextcloud/android/issues/13008) that is sometimes not wanted when the requirements for the autoupload folder (e.g. only upload in Wi-Fi) are not met since it drains the Akku. The problem here is that the scanning should still take place to be able to have a current and up-to-date upload queue, where the changed files are listed. To produce a compromise, we could simply make the interval between file system scans larger when the requirements are not met. So for example when the mobile phone is not connected to unmetered Wi-Fi run the background worker only every 1h and not every 15 min.

Issue #13008, opened 2024-05-15
Improve power usage by not trying to do stuff in background during flight mode


SvenLieber:
### Is your feature request related to a problem? Please describe.
There is a high battery usage by Nextcloud even when I am in flight mode. When I am already on a lower battery it can happen that the phone turns off at night and I don't even have an alarm in the morning which is frustrating. I am aware that there are obvious solutions which I also list below. But I still think that there could be also a technical solution in which the Nextcloud app behaves more smart.

### Describe the solution you'd like
I am aware that I explicitly turned of the "power save check" in Nextcloud to let auto-upload work efficiently.
However, for moments when I (also explicitly) turn the phone into flight mode I would expect Nextcloud to be smart and "knowing" that there currently is no Internet and thus no reason to perform auto-uploading or other background tasks.
I am not an Android developer, so I don't know if it is technically possible to fix this issue.
But I would imagine some sort of check that Nextcloud only wakes up and tries to do uploads when the phone is *not* in flight mode.

### Describe alternatives you've considered

* Obviously having the phone charged is a workaround. I try my best to have it charged
* Charging the phone overnight. Something I don't want to do.
* Enable the power save check in the auto-upload menu every time before I go to flight mode for saving battery.
* Additionally change the launch settings for Nextcloud in the global battery settings to avoid running in background every time before I go to flight mode for saving battery.

Both technical solutions to manually turn on power saving and later torn it off again seem a bit overkill considering that this need to be done every time.

### Additional context

I am aware that in the example shown in the following screenshot there is already a low battery. Something that some people might even consider a very low battery. But still, without active apps, a battery level of around 25% should be enough to survive the night if in flight mode.

![nextcloud-power-draining-flight-mode](https://github.com/nextcloud/android/assets/3501171/e32efcf9-11c8-4323-9fda-66d5dd514d30)

joshtrichards:
Hi @SvenLieber - Since many of these issues are specific to not only specific Android OS versions, but many other variables (some which are unknown or we'd have addressed them can you share:

- app version 
- OS version
- phone make/model
- your active auto-upload configs

I completely believe you're experiencing this, but as a point of reference I do not on my Pixel 7 and I never charge it overnight either.

> Nextcloud to be smart and "knowing" that there currently is no Internet and thus no reason to perform auto-uploading or other background tasks.

There really shouldn't be anything to upload if these are just media folders.

Are there any queued up or processed items in the *Uploads* tab overnight?

[...]

Issue #12954, opened 2024-04-30
Remove foreground worker for auto upload 


JonasMayerDev:
To help with reliability, we started using a foreground worker to scan the file system for changes. This is not good practice and we should try to get rid of the foreground worker since it leads to this [issue](https://github.com/nextcloud/android/issues/12596) for example.

The goal would be to change the auto upload mechanism to make it resumable (if the worker gets stopped by the system, it should not start scanning from scratch) and reduce the probability for the worker to get killed.

## TODO
- [X] Individual Worker for every Auto Upload Folder (https://github.com/nextcloud/android/pull/13015)
- - [x] Check if restarting all uploads all the time is necessary (every worker each time)
- - [x] Add a restart all method to FileSyncHelper 
- [ ] Make FileSyncWorker restartable without potential data loss (https://github.com/nextcloud/android/pull/13262)
- [ ] https://github.com/nextcloud/android/issues/13008 (https://github.com/nextcloud/android/pull/13262)

PR #13262, opened 2024-07-10
Ensure every file is scanned even when Worker is killed

JonasMayerDev:
As [this](https://github.com/nextcloud/android/pull/13217) PR failed because it produces too much overhead, a new approach consists of just shuffling the files before scanning to prevent a bias where files "further at the back" get not scanned in large folders. It is probably not the prettiest solution but since this problem is only theoretical (no known issues) and an edge case with the new improvements to auto upload this should be enough to prevent bias.  

Also Fixes https://github.com/nextcloud/android/issues/13008 and fixes https://github.com/nextcloud/android/issues/13078
Finally finishes https://github.com/nextcloud/android/issues/12954

Additionally, this PR includes:
1. Fixes a bug where the lastScannedTimestamp is not saved to the database -> Gaining a lot of performance
2. Make sure the files are not only scanned but also uploaded
3. Increase the FileSyncWorker execution interval if battery is low or conditions are not met (no internet)

To test, please test the whole auto upload feature, since there are a lot of small changes done that could affect the whole feature.


alperozturk96 (approved):
Tested

PR #13297, opened 2024-07-19
Add SSO Activity Theme Correctness Test

alperozturk96:

**Problem**

SsoGrantPermissionActivity not used in our Files app, however if any other app using Single Sign On e.g. Notes app this activity will be used. If theme for this activity is not correct, app is crashing or dialog is not appearing at all and this causing a huge problem for other apps such as other app can't authenticate the user.

_Android – You need to use a Theme.AppCompat theme (or descendant) with this activity_ error this appears in the logs.

We may change the styles and not aware of the side effects of it. Therefore, we have to be sure are we using correct theme or not.

Writing a screenshot test for this activity is not easy due to usage of the `ComponentName callingActivity = getCallingActivity()`. If we can create MaterialAlertDialogBuilder it means theme is correct. 

**Solution**

Check existence of the MaterialAlertDialogBuilder and binding for this activity.

**How to Test?**

Set wrong theme and test will fail. Set correct theme test will pass.

**How Can I Know Theme Is Correct or Not?**

Check [this](https://www.geeksforgeeks.org/android-you-need-to-use-a-theme-appcompat-theme-or-descendant-with-this-activity/) source for more detailed information.

### DIFF
--- a/app/src/androidTest/java/com/nextcloud/client/sso/SSOActivityIT.kt
+++ b/app/src/androidTest/java/com/nextcloud/client/sso/SSOActivityIT.kt
@@ -1,0 +1,28 @@
+package com.nextcloud.client.sso
+
+import androidx.test.espresso.intent.rule.IntentsTestRule
+import com.owncloud.android.AbstractIT
+import com.owncloud.android.ui.activity.SsoGrantPermissionActivity
+import org.junit.Assert.fail
+import org.junit.Rule
+import org.junit.Test
+
+class SSOActivityIT : AbstractIT() {
+
+    @Suppress("DEPRECATION")
+    @get:Rule
+    var activityRule = IntentsTestRule(SsoGrantPermissionActivity::class.java, true, false)
+
+    @Test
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/com/nextcloud/client/sso/SSOActivityIT.kt
```diff
@@ -0,0 +1,28 @@
+package com.nextcloud.client.sso
```
AndyScherzinger, line 1: lacks a SPDX header

AndyScherzinger:
@alperozturk96 if it is a screenshot test, shouldn't it take screenshots?

alperozturk96:
> @alperozturk96 if it is a screenshot test, shouldn't it take screenshots?

It's not a screenshot test. It's just checking theme is correct or not. MaterialAlertDialogBuilder was causing the problem. If we can create MaterialAlertDialogBuilder it means theme is correct.

AndyScherzinger (approved):

PR #13308, opened 2024-07-22
Share entries are updated, if available

tobiasKaminsky:

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
@@ -629,6 +629,9 @@ public class FileDataStorageManager {
                     } else {
                         deleted = getContentResolver().delete(file_uri, where, whereArgs);
                     }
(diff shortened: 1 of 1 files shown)


alperozturk96:
The actual bug has been resolved, but the share icon disappears at the 39th second. Then, at the 49th second, the share tab appears as if nothing has been shared before. However, this UI bug is disappears after the folder is refreshed.

https://github.com/user-attachments/assets/1cc1629a-aa28-4fb2-81df-075abbd0b93c

alperozturk96 (changes requested):
Actual issue resolved but UI bugs still exists.

alperozturk96:
@tobiasKaminsky Not valid after this PR: https://github.com/nextcloud/android/pull/15988.

DAO functions will be used.

alperozturk96:
The actual issue fixed within `FileDetailsSharingProcessFragment.kt`. Can't able to reproduce the issue anymore.

Thus closing this PR to keep things organized.

PR #13310, opened 2024-07-23
persist upload timestamp

tobiasKaminsky:


alperozturk96 (approved):
Tested. It's working without any issue for me. 

However, I noticed something, but I think it can't be related with this PR. Image previewing looks different.


https://github.com/user-attachments/assets/22b1d6dd-ff33-42aa-b52d-4df31a082e6a

ZetaTom (approved):

PR #13336, opened 2024-07-29
Limit Max File Upload Count

alperozturk96:

**Problem?**

We are using String array to pass file paths then upload those files. Some users trying to upload more than 5000 files at one time and when we put that String array inside the Intent we are exceeding the max limit then app is crashing, **Transaction Too Large**

**Solution**

We can store all file paths in the database and later use the table ID to retrieve and upload the files. Once the upload is complete, we will clean the database. Alternatively, we can limit the maximum file size. This PR implements a maximum file size limit.

**Example**

example_ss

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/UploadFilesActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/UploadFilesActivity.java
@@ -487,7 +487,7 @@ public class UploadFilesActivity extends DrawerActivity implements LocalFileList
             mCurrentDialog = null;
         }
 
-        if (!hasEnoughSpaceAvailable) {
+        if (hasEnoughSpaceAvailable) {
             // return the list of files (success)
             Intent data = new Intent();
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/owncloud/android/ui/activity/UploadFilesActivity.java
```diff
@@ -486,7 +487,7 @@ public void onCheckAvailableSpaceFinish(boolean hasEnoughSpaceAvailable, String.
             mCurrentDialog = null;
         }
 
-        if (hasEnoughSpaceAvailable) {
+        if (!hasEnoughSpaceAvailable) {
```
tobiasKaminsky, line 490: This seems wrong?

alperozturk96: https://github.com/nextcloud/android/pull/13336/commits/e2caaa182ebd0aa4a69dcc921748c1a6995391a0

AndyScherzinger (approved):

PR #13440, opened 2024-08-22
BugFix - Check Account Type Before Sync Operation Execution and Instead Crashing The App Inform User

alperozturk96:

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -539,7 +539,7 @@ public abstract class DrawerActivity extends ToolbarActivity
         } else if (itemId == R.id.nav_logout) {
             mCheckedMenuItem = -1;
             MenuItem isNewMenuItemChecked = menuItem.setChecked(false);
-            Log_OC.d(TAG,"onNavigationItemClicked nav_logout setChecked " + isNewMenuItemChecked);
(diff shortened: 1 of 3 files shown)


tobiasKaminsky:
Do we really want to show this?
Anonymous network call should not happen, thus we can silently cancel it?
Maybe only with internal logging?

app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -1229,60 +1229,60 @@ public void onAccountRemovedEvent(AccountRemovedEvent event) {
...
-                if ((getCapabilities() == null || getCapabilities().getAccountName().isEmpty())
-                    && getStorageManager() != null) {
+        if (!getBaseContext().getResources().getBoolean(R.bool.show_external_links)) {
+            return;
+        }
+
+        Thread t = new Thread(() -> {
+            User user = accountManager.getUser();
+            if (user.isAnonymous()) {
+                accountManager.startAccountCreation(this);
```
tobiasKaminsky, line 1239: as discussed, please keep change as little as possible and only fail silently.
-> this method opens up AuthenticatorActivity.
As we do not know why this is happening, it might be that then the user ends up in an endless loop
- open app
- fetch external link
- anonym -> AuthActivity
- enter server url
- fetch link
- -> step 1

So please remove all, but simply fail if no account can be used to fetch credentials or external links

tobiasKaminsky (changes requested):

tobiasKaminsky (approved):

Issue #925, opened 2017-05-02
Stop music when bluetooth headset disconnects


jsilence:
### Actual behaviour
- Playing music straight out of the app is awesome!  There is just one minor nitpick: When a bluetooth headset is turned off, the playback does not stop, but it continues over the internal speaker of the mobile phone. 

### Expected behaviour
- When disconnecting the BT device, stop/pause playback.
 
### Steps to reproduce
1. Upload a sound file via nextcloud sync
2. Pair a BT headset with the phone
3. Start playback by clicking on the file in nextcloud app
4. enjoy playback
5. turn off BT headset
6. Experience mobile phone speakers blaring at you

### Environment data
Android version: 7.1

Device model: Nexus 6P

Stock or customized system:  Stock Android

Nextcloud app version: 1.4.2

Nextcloud server version: current

tobiasKaminsky:
I am unsure if this can be handled by the app.
If you are playing a music file and then pair the BT heandset, it will also switch sound from internal to BT without an user interaction, or?
So I guess the opposite way is also done via android itself..

jsilence:
Other music playing apps like Spotify or VLC work the expected way: If the BT headset is turned off, playback of music pauses rather than switching to the internal speakers.

AndyScherzinger:
@esafirm from my point of view, sure, go ahead! :+1:

Issue #11810, opened 2023-07-16
Add support for media remote controls from Bluetooth


Spheerys:
### Is your feature request related to a problem? Please describe.
When a media is played within Nextcloud, the remote controls from any bluetooth device (play, pause, next...) and not working

### Describe the solution you'd like
I want to be able to use the remote controls from my bluetooth devices

### Describe alternatives you've considered
I have to awake my phone, and manually press pause/play/next/previous buttons on the interface

Issue #7462, opened 2020-11-24
Enhance PlayerService


tobiasKaminsky:
I hopefully fixed all 
```
android.app.RemoteServiceException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{496cd9c u0 com.nextcloud.client/.media.PlayerService}
```

But if some code tries to directly stop audio player it might fail, if service was not started previously:
https://proandroiddev.com/pitfalls-of-a-foreground-service-lifecycle-59f014c6a125

So it seems that Service.startForeground is only started if there is a playable file.
https://github.com/nextcloud/android/blob/277f08b9414707bc95514911d4cd71caec96bfe8/src/main/java/com/nextcloud/client/media/Player.kt#L86

For failing tests see: https://github.com/nextcloud/android/pull/6908

bapjg:
Added info : Problem occured several times when listening to the same audio file.
After re-syncing the audio file, the problem went away.
I conclude the the file was probably corrupt at the tail end.

joshtrichards:
My take is that this was addressed in #9535 and hence can be closed, but since this issue is sort of a "note to yourself", @tobiasKaminsky can you confirm?

Issue #5037, opened 2019-12-19
Display buffer progress in music player


NilsGriebner:
### Is your feature request related to a problem? Please describe.
After selecting a mp3 file with size of ~100mb the music player pops up but nothing happens apparently. It takes about 1 minute until the music starts to play. In the meantime there is no visual feedback for a user that something is in progress. It looks like the player is stuck.

### Describe the solution you'd like
It would be great to have a visual feedback like a loading screen or something else which tells the user that buffering (or whatever is going on in background) is in progress.

PR #13467, opened 2024-08-26
Improve Media Player

parneet-guraya:
Fixes:
#925
#5037
#11810
#7462 [ `PlayerService` that used legacy media library no longer used so it makes it obsolete I guess?]

### Changes Introduced: 
- [x]  Audio files now can be streamed over network.
- [x]  Audio playback in background with nice media style notification.
- [x]  Both Audio & Video players can handle commands from Bluetooth and Wired earphones. 
- [x] Both Audio & Video players implements [MediaSession](https://developer.android.com/media/media3/session/control-playback#media-session-choice) that means any media controller can connect to it and control it. Example Google Assistant, smart watch etc
- [x] Remove dependency on service that used legacy media for audio playback. yet to remove the files in separate PR.
- [x] Support managing audio focus i.e Audio ducking, Audio focus loss. 

https://github.com/user-attachments/assets/bceaee01-5f81-44e9-9fd4-1bbdc60f3d0a

### Oneplus Android 14
Screenshot_2024-09-04-04-53-55-21

### Stock Android 12
Screenshot from 2024-09-04 05-00-34

### Stock Android 13 & 14
Screenshot from 2024-09-04 05-02-39


alperozturk96 (changes requested):
Thank you for your contribution! I've created a Git patch to eliminate the use of the "!!" operator.

[force_unwrap1.patch](https://github.com/user-attachments/files/16930464/force_unwrap1.patch)

Could you apply this patch?

Additionally, the use of `runBlocking` blocks the current thread until it completes. `BackgroundPlayerService` runs on the main thread. I tested, general flow looks okay. It might be acceptable in this case, but for `initNextcloudExoPlayer()`, would it be possible to notify the service after this function finishes?

parneet-guraya:
> Thank you for your contribution! I've created a Git patch to eliminate the use of the "!!" operator.
> 
> [force_unwrap1.patch](https://github.com/user-attachments/files/16930464/force_unwrap1.patch)
> 
> Could you apply this patch?

Sure

> 
> Additionally, the use of `runBlocking` blocks the current thread until it completes. `BackgroundPlayerService` runs on the main thread. I tested, general flow looks okay. It might be acceptable in this case, but for `initNextcloudExoPlayer()`, would it be possible to notify the service after this function finishes?

Ah, I was aware of this forgot to add the comment. I know it would block the main thread it's just that the `Nextcloud client` needs to be created on background thread. but in that time service already starts and player doesn't get initialized. It doesn't do that much of a task while creating a client so I add a run blocking so to wait for the client. 

Also, we're not the one starting the service, it is started when session token is created. So no way we can tell it to start after the client init on background. 

since controlling when service starts not in our control. Can we have some other option to inject Exoplayer without having to create client instance? tell me if you have any ideas I will see.

thanks :+1:

alperozturk96:
@parneet-guraya Could you please address the Codacy Static Code Analysis issues and apply a git patch to eliminate the use of the `!!` operator so we can proceed with the merge? Thanks for your contribution! 💯

alperozturk96 (approved):

PR #13488, opened 2024-08-29
BugFix - Set Drawer For Favorite Search Type

alperozturk96:


alperozturk96:
Not valid anymore: https://github.com/nextcloud/android/pull/13513

PR #13493, opened 2024-08-30
Enforced servers

tobiasKaminsky:
similar to "enforce domain" this allows to build an app with a dropdown menu with pre-defined servers.

### DIFF
--- a/app/src/main/res/layout/account_setup.xml
+++ b/app/src/main/res/layout/account_setup.xml
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
   ~ Nextcloud - Android Client
   ~
   ~ SPDX-FileCopyrightText: 2019 Tobias Kaminsky <tobias@kaminsky.me>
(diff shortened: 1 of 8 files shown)


app/src/main/res/layout/account_setup.xml
```diff
@@ -79,6 +80,22 @@
                 </com.google.android.material.textfield.TextInputLayout>
             </FrameLayout>
 
+            <LinearLayout
```
alperozturk96, line 83: Why do you need to wrap Spinner with LinearLayout?

```
         <Spinner
                android:id="@+id/servers_spinner"
                android:visibility="gone"
                tools:visibility="visible"
                android:foregroundTint="#FFFFFF"
                android:spinnerMode="dialog"
                android:textColor="@color/white_helper_text" />
```

This should work without LinearLayout.

alperozturk96 (changes requested):

app/src/main/res/layout/account_setup.xml
tobiasKaminsky: I first thought, that I need more elements.
will try to remove it.

alperozturk96 (approved):

PR #13501, opened 2024-09-02
Feature - Offline File Creation

alperozturk96:

**Demo**

https://github.com/user-attachments/assets/dee497d1-779d-412a-a9ca-ac45c9bda8bc


tobiasKaminsky:
This fails for me with a SQLite upgrade error.
Can you rebase it and test 
- uninstall all
- install and run master
- install and run this PR?

alperozturk96:
> This fails for me with a SQLite upgrade error. Can you rebase it and test
> 
> * uninstall all
> * install and run master
> * install and run this PR?

Previously, I thought we needed the `parentPath` column, but it turns out we don't since we already have the `parent_oc_file_id` and the `offline_operation_type` sealed class, which effectively differentiates between operations. After adding the migration from version 84 to 85, the issue was resolved.

alperozturk96:
> What I tested
> 
> * be offline
> * create folder "offline" -> not sorted correctly (sort order was newest first), but at end
> * added 5 images files -> no thumbnail, but we could show one (similar like we do on upload view)
> * after online again: needed to refresh to see that files have been uploaded
> * date is wrong?
> 
> image

I set the creation and modification date with the correct UTC timestamp (e.g., 1727343315600) for `OCFile`, but the `FileStorageUtils.fillOCFile(RemoteFile remote)` function overrides it with an invalid UTC timestamp (e.g., 15032385535000).

To create a file in the `OfflineOperationWorker`, the correct values for the creation and modification timestamps are also used. Do you have any idea why the server returns this value?

alperozturk96:
@tobiasKaminsky The sorting order may change after the upload because the stored value is being overridden by remoteFile after a successful upload.

**Fixes:**

https://github.com/user-attachments/assets/d2fd0faa-4867-485f-b735-8f7b2bfb38aa

app/schemas/com.nextcloud.client.database.NextcloudDatabase/84.json
```diff
@@ -1237,7 +1237,7 @@
       },
       {
         "tableName": "offline_operations",
-        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`_id` INTEGER PRIMARY KEY AUTOINCREMENT, `offline_operations_parent_oc_file_id` INTEGER, `offline_operations_parent_path` TEXT, `offline_operations_type` TEXT, `offline_operations_path` TEXT, `offline_operations_file_name` TEXT, `offline_operations_created_at` INTEGER)",
```
AndyScherzinger, line 1240: why is an existing DB migration being changed? This will fail on any installation where the DB is already at v84 (!)

AndyScherzinger (changes requested):

app/schemas/com.nextcloud.client.database.NextcloudDatabase/84.json
alperozturk96: We didn't release any version contains this feature. However, master have that unnecessary field. Nevertheless, Migration84to85 should handle it, I will update the code, thanks.

tobiasKaminsky (approved):

PR #13513, opened 2024-09-04
BugFix - Unify Drawer Menu Item Handling

alperozturk96:

- Stores the menu item value in memory, eliminating the need for argument passing and state management when restoring the menu item.
- Calls `setDrawerMenuItemChecked()` only when necessary, ensuring a single source of truth for the menu item state.
- Utilizes `DrawerActivity` to manage the menu item, with sub-activities modifying the item as needed.
- Removes redundant logic related to menu item handling.

**How to Test?**

1. Navigate through multiple tabs, put the app in the background, then bring it back to the foreground and check the highlighted drawer menu.
2. Navigate to All Files → Favorites → Uploads → Press Back, and verify the highlighted drawer menu.
3. Navigate to All Files → Uploads and check the highlighted drawer menu.
4. Navigate to All Files → Community and check the highlighted drawer menu.
5. Navigate through multiple tabs, then rotate the device and check the highlighted drawer menu.
6. Navigate through multiple tabs, lock the screen, then unlock it and check the highlighted drawer menu.
7. Try previous scenarios with passcode.


ZetaTom (dismissed):
When selecting *Personal Files* from the *Activities* tab, *All files* will be selected instead.

On a side note, I don't think we should be using string matching to determine which tab is currently open.

alperozturk96:
> When selecting _Personal Files_ from the _Activities_ tab, _All files_ will be selected instead.
> 
> On a side note, I don't think we should be using string matching to determine which tab is currently open.

This is a workaround solution and @tobiasKaminsky we decided to fix it this way until we have better codebase and structure.

Obviously the whole structure, navigation flow is wrong and broken. That's not the case in here. Proper fix will be fixing navigation flow and have single source of truth for simple tasks like highlighting drawer menu item. That fix will consume much more time, and I can't decide to implement that fix by myself. Because of the priorities, time etc.

If you have the proper fix with much less code change, please open PR.

**Current codebase behaviors:** 

Currently, some menu items, such as “Settings,” in the drawer menu don’t extend from DrawerActivity, which is causing navigation issues (different action bar, different back button behavior when it compared to other activities). 

There are multiple methods being used to highlight the drawer menu, leading to some inconsistencies. 

Two menu item variable was responsible for holding menu item value.

setDrawerMenuItemChecked() gets called multiple times from different files and some value of it overriding correct value and applying wrong one.

ZetaTom (approved):
This change fixes the issue and works fine on my end.

But we should keep in mind that this is just a workaround. We will need to implement a proper at some point in the future, as @alperozturk96 mentioned.

PR #13517, opened 2024-09-05
Fix ScreenShot Tests After Offline Folder Creation Operations

alperozturk96:

**Conflict File Dialog Encrypted Folder (Before)**

Screenshot 2024-09-06 at 16 06 47

**Known Issues**

Previously, the server file name was used for the title, and to get the folder name from the remote path of the server file, a Java `File` instance was created, then the absolute path was retrieved. 

We can achieve the same result directly using the `decryptedRemotePath`. However, for encrypted folders, neither the old nor the new approach will work. We can address this issue in a separate PR.


alperozturk96:
continuous-integration/drone/pr CI fixed via https://github.com/nextcloud/android/pull/13541

tobiasKaminsky (approved):

tobiasKaminsky:
Just learned that if the last commit is by a bot no gh action will be triggered.
@alperozturk96 can you squash two last commits, so they are authored by you.

PR #13519, opened 2024-09-05
BugFix - OCShare Conversion

alperozturk96:

**Problem?**

- File size, thumbnail not visible in shared tab.
- Files are not filtered correctly for shared tab

**Solution**

- Path can be used for fetching OCFile rather than converting OCShare to OCFile manually


ZetaTom (changes requested):
Unfortunately this change breaks the *Shared* tab in some use cases.

If the user doesn't have any shared files in the root directory, the tab will be stuck at *Loading...*. If there is a shared file in the root folder, this will indeed be displayed correctly. However, any files contained within subdirectories will not be displayed.

alperozturk96:
https://github.com/nextcloud/server/issues/4456 API needs to be updated. This solution will solve the problem for only existing Files in device. When the device doesn't initially have an OCFile, the path can't be used to retrieve it. 

e.g. Shared file can be in a/b/c/d/a.txt --> to display a.txt in shared tab device must have OCFile. In first launch, app doesn't aware of that file and that file can be accessible when user enters the exact same path.

PR #13533, opened 2024-09-09
Flag FLAG_SECURE should only be active when in passcode screen

parneet-guraya:
From talk [issue](https://github.com/nextcloud/talk-android/issues/4164#issue-2510459429) 

FLAG_SECURE, should only be active in passcode screen unlike (current) master. This cause user unable to take screenshots or screen recording. If user is using passcode/device-credentials.


alperozturk96 (approved):
Tested working

Issue #8055, opened 2021-02-25
"No resized image available. Download full image?" -> always yes


tobiasKaminsky:
If an user scrolls through a lot of unsupported images, one has to click on "yes" to download every image where there is no preview available.

Ref: https://github.com/nextcloud/android/issues/8017#issuecomment-784164483

Issue #8017, opened 2021-02-21
WebP support


L-as:
### Is your feature request related to a problem? Please describe.
Whenever I open a WebP image, it asks me "No resized image available. Download full image?" 2~3 times, but even if I say "yes" it doesn't actually show the image, it just downloads it without showing it. You can view it if you go back to the directory tree and click on it again, but it's quite a PITA.

### Describe the solution you'd like
Optimally WebP would have the same level of support as JPEG and such.

### Describe alternatives you've considered
I took a look around, and I can see that normally images are resized before being sent (which also explains why I see no thumbnails), but if supporting resizing is non-trivial, you could have it automatically ephemerally download the full image and show that. Perhaps this could be a setting in the app, since it could be useful for JPEG and such too, if you don't care about how much data you're using and just want the highest quality image.

A personal solution for me would be to convert all my images to JPEG, but I don't have the storage space for that...

tobiasKaminsky:
@nextcloud/server-triage for the resized image feature we rely on server capability to show wepg images, is this somehow supported?

On my Android 10 I can see downloaded webp image as like any other image.
What Android version do you use?

L-as:
I am on Android 6, though I **can** see the downloaded webp image, it's just buggy in the sense that clicking "yes" when it gives the pop-up doesn't actually show it even though it downloads it. I need to exit the viewer and try to view the image again for it to actually show it.

tobiasKaminsky:
This also works for me.
The image previewer is loading next two upcoming images in background.
If one of those cannot be displayed, this "do you want to download" question is also shown.
So if it this the case and you click on yes, then you do not download the current image, but one of the next two.

You can check this, if you copy/move a webp image into a new folder, only containing this file.
If this then still fails, it is a new bug, otherwise it is the old one.

L-as:
You seem to be correct.
I tried with a single webp in a folder, and it asked me if I wanted the full image, which after accepting caused it to download and show it correctly.

I tried my folder with a lot of pictures again, and it does seem that it shows me the pop-up once for each picture it wants to download, and I could view the next (not the current) picture after clicking "yes" to the popup, but after swiping a bit around it crashed...

Would it be hard to have a setting where you can make it "yes" by default for the pop-up, so that one can view many pictures smoothly without rapidly accepting the pop-ups?

Thanks

tobiasKaminsky:
Thanks for confirmation.
So there are two issues:
[...]

PR #13534, opened 2024-09-09
Always download if image is unsupported

parneet-guraya:
Fix: #8055

- Image will always be downloaded if unsupported.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/preview/PreviewImageFragment.kt
+++ b/app/src/main/java/com/owncloud/android/ui/preview/PreviewImageFragment.kt
@@ -721,28 +721,14 @@ class PreviewImageFragment : FileFragment(), Injectable {
         binding.emptyListProgress.visibility = View.GONE
     }
 
-    fun setErrorPreviewMessage() {
+    fun handleUnsupportedImage() {
         try {
-            if (activity != null) {
+            (activity as? PreviewImageActivity)?.requestForDownload(file) ?: context?.let {
                 Snackbar.make(
                     binding.emptyListView,
-                    R.string.resized_image_not_possible_download,
+                    resources.getString(R.string.could_not_download_image),
                     Snackbar.LENGTH_INDEFINITE
-                )
-                    .setAction(
-                        R.string.common_yes
-                    ) { v: View? ->
-                        val activity = activity as PreviewImageActivity?
-                        if (activity != null) {
-                            activity.requestForDownload(file)
-                        } else if (context != null) {
-                            Snackbar.make(
-                                binding.emptyListView,
-                                resources.getString(R.string.could_not_download_image),
-                                Snackbar.LENGTH_INDEFINITE
-                            ).show()
-                        }
-                    }.show()
+                ).show()
             }
         } catch (e: IllegalArgumentException) {
             Log_OC.d(TAG, e.message)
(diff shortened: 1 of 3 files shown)


ZetaTom (commented):
Thank you for this contribution.

I do believe that a simple toggle in the preferences is a suitable solution for this problem. However, I would like to involve @nextcloud/designers in this discussion.

In my testing everything worked as expected. I've made suggestions for some minor improvements.

jancborchardt (dismissed):
To refer to our design guidelines from https://nextcloud.com/design/
> **Software should get out of the way.**
> Do things automatically instead of offering configuration options.

So in this case, what is the option supposed to do, and what is the default:
- If it’s off by default, only a tiny fraction of people will enable it anyway
- If it’s on by default, why don’t we just make it so, without the option?

@parneet-guraya what do you think?

EDIT: Also reading the original issue at https://github.com/nextcloud/android/issues/8017, the root cause does not seem to be best solved by introducing a setting, but rather by just doing it. :)

parneet-guraya:
@jancborchardt hi :wave: , 

> If it’s on by default, why don’t we just make it so, without the option?

By default we ask for the resized image from the backend instead of full resolution (and only fallback to download full resolution if not supported), to save user's bandwidth.  I introduced this preference just to keep the current default. 

But, if we keep it `always on` as only option that'd mean unsupported images will always be downloaded. If that's something we are okay with (maybe @tobiasKaminsky  can tell) then yes, we can keep this option on by default.

Let me know what you think.

jancborchardt:
@parneet-guraya thanks for the explanation. :)

> But, if we keep it `always on` as only option that'd mean unsupported images will always be downloaded. 

Yes, I would say that is fine – because people want to view the images. They don’t even know that we normally show only a resized version. So we should not expose implementation details to them and just load it.

AndyScherzinger:
I also agree with Jan's comment: **always on** 👍

parneet-guraya:
`always on` it is :+1:

parneet-guraya:
@ZetaTom , I pushed the change and as discussed in the comments, unsupported image will always be downloaded.

app/src/main/java/com/owncloud/android/ui/preview/PreviewImageFragment.kt
```diff
@@ -721,28 +721,19 @@ class PreviewImageFragment : FileFragment(), Injectable {
         binding.emptyListProgress.visibility = View.GONE
     }
 
-    fun setErrorPreviewMessage() {
+    fun handleUnsupportedImage() {
```
ZetaTom, line 724: To avoid the redundant check if `activity` is null, I'd suggest the following:

```kotlin
fun handleUnsupportedImage() {
    try {
        (activity as? PreviewImageActivity)?.requestForDownload(file) ?: context?.let {
            Snackbar.make(
                binding.emptyListView,
                resources.getString(R.string.could_not_download_image),
                Snackbar.LENGTH_INDEFINITE
            ).show()
        }
    } catch (e: IllegalArgumentException) {
        Log_OC.d(TAG, e.message)
    }
}
```

ZetaTom (commented):
Thank you for implementing the suggested changes.

Everything worked as expected.

ZetaTom (approved):
During my testing, these changes worked as expected. However, the preview of the actual image didn't work until I re-opened the file. As this behaves the same for me on master, I think that we can go ahead and merge anyway and address the issue of not loading at another date.

alperozturk96 (approved):

alperozturk96:
@jancborchardt Could you review it again? If it's okay, can we merge it?

PR #13541, opened 2024-09-09
BugFix - Dont Set Connectivity Cache True When Base Server Address Is Empty & Fix Screenshot Tests

alperozturk96:

**Problem**

When the app is installed for the first time and a file upload is attempted during the initial launch, the upload fails because the cached value for `isInternetWalled()` is set to true. 

UploadFileOperation

```
// check that internet is not behind walled garden
if (!connectivityService.getConnectivity().isConnected() || connectivityService.isInternetWalled()) {
     remoteOperationResult = new RemoteOperationResult(ResultCode.NO_NETWORK_CONNECTION);
}
```

This check causes the upload error during the first launch.


alperozturk96:
Screenshots tests fixed (except ContactsBackupIT) via https://github.com/nextcloud/android/pull/13517

ZetaTom (approved):
Worked as expected during my tests.

PR #13546, opened 2024-09-10
TextWebView: allow to upload multiple attachements

tobiasKaminsky:


juliushaertl:
Since Stephan pinged me on this, cannot really judge on the code but Arthur gave it a test run and confirmed that it works

alperozturk96 (approved):
When I try to open an .md file, the TextEditorWebView (which inherits from EditorWebView) gets called. I add multiple attachments to the .md file, they are added successfully, which triggers the else block without any issues.

PR #13551, opened 2024-09-11
Do not allow multiple instances of LoadingDialog

ZetaTom:
This change reintroduces more robust state handling when creating new loading dialogues. Previously it was possible to create multiple instances by issuing successive calls to `showLoadingDialog()`. Because of this one or more instances could end up stuck on screen, requiring a restart of the app.

**Testing these changes:**
1. Open Nextcloud App
2. Select two or more files
3. Download said files
4. Observe the *Wait a moment...* dialogue

With this change, the dialogue should eventually disappear.

---


alperozturk96 (approved):

alperozturk96:
Tested working

PR #13599, opened 2024-09-23
BugFix - Filter Root

alperozturk96:

**Problem?**

When users navigate to the Shared or Favorite files, the back button action does not display the correct file list.

**What does this PR?**

- Filters OCFileListAdapter according to the selected drawer menu item

**How to Test?**

1. Mark some files as Favorite or share some files.
2. Select Favorites or Shared Files from the drawer menu.
3. Navigate to a subdirectory.
4. Press the back button.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
@@ -916,7 +916,7 @@ public class OCFileListFragment extends ExtendedListFragment implements
         }
     }
 
-    private boolean shouldNavigateNormally(OCFile topParent) {
+    private boolean shouldNavigateWithoutFilter(OCFile topParent) {
         int menuItemId = DrawerActivity.menuItemId;
         return (menuItemId != R.id.nav_shared && menuItemId != R.id.nav_favorites) ||
             (menuItemId == R.id.nav_shared && topParent != null && topParent.isShared()) ||
@@ -928,7 +928,7 @@ public class OCFileListFragment extends ExtendedListFragment implements
         return menuItemId == R.id.nav_shared || menuItemId == R.id.nav_favorites;
     }
 
-    private Pair<Integer, OCFile> navigateNormally(FileDataStorageManager storageManager) {
+    private Pair<Integer, OCFile> getPreviousFileWithoutFilter(FileDataStorageManager storageManager) {
         int moveCount = 0;
         OCFile parentDir = null;
         String parentPath = null;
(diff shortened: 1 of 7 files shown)


tobiasKaminsky:
Hm. This looks for me like a strange workaround/fix.
If you are back at root folder, then Favorite/Shared search should be applied, so that no manual filtering will be done.

Favorites is searching across all folders, so e.g. those are favorited
- /test
- /a/1.pdf

If you first go to "favorites" you will see both.
But if with this PR, you go "back" from "test", you would only see "test" folder, but not file "1.pdf".

alperozturk96:
> Hm. This looks for me like a strange workaround/fix. If you are back at root folder, then Favorite/Shared search should be applied, so that no manual filtering will be done.
> 
> Favorites is searching across all folders, so e.g. those are favorited
> 
> * /test
> * /a/1.pdf
> 
> If you first go to "favorites" you will see both. But if with this PR, you go "back" from "test", you would only see "test" folder, but not file "1.pdf".

I guess [this](https://github.com/nextcloud/android/pull/13599/commits/a3521e11693a0764585914a6b84b66ab61174149) is fixing the problem you described?


https://github.com/user-attachments/assets/e6e57051-6f3b-405e-b762-f278d0833d1b

tobiasKaminsky:
This works now ✔

Can you check this:
- have a folder /a/b, where "b" is favorited
- delete database (adb shell pm clear com.nextcloud.client)
- start app
- go to favorites
- click on "b"
--> endless loading for me

But when I click back, I can directly see content of "/a"

tobiasKaminsky (dismissed):
to check

app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -915,50 +919,123 @@ private void updateSortAndGridMenuItems() {
...
+
+    /**
+     * Call this, when the user presses the up button.
+     * <p>
+     * Tries to move up the current folder one level. If the parent folder was removed from the database, it continues
+     * browsing up until finding an existing folders.
+     * <p>
+     * return       Count of folder levels browsed up.
+     */
+    public int onBrowseUp() {
```
alperozturk96, line 1018: Previously, onBrowseUp() was running on the main thread, which caused lag. This PR adds more I/O operations to the function due to previous file changes based on the filter type. Therefore, we need to run onBrowseUp() on the I/O thread and update the UI on the main thread.

Note: The endless loading issue is not part of this PR. It can also occasionally be observed on the stable or master branch. During my tests, I encountered it only once.

@tobiasKaminsky

tobiasKaminsky (approved):

Issue #13051, opened 2024-05-27
Re-order and group drawer items?


joshtrichards:
To better align with the evolving Web UI should we consider tidying up the [navigation drawer](https://m3.material.io/components/navigation-drawer/overview) a bit?

- splitting up the Files related items from the other ones?
- adding visual dividers between the logically separate groups of items
- related:
  - we could maybe also move Activities up to the ecosystem app menu that sits at the top, since it is, in fact an app (and that's consistent with the web UI approach
  - Media isn't exactly the Photos app so it's debatable about moving it there, but that's the only other one that is a candidate
  - Incidentally, maybe there is some way to revisit the design of the ecosystem apps list to make them more compact/list more of them
- where should *Notifications* go?

# e.g.

---

## Talk | Notes | More

- All Files
- Personal Files
- Favorites
- Recents <-- (future) 
- Shares <-- "Shared->Shares" w/ own section 
- Tags <-- (future)
- External Storage <-- (future)
----
- Activities
- Media
- Uploads
- On Device
---

- Deleted files
- 559 MB used
- Settings

---

### Reference: Current Android:

![image](https://github.com/nextcloud/android/assets/1731941/7cdd1f39-f5a9-418b-a662-75b2819e775e)

### Reference: Server v29 Web Client UI:

![image](https://github.com/nextcloud/android/assets/1731941/c1691dfc-9573-4fbd-9907-8289b3bd109c)

AndyScherzinger:
looping in @nextcloud/designers for input and feedback

jancborchardt:
@joshtrichards liking your suggestions a lot! :) 
- I tend to agree with moving Media up to the integrations section as if it’s Photos
- Activity I’m not sure about, since it’s similar to "Recents" (which we don’t have yet) and not used so often so doesn’t need to be that present
- I like the organization / splitting via separator.
- I agree that Notifications is too difficult to reach. If there are notifications, can’t we show a notification bell in the header, e.g. left of the avatar in the search field?

@tobiasKaminsky what do you think?

tobiasKaminsky:
![2024-09-24-084530](https://github.com/user-attachments/assets/7151528f-8a0e-404e-9a7b-c68abc30f5f6)

What do you think? There are some apps, e.g. groupfolder which are not always enabled. I tried to keep them either in first group (files) or second (rest).

Notification:
When we receive push notifications, then we can show a red bell/dot somewhere.
But then the user already sees it on OS level.

This notification section serves mainly two things
[...]

PR #13609, opened 2024-09-24
Re-sort drawer entries

tobiasKaminsky:
For https://github.com/nextcloud/android/issues/13051


tobiasKaminsky:
@marcoambrosini @Hyeyoung346 in above screenshot link you can see diff between new and old.
Are you fine with it?

PR #13624, opened 2024-09-25
Simplify FileDisplayActivity onBackPressed 

alperozturk96:

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -1019,27 +1019,22 @@ public class FileDisplayActivity extends FileActivity
     @SuppressFBWarnings("ITC_INHERITANCE_TYPE_CHECKING")
     @Override
     public void onBackPressed() {
-        final boolean isDrawerOpen = isDrawerOpen();
-        if (isDrawerOpen) {
+        if (isDrawerOpen()) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -997,24 +997,22 @@ private Boolean isRootDirectory() {
...
-        final boolean isSearchOpen = isSearchOpen();
-
-        final Fragment leftFragment = getLeftFragment();
+        if (isDrawerOpen()) {
+            super.onBackPressed();
+            return;
+        }
 
-        if (isSearchOpen) {
+        if (isSearchOpen()) {
```
tobiasKaminsky, line 1005: You are changing here the logic.
Previous: if searchOpen  else if drawerOpen
Now: if drawer open else if searchOpen.

Is this intended? 
A quick test still shows correct/expected behaviour.

tobiasKaminsky (approved):

app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
alperozturk96: https://github.com/nextcloud/android/pull/13624/commits/b057344e46d2254392553a96156e1d60c02371e8

AndyScherzinger (approved):

PR #13664, opened 2024-10-01
Update Dependency & Fix UI - Assistant Screen

alperozturk96:

**Changes**

- Fix wrong color usages
- Update code according to the latest dependency

**Demo**

https://github.com/user-attachments/assets/fc932ee2-401e-4f15-b79e-c04f759d20d6


AndyScherzinger:
Looks good to me, just as a general comment, we should be using the outlined style on input elements, as that is the standard we are aiming for

PR #13680, opened 2024-10-02
Do Not Throw An Exception in SyncOperation

alperozturk96:

I returned an ACCOUNT_EXCEPTION for testing purposes, and as can be seen in the video, only the capabilities cannot be updated.

https://github.com/user-attachments/assets/b279536f-6c9e-4dd5-8a32-30f4728b345a

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/common/SyncOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/common/SyncOperation.java
@@ -46,8 +46,7 @@ public abstract class SyncOperation extends RemoteOperation {
      */
     public RemoteOperationResult execute(Context context) {
         if (storageManager.getUser().isAnonymous()) {
-            throw new IllegalArgumentException("Trying to execute a sync operation with a " +
-                                                   "storage manager for an anonymous account");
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Accordin to exception message this is to prevent running a sync operation with a file data storage manager on a wrong/null account.
Have you tested such an operation?

alperozturk96:
> Accordin to exception message this is to prevent running a sync operation with a file data storage manager on a wrong/null account. Have you tested such an operation?

I'm not sure what you mean. I tested switching accounts, and the correct FileDataStorageManager was used.

https://github.com/user-attachments/assets/5f045bfd-b981-4a88-a253-b3986b9ffbd4

tobiasKaminsky:
The message you removed:
"Trying to execute a sync operation with a storage manager for an anonymous account"

So you have you tried it with a non-existing account, e.g. during first setup and when removing last account?

alperozturk96:
> Accordin to exception message this is to prevent running a sync operation with a file data storage manager on a wrong/null account. Have you tested such an operation?

I logged in and removed the same account. After pressing the back button, the login screen appeared again. Later, I logged in with the same account and created a folder with two files in it. I did not encounter any problems.

tobiasKaminsky (approved):

PR #13683, opened 2024-10-02
Only Download Necessary Files

alperozturk96:

**Problem**

Downloaded files are re-downloading, and notifications appear for files that have already been downloaded. The sync should only trigger downloads for new or editable files, as media files cannot be changed. 

The two-way sync triggers the FileDownloadWorker every 15 minutes. If a folder contains multiple files, even if all files are already downloaded, the worker will still show notifications for each file and attempt to re-download them again.

**Note:** Images can be edited, but any edited image will create a copy rather than modifying the original file. Thus, no need to download original image.

**Changes**

- Download only files that haven't been downloaded and aren't media files.
- Return early if requestDownloads is empty.

**Demo**

https://github.com/user-attachments/assets/88505d08-654b-4a4b-821c-0c622970dec8

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/download/FileDownloadWorker.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/download/FileDownloadWorker.kt
@@ -237,7 +237,7 @@ class FileDownloadWorker(
         } else {
             listOf(file)
         }.filterNot {
-            it.isDown && (MimeTypeUtil.isImageOrVideo(it) || MimeTypeUtil.isAudio(it))
+            it.isDown && MimeTypeUtil.isMedia(it.mimeType)
         }
     }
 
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Can you explain reason behind filtering media files?

This worker is also used when exporting files (FilesExportWork.kt), so this would then not work if you want to export a pdf file?

tobiasKaminsky (changes requested):
see comment

alperozturk96:
> Can you explain reason behind filtering media files?
> 
> This worker is also used when exporting files (FilesExportWork.kt), so this would then not work if you want to export a pdf file?


Screenshot 2024-10-08 at 11 49 09

FilesExportWork only triggers the FileDownloadWorker when a file hasn’t been downloaded, so the filter introduced in this PR won’t be applied.

Here’s a demo of FilesExportWork with already downloaded files.

https://github.com/user-attachments/assets/042674cc-aace-48aa-b353-e0dd31c991f3

You can check the PR description for more details on the purpose of this PR.

fegauthier:
I'm eagerly waiting for this PR. I have a Google Pixel 8, and files are constantly resyncing every 15 minutes. It drains the battery significantly and uses bandwidth unnecessarily. Do you guys know when it will be merged?

Thanks.

alperozturk96:
Not Valid after: https://github.com/nextcloud/android/pull/13897

PR #13695, opened 2024-10-04
Remove fido since we do not need it with login flow v2

tobiasKaminsky:


alperozturk96 (approved):
Great 💪 

Tested and working with a fresh install as well as with a re-launch of an existing logged-in account.

PR #13746, opened 2024-10-09
Fix Crashes When Opening File Details from Media and Other Tabs

alperozturk96:

**How to reproduce?**

Open an image from the Media tab and attempt to navigate to the file details.

**Known Issues:**

File details do not open from the Media tab; the same issue exists on the master branch.

**Changes:**

Fixed crashes when the user tries to open file details from other tabs and the Media tab.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/fragment/FileDetailFragment.java
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/FileDetailFragment.java
@@ -351,11 +351,12 @@ public class FileDetailFragment extends FileFragment implements OnClickListener,
             }
         });
 
-        binding.tabLayout.post(() -> {
-            TabLayout.Tab tab1 = binding.tabLayout.getTabAt(activeTab);
-            if (tab1 == null) return;
-            tab1.select();
-        });
+        // FIXME file detail not opening from Media tab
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):

app/src/main/java/com/owncloud/android/ui/fragment/FileDetailFragment.java
```diff
@@ -351,11 +351,12 @@ public void onTabReselected(TabLayout.Tab tab) {
             }
         });
 
-        binding.tabLayout.post(() -> {
-            TabLayout.Tab tab1 = binding.tabLayout.getTabAt(activeTab);
-            if (tab1 == null) return;
-            tab1.select();
-        });
+        // FIXME file detail not opening from Media tab
```
tobiasKaminsky, line 354: @alperozturk96 please create an issue for it and fix this.

alperozturk96: [Issue](https://github.com/nextcloud/android/issues/13748) created. I tried this morning a lot but somehow FileDetailFragment gets destroyed immediately, I don't know why.

PR #13755, opened 2024-10-10
Warn when file extension is renamed

tobiasKaminsky:
image

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/dialog/RenameFileDialogFragment.kt
+++ b/app/src/main/java/com/owncloud/android/ui/dialog/RenameFileDialogFragment.kt
@@ -197,8 +197,8 @@ class RenameFileDialogFragment : DialogFragment(), DialogInterface.OnClickListen
     override fun afterTextChanged(s: Editable) = Unit
 
     private fun checkExtensionRenamed(newFileName: String): Boolean {
-        mTargetFile?.fileName?.let {
-            val previousExtension = File(it).extension
+        mTargetFile?.fileName?.let { previousFileName ->
+            val previousExtension = File(previousFileName).extension
             val newExtension = File(newFileName).extension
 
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1247,4 +1247,5 @@
     <string name="sync">Sync</string>
     <string name="please_select_a_server">Please select a server…</string>
     <string name="notification_icon_description">Unread notifications exist</string>
+    <string name="warn_rename_extension">Renaming extension might make the file unusable</string>
```
ZetaTom, line 1250: Personally, I would suggest the following wording.

```suggestion
    <string name="warn_rename_extension">Changing the extension might cause this file to open in a different application</string>
```

The macOS Finder has a similar dialogue when renaming files:

> **Are you sure you want to change the extension from ".ext1" to ".ext2"?**
>
> If you make this change, your document may open in a different application.

app/src/main/java/com/owncloud/android/ui/dialog/RenameFileDialogFragment.kt
```diff
@@ -186,6 +191,17 @@ class RenameFileDialogFragment : DialogFragment(), DialogInterface.OnClickListen
 
     override fun afterTextChanged(s: Editable) = Unit
 
+    private fun checkExtensionRenamed(newFileName: String): Boolean {
+        mTargetFile?.fileName?.let {
+            val previousExtension = File(it).extension
```
ZetaTom, line 196:
```suggestion
        mTargetFile?.fileName?.let { previousFileName ->
            val previousExtension = File(previousFileName).extension
```

ZetaTom (commented):
Everything worked as expected during my tests. I've added some minor polishing suggestions.

PR #13847, opened 2024-10-22
2way sync: sync only on wifi

tobiasKaminsky:

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/InternalTwoWaySyncWork.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/InternalTwoWaySyncWork.kt
@@ -35,7 +35,9 @@ class InternalTwoWaySyncWork(
         var result = true
 
         if (powerManagementService.isPowerSavingEnabled ||
-            !connectivityService.isConnected || connectivityService.isInternetWalled
+            !connectivityService.isConnected ||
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nextcloud/client/jobs/InternalTwoWaySyncWork.kt
```diff
@@ -35,7 +35,9 @@ class InternalTwoWaySyncWork(
         var result = true
 
         if (powerManagementService.isPowerSavingEnabled ||
-            !connectivityService.isConnected || connectivityService.isInternetWalled
+            !connectivityService.isConnected ||
+            connectivityService.isInternetWalled ||
+            !connectivityService.connectivity.isWifi
```
bobemoe, line 40: Is this going to prevent sync on Ethernet?

rugk: Yeah should maybe check for metered connections, better? (Also for some wifi networks you don't want it to sync)

rugk: Though maybe open a new issue for that to not get it lost? or can @tobiasKaminsky et al bring some insight in here?

PR #13859, opened 2024-10-22
Added SHA256 & SHA1 application certificate fingerprints to README

DanyPM:
Add the app certificate fingerprint to the README file, enabling apps and users to check the integrity of apk installed from github.

The main purpose is to enable subsequent verification of the app through [AppVerifier](https://github.com/soupslurpr/AppVerifier), including adding the fingerprint to its internal database.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,11 @@ height="80">](https://play.google.com/store/apps/details?id=com.nextcloud.client
 alt="Get it on F-Droid"
 height="80">](https://f-droid.org/packages/com.nextcloud.client/)
 
+Signing certificate fingerprint to [verify](https://developer.android.com/studio/command-line/apksigner#usage-verify) the APK:
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,6 +13,12 @@ height="80">](https://play.google.com/store/apps/details?id=com.nextcloud.client
 alt="Get it on F-Droid"
 height="80">](https://f-droid.org/packages/com.nextcloud.client/)
 
+Signing certificate fingerprint to [verify](https://developer.android.com/studio/command-line/apksigner#usage-verify) the APK:
```
tobiasKaminsky, line 16: Can you add a note that this is only true for 
- Gplay (shipped via Google Play Store)
- assets found in https://github.com/nextcloud/android/releases

It is not working for Fdroid, as they sign on their own.

tobiasKaminsky:
Thank you for your contribution :tada:

README.md
tobiasKaminsky: @DanyPM may I do this for you?

PR #13897, opened 2024-10-29
BugFix - Unnecessary Sync

alperozturk96:

This regression occurred during the[ conversion from service to worker.](https://github.com/nextcloud/android/commit/3b2824e80e8febff9370a5b906204b51c4831472) Previously, while using the service, [we triggered FileDownloader in a loop](https://github.com/nextcloud/android/blob/54c6d519ff9ac02aced1dd6eb2188c9642ab4c93/app/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java#L448) for mFilesForDirectDownload. Now, [since FileDownloadWorker supports both folder and file sync](https://github.com/nextcloud/android/blob/master/app/src/main/java/com/nextcloud/client/jobs/download/FileDownloadWorker.kt#L225), we can directly use the mFilesForDirectDownload list with the worker.

**How to Test?**

1. Sync Folder
2. Only necessary files will be synchronized

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
@@ -40,6 +40,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Vector;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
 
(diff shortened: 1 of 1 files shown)


ZetaTom (approved):
In my tests, this fixed the endless re-downloading of synchronised folders.

Issue #12652, opened 2024-03-06
account login screen opens again on back press


mahibi:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1) login with new account in files app
2) after successful login, press back button

### Expected behaviour

app is closed

### Actual behaviour

login screen appears again.

Could be fixed by either set intent flags or by custom handling of onBackPressed

### Nextcloud android app version

3.28.0

aloksingh20:
I want to work on this issue, please assign it to me

yadnyeshdande:
Can I work on this issue?

alperozturk96:
Hello. Thanks for the messages, it's really great to see support @aloksingh20 @yadnyeshdande 🙂

raviSaharan07:
Is this issue still open?

neeldoshii:
Hello @alperozturk96 @mahibi, is this issue progress on stale? If yes can I work on this?

joshtrichards:
Related: #11392

Shr3yash:
Hi team, is this issue still open?

Issue #11392, opened 2023-02-16
Pressing back on the home screen does not exit app


smanilov:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Open the app without logging in
2. Press the system back button

### Expected behaviour

The app should exit. This would be normal Android app behavior.

### Actual behaviour

The home screen is presented again.

### Android version

13

### Device brand and model

Samsung Galaxy S22 Ultra

### Stock or custom OS?

Stock

### Nextcloud android app version

Cannot access

### Nextcloud server version

Irrelevant

### Using a reverse proxy?

I don't know

### Android logs

[bug.log](https://github.com/nextcloud/android/files/10755578/bug.log)

### Server error logs

```bash
Irrelevant.
```

### Additional information

Server info is irrelevant, because the bug appears before connecting.

AlvaroBrey:
This is not a bug, just a behaviour that is not implemented. Right now the app just delegates to the system default back pressed handler, which differs depending on Android version; in most modern versions this just moves the app to Recents and shows the launcher.

For the record, I don't think we should change this. But I'm leaving this open for discussion.

smanilov:
I wasn't sure whether to call it a bug, but thought that felt more accurate. I see how it could be seen as a feature request too.

You said: "in most modern versions this just moves the app to Recents and shows the launcher.". But in my case, this is not what I'm observing, while my phone and OS are fairly recent (less than a year old model; updated to Nov 22 Android).

What I'm observing is that the screen goes away, just to reappear immediately. That's why I think it's a bug.

smanilov:
The app is running in this code, which seems wrong to me:

```
    public void onBackPressed() {
        onFinish();

        if (getIntent().getBooleanExtra(EXTRA_ALLOW_CLOSE, false)) {
            Log.d("stanm", "onBackPressed: EXTRA_ALLOW_CLOSE == false");
            super.onBackPressed();
        } else {
            Log.d("stanm", "onBackPressed: EXTRA_ALLOW_CLOSE == true");
[...]

PR #13902, opened 2024-10-30
Fix: Launching LauncherActivity & FirstRunActivity properly

surinder-tsys:
This PR contains the following fixes:
1. Launching Launcher & FirstRun Activity Properly: https://github.com/nextcloud/android/issues/12652
2. Account login screen opens again on back press: https://github.com/nextcloud/android/issues/12652
3. Locking & Unlocking phone in FirstRunActivity redirects to AuthenticatorActivity: https://github.com/nextcloud/android/issues/4993

**Previous Behaviour:**
1. User is not logged In.
4. App launches and directly redirects to AuthenticatorActivity without showing Launcher and FirstRun Activity.
5. On doing successful authentication when user presses back button then AuthenticatorActivity shows again. (This happens only for First login but while adding account this works fine)

**Modified Behaviour:**
1. User is not logged In.
2. App launches and firstly shows LauncherActivity and then redirects to FirstRunActivity.
3. Now, from FirstRunActivity clicking on Login it opens AuthenticatorActivity and once user is authenticated, pressing back finishes the app instead of showing AuthenticatorActivity again.

**Changes Explained:**

-  Starting any Activity checks for User and startAccountCreation() if user not found. This bypass showing Launcher and FirstRun Activity.

To avoid this I have added check for both Activities and skipping redirection to AuthenticatorActivity in **UserAccountManagerImpl** class.
```
public void startAccountCreation(final Activity activity) {
        // skipping AuthenticatorActivity redirection when user is on Launcher or FirstRun Activity
        if (activity instanceof LauncherActivity || activity instanceof FirstRunActivity) return;

        Intent intent = new Intent(context, AuthenticatorActivity.class);

        intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
        context.startActivity(intent);
    }
```

- Making the above change will directly finishes the app as AuthenticatorActivity doesn't shows from LauncherActivity.

To fix this I have added AuthenticatorActivity Intent in **LauncherActivity** if user is not present.
```
private fun scheduleSplashScreen() {
        Handler(Looper.getMainLooper()).postDelayed({
            if (user.isPresent) {
                startActivity(Intent(this, FileDisplayActivity::class.java))
            } else {
                startActivity(Intent(this, AuthenticatorActivity::class.java))
            }
            finish()
        }, SPLASH_DURATION)
    }
```

- To show the FirstRunActivity, under **OnboardingServiceImpl** -> isFirstRun flag which checks currentAccount as shown below always returns false as for no user we are returning **Anonymous** account.

```
override val isFirstRun: Boolean
        get() {
            return accountProvider.currentAccount == null
        }
```
To fix this I have updated the logic as below:
``` 
override val isFirstRun: Boolean
        get() {
            return accountProvider.user.isAnonymous
        }
```

- When user is on FirstRunActivity and locks his phone then unlocks , it redirected to AuthenticatorActivity (mentioned here: https://github.com/nextcloud/android/issues/4993). To avoid this we have add a check in **BaseActivity** for onRestart method. 

```
@Override
    protected void onRestart() {
        Log_OC.v(TAG, "onRestart() start");
        super.onRestart();
        if (enableAccountHandling) {
            mixinRegistry.onRestart();
        }
    }
```

- Finally, to finish the **AuthenticatorActivity** after successful login at first time (Refer: https://github.com/nextcloud/android/issues/12652). I have updated logic as below:

```
private void endSuccess() {
        if (!onlyAdd) {
            Intent i = new Intent(this, FileDisplayActivity.class);
            i.setAction(FileDisplayActivity.RESTART);
            i.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
            startActivity(i);
        }
        finish();
    }
```

### DIFF
--- a/app/src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
+++ b/app/src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
@@ -1,7 +1,7 @@
 /*
  * Nextcloud - Android Client
  *
- * SPDX-FileCopyrightText: 2023 TSI-mc
+ * SPDX-FileCopyrightText: 2023-2024 TSI-mc <surinder.kumar@t-systems.com>
  * SPDX-FileCopyrightText: 2019 Chris Narkiewicz <hello@ezaquarii.com>
  * SPDX-License-Identifier: AGPL-3.0-or-later OR GPL-2.0-only
  */
@@ -19,8 +19,10 @@ import android.content.SharedPreferences;
 import android.preference.PreferenceManager;
 import android.text.TextUtils;
 
+import com.nextcloud.client.onboarding.FirstRunActivity;
 import com.nextcloud.common.NextcloudClient;
 import com.nextcloud.utils.extensions.AccountExtensionsKt;
+import com.nmc.android.ui.LauncherActivity;
 import com.owncloud.android.MainApp;
 import com.owncloud.android.R;
 import com.owncloud.android.authentication.AuthenticatorActivity;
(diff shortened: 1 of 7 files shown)


alperozturk96 (approved):
Tested working and thank you so much for this fix. 💯

joshtrichards:
Related: #11392

tobiasKaminsky (approved):

PR #13494, opened 2024-08-30
2way sync: Better view handling

tobiasKaminsky:
- empty state

- [x] Tests written, or not not needed

PR #13905, opened 2024-10-30
Add disable all button to InternalTwoWaySyncActivity menu

ZetaTom:
This change adds a button to disable two-way synchronisation for all folders. It also stops the two-way sync worker and already enqueued download workers. The latter is necessary, because the two-way sync worker will only check for changes and hand off the actual download to a separate worker, which will continue to run even when the two-way sync worker is no longer running.

The button is located in the overflow menu of the two-way sync settings activity.

### Testing
Start with a fresh install of the Nextcloud app. No folders should be scheduled for two-way sync (Settings → Internal two way sync). To schedule all folders for two-way sync, set a break point on line 41 in `InternalTwoWaySyncActivity.kt`, start the debugger and activity. Once the breakpoint is hit, evaluate the following code snippet:

```kt
fileDataStorageManager.getFolderContent(fileDataStorageManager.getFileById(1), false).forEach { it.internalFolderSyncTimestamp = 0; fileDataStorageManager.saveFile(it) }
```

Once this has finished, resume and relaunch the activity. All sub-directories of the root folder should be scheduled for two-way sync. To disable two-way sync for all folders, tap *Disable all folders* in the overflow menu. After a short time all folders should disappear.

Depends on #13494

---

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManager.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManager.kt
@@ -130,6 +130,11 @@ interface BackgroundJobManager {
         changedFiles: Array<String> = arrayOf<String>()
     )
 
+    /**
+     * Cancel two-way sync. Existing tasks might finish, but no new invocations will occur.
+     */
+    fun cancelTwoWaySyncJob(user: User)
+
     fun scheduleOfflineSync()
 
     fun scheduleMediaFoldersDetectionJob()
--- a/app/src/main/java/com/owncloud/android/ui/activity/InternalTwoWaySyncActivity.kt
+++ b/app/src/main/java/com/owncloud/android/ui/activity/InternalTwoWaySyncActivity.kt
@@ -8,13 +8,24 @@
 package com.owncloud.android.ui.activity
 
 import android.os.Bundle
+import android.view.Menu
+import android.view.MenuInflater
+import android.view.MenuItem
+import android.view.View
+import androidx.core.view.MenuProvider
 import androidx.recyclerview.widget.LinearLayoutManager
 import com.nextcloud.client.di.Injectable
+import com.owncloud.android.R
 import com.owncloud.android.databinding.InternalTwoWaySyncLayoutBinding
 import com.owncloud.android.ui.adapter.InternalTwoWaySyncAdapter
+import com.owncloud.android.utils.theme.ViewThemeUtils
+import javax.inject.Inject
 
-class InternalTwoWaySyncActivity : BaseActivity(), Injectable {
+class InternalTwoWaySyncActivity : DrawerActivity(), Injectable {
     lateinit var binding: InternalTwoWaySyncLayoutBinding
+
+    @Inject
+    lateinit var viewThemeUtils: ViewThemeUtils
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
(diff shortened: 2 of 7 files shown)


app/src/main/java/com/owncloud/android/ui/activity/InternalTwoWaySyncActivity.kt
```diff
@@ -7,24 +7,136 @@
...
+                    )
+                )
+            }
+        }
+    }
+
+    /**
+     * Disable two-way sync for all folders and stop all related workers.
+     */
+    private fun removeAllFolders() {
```
alperozturk96, line 98: A better to use lifecycleScope ([more detailed explanation](https://stackoverflow.com/a/65047492/16787943)). Also, avoid running I/O operations on the main thread; it’s best to update the view on the main thread instead.

Additionally, I think it would be clearer to remove comments within the functions since the function names should already explain what they’re doing. For instance, “removeAllFolders()” might be misleading, so I’d suggest changing it to “disableTwoWaySyncAndWorkers()


```
private fun disableTwoWaySyncAndWorkers() {
        lifecycleScope.launch(Dispatchers.IO) {
            backgroundJobManager.cancelTwoWaySyncJob(user.get())

            val folders = fileDataStorageManager.getInternalTwoWaySyncFolders(user.get())
            folders.forEach { folder ->
                FileDownloadWorker.cancelOperation(user.get().accountName, folder.fileId)
                backgroundJobManager.cancelFilesDownloadJob(user.get(), folder.fileId)

                folder.internalFolderSyncTimestamp = -1L
                fileDataStorageManager.saveFile(folder)
            }

            launch(Dispatchers.Main) {
                internalTwoWaySyncAdapter.update()
            }
        }
    }
```

app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManager.kt
```diff
@@ -130,6 +130,11 @@ interface BackgroundJobManager {
         changedFiles: Array<String> = arrayOf<String>()
     )
 
+    /**
+     * Cancel two-way sync. Existing tasks might finish, but no new invocations will occur.
+     */
+    fun cancelTwoWaySyncJob(user: User)
```
alperozturk96, line 136: I noticed that we didn’t schedule the JOB_INTERNAL_TWO_WAY_SYNC with the user, so why are we cancelling it for them? Should we consider using the user inside the scheduleInternal2WaySync() function instead?

app/src/main/java/com/owncloud/android/ui/activity/InternalTwoWaySyncActivity.kt
alperozturk96, line 98: Better to use lifecycleScope ([more detailed explanation](https://stackoverflow.com/a/65047492/16787943)). Also, avoid running I/O operations on the main thread; it’s best to update the view on the main thread instead.

Note: You have to add this dependency --> `implementation("androidx.lifecycle:lifecycle-runtime-ktx:2.8.6")`



Additionally, I think it would be clearer to remove comments within the functions since the function names should already explain what they’re doing. For instance, “removeAllFolders()” might be misleading, so I’d suggest changing it to “disableTwoWaySyncAndWorkers()


```
private fun disableTwoWaySyncAndWorkers() {
        lifecycleScope.launch(Dispatchers.IO) {
            backgroundJobManager.cancelTwoWaySyncJob(user.get())

            val folders = fileDataStorageManager.getInternalTwoWaySyncFolders(user.get())
            folders.forEach { folder ->
                FileDownloadWorker.cancelOperation(user.get().accountName, folder.fileId)
                backgroundJobManager.cancelFilesDownloadJob(user.get(), folder.fileId)

                folder.internalFolderSyncTimestamp = -1L
                fileDataStorageManager.saveFile(folder)
            }

            launch(Dispatchers.Main) {
                internalTwoWaySyncAdapter.update()
            }
        }
    }
```

app/src/main/res/values/strings.xml
```diff
@@ -1248,4 +1248,8 @@
     <string name="please_select_a_server">Please select a server…</string>
     <string name="notification_icon_description">Unread notifications exist</string>
     <string name="unset_internal_two_way_sync_description">Remove folder from internal two way sync</string>
+    <string name="internal_two_way_sync_list_empty_headline">Two way sync not set up</string>
+    <string name="internal_two_way_sync_text">To set up a two way sync folder, please enable it in the details tab of the folder in question.</string>
+    <string name="internal_two_way_sync_headline">Internal two way sync</string>
+    <string name="action_dismiss_two_way_sync">Disable all folders</string>
```
alperozturk96, line 1254: <string name="two_way_sync_activity_disable_all_button_title">Disable two-way sync for all folders</string>

"Disable all folders" could be misleading or confusing for the user. 

Quick note: The current translation keys don’t clearly indicate where they are used. We can improve this in future pull requests, but for now, perhaps we could adopt a new approach for upcoming PRs.

alperozturk96 (changes requested):

alperozturk96 (commented):
Spotbugs increased

app/src/main/res/values/strings.xml
ZetaTom: Given that we're already on a screen titled *Internal two way sync*, I think it is fair to assume that users know what *disable* means in this context.

As this is just an option in the overflow menu, I didn't want to use an overly verbose string. However, we could adjust this to read *Disable **for** all folders*, if you prefer.

alperozturk96: "Disable for all folders" not the best, but I guess acceptable.

alperozturk96: https://github.com/nextcloud/android/pull/13905/commits/4cec0fc6e02d9666632188f031d347722c018ecd

app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManager.kt
alperozturk96: https://github.com/nextcloud/android/pull/13905/commits/8431266caa0b6eed84eb1679166356365c0c36df

alperozturk96 (approved):

tobiasKaminsky (approved):

PR #13937, opened 2024-11-04
Check Nullable Fields During Folder Changes Check 

alperozturk96:

**How to Test?**

1. Place a debugger in the checkForChanges() function.
2. Try syncing both a changed and an unchanged folder; the debugger should hit the selected line, and the correct result should be returned.


alperozturk96:
@tobiasKaminsky Could you check the returned ResultCodes? Are they correct? Maybe some other result codes can suit better.

tobiasKaminsky (changes requested):
These changes are too risky without real tests.

alperozturk96:
outdated PR

PR #13942, opened 2024-11-04
Feature - Add More MDM

alperozturk96:

**How to Test?**

1. Modify the relevant settings in setup.xml, and ensure the app responds appropriately to the updated conditions.
2. Verify if conditions. It must behave as expected.
3. Test both MDM-related variables and legacy variables in combination to confirm compatibility.

**Disable Multi Account -- User can't add account. New accounts can’t be added from the Manage Account section.**

Screenshot

Screenshot

**Disable Sharing -- Removes share button from action list**

Screenshot

Screenshot

**Disable More External Site  -- Removes external sites from drawer menu**

Screenshot

**Disable Intro  -- Don't launch FirstRunActivity only if user not logged in before**

https://github.com/user-attachments/assets/327c6623-f451-4d6a-84d0-484a8f4c6d12


app/src/main/java/com/owncloud/android/authentication/AccountAuthenticator.java
```diff
@@ -70,7 +71,7 @@ public Bundle addAccount(AccountAuthenticatorResponse response,
 
         final Bundle bundle = new Bundle();
 
-        if (mContext.getResources().getBoolean(R.bool.multiaccount_support) || accounts.length < 1) {
```
alperozturk96, line 73: @tobiasKaminsky accounts.length must be bigger than one so that we can be sure we have multiple accounts, right? In else condition app using this message

` <string name="auth_unsupported_multiaccount">%1$s does not support multiple accounts</string>`

tobiasKaminsky:
MDMConfig.kt -> shareVia Link -> return false, but I can still see it in details:

app/src/main/java/com/owncloud/android/authentication/DeepLinkLoginActivity.kt
```diff
@@ -11,16 +11,14 @@ import android.os.Bundle
...
+import com.nextcloud.utils.mdm.MDMConfig
 import com.owncloud.android.R
 
 class DeepLinkLoginActivity : AuthenticatorActivity(), Injectable {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-
-        if (!resources.getBoolean(R.bool.multiaccount_support) &&
-            accountManager.accounts.size == 1
```
alperozturk96, line 22: @tobiasKaminsky Why we are checking accountManager.accounts.size == 1 here? Why only boolean flag not enough?

app/src/main/java/com/owncloud/android/ui/activity/FileActivity.java
```diff
@@ -718,15 +719,21 @@ public static void copyAndShareFileLink(FileActivity activity,
                                             OCFile file,
                                             String link,
                                             final ViewThemeUtils viewThemeUtils) {
-        ClipboardUtil.copyToClipboard(activity, link, false);
```
alperozturk96, line 721: @tobiasKaminsky Can copyToClipboard flag separately deactivated?

app/src/main/res/xml/app_config.xml
```diff
@@ -25,4 +25,40 @@
...
+
+    <restriction
+        android:key="disable_intro"
+        android:defaultValue="false"
+        android:restrictionType="bool"
+        android:title="@string/app_config_disable_intro_title" />
+
+    <restriction
+        android:key="disable_log"
+        android:defaultValue="true"
```
alperozturk96, line 60: @tobiasKaminsky Default value of the disable_log is true, fyi.

alperozturk96:
> sendFilesSupport -> false but you can still click on "+person" icon on a file in file list to trigger sharing dialog: 

I’ve updated the PR. I believe the correct condition is: if any of the functions—shareViaUser, shareViaLink, or sendFilesSupport—returns false, we should hide the icon, right?

app/src/main/java/com/owncloud/android/ui/activity/FileActivity.java
tobiasKaminsky: All MDM flags can be enabled/disabled individually, yes.

app/src/main/java/com/owncloud/android/ui/adapter/UserListAdapter.java
```diff
@@ -313,11 +319,12 @@ static class AddAccountViewHolderItem extends RecyclerView.ViewHolder {
          * @param accountListAdapterListener {@link Listener}
          */
         private void bind(Listener accountListAdapterListener) {
-            // bind action listener
```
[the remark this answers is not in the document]
tobiasKaminsky: Not sure. But this should be independant from this PR?
So, create a new issue and we can check/discuss.

tobiasKaminsky:
> I’ve updated the PR. I believe the correct condition is: if any of the functions—shareViaUser, shareViaLink, or sendFilesSupport—returns false, we should hide the icon, right?

yes. MDM is to disable sharing/send files/clipboard

So for MDM there is no need to have those 3 functions. But as this is already in place in our code, just combine those for MDM.

app/src/main/java/com/owncloud/android/authentication/AccountAuthenticator.java
alperozturk96: I’m not planning to change it. While adding new MDMs, I noticed that this “if” condition looks a bit strange, but I’m unsure about the correct condition. So I’m asking if it is correct or not.

alperozturk96: https://github.com/nextcloud/android/pull/13942/commits/c08471fc684a1c1514c1d3d3159355f63bef03cd

tobiasKaminsky:
image

Still sharing, but "details" as headline?

tobiasKaminsky:
Share via link:
- internal link share still should be shown (as this is not a real share)

sendFilesSupport:
- still can send it to another app
- also possible to see "DocumentStorageProvider" in Android's Files app

Rest ✔

tobiasKaminsky:
> Disable clipboard should also hide clipboard icon.
> Otherwise user might think it is just broken.

This is missing.

tobiasKaminsky:
Rest is fixed and tested ✔

PR #13974, opened 2024-11-07
BugFix - Sync Nested Folder

alperozturk96:

**How to Test?**

The sync folder action should sync all files within the selected folder, not just those one level below.

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
@@ -40,7 +40,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Vector;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.function.Consumer;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
```diff
@@ -414,23 +413,21 @@ private void classifyFileForLaterSyncOrDownload(OCFile remoteFile, OCFile localF
 
 
     private void prepareOpsFromLocalKnowledge() throws OperationCancelledException {
-        List<OCFile> children = getStorageManager().getFolderContent(mLocalFolder, false);
+        List<OCFile> children = getStorageManager().getAllFilesRecursivelyInsideFolder(mLocalFolder);
```
tobiasKaminsky, line 416: If we want to have nested folder sync, we should start with SynchronizeFolderOperation on folders, but not only on files.

tobiasKaminsky (changes requested):
SynchronizeFolderOperation needs to be used.

alperozturk96:
Duplication of https://github.com/nextcloud/android/pull/14166

pretubedongit:
Why is this closed? It's not a duplication of #14166. This is about nested folder or level 1 files only. It's still present in the latest version.

Issue #13906, opened 2024-10-30
FileDownloadWorker and TwoWaySync Notification


alperozturk96:
### ⚠️ Before posting ⚠️

- [x] This is a **bug**, not a question or an enhancement.
- [x] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [x] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Set up the folder for two-way sync.
2. Wait until the sync is complete.

### Expected behaviour

1. Notifications should not get stuck at 100%.
2. Notifications should reflect actual progress and not jump to 100% when progress is lower.
3. Empty notifications should not appear at all.

### Actual behaviour

Sometimes, notifications get stuck at 100% and don’t reflect actual progress, and empty notifications can also appear.

### Android version

15

### Device brand and model

Android Emulator

### Stock or custom OS?

Stock

### Nextcloud android app version

3.30.3

### Nextcloud server version

29.0.4

### Using a reverse proxy?

I don't know

joshtrichards:
> FileDownloadWorker crashes due to multiple asynchronous start for folder sync, causing android.app.RemoteServiceException$ForegroundServiceDidNotStartInTimeException.

Related reports: #12706

PR #13975, opened 2024-11-07
Improve File Download Worker

alperozturk96:

**Changes**

- Convert worker to CoroutineWorker

- Call setForeground() before starting any task

- Remove non-translatable texts

- Fixes FileDownloadWorker crashes due to an asynchronous start, causing android.app.RemoteServiceException$ForegroundServiceDidNotStartInTimeException.

- Remove total and current index tracking from DownloadNotificationManager. This feature can be reintroduced after addressing [this](https://github.com/nextcloud/android/issues/13906) issue. Refer to the Technical Explanation section for more details; without resolving this, the download status tracking feature is ineffective and nonfunctional.

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
@@ -631,6 +631,7 @@ internal class BackgroundJobManagerImpl(
             .setInputData(data)
             .build()
 
+        // Since for each file new FileDownloadWorker going to be scheduled, better to use ExistingWorkPolicy.KEEP policy.
         workManager.enqueueUniqueWork(tag, ExistingWorkPolicy.KEEP, request)
     }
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
```diff
@@ -626,7 +626,7 @@ internal class BackgroundJobManagerImpl(
             .setInputData(data)
             .build()
 
-        workManager.enqueueUniqueWork(tag, ExistingWorkPolicy.REPLACE, request)
+        workManager.enqueueUniqueWork(tag, ExistingWorkPolicy.KEEP, request)
```
alperozturk96, line 629: Since for each file new FileDownloadWork going to be scheduled, no need to replace with previous one.

app/src/main/java/com/nextcloud/client/jobs/download/DownloadNotificationManager.kt
```diff
@@ -25,17 +26,8 @@ class DownloadNotificationManager(
 ) : WorkerNotificationManager(id, context, viewThemeUtils, R.string.downloader_download_in_progress_ticker) {
 
     @Suppress("MagicNumber")
-    fun prepareForStart(operation: DownloadFileOperation, currentDownloadIndex: Int, totalDownloadSize: Int) {
```
alperozturk96, line 28: Since for each file new FileDownloadWork going to be scheduled, no need to track indexes.

ZetaTom (approved):
Everything worked during my tests.

PR #13993, opened 2024-11-11
BugFix - Don't Use File Download Worker for Two Way Sync

alperozturk96:

**Changes**

- Use FileDownloadWorker only for manual sync to ensure that notifications for downloads do not disappear. Additionally, observe the sync status in the adapter, and ensure the cancellation functionality for sync is not broken after these changes for Two-Way-Sync

- Add operation?.cancel() in the onStopped() method to improve cancellation handling through SynchronizeFolderOperations

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/OfflineSyncWork.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/OfflineSyncWork.kt
@@ -77,7 +77,8 @@ class OfflineSyncWork constructor(
                     user,
                     true,
                     context,
-                    storageManager
+                    storageManager,
+                    true
                 )
                 synchronizeFileOperation.execute(context)
             }
--- a/app/src/main/java/com/owncloud/android/operations/SynchronizeFileOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/SynchronizeFileOperation.java
@@ -114,7 +114,8 @@ public class SynchronizeFileOperation extends SyncOperation {
         User user,
         boolean syncFileContents,
         Context context,
-        FileDataStorageManager storageManager) {
+        FileDataStorageManager storageManager,
+        boolean syncForInternalTwoWaySyncWorker) {
         super(storageManager);
 
         mLocalFile = localFile;
@@ -134,6 +135,7 @@ public class SynchronizeFileOperation extends SyncOperation {
         mSyncFileContents = syncFileContents;
         mContext = context;
         mAllowUploads = true;
+        this.syncForInternalTwoWaySyncWorker = syncForInternalTwoWaySyncWorker;
     }
 
 
@@ -163,9 +165,9 @@ public class SynchronizeFileOperation extends SyncOperation {
         boolean syncFileContents,
         boolean allowUploads,
         Context context,
-        FileDataStorageManager storageManager) {
-
-        this(localFile, serverFile, user, syncFileContents, context, storageManager);
+        FileDataStorageManager storageManager,
+        boolean syncForInternalTwoWaySyncWorker) {
+        this(localFile, serverFile, user, syncFileContents, context, storageManager, syncForInternalTwoWaySyncWorker);
         mAllowUploads = allowUploads;
     }
 
--- a/app/src/main/java/com/owncloud/android/operations/SynchronizeFileOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/SynchronizeFileOperation.java
@@ -17,6 +17,7 @@ import android.content.Context;
 import android.text.TextUtils;
 
 import com.nextcloud.client.account.User;
+import com.nextcloud.client.jobs.download.FileDownloadHelper;
 import com.nextcloud.client.jobs.upload.FileUploadHelper;
 import com.nextcloud.client.jobs.upload.FileUploadWorker;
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/nextcloud/client/jobs/OfflineSyncWork.kt
```diff
@@ -77,7 +77,8 @@ class OfflineSyncWork constructor(
                     user,
                     true,
                     context,
-                    storageManager
+                    storageManager,
+                    true
```
tobiasKaminsky, line 81: boolean syncForInternalTwoWaySyncWorker -> this is for Internal2WaySync, why is then also "true" in OfflineSyncWork?
If this is intented, please rename boolean to .eg. "syncInBackgroundWorker".

app/src/main/java/com/owncloud/android/operations/SynchronizeFileOperation.java
```diff
@@ -295,13 +301,29 @@ private void requestForUpload(OCFile file) {
...
 
-        mTransferWasRequested = true;
+            try {
+                final var operation = new DownloadFileOperation(mUser, file, mContext);
+                var result = operation.execute(getClient());
+
+                mTransferWasRequested = true;
+
+                String filename = file.getFileName();
+                if (filename != null) {
```
tobiasKaminsky, line 314: How can filename be null?

app/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
```diff
@@ -441,8 +446,30 @@ private void syncContents() throws OperationCancelledException {
...
+        if (syncForInternalTwoWaySyncWorker) {
+            try {
+                for (OCFile file: mFilesForDirectDownload) {
+                    if (file == null) continue;
+
+                    final var operation = new DownloadFileOperation(user, file, mContext);
+                    var result = operation.execute(getClient());
+
+                    String filename = file.getFileName();
+                    if (filename == null) continue;
```
tobiasKaminsky, line 458:
```suggestion
if (filename == null)  {
   continue;
   }
```

better readable, and if we add any line it will not break logic.

app/src/main/java/com/nextcloud/client/jobs/OfflineSyncWork.kt
alperozturk96: https://github.com/nextcloud/android/pull/13993/commits/846b70191638cf9eb5a8ed546bbf08ab73a43f9f

app/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
```diff
@@ -441,8 +446,30 @@ private void syncContents() throws OperationCancelledException {
     }
 
     private void startDirectDownloads() {
-        final var fileDownloadHelper = FileDownloadHelper.Companion.instance();
-        mFilesForDirectDownload.forEach(file -> fileDownloadHelper.downloadFile(user, file));
+        if (syncForInternalTwoWaySyncWorker) {
+            try {
+                for (OCFile file: mFilesForDirectDownload) {
```
tobiasKaminsky, line 451: In this loop we also need the mCancellationRequested check?

app/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
alperozturk96: https://github.com/nextcloud/android/pull/13993/commits/c45d8394e25d6b960d55c92702132c6858242085

app/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
tobiasKaminsky: if (mCancellationRequested.get()) {
                throw new OperationCancelledException();
            }

alperozturk96: https://github.com/nextcloud/android/pull/13993/commits/b85affb3f975b087548a2655b55ecfdff01a0dba

app/src/main/java/com/owncloud/android/operations/SynchronizeFileOperation.java
alperozturk96: getFileName() calls getDecryptedFileName(), which then calls getDecryptedRemotePath(). getDecryptedRemotePath() uses decryptedRemotePath, and all of these return nullable strings, so it’s better to check for nullability.

How can filename be null?

resetData() might be called, which will set decryptedRemotePath to null.

FileDataStorageManager.createFileInstance() is setting decryptedRemotePath, and it may also set it to null.

tobiasKaminsky: :+1: Then we should have this in another PR marked as nullable. 
Else we can run into unseen NPE.

tobiasKaminsky (approved):

AndyScherzinger:
@tobiasKaminsky @alperozturk96 breaks FileIT test with a compile error

ZetaTom:
> @tobiasKaminsky @alperozturk96 breaks FileIT test with a compile error

Should be resolved with #13998.

PR #13995, opened 2024-11-11
Clear download jobs on first start

ZetaTom:
This will cause all pending download jobs to be cancelled once on the first start.

The reasoning behind this change is that all download jobs, including those issued by two-way sync, should be cancelled after updating the app.

---

### DIFF
--- a/app/src/main/java/com/nextcloud/client/preferences/AppPreferences.java
+++ b/app/src/main/java/com/nextcloud/client/preferences/AppPreferences.java
@@ -397,4 +397,7 @@ public interface AppPreferences {
 
     void setTwoWaySyncInterval(Long value);
     Long getTwoWaySyncInterval();
+
+    boolean shouldStopDownloadJobsOnStart();
+    void setStopDownloadJobsOnStart(boolean value);
 }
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/nextcloud/client/preferences/AppPreferences.java
```diff
@@ -397,4 +397,7 @@ default void onDarkThemeModeChanged(DarkMode mode) {
 
     void setTwoWaySyncInterval(Long value);
     Long getTwoWaySyncInterval();
+
+    boolean isStopDownloadJobsOnStart();
```
alperozturk96, line 401: `shouldStopDownloadJobsOnStart()` This function name might be better.

alperozturk96 (approved):

app/src/main/java/com/nextcloud/client/preferences/AppPreferences.java
ZetaTom: I agree and have renamed the function.

tobiasKaminsky (approved):

PR #13999, opened 2024-11-12
Optimization code

FjPlays:
added comment for clear details of the function


tobiasKaminsky:
Thank you very much for your contribution.
We usually try to have no in-line documentation, as we want to have the code self explaining.
If we chose to write it, then it must be a tricky code, or a hard to understand reason.

In your added documentation I do not see a real win, as the code itself explains what it wants to do.
Except of L357 and L359.

Can you change it, so that only those two lines remain?

tobiasKaminsky:
I will close this PR for now, but please feel free to update and re-open it.

Issue #2416, opened 2018-03-30
[BUG] Phantom Directory


c0fe:
### Actual behaviour
-Tell us what happens
A mysterious directory "0" appears after initializing file upload and disappears the moment after any interaction with the application.
### Expected behaviour
-Tell us what should happen
 This directory should not exist.
### Steps to reproduce
1. Go to folder
2. Select file upload

### Environment data
Android version:
8.1.0
Device model: 
Nexus 5x
Stock or customized system:
Stock
Nextcloud app version:
3.0.3
Nextcloud server version:
13
### Screenshots
https://i.imgur.com/RXpuTxT.png
https://i.imgur.com/4A4jKNr.png

tobiasKaminsky:
/storage/emulated/0/ is your sdcard/primary storage.
On android the storage system/structure is quite confusing.
Nextcloud open up on sdcard and therefore "0" is shown. Once you enter a subfolder this folder name is shown.

c0fe:
@tobiasKaminsky yeah that is a bit confusing. though I do not have any SD Cards installed onto the phone. Anyway to make this appear easier to end users?

tobiasKaminsky:
Fer first storage is always named this way, independent if it is sd card or or internal storage.
As this is the name of the folder we cannot do that much. Some other apps try to detect this and replace "0" by "sdcard", which can also lead to confusion and is only working for known storage paths...

tobiasKaminsky:
Maybe @nextcloud/designers have an idea?

jancborchardt:
We had this for the storage directory setting as well, where I brought up the same point. ;)

It is mighty confusing for anyone but programmers. We should do our best to try and "translate"/detect these known paths. Storage is after all our core competence so this is really important for flawless UX. :)

joshtrichards:
Related: #1088

PR #14005, opened 2024-11-12
Improved "Choose storage location" dialogue

### LINKED ISSUE #1088: "Storage path" setting is too technical

ZetaTom:
This change introduces a new dialogue that allows users to change the local storage location. It replaces the current dialogue, which may be too complicated for the average user.

Instead of providing the user with a list of paths to choose from, the user is presented with a simple UI allowing them to choose if they want to use internal or external storage. Additionally, users can choose if they want to enable media indexation or not. Enabling media indexation (since renamed to *Allow access from other apps*) allows other apps to access local files from Nextcloud.

### Screenshots

Old | New
--- | ---
data_storage_location_old | data_storage_location_new

### Known Issues
For some users the *External Storage* selection may behave unexpectedly. Even if no external storage is present, it may still be shown as an option. This is to be expected, as this dialogue only interprets the results from calling `DataStorageProvider.getAvailableStoragePoints()`.

Language strings are included since a string was renamed.

Fixes #1088 #2416

---


jancborchardt (commented):
Awesome @ZetaTom! :)

> Enabling media indexation allows other apps to access local files from Nextcloud.

Good point – should we then rename "Allow media indexing" to "Allow other apps to access local Nextcloud files"? Or something shorter of course. :) This will make it way more understandable since I was also unsure what it meant.

ZetaTom:
> Good point – should we then rename "Allow media indexing" to "Allow other apps to access local Nextcloud files"? Or something shorter of course. :) This will make it way more understandable since I was also unsure what it meant.

I've since changed the string to *Allow access from other apps*. In my opinion this still gets the point across and is reasonably short.

jancborchardt (approved):
Nice, looks great! :)

app/src/main/java/com/nextcloud/ui/ChooseStorageLocationDialogFragment.kt
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    private fun updateStorageTypeSelection() {
+        val hasInternalStorage = storagePoints.any { it.storageType == StorageType.INTERNAL }
+        val hasExternalStorage = storagePoints.any { it.storageType == StorageType.EXTERNAL }
+
+        binding.storageInternalRadio.isEnabled = hasInternalStorage
+        binding.storageInternalRadio.text = formatFreeSpaceString(StorageType.INTERNAL, selectedPrivacyType)
+
+        binding.storageExternalRadio.isEnabled = hasExternalStorage
```
alperozturk96, line 135: hasInternalStorage or hasExternalStorage can be false, in that case no need to call formatFreeSpaceString.

This approach could be better:

```
val hasInternalStorage = storagePoints.any { it.storageType == StorageType.INTERNAL }
binding.storageInternalRadio.isEnabled = hasInternalStorage
binding.storageInternalRadio.text = if (hasInternalStorage) formatFreeSpaceString(StorageType.INTERNAL, selectedPrivacyType) else ""
```

app/src/main/java/com/nextcloud/ui/ChooseStorageLocationDialogFragment.kt
```diff
@@ -0,0 +1,167 @@
...
+        val radioButton = when (getCurrentStorageLocation().storageType) {
+            StorageType.EXTERNAL -> binding.storageExternalRadio
+            else -> binding.storageInternalRadio
+        }
+
+        radioButton.isChecked = true
+        updateMediaIndexSwitch()
+    }
+
+    private fun formatFreeSpaceString(storageType: StorageType, privacyType: PrivacyType): String {
```
alperozturk96, line 99: This function does not just format a string but also returns it; therefore, I would name it something like `getFormattedFreeSpaceString`.

alperozturk96 (changes requested):
It looks good 💯. With a bit more polishing, I think it’ll be ready to merge.

app/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
```diff
@@ -802,34 +801,16 @@ private void setupGeneralCategory() {
...
-                }
-                StorageMigration storageMigration = new StorageMigration(this, user, storagePath, newPath, viewThemeUtils);
-                storageMigration.setStorageMigrationProgressListener(this);
-                storageMigration.migrate();
-
-                return false;
+        prefDataLoc = findPreference(AppPreferencesImpl.DATA_STORAGE_LOCATION);
+        if (prefDataLoc != null) {
+            prefDataLoc.setOnPreferenceClickListener(p -> {
+                Intent intent = new Intent(MainApp.getAppContext(), ChooseStorageLocationActivity.class);
```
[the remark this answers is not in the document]
ZetaTom: Since `SettingsActivity` extends the deprecated `PreferenceActivity`, we unfortunately need a intermediary Activity to show the fragment. We already have a similar workaround in place for `SyncedFoldersActivity`, `ContactsPreferenceActivity`, and `SetupEncryptionActivity`.

app/src/main/java/com/nextcloud/ui/ChooseStorageLocationDialogFragment.kt
ZetaTom: To provide a intuitive user experience, we still need all available options to be displayed with the respective labels. `formatFreeSpaceString` will return a String already properly formatted, ready to be displayed, even if the storage type isn't available.

The suggested approach would show a non-selectable option without any associated label to provide context:

untitled

app/src/main/java/com/nextcloud/ui/ChooseStorageLocationDialogFragment.kt
ZetaTom: I've renamed `formatFreeSpaceString` to `getStoragePointLabel`.

app/src/main/java/com/nextcloud/ui/ChooseStorageLocationDialogFragment.kt
alperozturk96: If a storage type isn’t available, I will hide its radio button and text. We shouldn’t attempt to calculate the free space for unavailable storage.

ZetaTom: We aren't performing any calculations for storage locations which aren't available.

Using `storagePoints.firstOrNull()` we figure out if any locations match the given criteria. If no location is found, `storagePath` is set to `null` and `typeString` is returned.

alperozturk96 (approved):

tobiasKaminsky (approved):

Eiernase:
nice that i need to dig into some random change on github to find where my files are stored because the UI doesn't show me :(

AndyScherzinger:
@Eiernase you don't, it is the first item if you navigate to the settings from the main menu.

Eiernase:
@AndyScherzinger the option is the first item in the settings, but there's no mention anymore of where the internal and external storage locations are.
I was expecting there to just be a "Nextcloud" folder in my files next to all other initial directories – you know, like most apps do.
But it doesn't matter anyway, since there's apparently no way to keep a folder offline (feature requested seven years ago).

AndyScherzinger:
> But it doesn't matter anyway, since there's apparently no way to keep a folder offline (feature requested seven years ago).

Yes there is since quite a while. Open the details tab of a folder and hit the checkbox "Sync"

Eiernase:
I tried using that too, but it doesn't download the folder.

AndyScherzinger:
looping in @tobiasKaminsky since you build the sync feature I if am not mistaken.

PR #14030, opened 2024-11-15
BugFix - Icon Visibilities

alperozturk96:

• Hide the people icon after sharing with at least one user (resolves share icon visibility issues).
• Use correct overlay icons for folders (e.g., shared with users, shared via link, locked, encrypted).
• Add offline operation folder type to screenshot tests.
• Fix and update broken screenshot tests.

| Before  | After |
| ------------- | ------------- |
| w | r |

**Folders With Overlay Icon**

Screenshot 2024-11-15 at 11 57 50


AndyScherzinger:
Screenshot Tests look good to me. Just replace the existing ones with the new ones. They are visually similar but fail due to the update of the common lib which has m3 theming so the blue and Grey tones are slightly different now 👍

AndyScherzinger:
LGTM

@alperozturk96 As a follow up maybe, did you see that the folder icons with an overlay are slightly smaller see their width than folders without an overlay? 🔍

alperozturk96:
@AndyScherzinger Thank you. Good point. I will update the PR.

app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -663,53 +658,47 @@ private void bindListItemViewHolder(ListItemViewHolder holder, OCFile file) {
...
             }
-        });
+        } else {
+            boolean isAutoUpload = SyncedFolderProvider.isAutoUploadFolder(syncedFolderProvider, file, user);
+            boolean isDarkModeActive = preferences.isDarkModeEnabled();
+            Drawable icon = MimeTypeUtil.getOCFileIcon(file, context, viewThemeUtils, isAutoUpload, isDarkModeActive);
+            holder.getThumbnail().setImageDrawable(icon);
+            if ("image/png".equalsIgnoreCase(file.getMimeType())) {
+                holder.getThumbnail().setBackgroundColor(context.getResources().getColor(R.color.bg_default, context.getTheme()));
+            }
```
ZetaTom, line 694: During testing, I've notices that transparent PNGs are always displayed with a white and WEBPs with a black background. Is there a reason why PNGs are handled differently in this case?

image

ZetaTom (changes requested):
During my testing I've noticed some inconsistencies compared to the previous implementation. Otherwise it seems to work as expected; the share icon is only displayed when expected.

1. Corners of thumbnails are no longer rounded (see screenshot below)
2. Thumbnails of images are only displayed after a manual reload (pull to refresh)
3. The `ic_movie` overlay for video files is scaled wrong (see screenshot below)
4. Appearance of auto-upload folder has changed (icon used to be taller)

image

jancborchardt (commented):
Just a small thing – the avatar seems to be a bit off to the right compared to the sharing icons. Not super important but might be easy to fix?
image

PR #14051, opened 2024-11-19
Fix FileDetailFragment not starting

ZetaTom:
This fixes an issue where the `FileDetailFragment` would not start or would start on the wrong tab.

---


alperozturk96 (approved):
Tested working

PR #14060, opened 2024-11-20
Check Storage Path Emptiness

alperozturk96:
- [x] Tests written, or not not needed

**Fixes**

image

**How to reproduce?**

1. Open new dialog
2. Press ok button close dialog
3. Try to open new dialog again
4. Crash

PR #14070, opened 2024-11-21
Fix exception when closing choose storage location dialogue

ZetaTom:
This change is an addition to #14005. Previously, the app would crash after closing the *Choose storage location* dialogue.

- read `storagePath` from shared preferences
- check if `storagePath` is null

Replaces #14060

---


ZetaTom:
I'm not sure why the screenshot tests and the subsequent upload are failing.

alperozturk96 (approved):
Tested working

PR #14086, opened 2024-11-25
Fix Status Bar & Action Bar Appereance for SDK 35

alperozturk96:


alperozturk96:
Better approach https://github.com/nextcloud/android/pull/14098

PR #14089, opened 2024-11-25
Increase Build Speed

alperozturk96:

- Avoid compiling unnecessary resources for debug build
- Use useBuildCache and enable incremental.apt for kapt
- Change order of repositories. Put gradlePluginPortal() in the last. Gradle searches repositories in the order that they're declared, so build performance is improved if the repositories listed first contain most of the plugins.
- Reduce heap size. Increasing heap size also can affect build speed negatively
- Enable parallel garbage collector

[Document](https://developer.android.com/build/optimize-your-build)


alperozturk96:
> Could you please explain why decreasing the heap size will result in a decreased build time?
> 
> I was under the impression that the general consensus was that _in_creasing the heap size should also increase overall performance.

It’s incorrect to assume a linear relationship. Increasing the IDE heap size does not result in an equivalent gain, as indicated by the informational message on the Android Studio memory settings page.

For example, if the IDE max heap size is configured to use 6GB of RAM, and your browser and operating system are also consuming significant memory, your system might quickly start using swap space. This can lead to slower performance as the system offloads data to disk when physical memory is insufficient.

Screenshot 2024-12-03 at 08 56 26

I increased the Daemon’s maximum heap size to address CI failures caused by low memory. You can find the corresponding error messages in the logs of failed CI's.

I decreased the IDE’s maximum heap size to 2GB, since I haven’t encountered any warnings from Android Studio at this level. If the IDE displays a warning, adjustments can be made accordingly.

alperozturk96:
In Android Studio, Memory Settings page indicates allocating too much memory also might affect the performance negatively. This relationship is not linear, as each project has unique requirements based on its modules, dependencies, and complexity.

mem_alloc

I increased the "kotlin.daemon.jvmargs" because the CI was failing and reporting low memory issues. You can observe the same issue in some failing PRs. Afterward, I conducted several tests on my local machine, and a 2G configuration performed better than the master branch.

After your question, I wanted to run benchmarks, and here are the results.

**PR 6G JVMARGS**

pr_6g_jvmargs

**PR 2G JVMARGS**

pr_2g_jvmargs

**Master**

master


It seems that 6G is performing well.

For each variant, I performed the following steps:

1. Opened Android Studio → File → Invalidate Caches and selected all three checkboxes.
2. Restart Android Studio 
3.  "/gradlew clean"
4. "./gradlew assembleDebug --profile"


Of course, further fine-tuning can be done. 16G is too much? What about 12G? Or any other Gradle option would help more? To be honest, I really don't know how much memory would be the best, and can't find any document that explain this well.

If you have any suggestions, please feel free to share.

tobiasKaminsky:
Adding sane defaults is good, but everyone is able to overwrite them, if they have e.g. way too much ram ;-)

tobiasKaminsky (approved):

PR #14095, opened 2024-11-27
Check BOOT_COMPLETED BroadcastReceiver Usage For API Level 35

alperozturk96:

Android 15 introduces new restrictions on BOOT_COMPLETED broadcast receivers launching foreground services. Specifically, BOOT_COMPLETED receivers are prohibited from starting the following types of foreground services: dataSync, mediaPlayback, mediaProjection, phoneCall, microphone

If a BOOT_COMPLETED receiver attempts to start any of these service types, the system will throw a ForegroundServiceStartNotAllowedException.

WorkManager provides a solution by managing and running a foreground service on your behalf to execute a WorkRequest, while also displaying a configurable notification.

MainApp.initSyncOperations() this function triggers a series of workers and is also invoked by the BOOT_COMPLETED broadcast receiver. The shouldSync flag differentiates between scenarios and ensures that WorkManager is triggered appropriately.

[BOOT_COMPLETED BroadcastReceiver Behavior Changes Doc](https://developer.android.com/about/versions/15/behavior-changes-15#fgs-boot-completed)

[Worker Document](https://developer.android.com/develop/background-work/background-tasks/persistent/how-to/long-running)

### DIFF
--- a/app/src/main/java/com/owncloud/android/files/BootupBroadcastReceiver.java
+++ b/app/src/main/java/com/owncloud/android/files/BootupBroadcastReceiver.java
@@ -14,6 +14,7 @@ package com.owncloud.android.files;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
+import android.os.Build;
 
 import com.nextcloud.client.account.UserAccountManager;
 import com.nextcloud.client.core.Clock;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/owncloud/android/files/BootupBroadcastReceiver.java
```diff
@@ -62,7 +63,10 @@ public void onReceive(Context context, Intent intent) {
         AndroidInjection.inject(this, context);
 
         if (Intent.ACTION_BOOT_COMPLETED.equals(intent.getAction())) {
-            MainApp.initSyncOperations(context,
+            boolean isApiLevel35OrHigher = (Build.VERSION.SDK_INT >= 35);
```
tobiasKaminsky, line 66: Maybe it is better to have isApiLevelLowerThan35.
As you always negate it.

alperozturk96: https://github.com/nextcloud/android/pull/14095/commits/2fd96ebe6512fa4d684ac5389a5bf1075988703a

tobiasKaminsky:
We discussed if this is needed at all for newer API levels, as work manager jobs might persist.
Have you checked this?

alperozturk96:
> We discussed if this is needed at all for newer API levels, as work manager jobs might persist. Have you checked this?

We can change the foregroundServiceType to "specialUse" and explain why it's necessary for our app. I will create new PR for that.

PR #14099, opened 2024-11-29
MANAGE_EXTERNAL_STORAGE not possible for gplay flavor anylonger

tobiasKaminsky:
- only for GPLAY
- shows a warning only for users that use autoupload with move/delete
- correctly show if files are read-only (folder read-only check seems not to be enough)

### DIFF
--- a/app/src/main/java/com/nextcloud/utils/BuildHelper.kt
+++ b/app/src/main/java/com/nextcloud/utils/BuildHelper.kt
@@ -1,0 +1,11 @@
+/*
+ * Nextcloud - Android Client
+ *
+ * SPDX-FileCopyrightText: 2024 Your Name <your@email.com>
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+package com.nextcloud.utils
+
+object BuildHelper {
+    const val GPLAY: String = "gplay"
+}
--- a/app/src/main/java/com/nextcloud/utils/BuildHelper.kt
+++ b/app/src/main/java/com/nextcloud/utils/BuildHelper.kt
@@ -1,7 +1,7 @@
 /*
  * Nextcloud - Android Client
  *
- * SPDX-FileCopyrightText: 2024 Your Name <your@email.com>
+ * SPDX-FileCopyrightText: 2024 Tobias Kaminsky <tobias.kaminsky@nextcloud.com>
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 package com.nextcloud.utils
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -13,6 +13,7 @@
  */
 package com.owncloud.android.ui.activity;
 
+import android.Manifest;
 import android.accounts.Account;
 import android.accounts.AuthenticatorException;
 import android.annotation.SuppressLint;
@@ -50,6 +51,7 @@ import com.nextcloud.appReview.InAppReviewHelper;
 import com.nextcloud.client.account.User;
 import com.nextcloud.client.appinfo.AppInfo;
 import com.nextcloud.client.core.AsyncRunner;
+import com.nextcloud.client.core.Clock;
 import com.nextcloud.client.di.Injectable;
 import com.nextcloud.client.editimage.EditImageActivity;
 import com.nextcloud.client.files.DeepLinkHandler;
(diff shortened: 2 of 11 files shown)


app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -284,6 +291,43 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                .setMessage(R.string.auto_upload_gplay_desc)
+                .setNegativeButton(R.string.dialog_close, (dialog, which) -> {
+                    dialog.dismiss();
+                })
+                .setIcon(R.drawable.nav_synced_folders)
+                .create()
+                .show();
+        }
+
+        preferences.setAutoUploadGPlayWarningShown(true);
```
alperozturk96, line 328: Shouldn't we set this true when showInfoDialog was also true?

tobiasKaminsky: If they set it up later, then they can only set it up with new behaviour, s to there is no need to inform them that something has changed.

alperozturk96 (approved):
Tested

app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
ZetaTom: I'm currently running the *gplayDebug* build variant and am still able to set the upload action to move or delete when setting up a custom auto upload folder. If I'm not mistaken, this option should be disabled.

tobiasKaminsky: On which API level? As older level will still work.
Hm. I did not thougth about it.
Then we should show this warning depending on API level?

ZetaTom: > On which API level?

I was using a Pixel 2 emulator running an AOSP image on API version 34 (Android 14). For testing, I did a fresh install each time, built using the `gplayDebug` variant.

tobiasKaminsky: Can you try it with Android 15?

app/src/main/java/com/nextcloud/utils/BuildHelper.kt
```diff
@@ -0,0 +1,11 @@
+/*
+ * Nextcloud - Android Client
+ *
+ * SPDX-FileCopyrightText: 2024 Your Name <your@email.com>
```
alperozturk96, line 4: Please fix license header :)

GuyFromMars365:
If MANAGE_EXTERNAL_STORAGE is not possible in gplay, what reason did Google for provide for declining?According to [this page](https://support.google.com/googleplay/android-developer/answer/10467955?hl=en), you would have to submit the [Permissions Declaration Form](https://support.google.com/googleplay/android-developer/answer/9214102?hl=en&sjid=16615896801329230378-NC#zippy=%2Cstep-specify-your-apps-core-functionality:~:text=special%20instructions.-,Complete%20the%20Permissions%20Declaration%20Form,-Step%201%3A%20Evaluate) to have this permission. I'm disappointed Google/Gplay/Android declined this request, as it almost renders the app partially useless.

DragonQ:
> If MANAGE_EXTERNAL_STORAGE is not possible in gplay, what reason did Google for provide for declining?According to [this page](https://support.google.com/googleplay/android-developer/answer/10467955?hl=en), you would have to submit the [Permissions Declaration Form](https://support.google.com/googleplay/android-developer/answer/9214102?hl=en&sjid=16615896801329230378-NC#zippy=%2Cstep-specify-your-apps-core-functionality:~:text=special%20instructions.-,Complete%20the%20Permissions%20Declaration%20Form,-Step%201%3A%20Evaluate) to have this permission. I'm disappointed Google/Gplay/Android declined this request, as it almost renders the app partially useless.

Is there any more info on what the Nextcloud team will do about this? As you say, it renders the app half useless.

Does the F-Droid version still work as it used to?

aptalca:
The app stopped uploading geotags with the latest update for me. Is it perhaps related to this PR?

sol8712:
If Nextcloud is open source and the community is expected to assist with development, bugs, maintaining various apps, the community needs some transparency on what Googles reasons are for declining so developers can work within Googles rules.

If Googles own rules or decisions maybe breaking some laws or regulations people need that info to be able to report them to the relevant authority.

PeeK1e:
> Is there any more info on what the Nextcloud team will do about this? As you say, it renders the app half useless.
> 
> Does the F-Droid version still work as it used to?

I installed the F-Droid Version, and I finally can upload my Signal backups again. So yes. It works.
Regarding the permissions, we have to hope for either the EU or the US to force Google to not be a monopolist. But you may wake me from my grave when that happens.

PR #14122, opened 2024-12-04
3.30.6

alperozturk96:

### DIFF
--- a/fastlane/metadata/android/en-US/changelogs/30300690.txt
+++ b/fastlane/metadata/android/en-US/changelogs/30300690.txt
@@ -1,0 +1,8 @@
+## 3.30.5 (December 4, 2024)
+
+- MANAGE_EXTERNAL_STORAGE permission removed for Google Play Store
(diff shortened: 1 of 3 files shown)


fastlane/metadata/android/en-US/changelogs/30300690.txt
```diff
@@ -0,0 +1,8 @@
+## 3.30.6 (December 4, 2024)
+
+- MANAGE_EXTERNAL_STORAGE permission removed for Google Play Store
```
tobiasKaminsky, line 3: Can you add that this needs to be done, something like:
Due to new restrictions imposed by Google, the auto upload feature will no longer be able to automatically remove uploaded files

alperozturk96: https://github.com/nextcloud/android/pull/14122/commits/e668d941eadf2145b2a5c41544aa12c987eaf8f8

tobiasKaminsky (approved):

Issue #11432, opened 2023-03-01
Possibility to use memories as gallery in the app instead of the default gallery if installed


MrLoverLoverMMMM:
### Is your feature request related to a problem? Please describe.
Well more or less.
The speed of the gallery for more than 1 Million Pictures is very slow. In have installed memories to solve this issue.
The problem is that in the nextcloud app, you cannot use the new memories, so the gallery there is very very slow.

### Describe the solution you'd like
Possibility to choose between memories and default gallery in the apps settings.
Alternative use memories if installed otherwise fallback to gallery default.

BluePixel4k:
I would also love to see an integration of the app "memories" in the nextcloud app :)

germebl:
just want to let you know that I am also very interested in this. I currently have to use another gallery app just to avoid using Google Photos because I can't select Memories to open photos or as the default gallery app.

I already tried to choose a picture in the filemanager and select "open with" and then wanted to choose Memories, but it seems that Memories is not supported as Gallery App. Just my banking app? wtf?

It's a Pixel 7 and it does not have a option for the default galery app, so i would need to use the way i mentioned above and then when Memories would be choosable, i would need to select "Always" to set it as default Gallery app.

PR #14146, opened 2024-12-09
Feature - Bottom Navigation View

alperozturk96:

Screenshot_20241210_142328

**How to Test?**

- Assistant bottom navigation option must be visible when user have that feature
- Navigation from drawer menu or bottom navigation bar must highlight correct drawer menu item and bottom navigation item

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -257,6 +257,9 @@ public abstract class DrawerActivity extends ToolbarActivity
             menuItemId = menuItem.getItemId();
 
             if (menuItemId == R.id.nav_all_files) {
+                MainApp.showOnlyPersonalFiles(false);
+                MainApp.showOnlyFilesOnDevice(false);
+
                 Intent intent = new Intent(this, FileDisplayActivity.class);
                 intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
                 intent.setAction(FileDisplayActivity.ALL_FILES);
--- a/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -241,7 +241,6 @@ public abstract class DrawerActivity extends ToolbarActivity
         viewThemeUtils.platform.colorBottomNavigationView(bottomNavigationView);
     }
 
-    @SuppressWarnings("unused")
     private void checkAssistantBottomNavigationMenu() {
         boolean isAssistantAvailable = getCapabilities().getAssistant().isTrue();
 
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1839,7 +1839,10 @@ protected void handleSearchEvent(SearchEvent event) {
 
         // avoid calling api multiple times if async task is already executing
         if (remoteOperationAsyncTask != null && remoteOperationAsyncTask.getStatus() != AsyncTask.Status.FINISHED) {
-            Log_OC.d(TAG, "OCFileListSearchAsyncTask already running skipping new api call for search event: " + searchEvent.getSearchType());
+            if (searchEvent != null) {
```
alperozturk96, line 1842: It was causing a crash during bottom nav bar changes

app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -155,7 +156,7 @@ public abstract class DrawerActivity extends ToolbarActivity
     /**
      * Reference to the navigation view.
      */
-    private NavigationView mNavigationView;
+    private NavigationView drawerNavigationView;
```
alperozturk96, line 159: Since we now have two navigation views, I had to rename them to make them distinguishable. Clicking on an item in the drawer menu or the bottom navigation view must correctly highlight the corresponding item in both the drawer menu and the bottom navigation view.

alperozturk96:
@tobiasKaminsky I don't want to use boolean value. How can we suppress it?

Screenshot 2024-12-10 at 14 27 56

tobiasKaminsky:
You can suppress the warnings with "@SuppressFBWarnings("RV")".
You need to add it above the function.

Hyeyoung346:
I also think there should always be text under the icons. And the iOS navigation bar is also the same.  I'm wondering if "Photos" should be changed to "Media" like on iOS, since that section actually contains videos as well.

tobiasKaminsky:
One small thing 

- click favorites
- click assistant
- go back
- see that assistant is still highlighted

tobiasKaminsky (commented):
Works great :+1:

tobiasKaminsky:
Rebasing as I fixed lint with #14185

alperozturk96:
https://github.com/nextcloud/android/pull/14146/commits/ce28b385c24ae6a8137d10a54935d41b8c699126

tobiasKaminsky (approved):

BluePixel4k:
Would it be possible to replace the Media tab with the Memories app if it's installed on the server and enabled for the user?
Because many features are missing in the media view.
The Memories app could initially be integrated via WebView.

The integration is also requested in this issue: https://github.com/nextcloud/android/issues/11432

PR #14151, opened 2024-12-10
Improve CI Performance

alperozturk96:

The CI process was failing with 1GB of memory allocation. 2GB resolved the issue intermittently, but it was still insufficient in certain cases. 

To address this, I allocated 6GB of memory, which may be excessive for the CI environment. @tobiasKaminsky What do you think?


tobiasKaminsky:
Seems to work.
If not, we can always shrink it.

tobiasKaminsky (approved):

PR #14161, opened 2024-12-11
BugFix - Get Files For Auto Upload

alperozturk96:

FilesystemDataProvider.getFilesForUpload → This function returns local paths for auto-upload.

The new function performs the same task. When I pass syncedFolder.localPath, the new function returns the expected result, but the old function returns an empty list.

**How to Test**

Set up auto-upload. The local path must contain nested folders, and each folder should have files.

Ensure all files are fetched from the given directory, even with subfiles

The correct name collision policy should be applied. For example:
	• Ask Every Time or Skip Upload: Files should not be re-uploaded.
	• Overwrite: Files should be re-uploaded.


tobiasKaminsky (approved):

tobiasKaminsky:
@alperozturk96 can this be a reason for  https://github.com/nextcloud/android/issues/14198 ?

Issue #14238, opened 2024-12-25
NetworkOnMainThreadException:  FileDataStorageManager.saveFileWithParent


tabi-ya:
### Cause of error
```java
Exception in thread "main" com.owncloud.android.operations.RemoteOperationFailedException: Error during saving file with parents: /Signal/signal-2024-12-24-21-51-40.backup / Unexpected exception / null
    at com.owncloud.android.datamodel.FileDataStorageManager.saveFileWithParent(FileDataStorageManager.java:549)
    at com.owncloud.android.datamodel.FileDataStorageManager.createPendingFile(FileDataStorageManager.java:209)
    at com.owncloud.android.datamodel.FileDataStorageManager.addCreateFileOfflineOperation(FileDataStorageManager.java:174)
    at com.owncloud.android.ui.activity.FileDisplayActivity.lambda$requestUploadOfFilesFromFileSystem$16(FileDisplayActivity.java:1023)
    at com.owncloud.android.ui.activity.FileDisplayActivity.$r8$lambda$GG51LwFMwIhUrmNJlkp9-QY2xq0(Unknown Source:0)
    at com.owncloud.android.ui.activity.FileDisplayActivity$$ExternalSyntheticLambda9.onComplete(D8$$SyntheticClass:0)
    at com.nextcloud.client.network.ConnectivityServiceImpl.lambda$isNetworkAndServerAvailable$0(ConnectivityServiceImpl.java:76)
    at com.nextcloud.client.network.ConnectivityServiceImpl$$ExternalSyntheticLambda1.run(D8$$SyntheticClass:0)
    at android.os.Handler.handleCallback(Handler.java:938)
    at android.os.Handler.dispatchMessage(Handler.java:99)
    at android.os.Looper.loop(Looper.java:233)
    at android.app.ActivityThread.main(ActivityThread.java:8063)
    at java.lang.reflect.Method.invoke(Native Method)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:631)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:978)
Caused by: Exception in thread "main" android.os.NetworkOnMainThreadException
    at android.os.StrictMode$AndroidBlockGuardPolicy.onNetwork(StrictMode.java:1605)
    at org.conscrypt.Platform.blockGuardOnNetwork(Platform.java:804)
[...]

PR #14268, opened 2024-12-30
BugFix - NetworkOnMainThreadException: FileDataStorageManager.saveFileWithParent

alperozturk96:

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
@@ -81,8 +81,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.FutureTask;
 
 import androidx.annotation.NonNull;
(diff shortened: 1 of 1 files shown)


tobiasKaminsky (changes requested):
Please fix in FileDisplayActivity

alperozturk96:
> I think FileDataStorageManager shall not handle this. Instead, as seen in the linked bug, the root cause is that requestUploadOfFilesFromFileSystem() from FileDisplayActivity runs on MainThread.

After your comment, I checked again, and the actual issue is with the method:

`void isNetworkAndServerAvailable(@NonNull GenericCallback<Boolean> callback);`

This callback runs on the main thread, which is necessary for some cases. However, in other cases, like this one, it must run on a background thread instead of the main thread.

**Current Usage of isNetworkAndServerAvailable:**

**Must run on a background thread:**
	•	fileOperationsHelper?.removeFiles
	•	fileDataStorageManager.addRemoveFileOfflineOperation
	•	FileUploadHelper.Companion.instance().uploadNewFiles
        •	fileDataStorageManager.addCreateFileOfflineOperation

**Must run on the main thread:**
	•	DisplayUtils.showSnackMessage
	•	ocFileListFragment.setEmptyListMessage
	•	fileDisplayActivity.refreshCurrentDirectory

To address this, it would be better to create a new PR. Before proceeding, I want to inform you because this PR will touch multiple files. While it won’t be a large PR, it will require careful testing.

What do you think? Is it okay for you?

alperozturk96:
@tobiasKaminsky How are we going to proceed? I explained earlier I have to change multiple files to fix [this](https://github.com/nextcloud/android/issues/14238) issue. Shall I create a PR for that?

alperozturk96:
outdated PR

PR #14270, opened 2024-12-30
BugFix - ThumbnailCacheManager Software rendering doesn't support hardware bitmaps

alperozturk96:

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -70,6 +70,7 @@ import java.util.List;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Can you add more info on this?
Maybe we can even write a test with some test images?

alperozturk96:
> Can you add more info on this? Maybe we can even write a test with some test images?

As I understand it, this issue occurs when an app tries to render or modify a bitmap that has been created as a hardware-accelerated bitmap, but the rendering or modification is attempted using software rendering.

Hardware rendering is handled by the GPU, while software rendering is handled by the CPU. Writing a test for this scenario can be hard. Question to be asked: why would a device attempt to render using the CPU, especially since hardware rendering is [enabled by default ](https://developer.android.com/topic/performance/hardware-accel)for target API levels greater than 13?

tobiasKaminsky:
I have no real insights, thus I am unsure about negative sideeffects.
We can merge it, and see if daily users complain.

tobiasKaminsky (approved):

PR #14299, opened 2025-01-03
enable power warning for emulators

tobiasKaminsky:

### DIFF
--- a/app/src/androidTest/java/com/nextcloud/client/SyncedFoldersActivityIT.java
+++ b/app/src/androidTest/java/com/nextcloud/client/SyncedFoldersActivityIT.java
@@ -92,7 +92,7 @@ public class SyncedFoldersActivityIT extends AbstractIT {
 
         AlertDialog sut = activity.buildPowerCheckDialog();
         
-        activity.runOnUiThread(sut::show);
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/com/nextcloud/client/SyncedFoldersActivityIT.kt
```diff
@@ -119,7 +119,8 @@ class SyncedFoldersActivityIT : AbstractIT() {
                 onIdleSync {
                     EspressoIdlingResource.increment()
                     val dialog = sut.buildPowerCheckDialog()
-                    dialog.show()
+                    scenario.onActivity { showPowerCheckDialog() }
```
alperozturk96: Please use following code. No need for another ` scenario.onActivity`

`sut.showPowerCheckDialog()`

alperozturk96 (changes requested):

app/src/androidTest/java/com/nextcloud/client/SyncedFoldersActivityIT.kt
tobiasKaminsky:
```suggestion
                   sut.showPowerCheckDialog()
```

PR #14342, opened 2025-01-08
BugFix - Auto Upload Check File Sent For Upload

alperozturk96:

**How to Test**

Set up auto-upload. The local path must contain nested folders, and each folder should have files.

Ensure all files are fetched from the given directory with subfiles. After successfully uploaded, files should not upload again.

Screenshot 2025-01-08 at 11 48 21


tobiasKaminsky:
Closing, as we revert the sync logic.
Might be back for RC2.

PR #14364, opened 2025-01-09
Stop Uploads When the Device Doesn’t Have an Internet Connection

alperozturk96:

**Problem**

When the device lacks an active internet connection or is in airplane mode, the app continues attempting to upload files. This results in numerous “upload failed” notifications appearing repeatedly.

**How to Test**

1. Attempt to upload multiple files
2. During the upload process, disconnect the internet or enable airplane mode.
3. Observe that multiple error notifications are generated.

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/upload/UploadNotificationManager.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/upload/UploadNotificationManager.kt
@@ -167,6 +167,8 @@ class UploadNotificationManager(private val context: Context, viewThemeUtils: Vi
         }
     }
 
+    fun dismissErrorNotification() = notificationManager.cancel(FileUploadWorker.NOTIFICATION_ERROR_ID)
+
     fun dismissOldErrorNotification(remotePath: String, localPath: String) {
         notificationManager.cancel(
             NotificationUtils.createUploadNotificationTag(remotePath, localPath),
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/nextcloud/client/jobs/upload/UploadNotificationManager.kt
```diff
@@ -143,6 +143,18 @@ class UploadNotificationManager(private val context: Context, viewThemeUtils: Vi
         )
     }
 
+    fun showConnectionErrorNotification() {
+        notificationBuilder.run {
+            setContentTitle(context.getString(R.string.file_upload_worker_error_notification_title))
+            setContentText("")
```
tobiasKaminsky, line 149: Will it retry once a network connection is detected?
If so, we can add this info here as text.

alperozturk96: If the FileUploadWorker is triggered for auto-upload, it will run again. However, if it is triggered for manual file upload, it will not. We could indicate this detail, but it might be too much information for the user.

alperozturk96:
backport to stable-3.31

tobiasKaminsky (approved):

Zoidberg2:
I believe that this introduced issues, now i cant upload any files because the app is always saying "no internet connection"
i do use my phone on airplane mode all the time(it has internet via WIFI).

Issue #10645, opened 2022-08-18
Externally deleted files uploaded again (because external app isn't really deleting them)


simon-eller:
### ⚠️ Before posting ⚠️

- [X] This is a **bug**, not a question or an enhancement.
- [X] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [X] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

Configure auto upload in the Nextcloud app for the camera folder of the smartphone
Shoot a photo and wait until it's uploaded on the Nextcoud server
Delete the photo in the Samsung gallery app
Go to your Nextloud app and click on "show hidden files" option
Now you can see that the photo was uploaded twice

### Expected behaviour

If the photo was deleted on the smartphone, nothing should actually happen with the current options on the Nextcloud itself.

Whether the photo should then also be deleted on the Nextcloud has already been discussed in [a number of other issues](https://github.com/nextcloud/android/issues/5225).

### Actual behaviour

If you delete a photo in Samsung Gallery, it will not be deleted directly. Instead, the file is renamed and only really deleted after 30 days (because of the recycle bin function).

Original file name: 20220717_1234.jpg
Deleted file name: .trashed-<someid>-20220717_1234.jpg

The deleted file will no longer appear in Samsung Gallery, but it is still there.
The problem now is that Nextcloud thinks it is a new file and uploads this file as well.

The photo is now available twice on Nextcloud, except that the deleted version is not displayed directly because the file is classified as a hidden file by the dot at the beginning of the name.

The second version is only displayed as soon as you select the "Show hidden files" option.
![original_file_and_deleted_file](https://user-images.githubusercontent.com/63782380/185386940-8a55a224-4cb6-4c85-8e22-b28a1ecb9179.png)

### Android version

12

### Device brand and model

Samsung Galaxy A51 SM-A515F/DSN

### Stock or custom OS?

Stock

### Nextcloud android app version

3.21.0

### Nextcloud server version

24.0.3

### Using a reverse proxy?

No

### Additional information

The Nextcloud app should recognize if it is a file with the name something like .trashed<...>.jpg and not synchronize it to the server.

AlvaroBrey:
[...]

PR #14384, opened 2025-01-13
Hard-code ignore files with prefix .trashed

0ranki:

Creating initally as a draft. Help on how to write tests for this is most welcome, if necessary, as other comments too.

Fixes issue #10645 for people using Google Photos for handling their local copies of photos. Samsung Photos seems to use the same trashcan method too, though this hasn't been tried on a Samsung phone.

Files with the prefix `.trashed` are excluded from the list of files eligible for sync using the same method as is used for thumbnail files (hard-coded).

### DIFF
--- a/app/src/main/java/com/owncloud/android/utils/SyncedFolderUtils.kt
+++ b/app/src/main/java/com/owncloud/android/utils/SyncedFolderUtils.kt
@@ -27,6 +27,7 @@ object SyncedFolderUtils {
     private const val THUMBNAIL_FOLDER_PREFIX = ".thumbnail"
     private const val THUMBNAIL_DATA_FILE_PREFIX = ".thumbdata"
     private const val SINGLE_FILE = 1
+    private const val EXTERNAL_TRASHCAN_PREFIX = ".trashed"
 
     /**
      * analyzes a given media folder if its content qualifies for the folder to be handled as a media folder.
@@ -129,7 +130,8 @@ object SyncedFolderUtils {
         return when {
             fileName != null -> {
                 !DISQUALIFIED_MEDIA_DETECTION_FILE_SET.contains(fileName.lowercase()) &&
-                    !fileName.startsWith(THUMBNAIL_DATA_FILE_PREFIX)
+                    !fileName.startsWith(THUMBNAIL_DATA_FILE_PREFIX) &&
+                    !fileName.startsWith(EXTERNAL_TRASHCAN_PREFIX)
             }
             else -> false
         }
--- a/app/src/main/java/com/owncloud/android/utils/SyncedFolderUtils.kt
+++ b/app/src/main/java/com/owncloud/android/utils/SyncedFolderUtils.kt
@@ -24,9 +24,9 @@ object SyncedFolderUtils {
     )
     private val DISQUALIFIED_MEDIA_DETECTION_FILE_SET: Set<String> = DISQUALIFIED_MEDIA_DETECTION_SOURCE.toSet()
     private val AUTO_QUALIFYING_FOLDER_TYPE_SET: Set<MediaFolderType> = setOf(MediaFolderType.CUSTOM)
-    private const val THUMBNAIL_FOLDER_PREFIX = ".thumbnail"
     private const val SINGLE_FILE = 1
-    private val EXCLUDE_FILENAME_PREFIXES = listOf(
+    private val EXCLUDE_PREFIXES = listOf(
+        ".thumbnail",
         ".thumbdata",
         ".trashed",
         ".pending",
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/utils/SyncedFolderUtils.kt
```diff
@@ -27,6 +27,7 @@ object SyncedFolderUtils {
     private const val THUMBNAIL_FOLDER_PREFIX = ".thumbnail"
     private const val THUMBNAIL_DATA_FILE_PREFIX = ".thumbdata"
     private const val SINGLE_FILE = 1
+    private const val EXTERNAL_TRASHCAN_PREFIX = ".trashed"
```
Aohzan, line 30:
```suggestion
    private const val EXTERNAL_TRASHCAN_PREFIX = ".trashed"
    private const val EXTERNAL_PENDING_PREFIX = ".pending"
```

app/src/main/java/com/owncloud/android/utils/SyncedFolderUtils.kt
```diff
@@ -129,7 +130,8 @@ object SyncedFolderUtils {
         return when {
             fileName != null -> {
                 !DISQUALIFIED_MEDIA_DETECTION_FILE_SET.contains(fileName.lowercase()) &&
-                    !fileName.startsWith(THUMBNAIL_DATA_FILE_PREFIX)
+                    !fileName.startsWith(THUMBNAIL_DATA_FILE_PREFIX) &&
+                    !fileName.startsWith(EXTERNAL_TRASHCAN_PREFIX)
```
Aohzan, line 134:
```suggestion
                    !fileName.startsWith(EXTERNAL_TRASHCAN_PREFIX) &&
                    !fileName.startsWith(EXTERNAL_PENDING_PREFIX)
```
Can you add `.pending` file exclusion too

tobiasKaminsky:
@alperozturk96 I have the same problem with ".nomedia".
I think this is because the logic of subfolder detection changed?

0ranki:
@tobiasKaminsky @Aohzan Sure, I'll include `.nomedia` and `.pending` to the list. What app creates those `.pending` files, by the way?

app/src/main/java/com/owncloud/android/utils/SyncedFolderUtils.kt
0ranki: As there are already 4 prefixes, instead of a const for each defined them in a list

0ranki:
> For `.pending`, I think it's done by Google Photos, I don't have any other app relating with Camera dir 🤔 , I see some discussion about those files and people owning Pixel phones. I have the issue with two Pixel myself

Ok, I have a Pixel too, but don't have Google stuff on it. Well, right now I do to test the trashcan stuff, but no sync to Google.

I did write a rudimentary unit test for the added `hasExcludePrefix` function, which isn't included yet. Should the branch be rebased to current `master`?

tobiasKaminsky:
Thanks for your help!

I think this should also be added to SyncedFolderUtils -> isQualifiedFolder (L108), as then it ignores all folders with these names.

In my case, I only have an empty ".nomedia" folder, but no files in it.

What do you think?

0ranki:
> Thanks for your help!
> 
> I think this should also be added to SyncedFolderUtils -> isQualifiedFolder (L108), as then it ignores all folders with these names.
> 
> In my case, I only have an empty ".nomedia" folder, but no files in it.
> 
> What do you think?

I'd guess folders with those prefixes would be safe to ignore too, so `.thumbdata` could be added to the same list and the same function used for both

0ranki:
SyncedFolderUtils -> hasExludePrefix is now used in SyncedFolderUtils -> isQualifiedFolder too.

Added a simple unit test.

app/src/main/java/com/owncloud/android/utils/SyncedFolderUtils.kt
```diff
@@ -155,18 +155,18 @@ object SyncedFolderUtils {
...
      */
-    private fun hasExcludePrefix(fileName: String?): Boolean {
-        if (fileName == null) {
-            return false
-        }
-        EXCLUDE_FILENAME_PREFIXES.forEach {
-            if (fileName.startsWith(it)) {
-                return true
+    fun hasExcludePrefix(name: String?): Boolean {
+        val hasPrefix = false
```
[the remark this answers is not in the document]
0ranki: If you're referring to the boolean, yeah, that was already removed in 46feea9. Sorry for the trouble, noticed it right after pushing 🙃

tobiasKaminsky (approved):

tobiasKaminsky:
@0ranki thanks for your help :blue_heart:

PR #14410, opened 2025-01-15
Don't Show Failed Upload Message For Same File and Handle Local Behavior

alperozturk96:

• Removes messages related to failed uploads
• Manages local behavior in cases of upload synchronization conflicts and when the remote directory contains the same file


bilogic:
thanks for the PR, it solves the problem, but because it is another fresh installation, I cannot re-use my config from the GA version or the DEV version and had to reconfigure the entire app from scratch...

alperozturk96:
@bilogic Thank you for the message. That might be a some other problem. Since it's not related with this PR. Please create a separate issue. So we can continue to discuss from that issue.

PR #14432, opened 2025-01-20
Show auto upload notification in all activities

ZetaTom:
This change amends previous changes to the *Changes to auto upload* notification. Previously the notification would sometimes not be displayed when an activity, other than the `FileDisplayActivity` was in foreground.

---


alperozturk96 (approved):
https://github.com/user-attachments/assets/5c69638b-5277-4d13-9318-151efe58c78b

PR #14461, opened 2025-01-24
Remove Login Flow v1

alperozturk96:

Android 8

https://github.com/user-attachments/assets/4a6fcf38-e536-4492-b543-f2df2b2bc78e

Android 15

https://github.com/user-attachments/assets/7a2731fb-3a7a-41c4-a568-268ec4793ce2

### DIFF
--- a/app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -24,25 +24,17 @@ import android.content.Intent;
 import android.content.ServiceConnection;
 import android.content.SharedPreferences;
 import android.content.pm.PackageManager;
-import android.graphics.Bitmap;
-import android.graphics.Color;
 import android.net.Uri;
-import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.IBinder;
 import android.preference.PreferenceManager;
 import android.text.TextUtils;
-import android.util.AndroidRuntimeException;
 import android.view.KeyEvent;
 import android.view.View;
 import android.view.inputmethod.EditorInfo;
 import android.webkit.CookieManager;
-import android.webkit.CookieSyncManager;
 import android.webkit.URLUtil;
-import android.webkit.WebResourceRequest;
-import android.webkit.WebResourceResponse;
-import android.webkit.WebView;
 import android.widget.AdapterView;
 import android.widget.ArrayAdapter;
 import android.widget.LinearLayout;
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -253,9 +225,6 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
     private String token;
 
     private boolean onlyAdd = false;
-    @SuppressLint("ResourceAsColor") @ColorInt
-    private int primaryColor = R.color.primary;
```
alperozturk96, line 257: Unused variable after removing the login flow v1

app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1043,24 +895,6 @@ private void onGetServerInfoFinish(RemoteOperationResult result) {
                 webViewPassword != null && !webViewPassword.isEmpty()) {
                 checkBasicAuthorization(webViewUser, webViewPassword);
             } else {
-                new Thread(() -> {
```
alperozturk96, line 1046: This remote operation only used for the setting the primaryColor. Since primaryColor is not used no need for this remote operation

app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -421,9 +384,8 @@ public void onNothingSelected(AdapterView<?> parent) {
 
     private void deleteCookies() {
         try {
-            CookieSyncManager.createInstance(this);
```
alperozturk96, line 424: No need to use CookieSyncManager. [Document](https://developer.android.com/reference/android/webkit/CookieManager.html)

app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -421,9 +384,8 @@ public void onNothingSelected(AdapterView<?> parent) {
...
             CookieManager.getInstance().removeAllCookies(null);
-        } catch (AndroidRuntimeException e) {
```
alperozturk96, line 426: On Android 8, if the device has WebView Shell as the default web browser.

`CookieManager.getInstance().removeAllCookies(null);`

This function throw the IllegalArgumentException.

app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
tobiasKaminsky: Delete it? :)

tobiasKaminsky (approved):

PR #14493, opened 2025-02-04
Open "on device" correctly when coming from "Uploads"

tobiasKaminsky:

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -589,7 +589,18 @@ public abstract class DrawerActivity extends ToolbarActivity
             startPhotoSearch(menuItem.getItemId());
         } else if (itemId == R.id.nav_on_device) {
             EventBus.getDefault().post(new ChangeMenuEvent());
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -589,7 +589,18 @@ private void onNavigationItemClicked(final MenuItem menuItem) {
             startPhotoSearch(menuItem.getItemId());
         } else if (itemId == R.id.nav_on_device) {
             EventBus.getDefault().post(new ChangeMenuEvent());
-            showFiles(true, false);
```
alperozturk96, line 592: `showFiles(true, false);` is doing exactly the same thing. I think the issue happens sometimes and sometimes not, which might be why you are unable to reproduce it with this PR.

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
tobiasKaminsky: Indeed. So not a solution.

PR #14501, opened 2025-02-05
Fix Image Scaling/Rotating

alexknop:
Problem:

Application is opening some images incorrectly. The image is being rotated incorrectly and the wrong heights/widths are being used for scaling.

For example, this image with a Width of 4096 pixels and Height of 3072 pixels and an orientation of upper-right was showing like this:
image
My screen size is 1080x2460

It appears after BitmapUtils.decodeSampledBitmapFromFile() is run, that the orientation is changed successfully. 

The problem then was the height/width used. When originally using my screen size, the scaling was affected by stretching the image to fit the height of my screen. It needs to maintain the original aspect ratio of the image.

The image was then being rotated again, thus resulting in the image above.

Solution:
Take original heights and widths of file.
Determine if this image is to be rotated 90 or 270 degrees. If so, the height and width needs to accommodate for that.
Now, decode the bitmap using the original image's height/width.
Next, scale the image based on your screen size.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/preview/PreviewImageFragment.kt
+++ b/app/src/main/java/com/owncloud/android/ui/preview/PreviewImageFragment.kt
@@ -64,7 +64,6 @@ import com.owncloud.android.ui.fragment.FileFragment
 import com.owncloud.android.ui.preview.PreviewMediaFragment.Companion.newInstance
 import com.owncloud.android.utils.BitmapUtils
 import com.owncloud.android.utils.DisplayUtils
-import com.owncloud.android.utils.MimeType
 import com.owncloud.android.utils.MimeTypeUtil
 import com.owncloud.android.utils.theme.ViewThemeUtils
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/ui/preview/PreviewImageFragment.kt
```diff
@@ -529,12 +529,37 @@ class PreviewImageFragment : FileFragment(), Injectable {
                         }
 
                         try {
+                            // Get the original dimensions of the bitmap
+                            val bitmapResolution = BitmapUtils.getImageResolution(storagePath);
```
alperozturk96, line 533: Could you extract whole logic to the BitmapUtils class and just call the result of it from the PreviewImageFragment?

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/ui/preview/PreviewImageFragment.kt
alexknop: Yes. Done

alperozturk96 (approved):
I tested various images with different sizes, including both landscape and portrait orientations. Throughout the testing process, I did not encounter any issues.

Thank you for the PR 💯

alperozturk96:
@alexknop Could you fix the spotlessKotlinCheck?

`./gradlew spotlessApply` This command can help.

Thank you.

alexknop:
@alperozturk96 Done, I believe.

Thanks.

tobiasKaminsky (approved):

PR #14502, opened 2025-02-05
Fix Image Loading When Offline

### LINKED ISSUE #14470: Cannot Open Offline Photos After Video
### ⚠️ Before posting ⚠️

### Steps to reproduce

1. Download a folder with photos and videos offline
2. Turn off Internet connection to where server is unavailable 
3. Open a video offline
4. Try to open a photo afterwards and find app unresponsive 

### Expected behaviour

Should be able to open photos after opening video.

### Actual behaviour

When trying to open a photo nothing happens 

### Android version

15

### Device brand and model

TCL 50 XE

### Stock or custom OS?

Stock

3.30.8

### Nextcloud server version

30.0.5

### Using a reverse proxy?

No

### Android logs

[logcatOutput.txt](https://github.com/user-attachments/files/18563044/logcatOutput.txt)

### Server error logs

```bash

```

### Additional information

_No response_

alexknop:
Problem:

Although Videos were loading immediately when offline, Images were not

Solution:
Found that checking for whether file is offline was passing, but the ! in the code was making it false, forcing a sync which would timeout (taking extra time) and preventing immediate access to images

Fixes #14470

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
@@ -1225,9 +1225,9 @@ public class OCFileListFragment extends ExtendedListFragment implements
                     case GALLERY_SEARCH -> VirtualFolderType.GALLERY;
                     default -> VirtualFolderType.NONE;
                 };
-                ((FileDisplayActivity) mContainerActivity).startImagePreview(file, type, !file.isDown());
+                ((FileDisplayActivity) mContainerActivity).startImagePreview(file, type, file.isDown());
             } else {
-                ((FileDisplayActivity) mContainerActivity).startImagePreview(file, !file.isDown());
+                ((FileDisplayActivity) mContainerActivity).startImagePreview(file, file.isDown());
             }
         } else if (file.isDown() && MimeTypeUtil.isVCard(file)) {
             ((FileDisplayActivity) mContainerActivity).startContactListFragment(file);


alexknop:
It still takes a long time to load images when on cellular.

My server address is 192.168.1.2 and as I test this on a guest Wi-fi network with 192.168.1.0/24 as well, it seems Nextcloud is quicker to detect my server unreachable on here versus on Cellular. However, load times are much better when on the guest Wi-Fi and loading images due to this code change.

Videos always load immediately regardless of WiFi or Cellular

alexknop:
I dug more into it and found the timeouts for ReadFileRemoteOperation were extensive by default but you can customize the timeouts. By lowering them, the connection error happens sooner and the app responds quicker.

Images are now loading fine when on cellular and I cannot reach my server. 
Please consider these changes and lets discuss.

alexknop:
Found another area to reduce timeouts and speed up the app, that is the timeouts for OCVersionCheck.

Changed them to 1000 and app is responding better when offline.

alperozturk96:
RefreshFolderOperation may need more time. If we lower the default session time out value, the user may not get the latest result. 

Default = SessionTimeOut(60000, 15000)
PR = SessionTimeOut(1000, 1000)

@tobiasKaminsky Right?

tobiasKaminsky:
@alexknop As I see that you are actively contribution, I'd like to invite you to our org, so that you can directly create PR within nextcloud repository.
This makes contributing and reviewing easier (as then CI directly runs).

alexknop:
> @alexknop As I see that you are actively contribution, I'd like to invite you to our org, so that you can directly create PR within nextcloud repository. This makes contributing and reviewing easier (as then CI directly runs).

@tobiasKaminsky I was actually invited into the org yesterday. I will be doing my other PRs directly from here on out. 
What do you think about my proposed changes?

tobiasKaminsky:
Timeout to 1s is not a real solution.
I think it is okay that the first request takes longer, but then we know that we are offline and can skip some calls.
So the better/real solution is to have a correct handling when we know that we are offline.

alexknop:
> Timeout to 1s is not a real solution. I think it is okay that the first request takes longer, but then we know that we are offline and can skip some calls. So the better/real solution is to have a correct handling when we know that we are offline.

Okay, I have reverted the timeout changes and kept it to just the bug i found.
We can work on a permanent solution for these offline scenarios

alexknop:
@alperozturk96 @tobiasKaminsky can I get a look at this? Looks like a small enough fix.

alexknop:
closing in order to do this under this repo instead of my forked one.
#14733

PR #14509, opened 2025-02-06
Increase PIN Enter Delay

alperozturk96:

https://github.com/user-attachments/assets/65e16547-2e1b-4584-8d4c-34368090749b


tobiasKaminsky:
delay needs to be persistent:
- type wrong code
- see 15s delay
- close app
- open app 
- I can directly enter another code

tobiasKaminsky (changes requested):
delay needs to be persistent

tobiasKaminsky (approved):

PR #14510, opened 2025-02-06
Open Document Timeout

### LINKED ISSUE #14478: OpenDocument Has Long Timeout For Opening Files When Offline
### ⚠️ Before posting ⚠️

### Steps to reproduce

1. Download a file offline
2. Then go offline
3. Attempt to open a file from an application which would call the OpenDocument function (In my case I used an offline .kdbx file and tried to open with KeepassDX)
4. Expect to wait up to a minute for the toast message to finally say "File could not be synced. Showing latest available version."

### Expected behaviour

After a short period of time (5 sec max), offline file should just load

### Actual behaviour

In DocumentsStorageProvider.java, openDocument() calls hasServerChange() on line 210, which is returning true based on its result of ETAG_CHANGED even though the Etag could not be found. This is handled by CheckEtagRemoteOperation.class in the event of an IOException.
This then triggers needsDownload to kick off, which is unnecessary in this scenario.

### Android version

15

### Device brand and model

TCL 50 XE

### Stock or custom OS?

Stock

3.32.0

### Nextcloud server version

30.0.5

### Using a reverse proxy?

No

### Android logs

_No response_

### Server error logs

```bash

```

### Additional information

In my opinion, in hasServerChange(), a SessionTimeout object with Timeouts of 1000 should then be passed to CheckEtagRemoteOperation() to shorten the wait time from 1 minute to 1 second. 

However, if the Etag is not found, we should not be trying to download it. Possibly another connection check needs to be factored into boolean needsDownload.

alexknop:
Problem:

When offline and attempting to open a document via the Document Provider, the wait time is excessive.

CheckEtagRemoteOperation is reporting ETAG_CHANGED when the connection times out to the server, thus triggering needsDownload.

Instead, if the file is saved offline and the etag cannot be read from the server, there is no need for download.

Solution:

Implemented shorter timeouts than the default, which would take longer than is necessary.

If the etag has changed and we have a connection to the server, then result's data will be populated.
We now check for the data to catch instances where the connection timed out when checking for the etag.
If null (which results from a timeout), then we return false. This skips the download.

Fixes #14478


tobiasKaminsky:
Timeout of 1s might lead to false positive errors?

On Files app we know if the app is offline, so maybe we can re-use this?
E.g. we have NetworkChangeReceiver

What do you think?

alexknop:
I do not see why not, but in terms of implementing I do not have enough knowledge of this app's architecture.
It seems like a good check to run and act on accordingly prior to determining whether to run hasServerChange.

I was simply bug fixing in noticing offline files take an egregious amount of time to load, and noticing the return value for the eTAG played a role in it. I do think timeouts can be reduced, in this age 1 second is a long time to reach your server, but I understand wanting a bit of a safeguard. I thought this function also retried a few times prior to returning a value, which is a safeguard in itself.

Hope this helps, but apologies for not being able to carry this idea out.

tobiasKaminsky:
Would it be okay for you to remove the timout changes, but keep the other?
Then we can merge it.
And in another issue we can discuss better offline handling?

alexknop:
closing this so I dont have to use my forked repo.
Opened #14734

Issue #14437, opened 2025-01-21
Thumbnail images not showing in the "Shared" tab


buidanhnam:
### ⚠️ Before posting ⚠️

- [x] This is a **bug**, not a question or an enhancement.
- [x] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [x] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

Steps to Reproduce:
1) Open the app.
2) Navigate to the "Shared" tab.

https://github.com/user-attachments/assets/9e879f10-c224-4561-9968-b1455aef9d10

### Expected behaviour

Thumbnail images should display normally, similar to other sections.

### Actual behaviour

Thumbnail images do not display in the "Shared" tab. I believe the issue is caused by cached thumbnail files.

### Android version

12

### Device brand and model

Samsung Note 20

### Stock or custom OS?

Stock

### Nextcloud android app version

30.0.8 final

### Nextcloud server version

30.0.5

### Using a reverse proxy?

I don't know

### Server error logs

```bash

```

tobiasKaminsky:
Server API is here: https://github.com/nextcloud/server/pull/50870

@ZetaTom can you check if this works for you?

PR #14513, opened 2025-02-07
BugFix - Use New Thumbnail Key for Thumbnail Generation

alperozturk96:

The `OCFile` `remoteId` is currently being used as the `imageKey` for thumbnail generation. This approach prevents the correct regeneration of thumbnails when a user uploads a different file with the same filename and extension.  

### Example:  
1. User uploads **aa.jpg** (479 KB) → Thumbnail is generated.  
2. User replaces it with **aa.jpg** (2124 KB) → Thumbnail is not regenerated because the `remoteId` remains unchanged.

### Proposed Solution:  
To ensure proper thumbnail generation, the `getThumbnailKey()` should be used as the `imageKey` instead of the `remoteId`.

### Known Issues:
The `ConflictResolveDialog` is currently unable to display thumbnails correctly in both the master branch and this PR. This issue will be addressed separately in a future PR.

**Before**

https://github.com/user-attachments/assets/61e24530-bf05-423d-848c-d3056c915ecb

**After**

https://github.com/user-attachments/assets/46df7a53-a596-4781-9579-7c5f999421e4

### DIFF
--- a/app/src/androidTest/java/com/owncloud/android/ui/fragment/GalleryFragmentIT.kt
+++ b/app/src/androidTest/java/com/owncloud/android/ui/fragment/GalleryFragmentIT.kt
@@ -80,6 +80,7 @@ class GalleryFragmentIT : AbstractIT() {
         val defaultSize = ThumbnailsCacheManager.getThumbnailDimension().toFloat()
         val file = OCFile("/$id.png").apply {
             fileId = id.toLong()
+            fileLength = (0..100).random().toLong()
             remoteId = "$id"
             mimeType = "image/png"
             isPreviewAvailable = true
--- a/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
@@ -272,8 +272,8 @@ class FileUploadWorker(
 
             val task = ThumbnailsCacheManager.ThumbnailGenerationTask(storageManager, user)
             val file = File(uploadFileOperation.originalStoragePath)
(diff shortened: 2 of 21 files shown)


tobiasKaminsky:
I think we need to discuss this approach.
First "problem" that comes to my mind:
- create an image
- let Android generate thumbnail with etag "123"
- rename file
--> new etag
-> thus new thumbnail

alperozturk96:
> I think we need to discuss this approach. First "problem" that comes to my mind:
> 
> * create an image
> * let Android generate thumbnail with etag "123"
> * rename file
>   --> new etag
>   -> thus new thumbnail

Exactly. In that case, it will regenerate the thumbnail. We can evaluate the pros and cons of this approach. If the advantages outweigh the disadvantages, we can proceed with this method. Alternatively, we can try to find a more distinct and reliable solution, maybe using a combination of `remoteId`, `size`, and `hash`, which may have fewer side effects.

ZetaTom:
Using only `remoteId` as a key for our thumbnail cache also causes other problems. The *Shared* tab is notoriously difficult to work with, as it solely relies on the data returned by the sharing API. In this case, `remoteId` is no set, which is why the default placeholder thumbnail is displayed instead. (See #14437) Because of this, I think we should re-evaluate what key to use.

alperozturk96:
@tobiasKaminsky @ZetaTom I updated the PR and added the tests. Please review. Now it's using fileLenght + hashCode().

alperozturk96:
I discussed with @tobiasKaminsky . `isUpdateThumbnailNeeded()` will be used for solving this problem.

alperozturk96:
https://github.com/nextcloud/android/pull/14556

PR #14546, opened 2025-02-13
BugFix - Don't Set Icon If Thumbnail Exists & Handle Thumbnails Properly

alperozturk96:

### Issue

1. After uploading an image, the thumbnail appears correctly. However, if the user downloads the image, the thumbnail disappears, and they must refresh the page to see it again.  

2. The offline operation thumbnail generation logic exists in `OCFileListAdapter.configureThumbnail()`. However, `holder.getThumbnail()` shares the same `ImageView` for both regular and offline files, causing re-set image for the regular files. To resolve this, thumbnail generation for offline operations should be handled in `DisplayUtils`, ensuring reusability across different lists while preventing unnecessary image updates for regular files.

### Changes

- Fixes unnecessary thumbnail set for regular files
- Handles thumbnail logic for the offline operations in the DisplayUtils just like for the regular files

### Demo

https://github.com/user-attachments/assets/1605784b-a984-426e-b704-bf2eba69b2e9


app/screenshots/gplay/debug/com.owncloud.android.ui.dialog.DialogFragmentIT_testFileActionsBottomSheet.png
alperozturk96, line 1: Retry option should not be visible because this file is not an offline file.

PR #14547, opened 2025-02-13
Feature - Check Auto Upload Sub Folder

alperozturk96:

Screenshot 2025-03-12 at 10 20 03


tobiasKaminsky:
as discussed
- do not automatically change it
- show an info it auto upload overview

jancborchardt (changes requested):
In both cases where it appears – but especially when clicking the info "i" icon – it would be more fitting to use a modal since the info is more important than to just show it in a toast.

app/src/main/java/com/owncloud/android/ui/dialog/ConfirmationDialogFragment.kt
```diff
@@ -69,10 +74,21 @@ open class ConfirmationDialogFragment : DialogFragment(), Injectable {
...
 
         val builder = MaterialAlertDialogBuilder(requireActivity())
-            .setIcon(com.owncloud.android.R.drawable.ic_warning)
-            .setIconAttribute(R.attr.alertDialogIcon)
             .setMessage(message)
 
+        if (titleIconId == defaultTitleIconId) {
+            builder
+                .setIcon(titleIconId)
+                .setIconAttribute(R.attr.alertDialogIcon)
```
alperozturk96, line 82: Only setIconAttribute for defaultTitleIconId otherwise it will override given titleIconId

tobiasKaminsky (approved):

carlhaupt:
its back again with version 3.33.0 of android com.nextcloud.client. I cannot use the camera-folder from actual camera aperture only because I have opencamera and others which store in DCIM. and it says there will be doublettes because there is an upload which saves /camera alread als subfolder, but I cannot see the definition of the main /DCIM-Folder which is mentioned. There are 2 or even more definitions of subfolders with the same path, which i can hide but do all the same. /download/DCIMALT for example from a former backup of the old phone is twice in the list,  the same with others. And no delete-Button for those, just hide-option, in my case 6 or 7 hidden folders already, 

I haven't tried to reset the client to get rid of those, cause chatgpt thought its the best to inform you this way that the bug is back again

plz fix again. thanks

PR #14557, opened 2025-02-17
BugFix - Update Recommended File List When Needed

alperozturk96:

### How to reproduce?

1. Have Nextcloud v31
2. Enable recommendations app
3. Create a bunch of new files and folder
4. Rename file
5. Recommended file list not updated

https://github.com/user-attachments/assets/a7df1587-ffe3-44d0-bf20-fa04d7df669a

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -1823,11 +1823,13 @@ public class FileDisplayActivity extends FileActivity
             onRestoreFileVersionOperationFinish(result);
         }
 
-        OCFileListFragment fileListFragment =
-            (ActivityExtensionsKt.lastFragment(this) instanceof OCFileListFragment fragment) ? fragment : getListOfFilesFragment();
+        if (operation instanceof RemoveFileOperation || operation instanceof RenameFileOperation) {
+            OCFileListFragment fileListFragment =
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1822,6 +1822,15 @@ public void onRemoteOperationFinish(RemoteOperation operation, RemoteOperationRe
         } else if (operation instanceof RestoreFileVersionRemoteOperation) {
             onRestoreFileVersionOperationFinish(result);
         }
+
+        if (operation instanceof RemoveFileOperation || operation instanceof RenameFileOperation) {
```
tobiasKaminsky, line 1826: Can you test if this is also needed for
- create folder
- upload file
- move file/folder

alperozturk96: Not needed for "create folder operation" because when we create a new folder, app automatically redirecting to the newly created folder. Thus, when the user goes back to the root directory, the recommended file list will be updated.

Not needed for "move file/folder operation" because recommended file list will not change. Also, when the user taps recommended file after move it will redirect to the correct directory.

Note: If I create a new file from web or client, that newly created file not appears in the recommended file list. I will check with `authors` of the recommendations app.

I would say it's okay to keep it like this.

ZetaTom (approved):

PR #14561, opened 2025-02-19
Prepare for build with *.kts files

hannesa2:
It's just about a preparation for kts build files


hannesa2:
Gradle crashes ! 
image

tobiasKaminsky:
welcome back :wave: 

Can you explain a bit why the changes from ' to " are needed for kts builds?

hannesa2:
kts needs `"` There is a `'` not valid

appscan/.gitignore
```diff
@@ -1,3 +0,0 @@
-# SPDX-FileCopyrightText: 2023-2024 Nextcloud GmbH and Nextcloud contributors
-# SPDX-License-Identifier: AGPL-3.0-or-later OR GPL-2.0-only
```
tobiasKaminsky, line 2: I think we should not remove it?

hannesa2: I at least see no deeper sense in it in a `.gitignore`

AndyScherzinger: Strange, reuse demands SPDX header in all files in a repo but the CI check didn't fail.

@hannesa2 reuse specifies the need to have a license and copyright info for any file within a repo, which would also include the ignore file (or any other file that seems less needed to have such an info)

hannesa2: Did you see, that i removed the complete file ?
The only significant content `/build` is already in an other file twice
https://github.com/nextcloud/android/blob/master/.gitignore#L16
and here
https://github.com/nextcloud/android/blob/master/.gitignore#L36

hannesa2:
@AndyScherzinger 
> Strange, reuse demands SPDX header in all files in a repo but the CI check didn't fail.

I deleted this file, that's why (I guess) it didn't find anything.

In my point of view these two errors are more a problem
https://github.com/nextcloud/android/actions/runs/13537380915/job/37831391867
https://github.com/nextcloud/android/actions/runs/13537380909/job/37831391864

AndyScherzinger:
> I deleted this file, that's why (I guess) it didn't find anything.

Ah, right, that explains it. I missed the deletion... @hannesa2 

> In my point of view these two errors are more a problem
> https://github.com/nextcloud/android/actions/runs/13537380915/job/37831391867
> https://github.com/nextcloud/android/actions/runs/13537380909/job/37831391864

Yes, one seems broken, the other seems to have an infra issue I suppose

hannesa2:
> Yes, one seems broken, the other seems to have an infra issue I suppose

@AndyScherzinger If I would be a maintainer, I would delete this actions and add them as a new pull request to fix them. With this, the pipeline would raise no errors anymore and you could concentrate to fix them separate

AndyScherzinger:
@tobiasKaminsky looping you in on the matter of the broken GH action for the lib-version check

> @AndyScherzinger If I would be a maintainer, I would delete this actions and add them as a new pull request to fix them. With this, the pipeline would raise no errors anymore and you could concentrate to fix them separate

AndyScherzinger:
@tobiasKaminsky I raised the number of hashes to 20. The issue though is that the check runs on the lib project and fetches the last 10 commit SHAs (now 20), if there hasn't been a lib update to master branch on android, not android-lib that is within these 10 commits, than the check will always fail. I think it would be more reliable to just check if the lib version in the gradle file of the app branch matches the one from master branch.

AndyScherzinger (approved):

appscan/.gitignore
AndyScherzinger: I am also in favor of removing it, having a single top-level ignore file helps to have a global overview on the ignores SO I am all for that way (both are generally okay, so I think it is a matter of preference)

AndyScherzinger:
@hannesa2 all changes looked good and CI was fine, except for some flaky test on drone, so I merged 👍

PR #14581, opened 2025-02-24
remove methods that only call super

alexknop:
Problem:

Identified functions that were simply just calling super.

Solution:

Removed those functions to clean up code.


AndyScherzinger (approved):
Looks good to me, any reason for the ktlint XML changes in the PR?

AndyScherzinger:
@alexknop see DCO check, we run a check for all git commits to be signed. You might want to look into that 😊
We can manually waive but in the long run it would be good if commits are signed.

alexknop:
@AndyScherzinger fixed both. That xml was changed when running inspect code in android studio but does not seem necessary at this time.

PR #14582, opened 2025-02-24
made functions void where their return values were never used

alexknop:
Problem:
There were functions that were declared to return values, but when used in the project, there was no use of the return values.

Solution:
Change these functions to void for now.


tobiasKaminsky (approved):

tobiasKaminsky:
Thanks for your PR :tada: 
Can you check and fix Spotbugs?

alexknop:
> Thanks for your PR 🎉 Can you check and fix Spotbugs?

Could use help in how to check that...
I'm seeing some in the Analysis tab of the Checks, but not sure what is related to this branch...as in what would have increased.

alexknop:
> Spotbugs has +1 issue. Best is two open both html files side by side and check for diff. Unfortunately there is no other easy way.
> 
> ```
> DB: Duplicate Branches (1: 0/1/0/0)
> Method uses the same code for two branches   (1: 0/1/0/0)
>  In class com.owncloud.android.datamodel.UploadsStorageManager
> In method com.owncloud.android.datamodel.UploadsStorageManager.updateDatabaseUploadResult(RemoteOperationResult, UploadFileOperation)
> At UploadsStorageManager.java:[line 802]
> At UploadsStorageManager.java:[line 805]
> com.owncloud.android.datamodel.UploadsStorageManager.updateDatabaseUploadResult(RemoteOperationResult, UploadFileOperation) uses the same code for two branches
> ```
> 
> So it seems that your change made then / else branch to be the same, thus not needed at all.

Fixed!

alexknop:
@tobiasKaminsky @AndyScherzinger There has been a bit of delay in my PRs. I dont want to bog you guys down with change requests, but for something small like this to be open since Feb, it's a bit discouraging to want to contribute more. Hopefully something can be done to speed up this process.

alperozturk96 (approved):

tobiasKaminsky:
Thank you @alexknop :tada:

PR #14584, opened 2025-02-24
Removal of Optional parameters and fields

alexknop:
Problem:

Use of Optional tags here is unnecessary and makes code more complicated.

Solution:

Remove optional tags and handle where user can just be null.


alexknop:
VOID

PR #14586, opened 2025-02-24
Fix control flow errors/warnings

alexknop:
Problem:

Various issues in control flow statements and their bodies were found

Solution:

Address each one as needed: moving default blocks to the bottom, removing redundant lines in if/else blocks, variables that can be declared without the if/else statement, running do/while in the case where the first character is not alphanumeric and the while loop would not run to increase the position.

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
@@ -1224,10 +1224,6 @@ public class UploadFileOperation extends SyncOperation {
                                       OwnCloudClient client) {
         switch (mLocalBehaviour) {
             case FileUploadWorker.LOCAL_BEHAVIOUR_FORGET:
-            default:
-                mFile.setStoragePath("");
-                saveUploadedFile(client);
-                break;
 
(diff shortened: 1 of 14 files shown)


tobiasKaminsky:
```
com.owncloud.android.UploadIT > testUploadWithForget[android(AVD) - 8.1.0] FAILED 
	junit.framework.AssertionFailedError
	at junit.framework.Assert.fail(Assert.java:48)
```

fails now. 
I will test locally.

app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -1225,10 +1225,6 @@ private void handleLocalBehaviour(File temporalFile,
                                       OwnCloudClient client) {
         switch (mLocalBehaviour) {
             case FileUploadWorker.LOCAL_BEHAVIOUR_FORGET:
-            default:
```
tobiasKaminsky, line 1228: This is the problematic change.
After change LOCAL_BEHAVIOUR_FORGET and LOCAL_BEHAVIOUR_DELETE are the same.

tobiasKaminsky (changes requested):

alexknop (commented):
complete

PR #14625, opened 2025-03-03
Remove Airplane mode Check Due to Configurable Unmetered Wi-Fi

alperozturk96:

Some users prefer not to initiate uploads while their device is in airplane mode. For instance, a user may have Wi-Fi during a flight to browse the internet, but triggering an auto-upload in such cases would be undesirable due to the typically slow and limited in-flight Wi-Fi.

After discussing with the team, we have determined that since Android 9, users have the ability to designate Wi-Fi networks as metered or unmetered. 

Image 1

Given that we already offer an option to upload only on unmetered Wi-Fi, there is no longer a need to check for airplane mode.

Image 2

As a result, we will remove the airplane mode check. If a user wishes to prevent auto-uploads when connected to a specific Wi-Fi network—such as in airplane mode—they can simply configure the network as metered in the Wi-Fi settings.

This approach supports both use cases—allowing uploads in airplane mode by default, while giving users the option to disable auto-uploads by setting the network as metered.

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
@@ -9,7 +9,6 @@ package com.nextcloud.client.jobs.upload
 
 import android.app.PendingIntent
 import android.content.Context
-import android.provider.Settings
 import androidx.localbroadcastmanager.content.LocalBroadcastManager
 import androidx.work.Worker
 import androidx.work.WorkerParameters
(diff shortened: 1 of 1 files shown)


ZetaTom (approved):

tobiasKaminsky (approved):

infinitytec:
Any ETA for when this fix will be in the Play Store?

alperozturk96:
The issue has been resolved with https://github.com/nextcloud/android/releases/tag/rc-3.31.1-01. RC1 version also released on Play Store.

PR #14630, opened 2025-03-04
Fix ecosystem app switcher spacing

ZetaTom:
This change introduces a more robust layout for the ecosystem app switcher present in the navigation drawer. The previous implementation made some assumptions which don't work well for  small screens.

### Screenshots

Currently | This Pull Request
:---: | :---:
old-narrow | new-narrow
old-wide | new-wide

---


alperozturk96 (approved):
Tested seems like ok.

PR #14646, opened 2025-03-07
showDirectoryWhenDeletionCompleted set to false?

tobiasKaminsky:

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/dialog/RemoveFilesDialogFragment.kt
+++ b/app/src/main/java/com/owncloud/android/ui/dialog/RemoveFilesDialogFragment.kt
@@ -24,7 +24,6 @@ import com.owncloud.android.datamodel.OCFile
 import com.owncloud.android.ui.activity.FileActivity
 import com.owncloud.android.ui.activity.FileDisplayActivity
 import com.owncloud.android.ui.dialog.ConfirmationDialogFragment.ConfirmationDialogFragmentListener
-import com.owncloud.android.ui.preview.PreviewImageActivity
 import javax.inject.Inject
 
 /**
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
@alperozturk96 I am not sure why this boolean is needed at all?

old behaviour
- have multible images in a folder
- click on one, to open slide show
- delete one
--> folder listing is shown

now:
- have multible images in a folder
- click on one, to open slide show
- delete one
--> another image is shown

I also tested if last image is deleted, then the empty directory is shown

alperozturk96:
> @alperozturk96 I am not sure why this boolean is needed at all?
> 
> old behaviour
> 
> * have multible images in a folder
> * click on one, to open slide show
> * delete one
>   --> folder listing is shown
> 
> now:
> 
> * have multible images in a folder
> * click on one, to open slide show
> * delete one
>   --> another image is shown
> 
> I also tested if last image is deleted, then the empty directory is shown

**PreviewImageActivity** extends from **FileActivity**, so the delete logic in **RemoveFilesDialogFragment** was not applied when deleting an image from the preview. I have addressed in this PR (https://github.com/nextcloud/android/pull/14498).  

I thought this was not the expected behavior. Since the user is in a folder rather than a gallery, I would expect the flow to return directly to the file list after deleting an image, rather than continuing to view other images in the folder.  

If the expected behavior is to continue viewing other images, you can remove the `showDirectoryWhenDeletionCompleted()` function and the `showDirectoryWhenDeletionCompleted` boolean variable.

tobiasKaminsky:
It is the expected (and previous behaviour), thus I will remove it.
Thanks for the confirmation :+1:

alperozturk96 (approved):

PR #14659, opened 2025-03-10
BugFix - Auto Rename Creating New folder

alperozturk96:

https://github.com/user-attachments/assets/98574b8e-2f31-4f0e-bb9c-13c9efe0798d

### DIFF
--- a/app/src/androidTest/java/com/nextcloud/utils/AutoRenameTests.kt
+++ b/app/src/androidTest/java/com/nextcloud/utils/AutoRenameTests.kt
@@ -206,4 +206,28 @@ class AutoRenameTests : AbstractOnServerIT() {
         val expectedFilename = "a_a.jpg"
         assert(result == expectedFilename) { "Expected $expectedFilename but got $result" }
     }
+
+    @Test
+    fun testRemovingLeadingWhitespace() {
+        val filename = " readme.txt"
+        val result = AutoRename.rename(filename, capability)
+        val expectedFilename = "readme.txt"
+        assert(result == expectedFilename) { "Expected $expectedFilename but got $result" }
+    }
+
+    @Test
+    fun testRemovingTrailingWhitespace() {
+        val filename = "readme.txt "
+        val result = AutoRename.rename(filename, capability)
+        val expectedFilename = "readme.txt"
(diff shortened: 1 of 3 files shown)


ZetaTom (commented):
During testing I found the error messages a bit misleading sometimes. Consider the following setup:

1. Create a folder called *A* in the root directory
2. Create another folder called *A⎵*
    - will fail because the stripped filename will be *A*
    - *You are not permitted to create this file* message appears
3. Create another folder called *A.part⎵*
    - will fail because the stripped filename will also be *A*
    - *Folder already exists* message appears

I'm not sure if we should not silently rename folders. We should at least provide an appropriate error message informing the user of what went wrong.

Otherwise it seems to work. I'm unable to create folders containing leading or trailing spaces, illegal characters and extensions.

alperozturk96:
> During testing I found the error messages a bit misleading sometimes. Consider the following setup:
> 
> 1. Create a folder called _A_ in the root directory
> 2. Create another folder called _A⎵_
>    
>    * will fail because the stripped filename will be _A_
>    * _You are not permitted to create this file_ message appears
> 3. Create another folder called _A.part⎵_
>    
>    * will fail because the stripped filename will also be _A_
>    * _Folder already exists_ message appears
> 
> I'm not sure if we should not silently rename folders. We should at least provide an appropriate error message informing the user of what went wrong.
> 
> Otherwise it seems to work. I'm unable to create folders containing leading or trailing spaces, illegal characters and extensions.

Please create a separate issue for this enhancement and share your suggestions. If it's necessary, assign it to the @jancborchardt

jancborchardt:
@ZetaTom sounds good – better wording in both cases would be:

> Folder was renamed to "A" to prevent synchronization issues due to characters not allowed in filename.
> A folder with the name "A" already exists

And then buttons for Cancel, Create "A (2)", and Open "A"

jancborchardt (commented):
@ZetaTom sounds good – better wording in both cases would be:

> Folder was renamed to "A" to prevent synchronization issues due to characters not allowed in filename.
> A folder with the name "A" already exists

And then buttons for Cancel, Create "A (2)", and Open "A"

PR #14670, opened 2025-03-13
BugFix - Use Previous eTag for Shared Folder If Exists

alperozturk96:

### Problem

In `SharedListFragment`, the `ReadFileRemoteOperation` is invoked, followed by `checkForChanges()` in `RefreshFolderOperation`, which also calls `ReadFileRemoteOperation`. As a result, the eTags remain unchanged, leading to an empty list being displayed. This occurs because the eTag is updated just before `checkForChanges()` is executed in `SharedListFragment`.

### How to Reproduce?

1. Create shared folder with items from web app
2. Open the Android app
3. Do not open the shared folder from the file list
4. Navigate to the shared file list from drawer menu
5. Open shared folder
6. Empty list will be visible

### Solution

In the application's database, the previous eTag is already stored. Overwriting it with the latest eTag results in an empty list being displayed, as the eTag remains unchanged. In `SharedListFragment`, the application needs to call `ReadFileRemoteOperation` to handle events; however, it does not need to update the eTag, since the previous eTag is already present in the database.

### Demo

https://github.com/user-attachments/assets/aeb1dc48-169f-442f-bdfd-02ece5386cc5

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
@@ -825,6 +825,7 @@ public class OCFileListAdapter extends RecyclerView.Adapter<RecyclerView.ViewHol
             ocFileListDelegate.setShowShareAvatar(true);
             this.user = account;
         }
+
         if (mStorageManager != null) {
             // TODO refactor filtering mechanism for mFiles
             mFiles = mStorageManager.getFolderContent(directory, onlyOnDevice);
@@ -860,7 +861,7 @@ public class OCFileListAdapter extends RecyclerView.Adapter<RecyclerView.ViewHol
         }
 
         searchType = null;
-        notifyDataSetChanged();
+        activity.runOnUiThread(this::notifyDataSetChanged);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/owncloud/android/ui/fragment/SharedListFragment.kt
```diff
@@ -99,10 +108,12 @@ class SharedListFragment : OCFileListFragment(), Injectable {
             isLoading = true
             val file = fetchFileData(partialFile)
             isLoading = false
-            if (file != null) {
-                block(file)
-            } else {
-                DisplayUtils.showSnackMessage(requireActivity(), R.string.error_retrieving_file)
+            withContext(Dispatchers.Main) {
+                if (file != null) {
+                    block(file)
```
alperozturk96, line 113: Callback should be called from the MainThread otherwise it will cause crash. e.g. `android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views. Expected: main Calling: DefaultDispatcher-worker-2`

app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
```diff
@@ -860,7 +861,7 @@ public void swapDirectory(
         }
 
         searchType = null;
-        notifyDataSetChanged();
+        activity.runOnUiThread(this::notifyDataSetChanged);
```
alperozturk96, line 864: Adapter must be updated on MainThread.

tobiasKaminsky (approved):

PR #14677, opened 2025-03-14
BugFix - AutoRename Folder With Emoji

alperozturk96:

### Problem

The logic for removing non-printable Unicode characters is causing the creation of a new folder instead of uploading the file to an existing folder.


alperozturk96:
Replaced with https://github.com/nextcloud/android/pull/14680

PR #14713, opened 2025-03-20
Migrate to Version Catalog

alperozturk96:

### [Gradle version catalogs](https://docs.gradle.org/current/userguide/platforms.html)

Given the numerous dependencies, I propose initiating the migration process in this PR. Over time, we can progressively transition all dependencies to the TOML file.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,9 +1,5 @@
-#
-# Nextcloud - Android Client
-#
-# SPDX-FileCopyrightText: 2016-2025 Nextcloud GmbH and Nextcloud contributors
-# SPDX-FileCopyrightText: 2025 Alper Ozturk <alper.ozturk@nextcloud.com>
-#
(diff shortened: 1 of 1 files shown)


alperozturk96:
@AndyScherzinger How can we fix [REUSE Compliance Check](https://github.com/nextcloud/android/actions/runs/13969058505/job/39106158216?pr=14713) in TOML file?

AndyScherzinger:
> @AndyScherzinger How can we fix [REUSE Compliance Check](https://github.com/nextcloud/android/actions/runs/13969058505/job/39106158216?pr=14713) in TOML file?

@alperozturk96 see 

https://github.com/nextcloud/android/blob/master/REUSE.toml#L1-L2

But with 2025 in your case. If that doesn't work (it should since it is a valid toml comment), at a <yourfile>.toml.license with the above header (no `#`) as content next to the toml file.

AndyScherzinger (approved):
Fine by me as long as it is ensured that renovate bot can still detect it and send update PRs

tobiasKaminsky (approved):

PR #14729, opened 2025-03-26
Feature - Add Share Permissions

alperozturk96:


app/src/main/java/com/owncloud/android/ui/fragment/FileDetailsSharingProcessFragment.kt
```diff
@@ -574,21 +678,6 @@ class FileDetailsSharingProcessFragment :
         }
     }
 
-    /**
-     *  get the permissions on the basis of selection
-     */
-    private fun getSelectedPermission() = when {
```
alperozturk96, line 580: No need this function anymore. clickEvents assigns value for the permission

mpivchev:
image

Download limit option should only be for link shares

mpivchev:
image

Share links must have File Request option.

image

mpivchev:
For folders with link share Read should be selectable:

image

image

mpivchev:
image

user1 should not have File request option

mpivchev:
﻿﻿
image

File request is written as "secure file drop", maybe just file request is enough. its like that on ios and web.

mpivchev:
Share link has a "Send new email" option that doesnt do anything?

image

mpivchev:
image

When selecting File request to email nothing appears in the quick status text.

mpivchev:
[Screen_recording_20250507_152150.webm](https://github.com/user-attachments/assets/d98f0b2c-6bd5-4ca7-959a-d606f6f2ff1c)

cant seem to add an email share to file

mpivchev:
Something weird is happening when setting share as file request:

1. When updating custom permissions, they don't update.
2. After closing share screen and opening again, pressing on secure file drop does not open the quick sheet

[Screen_recording_20250507_153831.webm](https://github.com/user-attachments/assets/b07a788f-1767-4dca-a788-c90c2ccc98fc)

alperozturk96:
> image
> 
> File request is written as "secure file drop", maybe just file request is enough. its like that on ios and web.

@tobiasKaminsky 

What's the difference between fileDrop (file request) and secureFileDrop? SecureFileDrop only has extra READ_PERMISSON_FLAG. It cannot be related with e2ee, no?

mpivchev:
image

View only is marked as Custom for links, i assume because of https://github.com/nextcloud/server/issues/52967?

PR #14733, opened 2025-03-27
Fix Image Loading When Offline

alexknop:
Problem:

Although Videos were loading immediately when offline, Images were not

Solution:
Found that checking for whether file is offline was passing, but the ! in the code was making it false, forcing a sync which would timeout (taking extra time) and preventing immediate access to images

Fixes https://github.com/nextcloud/android/issues/14470

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
@@ -1206,9 +1206,9 @@ public class OCFileListFragment extends ExtendedListFragment implements
                     case GALLERY_SEARCH -> VirtualFolderType.GALLERY;
                     default -> VirtualFolderType.NONE;
                 };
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Thanks for the fix.
Please see https://github.com/nextcloud/android/issues/14470#issuecomment-2851119624

I cannot reproduce it on master, thus I cannot test the fix.

alexknop:
> Thanks for the fix. Please see [#14470 (comment)](https://github.com/nextcloud/android/issues/14470#issuecomment-2851119624)
> 
> I cannot reproduce it on master, thus I cannot test the fix.

Check https://github.com/nextcloud/android/issues/14470#issuecomment-2858654273 where displaying the video is working fine but then I try to select an image after and it won't load.
However looking at my proposed changes, don't we want to properly check if the image is downloaded before passing to the startImagePreview() function?

tobiasKaminsky (approved):

tobiasKaminsky (approved):

PR #14759, opened 2025-04-04
Open PDF Files Directly

alperozturk96:

Open PDF files directly without redirecting to the file details.


ZetaTom (commented):
In general this seems to work fine and fails safe in the event of a failed download. However, there is one case where a race condition causes the wrong file to be opened.

To observe this, prepare two PDF documents. One of them needs to have a large file size. If one tries to open the larger one first, the download will be queued. When one now downloads the second file via overflow menu > *Download*, the smaller file will be displayed instead.

tobiasKaminsky:
- for all downloadable files (except video and image as they are opened as a preview)
- hide detail for small files (like <1Mb)
- show detail fragment for larger files
- fix download progress bar in detail fragment

alperozturk96:
Since all these changes should be handled in different PRs, we can close this.

alperozturk96:
The new issue: https://github.com/nextcloud/android/issues/14850

PR #14775, opened 2025-04-08
Handle Downloaded File Open Intent

alperozturk96:

Try to open .xlsx file, it should redirect to an app that can view .xlsx file.


tobiasKaminsky:
Smoke test
- download xls/office files
- download other file types

tobiasKaminsky (approved):

PR #14819, opened 2025-04-24
Add checks for DecryptedFolderMetadataFile object to see if actually encrypted

alexknop:
Problem:

updateFileNameForEncryptedFile will run even if the object in question is null. As a result, we get an error.

Solution:

Add check for object to see if not null before performing an operation on a file which may not be encrypted.


tobiasKaminsky (approved):

alperozturk96 (changes requested):
Hello

Thank you for the PR.

I recommend using instanceof, as it already checks for both nullability and type. There's no need to check nullability or version separately.

```
if (object instanceof DecryptedFolderMetadataFileV1 metadataFile) {
     RefreshFolderOperation.updateFileNameForEncryptedFileV1(storageManager, metadataFile, mLocalFolder);
} else if (object instanceof DecryptedFolderMetadataFile metadataFile) {
     RefreshFolderOperation.updateFileNameForEncryptedFile(storageManager, metadataFile, mLocalFolder);
}
```

alperozturk96 (approved):

PR #14820, opened 2025-04-24
Remove misleading error

alexknop:
Problem:

On startup, we check every folder to see if the folder is part of the synced folders database. If it isn't, we are getting an error how we expected to see an entry for the folder in the sync database, which isn't true.
Screenshot From 2025-04-24 09-04-20

Solution:
Hide (cosmetic) error


tobiasKaminsky:
Hm. It checks if we have more than 1 entry.
And as every folder can/should only be once in DB, it is okay to have this as an error?

tobiasKaminsky (approved):

PR #14822, opened 2025-04-25
BugFix - Track File Upload Index

alperozturk96:

### Problem

Currently, the logic implemented in FileUploadWorker processes file uploads in discrete batches, which limits our ability to track the overall upload progress accurately and total upload index is wrong. We have outer while loop to update uploads and inner for loop for uploads.

For example, when a user attempts to upload 164 files:

1. The first batch is processed, triggering a progress notification for files 1–2.
2. The second batch then triggers a notification for files 23–2.
3. The third batch triggers a notification for files 53–2
4. ... and final batch 164-2.

### How to reproduce?

1. Create a folder
2. Try to upload n times file
3. Observe file upload notification

### Solution

- Pass totalUploadSize to the worker and utilize it internally.
- Update FileUploadWorker to support parallel job execution instead of queuing. This allows users to initiate multiple folder uploads independently, enabling better performance and more accurate tracking in notifications.

### Demo

https://github.com/user-attachments/assets/80b0ecb4-f4cd-4200-b604-af2bedd09bee

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
@@ -36,7 +36,6 @@ import com.owncloud.android.operations.DownloadType
 import java.util.Date
 import java.util.UUID
 import java.util.concurrent.TimeUnit
-import kotlin.random.Random
 import kotlin.reflect.KClass
 
 /**
@@ -584,7 +583,7 @@ internal class BackgroundJobManagerImpl(
      * @param uploadIds array of upload ids
      */
     override fun startFilesUploadJob(user: User, uploadIds: LongArray) {
-        val tag = startFileUploadJobTag(user) + Random.nextLong()
+        val tag = startFileUploadJobTag(user)
         val dataBuilder = Data.Builder()
             .putString(FileUploadWorker.ACCOUNT, user.accountName)
             .putLongArray(FileUploadWorker.UPLOAD_IDS, uploadIds)
@@ -599,7 +598,7 @@ internal class BackgroundJobManagerImpl(
             .setConstraints(constraints)
             .build()
 
-        workManager.enqueueUniqueWork(tag, ExistingWorkPolicy.KEEP, request)
+        workManager.enqueueUniqueWork(tag, ExistingWorkPolicy.APPEND_OR_REPLACE, request)
     }
 
     private fun startFileDownloadJobTag(user: User, fileId: Long): String =
--- a/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
@@ -36,6 +36,7 @@ import com.owncloud.android.operations.DownloadType
 import java.util.Date
 import java.util.UUID
 import java.util.concurrent.TimeUnit
+import kotlin.random.Random
 import kotlin.reflect.KClass
 
 /**
@@ -576,7 +577,7 @@ internal class BackgroundJobManagerImpl(
         startFilesUploadJobInternal(user, totalUploadSize)
 
     private fun startFilesUploadJobInternal(user: User, totalUploadSize: Int?) {
-        val tag = startFileUploadJobTag(user)
+        val tag = startFileUploadJobTag(user) + Random.nextLong()
         val dataBuilder = Data.Builder()
             .putString(FileUploadWorker.ACCOUNT, user.accountName)
 
--- a/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
@@ -581,16 +581,13 @@ internal class BackgroundJobManagerImpl(
      * number of files so it's safer to treat each invocation as an independent job.
      *
      * @param user The user for whom the upload job is being created.
-     * @param totalUploadSize Optional total size of the files to upload to track upload progress
+     * @param uploadIds array of upload ids
      */
-    override fun startFilesUploadJob(user: User, totalUploadSize: Int?) {
+    override fun startFilesUploadJob(user: User, uploadIds: LongArray) {
         val tag = startFileUploadJobTag(user) + Random.nextLong()
         val dataBuilder = Data.Builder()
             .putString(FileUploadWorker.ACCOUNT, user.accountName)
-
-        totalUploadSize?.let {
-            dataBuilder.putInt(FileUploadWorker.TOTAL_UPLOAD_SIZE, it)
-        }
+            .putLongArray(FileUploadWorker.UPLOAD_IDS, uploadIds)
 
         val constraints = Constraints.Builder()
             .setRequiredNetworkType(NetworkType.CONNECTED)
--- a/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
@@ -571,6 +571,18 @@ internal class BackgroundJobManagerImpl(
     override fun isStartFileUploadJobScheduled(user: User): Boolean =
         workManager.isWorkScheduled(startFileUploadJobTag(user))
 
+    /**
+     * This method supports initiating uploads for various scenarios, including:
+     * - New upload batches
+     * - Failed uploads
+     * - FilesSyncWork
+     *
+     * A unique tag is generated for each upload job. This is intentional because this job may encapsulate an arbitrary
+     * number of files so it's safer to treat each invocation as an independent job.
+     *
+     * @param user The user for whom the upload job is being created.
+     * @param totalUploadSize Optional total size of the files to upload to track upload progress
+     */
     override fun startFilesUploadJob(user: User, totalUploadSize: Int?) {
         val tag = startFileUploadJobTag(user) + Random.nextLong()
         val dataBuilder = Data.Builder()
--- a/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
@@ -570,10 +570,19 @@ internal class BackgroundJobManagerImpl(
     override fun isStartFileUploadJobScheduled(user: User): Boolean =
         workManager.isWorkScheduled(startFileUploadJobTag(user))
 
-    override fun startFilesUploadJob(user: User) {
-        val data = workDataOf(FileUploadWorker.ACCOUNT to user.accountName)
+    override fun startFilesUploadJob(user: User) = startFilesUploadJobInternal(user, null)
 
+    override fun startFilesUploadJob(user: User, totalUploadSize: Int) =
+        startFilesUploadJobInternal(user, totalUploadSize)
+
+    private fun startFilesUploadJobInternal(user: User, totalUploadSize: Int?) {
         val tag = startFileUploadJobTag(user)
+        val dataBuilder = Data.Builder()
+            .putString(FileUploadWorker.ACCOUNT, user.accountName)
+
+        totalUploadSize?.let {
+            dataBuilder.putInt(FileUploadWorker.TOTAL_UPLOAD_SIZE, it)
+        }
 
         val constraints = Constraints.Builder()
             .setRequiredNetworkType(NetworkType.CONNECTED)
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadHelper.kt
```diff
@@ -253,16 +253,21 @@ class FileUploadHelper {
...
-    fun cancelAndRestartUploadJob(user: User) {
+    fun cancelAndRestartUploadJob(user: User, totalUploadSize: Int? = null) {
         backgroundJobManager.run {
             cancelFilesUploadJob(user)
-            startFilesUploadJob(user)
+
+            if (totalUploadSize != null) {
+                startFilesUploadJob(user, totalUploadSize)
+            } else {
+                startFilesUploadJob(user)
```
alperozturk96, line 269: Some calls (e.g., FilesSyncWork) to startFilesUploadJob() don't have totalUploadSize, so we need two versions of it.

ZetaTom (dismissed):
During testing I've noticed that multiple upload batches aren't handled correctly. When I started the first batch upload of 400 files and queued a second one, also containing 400 files, the upload counter would eventually exceed 400 while the total counter would not be increased. All tests were run on a Pixel 2 emulator with the API 30 AOSP image.

image

We could use `@JvmOverloads` to automatically generate functions with the default parameters excluded. This means, that one can call them from Java code without these provided explicitly.

alperozturk96:
During my recent testing, I found an issue with the use of `getCurrentUploadsForAccountPageAscById()`. This function returns uploads that are currently in progress or scheduled for execution. When a user triggers two separate `FileUploadWorker` instances consecutively it causes a problem.

**Example scenario**

Consider a user with a folder structure where _**folder A**_ contains 101 files, including a subfolder named _**inner_folder**_ which also contains 101 files. If the user selects all files in _**folder A**_ except the _**inner_folder**_ and initiates an upload, the first `FileUploadWorker` will be triggered and `getCurrentUploadsForAccountPageAscById()` will return 101 uploads. Before the first upload completes, if the user then selects inner_folder and triggers another upload, the second `FileUploadWorker` will call `getCurrentUploadsForAccountPageAscById()` again. This time, it will return more than 101 uploads because the previous job has not yet finished. As a result, the returned upload count will be inaccurate, causing notifications to be displayed incorrectly.

Using an external counter to track the number of uploads does not resolve this issue as well, since the counter increments based on the inaccurate item count retrieved from the same function.

**Changes**

The `startFilesUploadJob()` method accepts a list of specific upload IDs. This approach ensures the worker always has the correct number of items to process. For scenarios where the exact upload count cannot be determined upfront, such as in `FilesSyncWork` or `cancelAndRestartUploadJob()`. We can continue to rely on `getCurrentUploadsForAccountPageAscById()` or if it's possible to write query for these cases we can add new function to the `UploadsStorageManager` then we can use it.

**The following tests need to be done**

- Auto upload
- Uploading a folder
- Uploading multiple folders consecutively 
- Uploading multiple files and folders consecutively
- Retrying failed uploads
- Retrying canceled uploads

Please compare these changes against the master branch to identify which aspects are within the current scope and which are not.

**Demo:**

https://github.com/user-attachments/assets/d1e11e01-5a72-4ab8-b40f-92d4c79eddc1

app/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -627,18 +608,13 @@ public OCUpload[] getCurrentAndPendingUploadsForAccount(final @NonNull String ac
         return getUploads(inProgressUploadsSelection, accountName);
     }
 
-    public OCUpload[] getCurrentUploadsForAccount(final @NonNull String accountName) {
-        return getUploads(ProviderTableMeta.UPLOADS_STATUS + EQUAL + UploadStatus.UPLOAD_IN_PROGRESS.value + AND +
-                              ProviderTableMeta.UPLOADS_ACCOUNT_NAME + IS_EQUAL, accountName);
+    public long[] getCurrentUploadIds(final @NonNull String accountName) {
```
alperozturk96, line 611: The OCUpload class is a wrapper around the UploadEntity data class and it is redundant. A similar redundancy exists between OCFile and FileEntity. These entity data classes can be used directly and can have helper methods.

The getCurrentUploadsForAccountPageAscById and getCurrentUploadsForAccount functions currently uses the ContentResolver to obtain a cursor, which is then mapped to OCUpload objects an unnecessary step. Instead, current uploads can be fetched directly from the Room database via DAO methods.

As a result, the getCurrentUploadsForAccountPageAscById and getCurrentUploadsForAccount functions become obsolete and should no longer be used.

### Demo


https://github.com/user-attachments/assets/96a25a91-496f-4181-8d0e-a9840b7a37fd

app/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -494,28 +497,6 @@ private String getInProgressAndDelayedUploadsSelection() {
             " ) AND " + ProviderTableMeta.UPLOADS_ACCOUNT_NAME + IS_EQUAL;
     }
 
-    public int getTotalUploadSize(@Nullable String... selectionArgs) {
```
alperozturk96, line 497: Unused function

app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
```diff
@@ -579,18 +580,31 @@ internal class BackgroundJobManagerImpl(
...
+     * - FilesSyncWork
+     *
+     * A unique tag is generated for each upload job. This is intentional because this job may encapsulate an arbitrary
+     * number of files so it's safer to treat each invocation as an independent job.
+     *
+     * @param user The user for whom the upload job is being created.
+     * @param uploadIds array of upload ids
+     */
+    override fun startFilesUploadJob(user: User, uploadIds: LongArray) {
+        val tag = startFileUploadJobTag(user) + Random.nextLong()
```
tobiasKaminsky: Please try without unique tag, so that only one worker at the same time runs.

app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
```diff
@@ -579,18 +580,31 @@ internal class BackgroundJobManagerImpl(
...
+        val dataBuilder = Data.Builder()
+            .putString(FileUploadWorker.ACCOUNT, user.accountName)
+            .putLongArray(FileUploadWorker.UPLOAD_IDS, uploadIds)
```
tobiasKaminsky, line 589: Do not passs uploadIs, but instead access data inside worker

app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
```diff
@@ -121,23 +122,22 @@ class FileUploadWorker(
...
-        var uploadsPerPage = uploadsStorageManager.getCurrentUploadsForAccountPageAscById(-1, accountName)
-        val totalUploadSize = uploadsStorageManager.getTotalUploadSize(accountName)
+        val uploadIds = inputData.getLongArray(UPLOAD_IDS) ?: return Result.success()
+        val uploads = uploadIds.map { id -> uploadsStorageManager.getUploadById(id) }.filterNotNull()
+        val totalUploadSize = uploadIds.size
 
-        Log_OC.d(TAG, "Total upload size: $totalUploadSize")
-
-        while (uploadsPerPage.isNotEmpty() && !isStopped) {
+        for ((index, upload) in uploads.withIndex()) {
```
tobiasKaminsky, line 134: Inside loop check if upload still is pending in uploadsStorageManager

tobiasKaminsky (dismissed):
see inline

app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
alperozturk96: https://github.com/nextcloud/android/pull/14822/commits/f363bcd1c63a46948818fbf00e669ccf21ab8507

app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
alperozturk96: In all cases, `uploadIds` are not directly accessible via the account name. For example:

* `FileUploadHelper.retryUploads` receives a `failedUploads: Array<OCUpload>` parameter, where the failed uploads originate from `UploadsStorageManager`.
* `FileUploadHelper.uploadNewFiles` maps the `localPaths` parameter passed from eight different classes to create `OCUpload` instances. Therefore, this process cannot be handled directly within the worker.
* `FileUploadHelper.cancelFileUploads` takes a `uploads: List<OCUpload>` argument, which is provided by various classes, making it unsuitable for direct handling inside the worker.
* `FileUploadHelper.uploadUpdatedFile` uses the `existingFiles` parameter to generate `OCUpload` objects through mapping, which similarly cannot be managed directly within the worker.

app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
alperozturk96: `UPLOAD_IN_PROGRESS` (to see check`public enum UploadStatus` description) represents **both scheduled and currently uploading files**, which makes it unsuitable for filtering within the loop without unintentionally skipping valid uploads. When I tried using the following SQL command, I wasn’t able to upload any files because it represents both scheduled and in-progress uploads.

```
@Query(
        "SELECT EXISTS(" +
            "SELECT 1 FROM " + ProviderTableMeta.UPLOADS_TABLE_NAME +
            " WHERE " + ProviderTableMeta._ID + " = :id AND " +
            ProviderTableMeta.UPLOADS_ACCOUNT_NAME + " = :accountName AND " +
            ProviderTableMeta.UPLOADS_STATUS + " = :pendingStatus" +
            ")"
    )
    fun isPendingUpload(
        id: Int,
        accountName: String,
        pendingStatus: Int = UploadStatus.UPLOAD_IN_PROGRESS.value
    ): Boolean
```

**Solutions:**

Introducing a new status (e.g., `UPLOAD_PROCESSING_NOW`) would require significant changes to the existing codebase, affecting multiple classes.

In-memory tracking (storing a set of upload IDs currently being processed to prevent re-processing them in the same run) is also unreliable because if the worker restarts unexpectedly (for example, due to system resource constraints or OS killing background tasks), this in-memory state is lost, leading to possible duplicate uploads.

Based on current testing, the worker executes uploads sequentially without any issues. If a user tries to upload the same files, no conflict resolve notification appears, just like in the master branch. If the file name is the same, but the content has changed, the conflict resolve notification appears, consistent with the master branch behavior. 

I suggest addressing this in a separate PR.

**Demo:**


https://github.com/user-attachments/assets/7cd63dbe-f7cb-47c9-b024-1e011fbb0352

tobiasKaminsky:
For me it does not trac any progress, but only stays at 0%.

alperozturk96:
> 
> For me it does not trac any progress, but only stays at 0%.

What is the image size? If image size is too small, it goes 0 to %100 too fast and you won't notice it. Also, this PR doesn't touch the progress percent it fixes the track file index e.g.

**1** / 10 file.jpg
**2** / 10 file2.jpg
..
..

Please compare with master.

**Master**


https://github.com/user-attachments/assets/0d37e4e7-491b-4ded-ab9b-da05fbc5bf70

tobiasKaminsky:
Ah, I was under the impression that percentage is about how many files, not file transfer of one file.
But you are right :+1:

app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
```diff
@@ -146,18 +145,41 @@ class FileUploadWorker(
...
 
-            Log_OC.d(TAG, "Handling ${uploadsPerPage.size} uploads for account $accountName")
-            val lastId = uploadsPerPage.last().uploadId
-            uploadFiles(totalUploadSize, uploadsPerPage, accountName)
-            uploadsPerPage =
-                uploadsStorageManager.getCurrentUploadsForAccountPageAscById(lastId, accountName)
-        }
+            val user = userAccountManager.getUser(accountName)
+            if (!user.isPresent) {
+                uploadsStorageManager.removeUpload(upload.uploadId)
```
stefan2904, line 150: is this mean to remove uploads that are not for the currently active user? (see https://github.com/nextcloud/android/issues/15320)

alperozturk96: Hey, thanks for the message.

Get the account that wants to upload the file. If that user is not available, remove the upload that belongs to the same user from the queue, just like before.

Image

What do you think? How does this change affect the issue you linked to, it is same as before?

Maybe we can find the root cause of the issue please share your thoughts, thanks.

Issue #14821, opened 2025-04-24
Thumbnail not generated after upload if set to delete photo from device


alexknop:
### ⚠️ Before posting ⚠️

- [x] This is a **bug**, not a question or an enhancement.
- [x] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [x] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Allow auto-upload on a folder containing a photo(s).
2. Allow Nextcloud to delete the picture from your device after uploading
3. Observe Thumbnail cannot be generated because file is not found after deletion.

### Expected behaviour

I would expect Thumbnail to be generated prior to deletion

### Actual behaviour

File is deleted and then a thumbnail cannot be generated

![Image](https://github.com/user-attachments/assets/f23d1c66-10bd-4454-bc75-7bd4066f88e8)

### Android version

15

### Device brand and model

TCL 60 XE

### Stock or custom OS?

Stock

### Nextcloud android app version

3.31.1

### Nextcloud server version

31.0.4

### Using a reverse proxy?

I don't know

### Server error logs

```bash

```

PR #14823, opened 2025-04-25
Handle Sample Bitmap Decoding Better

alperozturk96:

### Issue

The decodeSampledBitmapFromFile() function may return null, and while null checks are implemented throughout the codebase, the associated log messages are often misleading. The current implementation lacks a check for file existence prior to decoding, making it difficult to accurately identify and debug the root cause of failures.

### Changes

1. Added a check to ensure the file exists before attempting to decode.
2. Replaced direct exception logging during decoding with conditional throwing of IOException when necessary.
3. Prevented ImageDecoder from attempting to decode non-existent files to avoid decoding failures.
4. Ensured BitmapFactory.decodeFile is only called when the target file exists.

### Related Issue

This PR helps us better understand the following issue

https://github.com/nextcloud/android/issues/14821

### DIFF
--- a/app/src/main/java/com/owncloud/android/utils/BitmapUtils.java
+++ b/app/src/main/java/com/owncloud/android/utils/BitmapUtils.java
@@ -35,8 +35,8 @@ import com.owncloud.android.lib.resources.users.Status;
 import com.owncloud.android.lib.resources.users.StatusType;
 import com.owncloud.android.ui.StatusDrawable;
 
-
 import java.io.File;
+import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/utils/BitmapUtils.java
```diff
@@ -85,18 +107,21 @@ public static Bitmap addColorFilter(Bitmap originalBitmap, int filterColor, int
      * @return decoded bitmap
      */
     public static Bitmap decodeSampledBitmapFromFile(String srcPath, int reqWidth, int reqHeight) {
```
tobiasKaminsky, line 109: Please "@Nullable"

app/src/main/java/com/owncloud/android/utils/BitmapUtils.java
```diff
@@ -75,6 +76,27 @@ public static Bitmap addColorFilter(Bitmap originalBitmap, int filterColor, int
         return resultBitmap;
     }
 
+    @RequiresApi(Build.VERSION_CODES.P)
+    private static Bitmap decodeSampledBitmapViaImageDecoder(@NonNull File file, int reqWidth, int reqHeight) {
```
tobiasKaminsky, line 80: Same here

tobiasKaminsky (changes requested):
nullable annotation

tobiasKaminsky (approved):

PR #14827, opened 2025-04-25
Fix error handling issues

alexknop:
Problem:

- Empty catch block
- Catch parameter that said ignored was actually used
- Throw inside of a finally block

Solution
- Change empty block to ignored
- Change parameter to e
- Logging error instead of a throw, as doing this in a finally is not recommended

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/EditorWebView.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/EditorWebView.java
@@ -75,7 +75,7 @@ public abstract class EditorWebView extends ExternalSiteWebView {
             new WebViewUtil().setProxyKKPlus(this.getWebView());
             try {
                 Thread.sleep(1000);
-            } catch (InterruptedException e) {
+            } catch (InterruptedException ignored) {
             }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/operations/CreateFolderOperation.java
```diff
@@ -375,8 +374,7 @@ private RemoteOperationResult encryptedCreateV2(OCFile parent, OwnCloudClient cl
                 RemoteOperationResult unlockFolderResult = EncryptionUtils.unlockFolder(parent, client, token);
 
                 if (!unlockFolderResult.isSuccess()) {
-                    // TODO E2E: do better
```
tobiasKaminsky, line 378: Those error are critical and should not be siltently catched and ignored.
Please revert this and the above.

Real solution would be to show an error to the user.
But this is not needed to do by you.

tobiasKaminsky (changes requested):
revert 2x removing of "throw"

alexknop:
> revert 2x removing of "throw"

done!

alexknop:
It looks like one of these was fixed in another PR, so just a small change is left regarding error handling

PR #14834, opened 2025-04-28
Warn Users About Invalid Filenames During Download

alperozturk96:

### How to Reproduce

1. Upload a file from the web application that contains the ">" character in its name.
2. Attempt to download the file from the client.

### Note
A toast message is used instead of a snackbar because the view is not accessible in the DownloadFileOperation.


app/src/main/java/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -69,6 +69,27 @@ private FileStorageUtils() {
...
+    private static boolean isValidExtFilenameChar(char c) {
+        if ((int) c <= 0x1F) {
+            return false;
+        }
+
+        return switch (c) {
+            case '"', '*', ':', '/', '<', '>', '?', '\\', '|', 0x7F -> false;
+            default -> true;
+        };
+    }
```
ZetaTom, line 91: Please provide references for these values in comments.

app/src/main/java/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -69,6 +69,27 @@ private FileStorageUtils() {
         // utility class -> private constructor
     }
 
+    public static boolean isValidExtFilename(String name) {
```
tobiasKaminsky, line 72: Please write a test.

tobiasKaminsky (changes requested):
see inline.

tobiasKaminsky (approved):

PR #14840, opened 2025-04-30
Fix Lint Result

alperozturk96:

**Changes**

Master was 58 before [this PR #14716](https://github.com/nextcloud/android/pull/14716) so it should remain at least 58 not 59.
Warnings should be 51, not 54 anymore, somehow lint-results.txt was not updated automatically. @tobiasKaminsky


app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -574,7 +572,7 @@ private void synchronizeData(List<Object> folderAndFiles) {
                 updateFileNameForEncryptedFileV1(fileDataStorageManager,
                                                  (DecryptedFolderMetadataFileV1) object,
                                                  updatedFile);
-            } else {
+            } else if (object != null) {
```
alperozturk96, line 575: Reduce Correctness spotbugs

app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -436,7 +436,7 @@ protected RemoteOperationResult run(OwnCloudClient client) {
         // parent file is not null anymore:
         // - it was created on fresh upload or
         // - resume of encrypted upload, then parent file exists already as unlock is only for direct parent
-
+        assert(parent != null);
```
alperozturk96, line 439: Since parent is no longer null, we can use an assert here; thus, the following line will not warn us about a null pointer exception (NPE).

tobiasKaminsky: Please move 432 to 435, then we do not need assert

tobiasKaminsky (dismissed):
see inline

tobiasKaminsky (approved):

PR #14857, opened 2025-05-02
Improve File Download Tracking

alperozturk96:

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/download/FileDownloadWorker.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/download/FileDownloadWorker.kt
@@ -153,8 +153,8 @@ class FileDownloadWorker(
         )
     }
 
-    private fun setWorkerState(user: User?) {
-        WorkerStateLiveData.instance().setWorkState(WorkerState.DownloadStarted(user, currentDownload))
+    private fun setWorkerState(user: User?, percent: Int) {
+        WorkerStateLiveData.instance().setWorkState(WorkerState.DownloadStarted(user, currentDownload, percent))
     }
 
     private fun setIdleWorkerState() {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1192,52 +1189,38 @@ private Integer checkFileBeforeOpen(OCFile file) {
...
-                setFabVisible(false);
-                ((FileDisplayActivity) mContainerActivity).startMediaPreview(file, 0, true, true, false, true);
-            } else {
-                mContainerActivity.getFileOperationsHelper().openFile(file);
-            }
+        if (PreviewImageFragment.canBePreviewed(file) && mContainerActivity instanceof FileDisplayActivity fda) {
+            fda.previewAndHandleImageFile(file, searchFragment, currentSearchType);
+        } else if (file.isDown() && mContainerActivity instanceof FileDisplayActivity fda) {
+            fda.previewDownloadedFile(file, this::setFabVisible);
         } else {
```
alperozturk96, line 1200: This entire preview logic belongs in FileDisplayActivity and needs to be accessible from there as well,
because after the file download is completed, the fragment is no longer an instance of OCFileListFragment,
so this function cannot be used from FileDisplayActivity.

app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1743,16 +1754,81 @@ private void observeWorkerState() {
...
+            return;
+        }
+
+        if (fileIDForImmediatePreview != currentFile.getFileId() || !currentFile.isDown()) {
+            return;
+        }
+
+        fileIDForImmediatePreview = -1;
+        if (PreviewImageFragment.canBePreviewed(currentFile)) {
+            startImagePreview(currentFile, !currentFile.isDown());
```
tobiasKaminsky, line 1784: Please check if the second paramater is correct.
I think it must be negated.

tobiasKaminsky (approved):

PR #14897, opened 2025-05-17
ThumbnailsCacheManager - AsyncTask to Executor

alexknop:
Problem:
-Android.os.AsyncTask is deprecated
-imageViewReference could not be null in some parts
-if/else could be optimized
-getAvatarDimension is unused

Solution
-Leverage Executor for InitDiskCacheTask
-Removed non-null check 
-Use requireNonNullElseGet
-Remove unused function

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
@@ -67,6 +67,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.lang.ref.WeakReference;
 import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
@@ -111,9 +114,10 @@ public final class ThumbnailsCacheManager {
     private ThumbnailsCacheManager() {
     }
 
-    public static class InitDiskCacheTask extends AsyncTask<File, Void, Void> {
-        @Override
-        protected Void doInBackground(File... params) {
+    private static final ExecutorService executor = Executors.newSingleThreadExecutor();
+
+    public static void initDiskCacheAsync() {
+        executor.execute(() -> {
             synchronized (mThumbnailsDiskCacheLock) {
                 mThumbnailCacheStarting = true;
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -480,40 +484,38 @@ protected Bitmap doInBackground(Object... params) {
         }
 
         protected void onPostExecute(Bitmap bitmap) {
-            if (imageViewReference != null) {
```
tobiasKaminsky, line 483: Indeed, I think we did the check wrong:
reference can never be null, but content can, right?
So we need something like "imageViewReference.get() != null"?

app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -814,11 +816,7 @@ private Bitmap doFileInBackground() {
             File file = (File)mFile;
 
             final String imageKey;
-            if (mImageKey != null) {
-                imageKey = mImageKey;
-            } else {
-                imageKey = String.valueOf(file.hashCode());
-            }
+            imageKey = Objects.requireNonNullElseGet(mImageKey, () -> String.valueOf(file.hashCode()));
```
tobiasKaminsky, line 819: Not sure if this is more readable, but if you want to keep it, then you can also combine it with L818 :+1:

app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -931,12 +929,7 @@ protected void onPostExecute(Bitmap bitmap) {
 
         private Bitmap doFileInBackground(File file, Type type) {
             final String imageKey;
-
-            if (mImageKey != null) {
-                imageKey = mImageKey;
-            } else {
-                imageKey = String.valueOf(file.hashCode());
-            }
+            imageKey = Objects.requireNonNullElseGet(mImageKey, () -> String.valueOf(file.hashCode()));
```
tobiasKaminsky, line 932: Same here

app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
alexknop: changed to .get() instead!

tobiasKaminsky (approved):

PR #14898, opened 2025-05-17
Fix Rotation of Thumbnails

alexknop:
Problem:
-Thumbnails were being rotated based on EXIF orientation, but these are already handled by BitmapUtils.retrieveBitmapFromFile() when we are extracting the image.

Solution:
-Take out rotateImage.

Screenshot: Top 2 were rotated, when taking out the rotateImage step, now 3rd thumbnail looks normal.
Screenshot_20250517-144740


tobiasKaminsky:
To check
- have two sets of images: one with exif, one without exif orientation
- upload to cloud (not via Android app)
- see if all of them are correctly orientated
- do the same while uploading within Android 
- check preview image 
- check after manual downloading

alperozturk96:
Is this PR still valid? I can only see deletion.

alperozturk96:

Please feel free to reopen it once the changes are addressed or PR is still valid. Thank you again for your contribution.

PR #14900, opened 2025-05-18
Decode and Scale Together

alexknop:
Problem:

Images are decoded in full resolution and then later scaled for the screen size when this can be done together.

Solution:

Replace Decoder Listener and pass in desired height/width based on screen size. 
Remove call to scale after decoding in full resolution. 

This saves seconds off of loading images on my device.

### DIFF
--- a/app/src/main/java/com/owncloud/android/utils/BitmapUtils.java
+++ b/app/src/main/java/com/owncloud/android/utils/BitmapUtils.java
@@ -84,14 +84,11 @@ public final class BitmapUtils {
 
             final var imageDecoderSource = ImageDecoder.createSource(file);
 
-            final var onDecoderListener = new ImageDecoder.OnHeaderDecodedListener() {
-                @Override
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/utils/BitmapUtils.java
```diff
@@ -56,10 +56,6 @@
 public final class BitmapUtils {
     public static final String TAG = BitmapUtils.class.getSimpleName();
 
-    private BitmapUtils() {
```
alperozturk96, line 59: All methods in this class are static, so removing the private constructor allows the class to be instantiated, which is not meaningful in this context. Since instantiation is unnecessary, please reintroduce the private constructor to prevent object creation.

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/utils/BitmapUtils.java
alexknop: Thank you for pointing this out, it has been changed back.

alperozturk96 (approved):
Looks good to me.

PR #14935, opened 2025-05-30
Migrate to Version Catalog III

alperozturk96:


ZetaTom (changes requested):
I think we'll have to change the android library version check in [`scripts/analysis/detectWrongSettings.sh`](https://github.com/nextcloud/android/blob/master/scripts/analysis/detectWrongSettings.sh#L9) for the CI to work. Otherwise everything works as expected.

PR #14939, opened 2025-05-31
Identify and remove unused assignments

alexknop:
Problem:

-Variables had unused assignments, either set and the variable was never used or they were set and then immediately overwritten.

Solution:

-Remove these variable assignments and only assign when needed.

### DIFF
--- a/app/src/main/java/com/owncloud/android/files/services/IndexedForest.java
+++ b/app/src/main/java/com/owncloud/android/files/services/IndexedForest.java
@@ -184,10 +184,7 @@ public class IndexedForest<V> {
     }
 
     private void removeDescendants(Node<V> removed) {
-        Iterator<Node<V>> childrenIt = removed.getChildren().iterator();
-        Node<V> child = null;
-        while (childrenIt.hasNext()) {
-            child = childrenIt.next();
+        for (Node<V> child : removed.getChildren()) {
             mMap.remove(child.getKey());
             removeDescendants(child);
         }
--- a/app/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
@@ -280,18 +280,7 @@ public class SynchronizeFolderOperation extends SyncOperation {
         }
 
         // get current data about local contents of the folder to synchronize
-        Map<String, OCFile> localFilesMap;
-        E2EVersion e2EVersion;
-
-        if (object instanceof DecryptedFolderMetadataFileV1) {
-            e2EVersion = E2EVersion.V1_2;
-            localFilesMap = RefreshFolderOperation.prefillLocalFilesMap((DecryptedFolderMetadataFileV1) object,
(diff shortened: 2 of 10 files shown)


app/src/main/java/com/owncloud/android/operations/SynchronizeFolderOperation.java
```diff
@@ -284,11 +284,9 @@ private void synchronizeData(List<Object> folderAndFiles) throws OperationCancel
         E2EVersion e2EVersion;
 
         if (object instanceof DecryptedFolderMetadataFileV1) {
```
alperozturk96: Please remove E2EVersion as it is no longer in use.

Additionally, the if-else block for prefillLocalFilesMap appears to be unnecessary. It can be replaced with the following simplified line:

`localFilesMap = RefreshFolderOperation.prefillLocalFilesMap(object,storageManager.getFolderContent(mLocalFolder, false));`

app/src/main/java/com/owncloud/android/ui/asynctasks/GallerySearchTask.java
```diff
@@ -170,7 +170,7 @@ private boolean parseMedia(long startDate, long endDate, List<Object> remoteFile
         }
 
         // existing files to remove
-        filesDeleted = localFilesMap.values().size();
+        long filesDeleted = localFilesMap.values().size();
```
alperozturk96: Could you please replace this line with the following code? Since localFilesMap is a map, there's no need to iterate over its values:

`long filesDeleted = localFilesMap.size();`

app/src/main/java/com/owncloud/android/files/services/IndexedForest.java
```diff
@@ -185,9 +185,8 @@ public Pair<V, String> removePayload(String accountName, String remotePath) {
 
     private void removeDescendants(Node<V> removed) {
```
alperozturk96, line 186: This `while` block can be replaced with the following code:

```
private void removeDescendants(Node<V> removed) {
        for (Node<V> child : removed.getChildren()) {
            mMap.remove(child.getKey());
            removeDescendants(child);
        }
    }
```

alperozturk96 (changes requested):

alperozturk96:
Hello

Thank you for the PR. I’ve added a few minor suggestions.

app/src/main/java/com/owncloud/android/ui/asynctasks/GallerySearchTask.java
alexknop: fixed

app/src/main/java/com/owncloud/android/files/services/IndexedForest.java
alexknop: fixed

alexknop:
@alperozturk96 thank you for the quick turnaround, everything is fixed per your suggestions

alperozturk96 (approved):

PR #14946, opened 2025-06-02
Activity/Trashbin: Show correct info if server not reachable

tobiasKaminsky:

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1348,4 +1348,6 @@
     <string name="auto_upload_sub_folder_warning">This folder is already included in the parent folder’s sync, which may cause duplicate uploads</string>
     <string name="sync_anyway">Sync anyway</string>
     <string name="sync_duplication">Sync duplication</string>
+    <string name="server_not_reachable">Server not reachable</string>
+    <string name="server_not_reachable_content">Device has no internet connection or server is not reachable!</string>
 </resources>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1348,4 +1348,6 @@
     <string name="auto_upload_sub_folder_warning">This folder is already included in the parent folder’s sync, which may cause duplicate uploads</string>
     <string name="sync_anyway">Sync anyway</string>
     <string name="sync_duplication">Sync duplication</string>
+    <string name="server_not_reachable">Server not reachable</string>
+    <string name="server_not_reachable_content">Device has no internet connection or server is not reachable!</string>
```
kra-mo:
```suggestion
    <string name="server_not_reachable">Could not load content</string>
    <string name="server_not_reachable_content">The device is likely not connected to the internet</string>
```

"Server not reachable" is very broad and from the user's standpoint, not very actionable. Instead, assuming that there is no internet is probably fine. If there is an issue, users will report it to their admins anyway. But we should probably discourage them from doing so, rather than encourage. Users will always blame sysadmins when prompted :)

kra-mo (changes requested):

app/src/main/res/values/strings.xml
tobiasKaminsky: Thank you!

PR #14960, opened 2025-06-06
BugFix - Use File ID Instead of File Object for EditorWebView Argument Passing

alperozturk96:

This PR fixes argument-passing logic to use the fileId (a Long) instead of the full `OCFile` object when navigating to `EditorWebView` and its subclasses.

Currently, `RichDocumentsEditorWebView` and `TextEditorWebView` extends from `EditorWebView`, which itself extends from `ExternalSiteWebView`, and it is also extends from `FileActivity`. Due to this inheritance chain, the `onCreate` methods of parent classes are executed before those of the child. As a result, any large `OCFile` object passed through an Intent will crash before `FileActivity` has the chance to handle it properly.

This issue arises because Android imposes a 1MB limit on Parcelable extras.

This PR resolves that by passing only the fileId, which is later used to retrieve the `OCFile` within `FileActivity`.

### How to test?

1. Create a new .md file in the app.
2. Attempt to open and edit the file using the editor.


alperozturk96:
Additional detailed findings are required. For now, I have reverted this PR to draft.

PR #14966, opened 2025-06-07
Update library to 2025-06-08

tobiasKaminsky:
Signed-off-by: Tobias Kaminsky <tobias@kaminsky.me>


alperozturk96:
Master already updated and compiling without any issue.

Screenshot 2025-06-12 at 15 49 56

`attributes` String should be used, not empty String.

Screenshot 2025-06-12 at 15 51 07

PR #14994, opened 2025-06-12
Drop Min Supported Versions

alperozturk96:

Min supported Android API Level is 26
Uses [NC Android lib 2.21.0 RC1](https://github.com/nextcloud/android-library/releases/tag/rc-2.21.0)
OUTDATED_SERVER_VERSION is 29

### DIFF
--- a/app/src/androidTest/java/com/owncloud/android/utils/theme/CapabilityUtilsIT.kt
+++ b/app/src/androidTest/java/com/owncloud/android/utils/theme/CapabilityUtilsIT.kt
@@ -19,8 +19,8 @@ class CapabilityUtilsIT : AbstractIT() {
     fun checkOutdatedWarning() {
         assertFalse(test(NextcloudVersion.nextcloud_31))
         assertFalse(test(NextcloudVersion.nextcloud_30))
-        assertFalse(test(NextcloudVersion.nextcloud_29))
 
+        assertTrue(test(NextcloudVersion.nextcloud_29))
         assertTrue(test(NextcloudVersion.nextcloud_28))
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/ui/asynctasks/LoadContactsTask.java
ZetaTom, line 1: Please remove unused imports.

alperozturk96:
@tobiasKaminsky 

Screenshot 2025-06-12 at 11 54 04


Lint increased. If I delete this folder, icon appears like below:

Screenshot 2025-06-12 at 11 54 38


Shall I delete or keep?

alperozturk96:
If I delete version check from this class AudioFocusManager, tests are failing.

tobiasKaminsky:
> Lint increased. If I delete this folder, icon appears like below:

Please test again.
I renamed the folder and tested with API35

tobiasKaminsky:
> If I delete version check from this class AudioFocusManager, tests are failing.

Hm. Maybe it never worked, since we test with an older Android version?

alperozturk96:
> > If I delete version check from this class AudioFocusManager, tests are failing.
> 
> Hm. Maybe it never worked, since we test with an older Android version?

I attempted to fix the issue but encountered a different problem (NPE for focusListener), and I'm not entirely sure what the current tests are validating. It might require more time to properly address. Would it be better to remove the tests for now and revisit them later?

alperozturk96:
> > Lint increased. If I delete this folder, icon appears like below:
> 
> Please test again. I renamed the folder and tested with API35

Working.

Screenshot_20250612_141013

PR #15007, opened 2025-06-16
BugFix - Handle Offline Remove Operation

alperozturk96:


app/src/main/java/com/nextcloud/client/jobs/upload/FileUploaderIntents.kt
```diff
@@ -57,32 +56,6 @@ class FileUploaderIntents(private val context: Context) {
         )
     }
 
-    fun resultIntent(resultCode: ResultCode, operation: UploadFileOperation): PendingIntent {
```
alperozturk96, line 60: Unused function

PR #15037, opened 2025-06-20
BugFix - Do Not Apply HideBottomViewOnScrollBehavior For Bottom Navigation View

alperozturk96:

### How to reproduce?

1. Ensure there are enough files in the list to cover the full height of the device screen.
2. Scroll down slightly.
3. Observe that the Bottom Navigation View and the Plus (+) button disappear.
4. Tap on any folder to navigate into it.
5. Notice that within the folder, the Bottom Navigation View remains hidden, and only the Plus (+) button is visible.

### Demo

https://github.com/user-attachments/assets/755bc4ed-2b27-4aae-a9c9-b68c745cfcc1

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/GalleryAdapter.kt
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/GalleryAdapter.kt
@@ -95,7 +95,7 @@ class GalleryAdapter(
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): SectionedViewHolder {
         val inflater = LayoutInflater.from(parent.context)
 
-        return when(viewType) {
+        return when (viewType) {
             VIEW_TYPE_HEADER -> {
                 val binding = GalleryHeaderBinding.inflate(inflater, parent, false)
                 GalleryHeaderViewHolder(binding)
--- a/app/src/main/java/com/owncloud/android/ui/adapter/GalleryAdapter.kt
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/GalleryAdapter.kt
@@ -90,30 +90,32 @@ class GalleryAdapter(
         }
     }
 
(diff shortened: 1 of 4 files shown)


app/src/main/res/layout/list_footer.xml
```diff
@@ -12,6 +12,7 @@
     android:layout_gravity="center_horizontal"
     android:gravity="center_horizontal"
     android:orientation="vertical"
+    android:layout_marginBottom="@dimen/standard_double_margin"
```
alperozturk96, line 15: Footer and last item in the list must be visible, thus bottom margin needed.

app/src/main/java/com/owncloud/android/ui/adapter/GalleryAdapter.kt
```diff
@@ -88,26 +90,32 @@ class GalleryAdapter(
...
-                storageManager,
-                this,
-                viewThemeUtils
-            )
+            }
+            VIEW_TYPE_FOOTER -> {
+                val binding = ListFooterBinding.inflate(inflater, parent, false)
+                OCFileListFooterViewHolder(binding)
+            }
+            else -> throw IllegalArgumentException("Unsupported view type: $viewType")
```
alperozturk96, line 118: The value must be one of the following: VIEW_TYPE_HEADER, VIEW_TYPE_ITEM, or VIEW_TYPE_FOOTER. You can confirm this by reviewing the SectionedViewHolder implementation. Unfortunately, since the type is represented as an integer rather than an enum, a default else clause with an exception is required.

tobiasKaminsky (changes requested):
Please try to mimick the same behaviour as "+" button, instead of always showing bottom bar.

alperozturk96:
https://github.com/nextcloud/android/pull/15063

Issue #14149, opened 2024-12-10
Navigation and Toolbar State Management into a Unified Enum-Based Approach


alperozturk96:
### Context

Our current implementation involves managing navigation state, toolbar configuration, and menu item highlighting using multiple independent sources. This fragmented approach introduces significant complexity and fragility, especially when states override one another or are not persisted properly.

**Toolbar Variants:**

With search:

<img width="463" height="134" alt="Image" src="https://github.com/user-attachments/assets/a697bf46-ea05-466f-aa7d-ee9e708526a6" />

Plain:

<img width="463" height="134" alt="Image" src="https://github.com/user-attachments/assets/5724d27d-6b7c-4b9d-a8ef-7703286aa982" />

### Current Implementation Issues

We currently rely on three separate mechanisms to determine and control the toolbar type, screen title, and the selected menu item (in both bottom navigation and drawer):

1. Static Booleans in `MainApp.java`:

```
private static boolean mOnlyOnDevice;
private static boolean mOnlyPersonalFiles;
```

2. SearchEvent in `com.owncloud.android.ui.events`:

Various search types are used to define the current context in fragments such as `OCFileListFragment`, `GalleryFragment`, and `SharedListFragment`, all managed by `FileDisplayActivity`:

```
 public static enum SearchType {
        FILE_SEARCH,
        FAVORITE_SEARCH,
        RECENTLY_MODIFIED_SEARCH,
        PHOTO_SEARCH,
        SHARED_SEARCH,
        GALLERY_SEARCH,
        FILE_ID_SEARCH,
        CONTENT_TYPE_SEARCH,
        RECENTLY_ADDED_SEARCH,
        SHARED_FILTER;
}
```

3. Static `menuItemId` from `DrawerActivity`:

Tracks user interaction from UI components within `DrawerActivity`. These mechanisms are not synchronized and often override one another, making state management brittle and error-prone.

### Problem Scenarios

Scenario 1:

A user navigates to the Media tab. The `SearchEvent` and `menuItemId` are correctly set. However, when the user backgrounds the app and later returns, the "All Files" are shown while the "Media" menu item remains highlighted. This occurs because `FileDisplayActivity` must handle all search types and ensure the toolbar and active menu item reflect the current state, but it often fails due to fragmented state checks.

Scenario 2:

[...]

PR #15049, opened 2025-06-24
BugFix - Fix Toolbar Inconsistency For Favorites Tab

alperozturk96:

The Favorites toolbar is inconsistent when switching between screens. 

**How to reproduce:**

1. Go to the Favorites tab.
2. Lock the device screen while on the Favorites tab (or switch to another tab and then return to Favorites)
3. Unlock the device and reopen the app.
4. Notice the search bar looks different.


alperozturk96:
> Please check if with new behaviour when you click on a folder, that then folder is shown.

Seems like okay.

https://github.com/user-attachments/assets/f7e81a5d-8c82-4b53-8541-a505e95a4a5b

ZetaTom (dismissed):
During our review meeting, @alperozturk96 and I found the following inconsistencies:

- unified search bar still appears
	1. Open Nextcloud App
	2. (Add folder to favourites, only if no favourites are set)
	3. Open *Favourites* tab
	4. Open folder from second step
	5. Press back button
	6. Wrong search bar appears
- wrong title in search bar
	1. Repeat steps above
	2. Lock device
	3. Unlock device
	4. *Favourites* is replaced with instance name ("Devcloud" in my case)

tobiasKaminsky:
@alperozturk96 can you update (and fix above finding by you and Tom)?

alperozturk96:
I have fixed the mentioned issues. However, I am not fully certain whether similar issues might occur in other scenarios. The current state is already an improvement over master. I have also provided a detailed explanation of the original issue earlier [here](https://github.com/nextcloud/android/issues/14149).

One of the reasons we don’t have a stable solution is that the toolbar is being set from multiple places, which can override each other. Currently, there is no single source of truth for managing it.

Screenshot 2025-09-03 at 13 47 25

@tobiasKaminsky

PR #15063, opened 2025-06-27
BugFix - Slide HideBottomViewOnScrollBehavior For BottomNavigationView

alperozturk96:

### How to reproduce?

1. Ensure there are enough files in the list to cover the full height of the device screen.
2. Scroll down slightly.
3. Observe that the Bottom Navigation View and the Plus (+) button disappear.
4. Tap on any folder to navigate into it.
5. Notice that within the folder, the Bottom Navigation View remains hidden, and only the Plus (+) button is visible.

### Demo

https://github.com/user-attachments/assets/587ebcaf-63f8-40ab-a402-f247ba7e5fc7

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
@@ -36,7 +36,6 @@ import android.view.ViewGroup;
 import android.widget.AbsListView;
 import android.widget.Toast;
 
-import com.google.android.material.behavior.HideBottomViewOnScrollBehavior;
 import com.google.android.material.bottomsheet.BottomSheetBehavior;
 import com.google.android.material.dialog.MaterialAlertDialogBuilder;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -2241,40 +2239,26 @@ public void setFabVisible(final boolean visible) {
...
+                viewThemeUtils.material.themeFAB(mFabMain);
+            } else {
+                mFabMain.hide();
             }
+
+            ViewExtensionsKt.slideHideBottomBehavior(mFabMain, visible);
+        });
+    }
+
+    public void slideHideBottomBehaviourForBottomNavigationView(boolean visible) {
```
alperozturk96, line 2259: The visibility of the bottom navigation view cannot be managed within the setFABVisible method, as it does not consistently align with the visibility of the FAB in all scenarios. Such as during multi-select operations. 

Therefore, the bottom navigation view visibility logic needs to be handled in a separate function.

PR #15067, opened 2025-06-27
🎨 Migrate to outlined icon style

AndyScherzinger:
- [x] Tests written, or not not needed

PR #15075, opened 2025-06-30
Align recommended files layout with grid item layout

AndyScherzinger:

Triggered by https://github.com/nextcloud/android/pull/15067#pullrequestreview-2971623961
Change the layout to be closer to the grid item layout. besides the extra, subline I think it would also be fine to simply use the same layout+subline.

|before|after|
|---|---|
|Screenshot_20250630_155232|Screenshot_20250630_183055|


jancborchardt (approved):
Much better already. And we need to improve more :O

kra-mo (approved):

Issue #14855, opened 2025-05-02
SyncBroadcastReceiver in FileDisplayActivity Triggers Multiple Times Causing Repeated Directory Listing


alperozturk96:
### Steps to reproduce

1. Upload files in a directory.
2. Observe FileDisplayActivity.SyncBroadcastReceiver being triggered multiple times during the process.
3. As a result, listDirectory() is called repeatedly.
4. These repeated calls lead to multiple invocations of OCFileListAdapter.swapDirectory().

### Suggestion

The broadcast receiver should be triggered only when necessary, and directory listing should occur a single time per meaningful change.

alperozturk96:
Currently, the following code retrieves the path of the remote folder that needs to be synchronized:

`String synchFolderRemotePath = intent.getStringExtra(FileSyncAdapter.EXTRA_FOLDER_PATH);`

Subsequently, we fetch the list of remote files inside that folder with:

`RemoteOperationResult synchResult = (RemoteOperationResult) DataHolderUtil.getInstance().retrieve(intent.getStringExtra(FileSyncAdapter.EXTRA_RESULT));`

Given that we already have access to the current list of files in the local directory, we can calculate and compare the hash codes of both local and remote file lists. If the hash codes match, we can safely skip the synchronization process. If they differ, we proceed with the update.

This approach would result in a significant performance improvement. **Even a single unnecessary folder update currently triggers a full directory listing, including database queries, thumbnail cache checks, and redundant rendering of thumbnails for each file.**

### Implementation Steps

**Convert FileDisplayActivity.SyncBroadcastReceiver to Kotlin (First PR)**

Refactor the logic for better readability, maintainability and performance.

~~**Introduce Hash Code Comparison (Second PR)**Implement the hash code comparison logic within FileDisplayActivity.SyncBroadcastReceiver.~~

**Analyze SyncBroadcastReceiver Usage (Third PR)**

Review the usage of FileDisplayActivity.SyncBroadcastReceiver throughout the codebase. Identify and remove any redundant or unnecessary triggers if it exists.

**Better OCFileListAdapter implementation - this is big one and should be discussed in different issue**

-------------------------

These steps will bring the Android implementation more in line with iOS in terms of performance and state management, and it's a noticeable user experience improvement.


**For each change, we must validate the following functional areas:**

- Folder and file listing behavior
- "Favorites", "Shared with you", "Personal", and "On-device files" filters.
- General sync behavior such as folder sync, file download.

I believe we should prioritize this task.

[...]

PR #15081, opened 2025-07-02
BugFix - Do not handleRemovedFolder for Root Directory

alperozturk96:

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
@@ -1355,10 +1355,11 @@ public class FileDisplayActivity extends FileActivity
 
         OCFile currentFile = (getFile() == null) ? null : getStorageManager().getFileByPath(getFile().getRemotePath());
         final OCFile currentDir = (getCurrentDir() == null) ? null : getStorageManager().getFileByPath(getCurrentDir().getRemotePath());
+        final boolean isSyncFolderRemotePathRoot = OCFile.ROOT_PATH.equals(syncFolderRemotePath);
 
-        if (currentDir == null) {
+        if (currentDir == null && !isSyncFolderRemotePathRoot) {
             handleRemovedFolder(syncFolderRemotePath);
-        } else {
+        } else if (currentDir != null) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.java
```diff
@@ -1323,10 +1323,11 @@ private void handleSyncEvent(String event, String syncFolderRemotePath, String i
 
         OCFile currentFile = (getFile() == null) ? null : getStorageManager().getFileByPath(getFile().getRemotePath());
         final OCFile currentDir = (getCurrentDir() == null) ? null : getStorageManager().getFileByPath(getCurrentDir().getRemotePath());
+        final boolean isSyncFolderRemotePathRoot = OCFile.ROOT_PATH.equals(syncFolderRemotePath);
 
-        if (currentDir == null) {
+        if (currentDir == null && !isSyncFolderRemotePathRoot) {
```
tobiasKaminsky, line 1360: While this works, the more important question is how can "currentDir" be null?
Can you set a breakpoint with condition "currendDir == null" and test it?
Related ticket might reveal some ideas how to reproduce, e.g. switching Activities

alperozturk96: The null check for currentDir already exists in the master branch. Purpose of this PR is not including the root path to the handleRemovedFolder function.


And this check logic is invoked from multiple places via SyncBroadcastReceiver. As previously discussed, this is related to the issue referenced here: https://github.com/nextcloud/android/issues/14855#issuecomment-2966196107.

Determining why currentDir is null depends on the specific broadcast event that triggers it and is beyond the scope of this pull request.

The relevant events:

```
FileSyncAdapter.EVENT_FULL_SYNC_START,
FileSyncAdapter.EVENT_FULL_SYNC_END,
FileSyncAdapter.EVENT_FULL_SYNC_FOLDER_CONTENTS_SYNCED,
RefreshFolderOperation.EVENT_SINGLE_FOLDER_CONTENTS_SYNCED,
RefreshFolderOperation.EVENT_SINGLE_FOLDER_SHARES_SYNCED

```

tobiasKaminsky (approved):

PR #15091, opened 2025-07-03
BugFix - Hide Keyboard After Search

alperozturk96:

### Issue

After performing a search and tapping the "DONE" button, the keyboard does not close and the cursor remains visible.

### Before

https://github.com/user-attachments/assets/527b63f8-6724-4b68-964c-052035eef02f

### After

https://github.com/user-attachments/assets/c83e11a4-a360-4477-ba5e-ce68bcce3a99

### Similar Behavior From Files, Phone, Maps Apps

https://github.com/user-attachments/assets/6292d7d0-9e6a-42c2-9e32-39860d783d80

https://github.com/user-attachments/assets/42b3d2b0-68bf-4ae8-8a6d-03498ee75938

https://github.com/user-attachments/assets/4a281e63-48e2-4261-95e0-b666d2ca233b

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/fragment/UnifiedSearchFragment.kt
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/UnifiedSearchFragment.kt
@@ -8,6 +8,7 @@
 package com.owncloud.android.ui.fragment
 
 import android.Manifest
+import android.content.Context
 import android.content.Intent
 import android.os.Bundle
 import android.view.LayoutInflater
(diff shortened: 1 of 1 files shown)


kra-mo:
LGTM, although one thing I notice is that search results (or at least suggestions) update live in these other apps without having to explicitly initiate a search.

I think this would be valuable to do for us too. For example, if after a maybe 500-1000ms timeout, no new input happened, initiate a search automatically.

The iOS client also does this.

tobiasKaminsky (approved):

PR #15107, opened 2025-07-07
BugFix - Share Permission Handling

alperozturk96:

### Issues

1. Since Nextcloud 30, the enabled key have been renamed to value and supports more than boolean. Thus, client needs to support both type.

[Documentation](https://docs.nextcloud.com/server/stable/developer_manual/client_apis/OCS/ocs-share-api.html#update-share)

2. When user change the permission from sharing details (quick permission bottom sheet ), share is not updating correctly.
3. Choosing `Can edit` permission from share settings screen and choosing `File request`from quick permission bottom sheet makes disable the bottom sheet actions.

### Notes

File request can only be set if share is public or mail and share is folder

Download limit can be only set if share is not folder and share is public

If user creates a folder share and use just read custom permission type it also means `View Only` thus showing `View Only `doesn't mean wrong, check web behavior.

If user creates a folder share and use just create custom permission type it also means `File request` thus showing `File request` doesn't mean wrong, check web behavior.

If user creates a file share and use just read custom permission type it also means `View Only` thus showing `View Only `doesn't mean wrong, check web behavior.

If user creates a file share and use read and edit custom permission type it also means `Can edit` thus showing ``Can edit` doesn't mean wrong, check web behavior.

### How to test?

**Scenario 1: Updating an Existing Share**

1. Set up two instances: one with Nextcloud 28 and another with Nextcloud version 30 or higher.
2. Create a share on both instances.
3. Update the share settings to enable **"Allow download and sync."**
5. Ensure the update is successfully completed.
6. Repeat for other share permission types

---

**Scenario 2: Creating a Share with Specific Permissions**

1. Set up two instances: one with Nextcloud 28 and another with Nextcloud version 30 or higher.
2. Create a share with **"Allow download and sync"** enabled.
3. Verify that the share is successfully created.
4. Repeat for other share permission types


tobiasKaminsky (dismissed):
sorting failes:
ShareeListAdapterTest -> testSorting()

PR #15132, opened 2025-07-11
Add Logs For ContentObserver

alperozturk96:

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/ContentObserverWork.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/ContentObserverWork.kt
@@ -32,10 +32,12 @@ class ContentObserverWork(
     override fun doWork(): Result {
         backgroundJobManager.logStartOfWorker(BackgroundJobManagerImpl.formatClassTag(this::class))
 
-        if (params.triggeredContentUris.size > 0) {
+        if (params.triggeredContentUris.isNotEmpty()) {
             Log_OC.d(TAG, "File-sync Content Observer detected files change")
             checkAndStartFileSyncJob()
             backgroundJobManager.startMediaFoldersDetectionJob()
+        } else {
+            Log_OC.w(TAG, "triggeredContentUris empty")
         }
         recheduleSelf()
 
@@ -60,6 +62,8 @@ class ContentObserverWork(
                 false,
                 changedFiles.toTypedArray()
             )
+        } else {
+            Log_OC.w(TAG, "cant startFilesSyncForAllFolders")
         }
     }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/nextcloud/client/jobs/ContentObserverWork.kt
```diff
@@ -32,10 +32,12 @@ class ContentObserverWork(
     override fun doWork(): Result {
         backgroundJobManager.logStartOfWorker(BackgroundJobManagerImpl.formatClassTag(this::class))
 
-        if (params.triggeredContentUris.size > 0) {
+        if (params.triggeredContentUris.isNotEmpty()) {
             Log_OC.d(TAG, "File-sync Content Observer detected files change")
             checkAndStartFileSyncJob()
             backgroundJobManager.startMediaFoldersDetectionJob()
+        } else {
+            Log_OC.w(TAG, "triggeredContentUris empty")
```
tobiasKaminsky: .d() is okay, since it can happen that uris are empty if e.g. a custom folder is synced

app/src/main/java/com/nextcloud/client/logger/ui/LogsAdapter.kt
```diff
@@ -16,23 +16,20 @@ import androidx.core.content.ContextCompat
...
     }
 
-    private val timestampFormat = SimpleDateFormat("yyyy-MM-dd HH:mm:ss", Locale.US)
     private val inflater = LayoutInflater.from(context)
 
     var entries: List<LogEntry> = listOf()
         @SuppressLint("NotifyDataSetChanged")
         set(value) {
-            field = value
+            field = value.sortedBy { it.timestamp }
```
tobiasKaminsky, line 32: Do we really wanna do this?
This means that oldest is on top, right?
So if you wanna see new stuff, you would need to scroll all the way down?

(I know it is tempting, since we both were confused about sorting order, but for quickly seeing latest logs, I think it is better to have newest first?)

alperozturk96: Newest to oldest is confusing. Since we can filter the logs, a relevant short log list sorted from oldest to newest will help us follow the code execution more clearly. Logcat also follows the same order, oldest to newest.

app/src/main/java/com/nextcloud/client/jobs/ContentObserverWork.kt
alperozturk96:
```suggestion
            Log_OC.d(TAG, "triggeredContentUris empty")
```

sowjanyakch (approved):

PR #15153, opened 2025-07-14
BugFix - External Share Icon Appreance After Delete

alperozturk96:

### Steps to Reproduce:
1. Launch the app and log in.
2. Tap the share icon on any file or folder.
3. Enter an external email address and create share. Return to the file list and observe the share icon.
4. Delete the external share from the share detail screen and observe the share icon again.

### Actual Result:
The share icon sometimes continues to display the "shared" state, or doesn't update at all even after a hard refresh.

### Expected Result:
After deleting the share, the icon should return to its default (unshared) state consistently.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/ShareeListAdapter.kt
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/ShareeListAdapter.kt
@@ -125,9 +125,7 @@ class ShareeListAdapter(
         1
     }
 
-    fun isAdapterEmpty(): Boolean {
-        return shares.isEmpty()
-    }
+    fun isAdapterEmpty(): Boolean = shares.isEmpty()
 
     @SuppressLint("NotifyDataSetChanged")
     fun toggleShowAll() {


app/src/main/java/com/owncloud/android/ui/fragment/FileDetailSharingFragment.java
```diff
@@ -650,6 +651,12 @@ public void sendNewEmail(OCShare share) {
     public void unShare(OCShare share) {
         unShareWith(share);
 
+        FileEntity entity = fileDataStorageManager.getFileEntity(file);
+        if (entity != null) {
+            entity.setSharedWithSharee(0);
```
tobiasKaminsky: Not tried it, but what if you have like this
- file with two sharees, e.g. share via mail and to user1
- unshare one of them
--> then this is triggered and set to false?
Then it will not show the sharees?

https://github.com/nextcloud/android/blob/11bc0daf57a067e87c0684051ebbef00ce4956a7/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java#L590

alperozturk96: Thank you for the feedback.

Since we are using `ShareeListAdapter` for both internal and external shares, and the fields `sharedWithSharee` and `sharedViaLink` are used as boolean indicators rather than actual share counts (where 1 represents that a share exists and 0 means no shares), we can simplify the logic.

After removing a share, we can check whether the corresponding adapter (sharesListInternal or sharesListExternal) is empty. If so, we can safely set `sharedWithSharee` or `sharedViaLink` to 0 to indicate that there are no remaining shares of that type.

tobiasKaminsky (approved):

PR #15157, opened 2025-07-15
BugFix - Hide Send Action If File Doesn't Have Correct Permission

alperozturk96:


alperozturk96:
"Send" button creates ShareDialog and two actions (action_send_file, action_send_share_file) used for it. These two actions needs to be checked.

Screenshot 2025-07-17 at 13 20 38

PR #15166, opened 2025-07-15
Bump search bar icon size 20->24dp for M3 styling

AndyScherzinger:

Icon size bumped for search bar according to https://m3.material.io/components/search/specs (file listing removed from screenshots since it contained production data)

|Before|After|
|---|---|
|before|after|


alperozturk96 (approved):

jancborchardt (approved):
Looks more consistent with the rest of the icons and not so tiny, nice.

PR #15172, opened 2025-07-16
Removed the jetifier flag to improve build performance

eeshanjamal:


tobiasKaminsky (approved):
Verified, we do not need it :+1:

PR #15174, opened 2025-07-16
Auto Sign off Git hook

eeshanjamal:

> Since I started contributing to this repo. I often miss doing Sign off for my commits and the reason for it's my habit of using Git GUI tools (such as SourceTree, Github Desktop) where you have to manually check Sign off for each commit. So, I've came up with this Git hook which auto Sign off my commits in-case I miss it doing manually. 
> 
> I found it useful and maybe it would be helpful for other developers as well contributing to this repo. So, I've parked it at the same place where other git hooks of this repo exist and created this pull request. You can review it and take final decision on wether to keep it or not.

### DIFF
--- a/scripts/hooks/prepare-commit-msg
+++ b/scripts/hooks/prepare-commit-msg
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+# Add Signed-off-by if not already present
+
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):
I like it 👍

PR #15184, opened 2025-07-17
Fix: App is ANR when uploadig 5000 items

tobiasKaminsky:


app/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -131,14 +131,14 @@ public long[] storeUploads(final List<OCUpload> ocUploads) {
...
 
-            OCUpload existingUpload = getPendingCurrentOrFailedUpload(ocUpload);
-            if (existingUpload != null) {
-                Log_OC.v(TAG, "Will update upload in db since " + ocUpload.getLocalPath() + " already exists as" +
-                    " pending, current or failed upload");
-                ocUpload.setUploadId(existingUpload.getUploadId());
-                updateUpload(ocUpload);
-                continue;
-            }
+//            OCUpload existingUpload = getPendingCurrentOrFailedUpload(ocUpload);
```
tobiasKaminsky: I am unsure if we really can remove that?

alperozturk96 (changes requested):
Manual upload not working.

1. Select some files to upload
2. Nothing happens

app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
```diff
@@ -196,13 +198,26 @@ class FileUploadWorker(
...
+            } else {
+                fileUploaderDelegate.sendBroadcastUploadFinished(
+                    operation,
+                    result,
+                    operation.oldFile?.storagePath,
+                    context,
+                    localBroadcastManager
+                )
+            }
         }
```
alperozturk96, line 216: ### Code suggestion

`private const val BROADCAST_BATCH_SIZE: Int = 100`

```
val shouldBroadcast = if (totalUploadSize > BROADCAST_BATCH_SIZE && currentUploadIndex > 0) {
    currentUploadIndex % BROADCAST_BATCH_SIZE == 0
} else {
    true
}

if (shouldBroadcast) {
    sendBroadcastUploadFinished(operation, result)
}
```


```
private fun sendBroadcastUploadFinished(operation: UploadFileOperation, result: RemoteOperationResult<*>) {
        fileUploaderDelegate.sendBroadcastUploadFinished(
            operation,
            result,
            operation.oldFile?.storagePath,
            context,
            localBroadcastManager
   )
}
```

alperozturk96:
@tobiasKaminsky One last thing :)

You can store as `private const val` inside the companion object and if you can add parentheses that would be good as well.

Screenshot 2025-08-18 at 18 02 44

tobiasKaminsky:
Please do not rebase others PRs.
I now changed the magic number, but cannot push it, since it is rebased.
Checking out latest status, will remove my change.

PR #15188, opened 2025-07-17
Add Debug Login

alperozturk96:

Allows login by long-pressing the thumbnail icon. Credentials can be configured in `.gradle/config.properties`, which is a local and ignored file intended for development purposes.

e.g: 

Screenshot 2025-07-17 at 15 46 23

Screenshot 2025-07-17 at 15 22 21

### Demo:

https://github.com/user-attachments/assets/6f8d52cf-f8ad-4413-95e3-3ad548fce468

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -85,6 +85,17 @@ file("$project.rootDir/ndk.env").readLines().each() {
 
 def perfAnalysis = project.hasProperty("perfAnalysis")
 
+def getConfigProperties() {
+    def props = new Properties()
+    def file = rootProject.file(".gradle/config.properties")
+    if (file.exists()) {
+        props.load(new FileInputStream(file))
+    }
+    return props
+}
+
+def configProps = getConfigProperties()
+
 android {
     // install this NDK version and Cmake to produce smaller APKs. Build will still work if not installed
     ndkVersion = "${ndkEnv.get("NDK_VERSION")}"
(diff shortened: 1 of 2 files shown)


tobiasKaminsky:
Hm. I think it is better to not have such a code in.
You can always use qrcode / login url: nc://login

This way I have a login.html, which I open on device and then have 20 different logins, with username/password.

alperozturk96:
> Hm. I think it is better to not have such a code in. You can always use qrcode / login url: nc://login
> 
> This way I have a login.html, which I open on device and then have 20 different logins, with username/password.

I do not see any issue or downside in using this code. It's for the debug build and is never going to be used in production. Technically, if you find any negative aspects, please feel free to share.

Additionally, I don’t want to use redirection or rely on any HTML file because I’m using different emulators (sometimes I need to wipe them) and the credentials keep changing, and I don't want to update that HTML file. In time, that HTML file will be huge, and I have to scroll and find a specific one. 

This way, I don’t have to do anything extra, no redirection, HTML file is needed. I can simply configure the credentials in the local Gradle file within Android Studio and build the app. That’s all.

This also allows anyone from the community to test faster, since they all use Android Studio. They can just adjust the local Gradle file and don’t need an HTML file.

alperozturk96:
@AndyScherzinger @mahibi @ZetaTom 

This improvement helping me a lot. I would like to get your opinion as well.

mahibi:
> @AndyScherzinger @mahibi @ZetaTom
> 
> This improvement helping me a lot. I would like to get your opinion as well.

i like the idea and can imagine to adapt it for the talk app.
Also, would like to know more about the login.html approach as well.

mahibi:
tested and works

PR #15210, opened 2025-07-21
Reduce HTTP Timeouts

alexknop:
I have a Nextcloud server with a private server address (192.168.1.2).

When launching the app on my cell phone when I had Wi-Fi off but Cellular data on, I noticed a spinning wheel for about a minute when Nextcloud starts.

I discovered there are many HTTP requests happening on app startup with timeouts of a minute. This can hinder the ability to open offline photos when in this state, possibly due to the many threaded processes.

I think 60 seconds is excessive and was able to manually set timeouts prior to these requests. I have them set to 3 seconds and the app is responding fine now, but I'm open to discussing a better solution to this.

The OwnCloud Client objects have a function to set the default timeouts, but the Nextcloud Client objects do not, and so I had to do so via reflection.

I do not see this issue when on other Wi-Fi networks, possibly because they also use the 192.168.1.0 network and the server availability is determined quicker, but since I use a private IP server, this behavior is noticeable when not on my home Wi-Fi.

Lastly, besides startup, I noticed longer times to load files via the DocumentStorageProvider when my phone is in this connectivity scenario and have a fix in here for that.


tobiasKaminsky:
Thanks for that idea.

NextcloudClient can be changed in android-library, thus then using the same timeouts.
I am not an expert on timeouts, thus I hesitate to set it to such a low value, just "because it works for you" (no offense).

However I am open for discussions how to improve/change.
Some ideas
- what is default on OKHTTP?
- how are other apps handling it?
- cache network state, so that only first request needs to time out?

alexknop:
> Thanks for that idea.
> 
> NextcloudClient can be changed in android-library, thus then using the same timeouts. I am not an expert on timeouts, thus I hesitate to set it to such a low value, just "because it works for you" (no offense).
> 
> However I am open for discussions how to improve/change. Some ideas
> 
>     * what is default on OKHTTP?
> 
>     * how are other apps handling it?
> 
>     * cache network state, so that only first request needs to time out?

Defaults for OkHttpClient appear to be 10 seconds from research, but I am not finding where or how it is changing to higher values in the app. The appcontext is passed into the builder and its possibly layered in the appcontext.

As for other apps, i am really not sure. There is a boolean in OkHttpClient for retryOnConnectionFailure which you could set to false and that would help the retry issue.

I know my case is unique, but I imagine there are other connection scenarios which could create this same pain point, such as having your server accessible over VPN to a private IP and this reacting the same way when not on VPN but connected to the internet. I am not sure if this same scenario applies to a publicly accessibly nextcloud server and the server is down.

alexknop:
@alperozturk96 @tobiasKaminsky  I changed this branch to only include one change to DocumentStorageProvider. Even though my KeePass Database.kdbx file is downloaded offline, when I go to open it from the KeePassDX app, the process to check eTag times out after 60 seconds when my server is not reachable. I've shortened the timeout to 3 seconds. It goes through 3 retries and now times out a lot sooner.

alperozturk96:
@alexknop 

Hey, thank you for the PR.

Since this change might behave differently depending on other use cases, I think it would be best to create a separate issue to discuss the implementation in detail. Once we align on that, we can move forward together.

I’m going to close this PR for now, just to keep things organized, but I really appreciate the effort you put in.

PR #15213, opened 2025-07-21
Add start your search to UnifiedSearchFragment

alperozturk96:

### Issue

After tapping the close button in the search bar within the UnifiedSearchFragment, the menu button appears in the toolbar menu instead of the back button.

### Before

Screenshot 2025-07-21 at 13 44 42

https://github.com/user-attachments/assets/a051e6e5-73d3-4cba-a1a8-2b250ac4880d

### After

https://github.com/user-attachments/assets/f029bc95-536c-4255-b881-0d18e252815b


alperozturk96:
> Looks good, but clicking on the `X` should also bring up the keyboard again as there is no point in clearing the text if you aren't going to search again :)

Ahh, miss that, thanks for the review.

PR #15217, opened 2025-07-21
BugFix - Prevent Empty All Files

alperozturk96:

### After 

https://github.com/user-attachments/assets/70a99d5f-f595-4680-87a4-d32c66b1e3ca

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
@@ -233,7 +233,7 @@ public class RefreshFolderOperation extends RemoteOperation {
         mConflictsFound = 0;
         mForgottenLocalFiles.clear();
 
-        if (mLocalFolder == null || mLocalFolder.getRemoteId() == null || mLocalFolder.getRemotePath() == null) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -437,7 +437,10 @@ private RemoteOperationResult fetchAndSyncRemoteFolder(OwnCloudClient client) {
         String remotePath = mLocalFolder.getRemotePath();
         RemoteOperationResult result = new ReadFolderRemoteOperation(remotePath).execute(client);
         Log_OC.d(TAG, "Refresh folder " + user.getAccountName() + remotePath);
-        Log_OC.d(TAG, "Refresh folder with remote id" + mLocalFolder.getRemoteId());
+
+        if (mLocalFolder.isRootDirectory()) {
```
tobiasKaminsky, line 441: Why only debug log when it is root folder?

tobiasKaminsky:
Do you have any steps to reproduce?
If local folder is null, then it should bubble up to an existing parent folder?

app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
alperozturk96: Root folder doesn't have remoteId. PR is still under development. I cannot reproduce it. I will add comment.

PR #15221, opened 2025-07-22
Show quota error message

tobiasKaminsky:

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
@@ -204,8 +204,9 @@ class FileUploadWorker(
             currentUploadFileOperation = null
 
             if (result.code == ResultCode.QUOTA_EXCEEDED) {
+                Log_OC.w(TAG, "Quota exceeded, stopping uploads")
                 notificationManager.showQuotaExceedNotification(operation, result.code)
-                continue
+                break
             }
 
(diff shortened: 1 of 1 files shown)


alperozturk96:
@tobiasKaminsky Please check this comment first before this PR: https://github.com/nextcloud/android/pull/14822#issuecomment-3101342416

app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
```diff
@@ -202,6 +202,13 @@ class FileUploadWorker(
 
             val result = upload(operation, user.get())
             currentUploadFileOperation = null
+
+            if (result.code == ResultCode.QUOTA_EXCEEDED) {
+                Log_OC.w(TAG, "Quota exceeded, stopping uploads")
+                notificationManager.showQuotaExceedNotification(operation, result.code)
+                break
```
alperozturk96, line 209: This check works, but it still allows one file to be uploaded.

For example, if the user has already exceeded their quota, the next time they attempt to upload one or more files, the first file will still be uploaded while the rest are stopped. To handle this correctly, we need to check the `QUOTA_EXCEEDED` result code in `UploadFileOperation` and prevent the upload at that stage.

tobiasKaminsky: My lazy idea was to always allow the next upload, so that we then check implicitly if enough quota is available again.
Else we would need to check before any upload via a separate call if enough space is available.

Issue #15218, opened 2025-07-21
Instant upload fails with "can't access local file" error


fatino:
### ⚠️ Before posting ⚠️

- [x] This is a **bug**, not a question or an enhancement.
- [x] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [x] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Have instant upload configured

### Expected behaviour

Photos and videos get uploaded 

### Actual behaviour

It shows error "can't access local file", even if access to storage is granted. Photos don't appear on next loud. Reconfiguring, forcing sync in settings. Nothing helps. Funny thing is, it shows thumbnails correctly in the app, so it "sees" the file. 

![Image](https://github.com/user-attachments/assets/d19668fb-10fe-4a6f-875d-c3df5d75454b)

### Android version

15

### Device brand and model

Oneplus 12

### Stock or custom OS?

Stock

### Nextcloud android app version

3.32.2

### Nextcloud server version

31.0.6

### Using a reverse proxy?

Yes

### Server error logs

```bash

```

### Additional information

I think it might be, because oneplus 12 system update that made photos and videos read only.

![Image](https://github.com/user-attachments/assets/72e57e46-c3bc-4f32-a97d-f4493f4d2422)

SanSiro8:
Same problem but on Xiaomi

schiessle:
I have the same problem on the Galaxy S24 Ultra, Android 15.

I just reconfigured the upload with the option "also upload already existing files" some where uploaded especially older ones. But newer pictures are not uploaded automatically. Also if I take new pictures. For me it looks like maybe the permissions changed on filesystem level at some point in time?

ChristophWurst:
> For me it looks like maybe the permissions changed on filesystem level at some point in time?

Same happened for Richard and me. This sounds like the issue discussed in https://github.com/nextcloud/android/issues/15039. Re-granting permissions manually solved new uploads for me.

AndyScherzinger:
^ @tobiasKaminsky @alperozturk96

schiessle:
@ChristophWurst removing and adding the permission again didn't solved the problem for me. But this one seems to work: https://github.com/nextcloud/android/issues/15039#issuecomment-3074371984

[...]

Issue #15039, opened 2025-06-20
Auto Upload: Local File not found


AndyXheli:
### ⚠️ Before posting ⚠️

- [x] This is a **bug**, not a question or an enhancement.
- [x] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [x] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

Enable autoupload
Taking pictures
Gives me error local file not found but the files are all there not sure what changed. 

### Expected behaviour

Should auto upload

### Actual behaviour

Not uploading 

### Android version

15

### Device brand and model

S25 Ultra

### Stock or custom OS?

Stock

### Nextcloud android app version

3.32.0 RC1

### Nextcloud server version

31.0.6

### Using a reverse proxy?

Yes

### Server error logs

```bash

```

### Additional information

![Image](https://github.com/user-attachments/assets/f7ba2664-1b59-4efe-9273-05f58ac388f2)

toinopt:
I have the same issue, only now did I realize that the local file not found was for files I actually had on the phone.
Noticed I am missing auto upload since a couple of days ago.

AndyXheli:
Yea not sure why the app sees the files then it says missing so for me AutoUpload dose not work at all

frxprince:
I also have the same issue.

CodeUnit:
The same problem happens under Android 16.

sprshp:
same issue here... 

phone : Oneplus 10 Pro NE2213_15.0.0.901 (EX01V80P01)
client : Nextcloud 3.32.0 RC1 (30320051)
server : Nextcloud Hub 10 (31.0.5)

Dirkxke:
I am experiencing the same problem on my S24 ultra.
My upload stopped somewhere after the 16th of June.

phone : S24 Ultra on Android 15
client : Nextcloud 3.32.0 RC1
server : Nextcloud Hub 10 (31.0.5snap1)

If someone can direct me to a log file, I am more than willing to upload it.

Edga-r:
Same issue.

phone : S20 Ultra on Android 13
client : Nextcloud 3.32.0 RC1
server : Nextcloud Hub 10 (31.0.6)

Nerodacles:
Same issue here!

[...]

PR #15240, opened 2025-07-23
Check Storage Permission Before Upload

alperozturk96:

### Issue

Previously granted `MANAGE_EXTERNAL_STORAGE` or `Only Media Access` permission can be revoked if the app hasn't been used for a few months (e.g. if user not opens the app just using the app for auto upload)  or after a system upgrade.

Screenshot 2025-07-23 at 13 35 00

Some other OS implementations may handle this differently, so we need to cover this edge case as well.
Also, it seems that some users have experienced this issue before and resolved it by re-granting the permission.

https://github.com/nextcloud/android/issues/15218#issuecomment-3106054821
https://github.com/nextcloud/android/issues/15039#issuecomment-3090533342

### Changes

<table>
  <tr>
    <td>Image 1</td>
  </tr>
</table>

### Notes

`PermissionUtil.requestExternalStoragePermission` cannot be used to display the same permission dialog that appears at the beginning of the application, **as it depends on `AppCompatActivity`**. Therefore, we inform the user and provide them with the option to choose between the standard app permission or full file access.

@ZetaTom 

Additionally, instead of handling the error after a failed upload, we proactively check for the necessary permissions before initiating the upload. This approach prevents unnecessary calls and simplifies the logic flow.

### All files access action

Image 1

### App permissions action

We cannot redirect to the directly to the storage permission page. There is no intent action for that.

Image 2


app/src/main/java/com/nextcloud/client/jobs/FilesSyncWork.kt
```diff
@@ -52,20 +56,29 @@ class FilesSyncWork(
...
 
     private lateinit var syncedFolder: SyncedFolder
+    private val notificationManager = AutoUploadNotificationManager(context, viewThemeUtils, NOTIFICATION_ID)
 
-    @Suppress("MagicNumber", "ReturnCount")
+    @Suppress("MagicNumber", "ReturnCount", "LongMethod")
     override fun doWork(): Result {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R && !Environment.isExternalStorageManager()) {
+            Log_OC.w(TAG, "AutoUpload process cancelled - missing storage permission")
+            notificationManager.showStoragePermissionNotification()
```
tobiasKaminsky: Would this then be shown on each run?
What if the user dismisses/ignores notification?

alperozturk96: If the user dismisses it, the notification will appear again on the next attempt (since it is periodic worker) if the permission is still not granted.

ZetaTom (changes requested):
During my testing everything worked as expected when I granted the *Manage all files* permission. However, when I choose *Media read-only* access I still get the notification and the upload is cancelled. Auto-Upload should also work when users choose media instead of full access.

To provide feedback for missing permissions and not break any use cases, I think that we need to check if we have the appropriate permissions for that folder on every sync run.

From a UI/UX point of view, I think that the button on the notification should say something more descriptive than *OK*.

alperozturk96:
We can check the file access permission from UploadFileOperation instead of checking the storage permission and notify user.

alperozturk96:
Looks like full file access permission fixed the issue, please read the issue and comments.

https://github.com/nextcloud/android/issues/15039#issuecomment-3037926879
https://github.com/nextcloud/android/issues/15039#issuecomment-3038263636
https://github.com/nextcloud/android/issues/15039#issuecomment-3038352257
https://github.com/nextcloud/android/issues/15039#issuecomment-3090533342
https://github.com/nextcloud/android/issues/15039#issuecomment-3160161130

app/src/main/java/com/owncloud/android/operations/upload/UploadFileOperationExtensions.kt
```diff
@@ -0,0 +1,67 @@
...
+            .setSmallIcon(android.R.drawable.stat_sys_warning)
+            .setContentTitle(context.getString(R.string.upload_missing_storage_permission_title))
+            .setContentText(context.getString(R.string.upload_missing_storage_permission_description))
+            .setPriority(NotificationCompat.PRIORITY_HIGH)
+            .addAction(allowAllFileAccessAction)
+            .addAction(appPermissionsAction)
+            .setAutoCancel(true)
+
+    ContextCompat.getSystemService(context, NotificationManager::class.java)?.run {
+        notify(MISSING_FILE_PERMISSION_NOTIFICATION_ID, notificationBuilder.build())
```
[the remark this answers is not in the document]
alperozturk96: fixed

alperozturk96:
Move checks to `checkConditions` function.

app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -405,9 +409,31 @@ public Context getContext() {
...
+        final var localFile = new File(getStoragePath());
+        if (!localFile.exists()) {
+            Log_OC.e(TAG, "Upload cancelled for " + getStoragePath() + ": local file not exists.");
+            return new RemoteOperationResult<>(ResultCode.LOCAL_FILE_NOT_FOUND);
+        }
+
+        if (!localFile.canRead()) {
+            Log_OC.e(TAG, "Upload cancelled for " + getStoragePath() + ": file is not readable or inaccessible.");
+            UploadFileOperationExtensionsKt.showStoragePermissionNotification(this);
+            missingPermissionThrown.set(true);
```
tobiasKaminsky, line 433: This is set for one operation, but what if we have multiple runs?
E.g. user uploads 5x, one after the other?

Then it is always being reset and the user will receive 5x the "error" message?

alperozturk96: It's using same notification id also I added check for active notification thus it will only display one notification.


https://github.com/user-attachments/assets/a3b21f7b-802b-47f4-a6c5-31c47d239374

app/src/main/java/com/owncloud/android/operations/upload/UploadFileOperationExtensions.kt
```diff
@@ -0,0 +1,76 @@
...
+        NotificationCompat.Builder(context, NotificationUtils.NOTIFICATION_CHANNEL_UPLOAD)
+            .setSmallIcon(android.R.drawable.stat_sys_warning)
+            .setContentTitle(context.getString(R.string.upload_missing_storage_permission_title))
+            .setContentText(context.getString(R.string.upload_missing_storage_permission_description))
+            .setPriority(NotificationCompat.PRIORITY_HIGH)
+            .addAction(allowAllFileAccessAction)
+            .addAction(appPermissionsAction)
+            .setAutoCancel(true)
+
+    notificationManager.notify(MISSING_FILE_PERMISSION_NOTIFICATION_ID, notificationBuilder.build())
```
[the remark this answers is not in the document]
alperozturk96: https://github.com/nextcloud/android/security/code-scanning/246

Screenshot 2025-08-14 at 11 08 46

---

Already immutable and has specific action.

```
private fun getActionPendingIntent(context: Context, actionType: UploadFileBroadcastReceiverActions): PendingIntent {
    val intent = Intent(context, UploadFileBroadcastReceiver::class.java).apply {
        action = "com.owncloud.android.ACTION_UPLOAD_FILE_PERMISSION"
        putExtra(UploadFileBroadcastReceiver.ACTION_TYPE, actionType)
    }

    return PendingIntent.getBroadcast(
        context,
        actionType.ordinal,
        intent,
        PendingIntent.FLAG_IMMUTABLE
    )
}
```

PR #15253, opened 2025-07-25
BugFix - Stop Re-Upload Same File Due To File Extension

### LINKED ISSUE #14996: upload changes case of file extension to lowercase
### ⚠️ Before posting ⚠️

### Steps to reproduce

1. navigate to a folder that you can uploada file
2. select a file to upload
3. goto upload screen (paused). FILE.JPG will now be FILE.jpg waiting to upload
4. this will create a duplicate file if you uploaded file previously 

### Expected behaviour

case of file extension should remain the same as stored on android device..JPG stays .JPG (uppercase) and not .jpg (lowercase)

### Actual behaviour

file extension is changed to lowercase. .JPG (uppercase) is being sent .jpg (lowercase).
files uploaded before as KIMG0001.JPG are now uploading and creating duplicates files named like KIMG0001.jpg. 
this is duplicated on 10,000+ images and files ...

### Android version

12

### Device brand and model

kyocera

### Stock or custom OS?

Stock

3.31.4

### Nextcloud server version

31.0.5

### Using a reverse proxy?

No

### Android logs

_No response_

### Server error logs

```bash

```

### Additional information

_No response_

alperozturk96:

### Issue

When uploading a file, the app changes the file extension to lowercase (e.g., .JPG becomes .jpg). This leads to duplicate uploads if the same file was previously uploaded with the original (uppercase) extension, since the system may treat `.JPG` and `.jpg` as different files. 

This behavior has resulted in thousands of unintended duplicates.

### How to reproduce?

1. Have a file in folder, e.g. File.JPG
2. Try to upload exact same file but with lowercased extension e.g. File.jpg

https://github.com/user-attachments/assets/01840d0a-1726-48f9-897a-867251d390da

### Solution

If one of the remote path variant (File.JPG or File.jpg) exists in remote and file didn't change, skip the upload process.

@mpivchev Please check this for iOS as well. @tobiasKaminsky fyi.

### After

https://github.com/user-attachments/assets/095e3e87-82fd-4d38-a413-8a4a1fd82e4a

### How to test?

1. Files with the exact same name, content and only differing in upper- or lowercase extensions should not be re-uploaded.
2. If a file has the same name but different content, a conflict resolution dialog must be shown.


app/src/main/java/com/nextcloud/client/jobs/operation/FileOperationHelper.kt
```diff
@@ -28,6 +33,54 @@ class FileOperationHelper(
...
+        }
+
+        return if (result.isSuccess) {
+            result.data[0] as? RemoteFile
+        } else {
+            null
+        }
+    }
+
+    fun isFileChanged(remoteFile: RemoteFile?, ocFile: OCFile?): Boolean =
```
alperozturk96, line 81: @tobiasKaminsky Is this enough? Since we are going to use it just before the upload.

app/src/main/java/com/nextcloud/client/jobs/operation/FileOperationHelper.kt
```diff
@@ -12,25 +12,85 @@ import com.nextcloud.client.account.User
...
+        return false
+    }
+
+    fun isSameFileOnRemote(ocFile: OCFile, upload: OCUpload): Boolean {
+        val localFile = File(upload.localPath)
+        if (!localFile.exists()) {
+            return false
+        }
+        val localSize: Long = localFile.length()
+        return ocFile.fileLength == localSize
```
alperozturk96: @tobiasKaminsky @ZetaTom 

I need to compare a local file intended for upload with an existing remote file to determine if they are exactly the same. While file size can be used as a basic check, I’m looking for a more reliable method.

Using `eTags` is not feasible in this case, as the file to be uploaded (`OCUpload`) does not have an `eTag` yet. Similarly, relying on the modification timestamp is unreliable, since it can change due to non-content-related operations such as permission updates or file copying.

At this point, I have a `java.io.File` object and an `OCFile` instance, and I need to determine whether they represent the same file content. 

Any suggestions for a more robust comparison method beyond file size?

app/src/main/java/com/nextcloud/client/jobs/offlineOperations/OfflineOperationsWorker.kt
```diff
@@ -126,10 +125,10 @@ class OfflineOperationsWorker(
             return@withContext null
         }
 
-        val remoteFile = getRemoteFile(path)
+        val remoteFile = fileOperationHelper.getRemoteFile(path, client)
```
alperozturk96, line 128: `OfflineOperationWorker` was updated to use the logic in `FileUtil` to prevent code duplication and unify similar functionality.

app/src/main/java/com/owncloud/android/authentication/LoginUrlInfo.kt
```diff
@@ -13,6 +13,8 @@ import com.nextcloud.model.HTTPStatusCodes
 data class LoginUrlInfo(var server: String, var loginName: String, var appPassword: String) {
```
alperozturk96: Changed because of the Kotlin spotless apply.

`./gradlew spotlessApply`

app/src/main/java/com/nextcloud/client/jobs/operation/FileOperationHelper.kt
```diff
@@ -10,27 +10,77 @@ package com.nextcloud.client.jobs.operation
...
+     * ```
+     */
+    fun isSameRemoteFileAlreadyPresent(upload: OCUpload, storageManager: FileDataStorageManager): Boolean {
+        val (lc, uc) = FileUtil.getRemotePathVariants(upload.remotePath)
+
+        val remoteFile = storageManager.run {
+            getFileByDecryptedRemotePath(lc) ?: getFileByDecryptedRemotePath(uc)
+        }
+
+        if (upload.toFile()?.length() == remoteFile?.fileLength) {
```
alperozturk96, line 52: @tobiasKaminsky @ZetaTom 

I need to compare a local file intended for upload with an existing remote file to determine if they are exactly the same. While file size can be used as a basic check, I’m looking for a more reliable method.

Using `eTags` is not feasible in this case, as the file to be uploaded (`OCUpload`) does not have an `eTag` yet. Similarly, relying on the modification timestamp is unreliable, since it can change due to non-content-related operations such as permission updates or file copying.

At this point, I have a `java.io.File` object and an `OCFile` instance, and I need to determine whether they represent the same file content. 

Any suggestions for a more robust comparison method beyond file size?

fabolhak:
Unfortunately, this change makes it worse for my use case :cry:. I use to following workflow to upload my pictures (with ending .JPG):

- open the inbuilt Files app
- select all images I want to store on my nextcloud instance
- move all files to the specific folder using the Nextcloud storage provider in files app

With the previous version two files were created: 
- empty .JPG file
- actual .jpg file

For some reason the .JPG files were never deleted and I had to clean them up manually. 

With the linked version above, I only get the following file:
- empty .JPG file

And the file with actual content is lost :(

So please consider reworking the whole "changing of file extension to lowercase feature" as it can cause a lot of trouble. 


[logs.txt](https://github.com/user-attachments/files/21744547/logs.txt)

tobiasKaminsky:
- auto rename should only be working when "enforce windows compatible mode" is enabled, which is not clear by server right now (server justs adds some additional config to capabilities)
- if auto-rename is enabled, then our app should check before upload, if a renamed file exists and if yes, it should show the conflict resolution screen

fabolhak:
I have "enforce windows compatible mode" disabled. Still the file extensions get renamed from upper case to lower case: JPG -> jpg or MP4 -> mp4

Uploading via the Nextcloud app works fine (still does the renaming), but uploading via the inbuilt File browser is broken.

PR #15255, opened 2025-07-25
Add Remove Upload Action to Conflict Resolve Notification

alperozturk96:

### Changes

When the user selects the `Remove Upload` action, the conflicting upload will be cleared. As a result, previously dismissed notifications will not reappear upon app launch or when the user refreshes the directory.

https://github.com/user-attachments/assets/85fd371e-1dd3-4525-b887-13dbd780a78a

<table>
  <tr>
    <td>Image 1</td>
  </tr>
</table>

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1071,6 +1071,7 @@
     <string name="uploader_upload_failed_sync_conflict_error">File upload conflict</string>
     <string name="uploader_upload_failed_sync_conflict_error_content">Pick which version to keep of %1$s</string>
     <string name="upload_list_resolve_conflict">Resolve conflict</string>
+    <string name="upload_list_remove_upload">Remove upload</string>
     <string name="upload_list_delete">Delete</string>
     <string name="create_new">New</string>
     <string name="editor_placeholder" translatable="false">%1$s %2$s</string>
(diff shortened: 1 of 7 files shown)


mahibi (approved):

kra-mo (dismissed):
> When the user dismisses the notification

That sounds scary! Usually, no action should happen on dismissal of a notification, but especially none that could lead to data loss.

I'm pro having the button but anti the dismissal behavior if I understand correctly.

kra-mo:
And especially because some vendors of Android might include features for bulk "cleaning up" notifications or whatnot.

app/src/main/res/values/strings.xml
```diff
@@ -1071,6 +1071,7 @@
     <string name="uploader_upload_failed_sync_conflict_error">File upload conflict</string>
     <string name="uploader_upload_failed_sync_conflict_error_content">Pick which version to keep of %1$s</string>
     <string name="upload_list_resolve_conflict">Resolve conflict</string>
+    <string name="upload_list_remove_upload">Remove upload</string>
```
kra-mo: It's also not entirely clear to me which version this button would remove. I'm guessing the newer one, but it could be made clearer by saying "cancel" as only the newer one is being actively uploaded.

```suggestion
    <string name="upload_list_remove_upload">Cancel upload</string>
```

alperozturk96:
> > When the user dismisses the notification
> 
> That sounds scary! Usually, no action should happen on dismissal of a notification, but especially none that could lead to data loss.
> 
> I'm pro having the button but anti the dismissal behavior if I understand correctly.

Thank you for the feedback. I was undecided but yes it makes sense.

kra-mo (approved):

Issue #15214, opened 2025-07-21
Deep dive Auto Upload


tobiasKaminsky:
- debug current problems, e.g. with debug build handed out to users
- fix bugs
- write down improvements

thawn:
Thank you for diving into this. As mentioned in #15039, for me the local file not found auto upload issues were triggered by an android system update:
Build BP2A.250605.031.A2
Pixel 7a
Android 16

Also, enabling `special permissions -> all file access` (as opposed to only `Photos and Videos`) worked for me.

alperozturk96:
### Worker Chaining Problems

Currently, the `ContentObserverWork` triggers `FilesSyncWork`, which in turn triggers `FileUploadWorker`, which finally calls `UploadFileOperation`. 

This chained architecture introduces fragility to the Auto Upload mechanism. If any call in the chain such as `FileUploadWorker` or `UploadFileOperation` fails or behaves unexpectedly due to some code changes in any of it, it can disrupt the entire auto upload process. This tightly coupled workflow increases the risk of failure and makes debugging more complex.

Additionally, deep worker chaining can cause this edge case, especially on Android 12 and above, [where the system limits each app to 100 scheduled jobs](https://developer.android.com/reference/android/app/job/JobScheduler). Exceeding this limit can lead to runtime exceptions, redundant executions, and premature worker terminations.

To improve reliability and maintainability, a single source of truth approach should be adopted. Ideally, there should be one worker responsible for handling the auto upload logic, and this worker should internally manage the execution of file upload operations, rather than delegating through chained workers.

### FilesSyncWork Internal Logic Problems

**Early Exit Logic (canExitEarly()):**

The `canExitEarly()` method is used to exit the auto upload process early. However, its current implementation, combined with scan interval logic, can lead to premature exits, causing uploads to be skipped. This logic should be reviewed and potentially moved outside the worker. e.g., handled via `PeriodicWorkRequest.Builder.`

**Lack of Failure Handling:**

`FilesSyncWork` currently does not return failure or retry results. This suppresses visibility into failed executions and prevents proper application of retry or backoff policies.

**Missing Backoff Policy:**

In cases of failure (e.g., due to poor network connectivity), the worker should automatically retry using a defined [`BackoffPolicy`](https://developer.android.com/reference/androidx/work/BackoffPolicy), either EXPONENTIAL or LINEAR. 

**Missing Constraints:**

[...]

Issue #15051, opened 2025-06-24
Force rescan and upload missing images (and run this regularly)


jacksbox:
A button to force a rescan of all auto upload folders (and compare to the remote folder) and upload missing files would be great. It seems like a easy fix for the regularly the occurring bugs regarding missing uploaded photos.

This could even run once a day (getting a file list from the server and comparing it to a local one should not be to performance impacting, no?) to catch any maybe occurring problems?

fvigotti:
photos missing in the nextcloud upload can lead to a silent loss of photos,
many use the application as a backup to the cloud, and it seems to often make upload errors ( ie: for example today I took 10 photos, it only uploaded the last one), 
but who knows how many are missing in the cloud, and not only does it make these errors but it seems to lack the possibility of reporting it to the application and also having the security of a scheduled rescan as it says @jacksbox

Th0masL:
I agree. I have AutoUpload enabled on my photo folder `DCIM/Camera` and I have sometimes to wait hours before the photo gets uploaded to my Nextcloud instance.

The AutoUpload on that folder is configured as "Only upload on unmetered Wi-Fi" and "Also upload existing files".

I'm at home most of the time (on Wi-Fi), so it should upload right away.

SolidTux:
I also would like this feature. I actually have a folder that is missing quite a lot of pictures that I had to manually upload.

microniko:
I think this feature could be very usefull !

doits:
👍 Sometimes some files/photos don't get uploaded, autoupload is a little bit buggy. A button to rescan the folder and upload all missing files on the server would solve this.

Shouldn't the logic be there already with the "also upload existing files" toggle when new auto upload is enabled? It could do the same, upload all existing files again with the only exception to automatically skip existing ones (and not asking the user in this case).

doits:
... I guess #15267 is doing this?

gbakeman:
Note [this](https://github.com/nextcloud/android/issues/15214) meta issue about auto upload. The last pull request is in the 3.35 milestone, so I think we can anticipate some big auto upload enhancements coming soon. Hopefully the need to manually rescan can be replaced entirely by the structual changes that have been made.

DenisBalan:
[...]

PR #15267, opened 2025-07-28
Add forceSync option to auto-upload items

qwe2:
Sometimes files can get missed if they get marked as FILESYSTEM_FILE_SENT_FOR_UPLOAD but then do not get updated for whatever reason, as FilesystemDataProvider#getFilesForUpload only considers files where FILESYSTEM_FILE_SENT_FOR_UPLOAD=0. This change adds an explicit "Force sync" button which can be used to invalidate the client-side cache and re-enumerate all files in a given folder.

Screenshots:
Screenshot_20250728_122933 Screenshot_20250728_123043

### DIFF
--- a/app/src/androidTest/java/com/nextcloud/client/SyncedFoldersActivityIT.java
+++ b/app/src/androidTest/java/com/nextcloud/client/SyncedFoldersActivityIT.java
@@ -9,17 +9,22 @@ package com.nextcloud.client;
 
 import android.content.Intent;
 import android.os.Looper;
+import android.view.Menu;
+import android.view.MenuItem;
 
 import com.nextcloud.client.preferences.SubFolderRule;
 import com.owncloud.android.AbstractIT;
+import com.owncloud.android.R;
 import com.owncloud.android.databinding.SyncedFoldersLayoutBinding;
 import com.owncloud.android.datamodel.MediaFolderType;
 import com.owncloud.android.datamodel.SyncedFolder;
 import com.owncloud.android.datamodel.SyncedFolderDisplayItem;
 import com.owncloud.android.ui.activity.SyncedFoldersActivity;
+import com.owncloud.android.ui.adapter.SyncedFolderAdapter;
 import com.owncloud.android.ui.dialog.SyncedFolderPreferencesDialogFragment;
 import com.owncloud.android.utils.ScreenshotTest;
 
+import org.junit.Assert;
 import org.junit.Rule;
 import org.junit.Test;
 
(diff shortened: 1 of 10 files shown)


tobiasKaminsky:
Thank you very much for your contribution.
I wonder what happens if the user has "what to do if file exists -> always ask" enabled and then force sync?
It should then show n time a notification?
This should ideally be not the case?

doits:
Thank you very much, this would be a really great feature to fix missing photos in auto upload!

>  I wonder what happens if the user has "what to do if file exists -> always ask" enabled and then force sync?
> It should then show n time a notification?
> This should ideally be not the case?

I think so, too. For those files that exist because of the forced sync it should not show a notification. But to get this feature out faster it would be OK to follow the setting if it is hard to implement to skip those files.

How is this handled when setting up a new auto upload folder with the "upload existing files too" toggle? Is it showing the notification in this case if a file exists already? Or is this case handled there already and the logic could be applied here, too?

Torbins:
This is extremely useful, I hope it will be merged soon.

qwe2:
I've been a bit busy lately, so didn't have much time for this, but I'm planning on following this up. 

> I wonder what happens if the user has "what to do if file exists -> always ask" enabled and then force sync?
> It should then show n time a notification?
> This should ideally be not the case?

I will need to test how the sync behaves with various options set to say for sure, but my assumption is that it should behave the same way as if you were setting up the folder for the same time. If that flow shows `n` notifcations as well, then this will too. For now, only "skip uploading" is tested, as that's the option I'm using.

I'd welcome suggestions on how to handle this. I have some ideas, but I don't really love any of them...
* "Foce-sync" could be disabled if the settings are not "optimal".
* The popup could have an "ignore notifications" chechbox. Not sure how difficult this would be to implement.
* The popup could offer the user to change the "existing file" behavior temproarily just for the duration of the "force sync" run. (This sounds best from UX POV, but not having experience with the codebase, I'm pretty afraid this is very complicated to implement. I don't know what to do about the transactionality of the folder settings, e.g. how to change one setting for a scheduled task, then guarantee to always revert it once the task is done, whatever "done" even means here...)

Torbins:
As a user, I would expect the same behavior that you get when you setup new synchronization, but choose already existing folder on server side.

doits:
Any updates on this? Got ~30 photos that didn't sync since July and cannot make auto upload upload them. Even removing the folder from auto upload and adding it again, telling it to upload existing images, does not make it work afaik.

alperozturk96:
@doits @Torbins 

Hi

We’re currently working on improving the auto upload feature, and we’d love your help testing a new QA version. You can download it [[here](https://cloud.nextcloud.com/s/rzH8xoem9WG3z66)] and share your experience with us.

We’re aware of the current limitations with auto upload and explained the details [[in this issue](https://github.com/nextcloud/android/issues/15214)]. Rest assured, we’re actively working on making it more reliable.

It would also be very helpful if you could share the related logs with us. To do this, open **Settings → Logs** and filter using the keywords **“AutoUpload”** and **“Upload.”** You can then upload your logs [[at this link](https://cloud.nextcloud.com/s/nXi8CKpfBJrSqKn)].

jpradelle:
I have completely messed up sync folders, that's exactly the feature I'm looking for

lpcvoid:
This is such a great feature, please add it!

barrelltitor:
> @doits @Torbins
> 
> Hi
> 
> We’re currently working on improving the auto upload feature, and we’d love your help testing a new QA version. You can download it [[here](https://cloud.nextcloud.com/s/rzH8xoem9WG3z66)] and share your experience with us.
> 
> We’re aware of the current limitations with auto upload and explained the details [[in this issue](https://github.com/nextcloud/android/issues/15214)]. Rest assured, we’re actively working on making it more reliable.
> 
> It would also be very helpful if you could share the related logs with us. To do this, open **Settings → Logs** and filter using the keywords **“AutoUpload”** and **“Upload.”** You can then upload your logs [[at this link](https://cloud.nextcloud.com/s/nXi8CKpfBJrSqKn)].

The share is not found, can you share this again? Would love to test

DenisBalan:
Any updates? Missing feature when sync state is broken

Re https://github.com/nextcloud/android/issues/15051

tobiasKaminsky:
I think we have a pull to refresh, so that this is not needed in this way.
@alperozturk96 can you check this?

barrelltitor:
A button is always better than a hidden gesture. I had no idea this even existed and decided to just switch to immich for all photos and videos

Torbins:
> I think we have a pull to refresh, so that this is not needed in this way. @alperozturk96 can you check this?

Auto upload screen does not implement pull to refresh. At least not in release build, that I have now.

batorjian:
Is this Feature already in stable Release? I can't find it, neither in f-droid stable nor nextcloud-dev-version. But i really miss this Feature!

PR #15299, opened 2025-08-04
Optimize upload file data loading

tablebird:
1: UploadListActivity now loads data once instead of three times after launching
2: RecyclerView refresh animation is now associated with data

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
@@ -57,8 +57,11 @@ import com.owncloud.android.utils.theme.ViewThemeUtils;
 
 import java.io.File;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
```diff
@@ -976,6 +975,54 @@ public void apply() {
...
+                }
+                notifyDataSetChanged();
+                for (LoadCompleteListener loadCompleteListener : loadCompleteListenerSet) {
+                    loadCompleteListener.onComplete();
+                }
+            });
+        }
+    }
+
+    private static class UploadGroupLoadPolicy implements RejectedExecutionHandler {
```
alperozturk96, line 1009: Hello, thank you for the PR.

Could you replace via this to make it more readable?

```
private static class UploadGroupLoadPolicy implements RejectedExecutionHandler {
        @Override
        public void rejectedExecution(Runnable runnable, ThreadPoolExecutor executor) {
            if (executor.isShutdown()) {
                return;
            }

            Runnable oldest = executor.getQueue().poll();

            if (oldest instanceof UploadGroupLoadRunnable oldUploadTask && runnable instanceof UploadGroupLoadRunnable newUploadTask) {
                if (!oldUploadTask.loadCompleteListenerSet.isEmpty()) {
                    newUploadTask.loadCompleteListenerSet.addAll(oldUploadTask.loadCompleteListenerSet);
                }
            }

            executor.execute(runnable);
        }
    }
```

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
tablebird: replace completed

alperozturk96:
Tested working.

PR #15318, opened 2025-08-06
Fix grid view spacing

ZetaTom:
This change addresses several issues reported to us by users of the grid view. Several changes have been made to increase usability and consistency.

- file names of folders and images behave the same
- long file names no longer overlap the *more* (⋮) button
- *more* button is no longer cut off when in four or five column mode
- *more* button isn't pushed off screen any more
- better scaling on larger displays
- increased space for file names
- reduced duplicated code by unifying layouts

### Screenshot

Current UX / UI | This Pull Request
--- | ---
old_01 | new_01
old_02 | new_02
old_03 | new_03
old_04 | new_04
old_05 | new_05
old_06 | new_06

### Design Question
Currently, when using the five column mode, all file names are hidden accept for images and videos. From a usability perspective, I think that we should either hide or show all file names without any exceptions as it is rather common to have multiple file types next to one another.

As I find it hard to distinguish two folders, which are displayed identically in the five column mode, without the corresponding file names, I think that we should always show the file name. Given that we already had code in place to hide the file name under certain conditions, I chose not to do any changes to that for the time being. If we decide on one option, it's an easy change to make.

If we decide to hide the file name, we need to consider if we should hide the *more* button as well.

---


app/src/main/java/com/owncloud/android/ui/adapter/LocalFileListAdapter.java
```diff
@@ -503,14 +484,16 @@ private LocalFileListItemViewHolder(View itemView) {
         }
     }
 
-    static class LocalFileListGridImageViewHolder extends RecyclerView.ViewHolder {
+    static class LocalFileListGridItemViewHolder extends RecyclerView.ViewHolder {
```
alperozturk96: This class can be private.

ZetaTom: Visibility has been reduced with 812f54e.

ZetaTom:
> Looks good, except these 2 before/after screenshots where the filename fully vanishes:
> ...

@jancborchardt, this was a quirk of a bit of logic that was already in place but just broken. Because I wasn't sure what to do, I posed the *Design Question* in the description.

During a design review with @kra-mo and @alperozturk96 we came to the conclusion that file names should always be displayed for now. As I haven't updated the screenshots to reflect the latest changes, they are now outdated.

This is what the UI looks like in its current state:

Five Column Mode | Four Column Mode
--- | ---
Five Column Mode | Four Column Mode

I'll open another issue later on to further enhance the grid UI using dynamic layouts and constraints based on device parameters like screen width and scaling.

jancborchardt (commented):
Sounds good to me @ZetaTom – I’d even go so far and say that if it doesn’t work nicely, then we shouldn’t have a 5 column mode at all. At least for devices where the overall width doesn’t allow for having enough width for each entry then.

jancborchardt (approved):
Sounds good to me @ZetaTom – I’d even go so far and say that if it doesn’t work nicely, then we shouldn’t have a 5 column mode at all. At least for devices where the overall width doesn’t allow for having enough width for each entry then.

ZetaTom:
> Sounds good to me @ZetaTom – I’d even go so far and say that if it doesn’t work nicely, then we shouldn’t have a 5 column mode at all. At least for devices where the overall width doesn’t allow for having enough width for each entry then.

Exactly. We discussed that we should somehow determine how may columns fit nicely on a given device and then set a limit based on that dynamically. However, I think that we should keep the higher column modes for now as these also apply to larger devices, like tablets, for example. :)

alperozturk96:
Screenshot 2025-08-12 at 10 48 45

PR #15319, opened 2025-08-06
Refactor ExtendedListFragment

ZetaTom:
Refactor `ExtendedListFragment.kt` to remove depreciations and warnings.

---


app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
```diff
@@ -121,9 +120,9 @@ open class ExtendedListFragment :
     private var mEmptyListIcon: ImageView? = null
 
     // Save the state of the scroll in browsing
-    private var mIndexes: ArrayList<Int?>? = ArrayList<Int?>()
-    private var mFirstPositions: ArrayList<Int?>? = ArrayList<Int?>()
-    private var mTops: ArrayList<Int?>? = ArrayList<Int?>()
+    private var mIndexes: ArrayList<Int?>? = ArrayList()
```
alperozturk96:
```
private var mIndexes = arrayListOf<Int>()
private var mFirstPositions = arrayListOf<Int>()
private var mTops = arrayListOf<Int>()
```

Please make it non-nullable and handle it accordingly. Since it doesn't need to be null, and we're already checking if it's empty, it's better to keep it non-nullable.

Image

app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
```diff
@@ -190,40 +190,36 @@ open class ExtendedListFragment :
...
-                        (getActivity() !is UploadFilesActivity)
-                    ) {
-                        if (getActivity() is FileDisplayActivity) {
-                            val fragment = (getActivity() as FileDisplayActivity).leftFragment
-                            if (fragment is OCFileListFragment) {
-                                fragment.setFabVisible(!hasFocus)
-                            }
+        searchView?.setOnQueryTextFocusChangeListener { _: View?, hasFocus: Boolean ->
+            lifecycleScope.launch(Dispatchers.Main) {
+                if (getActivity() != null &&
```
alperozturk96:
```
val activity = getActivity()

if (
    activity != null &&
    activity !is FolderPickerActivity &&
    activity !is UploadFilesActivity
) {
    val ocFileListFragment = (
        getTypedActivity(FileDisplayActivity::class.java)
            ?.leftFragment as? OCFileListFragment
    )
    ocFileListFragment?.setFabVisible(!hasFocus)

    if (searchView?.query.isNullOrEmpty()) {
        closeButton?.visibility = View.INVISIBLE
    }
}
```

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
ZetaTom: Changes implemented with 1459eb1.

app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
ZetaTom: Changes implemented with 1459eb1. I've tweaked some of your suggestions. Please check this part again.

app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
```diff
@@ -121,9 +117,9 @@ open class ExtendedListFragment :
     private var mEmptyListIcon: ImageView? = null
 
     // Save the state of the scroll in browsing
-    private var mIndexes: ArrayList<Int?>? = ArrayList<Int?>()
-    private var mFirstPositions: ArrayList<Int?>? = ArrayList<Int?>()
-    private var mTops: ArrayList<Int?>? = ArrayList<Int?>()
+    private var mIndexes: ArrayList<Int?> = ArrayList()
```
alperozturk96: Why they are still `ArrayList<Int?>`?

We can replace with `arrayListOf<Int>()`

app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
```diff
@@ -248,22 +241,25 @@ open class ExtendedListFragment :
...
+            performSearch(query, ArrayList(), false)
             return true
         }
         return false
     }
 
     fun performSearch(query: String, listOfHiddenFiles: ArrayList<String?>?, isBackPressed: Boolean) {
         val adapter = recyclerView?.adapter
-        val activity: Activity? = getActivity()
+        val activity = activity
```
alperozturk96: Elvis operator can be used instead of if null check.

`val activity = activity ?: return`

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
ZetaTom: My apologies, I overlooked that part. I've just pushed the new commit (b6e8431).

app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
ZetaTom: Suggestion implemented in b6e8431.

PR #15336, opened 2025-08-11
Refactor utils

alperozturk96:

### DIFF
--- a/app/src/main/java/com/owncloud/android/utils/EncryptionUtilsV2.kt
+++ b/app/src/main/java/com/owncloud/android/utils/EncryptionUtilsV2.kt
@@ -986,22 +986,17 @@ class EncryptionUtilsV2 {
 
     @Suppress("TooGenericExceptionCaught")
     fun verifySignedData(data: CMSSignedData, certs: List<X509Certificate>): Boolean {
-        val signer: SignerInformation = data.signerInfos.signers.iterator().next() as SignerInformation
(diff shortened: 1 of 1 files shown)


tobiasKaminsky (changes requested):
see inline

app/src/main/java/com/owncloud/android/utils/EncryptionUtilsV2.kt
```diff
@@ -982,20 +984,24 @@ class EncryptionUtilsV2 {
...
+                    verifiedCertCount += 1
                 }
-            } catch (e: java.lang.Exception) {
+            } catch (e: Exception) {
                 Log_OC.e(TAG, "Error caught at verifySignedData: $e")
             }
         }
 
-        return false
+        return verifiedCertCount == certs.count()
```
tobiasKaminsky: I think this is wrong.
In certs all users' certs that have access are stored.
But the current user is only one out of them.
So the test is, that the current user needs to be verified by one of the certs, not all of them.

tobiasKaminsky (changes requested):

app/src/main/java/com/owncloud/android/utils/EncryptionUtilsV2.kt
alperozturk96: https://github.com/nextcloud/android/pull/15336/commits/df57c7e3feaaa838822622355918e8152e53fbd0

PR #15388, opened 2025-08-18
fix: left to right filename sanitization

alperozturk96:

### Issue

The purpose of this PR is not to enhance the UI, but to ensure that the correct filenames are displayed. Certain invisible or special Unicode characters can be used to manipulate a filename’s visual representation, misleading users about its true file extension.

e.g. filename: invoiceexe.docRLO this will appear as invoice.doc, but actual filename is invoicedoc.exe. It is not a doc file, it's an EXE file.

Previously, the filename with extension in one text view component was shown; now, we also display the file extension as a separate text view component next to the filename, which is why the UI is changed.

The reason for using two separate text view elements is explained in detail in the mentioned issue.


kra-mo (approved):
Looks good!

Out of scope for this PR, but now that these are separate elements, it might be nice to dim the extensions slightly, same as it is done on the web:

Image

alperozturk96:
@AndyScherzinger Which file is causing this?


Screenshot 2025-08-19 at 16 43 11

alperozturk96:
> Looks good!
> 
> Out of scope for this PR, but now that these are separate elements, it might be nice to dim the extensions slightly, same as it is done on the web:
> 
> Image

Can you create an enhancement issue for that? Or I can do it as well.

AndyScherzinger:
@alperozturk96 the file mentioned in line 6 of your screenshot

alperozturk96:
@tobiasKaminsky Please add license in different PR.


Screenshot 2025-08-20 at 07 48 57

AndyScherzinger (approved):

alperozturk96:
The remaining issues:

1. In some cases, the “More” button appears to be slightly shifted to the right.

2. Very long filenames may occupy space without showing the extension (e.g., This PR: aaaaaaaaaa..... Master: aaa.....jpg). This occurs because the master currently uses a single text view element and ellipsis in the middle.

<table>
  <tr>
    <td>Image 1</td>
  </tr>
</table>

Achieving perfectly consistent alignment across all three elements in every scenario does not appear feasible. Using **two text view elements makes it impossible to fully match with the master.**

Considering that **this change will only be applied to bidirectional filenames**, I believe the current implementation is acceptable. 

Demo:


https://github.com/user-attachments/assets/38090711-596d-4a37-8e25-841548e45728




AndyScherzinger:
Fine by me 👍

PR #15392, opened 2025-08-19
fix: share updating

alperozturk96:

### Issue

When creating a new folder and attempting to share it via a link, the "Set password" and "Link label" fields do not save changes the first time. The changes are only saved on subsequent attempts, and occasionally even multiple attempts do not work until a full refresh of "All Files" is performed.

### Steps to Reproduce:

1. Tap the "+" button and create a new folder.
2. Tap the share icon for the newly created folder.
3. Tap "Create link".
4. Tap on link details and navigate to Settings.
5. Enter a value in Set password and Link label fields.
6. Tap the confirm button.
7. Observe that the changes are not saved on the first attempt.

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/UpdateShareInfoOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/UpdateShareInfoOperation.java
@@ -16,6 +16,7 @@ import com.owncloud.android.datamodel.FileDataStorageManager;
 import com.owncloud.android.lib.common.OwnCloudClient;
 import com.owncloud.android.lib.common.operations.RemoteOperation;
 import com.owncloud.android.lib.common.operations.RemoteOperationResult;
+import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.lib.resources.shares.GetShareRemoteOperation;
 import com.owncloud.android.lib.resources.shares.OCShare;
 import com.owncloud.android.lib.resources.shares.UpdateShareRemoteOperation;
@@ -29,6 +30,7 @@ public class UpdateShareInfoOperation extends SyncOperation {
 
     private OCShare share;
     private long shareId;
+    private long shareRemoteId;
     private long expirationDateInMillis;
     private String note;
     private boolean hideFileDownload;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/owncloud/android/services/OperationsService.java
```diff
@@ -626,9 +627,11 @@ private Pair<Target, RemoteOperation> newOperation(Intent operationIntent) {
 
                     case ACTION_UPDATE_SHARE_INFO:
                         shareId = operationIntent.getLongExtra(EXTRA_SHARE_ID, -1);
+                        long shareRemoteId = operationIntent.getLongExtra(EXTRA_SHARE_REMOTE_ID, -1);
 
                         if (shareId > 0) {
```
ZetaTom:
```suggestion
                        if (shareId > 0 || shareRemoteId > 0) {
```

If I understand correctly, supplying `shareRemoteId` should be enough to resolve the correct share given the changes in `UpdateShareInfoOperation`.

app/src/main/java/com/owncloud/android/operations/UpdateShareInfoOperation.java
```diff
@@ -107,9 +121,10 @@ protected RemoteOperationResult run(OwnCloudClient client) {
             if (result.isSuccess() && shareId > 0) {
                 OCShare ocShare = (OCShare) result.getData().get(0);
                 ocShare.setPasswordProtected(!TextUtils.isEmpty(password));
+                ocShare.setRemoteId(shareRemoteId);
```
ZetaTom, line 124: The remote id should already be set for any shares fetched from the server, if I'm not mistaken.

alperozturk96: When we don't set the remoteId updating share is failing.

app/src/main/java/com/owncloud/android/services/OperationsService.java
alperozturk96: https://github.com/nextcloud/android/pull/15392/commits/b29a9072b4d9adce7f89825c34c07effd0130873

tobiasKaminsky (approved):

Issue #15237, opened 2025-07-23
Enhance recommended files


tobiasKaminsky:
- cache recommended files, so that closing/opening app does not need to trigger new "search"
- debug why no icons/thumbnails are shown on c.nc.c
- use same grid menu as existing grid view, otherwise this looks a bit off
- check if it is possible to have horizontal view without margin -> edge to edge

alperozturk96:
To be able to implement proper caching mechanism for recommended files [this](https://github.com/nextcloud/recommendations/issues/895) needs to be implemented.

PR #15402, opened 2025-08-20
feat: enhance recommended file list

alperozturk96:

### Issue https://github.com/nextcloud/android/issues/15237

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>before</td>
    <td>after</td>
  </tr>
</table>

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>before</td>
    <td>after</td>
  </tr>
</table>

### Changes

The existing grid layout is being used.
Now, we also have a multi-select option.
The placement of the “More” button has been changed.

### DIFF
--- a/app/src/main/res/layout/list_header.xml
+++ b/app/src/main/res/layout/list_header.xml
@@ -7,19 +7,19 @@
   ~ SPDX-FileCopyrightText: 2019 Nextcloud GmbH
   ~ SPDX-License-Identifier: AGPL-3.0-or-later OR GPL-2.0-only
 -->
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<LinearLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     android:id="@+id/headerView"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
     android:orientation="vertical"
-    android:paddingStart="@dimen/standard_padding"
     android:paddingTop="@dimen/standard_padding"
-    android:paddingEnd="@dimen/standard_padding"
     android:showDividers="none">
 
     <TextView
         android:id="@+id/headerText"
+        android:layout_marginHorizontal="@dimen/standard_margin"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
         android:layout_marginBottom="@dimen/standard_margin"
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
> Maybe just put it in the ⋮ menu? No need for a string change either imho, just putting it there grayed-out is probably fine.

I'd argue to keep the secondary line with the recommended status cause on Android I would expect menu items to be actionable.

alperozturk96:
@tobiasKaminsky @kra-mo  @AndyScherzinger 

Currently, the implementation uses `OCFileListGridItemViewHolder`, which extends from `ListViewHolder` and is designed for file listing. The interface implementation includes features such as multi-select, file item click handling, and more actions. However, it does not include an additional `TextView` element for displaying a reason. Adding this feature would therefore require significant effort.

We have several options to consider:

1. **Add the reason text programmatically** – This approach would require handling alignment, spacing, and other layout details programmatically within `OCFileListGridItemViewHolder`. This approach is fragile, and I don't recommend due to the potential layout issues. I have to work on it too much to have good-looking layout.

2. **Extend the current `OCFileListGridItemViewHolder` to include a reason `TextView`** – This would require modifying the file listing UI, setting the visibility of the reason text to `GONE` for standard files, and `VISIBLE` for recommended files. This is also not ideal, as it introduces changes to the core file listing logic for a specific use case.

3. **Create a new view holder for recommended files** – We could create a new view holder that extends `ListViewHolder` and define a separate layout for the recommended file list. This is the most robust and maintainable solution, although it requires extra development effort but not that much.

**Note:** More button opens the file actions bottom sheet and that bottom sheet designed for the normal file, thus adding there reason text is not ideal and requires more effort.

I think we should leave it as it is since the reason text is always (for now) "Recently edited" or we can choose option 3.

kra-mo:
> since the reason text is always (for now) "Recently edited"

That is fair enough for now, but I'd add a comment for the future then :)

kra-mo (approved):

AndyScherzinger:
> I think we should leave it as it is since the reason text is always (for now) "Recently edited" or we can choose option 3.

Agreed, so option 3 but for the future

alperozturk96:
> > I think we should leave it as it is since the reason text is always (for now) "Recently edited" or we can choose option 3.
> 
> Agreed, so option 3 but for the future

https://github.com/nextcloud/android/issues/15452

PR #15407, opened 2025-08-21
folders/favorites first as option

tobiasKaminsky:

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
@@ -183,6 +183,9 @@ public class SettingsActivity extends PreferenceActivity
         // Synced folders
         setupAutoUploadCategory(preferenceScreen);
 
+        // Files
+        setupFilesCategory(preferenceScreen);
+
         // Details
         setupDetailsCategory(preferenceScreen);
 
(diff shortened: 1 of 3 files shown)


alperozturk96 (changes requested):
I have set both of the new flags to false, so that files appear first and favorites are not prioritized. When combined with A-Z sorting, the expected behavior is that files appear first, with any favorite files appearing last, followed by folders, right? 

Additionally, other sorting options (new to old, old to new, Z-A) should not break the sorting so we have to test those as well.

**As a side note**, I would prefer to use an enum to represent state. However, this does not seem feasible here because the two new flags are independent and can be combined, rather than representing a single shared state.

Image

app/src/main/java/com/owncloud/android/utils/FileSortOrder.kt
```diff
@@ -89,9 +89,45 @@ open class FileSortOrder(@JvmField var name: String, var isAscending: Boolean) {
...
+                    o2.isFolder -> 1
+                    else -> 0
+                }
+            }
+            return files
+        }
     }
 
-    open fun sortCloudFiles(files: MutableList<OCFile>): List<OCFile> = sortCloudFilesByFavourite(files)
+    open fun sortCloudFiles(files: MutableList<OCFile>,  foldersBeforeFiles: Boolean, favoritesFirst: Boolean): MutableList<OCFile> {
```
alperozturk96: Please remove unnecessary if statements.

```
open fun sortCloudFiles(files: MutableList<OCFile>,  foldersBeforeFiles: Boolean, favoritesFirst: Boolean): MutableList<OCFile> {
        if (foldersBeforeFiles && favoritesFirst) {
            return sortCloudFilesByFavourite(sortCloudFilesByFolderFirst(files))
        }
        
        if (foldersBeforeFiles) {
            return sortCloudFilesByFolderFirst(files)
        }
        
        if (favoritesFirst) {
            return sortCloudFilesByFavourite(files)   
        }

        return files
    }
```

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/utils/FileSortOrder.kt
tobiasKaminsky: Which ones? I might be blind today :sweat_smile:

alperozturk96: Just use the code I gave above, :D Yours has four, mine has three.

alperozturk96 (changes requested):

PR #15413, opened 2025-08-22
fix: empty user input checks

alperozturk96:

### Issue

Currently, we are using `isEmpty()` to validate user input. However, this check does not prevent inputs that consist only of whitespace. For example, a user could try to create a folder with the name " " (spaces), which passes the `isEmpty()` check but is clearly invalid.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/fragment/FileDetailsSharingProcessFragment.kt
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/FileDetailsSharingProcessFragment.kt
@@ -768,21 +768,21 @@ class FileDetailsSharingProcessFragment :
         }
 
         if (binding.shareProcessSetPasswordSwitch.isChecked &&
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/owncloud/android/ui/fragment/FileDetailsSharingProcessFragment.kt
```diff
@@ -768,21 +768,21 @@ class FileDetailsSharingProcessFragment :
...
-            binding.shareProcessSelectExpDate.text?.trim().isNullOrEmpty()
+            binding.shareProcessSelectExpDate.text?.isBlank() == true
         ) {
             showExpirationDateDialog()
             return
         }
 
         if (binding.shareProcessChangeNameSwitch.isChecked &&
-            binding.shareProcessChangeName.text?.trim().isNullOrEmpty()
+            binding.shareProcessChangeName.text?.isBlank() == true
```
alperozturk96, line 785: `isBlank()` already checks correctly no need to trim.

PR #15415, opened 2025-08-22
SpotBugs: Fix Method ignores exceptional return value

alexknop:
There are 16 SpotBugs under 

Bad practice:
   Method ignores exceptional return value   (16: 0/16/0/0) 

Solution:

### DIFF
--- a/app/src/gplay/java/com/owncloud/android/utils/PushUtils.java
+++ b/app/src/gplay/java/com/owncloud/android/utils/PushUtils.java
@@ -96,7 +96,11 @@ public final class PushUtils {
         if (!new File(privateKeyPath).exists() && !new File(publicKeyPath).exists()) {
             try {
                 if (!keyPathFile.exists()) {
-                    keyPathFile.mkdir();
+                    try {
+                        Files.createDirectories(keyPathFile.toPath());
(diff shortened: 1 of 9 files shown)


alperozturk96:
Hello

We really appreciate the time and effort you’ve put into improving the project. 🙏

When multiple files and different logics are changed together, it becomes difficult for us to test and review effectively, since each area can have its own impact (e.g: https://github.com/nextcloud/android/issues/15214#issuecomment-3101884063).

To make review and testing smoother, it would be best if the changes are split into smaller PRs, ideally one PR per logical change in each file.

**For example:**

A PR focused solely on the move functionality in the `private void move(File sourceFile, File targetFile)` method of `UploadFileOperations.java.`

A separate PR for changes to `UploadFileOperations.java` related to `LOCAL_BEHAVIOUR_DELETE` within the `handleLocalBehaviour` method.

A separate PR for modifications to `Executors.newCachedThreadPool()` in `OCFileListFragment.java.`

This way, we can test and approve each functionality independently, ensuring quality while avoiding unexpected side effects.

Thanks again for your work  🙌

alexknop:
@alperozturk96 Good idea, closing this to focus on each one individually.

PR #15424, opened 2025-08-24
Automated testing minor improvements

PhilLab:
- Minor improvements of testing documentation
- Instrumentation tests now automatically grab the notifications permission

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -217,7 +217,6 @@ Source code of app:
 #### Instrumented tests
 - tests to see larger code working in correct way
 - tests that require parts of Android SDK
-- best to avoid server communication, see https://github.com/nextcloud/android/pull/3624
 
 - run all tests ```./gradlew createGplayDebugCoverageReport -Pcoverage=true```
 - run selective test class: ```./gradlew createGplayDebugCoverageReport -Pcoverage=true
@@ -229,6 +228,12 @@ Source code of app:
   -Pandroid.testInstrumentationRunnerArguments.class=com.owncloud.android.datamodel.FileDataStorageManagerContentProviderClientIT#saveFile```
 - JaCoCo results are shown as html: firefox ./build/reports/coverage/gplay/debug/index.html
 
+#### Instrumented tests with server communication
+It is best to avoid server communication, see https://github.com/nextcloud/android/pull/3624. But if a test requires
+a server, this is how it is done:
+- Start the server
+  - TODO: find out how... is it maybe this command: `docker run --name=uiComparison nextcloudci/server --entrypoint '/usr/local/bin/initnc.sh' 1>/dev/null`
+- your test class should inherit from `AbstractOnServerIT`, e.g.: `public class DownloadIT extends AbstractOnServerIT { ...`
 
 #### UI tests
 We use [shot](https://github.com/Karumi/Shot) for taking screenshots and compare them 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -221,12 +221,12 @@ Source code of app:
 
 - run all tests ```./gradlew createGplayDebugCoverageReport -Pcoverage=true```
 - run selective test class: ```./gradlew createGplayDebugCoverageReport -Pcoverage=true
-  -Pandroid.testInstrumentationRunnerArguments.class=com.owncloud.android.datamodel.FileDataStorageManagerTest```
+  -Pandroid.testInstrumentationRunnerArguments.class=com.owncloud.android.datamodel.FileDataStorageManagerContentProviderClientIT```
 - run multiple test classes:
   -   separate by ","
-  - ```./gradlew createGplayDebugCoverageReport -Pcoverage=true -Pandroid.testInstrumentationRunnerArguments.class=com.owncloud.android.datamodel.FileDataStorageManagerTest,com.nextcloud.client.FileDisplayActivityIT```
+  - ```./gradlew createGplayDebugCoverageReport -Pcoverage=true -Pandroid.testInstrumentationRunnerArguments.class=com.owncloud.android.datamodel.FileDataStorageManagerContentProviderClientIT,com.nextcloud.client.FileDisplayActivityIT```
 - run one test in class: ```./gradlew createGplayDebugCoverageReport -Pcoverage=true
-  -Pandroid.testInstrumentationRunnerArguments.class=com.owncloud.android.datamodel.FileDataStorageManagerTest#saveNewFile```
+  -Pandroid.testInstrumentationRunnerArguments.class=com.owncloud.android.datamodel.FileDataStorageManagerContentProviderClientIT#saveFile```
 - JaCoCo results are shown as html: firefox ./build/reports/coverage/gplay/debug/index.html
 
 
--- a/app/src/androidTest/java/com/owncloud/android/AbstractOnServerIT.java
+++ b/app/src/androidTest/java/com/owncloud/android/AbstractOnServerIT.java
@@ -53,7 +53,7 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
 /**
- * Common base for all integration tests.
+ * Common base for all integration tests requiring a server connection
  */
 public abstract class AbstractOnServerIT extends AbstractIT {
     @BeforeClass
--- a/app/src/androidTest/java/com/owncloud/android/DownloadIT.java
+++ b/app/src/androidTest/java/com/owncloud/android/DownloadIT.java
@@ -29,7 +29,7 @@ import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertTrue;
 
 /**
- * Tests related to file uploads.
+ * Tests related to file downloads.
  */
 public class DownloadIT extends AbstractOnServerIT {
     private static final String FOLDER = "/testUpload/";
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ If you want to [contribute](https://nextcloud.com/contribute/) to the Nextcloud 
 *   reporting problems / suggesting enhancements by [opening new issues](https://github.com/nextcloud/android/issues/new/choose)
 *   implementing proposed bug fixes and enhancement ideas by submitting PRs (associated with a corresponding issue preferably)
 *   reviewing [pull requests](https://github.com/nextcloud/android/pulls) and providing feedback on code, implementation, and functionality
+*   Add [automated tests](CONTRIBUTING.md#testing) for existing functionality
 *   installing and testing [pull request builds](https://github.com/nextcloud/android/pulls), [daily/dev builds](https://github.com/nextcloud/android#development-version-hammer), or [RCs/release candidate builds](https://github.com/nextcloud/android/releases) 
 *   enhancing Admin, User, or Developer [documentation](https://github.com/nextcloud/documentation/)
 *   hitting hard on the latest stable release by testing fundamental features and evaluating the user experience
--- a/app/src/androidTest/java/com/nextcloud/test/GrantStoragePermissionRule.kt
+++ b/app/src/androidTest/java/com/nextcloud/test/GrantStoragePermissionRule.kt
@@ -15,6 +15,10 @@ import org.junit.rules.TestRule
 import org.junit.runner.Description
 import org.junit.runners.model.Statement
 
+/**
+ * Rule to automatically enable the test to write to the external storage.
+ * Depending on the SDK version, different approaches might be necessary to achieve the full access.
+ */
 class GrantStoragePermissionRule private constructor() {
 
     companion object {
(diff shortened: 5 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -217,18 +217,23 @@ Source code of app:
...
 - run one test in class: ```./gradlew createGplayDebugCoverageReport -Pcoverage=true
-  -Pandroid.testInstrumentationRunnerArguments.class=com.owncloud.android.datamodel.FileDataStorageManagerTest#saveNewFile```
+  -Pandroid.testInstrumentationRunnerArguments.class=com.owncloud.android.datamodel.FileDataStorageManagerContentProviderClientIT#saveFile```
 - JaCoCo results are shown as html: firefox ./build/reports/coverage/gplay/debug/index.html
 
+#### Instrumented tests with server communication
+It is best to avoid server communication, see https://github.com/nextcloud/android/pull/3624. But if a test requires
+a server, this is how it is done:
+- Start the server
+  - TODO: find out how... is it maybe this command: `docker run --name=uiComparison nextcloudci/server --entrypoint '/usr/local/bin/initnc.sh' 1>/dev/null`
```
PhilLab: @alperozturk96 If somebody could point me to the right direction how to develop and test locally any integration test which needs a reachable backend, I would like to explain it here. And how to get new test data, e.g. new expected remote files onto it.

Secondly, if there is an existing method of using a mocked server instead, please let me know and I can write it up here as well.

tobiasKaminsky: You can run any server you like. 
Best is to look into https://docs.nextcloud.com/server/latest/developer_manual/getting_started/devenv.html

There is a docker instruction, but also the "plain" php way described.
You can also use your real server, with a separate(!) user for testing.



> Secondly, if there is an existing method of using a mocked server instead, please let me know and I can write it up here as well.

There are some tests, which do not require a server. eg. FileDetailFragmentStaticServerIT or tests using TestActivity

PhilLab: @tobiasKaminsky thank you for your response. What is now missing for me is how I can set up such a server-based test, test it locally, and then have it executed as part of the regular Github CI.

For it to work on the CI, it would require that I run a server in a similar configuration locally (or use a separate test user on a real server, as you said) - but for both I need to know how to replicate on my own what will later be executed in the CI. That's the essence of my question.

PhilLab: @tobiasKaminsky how can I set up such an end-to-end integration test which uses a specific state on the server, e.g. a specific list of files to download, etc...? Is there an existing test doing such a thing, e.g. by first uploading some files in an `@Before` test method? My aim is to educate (and be educated) how to extend the test suite.

If this is nothing which exists already, I'd remove this readme section from the PR and have the rest merged, as it would already be a small improvement.

PhilLab: @tobiasKaminsky okay, I sorted it out and added explanations to the markdown file as well as the abstract test class.

Can you kindly let me know how and when these `AbstractOnServerIT` run in the CI? What I _think_ right now:
1. AbstractOnServerIT tests are run as part of runCombinedTest.sh
2. That script is called in .drone.yml
3. Drone CI executes them **after the merge**. You can access the results in https://drone.nextcloud.com
4. The status of master is continuously monitored and if tests start to fail, a new PR is created to fix the code (or the tests)

However, where I need help: 
1. when I tried to find runs of a specific test `DownloadIT` to verify my assumptions, I failed to find them.
2. I was wondering if there is an existing mechanism to run these tests _before_ broken code is merged?
3. How do I find the Drone run for a specific merge? I clicked through the activity feed to find my PR, which was tedious, especially if the PR is older
4. Verifying my PR `#15669` I found https://drone.nextcloud.com/nextcloud/android/25363 , but no tests ran there. Why? What are the controls there?

If you are wondering, I am trying to come up with automated tests for PR ``#15669`` and document what I am learning in the process

CONTRIBUTING.md
PhilLab: @tobiasKaminsky regarding the formatting I also [needed help](https://github.com/nextcloud/android/pull/15669#issuecomment-3414492465) on my previous PR. Shall I add the gradlew commands here? I notice that I keep looking for them again and again.

```
### Formatter setup
Our formatter setup is rather simple:
* Standard Android Studio
* Line length 120 characters (Settings->Editor->Code Style->Right margin(columns): 120; also set by EditorConfig
* Auto optimize imports (Settings->Editor->Auto Import->Optimize imports on the fly)

You can fix Check / check (spotlessKotlinCheck) via following commands:

    ./gradlew spotlessApply
    ./gradlew detekt                                
    ./gradlew spotlessCheck
    ./gradlew spotlessKotlinCheck
``

File: CONTRIBUTING.md (hunk 1 above)
Comment (PhilLab) [reply CONTRIBUTING.md]: @tobiasKaminsky @alperozturk96 can you take a look? This shouldn't be much effort. If anything, we could merge this already as it is, I think it would be a no-brainer improvement and help other developers

Comment (alperozturk96) [reply CONTRIBUTING.md]: @PhilLab 

Hello

I can share what I know about the tests.

`AbstractOnServerIT` reads the credentials from `gradle.properties` (`NC_TEST_SERVER_BASEURL`, `NC_TEST_SERVER_USERNAME`, `NC_TEST_SERVER_PASSWORD`) and uses them to create the networking clients. Any test class extending it, for example, `DownloadIT`, will rely on those same credentials.

You can point the tests either to a Nextcloud instance you already have or to a local instance created via Docker (e.g., 10.0.2.2, admin, admin), and then run the tests accordingly.

Regarding the failed test runs, scripts, and YAML files, @tobiasKaminsky can provide a better explanation.

Sorry for the delayed response.

File: CONTRIBUTING.md
Comment (alperozturk96) [reply CONTRIBUTING.md]: @PhilLab Hey.

That would be awesome. Thank you. 

"./gradlew jacocoTestGplayDebugUnitTest" → this run unit tests locally

"./gradlew spotlessApply" → this formats the kotlin code.

"./gradlew lint" → this checks lint

@tobiasKaminsky Maybe you have more?

File: app/src/androidTest/java/com/owncloud/android/DownloadIT.java [hunk 2]
```diff
@@ -98,10 +98,10 @@ private void verifyDownload(OCFile file1, OCFile file2) {
         assertTrue(new File(file2.getStoragePath()).exists());
 
         // test against hardcoded path to make sure that it is correct
-        assertEquals("/storage/emulated/0/Android/media/com.nextcloud.client/nextcloud/" +
+        assertEquals("/storage/emulated/0/Android/media/"+targetContext.getPackageName()+"/nextcloud/" +
                          Uri.encode(account.name, "@") + "/testUpload/nonEmpty.txt",
                      file1.getStoragePath());
-        assertEquals("/storage/emulated/0/Android/media/com.nextcloud.client/nextcloud/" +
+        assertEquals("/storage/emulated/0/Android/media/"+targetContext.getPackageName()+"/nextcloud/" +
```
tobiasKaminsky, line 104: Is there a reason to change that? 
We test "only" our official Nextcloud client.

PhilLab: @tobiasKaminsky Two reasons:
1. The issue is that when you switch between test development and app debugging (to fix the found issues), the _tests_ always deletes the data and when starting the _app_ you need to login again. This costs me 2-3 minutes every time and is quite annoying.
A good workaround for this is to start the tests under one build variant, and debug the app under another build variant. But for this to work, it needs to be respected everywhere, otherwise assertions will fail or even the wrong data be deleted. This is the common background of the two commits `Fixed AbstractIT deleting the wrong account` and `DownloadIT now works for all build variants`
2. Also conceptually it is clearer that instead of the magic string `com.nextcloud.client` it refers to what is semantically needed here - and that is the packageName. Sure, following that paradigm, the assertion should make use of even more functions (e.g. to retrieve the `/media/` path programmatically instead of hard-coded). But I didn't have an actual use case to justify that. Only the use case from `#1`.

CONTRIBUTING.md
PhilLab: @alperozturk96 thanks. I amended the commit, please take a look.

PhilLab: @alperozturk96 okay for you?

CONTRIBUTING.md
PhilLab: Doesn't need to be overly detailed - just some pointers how any developer can make use of the Github CI results. Isn't that part of your daily routine as well, @alperozturk96 ? So both you and @tobiasKaminsky should be able to provide the information?

tobiasKaminsky:
@PhilLab the CI does not work on forked repos.
May I add you to our Nextcloud org and then you create the PR again within our repo?

app/src/androidTest/java/com/owncloud/android/AbstractOnServerIT.java
```diff
@@ -54,7 +55,16 @@
...
- * Common base for all integration tests.
+ * Common base for all integration tests requiring a server connection.
+ * ATTENTION: Deletes ALL files of the test user on the server after each test run.
+ * So you MUST use a dedicated test user.
+ *
+ * Uses server, user and password given as `testInstrumentationRunnerArgument`
+ * - TEST_SERVER_URL
+ * - TEST_SERVER_USERNAME
+ * - TEST_SERVER_PASSWORD
+ * These are supplied via build.gradle, which takes them from gradle.properties.
```
PhilLab: @alperozturk96 @tobiasKaminsky I just noticed that the test setup is broken on master. The values in gradle.properties are ignored now, even though there are still default values to be filled in there:
```
NC_TEST_SERVER_BASEURL=http://server
NC_TEST_SERVER_USERNAME=test
NC_TEST_SERVER_PASSWORD=test
```
I just tried to run tests but couldn't reach my server until I noticed that it is not using my custom URL and credentials.

I am not fully clear how this happened, but this is what I found so far:
- 247d08595de75add7f7e301ad44a30b1f54572dd introduced an alternative parsing of the same values, but from .gradle/config.properties.
- later, 5fd2e2942c2b375b86c52256f7c8921395f5594c came and changed that `testInstrumentationRunnerArgument` wasn't taking these values from the gradle.properties but instead from the otherwise unused config.properties.

This is not the first weird side-effect of that Gradle rewrite PR 🤔 #15859 .

It is very confusing. We need to clean up two things
1. We should only have one way of passing these values to the code. Either as test argument or as BuildConfig.
    - To be honest, storing the password in the BuildConfig, even if it is only the debug config, seems very risky to me. This can easily be overlooked and cause unintended side-effects. E.g. testing .apks you are sharing with others will include these credentials.
    - Hence, I would suggest to remove them from the BuildConfig again
2. We should only parse these values from one file. `.gradle/config.properties` or `gradle.properties`
    - I would suggest `gradle.properties`. This is how it always has been and it seems to be common practice.

We can also fix this in a new PR, as it is more urgent and independent of the documentation update here.

app/src/androidTest/java/com/owncloud/android/DownloadIT.java
PhilLab: @tobiasKaminsky with that explanation - is this fine for you?

app/src/androidTest/java/com/owncloud/android/AbstractOnServerIT.java
PhilLab: In my local stash, I have replaced this line https://github.com/nextcloud/android/blob/master/app/build.gradle.kts#L78 with:
 
```
val file = rootProject.file("gradle.properties")
```

shall I add this to the PR here, or will you open a new one with more consequential changes?

PhilLab:
> @PhilLab the CI does not work on forked repos. May I add you to our Nextcloud org and then you create the PR again within our repo?

@tobiasKaminsky regarding this, I sent an email to the address mentioned in the Github profile (not the one from your commits).

CONTRIBUTING.md
PhilLab: Seems to be a more complicated topic than I thought. Let's not drag this PR any longer. If at some point in the future I learn how the test execution is supposed to work, I can document it in a new PR.

Closing the discussion.

app/src/androidTest/java/com/owncloud/android/AbstractOnServerIT.java
PhilLab: @tobiasKaminsky @alperozturk96 I added another commit to change this to
```
val file = rootProject.file("gradle.properties")
```
because my local testing was blocked by this. Can we merge? But I do think that it needs a clean up, as eluded to in my message above.

PhilLab:
@tobiasKaminsky @alperozturk96 happy new year and all the best for 2026! Could this PR be merged? I find myself dragging these changes accross my different local branches, so I can work on automated tests. It would help to merge this soon. Thanks 🙂

PhilLab:
@tobiasKaminsky given the existing, longer discussion here I guess it is less advisable to duplicate this one into a new PR in the main repository? Or shall I recreate nonetheless?

alperozturk96 (approved):
Thank you for great PR. Looks good to me.

app/src/androidTest/java/com/owncloud/android/DownloadIT.java
alperozturk96: @tobiasKaminsky Could you please look at it, and if it's okay, give approval?

tobiasKaminsky: Fine ✔

tobiasKaminsky (approved):

tobiasKaminsky:
Awesome help @PhilLab 

To have it run one time on CI, I pushed it to our repo: https://github.com/nextcloud/android/pull/16286

PR #15440, opened 2025-08-26
feat: metadata worker 🚀 

alperozturk96:

### Issue

https://github.com/nextcloud/files-clients/issues/81


app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -2985,6 +2987,17 @@ class FileDisplayActivity :
         })
     }
 
+    // region MetadataSyncJob
+    private fun startMetadataSyncForRoot() {
+        backgroundJobManager.startMetadataSyncJob(OCFile.ROOT_ID)
```
tobiasKaminsky: This will not work with multiple accounts.
Only first account will have id 1.

I think it is better to use storagemanager.getFileByDecryptedRemotePath("/")

app/src/main/java/com/nextcloud/client/jobs/metadata/MetadataWorker.kt
```diff
@@ -0,0 +1,57 @@
...
+    override suspend fun doWork(): Result {
+        val storageManager = FileDataStorageManager(user, context.contentResolver)
+        val id = inputData.getLong(FILE_ID, -1L)
+        if (id == -1L) {
+            Log_OC.e(TAG, "❌ Invalid folder ID. Aborting metadata sync.")
+            return Result.failure()
+        }
+        Log_OC.d(TAG, "🕒 Starting metadata sync for folder ID: $id")
+
+        val subfolders = storageManager.getSubFiles(id).filter { !it.isEncrypted && it.isFolder }
```
tobiasKaminsky: Since you created "getSubFiles" you could aready filter for folder inside DB?

alperozturk96: Do you mean directly from the dao?

app/src/main/java/com/nextcloud/client/jobs/metadata/MetadataWorker.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    @Suppress("DEPRECATION", "ReturnCount")
+    override suspend fun doWork(): Result {
+        val storageManager = FileDataStorageManager(user, context.contentResolver)
+        val id = inputData.getLong(FILE_ID, -1L)
+        if (id == -1L) {
+            Log_OC.e(TAG, "❌ Invalid folder ID. Aborting metadata sync.")
+            return Result.failure()
+        }
+        Log_OC.d(TAG, "🕒 Starting metadata sync for folder ID: $id")
```
tobiasKaminsky: I think there is one step missing.
You do not check the current folder, e.g. if we want to check "/test/", then we
- check for Etag change in "/test"
- if etag unchanged -> directly done
- if etag has changed -> sync, this way we will also receive new subfoder / old subfolders are deleted

This can all be done RefreshFolderOperation.

app/src/main/java/com/nextcloud/client/jobs/metadata/MetadataWorker.kt
```diff
@@ -0,0 +1,57 @@
...
+        val subfolders = storageManager.getSubFiles(id).filter { !it.isEncrypted && it.isFolder }
+        if (subfolders.isEmpty()) {
+            Log_OC.d(TAG, "📂 No subfolders found for folder ID: $id. Nothing to sync.")
+            return Result.success()
+        }
+
+        subfolders.forEach { subFolder ->
+            Log_OC.d(TAG, "⏳ Fetching metadata for: ${subFolder.remotePath}")
+
+            val operation = RefreshFolderOperation(subFolder, storageManager, user, context)
```
tobiasKaminsky: This operation would then check etag on server.
But we know the current state, since we have synced the parent folder.
So it it is better to have in L42 a filter to run the loop only for subfolders with changed eTag.

app/src/main/java/com/nextcloud/client/jobs/metadata/MetadataWorker.kt
alperozturk96: https://github.com/nextcloud/android/pull/15440/commits/0f1a54a810e6094732585ba73681303520b2fecd

app/src/main/java/com/nextcloud/client/jobs/metadata/MetadataWorker.kt
alperozturk96: https://github.com/nextcloud/android/pull/15440/commits/0f1a54a810e6094732585ba73681303520b2fecd

app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
alperozturk96: https://github.com/nextcloud/android/pull/15440/commits/0f1a54a810e6094732585ba73681303520b2fecd

app/src/main/java/com/nextcloud/client/jobs/metadata/MetadataWorker.kt
alperozturk96: How do you know the current state? It can change at any moment. 

Let's say user have 4 folders in the root. Worker checks 5 folders including the root directory and user navigates to the sub-folder and user goes back to the root directory again thus we don't have to sync again (probably), but how can you check it in the worker? `RefreshFolderOperation` itself already checking it.

tobiasKaminsky: Let us have this folder structure
- / etag: 123
- /test1: etag 200
- /test2: etag 300

We then start the app
- RefreshFolderOperation runs with "/"
- it sees that etag has changed -> syncs metadata of subfolders
- changed etag: "/" -> 124; "/test2" -> 301
- then we run a check on all subfolders: ocFile.etag is "old/current state on mobile" and ocFile.eTagOnServer is new
--> we compare and only the one that has changed will then be used within loop

tobiasKaminsky:
```
before:
etag = "68b030cbb3aea"
etagInConflict = null
etagOnServer = "68b030cbb3aea"

after:
etag = "68b030cbb3aea"
etagInConflict = null
etagOnServer = "68b032f97cfac"
```

This is a real example after changing a file in a subfolder.

tobiasKaminsky: Btw: since this is a bit of "hidden" function, please let us add real tests for it.
I can help, if needed :)

alperozturk96: We should compare like this?

Image

Or?

Image

app/src/main/java/com/owncloud/android/ui/activity/FileActivity.java
```diff
@@ -312,10 +312,14 @@ protected void onSaveInstanceState(@NonNull Bundle outState) {
         outState.putParcelable(FileActivity.EXTRA_FILE, mFile);
         outState.putBoolean(FileActivity.EXTRA_FROM_NOTIFICATION, mFromNotification);
         outState.putLong(KEY_WAITING_FOR_OP_ID, mFileOperationsHelper.getOpIdWaitingFor());
-        if(getSupportActionBar() != null && getSupportActionBar().getTitle() != null) {
-            // Null check in case the actionbar is used in ActionBar.NAVIGATION_MODE_LIST
-            // since it doesn't have a title then
-            outState.putString(KEY_ACTION_BAR_TITLE, getSupportActionBar().getTitle().toString());
+
+        final var actionBar = getSupportActionBar();
```
alperozturk96, line 316: SpotBugs showed me unrelated issues, so I reduced it.

app/src/main/java/com/owncloud/android/utils/EncryptionUtils.java
```diff
@@ -526,7 +526,6 @@ public static String decodeBase64BytesToString(byte[] bytes) {
         }
     }
 
-    @SuppressFBWarnings({"DM", "MDM"})
```
alperozturk96, line 529: SpotBugs showed me unrelated issues, so I reduced it.

app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -383,19 +414,28 @@ private RemoteOperationResult checkForChanges(OwnCloudClient client) {
         result = new ReadFileRemoteOperation(remotePath).execute(client);
 
         if (result.isSuccess()) {
-            OCFile remoteFolder = FileStorageUtils.fillOCFile((RemoteFile) result.getData().get(0));
```
alperozturk96, line 386: No need to create OCFile.

tobiasKaminsky (approved):

PR #15442, opened 2025-08-26
fix: remove redundant code to prevent unnecessary power management popups

alperozturk96:

### Issue

This change simplifies the handling of battery optimization settings for the app. Previously, we included logic to handle `setPowerCheckDisabled` and `isPowerSavingExclusionAvailable`, which were only applicable for certain devices listed under `OVERLY_AGGRESSIVE_POWER_SAVING_VENDORS`.

I tested with`Xiaomi HyperOS 2`, all battery settings can be managed by redirecting to the app's battery optimization page using the `Settings.ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS` intent. As a result, the previous device-specific logic is no longer necessary and has been removed.

<table>
  <tr>
    <td>Image 1</td>
  </tr>
</table>

### DIFF
--- a/app/src/main/java/com/nextcloud/client/device/DeviceModule.kt
+++ b/app/src/main/java/com/nextcloud/client/device/DeviceModule.kt
@@ -10,7 +10,6 @@ package com.nextcloud.client.device
 
 import android.content.Context
 import android.os.PowerManager
-import com.nextcloud.client.preferences.AppPreferences
 import dagger.Module
 import dagger.Provides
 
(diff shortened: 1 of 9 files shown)


tobiasKaminsky:
Marcel: test with Samsung
Tobi: test with Huawei
Alper: tested already with Xiaomi

mahibi:
I tested on Samsung and it worked as expected.

Side note 1:
in 2022 i came to the conclusion that ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS was unreliable. See https://github.com/nextcloud/talk-android/issues/741#issuecomment-1143812542
However that might have changed..

Side note 2:
If anyone asks why it's not done for the talk app: Unfortunately for the Talk app we are not allowed to use ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS because it's a chat app, see https://developer.android.com/training/monitoring-device-state/doze-standby#exemption-cases

tobiasKaminsky:
Tested on Huawei ✔

Issue #15420, opened 2025-08-23
Bug - The offline operation cannot be completed. Unknown error


Rainbow-1000:
### ⚠️ Before posting ⚠️

- [x] This is a **bug**, not a question or an enhancement.
- [x] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [x] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

It automatic pop up. Cannot be controlled. 

### Expected behaviour

Do not have any offline folder. Don't expect this happen

### Actual behaviour

Should not appear. Only happens with this latest version of uupdate on 21/08/25

### Android version

15

### Device brand and model

Samsung S23 Ultra

### Stock or custom OS?

Stock

### Nextcloud android app version

3.32.3

### Nextcloud server version

31.0.8

### Using a reverse proxy?

I don't know

### Android logs

After updated to latest version this error message keep popping up. I do not have offli e folder or any files offline. Cannot find any option to turn this off

### Server error logs

```bash

```

Rainbow-1000:
![Image](https://github.com/user-attachments/assets/61e0a702-5b65-41dc-9b15-1867d00f28fb)

Anyone please check and response.

This is getting annoying as keep popping up notification and no way to stop it.

The app does not have option for offline files while it tried to do it then.

Please advise way to fix. Every time app has new update, always comes with new issue

alperozturk96:
@Rainbow-1000 

Hello

Thank you for the report. I will check it.

alperozturk96:
@Rainbow-1000 Hello

With [this](https://github.com/nextcloud/android/pull/15451#issuecomment-3227577363) update, you now have the option to cancel offline operation. Additionally, you can choose to disable offline operation notifications entirely through the notification channel settings.

Rainbow-1000:
Thanks for this . I installed. Could you please guide me through how to change that setting as I can't find in settings

alperozturk96:
@Rainbow-1000 

https://github.com/user-attachments/assets/f02d18b3-7d34-4379-a071-0be1b2d93b54

Rainbow-1000:
https://github.com/user-attachments/assets/b39da47e-7399-4ac6-aa96-434bd54622ef
Look like I don't have that option

[...]

PR #15451, opened 2025-08-27
feat: better offline operations notification handling

alperozturk96:

### Issue

Multiple offline operations can fail, and users may receive notifications for each failure. If a user attempts to execute the same operation again, the notification may reappear (e.g. https://github.com/nextcloud/android/issues/15420#issuecomment-3221663342). In the worker, certain offline operations are canceled and removed from the database. For example, when an operation completes successfully, or if it is a rename or delete operation. In the case of rename or delete operations, conflicts may occur; to prevent overwriting the server’s latest changes, the worker only informs the user instead of executing the action automatically.

### Changes

This PR introduces a new notification action for failed offline operations, allowing users to manually remove them if they do not wish to retry.

Additionally, the offline operations worker now uses a dedicated notification channel, giving users the option to manage or disable these notifications independently.

<table>
  <tr>
    <th>Action</th>
    <th>New Channel</th>
  </tr>
  <tr>
    <td>before</td>
    <td>after</td>
  </tr>
</table>

**Note: Adding a more descriptive label for the cancel action, such as "Cancel Offline Operation," could make the notification appear cluttered, as the title already conveys the purpose of the notification.**

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/offlineOperations/OfflineOperationsNotificationManager.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/offlineOperations/OfflineOperationsNotificationManager.kt
@@ -81,6 +81,7 @@ class OfflineOperationsNotificationManager(private val context: Context, viewThe
             }
             setContentTitle(text)
             setOngoing(false)
+            setProgress(0, 0, false)
             notificationManager.notify(ERROR_ID, this.build())
         }
     }
--- a/app/src/main/java/com/nextcloud/client/jobs/offlineOperations/OfflineOperationsWorker.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/offlineOperations/OfflineOperationsWorker.kt
@@ -64,10 +64,10 @@ class OfflineOperationsWorker(
 
     override suspend fun doWork(): Result = withContext(Dispatchers.IO) {
         val jobName = inputData.getString(JOB_NAME)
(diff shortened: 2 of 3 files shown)


app/src/main/java/com/nextcloud/client/jobs/notification/WorkerNotificationManager.kt
```diff
@@ -31,15 +32,14 @@ open class WorkerNotificationManager(
...
-            NotificationUtils.NOTIFICATION_CHANNEL_BACKGROUND_OPERATIONS,
+            channelId,
             viewThemeUtils
         ).apply {
             setTicker(context.getString(tickerId))
             setSmallIcon(R.drawable.notification_icon)
             setLargeIcon(BitmapFactory.decodeResource(context.resources, R.drawable.notification_icon))
             setStyle(NotificationCompat.BigTextStyle())
             priority = NotificationCompat.PRIORITY_LOW
-            setChannelId(NotificationUtils.NOTIFICATION_CHANNEL_DOWNLOAD)
```
alperozturk96, line 42: This line was overriding the `NOTIFICATION_CHANNEL_BACKGROUND_OPERATIONS` thus it needs to be removed.

kra-mo (dismissed):
Sounds good to me, with the exception that the channel should probably be sentence-case like the others.

"Offline Operations" → "Offline operations"

kra-mo (approved):

PR #15453, opened 2025-08-27
perf: increase file deletion performance 🚀 

alperozturk96:

### Issues

Previously, when a user attempted to delete multiple files, the `generateResizedImage() `function was triggered for each image file. This behavior is unnecessary, as the user is performing a deletion, not an upload. Instead, the image files should be removed from the cache.

Additionally, `showLoadingDialog()` was being called for each file deletion, which prevented the dialog from appearing correctly and increased UI load.

### Changes

- Removing image files directly from the cache rather than generating resized images.
- Ensuring the loading dialog is displayed correctly during deletion.
- Converting `RemoveFileOperation.java` to Kotlin.


app/src/main/java/com/owncloud/android/operations/RemoveFileOperation.kt
```diff
@@ -0,0 +1,98 @@
...
+        val operation: RemoteOperation<*>?
+
+        // Remove image from thumbnail cache
+        if (MimeTypeUtil.isImage(file.mimeType)) {
+            ThumbnailsCacheManager.removeFromCache(file)
+        }
+
+        var localRemovalFailed = false
+
+        if (!onlyLocalCopy) {
```
tobiasKaminsky: flip if branches

app/src/main/java/com/owncloud/android/operations/RemoveFileOperation.kt
```diff
@@ -0,0 +1,98 @@
...
+            } else {
+                RemoveFileRemoteOperation(file.remotePath)
+            }
+
+            result = operation.execute(client)
+            if (result.isSuccess || result.code == RemoteOperationResult.ResultCode.FILE_NOT_FOUND) {
+                localRemovalFailed = !storageManager.removeFile(file, true, true)
+            }
+        } else {
+            localRemovalFailed = !storageManager.removeFile(file, false, true)
```
tobiasKaminsky: re-add local generation of resized images if image is deleted only locally, to save server request

```suggestion
          if (MimeTypeUtil.isImage(fileToRemove.getMimeType())) {
            // store resized image
            ThumbnailsCacheManager.generateResizedImage(fileToRemove);
        }
            localRemovalFailed = !storageManager.removeFile(file, false, true)
```

tobiasKaminsky:
```
 NP: In class com.owncloud.android.services.OperationsService
In class com.owncloud.android.services.OperationsService
In method com.owncloud.android.services.OperationsService.newOperation(Intent)
Value loaded from file
Method invoked at OperationsService.java:[line 699]
Known null at OperationsService.java:[line 692]
```

PR #15474, opened 2025-08-29
fix: empty list text

alperozturk96:

### Issue

The `setEmptyListLoadingMessage()` function was performing an asynchronous `isNetworkAndServerAvailable()` check before updating the UI. This introduced timing issues where the wrong text message could appear in certain scenarios.

### How to reproduce?

1. Open the app with an empty root directory.
2. Observe that the empty list message may display incorrectly.

### Changes

Removed the internal `isNetworkAndServerAvailable()` check from `setEmptyListLoadingMessage().`
The method now directly updates the UI with the loading message, since network availability is already handled by the calling logic (e.g., `FolderPickerActivity.mSyncInProgress`,` LocalFileListFragment.setLoading`).

Ensured `setBackgroundText()` is called whenever `mSyncInProgress` changes in `FileDisplayActivity`.

Updated `FileDisplayActivity` to set `mSyncInProgress` to false once operations are finished, preventing stale loading messages.

### Before

https://github.com/user-attachments/assets/3dd001e9-34a2-4e76-84d1-04e221ed09c6

### After

https://github.com/user-attachments/assets/e007ff54-41d6-482b-8329-c226374321c1

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
@@ -190,7 +190,11 @@ class FileDisplayActivity :
 
     private var mWaitingToPreview: OCFile? = null
 
-    private var mSyncInProgress = false
+    private var mSyncInProgress: Boolean = false
+        set(value) {
+            field = value
+            setBackgroundText()
+        }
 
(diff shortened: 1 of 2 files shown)


alperozturk96:
replaced with https://github.com/nextcloud/android/pull/15475

PR #15475, opened 2025-08-29
fix: empty list state handling

alperozturk96:

### Issue

The `setEmptyListLoadingMessage()` function was performing an asynchronous `isNetworkAndServerAvailable()` check before updating the UI. This introduced timing issues where the wrong text message could appear in certain scenarios.

Multiple functions were previously used to set the empty list state, leading to inconsistent behavior.

The same state was sometimes applied consecutively, resulting in redundant operations.

### How to reproduce?

1. Open the app with an empty root directory.
2. Observe that the empty list message may display incorrectly.

### Changes

Removed the internal `isNetworkAndServerAvailable()` check from `setEmptyListLoadingMessage().`
The method now directly updates the UI with the loading message, since network availability is already handled by the calling logic (e.g., `FolderPickerActivity.mSyncInProgress`,` LocalFileListFragment.setLoading`).

Ensured `setBackgroundText()` is called whenever `mSyncInProgress` changes in `FileDisplayActivity`.

Updated `FileDisplayActivity` to set `mSyncInProgress` to false once operations are finished, preventing stale loading messages.

Consolidate empty list state handling into a single function, `setEmptyListMessage(state: Parcelable)`

Consolidate empty list state handling by separating empty list state from search event into distinct `Parcelable` enum classes.

### Before

https://github.com/user-attachments/assets/3dd001e9-34a2-4e76-84d1-04e221ed09c6

### After

https://github.com/user-attachments/assets/4c01cfa8-dc9f-4c6f-95f1-bf00858fc3f9

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
@@ -21,6 +21,9 @@ import android.annotation.SuppressLint
 import android.content.Context
 import android.content.res.Configuration
 import android.os.Bundle
+import android.os.Handler
+import android.os.Looper
+import android.os.Parcelable
 import android.util.DisplayMetrics
 import android.view.KeyEvent
 import android.view.LayoutInflater
@@ -42,7 +45,6 @@ import androidx.annotation.StringRes
 import androidx.appcompat.widget.SearchView
 import androidx.core.content.ContextCompat
 import androidx.fragment.app.Fragment
-import androidx.lifecycle.lifecycleScope
 import androidx.recyclerview.widget.GridLayoutManager
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
```diff
@@ -570,46 +560,51 @@ open class ExtendedListFragment :
...
-        lifecycleScope.launch(Dispatchers.Main) {
+        val newData = EmptyListData(headline, message, icon, tintIcon)
+        if (previousEmptyListData == newData) {
+            Log_OC.d(TAG, "Skipping setMessageForEmptyList – same data as before")
+            return
+        }
+        previousEmptyListData = newData
+        Log_OC.d(TAG, "Updating setMessageForEmptyList")
+
+        Handler(Looper.getMainLooper()).post {
```
tobiasKaminsky, line 581: two times handler?
I guess one can be removed

tobiasKaminsky (changes requested):
see inline :)

app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
alperozturk96: https://github.com/nextcloud/android/pull/15475/commits/48ce57e631d3a93b713f5557fd070d494bf9f577

tobiasKaminsky (approved):

alperozturk96:
Unrelated spot bug

Screenshot 2025-09-04 at 16 16 07

PR #15495, opened 2025-09-01
feat: add recommended item view holder

alperozturk96:

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>before</td>
    <td>after</td>
  </tr>
</table>

### Changes

- Uses a new separate view holder for recommended files
- Filters out correct actions for recommended files since user cannot delete, move, set as favorite, encrypt etc...  recommend file.
- Fixes multi select
- Adds reason text view element

### Note

Since we’ve added a new text view element for the "reason," the "more" button now needs to align properly between the filename + extension and the reason. To ensure this alignment, I’ve increased the width allocation, which is why you’re now seeing three items instead of four.

If you have any suggestions for a better approach, I’d be happy to hear them. Thanks!


kra-mo (approved):

alperozturk96:
Fixes https://github.com/nextcloud/android/pull/15495/commits/f21edf06769c3f9470e7df5a328d97a16280757d

Screenshot 2025-09-02 at 16 41 50

AndyScherzinger (approved):

PR #15501, opened 2025-09-02
fix: add account name to fetch recommended files

alperozturk96:

### Issue

Current user seeing the another user's recommended file

### Changes

Use account name for caching and fetching


app/schemas/com.nextcloud.client.database.NextcloudDatabase/91.json
```diff
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 91,
```
alperozturk96: Since v91 never released, no need to create another DB version.

https://github.com/nextcloud/android/blob/stable-3.33/app/src/main/java/com/owncloud/android/ui/fragment/filesRepository/RemoteFilesRepository.kt

AndyScherzinger: Any daily build installation would break, so while true for stable releases, untrue for pre-release builds

alperozturk96: https://github.com/nextcloud/android/pull/15501/commits/b80d726fd06640f87b3a59708d97c696e361f6c4

AndyScherzinger: Ah, you meant only within this branch 👍

PR #15536, opened 2025-09-05
fix: manage space activity screen

alperozturk96:

### Changes

Remove deprecated `AsyncTask` use instead `Coroutines`.
Replace `LinearLayout` with `ConstraintLayout`.
Fix status bar alignment
Check nullable cache parent dir path

### Note

Since this screen is accessed through the app settings and functions more as an integration point rather than a core part of our app, I didn’t apply `viewThemeUtils` for branding. However, I can update it to use the `viewThemeUtils` if needed.

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>before</td>
    <td>after</td>
  </tr>
</table>

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -616,8 +616,7 @@
         <activity
             android:name=".ui.activity.ManageSpaceActivity"
             android:exported="false"
-            android:label="@string/manage_space_title"
-            android:theme="@style/Theme.ownCloud" />
+            android:label="@string/manage_space_title" />
 
         <service
             android:name=".services.AccountManagerService"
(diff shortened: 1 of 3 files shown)


kra-mo (approved):
Definitely an improvement (lol) although the wording on this screen is probably worth revisiting sometime.

My one question is

> I didn’t apply viewThemeUtils for branding

is this done elsewhere? Because if not, I wouldn't start here. But if yes, ignore what I said and just proceed :)

alperozturk96:
> Definitely an improvement (lol) although the wording on this screen is probably worth revisiting sometime.
> 
> My one question is
> 
> > I didn’t apply viewThemeUtils for branding
> 
> is this done elsewhere? Because if not, I wouldn't start here. But if yes, ignore what I said and just proceed :)

We are using `viewThemeUtils` for branding, such as adding icon tinting, background colors for the bottom sheet, alert dialogs, etc. However, on this screen, I didn’t use it because it’s an integration that users can access via App Info → Storage & Cache → Clear Storage, which exists for any Android app.

PR #15570, opened 2025-09-08
rename text name to avoid transifex problem

tobiasKaminsky:


AndyScherzinger (approved):
👍 But haven't understood the transifex issue.

tobiasKaminsky:
> 👍 But haven't understood the transifex issue.

on master the same message id "clear_status_message_after" is: "Clear status after"
But on stable-3.33: it is "Clear status message after"

Since we combine all stable branches + master to have one common file to translate via Transifex, this led then to a clash when compiling on Android, since the resulting file then has 2x clear_status_message_after as an entry within strings.xml

Thus it needs to be fixed on all stables or on master.

AndyScherzinger:
Ah, so it is that again

PR #15586, opened 2025-09-09
fix: auto upload

alperozturk96:

### Issue

Previously, `FilesSyncWork` used the `REPLACE` work policy, which would cancel any unfinished work whenever a new worker was enqueued. In scenarios such as users taking multiple photos quickly in the DCIM folder, this could lead to:

1. Ongoing uploads being killed prematurely.
2. `updateFilesystemFileAsSentForUpload` not being called, causing files to be re-uploaded unnecessarily.
3. Execution order issues, since files were first dispatched to `FileUploadWorker` and only later marked as uploaded.

### Changes

**Synchronous upload flow:**

Removed the dependency on `FileUploadWorker`. `FilesSyncWork` now directly uses `UploadFileOperation` per file.
Each file is marked as “sent for upload” before uploading, ensuring correct DB state and sequential execution.

**Foreground service for long-running jobs:**

`ForegroundInfo` is now set for `FilesSyncWork` with `ForegroundServiceType.DataSync.`
Ensures Android treats this as a long-running background task.


tobiasKaminsky:
> @tobiasKaminsky Do we need to show notification for auto upload? Since auto upload continuous process and showing notification for each upload can be too much?

No need to show it, I think :)

PR #15591, opened 2025-09-09
SpotBugs: Fix Method ignores exceptional return value - RenameFileOperation

alexknop:
There are 16 SpotBugs under

Bad practice:

Solution:

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/RenameFileOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/RenameFileOperation.java
@@ -185,7 +185,7 @@ public class RenameFileOperation extends SyncOperation {
         boolean result = Files.exists(testFile) && Files.isRegularFile(testFile);
 
         try {
-            Files.delete(testFile);
+            Files.deleteIfExists(testFile);
         } catch (Exception e) {
             Log_OC.e("Error deleting file: ", e.getMessage());
             return true;


tobiasKaminsky:
Can you check why spotbugs has increased?

alexknop:
@tobiasKaminsky I'd like to check it again, I couldn't open the link, probably expired.

app/src/main/java/com/owncloud/android/operations/RenameFileOperation.java
```diff
@@ -161,26 +164,27 @@ private boolean isValidNewName() throws IOException {
...
+            }
         }
-        if (!tmpFolder.isDirectory()) {
+        if (tmpFolder != null && !Files.isDirectory(tmpFolder)) {
             throw new IOException("Unexpected error: temporal directory could not be created");
         }
         try {
-            testFile.createNewFile();   // return value is ignored; it could be 'false' because
-            // the file already existed, that doesn't invalidate the name
+            Files.createFile(testFile);
```
alperozturk96, line 180: `Files.createFile` can throw `FileAlreadyExistsException`, `UnsupportedOperationException` these exceptions as well. Please return false in those cases.

app/src/main/java/com/owncloud/android/operations/RenameFileOperation.java
```diff
@@ -161,26 +164,27 @@ private boolean isValidNewName() throws IOException {
...
             Log_OC.i(TAG, "Test for validity of name " + newName + " in the file system failed");
             return false;
         }
-        boolean result = testFile.exists() && testFile.isFile();
+        boolean result = Files.exists(testFile) && Files.isRegularFile(testFile);
 
-        // cleaning ; result is ignored, since there is not much we could do in case of failure,
-        // but repeat and repeat...
-        testFile.delete();
+        Files.delete(testFile);
```
alperozturk96: `Files.delete(testFile);` can throw following exceptions. Please return true in those cases since previously 

We had this comment.

`// cleaning ; result is ignored, since there is not much we could do in case of failure,` @tobiasKaminsky Do you have any idea about this?

Image

alperozturk96 (changes requested):
Hello, thank you for submitting the PR. It looks good overall, but exception handling needs to be added or improved.

app/src/main/java/com/owncloud/android/operations/RenameFileOperation.java
```diff
@@ -161,26 +164,32 @@ private boolean isValidNewName() throws IOException {
...
         }
-        boolean result = testFile.exists() && testFile.isFile();
+        boolean result = Files.exists(testFile) && Files.isRegularFile(testFile);
 
-        // cleaning ; result is ignored, since there is not much we could do in case of failure,
-        // but repeat and repeat...
-        testFile.delete();
+        try {
+            Files.delete(testFile);
+        } catch (IOException e) {
```
alperozturk96:
```suggestion
        } catch (Exception e) {
```

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/operations/RenameFileOperation.java
```diff
@@ -161,26 +164,32 @@ private boolean isValidNewName() throws IOException {
...
             return false;
         }
-        boolean result = testFile.exists() && testFile.isFile();
+        boolean result = Files.exists(testFile) && Files.isRegularFile(testFile);
 
-        // cleaning ; result is ignored, since there is not much we could do in case of failure,
-        // but repeat and repeat...
-        testFile.delete();
+        try {
+            Files.delete(testFile);
```
alperozturk96:
```suggestion
            Files. deleteIfExists(testFile);
```

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/operations/RenameFileOperation.java
alexknop: fixed

alperozturk96 (approved):

PR #15595, opened 2025-09-09
SpotBugs: Fix Method ignores exceptional return value - PrintAsyncTask

alexknop:
There are 16 SpotBugs under

Bad practice:

Solution:

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/asynctasks/PrintAsyncTask.java
+++ b/app/src/main/java/com/owncloud/android/ui/asynctasks/PrintAsyncTask.java
@@ -58,69 +58,67 @@ public class PrintAsyncTask extends AsyncTask<Void, Void, Boolean> {
 
     @Override
     protected Boolean doInBackground(Void... voids) {
+        if (file == null) return Boolean.FALSE;
+
         HttpClient client = new HttpClient();
         GetMethod getMethod = null;
 
-        FileOutputStream fos = null;
         try {
             getMethod = new GetMethod(url);
             int status = client.executeMethod(getMethod);
-            if (status == HttpStatus.SC_OK) {
-                if (file.exists() && !file.delete()) {
-                    return Boolean.FALSE;
-                }
+            if (status != HttpStatus.SC_OK) return Boolean.FALSE;
 
-                Files.createDirectories(file.toPath().getParent());
+            if (file.exists() && !file.delete()) return Boolean.FALSE;
 
-                if (!Objects.requireNonNull(file.getParentFile()).exists()) {
-                    Log_OC.d(TAG, file.getParentFile().getAbsolutePath() + " does not exist");
-                    return Boolean.FALSE;
-                }
+            File parentFile = file.getParentFile();
+            if (parentFile == null) return Boolean.FALSE;
 
-                if (!file.createNewFile()) {
-                    Log_OC.d(TAG, file.getAbsolutePath() + " could not be created");
-                    return Boolean.FALSE;
-                }
+            Files.createDirectories(parentFile.toPath());
+            if (!parentFile.exists()) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/ui/asynctasks/PrintAsyncTask.java
```diff
@@ -68,7 +70,11 @@ protected Boolean doInBackground(Void... voids) {
                     return Boolean.FALSE;
                 }
 
-                file.getParentFile().mkdirs();
+                try {
+                    Files.createDirectories(Objects.requireNonNull(file.getParentFile()).toPath());
+                } catch (IOException e) {
+                    Log_OC.e(TAG, "Could not create directory: " + Objects.requireNonNull(file.getParentFile()).getAbsolutePath(), e);
```
tobiasKaminsky: If file or file.getParentFile() is null, and we catch it and then throw another one via requireNonNull, it is probably better to 
- return in catch block
- do not have those nestet try at all?

app/src/main/java/com/owncloud/android/ui/asynctasks/PrintAsyncTask.java
```diff
@@ -68,9 +70,9 @@ protected Boolean doInBackground(Void... voids) {
                     return Boolean.FALSE;
                 }
 
-                file.getParentFile().mkdirs();
+                Files.createDirectories(file.toPath().getParent());
```
alperozturk96: Thank you for the PR.
I recommend using the following implementation, as it includes proper nullability checks for both the file and its parent directory. It also uses try-with-resources for `BufferedInputStream` and `FileOutputStream`, and handles all potential exceptions.

```
@Override
    protected Boolean doInBackground(Void... voids) {
        if (file == null) return Boolean.FALSE;

        HttpClient client = new HttpClient();
        GetMethod getMethod = null;

        try {
            getMethod = new GetMethod(url);
            int status = client.executeMethod(getMethod);
            if (status != HttpStatus.SC_OK) return Boolean.FALSE;

            if (file.exists() && !file.delete()) return Boolean.FALSE;

            File parentFile = file.getParentFile();
            if (parentFile == null) return Boolean.FALSE;

            Files.createDirectories(parentFile.toPath());
            if (!parentFile.exists()) {
                Log_OC.d(TAG, parentFile.getAbsolutePath() + " does not exist");
                return Boolean.FALSE;
            }

            if (!file.createNewFile()) {
                Log_OC.d(TAG, file.getAbsolutePath() + " could not be created");
                return Boolean.FALSE;
            }

            Header contentLengthHeader = getMethod.getResponseHeader("Content-Length");
            long totalToTransfer = 0;
            if (contentLengthHeader != null && !contentLengthHeader.getValue().isEmpty()) {
                totalToTransfer = Long.parseLong(contentLengthHeader.getValue());
            }

            try (
                BufferedInputStream bis = new BufferedInputStream(getMethod.getResponseBodyAsStream());
                FileOutputStream fos = new FileOutputStream(file)
            ) {
                byte[] buffer = new byte[4096];
                long transferred = 0;
                int read;

                while ((read = bis.read(buffer)) != -1) {
                    fos.write(buffer, 0, read);
                    transferred += read;
                }

                if (totalToTransfer > 0 && transferred != totalToTransfer) {
                    Log_OC.d(TAG, "Transferred bytes (" + transferred +
                        ") != expected (" + totalToTransfer + ")");
                    return Boolean.FALSE;
                }
            }

            return Boolean.TRUE;
        } catch (Exception e) {
            Log_OC.e(TAG, "Error downloading file", e);
            return Boolean.FALSE;
        } finally {
            if (getMethod != null) {
                getMethod.releaseConnection();
            }
        }
    }
```

PR #15597, opened 2025-09-09
SpotBugs: Fix Method ignores exceptional return value - FileStorageUtils

alexknop:
There are 16 SpotBugs under

Bad practice:

Solution:

### DIFF
--- a/app/src/main/java/com/owncloud/android/utils/FileStorageUtils.java
+++ b/app/src/main/java/com/owncloud/android/utils/FileStorageUtils.java
@@ -27,9 +27,8 @@ import com.owncloud.android.datamodel.FileDataStorageManager;
 import com.owncloud.android.datamodel.OCFile;
 import com.owncloud.android.lib.common.utils.Log_OC;
 import com.owncloud.android.lib.resources.files.model.RemoteFile;
+import com.owncloud.android.lib.resources.shares.ShareeUser;
 import com.owncloud.android.ui.helpers.FileOperationsHelper;
-
-import org.apache.commons.io.FilenameUtils;
 
 import java.io.File;
 import java.io.FileInputStream;
@@ -40,6 +39,7 @@ import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


tobiasKaminsky (approved):

alperozturk96 (changes requested):
`containsBidiControlCharacters` and some other functions are missing.

alexknop:
@alperozturk96 This is fixed, not sure what happened before.

app/src/main/java/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -531,7 +532,7 @@ public static void deleteRecursively(File file, FileDataStorageManager storageMa
         }
 
         storageManager.deleteFileInMediaScan(file.getAbsolutePath());
-        file.delete();
+        Files.delete(file.toPath());
```
alperozturk96: The line `Files.delete(testFile);` can throw several exceptions. Please make sure to handle these cases. Logging the exception using `Log_OC.e` would be sufficient.

Since we don’t have separate handling logic for each exception type, you may consider catching the general `Exception` class instead of individual exceptions.

Image

app/src/main/java/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -535,7 +536,11 @@ public static void deleteRecursively(File file, FileDataStorageManager storageMa
         }
 
         storageManager.deleteFileInMediaScan(file.getAbsolutePath());
-        file.delete();
+        try {
+            Files.delete(file.toPath());
+        } catch (IOException e) {
```
alperozturk96:
```suggestion
        } catch (Exception e) {
```

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/utils/FileStorageUtils.java
alexknop: fixed

app/src/main/java/com/owncloud/android/utils/FileStorageUtils.java
```diff
@@ -535,7 +536,11 @@ public static void deleteRecursively(File file, FileDataStorageManager storageMa
         }
 
         storageManager.deleteFileInMediaScan(file.getAbsolutePath());
-        file.delete();
+        try {
+            Files.delete(file.toPath());
```
alperozturk96: I encountered this exception as well, although upon checking, the files were successfully deleted.

Image

To prevent this issue, we can replace the current line with: 

```suggestion
            Files.deleteIfExists(file.toPath());
```

Then I will merge it. Thank you for your help, really appreciate your effort! 💯

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/utils/FileStorageUtils.java
alexknop: fixed

Issue #15376, opened 2025-08-17
Option to have denser row spacing


golemusu:
Nextcloud has very large spaces between lines. It is fine as a default behaviour but it would be handy to give user the option to make UI much denser to fit more lines to screen. 

1. in main screen (files, folders)

2. in 3-dot menus

3. elsewhere...?

PR #15627, opened 2025-09-12
Implement dense layout option

eltos:
~Adds an option in settings to~ increase the layout density.

Closes https://github.com/nextcloud/android/issues/15376

**EDIT:** That PR was updated to have dense layout by default, but not remove the share button. Note that the screenshot is still the initial one where also the share button was removed.

|Default Layout (before)|Dense Layout (after)| 
|-|-|
|default|dense

### DIFF
--- a/app/src/main/java/com/nextcloud/client/preferences/AppPreferences.java
+++ b/app/src/main/java/com/nextcloud/client/preferences/AppPreferences.java
@@ -277,8 +277,6 @@ public interface AppPreferences {
      */
     DarkMode getDarkThemeMode();
 
-    boolean isDenseLayoutEnabled();
-
     /**
      * Saves the uploader behavior which the user has set last.
      *
--- a/app/src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
+++ b/app/src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
@@ -58,7 +58,6 @@ public final class AppPreferencesImpl implements AppPreferences {
     public static final String DATA_STORAGE_LOCATION = "data_storage_location";
     public static final String STORAGE_PATH_VALID = "storage_path_valid";
     public static final String PREF__DARK_THEME = "dark_theme_mode";
-    public static final String PREF__DENSE_LAYOUT = "dense_layout";
     public static final float DEFAULT_GRID_COLUMN = 3f;
 
     private static final String AUTO_PREF__LAST_UPLOAD_PATH = "last_upload_path";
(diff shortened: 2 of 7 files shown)


tobiasKaminsky:
@eltos thanks for that nice improvement.
Are you fine with Laura and Jans idea to make it the default one? (And also show share button again?)

eltos:
Can the option remain to allow hiding the share button (while spacings are always reduced as you suggest)? Or no option just dense layout with share button?

Personally, I find the share button quite useless, as the three dot menu contains an easily accessible share menu with more features (local share via other app & various nextcloud share options). The extra button is mainly taking space and causes the filename to be shortened. On PC there is plenty more space, but IMHO mobile UI should account for the fact that less space is available. This is of course my personal opinion. Your decision. Let me know and I can update the PR.

kra-mo:
> Personally, I find the share button quite useless, as the three dot menu contains an easily accessible share menu with more features (local share via other app & various nextcloud share options). The extra button is mainly taking space and causes the filename to be shortened. On PC there is plenty more space, but IMHO mobile UI should account for the fact that less space is available. This is of course my personal opinion. Your decision. Let me know and I can update the PR.

I do agree with you from this point of view, but I'm not sure what the full context for its inclusion is. It does certainly make the feature more discoverable, especially to users of the web interface.

For this reason, I suggest splitting the PR out into two. Leave this for just the dense layout and create one removing the share button altogether. Then this one can be merged and we can continue the discussion about the share button there.

cc @jancborchardt if you want to discuss the share icon later.

eltos:
> For this reason, I suggest splitting the PR out into two. Leave this for just the dense layout and create one removing the share button altogether. Then this one can be merged and we can continue the discussion about the share button there.

I have updated the PR accordingly and created https://github.com/nextcloud/android/pull/15714 for the share button topic.

jancborchardt (approved):
This is good from my side then. :) (not tested)

PhilLab:
Given this was approved, can it be merged, @kra-mo , @tobiasKaminsky ? Or has it gotten outdated now?

alperozturk96:
@eltos 

Hi

Thank you for the PR. I wanted to confirm `list_item.xml` is the only file that has been changed? Sometimes, changes in the master branch can make the PR appear differently (after force-pushes, merging to master etc).

eltos:
In the end yes only this

golemusu:
It is much better like this but IMO there could be option to make it even more denser (e.g. "very compact"). So that there is barely space between texts. Obviously it might require making the icons on left side a bit smaller and I am not suggesting it as default UI but as option for those for whom main consideration is to have as many lines of text to the screen as possible and visibility of icons is not very important.

PR #15648, opened 2025-09-16
Fix crash when decoding filenames with invalid % characters

chagaiB:
This PR fixes a crash caused by URLDecoder when filenames contain invalid '%' patterns (like "65% on Sonic Wave").
It adds a safeUrlDecode() method that catches IllegalArgumentException and returns the original string instead of crashing.


alperozturk96:
Hey, thank you for the PR. I can't test it. It is not building.

Screenshot 2025-10-01 at 09 40 37

alperozturk96 (changes requested):
Hey, thank you for the PR. I can't test it. It is not building.

Screenshot 2025-10-01 at 09 40 37

alperozturk96:
Since `containsBidiControlCharacters` function in `FileStorageUtils` updated, closing this PR.

PR #15661, opened 2025-09-19
Opt-in to Arm MTE

thgoebel:
This is an easy security improvement (at least for users with Google Pixel 8 and later, which have the hardware for MTE).

Nextcloud doesn't have any explicit native code itself. But there are some libandroidx.graphics.path.so files in the APK, and also Android framework code that Nextcloud calls might have native code. All of that runs in the app's process, so opting in to MTE will cover that. Nextcloud parses potentially attacker controlled data, so enabling MTE is an easy defense-in-depth.

I've been running Nextcloud on a Pixel 8a with GrapheneOS and MTE force-enabled for a few months now without any issues.
Still, I recommend you to test this again on an MTE-compatible device (Pixel 8 and later), just to be sure :)

I'm only sending PRs for Files and Notes for Android, because those I use a lot and can (at)test.
I recommend you enable MTE for other Android apps (Talk, Deck) as well, and also for the Nextcloud iOS Files app.

For background on MTE, see:

- Android docs: https://source.android.com/docs/security/test/memory-safety/arm-mte
- iOS docs: https://developer.apple.com/documentation/xcode/enabling-enhanced-security-for-your-app
- My blog post :innocent: : https://thore.io/posts/2025/09/introduction-to-arm-memory-tagging-extensions/


tobiasKaminsky:
I think this is the more accurate documentation: https://developer.android.com/ndk/guides/arm-mte

Your linked one is for platform devs.

I unfortunately do not have such a device.

@mahibi do you have a Pixel?

mahibi:
> @mahibi do you have a Pixel?

just a Pixel 6a with Graphene which is currently not usable (have to set it up again from scratch i think)

thgoebel:
You need a Pixel 8 or later. The Pixel 8a is currently the cheapest option (~400€).

tobiasKaminsky:
let us merge it and see via dev version if others have troubles with ti.

PR #15669, opened 2025-09-21
Fix deleting from slideshow

### LINKED ISSUE #15636: Slideshow is left when deleting file
### Steps to reproduce

1. have multiple images in a folder
2. click on one, to open slide show
3. delete one
4. --> folder listing is shown

### Expected behaviour

1. have multiple images in a folder
2. click on one, to open slide show
3. delete one
4. --> another image is shown

### Actual behaviour

The slideshow is left and the parent folder is shown again.

`3.32.3` . Also reproducible on master

### Additional information

This functionality seems to break from time to time:
1. Initially it was fixed in https://github.com/nextcloud/android/pull/11919
2. The logic was improved and made more explicit in [pull/13016](https://github.com/nextcloud/android/pull/13016/files#diff-8add18fc6eb736e5a6fbccada743e708fe24ffda050771284d562caa57dc35c9R277)
3. Somewhere it regressed again
4. Was fixed again in https://github.com/nextcloud/android/pull/14646
5. Then it broke again. I suspect https://github.com/nextcloud/android/pull/14782 , which added `finishPreviewImageActivity()` to the `RemoveFilesDialogFragment`, so that `PreviewImageActivity::onRemoteOperationFinish()` with the responsible logic is never called

**On a secondary notice**, the position handling in `PreviewImageActivity::onRemoteOperationFinish()` also seems to have some bugs, which we could fix in the same go:
1. When deleting the first image in a folder, it doesn't jump to the second, but the third image. If there is no third, an empty view appears
2. Sometimes (don't know yet, when) it also jumps out of the slideshow.

**Lastly** all this regressing and being faulty indicates we need an automated test for this.

PhilLab:

Fixes #15636.

1. The patch added in #14782 prevented that the callback `onRemoteOperationFinish()` was ever called.
The changes were removed. I was not able to reproduce the end-to-end encryption issue, so this seems fine.
    - @alperozturk96 if you could try yourself, as you had originally worked on that issue?
2. the manual update of adapter and pager is only needed if user is not present. When user is present, the `initViewPager()` takes care of everything.
3. The next image was inconsistent between user present yes/no

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.kt
+++ b/app/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.kt
@@ -277,8 +277,6 @@ class PreviewImageActivity :
         super.onRemoteOperationFinish(operation, result)
 
         if (operation is RemoveFileOperation) {
-            val deletePosition = viewPager?.currentItem ?: return
-            val nextPosition = if (deletePosition > 0) deletePosition - 1 else 0
 
             previewImagePagerAdapter?.let {
                 if (it.itemCount <= 1) {
@@ -289,10 +287,12 @@ class PreviewImageActivity :
 
             if (user.isPresent) {
                 initViewPager(user.get())
+            } else {
+                val deletePosition = viewPager?.currentItem ?: return
+                val nextPosition = if (deletePosition > 0) deletePosition - 1 else 0
+                viewPager?.setCurrentItem(nextPosition, true)
+                previewImagePagerAdapter?.delete(deletePosition)
             }
-
-            viewPager?.setCurrentItem(nextPosition, true)
-            previewImagePagerAdapter?.delete(deletePosition)
         } else if (operation is SynchronizeFileOperation) {
             onSynchronizeFileOperationFinish(result)
         }
--- a/app/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.kt
+++ b/app/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.kt
@@ -286,12 +286,19 @@ class PreviewImageActivity :
             }
 
             if (user.isPresent) {
+                // Re-init the view pager, which will advance to the next image
                 initViewPager(user.get())
-            } else {
+            } else if (result.isSuccess) {
+                // If deletion was successful, update adapter and display next image
                 val deletePosition = viewPager?.currentItem ?: return
-                val nextPosition = if (deletePosition > 0) deletePosition - 1 else 0
-                viewPager?.setCurrentItem(nextPosition, true)
-                previewImagePagerAdapter?.delete(deletePosition)
+                previewImagePagerAdapter?.let { adapter ->
+                    // advance to the next image if possible, since initViewPager() also advances forwards
+                    val nextPosition = if (deletePosition < (adapter.itemCount - 1)) deletePosition + 1 else max(deletePosition - 1, 0)
+                    viewPager?.setCurrentItem(nextPosition, true)
+                    adapter.delete(deletePosition)
+                    // Page needs to be reselected after the adapter has been updated. Otherwise, wrong title is shown
+                    selectPage(nextPosition)
+                }
             }
         } else if (operation is SynchronizeFileOperation) {
             onSynchronizeFileOperationFinish(result)
--- a/app/src/main/java/com/owncloud/android/ui/preview/PreviewImagePagerAdapter.kt
+++ b/app/src/main/java/com/owncloud/android/ui/preview/PreviewImagePagerAdapter.kt
@@ -194,4 +194,10 @@ class PreviewImagePagerAdapter : FragmentStateAdapter {
     override fun createFragment(position: Int): Fragment = getItem(position)
 
     override fun getItemCount(): Int = imageFiles.size
+
+    override fun getItemId(position: Int): Long {
+        // The item ID function is needed to detect whether the deletion of the current item needs a UI update
+        // Use the OCFile id, fallback to position if not available
+        return imageFiles.getOrNull(position)?.fileId?.toLong() ?: position.toLong()
+    }
 }
--- a/app/src/main/java/com/owncloud/android/ui/dialog/RemoveFilesDialogFragment.kt
+++ b/app/src/main/java/com/owncloud/android/ui/dialog/RemoveFilesDialogFragment.kt
@@ -24,7 +24,6 @@ import com.owncloud.android.datamodel.OCFile
 import com.owncloud.android.ui.activity.FileActivity
 import com.owncloud.android.ui.activity.FileDisplayActivity
 import com.owncloud.android.ui.dialog.ConfirmationDialogFragment.ConfirmationDialogFragmentListener
-import com.owncloud.android.ui.preview.PreviewImageActivity
 import javax.inject.Inject
 
 /**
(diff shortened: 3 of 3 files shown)


app/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.kt
```diff
@@ -289,11 +287,20 @@ class PreviewImageActivity :
             }
 
             if (user.isPresent) {
+                // Re-init the view pager, which will advance to the next image
                 initViewPager(user.get())
+            } else if (result.isSuccess) {
+                // If deletion was successful, update adapter and display next image
```
alperozturk96: Please remove comments and extract `else if (result.isSuccess) {` logic to the separate function, e.g. `updateViewPagerAfterDeletion()` thus it will be also clear without comments.

app/src/main/java/com/owncloud/android/ui/preview/PreviewImagePagerAdapter.kt
```diff
@@ -201,4 +201,10 @@ class PreviewImagePagerAdapter : FragmentStateAdapter {
     override fun createFragment(position: Int): Fragment = getItem(position)
 
     override fun getItemCount(): Int = imageFiles.size
+
+    override fun getItemId(position: Int): Long {
+        // The item ID function is needed to detect whether the deletion of the current item needs a UI update
+        // Use the OCFile id, fallback to position if not available
+        return imageFiles.getOrNull(position)?.fileId?.toLong() ?: position.toLong()
```
alperozturk96: `getItemId()` is part of `androidx.viewpager2.adapter.FragmentStateAdapter`. No need to add an explanation here since the method is already documented in the official API.

Additionally, please remove unnecessary `toLong()` cast for `fileId` since `fileId` is already Long.

alperozturk96 (changes requested):
Hello

I tested changes for encrypted folder as well. I didn't come across any problem. Besides two small things, PR looks okay.

Thank you for the PR 👍

app/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.kt
PhilLab: @alperozturk96 will do so. Two items, though:
### 1. Weird test experience
Please note, in my initial submission the relevant line was
```
val nextPosition = if (deletePosition < (adapter.itemCount - 1)) deletePosition + 1 else max(deletePosition - 1, 0)
```
and I could've sworn my developer tests (I commented out the enclosing if/else) were successful. However, when testing later, this didn't work and I had to change the part ` deletePosition + 1` to simply ` deletePosition`:
```
val nextPosition = if (deletePosition < (adapter.itemCount - 1)) deletePosition else max(deletePosition - 1, 0)
```
which naturally can be simplified to 
```
val nextPosition = min(deletePosition, adapter.itemCount - 1)
```

I don't know what happened there in my initial tests. Either I messed up or there is a very weird timing issue going on. But I tested both on an Emulator and a real device and both behaved the same.  Just wanted to let you know.

### 2. Speaking names vs comments
I found it quite relevant that the else branch shall advance forward due to consistency with the if branch. Because the prior code didn't respect this consistency (it specifically had `deletePosition -1`).

So to fulfill your request, I would need to name the method `updateViewPagerAfterDeletionAndAdvanceForward()` , but even then it wouldn't explain that the `initViewPager()` would also advance forward. Are you sure this makes the code more readable?

app/src/main/java/com/owncloud/android/ui/preview/PreviewImagePagerAdapter.kt
PhilLab: @alperozturk96 good that you pointed me to the documentation. Because it says
> When overriding, also override containsItem(long).

So I'll do that. And I'll also remove the cast.

However, I do think that the explanation of this override's purpose is still needed. The official API doesn't explain this, does it?
> The item ID function is needed to detect whether the deletion of the current item needs a UI update

this is how the official API doc reads to me:
```
/**
     * Default implementation works for collections that don't add, move, remove items.
     * <p>
     * When overriding, also override {@link #containsItem(long)}.
     * <p>
     * If the item is not a part of the collection, return {@link RecyclerView#NO_ID}.
     *
     * @param position Adapter position
     * @return stable item id {@link RecyclerView.Adapter#hasStableIds()}
     */
```
So can we keep at least the first line of the comment?

app/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.kt
alperozturk96: We can also rename the `initViewPager()` function. `updateViewPagerAfterDeletionAndAdvanceForward` is good also.

app/src/main/java/com/owncloud/android/ui/preview/PreviewImagePagerAdapter.kt
alperozturk96: Yes, sure thank you.

app/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.kt
PhilLab: Adjusted

app/src/main/java/com/owncloud/android/ui/preview/PreviewImagePagerAdapter.kt
PhilLab: @alperozturk96 adjusted

app/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.kt
```diff
@@ -296,11 +306,11 @@ class PreviewImageActivity :
             }
 
             if (user.isPresent) {
+                // Re-init the view pager, which will advance to the next image
                 initViewPager(user.get())
```
alperozturk96: @PhilLab 

After an image is deleted from the preview, the first condition is matched, and we end up re-initializing the `ViewPager`, but `updateViewPagerAfterDeletionAndAdvanceForward` isn’t triggered. I was thinking, since this logic is already inside the `if (operation is RemoveFileOperation)` block, could we consider calling only `updateViewPagerAfterDeletionAndAdvanceForward` when the result is successful instead of re-initializing the `ViewPager` entirely or any other reason to have `user.present` check and `initViewPager` ?

e.g.

```
if (result.isSuccess) {
     updateViewPagerAfterDeletionAndAdvanceForward()
}
```

What do you think?

alperozturk96:
You can fix `Check / check (spotlessKotlinCheck` via following commands.

```
./gradlew spotlessApply
./gradlew detekt                                
./gradlew spotlessCheck
./gradlew spotlessKotlinCheck
```

app/src/main/java/com/owncloud/android/ui/preview/PreviewImageActivity.kt
PhilLab: @alperozturk96 yes, good idea. Though the `if (user.isPresent) { initViewPager` has been added deliberately [here](https://github.com/nextcloud/android/pull/13340/files#diff-39df01679cf4271eddfde37ae854d16abb9cf1e05afd7ff3fc5d33d8fac611a3R262) and been kept ever since.

So I am wondering what this would break.

alperozturk96: @tobiasKaminsky What do you think?

PhilLab: I've adjusted it and got rid of the if (user.isPresent`. I assume it was only added as a "brute force" method to ensure correct update of the viewpager, whereas a more targeted approach is also working fine

PR #15714, opened 2025-09-29
Hide share button in portrait mode to save space

eltos:
Spin-off from https://github.com/nextcloud/android/pull/15627

Hides the additional share button in portrait mode, where not much space is available. Note that the three dot menu contains an easily accessible share menu with even more features (local share via other app & various nextcloud share options). 
In landscape mode there is no change, it is still shown as before (e.g. useful for tablets).

|Portrait|Landscape|
|-|-|
|portrait|landscape

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java
@@ -14,8 +14,10 @@ import android.accounts.AuthenticatorException;
 import android.accounts.OperationCanceledException;
 import android.annotation.SuppressLint;
 import android.app.Activity;
+import android.content.ComponentCallbacks;
 import android.content.ContentValues;
 import android.content.res.ColorStateList;
+import android.content.res.Configuration;
 import android.content.res.Resources;
 import android.graphics.Color;
 import android.graphics.drawable.Drawable;
@@ -198,6 +200,16 @@ public class OCFileListAdapter extends RecyclerView.Adapter<RecyclerView.ViewHol
         // initialise thumbnails cache on background thread
         ThumbnailsCacheManager.initDiskCacheAsync();
         isRTL = DisplayUtils.isRTL();
+
+        activity.registerComponentCallbacks(new ComponentCallbacks() {
+            @Override
+            public void onConfigurationChanged(@NonNull Configuration newConfig) {
+                notifyDataSetChanged(); // force update of orientation-dependent layout (e.g. share button visibility)
+            }
+            @Override
+            public void onLowMemory() {
+            }
+        });
     }
 
     public boolean isMultiSelect() {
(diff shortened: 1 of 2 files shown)


kra-mo (approved):
Makes sense from my side, keeping it on landscape is a nice touch :)

We hide it based on the width on the web too anyway.

tobiasKaminsky:
After a dicussion with our designers and @AndyScherzinger we want to revert this.
Sharing is a crucial part of our NC file sync and share solution and thus this needs to be kept visible (and as easy as possible).

eltos:
Sharing is also an essential part of the Android ecosystem, and a normal android user expects that there is a share option in file managers. It being in the 3-dot-menu or long-press menu is the usual place where share options are to be found in android apps, and as such users will expect it to be there.

The existing share button was also limited to the Nextcloud-internal sharing, and did not provide the share-as-file option android users usually expect (e.g. to send a file via messenger/email/Bluetooth/etc). For this, they anyway had to go to the 3-dot-menu or long-press menu, where a unified share dialog offering both options exists. 

In my personal opinion (and as it appears also the reviewer's opinion), saving one click for Nextcloud-internal sharing is not worth the permanently truncated file names all over the app.

@tobiasKaminsky A middle-ground could be to add a prominent floating action button to share when the user selects files/folders via long-press, and to promote the sharing / link action in the 3-dot-menu to the top

Another middle-ground could be to remove the permanently visible 3-dot menu for each file, and instead show a share button there (opening the unified share dialog), since the 3-dot menu can be accessed easily after a long-press on the file (which is the usual and common android workflow to open a menu).

kra-mo:
@eltos Hi.

I agree with a lot of your concerns (hence my initial approval), but a similar area was touched recently in the iOS app and it became a topic of broader discussion.

Basically, our sharing options are combined with sharing status, which is something we do in a rather unique way. The presentation of this is also inconsistent across our existing clients and there are definitely issues with it.

With all of that in mind, I will be happy to dig more into this and potentially lead an initiative to redesign how we present sharing options across the apps, including finding a way to consistently declutter the UI. This would relate to upcoming changes to context menus on iOS, the toolbar as well as potential changes to selection modes on the web.

So basically, while I generally agree with you, I think we'll touch this again as part of a more general initiative with more changes across clients to keep everyone happy and actually come up with something consistent.

eltos:
> Basically, our sharing options are combined with sharing status, which is something we do in a rather unique way.

Just to clarify, the PR did not touch the sharing status, for an already shared file/folder the recipients icon is still shown where the share button used to be, and clicking it still brings up the share options.





> With all of that in mind, I will be happy to dig more into this and potentially lead an initiative to redesign how we present sharing options across the apps, including finding a way to consistently declutter the UI.

This is a good approach and much better than simply reverting the PR 👍
Maybe my "middle-ground" suggestions above can be considered in the process :)

Issue #15718, opened 2025-09-30
Indicate when auto-upload is skipped due to no changes and scan interval not reached


alperozturk96:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

Currently, auto-uploads may be skipped if there are no new files and the next scheduled scan is too early. While this helps save resources, it’s invisible to users and may cause confusion when uploads don’t happen as expected.

We should clearly inform users on the upload list screen or in auto upload overview page when uploads are skipped for this reason. This will make the behavior transparent and reduce uncertainty.

jancborchardt:
> Currently, auto-uploads may be skipped if there are no new files

That would be expected and needs no explanation, right?

> and the next scheduled scan is too early.

Why don't we scan also when someone opens the autoupload view?

alperozturk96:
Yes indeed we don't need to indicate "if there are no new files" it was just for logic explanation.

Do you mean "ignore interval scan logic and scan anyway when someone open the auto-upload overview"?

jancborchardt:
> Yes indeed we don't need to indicate "if there are no new files" it was just for logic explanation.

Cool, got it. :) 

> Do you mean "ignore interval scan logic and scan anyway when someone open the auto-upload overview"?

Yep – we can just do a scan outside of the regular interval. Cause if you open the autoupload view, it might be to check, or if there’s issues anyway etc.

PR #15722, opened 2025-09-30
feat: indicate skipped auto upload due to battery saver mode

alperozturk96:
- [x] Tests written, or not not needed

<table>
  <tr>
    <th>Light</th>
    <th>Dark</th>
  </tr>
  <tr>
    <td>before</td>
    <td>after</td>
  </tr>
</table>

<table>
  <tr>
    <th>Before Light Auto Upload Overview</th>
    <th>After Light Auto Upload Overview</th>
  </tr>
  <tr>
    <td>before</td>
    <td>after light</td>
  </tr>
</table>

<table>
  <tr>
    <th>Before Dark Auto Upload Overview</th>
    <th>After Dark Auto Upload Overview</th>
  </tr>
  <tr>
    <td>before</td>
    <td>after dark</td>
  </tr>
</table>

PR #15749, opened 2025-10-07
feat: indicate auto upload scan interval

alperozturk96:

### Demo

https://github.com/user-attachments/assets/a7ac04e4-2982-4a3f-b2df-06b81b4a410f

### Design Discussion

1. Icon vs. Text:

Do you think the icon alone is sufficient to communicate the “next possible scan” status?
Initially, I considered adding a short text label, but space in the header section is limited, and adding text might make it feel cluttered.

2. Visual Cues (Color or State):

If we rely on the icon alone, should we consider using color variations to indicate scan timing or status — for example, grey when the next scan is far away and blue when it’s approaching?

Alternatively, do you have other suggestions for visually emphasizing or differentiating this state? @kra-mo @jancborchardt 

### Technical Discussion

@tobiasKaminsky Are we going to implement [this suggestion](https://github.com/nextcloud/android/issues/15718#issuecomment-3370849353), what do you think?


jancborchardt (commented):
Hmm, but didnt we say we just do it automatically, no indicator? See my comment on the issue

kra-mo:
I do have to say that this does feel more like debugging information than something users would actually want to see.

(Not a review of the feature, just more generally I'm not a fan of extra things for users to have to think about. In the era of Google Photos which "just works", this feels backwards.)

alperozturk96:
@jancborchardt Auto upload is already designed to operate automatically — it runs under specific conditions, regardless of whether the user has the auto upload overview page open or not. The purpose of this PR is simply to make the process more transparent by informing users about the next scheduled scan, rather than hiding the logic behind it. Without this visibility, users may assume that auto upload is not functioning correctly when, in fact, it is just waiting for the right conditions.

For example, the existing logic includes checks such as delaying uploads if the battery level is below 50%, retrying after 60 minutes, or postponing for 2 hours if the battery is below 20%. Similarly, if there is no internet connection, the app will retry after 30 minutes. These are intentional, technical decisions to optimize performance and device usage — not bugs or malfunctions, but our users are not aware of it.

**What will happen if we just ignore the checks and run anyway when user open the auto upload overview page?** This is a technical decision and may bring extra load on client side and affect negatively since we have those check for reason. So I would like to get @tobiasKaminsky feedback before implementing it.

@kra-mo This information is not intended as debugging data. As explained in the linked [issue](https://github.com/nextcloud/android/issues/15718), uploads may be delayed under certain conditions, and users currently have no visibility into why. By providing this information, we give users an optional way to understand the timing of auto uploads without having to guess or assume that something is broken.

Additionally, the previous “auto upload not working” behavior was caused by [technical issues](https://github.com/nextcloud/android/issues/15214#issuecomment-3101884063)
that have now been resolved. This PR builds on those fixes to improve clarity, that auto upload continues to “just work” — while making its behavior understandable and predictable.

If we don’t implement this improvement, users who take new photos might see that uploads are delayed due to these built-in checks and mistakenly believe that auto upload has failed — an issue we’ve observed frequently in user reports. With this change, they can easily see when the next scan is scheduled. I think this is a step forward.

kra-mo:
I understand this. I think the banners and other passive and general sources of information are valuable for these.

I just also know that syncing at specified intervals is mostly a thing of the past, as far as users are concerned. It reminds me of old mail clients and such. This has been pushed "further down" so to speak, where it is now the perceived responsibility of developers to care about this, and I think this is a good thing.

What I'd much rather see is either more informational messages such as the banners (ones that the user does not need to manually access and that inform about general conditions instead of specific intervals) or something like pull-to-refresh to "fix" this from the user's side, or both.

"Low power mode is enabled so certain features may be blah blah blah" is a lot more helpful and a lot less robotic than displaying specific intervals, which the user then has to parse and make the connection themself.

tobiasKaminsky:
I am undecided, on one hand we claim it automatic, which is true, but automatic does not mean instant (this is how it was named in past and some users still understand it like that).

On the other hand designers are right to not add too many "unneeded" info to that screen.

But we want to regain trust in our system for those users that currently struggle with a reliable auto upload.
For those it can help to have it.
Once we know it is 100% reliable and no user (new and old) have truobles using it, we could remove it.

For me it is now a good trade off between info and no info, since it is shown only if user opens config screen and clicks on that icon.

I think, we also dicussed to show this "delay" icon only if there is a larger delay than usual:
Normally scan runs every 15 min -> no icon
if batter saving on/battery level <20% -> show icon to inform user about delay

kra-mo:
> I think, we also dicussed to show this "delay" icon only if there is a larger delay than usual

Something like that I would be in favor of. But in general, just providing the actual information that is relevant to the user directly instead of them having to think and make the connection. You should design for your grandma etc. etc. :)

MystikReasons:
Sorry for chiming in, but I also have an idea.
What about doing it similar to how Google does with YouTube: they have a 'Stats for nerds' button that shows additional information.
This way, it’s not too prominent, as @kra-mo wishes, but users still have access to the details whenever they want to check them.
Another advantage is that you can easily add more information in the future.

This could be shown via the kebab menu, for example as an item called 'Additional information'. At first it could be grayed out and non-clickable, and once the folder has been set up it would become active. That way, the user immediately knows that this option is only available for folders that are set up.

alperozturk96:
@kra-mo I think we should comment on the current status of the application. Otherwise, as a user, I can — and, as I have, did — point out more UI/UX issues before, and as a software engineer I could fix. I would say this is not mandatory information for users to check, or something a grandmother could reach and complain about. Since it will only be shown for active folders and if the delay is significant, like one hour. It is just optional information who willing to see.

However, considering the current workload and the size of the codebase, a complete overhaul is not feasible at the moment. I could technically remove all the checks and rewrite the system to be “100% automatic,” but that would ignore our existing constraints.

Our approach should be incremental: we improve step by step, and when the “right” time comes, we can eliminate side information or make them better.

I believe the main discussion wasn’t about the suggestion to show the “delay” icon only for longer-than-usual delays — that’s something I’m already planning to address in the final version of the PR :)

Other than this, do you have any additional design feedback based on the current status of the application for this PR ?

kra-mo:
@MystikReasons I do think that there is a way to make this info more user-friendly instead. Like statuses displaying the conditions that lead to reduced intervals. I just think that the exact number doesn't matter and that if it is the only known piece of information, the user needs additional knowledge and thinking for that to be useful.

@alperozturk96 I didn't mean to suggest to rewrite the feature. To be clear: I have no problems inherent to how it is done now, I just fear that this specific button may be more confusing than helpful. It feels very technical and I'm afraid a significant number of people wouldn't make the connection.

Couldn't we instead maybe display a status message when it takes longer, just something like "{condition (eg. 'Low battery')}, upload might take longer than expected"? And not hidden behind a button, I was quite a fan of eg. the banner in #15722.

kra-mo:
> it will only be shown for active folders and if the delay is significant, like one hour

And if that is already the case, it is nice. Sorry, I didn't get that.

alperozturk96:
@kra-mo Yes, the button itself isn’t very clear to me either, and to be honest, I’m not sure how to make it more prominent.

Screenshot 2025-10-08 at 10 22 18


This information is specific to each folder, not the entire auto-upload, so the space available in the header is limited. That's why I put the button, but what else we can use in the header for each active folder?

kra-mo:
> This information is specific to each folder, not the entire auto-upload, so the space available in the header is limited. That's why I put the button, but what else we can use?

Couldn't it be a status message in a subtitle, something like this?

Status

alperozturk96:
> > This information is specific to each folder, not the entire auto-upload, so the space available in the header is limited. That's why I put the button, but what else we can use?
> 
> Couldn't it be a status message in a subtitle, something like this?
> 
> Status

That's more user friendly, nice. I will update the PR.

alperozturk96:
@kra-mo PR is updated. You can check it from description.

alperozturk96:
Updated Screenshot @kra-mo 


<table>
  <tr>
    <td>Image 1</td>
  </tr>
</table>

Issue #15733, opened 2025-10-01
l10n: Remove placeholder strings from Transifex


rakekniven:
https://github.com/nextcloud/android/blob/3e811defc564fffcea2272c98a8013f1b2458808/app/src/main/res/values/strings.xml#L454-L460

Reported by translators

rakekniven:
cc @tobiasKaminsky

PR #15750, opened 2025-10-07
make placeholders non translatable

alperozturk96:

### Issue: https://github.com/nextcloud/android/issues/15733

Removes placeholder strings from translations.
Adds other hardcoded placeholders into strings.xml as non translatable.

### DIFF
--- a/app/src/main/res/layout/account_item.xml
+++ b/app/src/main/res/layout/account_item.xml
@@ -93,7 +93,7 @@
                 android:text="@string/placeholder_sentence"
                 android:textColor="?android:attr/textColorSecondary"
                 android:visibility="gone"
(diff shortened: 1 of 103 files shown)


AndyScherzinger:
Translated files would be updated via transifex so these changes could be removed making the PR a lot smaller.

Change is fine but you'd loose rtl language translation of placeholders where it has been nice to check the layout in the idea directly without the need for launching an emulator with an rtl language

alperozturk96:
@AndyScherzinger Thank you for the suggestions. We can add the RTL non-translatable placeholders in a separate PR. The failing tests and increased lint are unrelated, and since this PR only changes `tools:text`, I’ll proceed with merging it now to avoid potential conflicts.

PR #15754, opened 2025-10-08
feat: add rtl placeholders

alperozturk96:

<table>
  <tr>
    <th>Default</th>
    <th>RTL</th>
  </tr>
  <tr>
    <td>before</td>
    <td>after</td>
  </tr>
</table>


AndyScherzinger:
Afaik this won't work since transifex sync will override any file except the main-file from `values` so you cannot add any data to any other language without it being overridden by the sync bot....

PR #15759, opened 2025-10-09
fix: auto upload db state

alperozturk96:

### Changes

Correct upload status handling: Previously, when auto-upload was triggered, all files (e.g., 500 files) were initially marked as `IN_PROGRESS` but their final status (`SUCCESSFUL` or `FAILED`) was not consistently updated based on the outcome of the upload operation. This PR ensures that each upload entity reflects the correct final status.

Batch upload support: Uploading large numbers of files (2,000–10,000) previously caused significant load on the client. This PR introduces batch processing to reduce memory usage and improve performance.

Improved data access: Replaced `ContentValues` operations with proper `DAO` methods, ensuring safer, more maintainable database interactions.

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
@@ -14,14 +14,9 @@
  */
 package com.owncloud.android.datamodel;
 
-import android.content.ContentProviderOperation;
-import android.content.ContentProviderResult;
 import android.content.ContentResolver;
 import android.content.ContentValues;
-import android.content.OperationApplicationException;
 import android.database.Cursor;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -94,97 +89,6 @@ private void initOCCapability() {
         }
     }
 
-    /**
-     * Stores an upload object in DB.
-     *
-     * @param ocUpload Upload object to store
-     * @return upload id, -1 if the insert process fails.
-     */
-    public long storeUpload(OCUpload ocUpload) {
```
alperozturk96, line 103: No need of these `ContentValues` functions, since `UploadDao` doing the same job.

PR #15775, opened 2025-10-11
fix: use accountName only if User present

### LINKED ISSUE #15771: crash on Account login
### ⚠️ Before posting ⚠️

### Steps to reproduce

New install on Android 16, pixel 9a
Login to my own server in home network, crash

### Expected behaviour

No crash

### Cause of error
```java
Exception in thread "main" java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.String com.nextcloud.client.account.User.getAccountName()' on a null object reference
    at com.owncloud.android.ui.activity.FileDisplayActivity.onStart(FileDisplayActivity.kt:2784)
    at android.app.Instrumentation.callActivityOnStart(Instrumentation.java:1699)
    at android.app.Activity.performStart(Activity.java:9287)
    at android.app.ActivityThread.handleStartActivity(ActivityThread.java:4344)
    at android.app.servertransaction.TransactionExecutor.performLifecycleSequence(TransactionExecutor.java:214)
    at android.app.servertransaction.TransactionExecutor.cycleToPath(TransactionExecutor.java:194)
    at android.app.servertransaction.TransactionExecutor.cycleToPath(TransactionExecutor.java:176)
    at android.app.servertransaction.TransactionExecutor.executeNonLifecycleItem(TransactionExecutor.java:129)
    at android.app.servertransaction.TransactionExecutor.executeTransactionItems(TransactionExecutor.java:103)
    at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:80)
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2830)
    at android.os.Handler.dispatchMessage(Handler.java:110)
    at android.os.Looper.dispatchMessage(Looper.java:315)
    at android.os.Looper.loopOnce(Looper.java:251)
    at android.os.Looper.loop(Looper.java:349)
    at android.app.ActivityThread.main(ActivityThread.java:9041)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:593)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:929)

```

### App information
* ID: `com.nextcloud.client`
* Version: `30340090`
* Build flavor: `generic`

### Device information
* Brand: `google`
* Device: `tegu`
* Model: `Pixel 9a`
* Id: `BP3A.250905.014`
* Product: `tegu`

### Firmware
* SDK: `36`
* Release: `16`
* Incremental: `13873947`

### Android version

16

### Device brand and model

pixel 9a

### Stock or custom OS?

Stock

3.34.0

### Nextcloud server version

31.0.8

### Using a reverse proxy?

No

### Android logs

_No response_

### Server error logs

```bash

```

### Additional information

_No response_

joshtrichards:
Should fix #15771

Despite appearances, basically a one like change. The rest is just indentation change because the early return eliminates the need to have the rest inside the replaced conditional.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
@@ -2753,53 +2753,54 @@ class FileDisplayActivity :
         super.onStart()
         val optionalUser = user
         val storageManager = getStorageManager()
-        if (optionalUser.isPresent && storageManager != null) {
-            /** Check whether the 'main' OCFile handled by the Activity is contained in the */
-            // current Account
-            var file = getFile()
-            // get parent from path
-            if (file != null) {
-                if (file.isDown && file.lastSyncDateForProperties == 0L) {
-                    // upload in progress - right now, files are not inserted in the local
-                    // cache until the upload is successful get parent from path
-                    val parentPath =
-                        file.remotePath.substring(0, file.remotePath.lastIndexOf(file.fileName))
-                    if (storageManager.getFileByPath(parentPath) == null) {
-                        file = null // not able to know the directory where the file is uploading
-                    }
-                } else {
-                    file = storageManager.getFileByPath(file.remotePath)
-                    // currentDir = null if not in the current Account
-                }
-            }
-            if (file == null) {
-                // fall back to root folder
-                file = storageManager.getFileByPath(OCFile.ROOT_PATH) // never returns null
-            }
-            setFile(file)
 
-            val user = optionalUser.get()
-            setupDrawer()
+        if (!optionalUser.isPresent || storageManager == null) return
 
-            mSwitchAccountButton.tag = user.accountName
-            DisplayUtils.setAvatar(
-                user,
-                this,
-                getResources().getDimension(R.dimen.nav_drawer_menu_avatar_radius),
-                getResources(),
-                mSwitchAccountButton,
-                this
-            )
-            val userChanged = (user.accountName != lastDisplayedAccountName)
-            if (userChanged) {
-                Log_OC.d(TAG, "Initializing Fragments in onAccountChanged..")
-                initFragments()
-                if (file.isFolder && TextUtils.isEmpty(searchQuery)) {
-                    startSyncFolderOperation(file, false)
+        /** Check whether the 'main' OCFile handled by the Activity is contained in the */
+        // current Account
+        var file = getFile()
+        // get parent from path
+        if (file != null) {
+            if (file.isDown && file.lastSyncDateForProperties == 0L) {
+                // upload in progress - right now, files are not inserted in the local
+                // cache until the upload is successful get parent from path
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -2732,54 +2732,55 @@ class FileDisplayActivity :
...
-            }
-            if (file == null) {
-                // fall back to root folder
-                file = storageManager.getFileByPath(OCFile.ROOT_PATH) // never returns null
-            }
-            setFile(file)
 
-            val user = optionalUser.get()
-            setupDrawer()
+        if (!optionalUser.isPresent || storageManager == null) return
```
alperozturk96, line 2757: Hey. Thank you for the PR.

`if (optionalUser.isPresent && storageManager != null) {`
`if (!optionalUser.isPresent || storageManager == null) return`

These two conditions are logically equivalent — they express the same control flow in opposite forms. Therefore, changing one to the other wouldn’t actually resolve the issue.

The real problem likely occurs when the app is suspended and then resumed, during which the state (such as `optionalUser` or `storageManager`) is lost or not properly restored.

joshtrichards: > These two conditions are logically equivalent — they express the same control flow in opposite forms. Therefore, changing one to the other wouldn’t actually resolve the issue.

Switching the conditions permits the early return, which avoids running the later NPE code path though, right? (which in the existing code happens outside the conditional)

> The real problem likely occurs when the app is suspended and then resumed, during which the state (such as optionalUser or storageManager) is lost or not properly restored.

I can't argue there. Admittedly did not go down that rabbit hole. 

I guess the path forward comes down to two questions:

- Is the true underlying problem (suspend/resume state) readily fixable in the near future?
- What's the impact (cons specifically) of returning w/o at least crashing (as proposed in this PR)? I.e., Are we likely introducing new problems or just working around one until we can identify and fix the state matter?

I don't have the answers. Just posing to think about these a bit myself. :) To answer I'd have to take a deeper dive, which admittedly I haven't on this matter.

alperozturk96: This PR doesn’t seem likely to introduce any new issues, so we can go ahead and merge it. It might even help by returning early instead of failing within the condition.

As for the actual fix, it’s not an easy issue to reproduce or verify because of the tightly coupled, nested fragment hierarchy and the complexity of tracking their lifecycles. The `NPE` seems related to a specific execution flow — in theory, if the optional user is present, the user object should be accessible, but it appears there’s a lifecycle-related issue causing the null reference.

I’ll investigate this further, but thank you for the PR 🙏

alperozturk96:
Since `FDA.initFile()` checking user existence in the beginning of the function we can close this PR.

PR #15806, opened 2025-10-17
use dao in upload list adapter

alperozturk96:

### Changes

Uses `DAO` functions instead of `ContentResolvers`

Fixes UI state handling in the uploads list: since `UploadGroupLoadRunnable` performs asynchronous operations, we now wait for all upload groups to complete their refresh before updating the UI. So that list reflects the latest data and avoids inconsistent results.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
@@ -8,6 +8,7 @@
  */
 package com.owncloud.android.ui.adapter;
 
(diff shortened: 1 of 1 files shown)


mhzawadi:
I have both the stable and this QA version installed, is that safe or even recomended?

- If its ok to have both, I may have a bug.
- If not recommended, I need to remove the stable version and see if that sorts my issue

TL;DR: re-upload of 5000+ images for a second time

jkufner:
See https://github.com/nextcloud/android/issues/15812#issuecomment-3429308183

PR #15816, opened 2025-10-20
Drone: test with NC32

tobiasKaminsky:


PhilLab:
@tobiasKaminsky I just realized that Drone is still testing against v30 for "stable" and then found this PR here. Maybe it could be rejuvenated?

PR #15819, opened 2025-10-20
fix: empty folder appearance

alperozturk96:

### Issue

Since `MetadataWorker` updates the `eTag` newly added files not determined by client any more, thus empty folder may appear.

https://github.com/user-attachments/assets/88348207-ae78-4615-9b31-e5a9b9f4fb52

### Demo

Screenshot 2025-10-22 at 15 04 30

### Changes

Only updates e-Tag after successful `RefreshFolderOperation` otherwise resets e-Tag.
Only stores new e-Tag after saving child files to storage.

### How to test?

1. Create sub folder in root directory
2. Add one file from web
3. Check metadata worker
4. Create another sub folder from web
5. Add big files to the first sub-folder
6. Open app and navigate to the newly created subfolder
7. Navigate back to the root (so that we can trigger worker to cancel previously unfinished job)
8. Navigate to the newly created subfolder again
9. Navigate back to the root
10. Check new files existence from DB or navigate to the first subfolder

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/metadata/MetadataWorker.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/metadata/MetadataWorker.kt
@@ -55,7 +55,7 @@ class MetadataWorker(private val context: Context, params: WorkerParameters, pri
         return Result.success()
     }
 
-    @Suppress("DEPRECATION")
+    @Suppress("DEPRECATION", "TooGenericExceptionCaught")
     private suspend fun refreshFolder(folder: OCFile, storageManager: FileDataStorageManager) =
         withContext(Dispatchers.IO) {
             Log_OC.d(
@@ -72,12 +72,23 @@ class MetadataWorker(private val context: Context, params: WorkerParameters, pri
 
             Log_OC.d(TAG, "⏳ Fetching metadata for: ${folder.remotePath}")
 
-            val operation = RefreshFolderOperation(folder, storageManager, user, context)
-            val result = operation.execute(user, context)
-            if (result.isSuccess) {
-                Log_OC.d(TAG, "✅ Successfully fetched metadata for: ${folder.remotePath}")
-            } else {
-                Log_OC.e(TAG, "❌ Failed to fetch metadata for: ${folder.remotePath}")
+            try {
+                val operation = RefreshFolderOperation(folder, storageManager, user, context)
+                val result = operation.execute(user, context)
+                if (result.isSuccess) {
+                    Log_OC.d(TAG, "✅ Successfully fetched metadata for: ${folder.remotePath}")
+                } else {
+                    Log_OC.e(TAG, "❌ Failed to fetch metadata for: ${folder.remotePath}, resetting e-Tag")
+                    resetETag(folder, storageManager)
+                }
+            } catch (e: Exception) {
+                Log_OC.e(TAG, "❌ Exception during to fetch metadata for: ${folder.remotePath}, resetting e-Tag $e")
+                resetETag(folder, storageManager)
             }
         }
+
+    private fun resetETag(folder: OCFile, storageManager: FileDataStorageManager) {
+        folder.etag = ""
+        storageManager.saveFile(folder)
+    }
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -411,10 +411,11 @@ private RemoteOperationResult checkForChanges(OwnCloudClient client) {
         Log_OC.d(TAG, "Checking changes in " + user.getAccountName() + remotePath);
 
         // remote request
-        result = new ReadFileRemoteOperation(remotePath).execute(client);
+        result = new ReadFolderRemoteOperation(remotePath).execute(client);
```
alperozturk96: @tobiasKaminsky Is there any way to make it faster? Since it is reading folder instead of file.

app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -428,6 +429,14 @@ private RemoteOperationResult checkForChanges(OwnCloudClient client) {
                             "  Remote eTag: " + remoteFolderETag + "\n" +
                             "  Changed:     " + mRemoteFolderChanged
                             );
+
+                    // result.getData() returns folder itself with child files
+                    final int remoteChildFileSize = childFiles.size() - 1;
+                    final int childFileSize = fileDataStorageManager.fileDao.getChildFileCount(mLocalFolder.getFileId());
+                    if (childFileSize != remoteChildFileSize) {
```
alperozturk96: @tobiasKaminsky How can we compare better local and remote folder since `eTag` is not enough and up-to-date most of the time because of the `MetadataWorker`?

If we insert child files via `MetadataWorker` it will impact the performance as well and we can't compare file sizes since we have correct file size for the folder itself.

tobiasKaminsky: Please solve it via MetadataWorker: all subfolders on next hierarchy level needs to be updated and only then the etag can be changed/stored.

tobiasKaminsky (dismissed):

app/src/main/java/com/nextcloud/client/jobs/metadata/MetadataWorker.kt
```diff
@@ -72,12 +72,23 @@ class MetadataWorker(private val context: Context, params: WorkerParameters, pri
...
-            } else {
-                Log_OC.e(TAG, "❌ Failed to fetch metadata for: ${folder.remotePath}")
+            try {
+                val operation = RefreshFolderOperation(folder, storageManager, user, context)
+                val result = operation.execute(user, context)
+                if (result.isSuccess) {
+                    Log_OC.d(TAG, "✅ Successfully fetched metadata for: ${folder.remotePath}")
+                } else {
+                    Log_OC.e(TAG, "❌ Failed to fetch metadata for: ${folder.remotePath}, resetting e-Tag")
+                    resetETag(folder, storageManager)
```
alperozturk96, line 82: We can safely reset the e-Tag if the `RefreshFolderOperation` fails. Since the operation is only executed when the e-Tag differs, setting it back to null in the event of a failure is acceptable and will not trigger any unnecessary additional operations.

app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -634,8 +629,13 @@ private void synchronizeData(List<Object> folderAndFiles) {
                                            mLocalFolder);
         }
         fileDataStorageManager.saveFolder(remoteFolder, updatedFiles, localFilesMap.values());
-
         mChildren = updatedFiles;
+
+        // update eTag
+        mLocalFolder.setEtag(remoteFolder.getEtag());
```
alperozturk96, line 635: Only update `mLocalFolder` e-Tag and file-length after saving all children files to the directory. 

Later on in `run` `fileDataStorageManager.saveFile(mLocalFolder);` will be called after `fetchAndSyncRemoteFolder`

tobiasKaminsky:
@alperozturk96  as discussed please double check via more logs / throwing exceptions

Issue #14681, opened 2025-03-14
Upload Failure with "No Internet Connection" Message After Upgrading to Version 3.31.0 on Android App


Grumbly4308:
### ⚠️ Before posting ⚠️

- [x] This is a **bug**, not a question or an enhancement.
- [x] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [x] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

After upgrading the app from version 3.30.8 to 3.31.0, I am experiencing an issue where attempting to upload a file results in an error message stating, "Upload failed, No internet connection." This problem does not occur when using the previous version (3.30.8), indicating that the upgrade may have introduced this bug.

### Expected behaviour

The app should allow users to upload files without showing an incorrect error message when an active internet connection is available.

### Actual behaviour

Upon attempting a file upload in version 3.31.0, users receive an erroneous "No internet connection" error despite having a stable internet connection.

### Android version

15

### Device brand and model

Pixel 8a 

### Stock or custom OS?

Custom (explain in "additional information")

### Nextcloud android app version

3.30.0

### Nextcloud server version

Nextcloud Hub 10 (31.0.0)

### Using a reverse proxy?

Yes

### Server error logs

```bash

```

### Additional information

GraphineOS

tsloms:
Same/similar issue here.

Exactly the same error message in app.

Nextcloud Hub 9 (30.0.6) Server
Android app version 3.31.0

sriperinkulam:
Same issue here. 
Android App from Google Play - v3.31.0
Nextcloud version	30.0.7.2

Phone isn't in airplane mode and has an active wifi connection.

ADB logs
`03-14 16:52:51.708 10741 10741 D WM-WorkerWrapper: Starting work for com.nextcloud.client.jobs.upload.FileUploadWorker
03-14 16:52:51.708 10741 10741 D WM-SystemJobService: onStartJob for WorkGenerationalId(workSpecId=2d2f9bf9-0d25-446e-8f2e-cff768dc6899, generation=0)
03-14 16:52:51.709 10741 10802 D WM-Processor: Work WorkGenerationalId(workSpecId=2d2f9bf9-0d25-446e-8f2e-cff768dc6899, generation=0) is already enqueued for processing
[...]

PR #15820, opened 2025-10-20
fix: connectivity service impl

alperozturk96:

### Fixes: https://github.com/nextcloud/android/issues/14681

Deduplicates `isNetworkAndServerAvailable` logic

Removes `c.isWifi() && !c.isMetered()` from internet walled logic since `isConnected` already covers local internet and internet access

Implements `ConnectivityManager.NetworkCallback` for better network state listening

Removes deprecated logics

Adds better documentation

### DIFF
--- a/app/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
+++ b/app/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
@@ -138,9 +138,13 @@ public class ConnectivityServiceImpl implements ConnectivityService {
         Server server = accountManager.getUser().getServer();
         String baseServerAddress = server.getUri().toString();
 
-        if (!currentConnectivity.isConnected() || baseServerAddress.isEmpty()) {
-            walledCheckCache.setValue(true);
-            return true;
+        if (!currentConnectivity.isConnected()
+            || baseServerAddress.isEmpty() ||
+            !currentConnectivity.isWifi() ||
+            currentConnectivity.isMetered()) {
+            final var result = !currentConnectivity.isConnected();
+            walledCheckCache.setValue(result);
+            return result;
         }
 
         boolean isWalled;


app/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
```diff
@@ -109,77 +127,55 @@ public boolean isInternetWalled() {
         final Boolean cachedValue = walledCheckCache.getValue();
         if (cachedValue != null) {
             return cachedValue;
-        } else {
-            Server server = accountManager.getUser().getServer();
-            String baseServerAddress = server.getUri().toString();
-
-            boolean result;
-            Connectivity c = getConnectivity();
-            if (c != null && c.isConnected() && c.isWifi() && !c.isMetered() && !baseServerAddress.isEmpty()) {
```
alperozturk96, line 118: Why is `getConnectivity()` used here, and why does it specifically require a Wi-Fi connection and not metered one? If the user has a connection, wouldn’t it be possible to simply check whether it is walled or not? @tobiasKaminsky 

And even further no connection check needed at all we can just try to fetch if It's failed it's walled that's it.

app/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
```diff
@@ -109,77 +127,55 @@ public boolean isInternetWalled() {
...
+        } catch (Exception e) {
+            Log_OC.e(TAG, "Exception while checking internet walled state", e);
+            walledCheckCache.setValue(true);
+            return true;
+        } finally {
+            get.releaseConnection();
         }
     }
 
-    private boolean hasNonCellularConnectivity() {
```
alperozturk96, line 176: Deprecated

app/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
```diff
@@ -109,77 +127,55 @@ public boolean isInternetWalled() {
...
         }
-    }
 
-    private boolean isNetworkMetered() {
-        final Network network = platformConnectivityManager.getActiveNetwork();
+        final GetMethod get = requestBuilder.invoke(baseServerAddress + CONNECTIVITY_CHECK_ROUTE);
         try {
-            NetworkCapabilities networkCapabilities = platformConnectivityManager.getNetworkCapabilities(network);
-            if (networkCapabilities != null) {
-                return !networkCapabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_NOT_RESTRICTED);
```
alperozturk96, line 166: `NET_CAPABILITY_NOT_METERED` should be used instead of `NET_CAPABILITY_NOT_RESTRICTED`

app/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
```diff
@@ -62,43 +64,59 @@ public GetMethod invoke(String url) {
...
-            actNw.hasTransport(NetworkCapabilities.TRANSPORT_CELLULAR) ||
-            actNw.hasTransport(NetworkCapabilities.TRANSPORT_ETHERNET) ||
-            actNw.hasTransport(NetworkCapabilities.TRANSPORT_VPN) ||
-            actNw.hasTransport(NetworkCapabilities.TRANSPORT_BLUETOOTH) ||
-            actNw.hasTransport(NetworkCapabilities.TRANSPORT_WIFI_AWARE)) {
-            return true;
-        }
+        // Verify that the network both claims to provide Internet
+        // and has been validated (i.e., Internet is actually reachable).
+        if (actNw.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET) && actNw.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)) {
```
alperozturk96: First we must check do we actually have access to public internet then we can filter result via transport types.

tobiasKaminsky: Intranet must not be needed, since we can also support air-gapped solutions.

app/src/main/java/com/nextcloud/client/network/ConnectivityServiceImpl.java
```diff
@@ -25,161 +24,154 @@
...
+
+        if (!currentConnectivity.isConnected() || baseServerAddress.isEmpty()) {
+            walledCheckCache.setValue(true);
+            return true;
         }
-    }
 
-    private boolean isNetworkMetered() {
-        final Network network = platformConnectivityManager.getActiveNetwork();
+        boolean isWalled;
```
alperozturk96, line 150: Add metered wifi check here...

alperozturk96:
replaced via https://github.com/nextcloud/android/pull/16232

PR #15869, opened 2025-10-30
Simplified for loop

alexknop:
We were using var5,var6, and var7 unnecessarily to iterate through children

Solution:

Simplify for loop


alperozturk96:
Since `walkFileTreeRandomly` replaced with `AutoUploadHelper`, closing this PR.

PR #15872, opened 2025-10-30
fix: choose correct upload ids for retry

alperozturk96:

### Changes

Upload conditions are validated before scheduling jobs, ensuring that only eligible uploads are sent to the worker.

Fixes an unnecessary chained call in `UploadListActivity` and refreshes the UI after worker completed.

Removes `startAutoUploadImmediately` from `UploadListActivity`, as auto uploads are already managed by a periodic job. Manually triggering them provides no benefit, if a previous scan was recently completed, it will simply be skipped.
Additionally, `retryFailedUploads` already handles retrying failed auto uploads.

Handles result of `retryFailedUploads` in `UploadListActivity`.

### Demo

Screenshot 2025-10-30 at 09 39 29

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadHelper.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadHelper.kt
@@ -88,18 +88,42 @@ class FileUploadHelper {
         fun buildRemoteName(accountName: String, remotePath: String): String = accountName + remotePath
     }
 
+    /**
+     * Retries all failed uploads across all user accounts.
+     *
+     * This function retrieves all uploads with the status [UploadStatus.UPLOAD_FAILED], including both
+     * manual uploads and auto uploads. It runs in a background thread (Dispatcher.IO) and ensures
+     * that only one retry operation runs at a time by using a semaphore to prevent concurrent execution.
+     *
+     * Once the failed uploads are retrieved, it calls [retryUploads], which triggers the corresponding
+     * upload workers for each failed upload.
+     *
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadHelper.kt
```diff
@@ -88,32 +88,39 @@ class FileUploadHelper {
         fun buildRemoteName(accountName: String, remotePath: String): String = accountName + remotePath
     }
 
+    fun getUploadsByStatus(accountName: String, status: UploadStatus, onCompleted: (Array<OCUpload>) -> Unit) {
+        ioScope.launch {
+            val result = uploadsStorageManager.uploadDao.getUploadsByStatus(
+                accountName,
```
tobiasKaminsky: Please remove the account dependency

tobiasKaminsky (dismissed):
see inline

zc-devs:
> Manually triggering them provides no benefit, if a previous scan was recently completed, it will simply be skipped

Do you mean this check?

https://github.com/nextcloud/android/blob/7fd5807fa94d2f39a4eb6d71ebdc84b0156acbfd/app/src/main/java/com/nextcloud/client/jobs/FilesSyncWork.kt#L179-L197

Shouldn't `startAutoUploadImmediately` be kept in `UploadListActivity` and `FilesSyncWork.canExitEarly()` not check scan interval at all in case of manual triggering?

https://github.com/zc-devs/nextcloud-android/commit/99847fd57c08e2af879767405c12fdd7682df556

PR #15874, opened 2025-10-30
fix: handle skip name collision policy

alperozturk96:

### Issue

User can choose different collision policy and this needs to be handled for both auto upload and normal upload.

For skip policy, the upload needs to marks as SUCCESSFUL so that it won't repeated. Additionally, user should be able to see in the `UploadList`.

### Demo

<table>
  <tr>
    <td>Image 1</td>
  </tr>
</table>

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/UploadListAdapter.java
@@ -223,7 +223,7 @@ public class UploadListAdapter extends SectionedRecyclerViewAdapter<SectionedVie
 
         headerViewHolder.binding.uploadListAction.setOnClickListener(v -> {
             switch (group.type) {
-                case CURRENT -> new Thread(() -> {
+                case CURRENT -> {
                     OCUpload ocUpload = group.getItem(0);
                     if (ocUpload == null) {
                         return;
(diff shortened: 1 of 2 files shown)


alperozturk96:
@kra-mo Could you check the screenshot and share your opinion for the title and reason text for the `SKIP` policy?

If the user wants to skip upload an existing file, what should be the title and description?

kra-mo:
> @kra-mo Could you check the screenshot and share your opinion for the title and reason text for the `SKIP` policy?

"Skipped" makes sense from the brief description and demo, but I think I would need a better explanation of what this feature is :)

> If the user wants to skip upload an existing file, what should be the title and description?

Where would this be seen exactly? And why exactly would it occur?

alperozturk96:
> > @kra-mo Could you check the screenshot and share your opinion for the title and reason text for the `SKIP` policy?
> 
> "Skipped" makes sense from the brief description and demo, but I think I would need a better explanation of what this feature is :)
> 
> > If the user wants to skip upload an existing file, what should be the title and description?
> 
> Where would this be seen exactly? And why exactly would it occur?

When a user enables auto upload for a folder, they can configure a name collision policy, which determines how the app handles files that already exist on the server with the same name.

Possible behaviors:

**Overwrite:** Replace the existing file on the server with the new local version.
**Skip:** Do not upload the file if a file with the same name already exists on the server.
**Ask User:** Prompt the user with a conflict resolution dialog to decide what to do for each conflicting file.

For example, suppose a user has set up auto uploads from a local folder to a corresponding folder on the server, and 50 of those files already exist on the server with the same names. If the user previously selected the “Skip” policy for this folder, all 50 conflicting files will be skipped during upload, no files will be overwritten or uploaded in their place.

User can configure name collision policy for auto upload from Settings → Auto upload.

kra-mo:
I see. For the title then, Skip makes sense. For the description, something like:

Don't overwrite existing files with the same name.

kra-mo:
Unless you mean the text in the screenshot? The reason why it was skipped? Then:

A file with the same name already exists.

PR #15895, opened 2025-11-03
fix: dont ask storage permission again if user chooses limited file access

alperozturk96:

### How to reproduce?

1. Click on plus button from bottom right
2. Select Upload Files
3. Storage permission dialog will be shown and select the limited access.
4. Tap on upload files again, it will ask again the permission.


alperozturk96:
replaced with https://github.com/nextcloud/android/pull/15977

PR #15899, opened 2025-11-04
local videos cannot be used for preview

tobiasKaminsky:
How to test
- upload a video
- go directly to "media"
- thus no preview is generated within file listing
- wait for upload
- see that now video thumbnail is fetched from server
- Before: it tried to generate locally a video preview, which failed and thus only placeholder was shown


alperozturk96 (changes requested):
I couldn't able to see thumbnail for the videos in the media tab?

https://github.com/user-attachments/assets/1c61ff43-4059-4ccb-834c-bbd993ea3ef1

PR #15903, opened 2025-11-05
chore: suppress notify data set changed lint

alperozturk96:

For each `notifyDataSetChanged` logic needs to be checked and replaced with `notifyItemChanged`. Until then, we can reset `notifyDataSetChanged` lint count.

Screenshot 2025-11-05 at 09 10 17


alperozturk96:
replaced with https://github.com/nextcloud/android/pull/15911

PR #15921, opened 2025-11-06
fix(i18n): Plural wasn't working on i18n system

rakekniven:
Reported at Transifex

Singular is not present at Tx.


rakekniven:
No backports necessary?

PR #15923, opened 2025-11-07
fix: observe worker state

alperozturk96:

### Issue

Local file indicator icons were not updating correctly when files or folders finished downloading, and switching between list and grid layouts caused icons to disappear.

### Fixes

Fix observation of multiple workers across all relevant screens to ensure each worker completion is reflected in the UI.

Preserves local file indicator state during layout switches (list ↔ grid).

Fix logic for both folder and file downloads to display local file indicator icons consistently.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListDelegate.kt
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListDelegate.kt
@@ -193,12 +193,7 @@ class OCFileListDelegate(
     }
 
     @Suppress("MagicNumber")
-    fun bindViewHolder(
-        viewHolder: ListViewHolder,
-        file: OCFile,
-        currentDirectory: OCFile?,
-        searchType: SearchType?
-    ) {
+    fun bindViewHolder(viewHolder: ListViewHolder, file: OCFile, currentDirectory: OCFile?, searchType: SearchType?) {
         // thumbnail
         viewHolder.imageFileName?.text = file.fileName
         viewHolder.thumbnail.tag = file.fileId


alperozturk96:
@tobiasKaminsky 

### Note: 

Events are delivered only to the currently active activity.

For example, this means that when the user is in `FileDisplayActivity`. `ManageAccountsActivity`, cannot receive an event.

-----

FolderDownloadCompleted is used:

* In **FileDisplayActivity** to mark the green tick for a single OCFile item via `notifyItemChanged`, indicating that the folder has been downloaded.

FileDownloadStarted is used:

* In **FileDisplayActivity**, if `mWaitingToPreview` is not null, to request the download again if it has not already started.
* In **ManageAccountsActivity**, when the user logs out, to cancel all current downloads for the active account.
* In **PreviewImageActivity**, if the screen state is `PreviewImageActivityState.WaitingForBinder`, to select the appropriate page in the ViewPager.

FileDownloadCompleted is used:

* In **FileDisplayActivity** to preview an image immediately after the download completes, but only if `fileIDForImmediatePreview` is valid.
* In **PreviewImageActivity**, if the screen state is `PreviewImageActivityState.Edit`, to start the image editing activity; otherwise, it reinitializes the ViewPager for the newly downloaded file.

FileUploadStarted is used:

* In **UploadListActivity** (the Upload screen) to refresh the items in the list.
* In **FileDetailFragment** to update the text view and inform the user that the file is uploading.

FileUploadCompleted is used:

* In **UploadListActivity** (the Upload screen) to refresh the items in the list.
* In **FileDisplayActivity** to insert the related item via `notifyItemInserted`.

OfflineOperationsCompleted is used:

* In **FileDisplayActivity** to refresh the current directory and reload the OCFile list adapter.

alperozturk96:
@tobiasKaminsky 

**How to Test the Green Tick for a Folder & Slow Grid View**

1. Ensure the folder contains at least a few files.
2. Open the three-dot menu for that folder.
3. Tap **"Sync"**.
4. The folder will be downloaded via **FolderDownloadWorker**.
5. The green tick icon is then displayed through `OCFileListDelegate.showLocalFileIndicator`.

   * This method uses a Coroutine IO scope to call `isFolderFullyDownloaded`, which queries the database asynchronously.
   * Other icon checks (e.g., cancel, syncing states) can be handled synchronously, but the folder-completion check cannot because it is suspended.

**Why Some Approaches Do Not Work**

1. Setting synchronous icons first, then calling the async check. This shows the quick icons immediately, but the green tick never appears afterward because the adapter is not notified when the async result returns.

2. Calling `isFolderFullyDownloaded` asynchronously from `OCFileListFragment` or `FileDisplayActivity`, then using `notifyItemChanged`

3. Running the check on an `ExecutorService` or a Thread. This does not change the behavior, since the core issue is the async nature of the database call and the lack of a proper UI update mechanism tied to it.

4. Switching from `setAdapter` to `swapAdapter` in the switchLayoutManager function. While this make UI faster but changing the adapter-swap logic breaks item alignment, styles, and layout behaviors.

tobiasKaminsky:
To reproduce:
- go into a folder wtih downloaded files

tobiasKaminsky:
I debugged it and I see that we set visibility of the "container" "@+id/file_features_layout" only after setting the sync state.
This is very specific to the grid view:
https://github.com/nextcloud/android/blob/8ba9015281eff792bc1f86daf5ead009bee13532/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java#L498

If I always set this "container" to be visible, then there is no delay.

alperozturk96:
> I debugged it and I see that we set visibility of the "container" "@+id/file_features_layout" only after setting the sync state. This is very specific to the grid view:
> 
> https://github.com/nextcloud/android/blob/8ba9015281eff792bc1f86daf5ead009bee13532/app/src/main/java/com/owncloud/android/ui/adapter/OCFileListAdapter.java#L498
> 
> If I always set this "container" to be visible, then there is no delay.

Thank you awesome 👍

PR #15925, opened 2025-11-08
Move or Copy files: Start in the parent folder instead of root

PhilLab:

Previous behavior always moved up to the root folder, which meant the user
lost the context of their current files. Especially in complicated nested
folder structures, this made standard operations overly complicated, like
duplicating in the same folder or duplicating/moving to a sibling folder.
    
This is now also in sync with the behavior of the iOS app and the web UI

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
@@ -276,7 +276,7 @@ open class FolderPickerActivity :
         super.onResume()
         Log_OC.e(TAG, "onResume() start")
 
-        val extraFolder = intent.getParcelableArgument(EXTRA_FOLDER.toString(), OCFile::class.java)
+        val extraFolder = intent.getParcelableArgument(EXTRA_FOLDER, OCFile::class.java)
         if (extraFolder != null) {
             file = extraFolder
         } else {
@@ -658,8 +658,7 @@ open class FolderPickerActivity :
     }
 
     companion object {
-        @JvmField
-        val EXTRA_FOLDER = FolderPickerActivity::class.java.canonicalName?.plus(".EXTRA_FOLDER")
+        const val EXTRA_FOLDER = "com.owncloud.android.ui.activity.FolderPickerActivity".plus(".EXTRA_FOLDER")
 
         @JvmField
         @Deprecated(


app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
```diff
@@ -253,7 +253,7 @@ open class ExtendedListFragment :
             is FileDisplayActivity -> {
                 if (isBackPressed && query.isEmpty()) {
                     activity.resetSearchView()
-                    activity.updateListOfFilesFragment(true)
+                    activity.updateListOfFilesFragment()
```
PhilLab, line 257: Note to reviewer: This is the only instance where the parameter was `true`. It didn't have any effect anymore - but maybe something specific _was_ desired here and got lost over time?

PhilLab:
Let me know if a ticket shall be created to track the behavior change

app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
PhilLab: @alperozturk96 what do you think about this PR? Please review and let me know.

alperozturk96: @tobiasKaminsky Make sense to me, what do you think about new behavior?

PhilLab: @tobiasKaminsky can you take a look? With this change, the Android app finally behaves like the iOS app and the web app behaves in that aspect.

app/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
```diff
@@ -275,8 +276,13 @@ open class FolderPickerActivity :
         super.onResume()
         Log_OC.e(TAG, "onResume() start")
 
-        refreshListOfFilesFragment(false)
-        file = listOfFilesFragment?.currentFile
+        val extraFolder = intent.getParcelableArgument(EXTRA_FOLDER.toString(), OCFile::class.java)
```
alperozturk96: Could you replace this logic with the function below or something similar?

Since `EXTRA_FOLDER.toString()` returns `"null"` when `EXTRA_FOLDER` itself is null, it’s better to check whether the argument actually exists. If it doesn’t, we should fall back to `listOfFilesFragment?.currentFile`.

```
private fun initFileFromArg() {
        val fileArg: String? = EXTRA_FOLDER
        val extraFolder = fileArg?.let { intent.getParcelableArgument(it, OCFile::class.java) }
        file = extraFolder ?: listOfFilesFragment?.currentFile
    }
```

---

Other than that, the PR looks good. It opens the current directory for move or copy operations, just like the web version. The PR title says “Start in the parent folder instead of root,” but when I tested it, it opened the current directory of the selected file. I also tested this on the web version, and it behaves the same, opening the current directory and allowing the user to navigate back if needed.

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/ui/activity/FolderPickerActivity.kt
PhilLab: @alperozturk96 I phrased the PR title like this because I meant that it starts in the parent folder of the files which are going to be moved/copied. If required, I can rephrase it to "Move or Copy: Start in the **files'** parent folder instead of root"

### Regarding the suggested code change
The same holds for all other keys, `EXTRA_FILES`, `EXTRA_FILE_PATHS`, `EXTRA_ACTION`. And all four of these keys are often accessed as if they are never null.

So how could they become null? `FolderPickerActivity::class.java.canonicalName` could only be null, if the class is hidden, a local class or an an anonymous class. None of this applies to FolderPickerActivity. Also, if any of these keys is null, all of the keys would be null, as they are constructed in a similar pattern. And then they would be indistinguishable from each other, negating the purpose of keys.

What if we would instead assert their non-nullness at time of definition, using `1!` like this:
```
val EXTRA_FOLDER: String = FolderPickerActivity::class.java.canonicalName!!.plus(".EXTRA_FOLDER")
```
and similar for the other three `EXTRA_` keys? I've also found the same pattern in IndeterminateProgressDialog.kt`.

**I added a separate commit for this. Let me know if you like it, otherwise I can remove it.**

alperozturk96: Thanks for the update and the improvements you’ve made here. For this PR, I’d like us to keep the scope limited to `EXTRA_FOLDER` only.

Although `EXTRA_FOLDER` is probably never going to be null at runtime, since this PR specifically uses `EXTRA_FOLDER`, we have the opportunity to handle it more cleanly. Rather than assuming it will never be null or relying on the `!!` operator, we can explicitly handle it.

For the other keys and classes: while they may have been handled differently in the past, that doesn’t necessarily mean we should continue the same pattern. However, changing them in this PR would increase the scope and require additional testing, as some classes can be instantiated in multiple ways. To avoid unintended side effects, I’d prefer we leave those untouched for now.

So, could you please roll back the changes related to the keys and keep this PR as minimal as possible? Since this PR is meant to fix the Move / Copy behavior only, we should avoid modifying keys here. We can always improve the handling of the remaining keys incrementally in future PRs.

Thanks for understanding, and I appreciate the effort you’ve put into this.

PhilLab: @alperozturk96 okay, so I'll limit the changes to `EXTRA_FOLDER`. However, I strongly believe that this should be not a runtime check, because the key being null would be a catastrophic error. It simply should never happen, no matter the instantiation. Because if it happens, all keys will have the same value and the lookups will be completely wrong. We are not talking about arguments and _values_ being null here (which is perfectly valid), but the keys themselves.

With all that being said, just to confirm, as I wasn't sure what I can leave in there and what not: I should
1. Revert the changes to all key definitions.
2. Change the folder parsing to
    `val extraFolder = fileArg?.let { intent.getParcelableArgument(it, OCFile::class.java) }`
    `file = extraFolder ?: listOfFilesFragment?.currentFile`

Did I get that right?

alperozturk96: Yes, the keys should not be null. You may update `EXTRA_FOLDER` to be a non-nullable String (for example, "FOLDER_PICKER_ACTIVITY_ARG_EXTRA_FOLDER"), or keep it as is and explicitly handle the nullable case, like I showed above. The main point is that if the key is not nullable, we can simply retrieve the argument directly. If it remains nullable for certain reasons (such as `FolderPickerActivity::class.java.canonicalName`), then I’d just like to see the nullable case handled explicitly in the code. Either approach works for me, but I prefer non-nullable keys.

Please revert the changes made to all other key definitions, only the related class should be modified.

Since I will be testing again after your changes, modifying only the `EXTRA_FOLDER` key is okay.

Thank you 👍

PhilLab: @alperozturk96 take a look at the latest commit, I followed what I think is the minimal viable change. I touched the one key only and made it a constant, using the value it currently always has, when evaluated.

`FolderPickerActivity::class.java.canonicalName` results in `com.owncloud.android.ui.activity.FolderPickerActivity`

Please review again.

PR #15529, opened 2025-09-04
feat: do not show loading spinner

alperozturk96:
- [x] Tests written, or not not needed

PR #15932, opened 2025-11-10
fix: loading indicator

alperozturk96:

### Issue

Loading spinner stuck when user chooses the empty folder (from media → set media folder) and pull-to-refresh.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
@@ -1153,48 +1153,30 @@ public class ReceiveExternalFilesActivity extends FileActivity
                 boolean sameAccount = getAccount() != null && accountName.equals(getAccount().name)
                     && getStorageManager() != null;
 
-                if (sameAccount) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -856,7 +855,6 @@ private void startSyncFolderOperation(OCFile folder) {
 
         executorService.execute(() -> {
             long currentSyncTime = System.currentTimeMillis();
-            mSyncInProgress = true;
```
alperozturk96, line 859: Unused unnecessary duplicated logic that we have in `FileDisplayActivity`

app/src/main/java/com/owncloud/android/ui/fragment/ExtendedListFragment.kt
```diff
@@ -720,6 +721,10 @@ open class ExtendedListFragment :
                     true
                 )
             }
+        }.also {
+            mainThread {
+                mRefreshListLayout?.isRefreshing = false
```
alperozturk96, line 726: `isRefreshing` only set via pull-to-refresh since from https://github.com/nextcloud/android/pull/15529, no need to keep it as true.

AndyScherzinger (approved):

PR #15934, opened 2025-11-10
feat: chat ui

alperozturk96:

Library PR: https://github.com/nextcloud/android-library/pull/1881


kra-mo:
> * The x on the right side of the chat list is odd. Not sure how to test this PR, but just looking at it, I can't figure out what it would do.

I'm guessing it is

> `close conversations list`

Which returns you to the chat?

kra-mo:
Ok, I now tested it and it indeed, returns you to the chat UI.

Since this is now secondary, I would instead replace the x with a back button on the left side and the icon to open the chat list in general with something like `history`:

image

kra-mo:
Also, creating a new chat should immediately open it instead of keeping you in the conversation list.

alperozturk96:
@kra-mo 

https://github.com/user-attachments/assets/5f24b666-7183-43f5-9e06-e7e82b81f361

kra-mo:
It looks nice, thank you :)

One last comment is that I guess that the back button looks a bit odd without anything else in that bar.

Maybe include a page title there, like "Conversations"

alperozturk96:
@kra-mo 

Screenshot 2025-11-14 at 14 40 21

kra-mo (approved):

PR #15949, opened 2025-11-12
fix: back press handling in settings activity

alperozturk96:

### Fixes

Fixes drawer menu highlighting and transition from `SettingsActivity`.

Old related PR: https://github.com/nextcloud/android/pull/15384

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
@@ -217,16 +217,16 @@ public class SettingsActivity extends PreferenceActivity
         }
     }
 
-    public static boolean onBackPressed = false;
+    public static boolean isBackPressed = false;
 
     @SuppressLint("GestureBackNavigation")
     @Override
     public void onBackPressed() {
-        onBackPressed = true;
+        isBackPressed = true;
         super.onBackPressed();
         new Handler(Looper.getMainLooper()).postDelayed(() -> {
             Log_OC.d(TAG, "User returned from settings activity, reset onBackPressed flag.");
-            onBackPressed = false;
+            isBackPressed = false;
         }, 2000);
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
```diff
@@ -358,16 +357,6 @@ private void setupAboutCategory(String appVersion) {
         }
     }
 
-    @SuppressLint("GestureBackNavigation")
-    @Override
-    public void onBackPressed() {
```
alperozturk96, line 367: No need to override the back press behavior, since the back press will handle itself and display whatever is in the stack.

app/src/main/java/com/owncloud/android/ui/fragment/contactsbackup/BackupFragment.kt
```diff
@@ -336,7 +336,6 @@ class BackupFragment :
                 contactsPreferenceActivity.finish()
             } else {
                 val settingsIntent = Intent(context, SettingsActivity::class.java)
-                settingsIntent.flags = Intent.FLAG_ACTIVITY_SINGLE_TOP or Intent.FLAG_ACTIVITY_CLEAR_TOP
```
alperozturk96, line 339: Don't clear stack so that we can go back from `SettingsActivity`

app/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
```diff
@@ -1056,8 +1045,6 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
             handleMnemonicRequest(data);
         } else if (requestCode == ACTION_E2E && data != null && data.getBooleanExtra(SetupEncryptionDialogFragment.SUCCESS, false)) {
             Intent i = new Intent(this, SettingsActivity.class);
-            i.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
```
alperozturk96, line 1081: Don't clear stack so that we can go back from `SettingsActivity`

tobiasKaminsky (changes requested):
As dicussed

PR #15977, opened 2025-11-17
improve storage permission ui ux

alperozturk96:

### Changes

This PR simplifies and consolidates the app’s storage-permission handling to improve clarity, reduce duplicated logic.

**StoragePermissionDialog**

- Removed the dependency on an external listener.
- Ask media permission as well after full file access. (e.g. `MEDIA_LOCATION`)

**UploadListActivity**

- Removed the repeated permission-request flow.
- Added a non-intrusive Storage Permission Banner that appears when required permissions are missing.

**PermissionUtil**

Remove the external override flag.
Consolidated storage-permission checks.
Reduced code duplication and improved readability.

### Storage Permission Banner Behavior

@kra-mo 

When the user denies storage permissions, including full file access, media access, or limited access, the app can no longer trigger the native permission dialog. Because these permissions are essential for both auto and manual uploads, the banner provides clear user guidance.

The banner offers the following options:

Media Access → Opens the system settings page for granting media access.
Full File Access → Redirects the user to the app-specific full-file access settings page.
Don’t Ask Again → Allows the user to permanently dismiss the banner if they do not wish to see it again.

Screenshot_20251119_130958

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -122,7 +122,7 @@
     <string name="prefs_sycned_folders_summary">Manage folders for auto upload</string>
 
     <string name="prefs_all_files_access_title">All files access</string>
-    <string name="prefs_all_files_access_summary">Allow the app to access and manage all files on your device.</string>
+    <string name="prefs_all_files_access_summary">Allow the app to access and manage all files on your device</string>
 
     <string name="prefs_help">Help</string>
     <string name="prefs_recommend">Recommend to a friend</string>
@@ -1307,7 +1307,7 @@
     <string name="error_creating_file_from_template">Error creating file from template</string>
     <string name="no_send_app">No app available for sending the selected files</string>
     <string name="pdf_zoom_tip">Tap on a page to zoom in</string>
-    <string name="storage_permission_full_access">Full access</string>
+    <string name="storage_permission_all_files_access">All files access</string>
     <string name="storage_permission_media_read_only">Media read-only</string>
     <string name="storage_permission_dont_ask">Don\'t ask</string>
     <string name="subtitle_photos_videos">Photos and videos</string>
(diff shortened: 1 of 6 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1302,8 +1301,10 @@
     <string name="error_creating_file_from_template">Error creating file from template</string>
     <string name="no_send_app">No app available for sending the selected files</string>
     <string name="pdf_zoom_tip">Tap on a page to zoom in</string>
+    <string name="storage_permission_dont_show_again">Dont show again</string>
```
kra-mo:
```suggestion
    <string name="storage_permission_dont_show_again">Don't show again</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -1302,8 +1301,10 @@
...
     <string name="storage_permission_full_access">Full access</string>
     <string name="storage_permission_media_read_only">Media read-only</string>
+    <string name="storage_permission_dont_ask">Dont ask</string>
```
kra-mo:
```suggestion
    <string name="storage_permission_dont_ask">Don't ask</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -1290,9 +1290,8 @@
     <string name="no_actions">No actions for this user</string>
     <string name="search_error">Error getting search results</string>
     <string name="load_more_results">Load more results</string>
-    <string name="file_management_permission">Permissions needed</string>
+    <string name="storage_permission_banner_text">Storage permission is required for auto-upload and file uploads.</string>
```
kra-mo:
```suggestion
    <string name="storage_permission_banner_text">Storage permission is required for Auto-upload and file uploads.</string>
```

I think that since this is a feature, it should be capitalized

kra-mo (changes requested):
Oops, accidentally approved instead of requesting changes. In any case, it's just a couple of typos mostly.

alperozturk96:
The apostrophe (') is a special character. To show it in the UI `\` is needed, but backslash not visible for user, FYI. @kra-mo

PR #16005, opened 2025-11-25
Fix typos and grammar in strings.xml

duvvuvenkataramana:

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -120,7 +120,7 @@
     <string name="prefs_calendar_contacts_sync_setup_successful">Calendar and contacts sync set up</string>
     <string name="prefs_daily_backup_summary">Daily backup of your calendar and contacts</string>
     <string name="prefs_daily_contact_backup_summary">Daily backup of your contacts</string>
-    <string name="prefs_synced_folders_summary">Manage folders for auto upload</string>
+    <string name="prefs_sycned_folders_summary">Manage folders for auto upload</string>
     <string name="prefs_help">Help</string>
     <string name="prefs_recommend">Recommend to a friend</string>
     <string name="prefs_imprint">Imprint</string>
@@ -343,7 +343,7 @@
         <item quantity="one">You can upload only %d file at once.</item>
         <item quantity="other">You can upload up to %d files at once.</item>
     </plurals>
-    <string name="sync_foreign_files_forgotten_explanation">As of version 1.3.16, files uploaded from this device are copied into the local %1$s folder to prevent data loss when a single file is synced with multiple accounts.\n\nDue to this change, all files uploaded with earlier versions of this app were copied into the %2$s folder. However, an error prevented the completion of this operation during account synchronization. You may either leave the file(s) as is and delete the link to %3$s, or move the file(s) into the %1$s folder and retain the link to %4$s.\n\nListed below are the local file(s), and the remote file(s) in %5$s they were linked to.</string>
+    <string name="sync_foreign_files_forgotten_explanation">As of version 1.3.16, files uploaded from this device are copied into the local %1$s folder to prevent data loss when a single file is sycned with multiple accounts.\n\nDue to this change, all files uploaded with earlier versions of this app were copied into the %2$s folder. However, an error prevented the completion of this operation during account synchronization. You may either leave the file(s) as is and delete the link to %3$s, or move the file(s) into the %1$s folder and retain the link to %4$s.\n\nListed below are the local file(s), and the remote file(s) in %5$s they were linked to.</string>
     <string name="sync_current_folder_was_removed">The folder %1$s does not exist anymore</string>
     <string name="foreign_files_move">Move all</string>
     <string name="foreign_files_success">All files were moved</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -6,7 +6,7 @@
   ~ SPDX-License-Identifier: AGPL-3.0-or-later OR GPL-2.0-only
 -->
 <resources>
-    <string name="bottom_navigation_menu_files_label">All files</string>
+    <string name="bottom_navigation_menu_files_label">All Files</string>
```
AndyScherzinger: unsure about this line @rakekniven

rakekniven: No, this should be "All files".
We have strings like "Open in Files", which contains the name of the app.

duvvuvenkataramana:
@AndyScherzinger @rakekniven I changed and corrected some spellings. Review and Merge

app/src/main/res/values/strings.xml
```diff
@@ -119,7 +120,7 @@
     <string name="prefs_calendar_contacts_sync_setup_successful">Calendar and contacts sync set up</string>
     <string name="prefs_daily_backup_summary">Daily backup of your calendar and contacts</string>
     <string name="prefs_daily_contact_backup_summary">Daily backup of your contacts</string>
-    <string name="prefs_sycned_folders_summary">Manage folders for auto upload</string>
+    <string name="prefs_synced_folders_summary">Manage folders for auto upload</string>
```
AndyScherzinger: This will fail without changing code that referenced the identifier _and_ will delete all existing translations, so I would not change keys for the moment.

app/src/main/res/values/strings.xml
```diff
@@ -1031,7 +1032,7 @@
     <string name="no_browser_available">No app available to handle links</string>
     <string name="no_pdf_app_available">No App available to handle PDF</string>
     <string name="no_email_app_available">No App available to handle mail address</string>
-    <string name="no_map_app_availble">No App available to handle maps</string>
+    <string name="no_map_app_available">No App available to handle maps</string>
```
AndyScherzinger: Same here. This will fail without changing code that referenced the identifier and will delete all existing translations, so I would not change keys for the moment.

app/src/main/res/values/strings.xml
```diff
@@ -1454,7 +1455,7 @@
     <string name="document_storage_provider_cannot_create_file_inside_folder">Can’t create file: missing permission.</string>
     <string name="document_storage_provider_cannot_create_folder_inside_folder">Can’t create folder: missing permission.</string>
     <string name="link_not_followed_due_to_security_settings">Link not followed due to security settings.</string>
-    <string name="extented_list_fragment_empty_list_message_no_write_permission">You don’t have permission to create or upload files in this folder.</string>
+    <string name="extended_list_fragment_empty_list_message_no_write_permission">You don’t have permission to create or upload files in this folder.</string>
```
AndyScherzinger: And same here. This will fail without changing code that referenced the identifier and will delete all existing translations, so I would not change keys for the moment.

AndyScherzinger:
@duvvuvenkataramana thanks for the changes, please see my comments. This won't compile since changing the ID for a string basically breaks the ID-reference in the java/kotlin code, so the app won't even compile anymore.

app/src/main/res/values/strings.xml
duvvuvenkataramana: ok

app/src/main/res/values/strings.xml
duvvuvenkataramana: ok

app/src/main/res/values/strings.xml
duvvuvenkataramana: ok

app/src/main/res/values/strings.xml
```diff
@@ -342,7 +343,7 @@
...
+    <string name="sync_foreign_files_forgotten_explanation">As of version 1.3.16, files uploaded from this device are copied into the local %1$s folder to prevent data loss when a single file is sycned with multiple accounts.\n\nDue to this change, all files uploaded with earlier versions of this app were copied into the %2$s folder. However, an error prevented the completion of this operation during account synchronization. You may either leave the file(s) as is and delete the link to %3$s, or move the file(s) into the %1$s folder and retain the link to %4$s.\n\nListed below are the local file(s), and the remote file(s) in %5$s they were linked to.</string>
```
AndyScherzinger, line 346: This adds a typo and is not removing one!

app/src/main/res/values/strings.xml
```diff
@@ -548,8 +549,8 @@
     <string name="preview_image_error_no_local_file">There is no local file to preview</string>
 
     <string name="error__upload__local_file_not_copied">%1$s could not be copied to %2$s local folder</string>
-    <string name="prefs_synced_folders_local_path_title">Local folder</string>
-    <string name="prefs_synced_folders_remote_path_title">Remote folder</string>
+    <string name="prefs_sycned_folders_local_path_title">Local folder</string>
```
AndyScherzinger, line 552: again changing string IDs will break code!

app/src/main/res/values/strings.xml
duvvuvenkataramana: this is correct one

app/src/main/res/values/strings.xml
duvvuvenkataramana: this is correct one only sycned

AndyScherzinger:
Hi @duvvuvenkataramana quick question, are you vibe coding this? I am asking because 
* strings-xml is US-English and the GB/en equivalent in the repo runnign through a translation service would be British English
* `synced` got changed to `sycned` now which is introducing typos
* new commits add changes to String identifiers which I stated breaks the code from compiling!

PR #16009, opened 2025-11-25
fix(calendar): correct ICS event formatting and time zone usage

Gorlesunilkumar:
The ICS event file has been updated to follow proper RFC 5545 formatting and improve compatibility with calendar applications.

### Changes
- Replaced the empty DESCRIPTION field with a valid value
- Unified DTSTART and DTEND to both use TZID=Europe/Berlin for consistent time zone handling
- Corrected the event duration (09:00–10:00 Berlin time)
- Removed unnecessary blank lines to avoid parsing inconsistencies
- Kept the full VTIMEZONE block to ensure compatibility with Nextcloud, Google Calendar, and iOS

### Why
These adjustments improve interoperability with various calendar clients and prevent potential parsing issues caused by mixed time formats or empty fields.

### Testing
The change updates only static test assets. No functional logic was modified.
- [x] Tests not needed (data file correction only)


Gorlesunilkumar:
@alperozturk96 Review and merge

PR #16023, opened 2025-11-27
fix: screenshot tests logic

alperozturk96:


alperozturk96:
@tobiasKaminsky 

How can we fix this? 

Screenshot 2025-11-28 at 08 54 11

alperozturk96:
The `EventBus` handling still needs to be addressed for the tests. Since this issue is also blocking the master branch and the screenshot tests are consistently bypassing the inner assertions, I will proceed with merging this PR.

Screenshot 2025-12-03 at 14 58 30

PR #16026, opened 2025-11-28
fix user account switch, use account name instead of hash

tobiasKaminsky:
It seems that after app restart the hash is different and thus account switching for the very first time does not work.

To test:
- install & config master
- adb shell pm clear com.nextcloud.client
- open app and try to switch account
--> fails
- test above with this PR -> works


AndyScherzinger:
@tobiasKaminsky could this also be the issue with account switching via the ecosystem header in the notes app drawer?
Also didn't work in the recent past and also used a hash.

PR #16040, opened 2025-12-01
feat: show feedback only on error

alperozturk96:

This pull request removes several messages to reduce visual noise in the app.

**Removed messages**

Exporting files:
“File will be exported. See notification for details.”

Moving files:
“All files were moved”

Switching accounts:
“Logged in as …”

Restoring a file version:
“Successfully restored file version.”

Displaying helper text in PDF preview: “Tap on a page to zoom in”.


kra-mo:
> Copying text to clipboard (e.g., when reporting an issue):
> “Copied to clipboard”

There should be *some* form of indication when something has been copied so it might be fine to keep this one imo. But you could make it shorter, just "Copied". "to clipboard" is redundant.

> Should we keep this

Not sure, it should be clear that you can do so and if it isn't, that's a separate design issue so it's probably fine to drop it :)

PR #16043, opened 2025-12-02
fix: metadata photo size

alperozturk96:

### Issue

The `metadata-photo-size` property is being overwritten by the gallery search task. This property is provided through the `WebDAV API`, and the application should not override it with incorrect or null values.

**GallerySearchTask remote file:**

Screenshot 2025-12-02 at 10 23 13

### How to reproduce?

1. Have image files
2. Fresh install app
3. Check `filelist` database table before opening media tab `metadata_size` column has correct data but after opening media tab it becomes null.

Screenshot 2025-12-02 at 10 33 22


app/src/main/java/com/owncloud/android/datamodel/ThumbnailsCacheManager.java
```diff
@@ -1310,13 +1310,6 @@ public static Bitmap doResizedImageInBackground(OCFile file, FileDataStorageMana
             }
         }
 
-        // resized dimensions and set update thumbnail needed to false to prevent rendering loop
-        if (thumbnail != null) {
```
alperozturk96, line 1316: We should not override actual image size with thumbnail size.

PR #16048, opened 2025-12-03
fix: upload cancel & remove

alperozturk96:

### Changes

Fix cancel upload
Add remove upload

**This update improves the handling of file uploads by clearly distinguishing between Cancel and Remove actions:**

**Cancel Upload:**

1. Stops the current upload in progress.
2. Marks the upload as cancelled.
3. The upload remains visible in the upload list.
4. On app restart, cancelled uploads automatically resume.

**Remove Upload:**

1. Deletes the upload from the upload list entirely.
2. The file will never be uploaded again.

These two actions provide necessary functionality for managing uploads, especially in cases such as `SYNC_CONFLICT`, where users need to either resolve the conflict or remove the upload.

This update also ensures that users can clearly distinguish between cancelling an upload and permanently removing it, reducing confusion and improving overall upload management.

-----

How can we make it easier for users to understand the difference between these two actions? @kra-mo 

Screenshot_20251203_101634

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadBroadcastReceiver.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadBroadcastReceiver.kt
@@ -38,6 +38,7 @@ class FileUploadBroadcastReceiver : BroadcastReceiver() {
         }
     }
 
+    @Suppress("ReturnCount")
     private fun cancelUpload(context: Context, intent: Intent) {
         val uploadId = intent.getLongExtra(UPLOAD_ID, -1L)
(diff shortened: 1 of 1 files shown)


kra-mo:
> How can we make it easier for users to understand the difference between these two actions?

I guess this is merged now, but more accurate terms would be:

Cancel → Pause and
Remove → Cancel

as "cancel" sounds final.

kra-mo:
Although the first one could also be called "Stop" and the second could remain "Remove". Whatever you think makes more sense with the rest of the feature's existing wording.

alperozturk96:
@kra-mo I will create another PR for that, thanks.

Issue #16039, opened 2025-12-01
Download limit resets when changing anything in a link share


mpivchev:
### ⚠️ Before posting ⚠️

- [x] This is a **bug**, not a question or an enhancement.
- [x] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [x] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Set a link share with a download limit
2. Download a few times to see limit change
3. Change something on the share (ex name)

### Expected behaviour

Download limit should not change unless modified
Server works the same.

### Actual behaviour

Download limit changes no matter what

### Android version

/

### Device brand and model

/

### Stock or custom OS?

Stock

### Nextcloud android app version

/

### Nextcloud server version

/

### Using a reverse proxy?

I don't know

### Server error logs

```bash

```

### Additional information

iOS has a fix for it: https://github.com/nextcloud/ios/pull/3899

PR #16059, opened 2025-12-04
fix: update file download limit

alperozturk96:

### Issue: https://github.com/nextcloud/android/issues/16039

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/fragment/FileDetailsSharingProcessFragment.kt
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/FileDetailsSharingProcessFragment.kt
@@ -198,7 +198,6 @@ class FileDetailsSharingProcessFragment :
         setCheckboxStates()
         themeView()
         setVisibilitiesOfShareOption()
-        toggleNextButtonAvailability(isAnySharePermissionChecked())
         logShareInfo()
     }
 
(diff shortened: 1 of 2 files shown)


mpivchev:
Trying to add share to this file:

image

Title shows ROOT instead of the name of the file?
Also, the download limit option is not visible:

image

app/src/main/java/com/owncloud/android/ui/fragment/FileDetailsSharingProcessFragment.kt
```diff
@@ -610,13 +613,6 @@ class FileDetailsSharingProcessFragment :
                 )
     }
 
-    private fun toggleNextButtonAvailability(value: Boolean) {
```
alperozturk96, line 613: User can always update current share just like iOS and Web

app/src/main/java/com/owncloud/android/ui/fragment/FileDetailsSharingProcessFragment.kt
```diff
@@ -779,14 +774,6 @@ class FileDetailsSharingProcessFragment :
             return
         }
 
-        if (!isSharePermissionChecked() && !isCustomPermissionSelectedAndAnyCustomPermissionTypeChecked()) {
```
alperozturk96, line 782: User can always update current share just like iOS and Web

PR #16062, opened 2025-12-04
change upload buttons wording

alperozturk96:

Screenshot_20251204_160949


AndyScherzinger:
Hmm, just checked Firefox, and they also use "cancel" not "stop"

kra-mo:
I mean, another data point: Safari says "stop". I just think that in our case, stop or pause makes more sense as it will be resumed later, cancel really implies "don't do this" to me.

kra-mo:
Chrome has Pause and Cancel as well, Cancel for removing it.

kra-mo:
Right, and I mean Firefox doesn't have a separate way to remove the download either. And after clicking on "Cancel", the option becomes "Retry", which seems weird and more like the result of a failure state imo.

kra-mo (approved):
Fair enough. I'm leaning a bit more towards Pause and Cancel, especially after learning that Chrome does it but if you think this makes more sense, go ahead :)

alperozturk96:
Screenshot_20251205_102420

PR #16065, opened 2025-12-05
drop not supported api levels

alperozturk96:

Drop Android 8.1 support
Increase outdated server version to 30
Use android library version 2.22.2.


AndyScherzinger:
lemme check

AndyScherzinger:
Works for me when I set the lib version. I didn't for this PR but locally and only committed the keys and checksums for the dependencies

PR #16097, opened 2025-12-08
fix: action bar appereance

alperozturk96:

### Fixes

Sets the correct toolbar style for specific screen
Sets the correct leading icon for action bar
Sets the correct `currentSearchType`
Fixes multiple navigation item highlight
Fixes wrong navigation item highlighting if user comes back from `TrashbinActivity`
Fixes file name usage in action bar in detail screens
Sets correct search fragment value
Fixes `getPreviousFile()`. Since `currentSearchType` and `mSearchFragment` value is correct, no need to check filter type for `onBrowseUp`

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -713,19 +713,6 @@ public abstract class DrawerActivity extends ToolbarActivity
         launchActivityForSearch(searchEvent, id);
     }
 
-    private void handleSearchEvents(SearchEvent searchEvent, int menuItemId) {
-        if (this instanceof FileDisplayActivity) {
-            final Fragment leftFragment = ((FileDisplayActivity) this).getLeftFragment();
-            if (leftFragment instanceof GalleryFragment || leftFragment instanceof SharedListFragment) {
-                launchActivityForSearch(searchEvent, menuItemId);
-            } else {
-                EventBus.getDefault().post(searchEvent);
-            }
-        } else {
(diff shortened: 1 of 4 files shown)


alperozturk96:
@AndyScherzinger @tobiasKaminsky 

Expected behavior:

- The user opens the Favorites tab and sees the root favorite folder.
- The user navigates into a favorite folder and sees the files and subfolders that belong to that favorite folder.
- When the user goes back, they must return to the root favorite view and see the root favorite folder again.
- This is the same for the Shared tab and the All Files tab.

### Demo


https://github.com/user-attachments/assets/5954e70c-06e8-4e57-aaa7-c1976eb9c889

tobiasKaminsky:
Quick test with user Christine:
- open app
- go to "favorites"
- open "holidays & trips"
- on Android: only favorites files are shown
- on Web: all files are shown

AndyScherzinger:
> When the user goes back, they must return to the root favorite view and see the root favorite folder again.

@alperozturk96 When navigating via favorite into a subfolder navigating back up navigates you through the folder's hiearchy

alperozturk96:
> > When the user goes back, they must return to the root favorite view and see the root favorite folder again.
> 
> @alperozturk96 When navigating via favorite into a subfolder navigating back up navigates you through the folder's hiearchy

Isn't it expected behavior? Since favorites, shared are shortcuts and go up means navigate one folder level up. Because user choose to go to subdirectory of some favorite folder from favorite root.

Favorite root:

A
B (A/B)

If user chooses B directly from Favorite root, one level up is A not root favorite again?

AndyScherzinger:
> Isn't it expected behavior?

I don't know tbh, fine as is I guess then 👍

PR #16108, opened 2025-12-09
fix: sub file visibility

alperozturk96:

### Fixes

If folder is favorite or shared, sub-files must be visible


app/src/main/java/com/owncloud/android/ui/adapter/helper/OCFileListAdapterHelper.kt
```diff
@@ -90,11 +90,11 @@ class OCFileListAdapterHelper {
                 }
             }
 
-            if (isSharedView && !file.isShared) {
+            if (isSharedView && !directory.isShared && !file.isShared) {
```
alperozturk96: Since sub files are not marked as shared or favorite checking parent dir needed as well

AndyScherzinger: What happens in a sub sub folder? Or with any depth larger than 1?

alperozturk96:
Correct `currentSearchType` needed, thus let's check via this PR https://github.com/nextcloud/android/pull/16097 since it is providing it. @AndyScherzinger @tobiasKaminsky

PR #16115, opened 2025-12-10
fix: upload completion listener

alperozturk96:

### Issue

Once upload completed, adapter needs to be refreshed via correct data.

### Fixes

Sets worker state in finally blocks
Ensure all database operations are completed before signaling completion
Notify adapter only once when `FileUploadWorker` finished

### Demo

https://github.com/user-attachments/assets/ff191eee-69e3-4167-866c-1742e2c4884c


alperozturk96:
or `UploadFinishReceiver` can be used. Check with @tobiasKaminsky

alperozturk96:
replaced via https://github.com/nextcloud/android/pull/16317

PR #16116, opened 2025-12-10
Enhanced media player implementation

AlexViseven:
This merge request adds a persistent "queue" / playlist feature to the Nextcloud Media Player and implements a set of frequently requested playback behaviors and UI improvements:
- Shuffle mode (randomize playback order)
- Repeat modes (repeat one / repeat all / off)
- Automatically play the next file when the current media ends (so folders with mixed content will behave as playlists containing only media files)
- Automatic filtering to exclude non-media files when generating a playlist from a folder
- Improved media player notification controls while the app is minimized (skip to next / previous, play/pause)
The proposed implementation has been used in the HiDrive application for a long time.

image

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -49,6 +49,7 @@ import com.google.android.material.bottomnavigation.BottomNavigationView;
 import com.google.android.material.button.MaterialButton;
 import com.google.android.material.navigation.NavigationView;
 import com.google.android.material.progressindicator.LinearProgressIndicator;
+import com.ionos.player.model.PlaybackModel;
 import com.nextcloud.client.account.User;
 import com.nextcloud.client.di.Injectable;
 import com.nextcloud.client.files.DeepLinkConstants;
@@ -208,6 +209,9 @@ public abstract class DrawerActivity extends ToolbarActivity
     @Inject
     ClientFactory clientFactory;
 
+    @Inject
+    PlaybackModel playbackModel;
+    
     @Override
     public void onCreate(@Nullable Bundle savedInstanceState, @Nullable PersistableBundle persistentState) {
         super.onCreate(savedInstanceState, persistentState);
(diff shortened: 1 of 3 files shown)


AndyScherzinger:
pinged Tobias and Alper for a review - thanks for the contribution, looks great and interesting :)

@AlexViseven @dbotvin for the SPDX license identifier you should be using

`AGPL-3.0-or-later OR GPL-2.0-only` instead of `GPL-2.0`, the later is not a valid on per SPDX list and also wouldn't match the apps used license. You might want to add github.com/nextcloud/.github/blob/master/workflow-templates/reuse.yml to your fork or rebase to latest `master` branch of the nextcloud Android repo. Since the PR has conflicts you might want to rebase anyways.

app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -1851,6 +1977,35 @@ private ArrayList<ContentProviderOperation> prepareRemoveSharesInFile(
 
     }
 
+    public List<OCShare> getShares() {
```
alperozturk96, line 1980: Can we use `ShareDao` and get shares just like `fileDao.getFavoriteFiles`? So that we can remove the content resolver logic.

app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -2556,18 +2546,9 @@ class FileDisplayActivity :
...
-        // Safely handle the absence of a user
-        if (user != null) {
-            previewMediaIntent.putExtra(PreviewMediaActivity.EXTRA_USER, user)
-        }
-
-        previewMediaIntent.putExtra(PreviewMediaActivity.EXTRA_START_POSITION, startPlaybackPosition)
-        previewMediaIntent.putExtra(PreviewMediaActivity.EXTRA_AUTOPLAY, autoplay)
-        startActivity(previewMediaIntent)
+    private fun startMediaActivity(file: OCFile, startPlaybackPosition: Long, autoplay: Boolean, user: User?) {
+        val searchType = listOfFilesFragment?.currentSearchType
```
alperozturk96, line 2550: As far as I can see after this PR `PreviewMediaActivity` and `com.nextcloud.client.media.PlayerService` becomes obsolete if yes can we delete them?

alperozturk96 (changes requested):
Hey

Thank you for the great improvement. Could you please rebase to latest master and check the two small findings?

tobiasKaminsky:
Closing in favor of https://github.com/nextcloud/android/pull/16525

PR #16117, opened 2025-12-11
Fix Screenshots

tobiasKaminsky:


PhilLab:
@tobiasKaminsky it is only a small piece of the puzzle, but when extending tests for another PR, I noticed that this will be necessary to fix the screenshot tests on master as well: 55acbba190fbf02c09e0de4a3b2384cf699fb1fd (part of #16215). Cherry-pick it, if you like.

When executing locally, the test did still fail, because one of the icons changed from filled to outline. I didn't touch the reference screenshots, because I wasn't sure what the expected style actually is:
image

AndyScherzinger:
Hi @PhilLab,

to answer the question, we switched to outline, so updating the reference screenshot is correct 👍

PR #16119, opened 2025-12-11
Fix Gradle project sync error

PhilLab:
The commit 5fd2e2942c2b375b86c52256f7c8921395f5594c introduced stricter dependency verification, but this resulted in the Gradle project syncing to fail with the error:
> Multiple build operations failed.
> Dependency verification failed for configuration ':app:androidTestUtil':
>   - On artifact gradle-9.2.1-src.zip (gradle:gradle:9.2.1) in repository
> 'Gradle distributions': Artifact was signed with key 'E2F38302C8075E3D'
> but it wasn't found in any key server so it couldn't be verified
> This can indicate that a dependency has been compromised. Please
> carefully verify the signatures and checksums

This happened independent of the used gradle version..

The followup 32545d42f27fea81eff1d2d13cff966fb4e91640 added the key's fingerprint, but this was not enough to establish trust. Adding the public key to the keyring. The key was taken from
https://github.com/gradle/gradle/blob/b99fe7d304e68b978011840c31c08068f671f794/KEYS.md
and cross-checked with https://gradle.org/keys/


PhilLab:
@alperozturk96 I am unsure why this didn't cause issues for more developers but at least for me, building was impossible on two different machines

Issue #16126, opened 2025-12-13
3.35.0 RC2 Auto upload bugs


bilogic:
### ⚠️ Before posting ⚠️

- [x] This is a **bug**, not a question or an enhancement.
- [x] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [x] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

<img width="703" height="329" alt="Image" src="https://github.com/user-attachments/assets/72aeb530-a8db-490f-9f5d-4a32a52a6e77" />

1. My nextcloud is configured to auto upload and delete files from the camera folder
2. I took a picture, quickly deleted it from the folder before nextcloud has a chance to upload it
3. But nextcloud is unable to let go and forget about it, the screenshot keeps re-appearing even after I clicked on the trashcan MULTIPLE TIMES
4. What kind of use case calls for such behavior? Is it expecting the file to magically re-appear for it to upload again?
5. AND there are still photos in the camera folder, why is 3.35.0 RC2 not bothering to upload them? https://github.com/nextcloud/android/issues/16027
6. These behaviors defying basic logical thinking

### Expected behaviour

1. If a file no longer exists, nextcloud should LET GO and FORGET ABOUT IT
2. If files are still in the folder, nextcloud should RETRY to UPLOAD (and optionally delete it according to how the folder was configured)

### Actual behaviour

1. Nextcloud keeps trying to upload files that no longer exists and creates an Android notification to harass me
2. Nextcloud randomly refuses to upload 3 files that still exists in the folder https://github.com/nextcloud/android/issues/16027
3. This is utterly frustrating!

### Android version

16

### Device brand and model

Pixel 8

### Stock or custom OS?

Stock

### Nextcloud android app version

3.35.0 RC2

### Nextcloud server version

32.0.2

### Using a reverse proxy?

Yes

### Android logs

Cannot find

### Server error logs

```bash

```

### Additional information

In `Internal storage`, I cannot find `data` folder

bilogic:
Do we have to close now? There is no new nightly/beta whatever release, and I have not way to test/verify the fix...

binarydad:
[...]

PR #16132, opened 2025-12-15
fix(auto-upload): remove non-existing local files without restarting

alperozturk96:

### Issue: https://github.com/nextcloud/android/issues/16126

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/autoUpload/FileSystemRepository.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/autoUpload/FileSystemRepository.kt
@@ -28,6 +28,10 @@ class FileSystemRepository(private val dao: FileSystemDao, private val context: 
         const val BATCH_SIZE = 50
     }
 
+    suspend fun deleteByLocalPathAndId(path: String, id: Int) {
+        dao.deleteByLocalPathAndId(path, id)
+    }
+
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nextcloud/client/jobs/autoUpload/AutoUploadWorker.kt
```diff
@@ -322,6 +322,15 @@ class AutoUploadWorker(
 
                 try {
                     var (uploadEntity, upload) = createEntityAndUpload(user, localPath, remotePath)
+
+                    // if local file deleted, upload cannot be or retriable thus needs to be removed
+                    if (path.isEmpty() || !file.exists()) {
+                        Log_OC.w(TAG, "detected not existing local file, removing entity")
```
AndyScherzinger:
```suggestion
                        Log_OC.w(TAG, "detected non-existing local file, removing entity")
```

app/src/main/java/com/nextcloud/client/jobs/autoUpload/AutoUploadWorker.kt
```diff
@@ -322,6 +322,15 @@ class AutoUploadWorker(
 
                 try {
                     var (uploadEntity, upload) = createEntityAndUpload(user, localPath, remotePath)
+
+                    // if local file deleted, upload cannot be or retriable thus needs to be removed
```
alperozturk96:
```suggestion
                    // if local file deleted, upload process cannot be started or retriable thus needs to be removed
```

PR #16142, opened 2025-12-15
fix: search task root sub dir content fetch

alperozturk96:

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
@@ -496,7 +496,7 @@ public class FileDataStorageManager {
     }
 
     public boolean saveFile(OCFile ocFile) {
-        Log_OC.d(TAG, "saving file: " + ocFile.getRemotePath() + " etag: " + ocFile.getEtag());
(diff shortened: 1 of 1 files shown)


tobiasKaminsky (changes requested):
Replace with FileDataStorageManager -> saveFileWithParent

app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -2348,74 +2349,68 @@ class FileDisplayActivity :
...
+                        storageManager,
+                        user.get(),
+                        applicationContext
+                    )
+                    operation.execute(user.get(), this@FileDisplayActivity)
+                }
+
+                if (result.isSuccess) {
+                    Log_OC.d(TAG, "folder is refreshed: ${folder?.remotePath}")
+                    ocFileListFragment?.refreshDirectory()
```
alperozturk96: This prevents opening an empty directory (which only happens when the user tries to open directories quickly, such as going back and forth). First, we need to complete the folder refresh operation before displaying data.

**To test:**

1. Perform a fresh installation of the app.
2. Try opening directories quickly.

alperozturk96:
Object not found for spotbugs increased message

Screenshot 2025-12-19 at 12 45 57

PR #16145, opened 2025-12-15
Fixing timing issue with FileActivity's loading dialog show and dismiss

PhilLab:

Before dismissing the dialog, we need to wait for a potentially pending transaction. As showing the dialog also includes the dismissing of prior instances, we need to wait there as well.
Both is needed to satisfy the test case added in the previous commit.
    
Otherwise, the dialog might be shown after it was meant to be dismissed already. This issue was observed when testing RemoveFilesDialogFragment's removeFiles() and also sporadically "in the wild" when using the app productively

If you want to see for yourself, first checkout the test commit and see the test fail. Then checkout the fix commit and see the test succeed

### DIFF
--- a/app/src/androidTest/java/com/nextcloud/test/GrantStoragePermissionRule.kt
+++ b/app/src/androidTest/java/com/nextcloud/test/GrantStoragePermissionRule.kt
@@ -15,6 +15,10 @@ import org.junit.rules.TestRule
 import org.junit.runner.Description
 import org.junit.runners.model.Statement
 
+/**
+ * Rule to automatically enable the test to write to the external storage.
+ * Depending on the SDK version, different approaches might be necessary to achieve the full access.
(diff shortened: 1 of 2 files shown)


PhilLab:
@alperozturk96 can you review?

PhilLab:
@alperozturk96 Indeed. This should be fixed on a global level by the changes of #15424, which grabs the notification permission for all tests. If we merge that one first, the test here will also succeed.

PhilLab:
@alperozturk96 I cherry-picked the relevant commit from said PR, `Instrumentation tests now automatically grab the notifications permission`, onto this PR here now. Can you re-test, it should work now.

PhilLab:
@alperozturk96 I recreated the PR in the main repository here, so it can run on the CI: #16296

Issue #16139, opened 2025-12-15
Empty Folder Content


alperozturk96:
### ⚠️ Before posting ⚠️

- [x] This is a **bug**, not a question or an enhancement.
- [x] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [x] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Open root directory
2. Create a folder from web and add some file in it
3. Open same folder from app
4. Empty folder will be shown, however file must be visible.

### Expected behaviour

Folder contents should be displayed correctly.

### Actual behaviour

Sometimes folder contents looks empty.

### Android version

11

### Device brand and model

Emulator

### Stock or custom OS?

Stock

### Nextcloud android app version

3.34.0

### Nextcloud server version

30

### Using a reverse proxy?

I don't know

### Server error logs

```bash

```

### Additional information

`MetadataWorker` and normal folder navigation triggers `RefreshFolderOperation.java` in this class we check eTag and update content of the folder. `synchronizeData(List<Object> folderAndFiles)` responsible for the updating logic.

However, in some cases child file saved into the DB but parent ID of it becomes 0 thus it is causing connection lost between child file and parent dir.

PR #16148, opened 2025-12-16
fix: use existing parent id for gallery search task

alperozturk96:

### Issue

Helps: https://github.com/nextcloud/android/issues/16139. Please read the issue first.

### How to reproduce?

1. Have a large number of images in nested directories
2. Perform a fresh installation of the app
3. Open the Media tab
4. Navigate back to All Files
5. Observe the database using App Inspection: files that previously had a valid parent ID are set to `0` after visiting the Media tab

### Fixes

Overwriting parent id value

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/asynctasks/GallerySearchTask.java
+++ b/app/src/main/java/com/owncloud/android/ui/asynctasks/GallerySearchTask.java
@@ -51,7 +51,7 @@ public class GallerySearchTask extends AsyncTask<Void, Void, GallerySearchTask.R
     }
 
     @Override
-    protected GallerySearchTask.Result doInBackground(Void... voids) {
+    protected Result doInBackground(Void... voids) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/ui/asynctasks/GallerySearchTask.java
```diff
@@ -160,8 +162,15 @@ private boolean parseMedia(long startDate, long endDate, List<Object> remoteFile
                 remoteFile.setUploadTimestamp(existingFile.getUploadTimestamp());
             }
 
+            // initialize oc file from remote file
             OCFile ocFile = FileStorageUtils.fillOCFile(remoteFile);
 
+            // since remote does not contains parent id information use existing file to prevent overriding parent id with 0
```
alperozturk96, line 168: Remote file does not contains the parent id thus fix needs to be in client. @tobiasKaminsky What do you think?

PR #16165, opened 2025-12-18
fix: register/unregister broadcast receivers

alperozturk96:

### Issue

When the user navigates to a subdirectory from the Favorites or Shared tab, the receivers are unregistered, which causes the folder to not refresh correctly.

### Changes

Receivers needs to be registered and unregistered on onStart/onStop.

**Documentation:**

Screenshot 2025-12-18 at 13 51 01

**Result**

Screenshot 2025-12-18 at 13 47 37

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
@@ -1401,10 +1401,10 @@ class FileDisplayActivity :
     }
 
     private fun registerUploadFinishReceiver() {
-        val uploadIntentFilter = IntentFilter(getUploadFinishMessage())
+        val filter = IntentFilter(getUploadFinishMessage())
         mUploadFinishReceiver = UploadFinishReceiver()
         mUploadFinishReceiver?.let {
-            localBroadcastManager.registerReceiver(mUploadFinishReceiver!!, uploadIntentFilter)
+            localBroadcastManager.registerReceiver(it, filter)
         }
     }
 


app/src/main/java/com/owncloud/android/ui/fragment/SharedListFragment.kt
```diff
@@ -63,7 +63,6 @@ class SharedListFragment :
                 val fileDisplayActivity = activity as FileDisplayActivity
                 fileDisplayActivity.updateActionBarTitleAndHomeButtonByString(getString(R.string.drawer_item_shared))
                 fileDisplayActivity.setMainFabVisible(false)
-                fileDisplayActivity.initSyncBroadcastReceiver()
```
alperozturk96, line 66: No need to register from here again since SharedListFragment child of OCFileListFragment and OCFileListFragment used in FileDisplayActivity.

app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1411,9 +1391,36 @@ class FileDisplayActivity :
...
+        registerSyncBroadcastReceiver()
+        registerDownloadFinishReceiver()
+        registerUploadFinishReceiver()
+    }
+
+    private fun registerUploadFinishReceiver() {
+        val uploadIntentFilter = IntentFilter(getUploadFinishMessage())
+        mUploadFinishReceiver = UploadFinishReceiver()
+        mUploadFinishReceiver?.let {
+            localBroadcastManager.registerReceiver(mUploadFinishReceiver!!, uploadIntentFilter)
```
ZetaTom:
```suggestion
        val filter = IntentFilter(getUploadFinishMessage())
        mUploadFinishReceiver = UploadFinishReceiver()
        mUploadFinishReceiver?.let {
            localBroadcastManager.registerReceiver(it, filter)
```

ZetaTom (approved):

PR #16166, opened 2025-12-18
Apppassword onetime

tobiasKaminsky:


app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -1594,11 +1597,43 @@ private void startQRScanner() {
...
+        new Thread(() -> {
+            RemoteOperationResult<String> otpResult = nextcloudClient.execute(generateOneTimeAppPasswordRemoteOperation);
+
+            if (otpResult.isSuccess()) {
+                mServerInfo.mBaseUrl = AuthenticatorUrlUtils.INSTANCE.normalizeUrlSuffix(loginUrlInfo.getServer());
+                webViewUser = loginUrlInfo.getLoginName();
+                webViewPassword = otpResult.getResultData();
+
+                runOnUiThread(this::checkOcServer);
+            } else {
```
ZetaTom, line 1628: From a UX perspective, checking for `401 Unauthorized` and showing a better fitting error message may be worth it. This should only happen when the scanned code has expired.

tobiasKaminsky: Expired code can only be scanned when they do a screenshot from it, else it gets removed.
So I guess we can first skip that :)

ZetaTom (approved):

PR #16184, opened 2025-12-20
Bugfix and automated tests for #15636: deleting from slideshow

PhilLab:

This tests the deletion behavior when deleting from slide show, fixed in #15636. It also fixes another bug which was uncovered during the automated tests, where the view pager was not correctly updated.

This PR requires #16145 to be merged first, as the automated tests also stumble upon the same timing issue fixed (and tested!) in said PR.


app/src/main/java/com/owncloud/android/ui/dialog/RemoveFilesDialogFragment.kt
PhilLab, line 1: @alperozturk96 during developing the automated tests, a bug was discovered and fixed in the same PR. This is why this file was touched.

Can you review the fix and the tests, so both can be merged?

PhilLab:
Recreated the PR in the main repository: #16297

PR #16188, opened 2025-12-22
Test that move or copy files starts in the file's parent folder

PhilLab:
This tests the changes of #15925, where the old behavior of moving up to the root folder was changed to stay in the parent folder of the file(s) to move/copy.
The new behavior is in sync with iOS and the Web UI, hence this test ensures it doesn't break


tobiasKaminsky:
Since this will not run on our CI, I copied it to https://github.com/nextcloud/android/pull/16188

since you are now a member of our team, please create new PRs directly within our repo :)

PhilLab:
Makes sense. Just for future reference: typo in above message, rhe new PR is actually #16262

PR #16215, opened 2025-12-29
When uploading a single file, offer input field to change file name

### LINKED ISSUE #9215: when upload a file alllow modify file name
When I send a file to NC for android app to upload it, I can chose Just the destination folder.
It would be useful to be able to change/choose the file name too. This avoid me to load a file, for example from email Attachment, and go back to NC app ti find and rename it.

The use case should be:
1) the user uses the share function of an external app to send a file to NC
2) NC app shows a view with new file name and last used destination folder
3) user optionally changes file name
4) user optionally clicks on the destination folder field to change it: the NC app then shows the "select destination" picker view
5) user loads the file

PhilLab:

This streamlines the upload from files, e.g. from email attachments.
Closes #9215.
Also offers a convenient pre-selection of the filename without extension when entering the input field, as well as file name validation.
| New input field | No input field (e.g multi file upload). Also old behavior |
|-|-|
| Screenshot_20260102_121002 | Screenshot_20260102_121341 |

Recording of the automated test:
(The upload failure in the end is expected, as there is no server mocked)

https://github.com/user-attachments/assets/a15ad2ee-87c3-46b2-8473-1701e3ba1d3a

### Technical details

1. The approach via the listener mFileDisplayNameTransformer was chosen because this allows a potential extension to multiple files (with multiple input fields). Alternatively, ReceiveExternalFilesActivity#uploadFile() could've been used, which requires the storage to a temporary file first
2. nextFocusForward and nextFocusDown are both needed to support navigation with the Tab key as well as the Enter key.
3. The automated test is not yet verifying that the upload is actually using the user-provided name. I attempted to mock parts of the involved utility classes, but spent an inappropriate ammount of time without achieving anything.
4. The automated test also changes the two screenshots, because the test files are prepared in a `@Before` method which apply to all tests. Screenshot tests are broken on master right now, so I did not feel confident to provide updated screenshots with no way of verifying them. Either they need to be updated later, or the tests should be separated in two files

### DIFF
--- a/app/src/androidTest/java/com/nextcloud/test/NextcloudViewMatchers.kt
+++ b/app/src/androidTest/java/com/nextcloud/test/NextcloudViewMatchers.kt
@@ -18,11 +18,13 @@ fun withSelectedText(expected: String): Matcher<View> = object : TypeSafeMatcher
         description.appendText("with selected text \"$expected\"")
     }
 
+    @Suppress("ReturnCount")
     override fun matchesSafely(view: View): Boolean {
         if (view !is TextView) return false
         val text = view.text?.toString() ?: ""
         val s = view.selectionStart
         val e = view.selectionEnd
+        @Suppress("ComplexCondition")
         if (s < 0 || e < 0 || s > e || e > text.length) return false
         return text.substring(s, e) == expected
     }


app/src/androidTest/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivityIT.kt
```diff
@@ -40,4 +61,109 @@ class ReceiveExternalFilesActivityIT : AbstractIT() {
...
+                .perform(ViewActions.clearText())
+                .perform(ViewActions.typeTextIntoFocusedView(thirdFileName))
+                .check(matches(withText(thirdFileName)))
+                // Leave the field and come back to verify the pre-selection behavior correctly handles the new name
+                .perform(ViewActions.pressKey(KeyEvent.KEYCODE_TAB))
+                .perform(ViewActions.click())
+                .check(matches(withSelectedText(thirdFileName)))
+
+            /*
+             * TODO - NOMERGE: check that the new name is actually used during the upload!
```
PhilLab: WIP, because
1. this verification is missing
2. The following lines uncover a bug in the implementation which needs to be fixed.

alperozturk96 (commented):
Hello. Thank you for the PR and the test.

I tested and codewise looks okay. However, better to get feedback from the designer.

@kra-mo Enhancement issue: https://github.com/nextcloud/android/issues/9215

@tobiasKaminsky 

What do you think?

kra-mo (changes requested):
There should be some more padding atop the entry since the label is quite close to the edge now, but other than that, looks good from the design side. Neat :)

PhilLab:
Recreated PR in main repository: #16298

PR #16222, opened 2025-12-30
FolderPickerActivity: show correct folder

tobiasKaminsky:


PhilLab:
@tobiasKaminsky See #16188 for a new automated test. Though I am unsure why it didn't catch the issue last time I ran it on my machine.

PR #16233, opened 2026-01-02
Screenshot tests tooling

PhilLab:

The existing documentation sent you on the wrong track, as the 'shot' dependency is excluded from the normal build, so it requires appropriate preconditions to be executed. Otherwise, the gradle task `genericDebugExecuteScreenshotTests` will not even be  available. 

- Removed left-over usages of gplay flavor for screenshot tests
- Added documentation for local execution of screenshot tests
- androidScreenshotTest now sets up the uiComparison AVD, if not present


tobiasKaminsky:
Thank you :tada:

PR #16236, opened 2026-01-02
perf: Implement streaming XML parsing and improve file upload collision handling

dmiales:
Problem:
Users with large photo and video libraries (5000+ files per folder) experience severe performance issues and OutOfMemoryError (OOM) crashes when opening folders. The application receives a massive XML file from the server and attempts to parse it entirely in memory, causing:
- Application freezes and lagging (observed on Samsung S23+)
- OOM errors during XML parsing
- OOM errors in background synchronization processes
- Data loss due to crashes during sync operations

Solution:
Rewrote the parsing block from the old ownCloud approach to a new streaming parser implementation. Additionally implemented batch processing for folders. This eliminates OOM errors and significantly improves performance and stability, even with 20,000+ photos and videos in a single folder.

Implementation details:
- Add StreamingReadFolderRemoteOperation for memory-efficient folder reading using SAX parser
- Add PropFindSaxHandler for streaming XML parsing of PROPFIND responses
- Prevent OutOfMemoryError for folders with thousands of files by avoiding full XML loading
- Integrate streaming parser into RefreshFolderOperation with fallback to legacy method
- Add batch processing for large folders (500 files per batch) in FileDataStorageManager
- Improve file upload collision handling in UploadFileOperation
  - Add content comparison for non-encrypted files when SKIP policy is used
  - Report SYNC_CONFLICT when file with same name has different content
  - Preserve old behavior for encrypted files and edge cases
- Improve logging levels and add thread name tracking for better debugging

### DIFF
--- a/app/src/main/java/com/owncloud/android/lib/resources/files/PropFindSaxHandler.java
+++ b/app/src/main/java/com/owncloud/android/lib/resources/files/PropFindSaxHandler.java
@@ -20,6 +20,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 /**
  * SAX Handler for parsing WebDAV PROPFIND responses.
  * Handles streaming XML parsing to avoid OutOfMemoryError for large folders.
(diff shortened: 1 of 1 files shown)


alperozturk96:

Please feel free to reopen it once the changes are addressed. Thank you again for your contribution.

PR #16248, opened 2026-01-06
Media: handle back press on selection correct

tobiasKaminsky:


alperozturk96 (changes requested):
Breaks this https://github.com/nextcloud/android/pull/15900 ?

PR #16249, opened 2026-01-06
fix: keep FAB hidden in sharing options

ZetaTom:
This change addresses an issue where the FAB would become visible in the file details section.

### Steps to reproduce the issue
1. Open Nextcloud client
2. Navigate to file details and open *Sharing* tab
3. Set up public link share via *Create link* with default options
4. Open *Settings* of the link share
5. Close settings by either selecting *Cancel* or *Share and copy link*

#### Expected behaviour
FAB continues to be hidden when retuning to file details/sharing tab.

#### Actual behaviour
FAB becomes visible.

---


alperozturk96 (approved):
Tested.

PR #16261, opened 2026-01-09
FileDetails: set/unset favorite also via bottom menu

tobiasKaminsky:


ZetaTom (approved):

app/src/main/java/com/nextcloud/ui/fileactions/FileAction.kt
```diff
@@ -122,8 +122,6 @@ enum class FileAction(
                 R.id.action_lock_file,
                 R.id.action_unlock_file,
                 R.id.action_edit,
-                R.id.action_favorite,
```
alperozturk96, line 125: Why we remove from file details actions?

PR #16262, opened 2026-01-09
Test that move or copy files starts in the file's parent folder

tobiasKaminsky:
Original PR: #16188

This tests the changes of https://github.com/nextcloud/android/pull/15925, where the old behavior of moving up to the root folder was changed to stay in the parent folder of the file(s) to move/copy.
The new behavior is in sync with iOS and the Web UI, hence this test ensures it doesn't break

### DIFF
--- a/app/src/androidTest/java/com/owncloud/android/ui/fragment/OCFileListFragmentStaticServerIT.kt
+++ b/app/src/androidTest/java/com/owncloud/android/ui/fragment/OCFileListFragmentStaticServerIT.kt
@@ -437,9 +437,7 @@ class OCFileListFragmentStaticServerIT : AbstractIT() {
                 activity.supportFragmentManager.executePendingTransactions()
 
                 fragment.listDirectory(testFolder, false)
-                activity.supportFragmentManager.executePendingTransactions()
                 fragment.onFileActionChosen(R.id.action_move_or_copy, setOf(testFile))
-                activity.supportFragmentManager.executePendingTransactions()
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/com/owncloud/android/ui/fragment/OCFileListFragmentStaticServerIT.kt
```diff
@@ -397,4 +415,50 @@ class OCFileListFragmentStaticServerIT : AbstractIT() {
...
+                val testFile = OCFile("${testFolder.remotePath}myImage.png").apply {
+                    parentId = testFolder.fileId
+                    activity.storageManager.saveNewFile(this)
+                }
+
+                activity.addFragment(fragment)
+                activity.supportFragmentManager.executePendingTransactions()
+
+                fragment.listDirectory(testFolder, false)
+                activity.supportFragmentManager.executePendingTransactions()
```
alperozturk96: The first `activity.supportFragmentManager.executePendingTransactions()` is needed after adding fragment other two is not needed because we are not showing another fragment via `listDirectory` or `onFileActionChosen(R.id.action_move_or_copy, setOf(testFile))`

When I tested with only first one it worked as well.

PhilLab: @alperozturk96 thanks, I removed them.

Issue #10124, opened 2022-04-28
Parallelize tasks


tobiasKaminsky:
Currently we do this one by one, but iOS uses 5 parallel upload/downloads.

- [ ] parallelize upload
- [ ] parallelize download

ayushbhanu:
@tobiasKaminsky is this enhancement still open?

BluePixel4k:
Any news on this feature? :)

Currently, uploading a batch of files takes a very long time. A parallel upload would help to improve this.

SodaWithoutSparkles:
IMHO even zipping/taring the files on the client and upload a big file would be much more efficient than the current (1 by 1) solution. The overhead of constantly starting and stopping is way more resource-intensive for small files (e.g. photos). I can see 30+ MB/s on local network uploading a single video, but less than 1/10 of it when uploading photos.

This shouldnt be _too_ hard to implement either, for all files smaller than X MB. Combining this with parallel uploads/downloads should speed things up a lot.

Peacemanguy:
Hi!
This would be really useful for regular users that want to sync all the photos that they have on nextcloud.

I am trying to sync 2000+ photos and videos and currently I am at the 3rd DAY of waiting for all files to get uploaded.
I HAVE to keep the device open and not locked otherwise the files will not upload themselves.

Having to bundle them all together in an archive is a hassle that most people will not perform, myself included.

It would be a pretty good quality of life improvement to have this feature.

MightyHelper:
Hello, any news on this?

My use case for this is that I am on a trip and I have 300+ ms latency which makes the negotiation to even start the upload take longer than the actual upload itself.

I have seen writes higher than 10MB/s for videos, but for small photos it is below 10KB/s on average since it has to negotiate one by one... And I have 2000+ photos too...

k0raph:
I've been watching this issue for over 2 years now. I come back to it every so often when I need to upload large sets of files :sweat_smile:.

Although I have basically no experience in Kotlin I thought I would give it a go

It's a bit premature but I would appreciate any feedback on this draft https://github.com/nextcloud/android/pull/16268. I've only manually tested it so far, but it's significantly faster than the sequential uploading from before

Hopefully this gives hope to the others lurking around this issue, and even if my solution is not ideal, at the very least I hope it reignites the conversation again

Edit: It's now ready for review https://github.com/nextcloud/android/pull/16268

[...]

PR #16268, opened 2026-01-12
Parallelise uploads

k0raph:
Issue: https://github.com/nextcloud/android/issues/10124

This PR aims to introduce improvements to the file upload process by parallelizing uploads using Kotlin Coroutines and a Semaphore (which I defaulted to 5 concurrent tasks) to improve throughput.

Key Changes
1. Changing BackgroundJobManagerImpl.startFilesUploadJob() to launch multiple FileUploadWorkers
2. Compatibility change in FileUploadHelper.kt
3. Unit tests for FileUploadWorker.kt (extracted out dependencies to make testing easier)
4. Unit tests for BackgroundJobManagerImpl.startFilesUploadJob()
5. Unit test for FileUploadHelper

parallel-uploads-notification-bar
parallel-uploads-uploads-list


app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
```diff
@@ -73,10 +81,15 @@ class FileUploadWorker(
         const val UPLOAD_IDS = "uploads_ids"
         const val CURRENT_BATCH_INDEX = "batch_index"
         const val TOTAL_UPLOAD_SIZE = "total_upload_size"
-        const val SHOW_SAME_FILE_ALREADY_EXISTS_NOTIFICATION = "show_same_file_already_exists_notification"
 
-        var currentUploadFileOperation: UploadFileOperation? = null
+        /**
+         * The maximum number of concurrent parallel uploads
+         */
+        const val MAX_CONCURRENT_UPLOADS = 10
```
k0raph: Should the user be given flexibility over this field?

Peacemanguy: It would be helpfull as depending on the phone specs and "bloat" on the phone it may freeze or lag substantially.

k0raph: I've added a shared preference - "Max concurrent uploads", that the user can configure.

alperozturk96 (changes requested):
Hello

Thank you for the PR 👍 

I really appreciate the time and effort you’ve put into this.

I wanted to share some feedback after reviewing the changes. Using `ConcurrentHashMap`, `Semaphore`, `atomics`, and `synchronized` blocks to achieve parallel uploads makes the implementation quite fragile and hard to maintain in the long run. It also increases complexity in an area of the codebase that is already fairly sensitive.

In fact, we can achieve parallel uploads with a much simpler change. Currently, in `BackgroundJobManagerImpl`, the `startFilesUploadJob` function uses the same worker tag together with the `ExistingWorkPolicy.APPEND_OR_REPLACE` policy. Because of this, a new worker is not used for each upload attempt; instead, uploads are queued and executed sequentially. 

By making the tag unique (for example, by appending a random value), a new worker will be started for each upload, enabling parallelism with minimal changes:

`val tag = startFileUploadJobTag(user.accountName + Random.nextInt())`

----

Introducing parallel uploads is definitely a good idea, but it’s something we should approach incrementally and carefully to ensure the overall robustness of the app. `FileUploadWorker` is a critical component: it also triggers UI updates via the broadcast manager and is observed by multiple screens. Making it parallel will likely have side effects in those areas (notifications, upload list, file listing, etc.), so we need to be mindful of that impact.

I’d suggest the following approach: 

1. Enable parallel uploads with the minimal change (unique worker tags). 

2. Observe and adjust the behavior of the caller functions, and add appropriate tests. 

3. Observe UI behavior (notifications, file list, upload list) and address any issues that arise. 

Thanks again for your contribution and initiative on this topic. :)

@ZetaTom @tobiasKaminsky  fyi.

k0raph:
> Hello
> 
> Thank you for the PR 👍
> 
> I really appreciate the time and effort you’ve put into this.
> 
> I wanted to share some feedback after reviewing the changes. Using `ConcurrentHashMap`, `Semaphore`, `atomics`, and `synchronized` blocks to achieve parallel uploads makes the implementation quite fragile and hard to maintain in the long run. It also increases complexity in an area of the codebase that is already fairly sensitive.
> 
> In fact, we can achieve parallel uploads with a much simpler change. Currently, in `BackgroundJobManagerImpl`, the `startFilesUploadJob` function uses the same worker tag together with the `ExistingWorkPolicy.APPEND_OR_REPLACE` policy. Because of this, a new worker is not used for each upload attempt; instead, uploads are queued and executed sequentially.
> 
> By making the tag unique (for example, by appending a random value), a new worker will be started for each upload, enabling parallelism with minimal changes:
> 
> `val tag = startFileUploadJobTag(user.accountName + Random.nextInt())`
> 
> Introducing parallel uploads is definitely a good idea, but it’s something we should approach incrementally and carefully to ensure the overall robustness of the app. `FileUploadWorker` is a critical component: it also triggers UI updates via the broadcast manager and is observed by multiple screens. Making it parallel will likely have side effects in those areas (notifications, upload list, file listing, etc.), so we need to be mindful of that impact.
> 
> I’d suggest the following approach:
> 
>     1. Enable parallel uploads with the minimal change (unique worker tags).
> 
>     2. Observe and adjust the behavior of the caller functions, and add appropriate tests.
> 
>     3. Observe UI behavior (notifications, file list, upload list) and address any issues that arise.
> 
> 
> Thanks again for your contribution and initiative on this topic. :)
> 
> @ZetaTom @tobiasKaminsky fyi.

Thank you @alperozturk96 for taking the time to review and I appreciate you sharing your expertise in this area. I don't disagree with any of the comments and I will happily explore your suggestion.

k0raph:
@alperozturk96 I have looked into your suggestion, and managed to get it working. It's significantly less changes and anecdotally seems to run much faster =]. Unfortunately it has had the side effects you mentioned. Notably the upload list is not reflective of the parallel nature (multiple files should be marked as uploading) - I am not sure how to address this without the ConcurrentHashMap solution I previously incorporated into FileUploadWorker. Could you please advise in this case?
broken-uploads-list-only-marks-1-uploading-at-a-time
Additionally the notifications are now also broken, the numerators are all over the place - I don't have a solution for this yet. Will investigate
broken-notifications

May I also clarify what you mean by "file list"?

kra-mo:
From a design POV, "max concurrent uploads" should definitely not be a user-configurable setting. Just pick a sensible default and stick with that. If it doesn't work for everyone, we should investigate then.

k0raph:
> From a design POV, "max concurrent uploads" should definitely not be a user-configurable setting. Just pick a sensible default and stick with that. If it doesn't work for everyone, we should investigate then.

Hey @kra-mo happy to remove the configurability and simply default, but could you help me understand why it "should definitely not be user configurable?". Seems like a useful reasonable option at a glance

Alternatively if the fear is that we are giving the user "a foot gun", we can provide a limited selection of options 1,2,5,10,20?

kra-mo:
It's against our design principles: 

> Software should get out of the way. Do things automatically instead of offering configuration options. When people ask for a setting, find out what the root of the problem is and fix that instead. Also read [Choosing our Preferences](http://ometer.com/preferences.html).

From https://docs.nextcloud.com/server/27/developer_manual/design/introduction.html

But TL;DR the more preferences you have that most users shouldn't touch, the harder it is to find ones that they do care about. If there is no reason for something to be configurable other than "why not", it's best to just not :)

kra-mo:
If there *is* a need to have different numbers of for different scenarios, it'll probably come up, and we can probably solve it for that scenario automatically, as the developers. We shouldn't put the burden on the user.

k0raph:
@kra-mo great response, consider me convinced. I will revert the changes and default

k0raph:
@alperozturk96 @kra-mo @tobiasKaminsky could I get a re-review

tobiasKaminsky:
Hello,

thank you for this valuable PR.
Since this is a quite big (and "severe") change, I am a bit of hesitating to merge this soon.
Imagine if in some corner cases (like slow network) none of the parallel uploads finish.

Thus I think we need to have a way to test/make sure that it is robust in many cases.

We have our dev app, which we could use for that.
So I suggest to check if it is possible to bound this to "is_beta" in setup.xml?

This way we can enable it for dev versions, monitor the replies and then at some point enable it.
Before we merge it, we should also do intensive tests on our own.

k0raph:
> Hello,
> 
> thank you for this valuable PR. Since this is a quite big (and "severe") change, I am a bit of hesitating to merge this soon. Imagine if in some corner cases (like slow network) none of the parallel uploads finish.
> 
> Thus I think we need to have a way to test/make sure that it is robust in many cases.
> 
> We have our dev app, which we could use for that. So I suggest to check if it is possible to bound this to "is_beta" in setup.xml?
> 
> This way we can enable it for dev versions, monitor the replies and then at some point enable it. Before we merge it, we should also do intensive tests on our own.

Hi @tobiasKaminsky I'm sure it's possible, I am not familiar with the process here, it would be nice to have a clear plan to avoid this feature getting stuck in limbo. By any chance do we have any documentation on the beta/dev release process?

PhilLab:
@k0raph regarding the is_beta, which Tobias mentioned: Can you guard the new behavior with a `if(getResources().getBoolean(R.bool.is_beta)) { //.. your awesome new behavior` ?

It is set for the qa build flavor ([here](https://github.com/nextcloud/android/blob/master/app/src/qa/res/values/setup.xml#L10)). This build flavor is e.g. compiled pre-merge, and also plays a role in some  automated tests.

PR #16273, opened 2026-01-12
fix(auto-upload): handle existing sync conflicts

alperozturk96:

### Issue: https://github.com/nextcloud/android/issues/16126#issuecomment-3708201252

When a sync conflict occurs during auto-upload, the client repeatedly retries uploading the same file. This results in the user seeing the same conflict notification multiple times, even though the conflict already exists and requires manual resolution. 

### Changes

After the first auto-upload attempt that results in a sync conflict, the file is marked as handled by the auto-upload system. This prevents further automatic retries for the same file. The conflicted upload remains visible in the Uploads list, where the user can manually resolve the conflict or retry the upload if desired.

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/autoUpload/AutoUploadWorker.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/autoUpload/AutoUploadWorker.kt
@@ -401,7 +401,11 @@ class AutoUploadWorker(
         uploadsStorageManager.removeUpload(upload)
     }
 
-    private fun createEntityAndUpload(user: User, localPath: String, remotePath: String): Pair<UploadEntity, OCUpload>? {
+    private fun createEntityAndUpload(
+        user: User,
+        localPath: String,
+        remotePath: String
+    ): Pair<UploadEntity, OCUpload>? {
         val (needsCharging, needsWifi, uploadAction) = getUploadSettings(syncedFolder)
         Log_OC.d(TAG, "creating oc upload for ${user.accountName}")
 


app/src/main/java/com/nextcloud/client/jobs/autoUpload/AutoUploadWorker.kt
```diff
@@ -393,7 +401,11 @@ class AutoUploadWorker(
         uploadsStorageManager.removeUpload(upload)
     }
 
-    private fun createEntityAndUpload(user: User, localPath: String, remotePath: String): Pair<UploadEntity, OCUpload> {
+    private fun createEntityAndUpload(
+        user: User,
+        localPath: String,
+        remotePath: String
+    ): Pair<UploadEntity, OCUpload>? {
```
ZetaTom, line 408: Returning `null` in this case means that a prior upload failed due to a conflict. In my opinion we shouldn't introduce overall nullability to mean one specific thing. I would prefer if either the `OCUpload` becomes nullable, or if we return `UploadEntity` as is and check for `uploadEntity?.lastResult` in the calling function instead. I think that this would improve readability in the long term.

alperozturk96:
> If I understand correctly, this means that any files scheduled for Auto Upload which failed to upload due to a conflict, will not be rescheduled.
> 
> This improves usability, as users will not receive _repeated_ notifications or prompts to manually resolve conflicts for Auto Upload.
> 
> However, it also means that no further attempts will be made once an upload conflict is detected. Conflicts may also be resolved on the server, e.g. by renaming or deleting the file in conflict.
> 
> We need to ask ourselves, if this is the right approach. Another option would be to silently retry the upload without telling the user in case it fails again. In either case, the original notification or entry in the failed uploads list would still remain.
> 
> If we decide that failed uploads should not be reattempted, this would be an acceptable solution. I've left a comment suggesting a minor improvement.

“Conflicts may also be resolved on the server.” IMO, this scenario is not common. If a sync conflict is not resolved, retrying the upload will only cause unnecessary load. The app will silently hide it and keep retrying in the background. For each attempt on that specific folder, the app will try to upload a file that already has a conflict, which is very unlikely to be resolved automatically (due to changes on remote). 

Since users can manage these from the Upload List screen, this PR's approach makes more sense and is the right one to take. @tobiasKaminsky what do you think?

app/src/main/java/com/nextcloud/client/jobs/autoUpload/AutoUploadWorker.kt
alperozturk96: I agree, and...

The `createEntityAndUpload` function should either create both the entity and the upload together or not create them at all. We should avoid partially successful executions; it should either succeed completely or fail entirely.

One approach would be to only pass items that are eligible for mapping, ensuring the function always returns valid values.

Alternatively, we could use a sealed class or a `kotlin.Result` wrapper so the function would never return `null`.

The current approach is okay, and this logic is only used inside the worker; it is not a general function, so it’s fine to keep it as is for now. In the future, we can improve this as needed, for example, if another condition prevents mapping and we have to handle it differently in the caller function.

tobiasKaminsky:
Yes, no re-schedule.
We wait for any user interaction.

(If people then complain that they forgot to resolve the conflict, we could add a recurring notification like "you have pending conflicts in your upload queue". But this is out of scope for that PR.

So fine from my side ✔

PR #16274, opened 2026-01-12
do not show all sync conflict notifications during app launch

alperozturk96:

### Issue: https://github.com/nextcloud/android/issues/16126#issuecomment-3708201252

When a user encounters a sync conflict through manual or automatic uploads, the `retryUploads` function currently triggers a separate notification for each conflicted file at app launch. This can result in multiple notifications appearing at app launch, which can be overwhelming and annoying. 

### Changes 

This PR ensures that only a single notification is shown for sync conflicts **at app launch**. Tapping the notification opens the Uploads screen, allowing the user to review and resolve conflicts. This eliminates the repetitive notifications during app launch while keeping the user informed.

Note: User would still see sync conflict message for each file after upload attempt via auto or manual upload.

### This will be shown instead of multiple sync conflict notification at app launch

Screenshot_20260112_093332

### DIFF
--- a/app/src/main/java/com/nextcloud/client/notifications/AppWideNotificationManager.kt
+++ b/app/src/main/java/com/nextcloud/client/notifications/AppWideNotificationManager.kt
@@ -12,7 +12,6 @@ import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
 import android.content.pm.PackageManager
-import androidx.annotation.RequiresPermission
 import androidx.core.app.ActivityCompat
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
Very minor nitpick:
In screenshot we mix uploads and upload.

I think it is overkill to first compute if we have more than one conflict, so we should always use plural:
File file upload conflict"s"
Upload conflicts detected. Open uploads to resolve. (with .)
Resolve conflict"s"

PR #16276, opened 2026-01-12
fix(auto upload): file skip check

alperozturk96:

### Issue

`lastScanTimestampMs != -1L && lastModified < lastScanTimestampMs` 

This skip logic must be applied for only `fileSentForUpload` is true, otherwise it will skip newly added files as well.

### How to test?

1. Create a folder that has images from local file manager
2. Configure that folder for auto-upload
3. Auto upload will upload all images in it
4. Add new images from local file manager
5. Wait for next auto-upload attempt

### Result

Screenshot 2026-01-12 at 12 27 33

Screenshot 2026-01-12 at 12 26 57

### DIFF
--- a/app/src/test/java/com/owncloud/android/utils/AutoUploadHelperTest.kt
+++ b/app/src/test/java/com/owncloud/android/utils/AutoUploadHelperTest.kt
@@ -132,10 +132,10 @@ class AutoUploadHelperTest {
             type = MediaFolderType.CUSTOM
         )
 
-        val shouldSkipOldFile = folder.shouldSkipFile(oldFile, oldFileLastModified, null)
+        val shouldSkipOldFile = folder.shouldSkipFile(oldFile, oldFileLastModified, null, true)
         assertTrue(shouldSkipOldFile)
 
-        val shouldSkipNewFile = folder.shouldSkipFile(newFile, currentTime, null)
+        val shouldSkipNewFile = folder.shouldSkipFile(newFile, currentTime, null, false)
         assertTrue(!shouldSkipNewFile)
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nextcloud/client/jobs/autoUpload/FileSystemRepository.kt
```diff
@@ -159,20 +159,19 @@ class FileSystemRepository(private val dao: FileSystemDao, private val context:
...
+            }
+
             val fileModified = (lastModified ?: file.lastModified())
-            val shouldSkipFileBasedOnFolderSettings = syncedFolder.shouldSkipFile(file, fileModified, creationTime)
-            if (shouldSkipFileBasedOnFolderSettings) {
+            if (syncedFolder.shouldSkipFile(file, fileModified, creationTime, fileSentForUpload)) {
                 return
             }
 
-            val entity = dao.getFileByPathAndFolder(localPath, syncedFolder.id.toString())
```
alperozturk96, line 168: `getAutoUploadFilesEntities` already returns file needs to be uploaded and if we skip when `fileSentForUpload` is true, changed files will not be uploaded.

app/src/main/java/com/nextcloud/client/jobs/autoUpload/FileSystemRepository.kt
```diff
@@ -159,20 +159,19 @@ class FileSystemRepository(private val dao: FileSystemDao, private val context:
...
+            val entity = dao.getFileByPathAndFolder(localPath, syncedFolder.id.toString())
+            val fileSentForUpload = (entity != null && entity.fileSentForUpload == 1)
+            if (fileSentForUpload) {
+                Log_OC.d(TAG, "File was sent for upload, checking if it changed...")
+            }
+
             val fileModified = (lastModified ?: file.lastModified())
-            val shouldSkipFileBasedOnFolderSettings = syncedFolder.shouldSkipFile(file, fileModified, creationTime)
-            if (shouldSkipFileBasedOnFolderSettings) {
+            if (syncedFolder.shouldSkipFile(file, fileModified, creationTime, fileSentForUpload)) {
```
alperozturk96, line 169: When I copied or moved the file to another directory, the `lastModified` value didn’t change, and I couldn’t find a more reliable alternative. 

I think having this PR is better than not having it. As we discussed, though, it can be interpreted differently depending on the context and where you’re looking from.

@tobiasKaminsky

PR #16282, opened 2026-01-13
fix(auto-upload): show notification only when conditions are met

alperozturk96:

### Issue

`trySetForeground()` called before ensuring conditions, this causing to see multiple notifications. Conditions need to be checked first.


app/src/main/java/com/nextcloud/client/jobs/folderDownload/FolderDownloadWorkerNotificationManager.kt
```diff
@@ -109,13 +107,6 @@ class FolderDownloadWorkerNotificationManager(private val context: Context, view
         return getForegroundInfo(notification)
     }
 
-    fun getForegroundInfo(notification: Notification): ForegroundInfo =
```
alperozturk96, line 112: Since this is needed for other workers, we can move this logic to the `WorkerNotificationManager`

ZetaTom (approved):
Tested on AOSP Android 11 using a Pixel 2 emulator. Works while app is in foreground and background.

PR #16283, opened 2026-01-13
Respect avatar border for TextDrawable

AndyScherzinger:
...since they can't resize automatically, so the border must be calculated into the radius.

Follow-up to #16227

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before|after

For testing have a file shared with you by a user that got deactivated, so you don't get an avatar from the server. The fallback is a TextDrawable but that doesn't resize, so in case of the avatar group the border needs to be calculated into the radius, else the avatar is (slightly) to big.

### DIFF
--- a/app/src/main/java/com/owncloud/android/utils/DisplayUtils.java
+++ b/app/src/main/java/com/owncloud/android/utils/DisplayUtils.java
@@ -460,6 +460,7 @@ public final class DisplayUtils {
      * @param avatarRadius the avatar radius
      * @param resources    reference for density information
      * @param callContext  which context is called to set the generated avatar
+     * @param context      general context
      */
     public static void setAvatar(@NonNull User user,
                                  @NonNull String userId,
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -469,6 +470,30 @@ public static void setAvatar(@NonNull User user,
...
+     */
+    public static void setAvatar(@NonNull User user,
+                                 @NonNull String userId,
+                                 String displayName,
+                                 AvatarGenerationListener listener,
+                                 float avatarRadius,
+                                 Resources resources,
+                                 Object callContext,
+                                 Context context,
+                                 int avatarBoarder) {
```
alperozturk96:
```suggestion
                                 int avatarBorder) {
```

app/src/main/java/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -495,7 +520,8 @@ public static void setAvatar(@NonNull User user,
             // if no one exists, show colored icon with initial char
             if (avatar == null) {
                 try {
-                    avatar = TextDrawable.createAvatarByUserId(displayName, avatarRadius);
+                    avatar = TextDrawable.createAvatarByUserId(displayName,
+                                                               (avatarRadius - avatarBoarder));
```
alperozturk96:
```suggestion
                                                               (avatarRadius - avatarBorder));
```

app/src/main/java/com/owncloud/android/utils/DisplayUtils.java
```diff
@@ -469,6 +470,30 @@ public static void setAvatar(@NonNull User user,
...
+     * fetches and sets the avatar of the given account in the passed callContext
+     *
+     * @param user           the account to be used to connect to server
+     * @param userId         the userId which avatar should be set
+     * @param displayName    displayName used to generate avatar with first char, only used as fallback
+     * @param avatarRadius   the avatar radius
+     * @param resources      reference for density information
+     * @param callContext    which context is called to set the generated avatar
+     * @param context        general context
+     * @param avatarBoarder  value in case the avatar has a border, like in the case of the AvatarGroupLayout
```
alperozturk96:
```suggestion
     * @param avatarBorder  value in case the avatar has a border, like in the case of the AvatarGroupLayout
```

AndyScherzinger:
Fixed the typo @alperozturk96 - thanks for the review 🙏

alperozturk96 (approved):

PR #16284, opened 2026-01-13
fix(navigation): item highlight, content displaying

alperozturk96:

### Issue

- When the user navigates back from the Trashbin screen, the previously selected item's text remains bold. 

- When the user goes to Settings, presses back, then navigates to the Trash Bin and finally to Shared, the Trash Bin item is highlighted instead of Shared, and displayed content belongs to All files instead of Shared.

Screenshot_20260113_144313

### Changes

No need to `colorNavigationView` every time when an item is selected
Unset all navigation items, including sub-menu items
Don't expose the menu item id
Force child activities of the `DrawerActivity` to implement and handle menu item ID.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
@@ -1328,7 +1328,7 @@ class FileDisplayActivity :
 
         super.onResume()
 
-        if (ocFileListFragment is SharedListFragment || ocFileListFragment?.isSearchFragment == true) {
+        if (ocFileListFragment?.isSearchFragment == true) {
             ocFileListFragment?.setSearchArgs(ocFileListFragment?.arguments)
         }
         highlightNavigationViewItem(menuItemId)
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -911,7 +936,6 @@ private void updateQuotaLink() {
                         externalWebViewIntent.putExtra(ExternalSiteWebView.EXTRA_TITLE, firstQuota.getName());
                         externalWebViewIntent.putExtra(ExternalSiteWebView.EXTRA_URL, firstQuota.getUrl());
                         externalWebViewIntent.putExtra(ExternalSiteWebView.EXTRA_SHOW_SIDEBAR, true);
-                        menuItemId = Menu.NONE;
```
alperozturk96: Unnecessary logic

app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -877,22 +918,6 @@ private void setQuotaInformation(long usedSpace, long totalSpace, int relative,
         showQuota(true);
     }
 
-    private void unsetAllDrawerMenuItems() {
```
alperozturk96: Unnecessary logic

PR #16286, opened 2026-01-14
Minor improvements for automated testing

tobiasKaminsky:
Original PR: #15424

- Minor improvements of testing documentation
- Instrumentation tests now automatically grab the notifications permission
- DownloadIT now works for all build variants
- Fixed AbstractIT deleting the wrong account
- AbstractOnServerIT file deletion now ignores non-empty encrypted folders
- Fixed some static warnings in AbstractIT and AbstractOnServerIT
- Fixed incorrect lookup of gradle properties file

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -136,12 +136,6 @@ android {
             debug {
                 enableUnitTestCoverage = project.hasProperty("coverage")
                 resConfigs("xxxhdpi")
-
-                buildConfigField(
-                    "String",
-                    "NC_TEST_SERVER_DATA_STRING",
-                    "\"nc://login/user:${ncTestServerUsername}&password:${ncTestServerPassword}&server:${ncTestServerBaseUrl}\""
-                )
             }
         }
 
--- a/app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -716,22 +716,6 @@ public class AuthenticatorActivity extends AccountAuthenticatorActivity
         } else {
             accountSetupBinding.scanQr.setVisibility(View.GONE);
         }
-
-        addDebugLogin();
-    }
-
-    private void addDebugLogin() {
-        if (BuildConfig.DEBUG) {
-            try {
-                accountSetupBinding.thumbnail.setOnLongClickListener(v -> {
-                    final String dataString = BuildConfig.NC_TEST_SERVER_DATA_STRING;
-                    parseAndLoginFromWebView(dataString);
-                    return false;
-                });
-            } catch (Throwable t) {
-                Log_OC.w(TAG, "Test server data string not available in this build");
-            }
-        }
     }
 
     /**


app/build.gradle.kts
```diff
@@ -75,7 +75,7 @@ val ndkEnv = buildMap {
 }
 
 val configProps = Properties().apply {
-    val file = rootProject.file(".gradle/config.properties")
+    val file = rootProject.file("gradle.properties")
```
ZetaTom: We'll probably have to adjust our internal build infrastructure to use the new path.

alperozturk96, line 78: Hello

Thank you for the PR. It is really helpful.

This change affects the existing debug login mechanism implemented in `AuthenticatorActivity.addDebugLogin`.

**Side note:**
The debug login allows developers to quickly sign in by long-pressing the app icon on the login screen.
The `.gradle/config.properties` file is local, so each developer can maintain their own test credentials.

**Suggestion:**
We can introduce a new value for the debug login, for example:

`val debugLoginProps = ...`

This should point to the local `.gradle/config.properties` local file.

Meanwhile, `configProps` should continue to point to gradle.properties so that `testInstrumentationRunnerArgument `can correctly consume those values.

-----

Under what conditions this can be risky? Local file is not bundled with any apk.

If you could provide an example that clearly demonstrates how this could cause an issue/risk, we might consider removing it. Otherwise, it only provides a shortcut to a file on the developer's computer and is only triggered in the debug environment (`BuildConfig.DEBUG`).

If you have an idea for a safer approach, that would be great as well. Nevertheless good point, thanks 👍 

@PhilLab 

Old reference: https://github.com/nextcloud/android/pull/15424#discussion_r2616410414

PhilLab: @alperozturk96 
Without the patch from this PR here, the existing `gradle.properties` is simply unused, even though it contains explicit guidance on how to fill it for the automated tests. But the data never reaches the automated tests.

Why is it important to parse the data from `.gradle/config.properties` ? Why not from the `gradle.properties`, for which a ready-to-use file with dummy values and documentation already exists?

If you want to have a _separate_ parsing for different login information for debugging, then we need two parsing paths, pointing two two distinct files. E.g. `gradle.properties` for the automated tests and something clearly named as such, e.g.`debuglogin.gradle.properties`

-----------------

Regarding riskiness: I think it is unexpected to ship login data with any apk build. Unsuspecting developers might share a test build from their local system with others, which also contains this login data. It can easily be retrieved by decompiling the apk:
image

Or simply access this by long-pressing the icon on the login screen, as you just explained (a behavior which was entirely unknown to me).

So if we really want to keep it, then the file where this is parsed from must be a) separate from the test credential files and b) very clearly and explicitly explained - e.g. by a describing filename and a comment inside, explaining what entering credentials there will do.

-----------------

Ironically, my change now does **increase** the risk I am talking about. It does fix the data flow for the automated tests, so it is necessary. But at the same time, it injects these credentials in every local build now.

alperozturk96: `Why is it important to parse data from .gradle/config.properties?`, `Why not use gradle.properties instead?`

`gradle.properties` is not a local-only file but `.gradle/config.properties` only stored in your device. If we store test credentials there, they will be visible to everyone.

---

I also verified the debug APK as shown earlier: it does pick up values from the `configProps`. However, since these values are currently compiled into the debug APK, they can be extracted.

It would be better to remove the debug login mechanism in this PR. We can reintroduce it later in a separate PR using a more safer approach. 

With this change, the `debug {}` block will no longer define any `BuildConfig` fields, and values will not be extractable from the debug APK.

PhilLab: @alperozturk96 okay, sorry to hear that I jeopardized the debug login feature by my patch... As I am not entirely clear how the requested change to remove the feature temporarily would look like - could you push a commit for this on the branch here?

alperozturk96: No. Your finding is really good. Great point. 👍

alperozturk96: https://github.com/nextcloud/android/pull/16286/changes/eff8e77cb1b69512dc3d73ff75860b5a6b43fef4

PhilLab: @alperozturk96 thanks, looks good!

PR #16297, opened 2026-01-15
Bugfix and automated tests for #15636: deleting from slideshow

PhilLab:
Original PR: #16184

This tests the deletion behavior when deleting from slide show, fixed in #15636. It also fixes another bug which was uncovered during the automated tests, where the view pager was not correctly updated.

This PR requires #16296 to be merged first, as the automated tests also stumble upon the same timing issue fixed (and tested!) in said PR.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
@@ -122,7 +122,6 @@ import com.owncloud.android.ui.dialog.SendShareDialog.SendShareDialogDownloader
 import com.owncloud.android.ui.dialog.SortingOrderDialogFragment.OnSortingOrderListener
 import com.owncloud.android.ui.dialog.StoragePermissionDialogFragment
 import com.owncloud.android.ui.dialog.TermsOfServiceDialog
-import com.owncloud.android.ui.events.FilesRefreshEvent
 import com.owncloud.android.ui.events.SearchEvent
 import com.owncloud.android.ui.events.SyncEventFinished
 import com.owncloud.android.ui.events.TokenPushEvent
@@ -187,6 +186,7 @@ class FileDisplayActivity :
     OnEnforceableRefreshListener,
     OnSortingOrderListener,
     SendShareDialogDownloader,
+    OnFilesRemovedListener,
     Injectable {
     private lateinit var binding: FilesBinding
 
@@ -1455,6 +1455,12 @@ class FileDisplayActivity :
         }
     }
     // endregion
+
+    override fun onStop() {
+        Log_OC.v(TAG, "onStop()")
+        unregisterReceivers()
+        super.onStop()
+    }
 
     override fun onSortingOrderChosen(selection: FileSortOrder?) {
         val ocFileListFragment = this.listOfFilesFragment
@@ -2764,26 +2770,12 @@ class FileDisplayActivity :
 
         registerReceivers()
 
-        // Register for the FilesRefreshEvent, only if not already registered
-        if (!EventBus.getDefault().isRegistered(this)) {
-            EventBus.getDefault().register(this)
-        }
-
         if (SettingsActivity.isBackPressed) {
             Log_OC.d(TAG, "User returned from settings activity, skipping reset content logic")
             return
         }
 
         initFile()
-    }
-
-    public override fun onStop() {
-        Log_OC.v(TAG, "onStop()")
-        if (EventBus.getDefault().isRegistered(this)) {
-            EventBus.getDefault().unregister(this)
-        }
-        unregisterReceivers()
-        super.onStop()
     }
 
     private fun initFile() {
@@ -3036,13 +3028,9 @@ class FileDisplayActivity :
         })
     }
 
-    // region EventBus
-    @Suppress("unused")
-    @Subscribe(threadMode = ThreadMode.MAIN)
-    fun onFilesRefreshNeeded(event: FilesRefreshEvent) {
+    override fun onFilesRemoved() {
         refreshCurrentDirectory()
     }
-    // endregion
 
     private fun handleEcosystemIntent(intent: Intent?) {
         ecosystemManager.receiveAccount(
--- a/app/src/main/java/com/owncloud/android/ui/activity/OnFilesRemovedListener.kt
+++ b/app/src/main/java/com/owncloud/android/ui/activity/OnFilesRemovedListener.kt
@@ -5,9 +5,8 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
-package com.owncloud.android.ui.events
+package com.owncloud.android.ui.activity
 
-/**
- * Event denoting the need to refresh the files displayed in the current view
- */
-class FilesRefreshEvent
+interface OnFilesRemovedListener {
+    fun onFilesRemoved()
+}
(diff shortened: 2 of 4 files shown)


PhilLab:
@tobiasKaminsky @alperozturk96  Now I am concerned. I checked this PR's drone run and saw that the PreviewImageActivityIT.kt was checked out, but apparently did not actually run.

I then created a PR to provoke a failure, https://github.com/nextcloud/android/pull/16329#discussion_r2711171095, and its drone test ran through: https://drone.nextcloud.com/nextcloud/android/27697

Where is the error?

**Update** Seems this is a more general problem, as many more provoked failures are not catched in the CI. I created #16350 to track this. So it is more or less independent of the PR here, as it is a more widespread issue.

alperozturk96 (changes requested):
Hello 

Thank you for the PR and for the updates. 

I tested it with the latest master branch. I noticed that when a user deletes a picture from the preview, the next image appears instead of the file listing. 

Also, the `onRemoveFileOperationFinish` function updates the adapter and should trigger the UI updates, rather than handling it directly in `RemoveFilesDialogFragment.kt‎.` 

As a side note, not directly related to this PR: the line `fda?.refreshCurrentDirectory()` may be unnecessary, since `onRemoveFileOperationFinish` already performs the update. 

I suggest to remove unnecessary lines (`refreshDirectory`, `initPager`)  from `RemoveFilesDialogFragment.kt‎` , test again and add tests for these scenarios.

Thanks again for your work on this.

PhilLab:
@alperozturk96 
> I tested it with the latest master branch. I noticed that when a user deletes a picture from the preview, the next image appears not file listing as expected so I couldn't able to reproduce the bug.

Exactly. This is because this PR is a follow-up of the already merged bug fix, see the PR description:
> > This tests the deletion behavior when deleting from slide show, fixed in #15636.

I created the new PR to "guard" my fix :sweat_smile: against future regressions. If you want to experience the bug, you would need to revert the mentioned PR. Then, you'll also see the automated `PreviewImageActivityIT` fail.

And you can reproduce 
>> another bug which was uncovered during the automated tests, where the view pager was not correctly updated.

by checking out this branch here, locally reverting 74a9a3d4bc773e7e25bb154d6b3958ba522bf1e9 and then run the `PreviewImageActivityIT`. It will then fail and prove the existence of another bug.

Regarding your code suggestion: Trusting on `onRemoveFileOperationFinish()` doesn't work, I tried. It is not called for local only deletions, because it is only called from `FileDisplayActivity.onRemoteOperationFinish()`.
So I matched the existing `fda?.refreshCurrentDirectory()` and added `pia?.initViewPager()` alongside.

I'd prefer to keep the changes as small as possible - a refactoring to have the file operation finish callback work for every type of deletion feels like a significantly larger scope. Maybe a follow-up of this PR?

alperozturk96:
> @alperozturk96
> 
> > I tested it with the latest master branch. I noticed that when a user deletes a picture from the preview, the next image appears not file listing as expected so I couldn't able to reproduce the bug.
> 
> Exactly. This is because this PR is a follow-up of the already merged bug fix, see the PR description:
> 
> > > This tests the deletion behavior when deleting from slide show, fixed in #15636.
> 
> I created the new PR to "guard" my fix 😅 against future regressions. If you want to experience the bug, you would need to revert the mentioned PR. Then, you'll also see the automated `PreviewImageActivityIT` fail.
> 
> And you can reproduce
> 
> > > another bug which was uncovered during the automated tests, where the view pager was not correctly updated.
> 
> by checking out this branch here, locally reverting [74a9a3d](https://github.com/nextcloud/android/commit/74a9a3d4bc773e7e25bb154d6b3958ba522bf1e9) and then run the `PreviewImageActivityIT`. It will then fail and prove the existence of another bug.
> 
> Regarding your code suggestion: Trusting on `onRemoveFileOperationFinish()` doesn't work, I tried. It is not called for local only deletions, because it is only called from `FileDisplayActivity.onRemoteOperationFinish()`. So I matched the existing `fda?.refreshCurrentDirectory()` and added `pia?.initViewPager()` alongside.
> 
> I'd prefer to keep the changes as small as possible - a refactoring to have the file operation finish callback work for every type of deletion feels like a significantly larger scope. Maybe a follow-up of this PR?

I think adding a guard here isn’t the right direction. The fix should address the underlying bug directly, without needing additional defensive logic.

`So I matched the existing fda?.refreshCurrentDirectory() and added pia?.initViewPager() alongside.`

That’s actually what concerns me. It introduces more conditional and tightly coupled behavior into a dialog fragment.

`RemoveFilesDialogFragment` is a dialog fragment, and its responsibility should remain limited to presenting the dialog. Expanding its role increases complexity and makes the flow harder to reason about.

If `onRemoteOperationFinish` currently doesn’t handle local-only deletions, a solution would be for `RemoveFilesDialogFragment` to implement an interface to handle that case explicitly. That would solve the issue with fewer changes and clearer ownership. 

This approach would keep the architecture simpler and make the fix easier to maintain going forward.

PhilLab:
> I think adding a guard here isn’t the right direction. The fix should address the underlying bug directly, without needing additional defensive logic.

@alperozturk96 okay, take a look I just added two commits:
1. fix(slideshow) Dependency-inject ConnectivityService
2. fix(slideshow): Use EventBus to notify activies of file deletion

I used the EventBus, as I saw it being used elsewhere in the repository, so I would use the same here.

alperozturk96 (changes requested):
Hello

Thank you very much for your effort here, we truly appreciate the work you’ve put into this. Improving testability is very valuable for us, and your contribution definitely helps move things in that direction.

That said, we’ve previously aligned on gradually eliminating `EventBus` and `LocalBroadcastManager` usage over time. This PR appears to introduce additional usages, though I completely understand that you are not aware of it. @tobiasKaminsky  fyi.

Please correct me if I’m misunderstanding, but as I see it, this PR does not fix a new issue and instead focuses on adding tests and making the current logic testable. Is that correct? As far as I can tell, the related bug was already addressed in [https://github.com/nextcloud/android/pull/15669](https://github.com/nextcloud/android/pull/15669), and the loading dialog timing issue was handled in [https://github.com/nextcloud/android/pull/16296](https://github.com/nextcloud/android/pull/16296).

If that is correct, I would suggest that we first align on how we want to approach improving testability in these areas without introducing new `EventBus` dependencies. I’d be very happy to hear other opinions and discuss possible alternatives so we can decide on the most appropriate approach.

If this PR does address a specific bug that is not covered by the previous fixes, could you please share the steps to reproduce it? That would help us better understand the context and evaluate the changes more accurately.

Once again, thank you for your contribution. It’s great to see the focus on improving testability, and we genuinely appreciate the effort you’ve put into this.

PhilLab:
@alperozturk96 the second commit
```
fix(slideshow): After file deletion, properly refresh view pager

Otherwise, the localOnly file deletion will not update the UI.
```
does fix a real issue with the local-only file deletion. This can be reproduced by running the test without the fix (checkout the first commit)

PhilLab:
@alperozturk96 take a look, I now replaced the EventBus with an interface.

app/src/main/java/com/owncloud/android/ui/activity/FilesRefreshNeededListener.kt
```diff
@@ -0,0 +1,12 @@
+/*
+ * Nextcloud - Android Client
+ *
+ * SPDX-FileCopyrightText: 2026 Your Name <your@email.com>
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+package com.owncloud.android.ui.activity
+
+interface FilesRefreshNeededListener {
```
alperozturk96: Could you please update the license header, as well as the interface name and its function?
Once that’s done and the PR looks good, I’ll run one more round of testing and we can proceed with merging.

Thank you


```
interface OnFilesRemovedListener {
    fun onFilesRemoved()
}
```

PhilLab: @alperozturk96 sure, I amended the last commit, please take a look.

PR #16298, opened 2026-01-15
When uploading a single file, offer input field to change file name

### LINKED ISSUE #9215: when upload a file alllow modify file name

PhilLab:
Original PR: #16215

This streamlines the upload from files, e.g. from email attachments.
Closes #9215.
Also offers a convenient pre-selection of the filename without extension when entering the input field, as well as file name validation.
| New input field | No input field (e.g multi file upload). Also old behavior |
|-|-|
| Screenshot_20260102_121002 | Screenshot_20260102_121341 |

Recording of the automated test:
(The upload failure in the end is expected, as there is no server mocked)

https://github.com/user-attachments/assets/a15ad2ee-87c3-46b2-8473-1701e3ba1d3a

### Technical details

1. The approach via the listener mFileDisplayNameTransformer was chosen because this allows a potential extension to multiple files (with multiple input fields). Alternatively, ReceiveExternalFilesActivity#uploadFile() could've been used, which requires the storage to a temporary file first
2. nextFocusForward and nextFocusDown are both needed to support navigation with the Tab key as well as the Enter key.
3. The automated test is not yet verifying that the upload is actually using the user-provided name. I attempted to mock parts of the involved utility classes, but spent an inappropriate ammount of time without achieving anything.
4. The automated test also changes the two screenshots, because the test files are prepared in a `@Before` method which apply to all tests. Screenshot tests are broken on master right now, so I did not feel confident to provide updated screenshots with no way of verifying them. Either they need to be updated later, or the tests should be separated in two files

### DIFF
--- a/app/src/androidTest/java/com/nextcloud/utils/FileExtensionTest.kt
+++ b/app/src/androidTest/java/com/nextcloud/utils/FileExtensionTest.kt
@@ -12,6 +12,7 @@ import junit.framework.TestCase.assertFalse
 import junit.framework.TestCase.assertTrue
 import org.junit.Test
 
+@Suppress("TooManyFunctions")
 class FileExtensionTest {
 
     @Test
--- a/app/src/main/java/com/nextcloud/utils/extensions/StringExtensions.kt
+++ b/app/src/main/java/com/nextcloud/utils/extensions/StringExtensions.kt
@@ -38,7 +38,7 @@ fun String?.eTagChanged(eTagOnServer: String?): Boolean {
 fun String.extension(): String {
     val lastDot = lastIndexOf('.')
 
-    //  return empty string for filenames like ".gitignore"
+    // return empty string for filenames like ".gitignore"
     if (lastDot <= 0 || lastDot == length - 1) {
         return ""
     }


PhilLab:
@kra-mo I recreated the PR in the main repo now, so the automated tests are executed.

Regarding your [comment](https://github.com/nextcloud/android/pull/16215#pullrequestreview-3655363545):
> There should be some more padding atop the entry since the label is quite close to the edge now, but other than that, looks good from the design side. Neat :)

I added one commit which introduces padding at the top. See the comparison:

| No padding | Padding (standard_half_margin) |
|-|-|
| image | image |

kra-mo (approved):

PhilLab:
@alperozturk96 your comment on the old PR ([link](https://github.com/nextcloud/android/pull/16215#pullrequestreview-3650859341)) said
> I tested and codewise looks okay. However, better to get feedback from the designer.

Given that we have positive feedback from @kra-mo above, can we merge this now?

alperozturk96:
@tobiasKaminsky Could you please check this?

PhilLab:
@tobiasKaminsky can this be merged?

app/src/main/java/com/owncloud/android/ui/activity/ReceiveExternalFilesActivity.java
```diff
@@ -322,6 +334,53 @@ public void selectFile(OCFile file) {
...
 
+    @Override
+    public void afterTextChanged(Editable editable) {}
+
+    @Override
+    public void beforeTextChanged(CharSequence charSequence, int i, int i1, int i2) {}
+
+    // TODO: this is copy-paste from RenameFileDialogFragment.kt
+    @Override
+    public void onTextChanged(CharSequence s, int start, int before, int count) {
```
alperozturk96: Could you please de-duplicate this logic? Since both classes using similar logic we can create an interface and implement it via different class then `RenameFileDialogFragment`, `ReceiveExternalFilesActivity` just passes the UI elements to that class?

PhilLab: @alperozturk96 sure. I now introduced `FileNameTextWatcher.kt` to deduplicate this, please take a look

app/src/main/java/com/nextcloud/utils/fileNameValidator/FileNameValidator.kt
```diff
@@ -147,6 +148,19 @@ object FileNameValidator {
...
     }
 
+    /**
+     * @return True, if the extension of both filenames is different. If either filename is null, function returns false
+     */
+    fun isExtensionChanged(previousFileName: String?, newFileName: String?): Boolean {
+        if (previousFileName == null || newFileName == null) {
+            return false
+        }
+        val previousExtension = File(previousFileName).extension
```
alperozturk96: Please use following extension and check nullability of the file.

`fun String.toFile(): File?`

app/src/main/java/com/owncloud/android/ui/dialog/RenameFileDialogFragment.kt
```diff
@@ -99,7 +95,28 @@ class RenameFileDialogFragment :
             fileNames?.add(file.fileName)
         }
 
-        binding.userInput.addTextChangedListener(this)
+        binding.userInput.addTextChangedListener(
```
alperozturk96, line 98: Could you please add parameter names to make it clearer which section is responsible for what? Also, instead of using `s ->`, could you use a more descriptive variable name for better readability?

app/src/main/java/com/nextcloud/utils/fileNameValidator/FileNameTextWatcher.kt
```diff
@@ -0,0 +1,62 @@
...
+ */
+@Suppress("LongParameterList")
+class FileNameTextWatcher(
+    private val previousFileName: String?,
+    private val context: Context,
+    private val getCapabilities: () -> OCCapability,
+    private val getExistingFileNames: () -> Set<String>?,
+    private val onError: Consumer<String>,
+    private val onWarning: Consumer<String>,
+    private val onOkay: Runnable
```
alperozturk96: Could you please use these parameter names?

```
private val capabilitiesProvider: () -> OCCapability,
private val existingFileNamesProvider: () -> Set<String>?,
private val onValidationError: Consumer<String>,
private val onValidationWarning: Consumer<String>,
private val onValidationSuccess: Runnable
```

app/src/main/java/com/nextcloud/utils/fileNameValidator/FileNameTextWatcher.kt
```diff
@@ -0,0 +1,62 @@
...
+) : TextWatcher {
+
+    private var isOkay: Boolean = true // Used to trigger the onOkay callback only once
```
alperozturk96: Could you please use this variable name?

`isNameCurrentlyValid` instead of `isOkay`

app/src/main/java/com/nextcloud/utils/fileNameValidator/FileNameTextWatcher.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    override fun afterTextChanged(s: Editable?) {
+        var newFileName = ""
+        if (s != null) {
+            newFileName = s.toString()
+        }
+
+        val errorMessage = checkFileName(newFileName, getCapabilities(), context, getExistingFileNames())
+
+        if (isFileHidden(newFileName)) {
```
alperozturk96: Can we use `when` block instead of `if, else` chain?

for example:

```kotlin
when {
  isHiddenFileName(currentFileName) -> {
    isNameCurrentlyValid = false
    onValidationWarning.accept(
      context.getString(R.string.hidden_file_name_warning)
    )
  }

  errorMessage != null -> {
    isNameCurrentlyValid = false
    onValidationError.accept(validationError)
  }

  hasFileExtensionChanged(originalFileName, currentFileName) -> {
    isNameCurrentlyValid = false
    onValidationWarning.accept(
      context.getString(R.string.warn_rename_extension)
    )
  }

  !isNameCurrentlyValid -> {
    isNameCurrentlyValid = true
    onValidationSuccess.run()
  }
}
```

app/src/main/java/com/nextcloud/utils/fileNameValidator/FileNameValidator.kt
PhilLab: @alperozturk96 good suggestion, I wasn't aware of this extension. But this extension returns a null file, if it doesn't exist, so it will not work for the newExtension. So I would need to write:

        val previousExtension = previousFileName.toFile()?.extension
        val newExtension = File(newFileName).extension

I think this reads weird and having similar code path makes the code more readable:

        val previousExtension = File(previousFileName).extension
        val newExtension = File(newFileName).extension

Let me know if you still want to have it changed, then I'll do it.

PhilLab:
@alperozturk96 thank you for your review, I addressed the remarks. Please take a look

app/src/main/java/com/nextcloud/utils/fileNameValidator/FileNameValidator.kt
alperozturk96: https://github.com/nextcloud/android/pull/16298/changes/e3eda3fa58365c6a73d10369d2ae1f42a49fda3c

PR #16305, opened 2026-01-16
fix(tabs-list-mode): layout switch

alperozturk96:

### Issue

Changing the layout mode in the **Favorites** tab currently affects the **Shared** tab as well.
The Shared tab does **not** support grid view and should always display files in list view.

**Actual behavior:**
Files in the Shared tab are displayed in grid view when the layout is toggled in the Favorites tab, even though no toggle button is available in the Shared tab.

**Expected behavior:**
Layout changes in the Favorites tab should **not** impact the Shared tab, which should always remain in list view.

---

### Changes

* Use the default layout value when the folder is `null` to determine the correct file list mode.
* Include the **Shared filter** when determining the appropriate layout mode to ensure Shared files always use list view.

---

### Out of scope

- The app has multiple roots (e.g., All Files, Favorites, Shared). Currently, the layout preference for Favorites and All Files roots is not stored separately. Therefore, remembering the last picked layout for these roots is **out of scope** for this PR.

- Introducing layout switch mode for shared tab.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
@@ -557,6 +557,8 @@ public abstract class DrawerActivity extends ToolbarActivity
             }
 
             closeDrawer();
+            setupHomeSearchToolbarWithSortAndListButtons();
+            updateActionBarTitleAndHomeButton(null);
         } else if (itemId == R.id.nav_favorites) {
             openFavoritesTab();
         } else if (itemId == R.id.nav_gallery) {
@@ -630,6 +632,8 @@ public abstract class DrawerActivity extends ToolbarActivity
                     fda.browseToRoot();
                 }
                 EventBus.getDefault().post(new ChangeMenuEvent());
+                setupHomeSearchToolbarWithSortAndListButtons();
+                updateActionBarTitleAndHomeButton(null);
             } else if (menuItemId == R.id.nav_favorites) {
                 openFavoritesTab();
             } else if (menuItemId == R.id.nav_assistant && !(this instanceof ComposeActivity)) {
@@ -706,8 +710,7 @@ public abstract class DrawerActivity extends ToolbarActivity
         }
     }
 
-    protected void openSharedTab() {
-        resetFileDepth();
+    private void openSharedTab() {
         resetOnlyPersonalAndOnDevice();
         SearchEvent searchEvent = new SearchEvent("", SearchRemoteOperation.SearchType.SHARED_FILTER);
         launchActivityForSearch(searchEvent, R.id.nav_shared);
--- a/app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
@@ -435,7 +435,7 @@ public class OCFileListFragment extends ExtendedListFragment implements
                 } else {
                     setGridAsPreferred();
                 }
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1573,17 +1577,9 @@ public void updateOCFile(@NonNull OCFile file) {
...
-        }
-
+        setLayoutViewMode();
         updateSortButton();
-        if (mSwitchGridViewButton != null) {
-            setGridSwitchButton();
-        }
+        setLayoutSwitchButton();
 
         setFabVisible(!mHideFab);
```
alperozturk96, line 1588: `setGridSwitchButton` internally checks nullability

app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -697,8 +701,7 @@ private void resetFileDepth() {
         }
     }
 
-    protected void openSharedTab() {
-        resetFileDepth();
```
alperozturk96, line 710: The call of `openSharedTab` already resets file depth.

app/src/main/java/com/owncloud/android/ui/activity/DrawerActivity.java
```diff
@@ -548,6 +548,8 @@ private void onNavigationItemClicked(final MenuItem menuItem) {
             }
 
             closeDrawer();
+            setupHomeSearchToolbarWithSortAndListButtons();
```
alperozturk96, line 560: Needed to display the correct action bar when the user comes back from the shared tab. In the future, we can deduplicate the navigation logic.

app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1857,11 +1873,7 @@ protected void handleSearchEvent(SearchEvent event) {
 
         new Handler(Looper.getMainLooper()).post(() -> {
             updateSortButton();
-            if (isGridViewPreferred(mFile) && !isGridEnabled()) {
-                switchToGridView();
-            } else if (!isGridViewPreferred(mFile) && isGridEnabled()) {
-                switchToListView();
-            }
```
alperozturk96, line 1864: `isGridEnabled` already used inside the switch functions.

app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -1619,14 +1615,34 @@ public void sortFiles(FileSortOrder sortOrder) {
...
-     * @param folder Folder to check or null for root folder
-     * @return 'true' is folder should be shown in grid mode, 'false' if list mode is preferred.
+     * @param folder The folder to check, or {@code null} to refer to the root folder.
+     * @return {@code true} if the folder should be displayed in grid mode, {@code false} if list mode is preferred.
      */
-    public boolean isGridViewPreferred(@Nullable OCFile folder) {
-        return FOLDER_LAYOUT_GRID.equals(preferences.getFolderLayout(folder));
+    private boolean isGridViewPreferred(@Nullable OCFile folder) {
+        if (searchEvent != null) {
+            return (searchEvent.toSearchType() != SHARED_FILTER) &&
```
alperozturk96, line 1625: The Shared tab can only display files in list mode for now, so folder preferences cannot be applied there since there is no toggle to change it.

PR #16326, opened 2026-01-20
feat(assistant): translate

alperozturk96:

### Implements second part of it: https://github.com/nextcloud/files-clients/issues/89#issuecomment-3772489560

Library PR: https://github.com/nextcloud/android-library/pull/1933


nimishavijay:
Good question. It seems very cumbersome to read long text in this half-screen-scrolling UI. I am assuming that most people will paste the text from somewhere else, in which case the specifics of the original text doesn't matter much. We can just keep the header bar sticky and scroll the rest like usual.

Other than that, 2 small points:
- [x] "Source language" seems a bit technical, we can change that to "Translate from"
- [x] It is not obvious from the UI that there is a dropdown for the languages, so we should add a small chevron-down icon next to the language. 
- [ ] There's also a floating button and I am not sure what it does (is it a button to force translation or create a "new translation task"?). would it be possible to just automatically translate whenever the source text changes (maybe throttled to 500ms or something to avoid sending too many requests)?

alperozturk96:
@kra-mo Here is the UI/UX.


https://github.com/user-attachments/assets/c9135f01-505e-4a3c-a46b-f9b799aea240

PR #16336, opened 2026-01-22
fix(auto-upload): sync conflict handling

alperozturk96:

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/utils/UploadErrorNotificationManager.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/utils/UploadErrorNotificationManager.kt
@@ -98,7 +98,6 @@ object UploadErrorNotificationManager {
         Log_OC.d(TAG, "🔔" + "notification created")
 
         withContext(Dispatchers.Main) {
-
             // if error code is file specific show new notification for each file
             if (result.code.isFileSpecificError()) {
                 notificationManager.showNotification(operation.ocUploadId.toInt(), notification)


app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -1262,27 +1262,6 @@ private RemoteOperationResult checkNameCollision(OCFile parentFile,
         return null;
     }
 
-    public void handleLocalBehaviour() {
```
alperozturk96, line 1265: `private handleLocalBehaviour` already used internally no need to call this outside from the `UploadFileOperation`.

app/src/main/java/com/nextcloud/client/jobs/utils/UploadErrorNotificationManager.kt
```diff
@@ -31,39 +31,74 @@ import java.io.File
...
-                }
-            )
-        } ?: return
+        val silentCodes = setOf(
+            ResultCode.DELAYED_FOR_WIFI,
+            ResultCode.DELAYED_FOR_CHARGING,
+            ResultCode.DELAYED_IN_POWER_SAVE_MODE,
+            ResultCode.LOCAL_FILE_NOT_FOUND,
+            ResultCode.LOCK_FAILED
+        )
```
ZetaTom, line 63: I feel like we should discuss these *silent codes*. Especially as users keep having issues with `ResultCode.LOCAL_FILE_NOT_FOUND`.

During my testing, removing a file during upload causes the notification 'to be silent'. In the *Uploads* tab, it shows up as *Pending, Unknown error*.

alperozturk96: Out of scope for this PR and we can discuss for next version.

PR #16340, opened 2026-01-22
handle oc upload construct

alperozturk96:

### DIFF
--- a/app/src/main/java/com/nextcloud/client/database/entity/UploadEntity.kt
+++ b/app/src/main/java/com/nextcloud/client/database/entity/UploadEntity.kt
@@ -66,8 +66,8 @@ fun UploadEntity.toOCUpload(capability: OCCapability? = null): OCUpload? {
     val upload = try {
         OCUpload(localPath, remotePath, accountName)
     } catch (_: IllegalArgumentException) {
-        null
-    } ?: return null
+        return null
+    }
 
     fileSize?.let { upload.fileSize = it }
     id?.let { upload.uploadId = it.toLong() }


app/src/main/java/com/nextcloud/client/database/entity/UploadEntity.kt
```diff
@@ -56,13 +57,17 @@ data class UploadEntity(
...
     var remotePath = remotePath
     if (capability != null && remotePath != null) {
         remotePath = AutoRename.rename(remotePath, capability)
     }
-    val upload = OCUpload(localPath, remotePath, accountName)
+    val upload = try {
+        OCUpload(localPath, remotePath, accountName)
+    } catch (_: IllegalArgumentException) {
+        null
+    } ?: return null
```
ZetaTom:
```suggestion
        return null
    }
```

app/src/main/java/com/nextcloud/client/jobs/autoUpload/AutoUploadWorker.kt
```diff
@@ -404,13 +405,18 @@ class AutoUploadWorker(
...
+        } catch (_: IllegalArgumentException) {
+            null
+        }
+
+        if (upload == null) {
+            Log_OC.e(TAG, "cannot construct oc upload")
+            return null
+        }
+
+        upload = upload.apply {
```
ZetaTom:
```suggestion
        val upload = try {
            uploadEntity?.toOCUpload(null) ?: OCUpload(localPath, remotePath, user.accountName)
        } catch (_: IllegalArgumentException) {
            Log_OC.e(TAG, "cannot construct oc upload")
            return null
        }

        upload.apply {
```

PR #16345, opened 2026-01-23
fix(e2ee): handle E2EE v2.1

alperozturk96:

### Issue

The introduction of `E2EE v2.1` exposed multiple code paths that relied on exact version matching `(== 2.0)`, causing logic meant for v2.x to be skipped. As a result, the app crashes when interacting with encrypted folders created using E2EE 2.1.

### How to reproduce

1. Install End-to-End Encrypted app [v2.0.0-rc.6.](https://github.com/nextcloud/end_to_end_encryption/releases/tag/v2.0.0-rc.6)
2. Create an encrypted folder
3. Refresh the newly created folder
4. App crashes; the folder remains locked and uploads are blocked

### Changes

- Treats E2EE v2.1 as compatible with v2.0 logic (v2 or above)
- Replaces exact version checks with consolidated helpers to avoid missing logic for future versions
- Deduplicates repeated E2EE version checks and centralizes them in a single helper
- Adds extension helpers (e.g. resolving E2EE version from capabilities) to reduce code duplication
- Adds unit tests for the new version helper to prevent regressions

### DIFF
--- a/app/src/main/java/com/nextcloud/utils/e2ee/E2EVersionHelper.kt
+++ b/app/src/main/java/com/nextcloud/utils/e2ee/E2EVersionHelper.kt
@@ -16,22 +16,45 @@ import com.owncloud.android.utils.EncryptionUtils
 
 object E2EVersionHelper {
 
-    fun isV2orAbove(capability: OCCapability): Boolean = isV2orAbove(capability.endToEndEncryptionApiVersion)
+    /**
+     * Returns true if the given E2EE version is v2 or newer.
+     */
+    fun isV2Plus(capability: OCCapability): Boolean = isV2Plus(capability.endToEndEncryptionApiVersion)
 
-    fun isV2orAbove(version: E2EVersion): Boolean = version == E2EVersion.V2_0 || version == E2EVersion.V2_1
+    /**
+     * Returns true if the given E2EE version is v2 or newer.
+     */
+    fun isV2Plus(version: E2EVersion): Boolean = version == E2EVersion.V2_0 || version == E2EVersion.V2_1
 
+    /**
+     * Returns true if the given E2EE version is v1.x.
+     */
     fun isV1(capability: OCCapability): Boolean = isV1(capability.endToEndEncryptionApiVersion)
 
+    /**
+     * Returns true if the given E2EE version is v1.x.
+     */
     fun isV1(version: E2EVersion): Boolean =
         version == E2EVersion.V1_0 || version == E2EVersion.V1_1 || version == E2EVersion.V1_2
 
-    fun getLatestE2EVersion(isV2: Boolean): E2EVersion = if (isV2) {
+    /**
+     * Returns the latest supported E2EE version.
+     *
+     * @param isV2 indicates whether the E2EE v2 series should be used
+     */
+    fun latestVersion(isV2: Boolean): E2EVersion = if (isV2) {
         E2EVersion.V2_1
     } else {
         E2EVersion.V1_2
     }
 
-    fun determineE2EFromVersionString(version: String?): E2EVersion = when (version?.trim()) {
+    /**
+     * Maps a raw version string to an [E2EVersion].
+     *
+     * @param version version string
+     * @return resolved [E2EVersion] or [E2EVersion.UNKNOWN] if unsupported
+     */
+    fun fromVersionString(version: String?): E2EVersion = when (version?.trim()) {
         "1.0" -> E2EVersion.V1_0
         "1.1" -> E2EVersion.V1_1
         "1.2" -> E2EVersion.V1_2
@@ -40,13 +63,19 @@ object E2EVersionHelper {
         else -> E2EVersion.UNKNOWN
     }
 
-    fun determineE2EVersion(metadata: String): E2EVersion = runCatching {
+    /**
+     * Determines the E2EE version by inspecting encrypted folder metadata.
+     *
+     * Supports both V1 and V2 metadata formats and falls back safely
+     * to [E2EVersion.UNKNOWN] if parsing fails.
+     */
+    fun fromMetadata(metadata: String): E2EVersion = runCatching {
         val v1 = EncryptionUtils.deserializeJSON<EncryptedFolderMetadataFileV1>(
             metadata,
             object : TypeToken<EncryptedFolderMetadataFileV1>() {}
         )
 
-        determineE2EFromVersionString(v1?.metadata?.version.toString()).also {
+        fromVersionString(v1?.metadata?.version.toString()).also {
             if (it == E2EVersion.UNKNOWN) {
                 throw IllegalStateException("Unknown V1 version")
             }
@@ -57,6 +86,6 @@ object E2EVersionHelper {
             object : TypeToken<EncryptedFolderMetadataFile>() {}
         )
 
-        determineE2EFromVersionString(v2.version)
+        fromVersionString(v2.version)
     }.getOrDefault(E2EVersion.UNKNOWN)
 }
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/nextcloud/utils/e2ee/E2EVersionHelper.kt
```diff
@@ -0,0 +1,62 @@
...
+    fun determineE2EFromVersionString(version: String?): E2EVersion = when (version?.trim()) {
+        "1.0" -> E2EVersion.V1_0
+        "1.1" -> E2EVersion.V1_1
+        "1.2" -> E2EVersion.V1_2
+        "2", "2.0" -> E2EVersion.V2_0
+        "2.1" -> E2EVersion.V2_1
+        else -> E2EVersion.UNKNOWN
+    }
+
+    fun determineE2EVersion(metadata: String): E2EVersion = runCatching {
```
alperozturk96: Replacement of `public static E2EVersion determinateVersion(String metadata) {`

app/src/main/java/com/owncloud/android/datamodel/e2e/v2/decrypted/DecryptedFolderMetadataFile.kt
```diff
@@ -15,5 +17,5 @@ data class DecryptedFolderMetadataFile(
     var users: MutableList<DecryptedUser> = mutableListOf(),
     @Transient
     val filedrop: MutableMap<String, DecryptedFile> = HashMap(),
-    val version: String = "2.0"
+    val version: String = E2EVersionHelper.getLatestE2EVersion(true).value
```
alperozturk96: Replaces hardcoded hidden strings with actual latest version so that in future we dont come across issues like this. @tobiasKaminsky

app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -548,10 +551,10 @@ private void synchronizeData(List<Object> folderAndFiles) {
         Map<String, OCFile> localFilesMap;
         E2EVersion e2EVersion;
         if (object instanceof DecryptedFolderMetadataFileV1 metadataFileV1) {
-            e2EVersion = E2EVersion.V1_2;
+            e2EVersion = E2EVersionHelper.INSTANCE.getLatestE2EVersion(false);
             localFilesMap = prefillLocalFilesMap(metadataFileV1, fileDataStorageManager.getFolderContent(mLocalFolder, false));
         } else {
-            e2EVersion = E2EVersion.V2_0;
+            e2EVersion = E2EVersionHelper.INSTANCE.getLatestE2EVersion(true);
```
alperozturk96: Again, for each version iteration, these hidden hardcoded values need to be upgraded. With `getLatestE2EVersion()` function we can just replace from there. In this case, `E2EVersion.V2_1` should be the new value.

app/src/main/java/com/owncloud/android/utils/EncryptionUtils.java
```diff
@@ -404,105 +405,64 @@ public static DecryptedFolderMetadataFileV1 decryptFolderMetaData(EncryptedFolde
...
-            }
-        } catch (Exception e) {
-            EncryptedFolderMetadataFile v2 = EncryptionUtils.deserializeJSON(
-                metadata,
-                new TypeToken<>() {
+        E2EVersion version = E2EVersionHelper.INSTANCE.determineE2EVersion(serializedEncryptedMetadata);
+
+        if (E2EVersionHelper.INSTANCE.isV2orAbove(version)) {
+            EncryptionUtilsV2 encryptionUtilsV2 = new EncryptionUtilsV2();
+            return encryptionUtilsV2.parseAnyMetadata(getMetadataOperationResult.getResultData(),
```
alperozturk96, line 421: `encryptionUtilsV2.parseAnyMetadata` needs to be used for v2.1 and v2.0.

app/src/main/java/com/owncloud/android/utils/EncryptionUtils.java
```diff
@@ -1352,7 +1317,7 @@ public static void uploadMetadata(ServerFileInterface parentFile,
         RemoteOperationResult<String> uploadMetadataOperationResult;
         if (metadataExists) {
             // update metadata
-            if (version == E2EVersion.V2_0) {
```
alperozturk96, line 1355: Exact version checks break the app and require manual updates for every new version. This logic has been replaced with `E2EVersionHelper`.

app/src/main/java/com/owncloud/android/utils/EncryptionUtilsV2.kt
```diff
@@ -606,7 +607,9 @@ class EncryptionUtilsV2 {
             object : TypeToken<EncryptedFolderMetadataFile>() {}
         )
 
-        val decryptedFolderMetadata = if (v2.version == "2.0" || v2.version == "2") {
```
alperozturk96, line 609: `v2.version == "2.0" || v2.version == "2"` Fragile and dangerous check and it will not work with 2.1 replaced with `E2EVersionHelper.fromVersionString`

app/src/main/java/com/nextcloud/utils/e2ee/E2EVersionHelper.kt
```diff
@@ -0,0 +1,91 @@
...
+        else -> E2EVersion.UNKNOWN
+    }
+
+    /**
+     * Determines the E2EE version by inspecting encrypted folder metadata.
+     *
+     * Supports both V1 and V2 metadata formats and falls back safely
+     * to [E2EVersion.UNKNOWN] if parsing fails.
+     */
+    fun fromMetadata(metadata: String): E2EVersion = runCatching {
```
alperozturk96, line 72: Replacement of `public static E2EVersion determinateVersion(String metadata) {`

Tries v1 if fails v2 if fails E2EVersion.UNKNOWN

app/src/main/java/com/owncloud/android/datamodel/e2e/v2/decrypted/DecryptedFolderMetadataFile.kt
```diff
@@ -15,5 +17,5 @@ data class DecryptedFolderMetadataFile(
     var users: MutableList<DecryptedUser> = mutableListOf(),
     @Transient
     val filedrop: MutableMap<String, DecryptedFile> = HashMap(),
-    val version: String = "2.0"
+    val version: String = E2EVersionHelper.latestVersion(true).value
```
alperozturk96, line 20: Replaces hardcoded hidden strings with actual latest version so that in future we dont come across issues like this. @tobiasKaminsky

app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -548,10 +551,10 @@ private void synchronizeData(List<Object> folderAndFiles) {
         Map<String, OCFile> localFilesMap;
         E2EVersion e2EVersion;
         if (object instanceof DecryptedFolderMetadataFileV1 metadataFileV1) {
-            e2EVersion = E2EVersion.V1_2;
+            e2EVersion = E2EVersionHelper.INSTANCE.latestVersion(false);
             localFilesMap = prefillLocalFilesMap(metadataFileV1, fileDataStorageManager.getFolderContent(mLocalFolder, false));
         } else {
-            e2EVersion = E2EVersion.V2_0;
+            e2EVersion = E2EVersionHelper.INSTANCE.latestVersion(true);
```
alperozturk96, line 557: Again, for each version iteration, these hidden hardcoded values need to be upgraded. With `latestVersion` function we can just replace from there. In this case, `E2EVersion.V2_1` should be the new value.

PR #16356, opened 2026-01-25
Ensure images always have a `contentDescription`

AndyScherzinger:
... or `null` in case it is decorative.

adding @rakekniven for approval, because we have a string-freeze for Hub 26 Winter and merging this would add some new Strings which is bad for translation completeness but good from an accessibility point of view. See below. 
(I didn't find the string to be in need of a descriptive translator comment, but if so, please let me know and I add them)

This info is important for screenreaders, so hard-coded strings should not be used but values in `string.xml` so there is no mix of languages when read to the users using screeenreader.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -64,9 +64,9 @@
     <string name="assistant_screen_delete_task_alert_dialog_title">Delete task</string>
     <string name="assistant_screen_delete_task_alert_dialog_description">Are you sure you want to delete this task?</string>
     <string name="assistant_screen_task_more_actions_bottom_sheet_delete_action">Delete Task</string>
-    <string name="assistant_screen_task_create_success_message">Task successfully created</string>
+    <string name="assistant_screen_task_create_success_message">Task created</string>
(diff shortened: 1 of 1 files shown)


rakekniven:
Hello,

From my side, that's OK.
I have no knowledge of the release planning from the Android app.

AndyScherzinger:
Thanks for the swift feedback @rakekniven 🙏 

Plan would be to release or around the Hub release, hence around 18th February 🚀

app/src/main/res/values/strings.xml
```diff
@@ -68,8 +68,17 @@
     <string name="assistant_screen_task_create_fail_message">An error occurred while creating the task</string>
     <string name="assistant_screen_task_delete_success_message">Task successfully deleted</string>
```
rakekniven: Please avoid using words like "successfully" or "really". Keep messages short.

AndyScherzinger: Fixed via https://github.com/nextcloud/android/pull/16356/changes/fa4d85f86c0926d6eaad49cfa5a50c770d1ae139 👍 
Extra commit so I can just backport the string change 💡

PR #16393, opened 2026-01-28
fix(auto-upload): calls

alperozturk96:

### Issue

On Android, long-running foreground jobs cannot be started while the app is in the background. Attempting to do so results in the following exception:

```
E  ❌ Exception in ContentObserverWork: startForegroundService() not allowed due to mAllowStartForeground false: service com.nextcloud.client/androidx.work.impl.foreground.SystemForegroundService
android.app.ForegroundServiceStartNotAllowedException: startForegroundService() not allowed due to mAllowStartForeground false: service com.nextcloud.client/androidx.work.impl.foreground.SystemForegroundService
    at android.app.ForegroundServiceStartNotAllowedException$1.createFromParcel(ForegroundServiceStartNotAllowedException.java:54)
    at android.app.ForegroundServiceStartNotAllowedException$1.createFromParcel(ForegroundServiceStartNotAllowedException.java:50)
```

`ContentObserverWork` triggers `AutoUploadWorker` to upload files. When this happens while the app is in the background, the foreground service cannot be started, causing the auto-upload process to fail.

Additionally, when the app is removed from recent apps, `ContentObserverWork` may not be triggered at all. In that case, changed content URIs are never passed to `AutoUploadWorker`, which can cause auto-upload to not work.

### Changes

Remove duplicated and repeated worker-trigger logic, keeping only the required and correct trigger points.

Make `AutoUploadWorker` a long-running foreground worker, as required for upload operations.

Trigger `AutoUploadWorker` again when the app returns to the foreground. This ensures pending uploads resume in cases where the user previously killed the app and the scheduled execution of `ContentObserverWork` did not occur.

### DIFF
--- a/app/src/main/java/com/owncloud/android/MainApp.java
+++ b/app/src/main/java/com/owncloud/android/MainApp.java
@@ -17,7 +17,6 @@ package com.owncloud.android;
 
 import android.annotation.SuppressLint;
 import android.app.Activity;
-import android.app.ActivityManager;
 import android.app.Application;
 import android.app.NotificationChannel;
 import android.app.NotificationManager;
@@ -31,7 +30,6 @@ import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.content.res.Resources;
 import android.net.ConnectivityManager;
-import android.os.Build;
 import android.os.Bundle;
 import android.os.Environment;
 import android.os.StrictMode;
@@ -389,10 +387,14 @@ public class MainApp extends Application implements HasAndroidInjector, NetworkC
     private final LifecycleEventObserver lifecycleEventObserver = ((lifecycleOwner, event) -> {
         if (event == Lifecycle.Event.ON_START) {
             Log_OC.d(TAG, "APP IN FOREGROUND");
-            FilesSyncHelper.startAutoUploadForEnabledSyncedFolders(syncedFolderProvider,
-                                                                   backgroundJobManager,
-                                                                   new String[]{},
-                                                                   true);
+
+            if (preferences.startAutoUploadOnStart()) {
+                FilesSyncHelper.startAutoUploadForEnabledSyncedFolders(syncedFolderProvider,
+                                                                       backgroundJobManager,
+                                                                       new String[]{},
+                                                                       true);
+                preferences.setLastAutoUploadOnStartTime(System.currentTimeMillis());
+            }
         } else if (event == Lifecycle.Event.ON_STOP) {
             passCodeManager.setCanAskPin(true);
             Log_OC.d(TAG, "APP IN BACKGROUND");
--- a/app/src/main/java/com/nextcloud/client/preferences/AppPreferences.java
+++ b/app/src/main/java/com/nextcloud/client/preferences/AppPreferences.java
@@ -396,4 +396,7 @@ public interface AppPreferences {
 
     String getLastDisplayedAccountName();
     void setLastDisplayedAccountName(String lastDisplayedAccountName);
+
+    boolean startAutoUploadOnStart();
+    void setLastAutoUploadOnStartTime(long timeInMillisecond);
 }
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/nextcloud/client/jobs/autoUpload/AutoUploadHelper.kt
```diff
@@ -31,16 +31,6 @@ class AutoUploadHelper {
     }
 
     fun insertEntries(folder: SyncedFolder, repository: FileSystemRepository) {
-        val enabledTimestampMs = folder.enabledTimestampMs
```
alperozturk96, line 34: `fun SyncedFolder.shouldSkipFile` already checking and `isExisting` is not representing existence of the sync folder it means "Also upload existing files". Function name will be change in different PR.

app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
```diff
@@ -269,13 +268,15 @@ internal class BackgroundJobManagerImpl(
...
 
+    @Suppress("MagicNumber")
     override fun scheduleContentObserverJob() {
         val constrains = Constraints.Builder()
             .addContentUriTrigger(MediaStore.Images.Media.INTERNAL_CONTENT_URI, true)
             .addContentUriTrigger(MediaStore.Images.Media.EXTERNAL_CONTENT_URI, true)
             .addContentUriTrigger(MediaStore.Video.Media.INTERNAL_CONTENT_URI, true)
             .addContentUriTrigger(MediaStore.Video.Media.EXTERNAL_CONTENT_URI, true)
-            .setTriggerContentMaxDelay(MAX_CONTENT_TRIGGER_DELAY_MS, TimeUnit.MILLISECONDS)
+            .setTriggerContentUpdateDelay(Duration.ofSeconds(5))
```
alperozturk96, line 278: Creating 5, 10 as a const value to fix `MagicNumber` is not good for this use case. `Duration.ofSeconds(5)` clearly indicates what it is.

app/src/main/java/com/nextcloud/client/jobs/ContentObserverWork.kt
```diff
@@ -69,46 +56,19 @@ class ContentObserverWork(
...
         } catch (e: Exception) {
             Log_OC.e(TAG, "❌ Exception in ContentObserverWork: ${e.message}", e)
             Result.retry()
+        } finally {
+            Log_OC.d(TAG, "🔄" + "re-scheduling job")
+            backgroundJobManager.scheduleContentObserverJob()
         }
     }
 
-    private suspend fun updateForegroundInfo(notification: Notification) {
```
alperozturk96, line 84: `ContentObserverWork` should be background job.

app/src/main/java/com/owncloud/android/MainApp.java
```diff
@@ -386,6 +387,14 @@ public void disableDocumentsStorageProvider() {
     private final LifecycleEventObserver lifecycleEventObserver = ((lifecycleOwner, event) -> {
         if (event == Lifecycle.Event.ON_START) {
             Log_OC.d(TAG, "APP IN FOREGROUND");
+
+            if (preferences.startAutoUploadOnStart()) {
+                FilesSyncHelper.startAutoUploadForEnabledSyncedFolders(syncedFolderProvider,
```
alperozturk96, line 392: Can triggering the auto-upload worker every 10 minutes on app start cause issues? Ideally, failed uploads/downloads shouldn’t happen often, but there are cases where some files might get stuck. With this approach, those stuck files would be retried and that can overwhelm the user.

On the other hand, relying on a content observer isn’t fully reliable, there’s no guarantee it will always trigger the auto-upload worker correctly. For example, the app could be killed, or the worker could be delayed or restricted by the OS or by custom Android skins.

@tobiasKaminsky @ZetaTom lets discuss this.

alperozturk96: @tobiasKaminsky as discussed lets merge only to the master for now.

PR #16418, opened 2026-01-30
fix(share): icons

alperozturk96:

### Fixes

Share internal link alignment
Dropdown icon visibility (only visible if permission name exists)
Converts image views to material icon button since they are button not icon

### DIFF
--- a/app/src/main/res/layout/file_details_share_link_share_item.xml
+++ b/app/src/main/res/layout/file_details_share_link_share_item.xml
@@ -89,7 +89,7 @@
         android:paddingEnd="@dimen/standard_padding"
         android:paddingBottom="@dimen/standard_quarter_margin"
         android:scaleType="fitCenter"
-        app:iconTint="@color/text_color"
+        app:iconTint="@color/secondary_text_color"
         app:icon="@drawable/ic_content_copy"
         app:iconGravity="textStart" />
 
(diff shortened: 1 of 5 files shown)


AndyScherzinger:
@alperozturk96 What is the webUI displaying? Is it also hiding the element or are we missing something on the client-side?

AndyScherzinger:
@alperozturk96 the icon colors changed, but didn't yet check what M3 would define for icon-only buttons

alperozturk96:
### Before

Screenshot 2026-02-04 at 15 32 05
Screenshot 2026-02-04 at 15 32 18
Screenshot 2026-02-04 at 15 32 28
Screenshot 2026-02-04 at 15 32 32



### After

Screenshot 2026-02-04 at 15 27 46
Screenshot 2026-02-04 at 15 27 49
Screenshot 2026-02-04 at 15 28 20
Screenshot 2026-02-04 at 15 28 24

alperozturk96:
> @alperozturk96 the icon colors changed, but didn't yet check what M3 would define for icon-only buttons


Maybe this `app:iconTint="?attr/colorOnSurfaceVariant"` ?


Screenshot 2026-02-04 at 15 39 28
Screenshot 2026-02-04 at 15 39 32

AndyScherzinger:
If you want a declarative way, current you'd need to set up the Grey color we defined, afaik #666666

The material color is dynamic, so you'd need to call a theming util function for that.

Switching to colorOnSurfaceVariant would force you to change that everywhere in the UI so I wouldn't not do it in this PR but just make the changed one Grey again.

alperozturk96:
> If you want a declarative way, current you'd need to set up the Grey color we defined, afaik #666666
> 
> The material color is dynamic, so you'd need to call a theming util function for that.
> 
> Switching to colorOnSurfaceVariant would force you to change that everywhere in the UI so I wouldn't not do it in this PR but just make the changed one Grey again.

I kept the icon colors as it is (@color/secondary_text_color - #666666). Then we can save it for other PR.

I will check your first message.

Issue #16421, opened 2026-01-31
App build fails in encrypted Linux home dir since upgrade to AGP 9


PhilLab:
I can't build the app anymore. When building, the task `:app:dexBuilderGenericDebug`
fails with dozens of "File name too long" errors like:

> java.nio.file.FileSystemException: /home/myuser/devrepo/nextcloud-android/app/build/intermediates/global_synthetics_project/genericDebug/dexBuilderGenericDebug/out/com/owncloud/android/ui/adapter/ReceiveExternalFilesAdapter$ReceiveExternalViewHolder$$InternalSyntheticLambda$2$6636dd2b6ebd75843672790eb9fec557e7c7714a7c99b5385b15a5972ca7d9f8$0.globals: File name too long

I traced this back to 6ac2f2235983b4c98a21fa946fcef3b6440f6310:
```
-androidPluginVersion = '8.13.2'
+androidPluginVersion = "9.0.0"
```
When I set the version back to 8.13.2, I don't have this problem anymore and it builds just fine.

The problem seems to be that the Android Gradle Plugin generates synthetic lambda class names that can exceed Linux's file name length limit. Typically, a file name can be 255 characters (and full path 4096), but this is different if you layer eCryptFS with filename encryption on top of these file systems (**as installed in Ubuntu with encrypted home dir option**).

Then, the max filename is only 143 characters ([source](https://unix.stackexchange.com/questions/32795/what-is-the-maximum-allowed-filename-and-folder-size-with-ecryptfs/32834#32834)), which many of the lambda class names exceed (the one from the example above is 156 characters long)

I did verify this, and the console command confirmed this:
```bash
$ getconf NAME_MAX /home/myuser/
143
```

This seems like a bug which can only be fixed upstream, by Google. So I am not sure how to handle this - can we downgrade to 8.13.2 for the meantime?

**Update** A bug ticket has been created for Google: https://issuetracker.google.com/issues/481432927

PR #16306, opened 2026-01-16
chore(agp): Android Gradle plugin 9.0.0 

tobiasKaminsky:
- [ ] Tests written, or not not needed

PR #16422, opened 2026-01-31
fix(gradle): Fixed broken handling of test server and credentials

PhilLab:
This was a refactoring error from a17e1385f2049cb0f3597dc620894868d8fdba39

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -94,9 +94,9 @@ android {
 
     defaultConfig {
         testInstrumentationRunnerArguments += mapOf(
-            "TEST_SERVER_URL" to "ncTestServerBaseUrl.toString()",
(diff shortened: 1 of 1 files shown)


PhilLab:
@tobiasKaminsky 
1. For what its worth, I also had other build troubles caused by the Gradle update, but I didn't include them in the PR here, because their solution is less obvious: #16421
2. I am not exactly sure what it means that the PR I am hereby amending: https://github.com/nextcloud/android/pull/16306 , was added to the **3.36.0 milestone**. Maybe the fix here should be in the milestone as well, if this is required to be part of the release code.

PhilLab:
@tobiasKaminsky this is needed to run automated tests locally, can you take a look?

PR #16425, opened 2026-02-01
Media Sorting Improvement

Leo-Berman:

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/GalleryAdapter.kt
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/GalleryAdapter.kt
@@ -23,6 +23,7 @@ import com.afollestad.sectionedrecyclerview.SectionedRecyclerViewAdapter
 import com.afollestad.sectionedrecyclerview.SectionedViewHolder
 import com.nextcloud.client.account.User
 import com.nextcloud.client.preferences.AppPreferences
+import com.nextcloud.utils.date.DateFormatter
 import com.owncloud.android.databinding.GalleryHeaderBinding
 import com.owncloud.android.databinding.GalleryRowBinding
 import com.owncloud.android.datamodel.FileDataStorageManager
@@ -42,7 +43,6 @@ import com.owncloud.android.utils.theme.ViewThemeUtils
 import me.zhanghai.android.fastscroll.PopupTextProvider
 import java.util.Calendar
 import java.util.Date
-import java.util.regex.Pattern
 
 @Suppress("LongParameterList", "TooManyFunctions")
 class GalleryAdapter(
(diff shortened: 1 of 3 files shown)


Leo-Berman:
maybe this can't be done in the frontend because if we load by timestamp in the database part then the sections will be loaded weirdly anyways?

Leo-Berman:
I made a change so it loads everything all at once which saves the categories but I don't really understand exactly the load it places

christianlupus:
Cross-linking discussion from forum here:  https://help.nextcloud.com/t/first-pull-request-help-regarding-testing/239607

@tobiasKaminsky, maybe you can have a quick look at this if something like it would be a good idea to implement. Then, maybe you could either give some good advice on what is needed to get this merged or what would be needed to have good chances? Thanks a lot!

Leo-Berman:
Hi @alperozturk96 or @tobiasKaminsky I did some more work to make sure the code analysis was passing if that was a blocker. Let me know if there is anything else you'd like for me to do? Would this being a setting make it more likely to make it into master? I'm not super pressed about it, but it's more convenient for me so I don't have to keep building myself an SDK, so if there is a relatively easy way for this to happen that'd be great, and I'm willing to do a bit more work for it!

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
```diff
@@ -59,7 +59,8 @@ public class GalleryFragment extends OCFileListFragment implements GalleryFragme
     private boolean photoSearchQueryRunning = false;
     private AsyncTask<Void, Void, GallerySearchTask.Result> photoSearchTask;
     private long endDate;
-    private int limit = 150;
+    // Use 0 for unlimited - fetch all metadata at once; thumbnails load lazily
+    private int limit = 0;
```
alperozturk96, line 63: This may introduce unforeseen side effects on slow and very large instances, as forcing `SearchRemoteOperation` to fetch everything could cause performance issues.

@tobiasKaminsky Please share your thoughts as well.

app/src/main/java/com/owncloud/android/ui/adapter/GalleryAdapter.kt
```diff
@@ -59,6 +60,52 @@ class GalleryAdapter(
...
+
+        /**
+         * Extract folder date from path (YYYY, YYYY/MM, or YYYY/MM/DD).
+         * Uses LocalDate for calendar-aware validation (leap years, days per month).
+         * Invalid month/day values fall back to defaults. Future dates are rejected.
+         * @return timestamp or null if no folder date found or date is in the future
+         */
+        @VisibleForTesting
+        @Suppress("TooGenericExceptionCaught")
+        fun extractFolderDate(path: String?): Long? {
```
alperozturk96: Please move this folder date extraction logic to the `DateFormatter` object and add logs for exceptions cases.

alperozturk96 (changes requested):
Hello

Thank you for the PR.

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.java
Leo-Berman: I have tried with my local bench, but I think it would have to be so many images before this meta data loading actually started to be a bottleneck. Of course it's just a hunch, I'm not sure what metrics to look at

app/src/main/java/com/owncloud/android/ui/adapter/GalleryAdapter.kt
Leo-Berman: 279cff046cf00bc67b8024b49a4413663eddfa25

app/src/main/java/com/owncloud/android/ui/adapter/GalleryAdapter.kt
```diff
@@ -370,12 +371,43 @@ class GalleryAdapter(
         }
     }
 
+    /**
+     * Get the grouping date for a file: use folder date from path if present,
+     * otherwise fall back to modification timestamp month.
+     */
+    private fun getGroupingDate(file: OCFile): Long =
+        firstOfMonth(DateFormatter.extractFolderDate(file.remotePath) ?: file.modificationTimestamp)
```
alperozturk96, line 379: `extractFolderDate` seems to rely on the file path matching a very specific pattern, so it may not work correctly if the structure differs. I’m also wondering if this is the right place to handle sorting, since we’re already sorting by month and then again by date inside `transformToRows`.

Could you please share a bit more context about the issue this change is addressing? It would be really helpful if you could include the reproduction steps or an example scenario so we can better understand this change.

Leo-Berman: The specific file pathing thing this to math the autoupload subfolders structure. Since it only uses time stamp, if they are off then they don't match
 This would just be changing priority to use the subfolders structure and then if that fails fallback to timestamp. I think I used upload existing files or something and so it showed all my pictures under one month. Then after I added the sorting, there was still some errors because it would still load in batch by the timestamp, so I removed the limit because in my mind the meta data would rarely be a large amount even if we loaded every image and the thumbnails seem to already lazy load.

alperozturk96:
Hello

Thank you for your contribution and the effort you’ve put into this PR.

After reviewing the changes, we would like to highlight that `GalleryAdapter.kt` should not contain logic related to the auto-upload subfolder structure or different pattern. Introducing this responsibility into the adapter increases its complexity, makes it harder to maintain, and goes beyond its intended scope.

Additionally, setting the limit to `0` is not a viable solution, as it can lead to issues in larger instances. We kindly ask that edge cases like these be carefully considered when proposing changes.

Given the sensitivity of this area and the existing implementations already in place, we are unfortunately unable to merge this PR in its current form.

For future contributions, we recommend keeping changes as minimal and focused as possible, along with providing thorough test coverage. This helps us review changes more effectively and ensures stability in critical parts of the application.

We truly appreciate your work. Thank you.

Issue #12624, opened 2024-02-29
Build scanner binary component (for F-Droid/etc scanning support)


joshtrichards:
the scanner binary component is OSS yet it means that nextcloud would need to build and include a build process for this binary shipped by the scanner lib dependency ourselves.

So anybody could build and contribute this but it is currently not a priority for Nextcloud. So we are happy to see folks contributing this in order to also have document scanning capabilities in the f-droid version. 👍

_Originally posted by @AndyScherzinger in https://github.com/nextcloud/android/issues/10195#issuecomment-1881836396_

stefan-niedermann:
Maybe using the [Document Scanner API](https://android-developers.googleblog.com/2024/02/ml-kit-document-scanner-api.html?m=1) might be a good solution? It's used by Google Drive but depends on proprietary Play Services. ([More information](https://developers.google.com/ml-kit/vision/doc-scanner?hl=en))

PhilLab:
@tobiasKaminsky I'd like to invest some time into building a tiny OpenCV from source to have the document scanning available on F-Droid.
Is this something which would get reviewed, if I had a PR for it? Just wanting to confirm that I don't waste my time experimenting with this?

And what about a more fundamental improvement of the scanner, updating the computer vision to newer standards - is that also something which would get reviewed?

AndyScherzinger:
@PhilLab yes, definitely. The answer to both questions. The potential effort to build the native components of the use scanner library have always been rated too high. So yeah, a contribution would be highly welcomed and even step one would be a huge improvement since it would also the scanner feature to ship with f-droid as well. 👍

CC @tobiasKaminsky

PhilLab:
@AndyScherzinger @tobiasKaminsky Great to hear. The root issue is that to keep Nextcloud's apk small, a custom [Tiny OpenCV](https://github.com/Hazzatur/Tiny-OpenCV/blob/master/how_to.md) is used, which is not built from source when included from the [Document Scanning SDK](https://github.com/Hazzatur/Document-Scanning-Android-SDK) which Nextcloud uses.
Both of which are not really maintained and use old Computer Vision approaches.

To make it easier to upgrade to a more modern OSS scanning library, would it be an alternative approach for me to **split the generated .apks per ABI (x86, x86_64, arm64-v8a, armeabi-v7a), which will make it much smaller, and accept a larger scanner library instead?**. Of course, still _trying_ to keep the library small, but with a bit more wiggle room, not sacrificing everything to this one requirement.

[...]

PR #16427, opened 2026-02-01
feat(doc-scan): Query for FairScan, if built-in scanning not available

PhilLab:

Built-in scanning is not available for all build variants - e.g. not for "generic", which is the flavor for the F-Droid release. It doesn't allow the non-reproducible TinyOpenCV build.

If this is the case, we are checking whether the open source scanning app FairScan is available (https://github.com/pynicolas/FairScan) and open that one for scanning.

This approach was selected based on the discussion here: https://github.com/nextcloud/android/issues/12624#issuecomment-3727593158

### Result
[Screen_recording_20260201_143916.webm](https://github.com/user-attachments/assets/83c59815-acb1-4c91-b48a-cdfaa7c4bc71)

1. I thought about giving this menu entry a dedicated title, e.g. "Scan document from FairScan", but I decided against it - it's just more code to maintain and at the same time doesn't provide value to the user. Also it is not scalable: what if we add fallbacks to other scanner apps as well? So let's stick to the standard menu entry text.
2. I also had in mind to make this menu entry an "Install FairScan" button, in case there is no appscan and FairScan also is not installed. But that felt too intrusive.

### Behind the scenes
1. To simplify the detection of the appscan project, a reflection-based approach was implemented. See the messages of the first two commits for why I am proposing this. If you don't agree with this approach, I can remove the commits from the PR, they don't impact the feature itself
2. FairScan delivers a short-lived URI which we then use to upload the file. Originally, the filename is an "ugly" unix timestamp - you can also see this in above recording, from a prior attempt: `1769952395766.pdf`.
    - To make this a little prettier, the filename is transformed to our standard timestamped name, just like we do for images and videos captured via the camera intent.
    - Technically, this is done with the same approach of a `fileDisplayNameTransformer` as in my other current PR #16298.
3. Notably, FairScan's Intent feature is still experimental (see their [README](https://github.com/pynicolas/FairScan?tab=readme-ov-file#experimental-scan-to-pdf-via-intent)) but I'd still propose to include it, because:
    - For Non-F-Droid users, all stays the same and uses the built-in scanner
    - For F-Droid users, the menu entry will only appear if FairScan is installed, so the area of impact is quite small.
    - The intent integration is very basic so it is not very likely that FairScan's next iteration will change it in a way which is incompatible with this.
    - If it breaks by changing the intent URI - the button will simply not show anymore so we are not worse off than the status-quo.
    - If it breaks otherwise, then we are still not worse off than having no scanner in the F-Droid variant at all.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
+++ b/app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
@@ -45,6 +45,7 @@ import android.view.inputmethod.InputMethodManager
 import androidx.activity.OnBackPressedCallback
 import androidx.annotation.VisibleForTesting
 import androidx.appcompat.widget.SearchView
+import androidx.core.util.Function
 import androidx.core.view.MenuItemCompat
 import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
@@ -157,10 +158,12 @@ import com.owncloud.android.utils.PermissionUtil.requestNotificationPermission
 import com.owncloud.android.utils.PermissionUtil.requestStoragePermissionIfNeeded
 import com.owncloud.android.utils.PushUtils
 import com.owncloud.android.utils.StringUtils
+import com.owncloud.android.utils.UriUtils
 import com.owncloud.android.utils.theme.CapabilityUtils
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import org.apache.commons.io.FilenameUtils
 import org.greenrobot.eventbus.EventBus
 import org.greenrobot.eventbus.Subscribe
 import org.greenrobot.eventbus.ThreadMode
@@ -971,10 +974,13 @@ class FileDisplayActivity :
      */
     override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
         if (data != null &&
-            requestCode == REQUEST_CODE__SELECT_CONTENT_FROM_APPS &&
+            (
+                requestCode == REQUEST_CODE__SELECT_CONTENT_FROM_APPS ||
+                    requestCode == REQUEST_CODE__SELECT_CONTENT_FROM_APPS_AUTO_RENAME
+                ) &&
             (resultCode == RESULT_OK || resultCode == UploadFilesActivity.RESULT_OK_AND_MOVE)
         ) {
-            requestUploadOfContentFromApps(data, resultCode)
+            requestUploadOfContentFromApps(requestCode, resultCode, data)
         } else if (data != null &&
             requestCode == REQUEST_CODE__SELECT_FILES_FROM_FILE_SYSTEM &&
             (
@@ -1108,7 +1114,7 @@ class FileDisplayActivity :
         }
     }
 
-    private fun requestUploadOfContentFromApps(contentIntent: Intent, resultCode: Int) {
+    private fun requestUploadOfContentFromApps(requestCode: Int, resultCode: Int, contentIntent: Intent) {
         val streamsToUpload = ArrayList<Parcelable?>()
 
         if (contentIntent.clipData != null && (contentIntent.clipData?.itemCount ?: 0) > 0) {
@@ -1130,6 +1136,17 @@ class FileDisplayActivity :
 
         val currentDir = getCurrentDir()
         val remotePath = if (currentDir != null) currentDir.remotePath else OCFile.ROOT_PATH
+        var fileDisplayNameTransformer: Function<Uri, String?>? = null
+        if (requestCode == REQUEST_CODE__SELECT_CONTENT_FROM_APPS_AUTO_RENAME) {
+            fileDisplayNameTransformer = { uri: Uri ->
+                val displayName = UriUtils.getDisplayNameForUri(uri, applicationContext)
+                if (displayName != null && displayName.isNotEmpty()) {
+                    FileOperationsHelper.getTimestampedFileName("." + FilenameUtils.getExtension(displayName))
+                } else {
+                    null
+                }
+            }
+        }
 
         val uploader = UriUploader(
             this,
@@ -1140,7 +1157,8 @@ class FileDisplayActivity :
             ),
             behaviour,
             false, // Not show waiting dialog while file is being copied from private storage
-            null // Not needed copy temp task listener
+            null, // Not needed copy temp task listener
+            fileDisplayNameTransformer
         )
 
         uploader.uploadUris()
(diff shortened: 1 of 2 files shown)


alperozturk96:
@tobiasKaminsky What do you think, could you share your opinion?

PhilLab:
@tobiasKaminsky @alperozturk96 can you let me know your thoughts about the implementation?
As I said, I can easily drop the reflection-based appscan check, if you think this is entirely the wrong direction. It has its pros and cons, as explained in the respective commit message.

tobiasKaminsky:
I checked again with Fdroid, if we could use tiny cv:
https://gitlab.com/fdroid/fdroiddata/-/issues/3822

But since this will not work / is too complicated, I am fine with this approach ✔

app/src/androidTestGeneric/java/com/nextcloud/client/di/VariantModuleTest.kt
```diff
@@ -0,0 +1,57 @@
...
+        // Verify that calling getScanContract on stub throws UnsupportedOperationException
+        try {
+            feature.getScanContract()
+            throw AssertionError("Expected UnsupportedOperationException")
+        } catch (e: UnsupportedOperationException) {
+            assertTrue(e.message?.contains("not available") == true)
+        }
+    }
+
+    /**
```
alperozturk96: Please remove unnecessary comment since interface and what it does already clear.

app/src/androidTestGeneric/java/com/nextcloud/client/di/VariantModuleTest.kt
```diff
@@ -0,0 +1,57 @@
...
+     * In this variant, app scan should not be available
+     */
+    @Test
+    fun testAppScanOptionalFeatureAvailability() {
+        val feature = component.appScanOptionalFeature()
+
+        assertFalse(feature.isAvailable)
+        assertEquals(AppScanOptionalFeature.Stub, feature)
+
+        // Verify that calling getScanContract on stub throws UnsupportedOperationException
```
alperozturk96: Please remove comment logic already explains itself.

app/src/androidTestGplay/java/com/nextcloud/client/di/VariantModuleTest.kt
```diff
@@ -0,0 +1,54 @@
...
+
+        assertTrue(feature.isAvailable)
+        assertNotEquals(AppScanOptionalFeature.Stub, feature)
+
+        // Verify that calling getScanContract returns without raising an exception
+        assertNotNull(feature.getScanContract())
+    }
+
+    /**
+     * Dagger component for testing VariantModule in isolation
```
alperozturk96: Please remove unnecessary comment since interface and what it does already clear.

app/src/androidTestGeneric/java/com/nextcloud/client/di/VariantModuleTest.kt
```diff
@@ -0,0 +1,57 @@
...
+    @Before
+    fun setup() {
+        component = DaggerVariantModuleTest_TestVariantComponent.create()
+    }
+
+    /**
+     * In this variant, app scan should not be available
+     */
+    @Test
+    fun testAppScanOptionalFeatureAvailability() {
```
alperozturk96: Please rename function name to `testAppScanWhenNotAvailableShouldReturnError` and remove function comment.

app/src/androidTestGplay/java/com/nextcloud/client/di/VariantModuleTest.kt
```diff
@@ -0,0 +1,54 @@
...
+    @Before
+    fun setup() {
+        component = DaggerVariantModuleTest_TestVariantComponent.create()
+    }
+
+    /**
+     * In this variant, app scan should be available
+     */
+    @Test
+    fun testAppScanOptionalFeatureAvailability() {
```
alperozturk96: Please rename function name to `testAppScanWhenAvailableShouldReturnContract` and remove function comment.

app/src/androidTestGplay/java/com/nextcloud/client/di/VariantModuleTest.kt
```diff
@@ -0,0 +1,54 @@
...
+        val feature = component.appScanOptionalFeature()
+
+        assertTrue(feature.isAvailable)
+        assertNotEquals(AppScanOptionalFeature.Stub, feature)
+
+        // Verify that calling getScanContract returns without raising an exception
```
alperozturk96: Please remove the comment. Since function name will be changed and what it is doing clear.

app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1130,6 +1136,17 @@ class FileDisplayActivity :
 
         val currentDir = getCurrentDir()
         val remotePath = if (currentDir != null) currentDir.remotePath else OCFile.ROOT_PATH
+        var fileDisplayNameTransformer: androidx.core.util.Function<Uri, String?>? = null
```
alperozturk96: Please remove or simplify `androidx.core.util.Function<Uri, String?>?` variable type if possible.

app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
```diff
@@ -226,6 +227,8 @@ public class OCFileListFragment extends ExtendedListFragment implements
     private FloatingActionButton mFabMain;
     public static boolean isMultipleFileSelectedForCopyOrMove = false;
 
+    private static final Intent mFairScanIntent = new Intent("org.fairscan.app.action.SCAN_TO_PDF");
```
alperozturk96: `mFairScanIntent` please replace this variable name to `scanIntent`.

I suggest to avoid using `m` prefix for new variable names to match with rest of the code. Also no need to indicate specific package name in the variable since it's is just a scan intent.

app/src/main/AndroidManifest.xml
```diff
@@ -107,6 +107,11 @@
         <intent>
             <action android:name="android.intent.action.GET_CONTENT" />
         </intent>
+        <intent>
+            <!-- Since Android 11 (API level 30), this is required to detect the availability of FairScan to use
+             for external document scanning. See https://developer.android.com/training/package-visibility -->
+            <action android:name="org.fairscan.app.action.SCAN_TO_PDF" />
```
alperozturk96: Please remove the comment.  

`<action android:name="org.fairscan.app.action.SCAN_TO_PDF" />` this already explains to developer what it is doing.

alperozturk96 (changes requested):

app/src/main/AndroidManifest.xml
PhilLab: @alperozturk96 I moved it up and adjusted a little to fit the other declarations

app/src/main/java/com/owncloud/android/ui/fragment/OCFileListFragment.java
PhilLab: @alperozturk96 renamed, though made it explicit in the new name that it is an external app. This way you cannot confuse it with any internal intent.

PhilLab:
@alperozturk96 thanks for the helpful review, I force-pushed an update just now.

alperozturk96 (approved):

PR #16435, opened 2026-02-03
fix(manage-space-activity): app in foreground

alperozturk96:

### Issue

App is crashing when user trying to open the app after deleting data.

### How to reproduce?

1. Open the app and log in.
2. Send the app to the background.
3. Go to the app’s settings.
4. Begin the process to delete the app’s data and confirm.
5. Attempt to reopen the app afterward.

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/ManageSpaceActivity.kt
+++ b/app/src/main/java/com/owncloud/android/ui/activity/ManageSpaceActivity.kt
@@ -11,6 +11,7 @@ import android.os.Bundle
 import android.view.MenuItem
 import androidx.appcompat.app.AppCompatActivity
 import androidx.lifecycle.lifecycleScope
+import androidx.work.WorkManager
 import com.google.android.material.snackbar.Snackbar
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/owncloud/android/ui/activity/ManageSpaceActivity.kt
```diff
@@ -87,6 +91,7 @@ class ManageSpaceActivity :
             val result = clearApplicationData()
             withContext(Dispatchers.Main) {
                 if (result) {
+                    finishAffinity()
                     finishAndRemoveTask()
                     exitProcess(0)
```
alperozturk96, line 96: Ideally, `FileDisplayActivity` should not be launched after all accounts are removed on `APP_RESUME`. However, the app uses quite a bit of complex logic during startup.

Currently, `FirstRunActivity`, `OnboardingServiceImpl`, `LauncherActivity`, and `AuthenticatorActivity` all play a role in triggering the login flow and eventually `FileDisplayActivity`. To be honest, I’m not entirely sure why the app has four different launcher-related activities.

If you have any suggestions or ideas on a better approach to handle this, I’d really appreciate your feedback.

@ZetaTom @tobiasKaminsky

PR #16436, opened 2026-02-03
fix(file-listing): action bar title after preview

alperozturk96:

### Issue

Last opened PDF file name appears in action bar.

### How to reproduce?

1. Upload large pdf file along with multiple images
2. Open large pdf file press back
3. Open image file press back
4. Action bar title still showing the pdf file's name


app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1362,6 +1362,7 @@ class FileDisplayActivity :
         } else if (!ocFileListFragment.isSearchFragment && startFile == null) {
             updateListOfFilesFragment()
             ocFileListFragment.registerFabListener()
+            updateActionBarTitleAndHomeButton(currentDir)
```
alperozturk96, line 1371: Since this line will be executed only for file listing using current directory for action bar title should be okay.

@ZetaTom Could you please double-check this?

PR #16438, opened 2026-02-04
If enforced server, allow QR scanner, but validate url

tobiasKaminsky:

### DIFF
--- a/app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
+++ b/app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
@@ -105,6 +105,8 @@ import com.owncloud.android.utils.PermissionUtil;
 import com.owncloud.android.utils.WebViewUtil;
 import com.owncloud.android.utils.theme.ViewThemeUtils;
 
+import org.jetbrains.annotations.NotNull;
+
 import java.io.InputStream;
 import java.net.URLDecoder;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
```diff
@@ -649,10 +654,36 @@ private void parseAndLoginFromWebView(String dataString) {
...
     }
 
+    private boolean checkAllowedServers(@NonNull String server) {
+        String webviewLogin = getString(R.string.webview_login_url);
+
+        if (!webviewLogin.isEmpty() && webviewLogin.startsWith(server)) {
+            return true;
+        }
+
+        String enforcedServerList = getString(R.string.enforce_servers);
```
alperozturk96, line 681: `R.string.enforce_servers` this will be empty and it is breaking debug login script. Please check.

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/authentication/AuthenticatorActivity.java
tobiasKaminsky: It is empty in regular builds, but can be set via brander.

What login script do you mean?

alperozturk96: Debug login script we have for fast login but this changed and outdated thus I'm resolving the conversation.

alperozturk96 (approved):

PR #16450, opened 2026-02-05
fix(device): improve battery charging status detection and robustness

joshtrichards:
Mostly a refactor (code modernization and readability), but also:

- improved charge status logic (uses both EXTRA_STATUS and EXTRA_PLUGGED) 
- adds logging for diagnostics
- doc/comment improvements 

The charge status logic prevents false positives for "charging," such as when the device is plugged in but not actually charging and also handles weird devices and status results as robustly as possible.


alperozturk96:
Hi! Thanks for your contribution.

This PR has been in draft for some time, so we'd like to check whether you're still planning to work on it.

If the PR is still valid and you intend to continue, please mark it as Ready for Review (remove the draft status) and add the review label. We'll then be happy to take a look.

If you no longer plan to continue with this PR, please let us know so we can keep the backlog up to date.

Thanks!

PR #16459, opened 2026-02-06
fix(auto-upload): file detection

alperozturk96:

### Changes

The upload operation only requires read access to the file channel. Exclusive locking is unnecessary and can cause `LOCK_FAILED` errors when multiple uploads are triggered for the same local path. This change allows concurrent read access.

When a local file is detected as not eligible for auto-upload, both the auto-upload entity and the corresponding upload entity are removed from the database before continuing.

The upload file operation uses a remote existence check to determine whether a sync conflict has occurred. However, this approach is not accurate when the exact same file already exists on the remote; in such cases, there is no need to block the upload or throw an errors.

Update last scan time after uploading files since updating right after collecting files will always cause skip (`lastModified < lastScanTimestampMs`) logic to fail.

Update database results only in `UploadFileOperation`

### How to reproduce?

1. Configure auto-upload for the same local path with different remote destinations.
2. Trigger auto-upload for both configurations simultaneously.
3. One of the uploads may fail with a `LOCK_FAILED` error.

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
@@ -562,6 +562,8 @@ public class UploadsStorageManager extends Observable {
         String localPath = (upload.getLocalBehaviour() == FileUploadWorker.LOCAL_BEHAVIOUR_MOVE)
             ? upload.getStoragePath() : null;
 
+
+        Log_OC.d(TAG, "local behaviour: " + upload.getLocalBehaviour());
         Log_OC.d(TAG, "local path of upload: " + localPath);
 
         UploadStatus status = UploadStatus.UPLOAD_FAILED;
--- a/app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
@@ -1138,12 +1138,11 @@ public class UploadFileOperation extends SyncOperation {
                         Log_OC.e(TAG, "Could not delete temporal file");
                     }
                 } else {
-                    Log_OC.e(TAG, "temporal file is null, cannot delete it");
+                    Log_OC.d(TAG, "temporal file is null - internal storage is used instead of sd-card");
                 }
             } catch (Exception e) {
                 Log_OC.e(TAG, "an exception occurred during deletion of temporal file: ", e);
             }
-
 
             if (result == null) {
                 result = new RemoteOperationResult<>(ResultCode.UNKNOWN_ERROR);


app/src/main/java/com/nextcloud/client/jobs/autoUpload/AutoUploadWorker.kt
```diff
@@ -367,10 +341,6 @@ class AutoUploadWorker(
                             sendUploadFinishEvent(operation, result)
                         }
                     } catch (e: Exception) {
-                        uploadsStorageManager.updateStatus(
```
tobiasKaminsky, line 370: Please re-add, so that we have a final catch if something fails not related to UFO

app/src/main/java/com/nextcloud/client/jobs/autoUpload/FileSystemRepository.kt
```diff
@@ -160,22 +181,29 @@ class FileSystemRepository(private val dao: FileSystemDao, private val context:
             }
 
             val entity = dao.getFileByPathAndFolder(localPath, syncedFolder.id.toString())
-            val fileSentForUpload = (entity != null && entity.fileSentForUpload == 1)
-            if (fileSentForUpload) {
-                Log_OC.d(TAG, "File was sent for upload, checking if it changed...")
-            }
 
             val fileModified = (lastModified ?: file.lastModified())
```
tobiasKaminsky, line 185: Put this please in a different PR, which we then can directly merge to dev.

app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
```diff
@@ -346,10 +343,9 @@ class FileUploadWorker(
             fileUploadBroadcastManager.sendStarted(operation, context)
         } catch (e: Exception) {
             Log_OC.e(TAG, "Error uploading", e)
-            result = RemoteOperationResult<Any?>(e)
+            result = RemoteOperationResult(e)
```
tobiasKaminsky, line 356: Just like AutoUploadWorker add here a ERROR status if UFO failed.

app/src/main/java/com/owncloud/android/datamodel/UploadsStorageManager.java
```diff
@@ -550,59 +550,49 @@ public void clearSuccessfulUploads() {
...
+                upload.getUser(), new File(upload.getStoragePath()), upload.getRemotePath(), upload.getContext());
+
+            if (isSame) {
+                result = UploadResult.SAME_FILE_CONFLICT;
+                status = UploadStatus.UPLOAD_SUCCEEDED;
+            } else {
+                result = UploadResult.SYNC_CONFLICT;
             }
+        } else if (code == RemoteOperationResult.ResultCode.LOCAL_FILE_NOT_FOUND) {
+            result = UploadResult.FILE_NOT_FOUND;
```
tobiasKaminsky, line 589: re-add UploadStatus.UPLOAD_SUCCEEDED

otherwise a file not found upload will always be in failed/retry upload queue

app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -1056,105 +1052,94 @@ private RemoteOperationResult normalUpload(OwnCloudClient client) {
...
+                }
 
-            if (mCancellationRequested.get()) {
-                throw new OperationCancelledException();
-            }
+                if (mCancellationRequested.get()) {
+                    throw new OperationCancelledException();
+                }
 
-            if (result.isSuccess() && mUploadOperation != null) {
```
tobiasKaminsky, line 1112: please re-add rhe result.isSuccess check

app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -1056,105 +1052,94 @@ private RemoteOperationResult normalUpload(OwnCloudClient client) {
...
-                    Log_OC.w(TAG, "Failed to close file channel");
-                }
-            }
-
-            if (temporalFile != null && !originalFile.equals(temporalFile)) {
-                boolean isTempFileDeleted = temporalFile.delete();
-                Log_OC.d(TAG, "normalUpload, temp folder deletion: " + isTempFileDeleted);
+            // Clean up temporal file if it exists
+            if (temporalFile != null && !temporalFile.delete() && temporalFile.exists()) {
+                Log_OC.e(TAG, "Could not delete temporal file");
```
tobiasKaminsky: a bit misleading, if temporalFile is not existing

PR #16468, opened 2026-02-07
fix(manifest): Move camera2 feature declaration to scan-enabled flavors

joshtrichards:
Moving to flavor-specific manifests to avoid warning about `tools:node="replace"` for builds that don't have scanning/camera2

Eliminates this check (lint) warning:

```
> Task :app:processGenericDebugMainManifest
/home/runner/work/android/android/app/src/debug/AndroidManifest.xml Warning:
	uses-feature#android.hardware.camera2 was tagged at AndroidManifest.xml:0 to replace another declaration but no other declaration present
```

### DIFF
--- a/app/src/gplay/AndroidManifest.xml
+++ b/app/src/gplay/AndroidManifest.xml
@@ -15,6 +15,12 @@
 
     <uses-permission android:name="android.permission.READ_MEDIA_VISUAL_USER_SELECTED" />
 
+    <!-- Used for document scanning, but lib declares it as required, which it's not -->
+    <uses-feature
+        android:name="android.hardware.camera2"
+        android:required="false"
+        tools:node="replace" />
+
     <application
         android:name=".MainApp"
         android:icon="@mipmap/ic_launcher"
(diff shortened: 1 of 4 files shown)


AndyScherzinger:
Hi @joshtrichards @alperozturk96 since we have a general upload Pic via cam in the fab's bottom sheet I would have expected that it is needed in all flavors. Or are we using something else there?

joshtrichards:
@AndyScherzinger I [checked the upload-via-camera path](https://github.com/nextcloud/android/blob/2b84ff514a655f8ca4388c90e964d35a11f8f3ac/app/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java#L1089-L1097): it uses the system camera app via `ACTION_IMAGE_CAPTURE` / `ACTION_VIDEO_CAPTURE`,  so it doesn't use Camera2. The manifest override is only for the scanning dependency, which declares `camera2` as required in scan-enabled flavors.

tobiasKaminsky (approved):

PR #16481, opened 2026-02-08
build(app): remove redundant room schema location from annotationProc…

joshtrichards:
…essorOptions

Removes the 'room.schemaLocation' argument from the general annotationProcessorOptions block in app/build.gradle.kts. Fixes this warning in the build chain:
```
warning: The following options were not recognized by any processor: '[room.schemaLocation, kapt.kotlin.generated]'
```

Since KSP is already used for the Room compiler, the schema location is already correctly provided via 'ksp.arg'. Leaving it in the general javaCompileOptions caused Kapt to pass the argument to other processors (like prism4j), resulting in "unrecognized options" warnings during compilation.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist
 
### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -107,10 +107,6 @@ android {
         buildConfigField("boolean", "CI", ciBuild.toString())
         buildConfigField("boolean", "RUNTIME_PERF_ANALYSIS", perfAnalysis.toString())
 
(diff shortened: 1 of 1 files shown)


alperozturk96:
Tested building. Thank you.

PR #4853, opened 2019-11-18
Migrate simple cases of get current account to get user

ezaquarii:
This branch is an attempt to migrate all **trivial** use cases of `UserAccountManager.getCurrentAccount()` to `UserAccountManager.getUser()`, leveraging `User.toPlatformAccount()` and `User.toOwnCloudAccount()` APIs whenever refactoring becomes problematic or too risky.

Since `UserAccountManager` is injected in top-level component, such as `Activities` (**top**), this PR shall open path to migration of internal components, deeper in the object hierarchy (**bottom**).

My initial plan to start from internal componentsand taking migration up to `Activity`/`Fragment`/`Service` level - botton-top approach - turned out to not be feasible, because various internal componetns require refactoring of the callers anyway (top) atd those refactorings spill horribly all over the codebase.

I try to keep the risk of this change in reasonable bounds, but it **will** require complehensive application testing nevertheless.

PR #16484, opened 2026-02-08
refactor: remove deprecated getCurrentAccount() from interface

joshtrichards:
Remove `getCurrentAccount()` method from `CurrentAccountProvider` interface.
The method was deprecated in favor of `getUser()` and had only one external
caller (`DrawerActivity`), which has been migrated to use `getUser()` instead.

The implementation in `UserAccountManagerImpl` remains as it's still used
internally by `getUser()`, but removing it from the interface prevents new
code from depending on the deprecated Account-based API.

Refs: #4853, #4835

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist
 
### DIFF
--- a/app/src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
+++ b/app/src/main/java/com/nextcloud/client/account/UserAccountManagerImpl.java
@@ -187,9 +187,8 @@ public class UserAccountManagerImpl implements UserAccountManager {
         }
     }
 
-    @Override
     @NonNull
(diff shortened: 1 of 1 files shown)


alperozturk96 (changes requested):
Hello

Thank you for the PR. `SessionMixing` still using.

Image

and eventually other classes...

Image

joshtrichards:
Oops. I should have checked Kotlin property access paths. Doh!

PR #16507, opened 2026-02-13
fix(file-listing): local file indicator icon

alperozturk96:
### Issue

The folder/file sync icon does not appear while a folder is being downloaded. 

After the folder/file download completes, if the folder/file is deleted, the checkmark (tick) icon remains visible and does not disappear as expected

### Fix

https://github.com/user-attachments/assets/8e576437-7980-4e8e-862c-5fe7d91bc5f4


app/src/main/java/com/nextcloud/client/files/FileIndicatorManager.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.owncloud.android.R
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.update
+
+sealed class FileIndicator(val iconRes: Int?) {
+    data object Idle : FileIndicator(null)
+    data object Downloading : FileIndicator(R.drawable.ic_synchronizing)
+    data object Error : FileIndicator(R.drawable.ic_synchronizing_error)
+    data object Downloaded : FileIndicator(R.drawable.ic_synced)
```
tobiasKaminsky: please change to enum, so that this does not rely on runtime-icon-res-ids

app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -1690,6 +1697,12 @@ class FileDisplayActivity :
                 sameFile = file?.remotePath == uploadedRemotePath || renamedInUpload
             }
 
+            if (uploadWasFine) {
```
tobiasKaminsky: depending on upload move operation  it will be either green / not green

Image

tobiasKaminsky: Please also check this
- for auto-upload
- internal two way sync
- contacts and calendar backup

tobiasKaminsky: Please also add tests, at least for Upload via UploadIT.

app/src/main/java/com/owncloud/android/ui/adapter/OCFileListDelegate.kt
```diff
@@ -315,53 +313,26 @@ class OCFileListDelegate(
...
-            val icon = when {
-                isSyncing -> R.drawable.ic_synchronizing
-                hasConflict -> R.drawable.ic_synchronizing_error
-                isDown || isFullyDownloaded -> R.drawable.ic_synced
-                else -> null
+    private fun showFileIndicator(file: OCFile, holder: ListViewHolder) {
+        holder.localFileIndicator.run {
+            var indicator = file.fileIndicator
+            if (file.etagInConflict != null) {
+                indicator = FileIndicator.Error.iconRes
```
tobiasKaminsky: Please re-add sync icon for sync folder/file and upload file operation.

tobiasKaminsky: show sync icon on parent folder when uploading a file(s)

alperozturk96:
@tobiasKaminsky how can we generate screenshots to compare?

tobiasKaminsky (changes requested):
Please do it for 
calendar backup
contacts backup

PR #16523, opened 2026-02-16
Forbid to change extension with template

tobiasKaminsky:
This will create otherwise a non-expected file, e.g template with ".md" is selected, but ".txt" is file name.

To test:
- select template or no
- see default extension
- change extension
--> dialog cannot be confirmed


alperozturk96:
@kra-mo Do you think message is okay?

kra-mo:
> @kra-mo Do you think message is okay?

Sure, it works fine :)

PR #16526, opened 2026-02-16
Feat: Add UnifiedPush support

### LINKED ISSUE #3333: Use microG's Google Play library
F-Droid users cannot get "real" push notifications.

MicroG is developing a replacement library for Play Services client library: https://github.com/microg/android_external_GmsLib

It's FLOSS so it would AFAIK be allowed to be included in the F-Droid build. And if the F-Droid users do have the original Google Play Services or [microG](https://microg.org/) installed, they could use a real FLOSS app, but also get push notifications. (so stuff like https://github.com/nextcloud/android/issues/922 is not needed)

Leave it as it is; or remove Google Play Push services entirely. Both not ideal.

### Additional context
It seems to be still in development, but it's an idea, at least.

### LINKED ISSUE #5510: OpenPush instead of FCM
1) The NC app from the playstore uses Google Firebase Cloud Messaging to deliver push notifications. As Nextcloud is used by many privacy aware users, this is not ideal.
2) The NC (and NC Talk) app from f-droid lack in push notifications completely.

There is a project funded by the [PrototypeFund](https://prototypefund.de/project/open-source-push-service-fuer-android-apps/) that aims to create a free and open source self-hosted replacement for Android Push Notifications. Even though this project is not yet ready to use, we could keep track of it and maybe at some point
a) include an option in the settings to use a custom OpenPush instance instead of FCM for notifications.
b) replace Google FCM with OpenPush completely and run an OpenPush instance that is preconfigured in the Android app.

### Alternatives
The usage of microG's Google Play Services library replacement on the f-droid version (see  #3333) would solve (2).

### Additional notes
(1) is not of high priority, same goes for providing a dedicated OpenPush server instance (as proposed in (b)). But having the option to use a custom OpenPush instance would be a good thing to have in the future!

### LINKED ISSUE #8684: [Feature Request] UnifiedPush for push notifications
- The NC app from the Play-store uses Google Firebase Cloud Messaging to deliver push notifications. As Nextcloud is used by many privacy aware users, this is not ideal.
- The NC (and NC Talk) app from f-droid lack in push notifications completely.
- Given that NC is mainly relying on FCM for push notifications, as A FOSS app, the majority of people using NextCloud are those who are running away from proprietary Cloud Services likes Google Drive,... and value their privacy so there is high chance that they don't use Google Services in their phones and not many of them are using Micro-G either.

If every FOSS app that needs to push notifications implement their own push notification system, battery will drain faster because every app will have to check every x seconds, even if when optimizing the process.
I propose that NC uses the UnifiedPush project [https://unifiedpush.org/](url) which is being actively developed and is working properly. It also allows the use of FCM, so both worlds and there's a push provider for it (Gotify-UP) available at F-Droid .

The design flow would be something like:
- The Google Play version will use FCM if it is available.
- The F-Droid version will use UnifiedPush distributor (Gotify-UP, NoProvider2Push, ...) if it available.

Use Micro-G not sure if it supports FCM implementation but even that, metadata about notification are still sent via Google

### Additional context
UnifiedPush currently supports the open source push provider Gotify-UP available on F-Droid and Google FCM. It's being used by IM apps like FluffyChat and others are working on it like Element.

### LINKED ISSUE #11898: Open sources push notification
I'm using a "de-googled" android phone and I'm using Nextcloud as a open-source and independent option for "like-google" and "like-apple" services.
In the android app, under notification I got an error message "Push notifications are disabled because they need Google Play Services".

I would like to have a more free app that can work out his push notification without needing Google Play Services help. In more recent years many many applications started to develop their independent push services, so I hope it is "easy" to implement it.

Thank you.

(Easy step to reproduce the "problem": open Nextcloud android app with a phone without play services or MicroG and open "notification" tab).

p1gp1g:
Following [web push support in nextcloud/notifications](https://github.com/nextcloud/notifications/pull/2662), we can add [UnifiedPush](https://unifiedpush.org) support to the Android application\*, to get push notifications even without the Play Services.

The implementation follows this guide: https://unifiedpush.org/developers/ux/

- During startup:
    - The GPlay flavor (and the others) uses FCM by default
    - The generic flavor uses UnifiedPush by default, and will use it if a distributor (= an application that can be used as a Push Service) is available
    - With the generic flavor, if the user has many distributors and hasn't selected a default yet (nearly never the case - the default is a OS wide default), a dialog is shown to introduce the OS screen.
- Settings:
    - On the 2 flavors, if a distributor is available, you have a new setting that allows to enable/disable UnifiedPush
    - If the user has many distributors (nearly never the case - but that means the user knows what UnifiedPush is!), a 2nd setting is offered, to change the distributor

This feature also gives the possibility to get Push Notifications with the Play Services, without the proxy (Nextcloud servers directly push to Google FCM servers), and without a proprietary library. So this is an accepted way to use FCM for application in F-Droid. => This is in a 2nd PR, already ready.

Fix #11898
Fix #5510
Fix #8684 
Fix #3333 (with the 2nd PR)

\* We could add it to the Linux app too, but I haven't, at least yet.

### 🖼️ Screenshots

Coming, I will update this description

🏚️ Before | 🏡 After
---|---
image | Screenshot From 2026-02-16 23-11-48

### 🏁 Checklist
 
- [ ] Tests written, or not not needed : I don't know what can be done


p1gp1g:
And I'll start by fixing some conflicts :+1:

app/src/main/java/com/nextcloud/client/jobs/UnifiedPushWork.kt
```diff
@@ -0,0 +1,192 @@
...
+            uaPublicKey = uaPublicKey,
+            appTypes = appTypes()
+        ).execute(mClient)
+    }
+
+
+    private fun appTypes(): List<String> = context.packageManager
+        .getLaunchIntentForPackage(APP_NEXTCLOUD_TALK)?.let {
+            listOf("all", "-talk")
+        } ?: listOf("all")
```
p1gp1g, line 94: That is to get all notifications, except talk ones if talk is installed. But:
- We can decide to never receive talk notifications (if a user wants talk notif, they probably use Talk app)
- And it would be better to list the applications we want to receive notifications from

tuxArg: What about talk notifications? Is it expected that the talk app will independently handle unifiedpush notifications too?

p1gp1g: I'm cleaning up the code to open the pull request

app/src/main/java/com/owncloud/android/utils/CommonPushUtils.kt
```diff
@@ -0,0 +1,284 @@
...
+     *
+     * This is run when the application starts, and this is also where push notifications
+     * are set up for the first time
+     *
+     * Push notifications are set up for the first time with this function
+     */
+    @JvmStatic
+    fun registerCurrentPushConfiguration(activity: Activity, accountManager: UserAccountManager, preferences: AppPreferences) {
+        if (
+            (!preferences.isPushInitialized && BuildConfig.DEFAULT_PUSH_UNIFIEDPUSH)
```
p1gp1g, line 54: Instead of using a BuildConfig, we could add a function `isPlayServiceAvailable` to PushUtils (Implemented per flavor, dedicated to FCM and proxy-push) to check if the Play Services are available.

That way, a user using the gplay flavor on a phone without the Play Services can use by default UnifiedPush.

But it introduces a bit more change, and it is a bit less readable. As you want

o-kotb:
> * We could add it to the Linux app too, but I haven't, at least yet.

I think the Linux app uses websocket from [notify_push](https://github.com/nextcloud/notify_push). But maybe adapting it for UnifiedPush would be better? If so, is it possible for the website to also use UnifiedPush, and then notify_push would not be needed for both?

p1gp1g:
> > * We could add it to the Linux app too, but I haven't, at least yet.
> 
> I think the Linux app uses websocket from [notify_push](https://github.com/nextcloud/notify_push). But maybe adapting it for UnifiedPush would be better? If so, is it possible for the website to also use UnifiedPush, and then notify_push would not be needed for both?

The web app PR is being reviewed, it uses web push (this is the server->server protocol used by UnifiedPush): https://github.com/nextcloud/notifications/pull/2663

Regarding Linux, it may be interesting if used with DAV-push. But that's a lot of work

Elsensee:
> > * We could add it to the Linux app too, but I haven't, at least yet.
> 
> I think the Linux app uses websocket from [notify_push](https://github.com/nextcloud/notify_push). But maybe adapting it for UnifiedPush would be better? If so, is it possible for the website to also use UnifiedPush, and then notify_push would not be needed for both?

This is a bit off-topic, but I don't think using WebPush for the Linux app makes sense. notify_push is available, establishes a direct connection between Nextcloud server and client. With WebPush you need a middleman (a distributor).

It does not add any benefits. WebPush for Android allows to use something else than Google FCM. WebPush for Browsers allow to get push notifications if the tab is closed (and the Browser can save one connection).
But for the Linux app? It doesn't use Google FCM and there is no tab to close.

o-kotb:
My rationale was if you wanted push for both Android and desktop, you're going to have both webpush and notify_push for them respectively. But if the desktop client supported webpush, you only need one. Also, desktop doesn't connect to server directly but uses separate notify_push binary which can be on different server but most people have it on same server. I think it's same for webpush as in you can host you're own distributor on same server, but it'll be more useful for nc android and other apps like tusky and element with unifiedpush. Anyways, this is way more work so it's only thoughts. I'm glad f-droid nc finally can have push.

And yeah, since it's off-topic, I'll mark my comments as such.

p1gp1g:
There is a benefit for Linux phones. For desktop, it's more limited.

Let's mark our comments for that "thread" as off-topic :+1:

alperozturk96:
Hi! Thanks for your contribution.

Could you please resolve conflict? Then we can review.

p1gp1g:
I'm doing it

p1gp1g:
BTW, we need this PR too: https://github.com/nextcloud/android-library/pull/1949

p1gp1g:
The PR is rebased, and conflict free :+1:

Fuseteam:
hmmm github still complains about conflicts :thinking: or is the webui out of date?

app/src/main/java/com/owncloud/android/datamodel/PushConfigurationState.java
```diff
@@ -17,6 +17,8 @@ public class PushConfigurationState {
     public String deviceIdentifierSignature;
     public String userPublicKey;
     public boolean shouldBeDeleted;
+    public boolean shouldBeDisabled = false;
```
alperozturk96, line 20: Why do we need two boolean flag?

app/src/main/java/com/owncloud/android/services/UnifiedPushService.kt
```diff
@@ -0,0 +1,66 @@
+/*
+ * Nextcloud - Android Client
+ *
+ * SPDX-FileCopyrightText: 2025 Your Name <your@email.com>
```
alperozturk96: Please update the license header.

app/src/gplay/java/com/owncloud/android/utils/PushUtils.java
```diff
@@ -141,7 +158,8 @@ private static void deleteRegistrationForAccount(Account account) {
             RemoteOperationResult<Void> remoteOperationResult =
                 new UnregisterAccountDeviceForNotificationsOperation().execute(mClient);
 
-            if (remoteOperationResult.getHttpCode() == HttpStatus.SC_ACCEPTED) {
+            int status = remoteOperationResult.getHttpCode();
+            if (deleteLocalData && status == HttpStatus.SC_ACCEPTED) {
```
alperozturk96, line 162: Why `deleteLocalData` needs to be true for `UnregisterAccountDeviceForProxyOperation`.

app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
```diff
@@ -592,6 +595,84 @@ internal class BackgroundJobManagerImpl(
...
+            .build()
+
+        val request = oneTimeRequestBuilder(NotificationWork::class, JOB_NOTIFICATION)
+            .setInputData(data)
+            .build()
+
+        workManager.enqueue(request)
+    }
+
+    override fun registerWebPush(accountName: String, url: String, uaPublicKey: String, auth: String) {
```
alperozturk96: Instead of having three `registerWebPush`, `activateWebPush`, `unregisterWebPush` worker function, can't we use sealed class for three type and just internally provides data and we can just call one function.

e.g.

```
fun configureWebPush(data: WebPushData) {
      val request = oneTimeRequestBuilder(UnifiedPushWork::class, JOB_UNIFIEDPUSH)
            .setInputData(data.inputData)
            .build()

        workManager.enqueue(request)
}
```

```
sealed class WebPushData {
  data class Register(accountName: String, url: String, uaPublicKey: String, auth: String): WebPushData()
   ...
   ...

}
```

app/src/main/java/com/nextcloud/client/jobs/NotificationWork.kt
```diff
@@ -135,6 +142,22 @@ class NotificationWork constructor(
         cipher.doFinal(base64DecodedSubject)
     }
 
+    private fun handlePushMessage(accountName: String?, decryptedPushMessage: DecryptedPushMessage) {
```
alperozturk96: Please use `when`

e.g.

```
private fun handlePushMessage(
    accountName: String?,
    decryptedPushMessage: DecryptedPushMessage,
) = when {
    decryptedPushMessage.delete ->
        notificationManager.cancel(decryptedPushMessage.nid)

    decryptedPushMessage.deleteMultiple ->
        decryptedPushMessage.nids.forEach(notificationManager::cancel)

    decryptedPushMessage.deleteAll ->
        notificationManager.cancelAll()

    else ->
        fetchCompleteNotification(
            accountManager.getUser(accountName).orElseThrow { RuntimeException() },
            decryptedPushMessage,
        )
}
```

app/src/main/java/com/nextcloud/client/jobs/UnifiedPushWork.kt
```diff
@@ -0,0 +1,193 @@
...
+        val uaPublicKey = inputData.getString(EXTRA_UA_PUBKEY) ?: run {
+            Log.w(TAG, "No uaPubkey supplied")
+            return
+        }
+        val auth = inputData.getString(EXTRA_AUTH) ?: run {
+            Log.w(TAG, "No auth supplied")
+            return
+        }
+        val ocAccount = OwnCloudAccount(accountManager.getAccountByName(accountName), context)
+        val mClient = OwnCloudClientManagerFactory.getDefaultSingleton().getNextcloudClientFor(ocAccount, context)
```
alperozturk96: Instead of creating client can we make `com.nextcloud.repository.ClientRepository` injectable and inject worker.

Then call 

`val client = clientRepository.getOwncloudClient()`

app/src/main/java/com/nextcloud/client/jobs/UnifiedPushWork.kt
```diff
@@ -0,0 +1,193 @@
...
+        val accountName = inputData.getString(EXTRA_ACCOUNT) ?: run {
+            Log.w(TAG, "No account supplied")
+            return
+        }
+        val token = inputData.getString(EXTRA_TOKEN) ?: run {
+            Log.w(TAG, "No account supplied")
+            return
+        }
+        val ocAccount = OwnCloudAccount(accountManager.getAccountByName(accountName), context)
+        val mClient = OwnCloudClientManagerFactory.getDefaultSingleton().getNextcloudClientFor(ocAccount, context)
```
alperozturk96: We should unify this logic and use `ClientRepository`.

```
val ocAccount = OwnCloudAccount(accountManager.getAccountByName(accountName), context)
val mClient = OwnCloudClientManagerFactory.getDefaultSingleton().getNextcloudClientFor(ocAccount, context)
```

app/src/main/java/com/owncloud/android/services/UnifiedPushService.kt
```diff
@@ -0,0 +1,66 @@
...
+        Log.d(TAG, "Received new endpoint for $instance")
+        // No reason to fail with the default key manager
+        val key = endpoint.pubKeySet ?: return
+        backgroundJobManager.registerWebPush(instance, endpoint.url,key.pubKey, key.auth)
+    }
+
+    override fun onMessage(message: PushMessage, instance: String) {
+        Log.d(TAG, "Received new message for $instance")
+        try {
+            val mObj = JSONObject(message.content.toString(Charsets.UTF_8))
```
alperozturk96: Please avoid using the m prefix for variable names in new Kotlin classes.

app/src/main/java/com/owncloud/android/services/UnifiedPushService.kt
```diff
@@ -0,0 +1,66 @@
...
+            val token = mObj.getString("activationToken")
+            backgroundJobManager.activateWebPush(instance, token)
+        } catch (_: JSONException) {
+            // Messages are encrypted following RFC8291, and UnifiedPush lib handle the decryption itself:
+            // message.content is the cleartext
+            backgroundJobManager.startDecryptedNotificationJob(instance, message.content.toString(Charsets.UTF_8))
```
alperozturk96, line 55: Can you add `Log_OC.e` here?

app/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
```diff
@@ -355,6 +361,73 @@ private void setupSyncCategory() {
         setupAllFilesAccessPreference(preferenceCategorySync);
     }
 
+    private void setupPushCategory(PreferenceScreen preferenceScreen) {
```
alperozturk96: Please create new class and add `setupPushCategory`, `setUnifiedPushPreference` functions there and call the class in `SettingsActivity.java` so that we don't make the `SettingsActivity.java` more bigger.

app/src/main/java/com/owncloud/android/utils/CommonPushUtils.kt
```diff
@@ -0,0 +1,289 @@
...
+        accountManager: UserAccountManager,
+        preferences: AppPreferences
+    ) {
+        if (
+            (!preferences.isPushInitialized && BuildConfig.DEFAULT_PUSH_UNIFIEDPUSH)
+            || preferences.isUnifiedPushEnabled
+            ){
+            tryUseUnifiedPush(activity, accountManager, preferences) {}
+        } else {
+            CoroutineScope(Dispatchers.IO).launch {
```
alperozturk96: Please use lifecycle scope of passed activity.

app/src/main/java/com/owncloud/android/utils/CommonPushUtils.kt
```diff
@@ -0,0 +1,289 @@
...
+        }
+    }
+
+    @JvmStatic
+    fun unregisterUnifiedPushForAccount(
+        context: Context,
+        accountManager: UserAccountManager,
+        account: OwnCloudAccount
+    ) {
+        CoroutineScope(Dispatchers.IO).launch {
```
alperozturk96: Creates an unmanaged scope. Every call creates a new independent CoroutineScope that isn't tied to any lifecycle. If the Activity is destroyed or the operation should be cancelled, these coroutines keep running.

app/src/main/java/com/owncloud/android/utils/CommonPushUtils.kt
```diff
@@ -0,0 +1,289 @@
...
+     * if a server doesn't support web push, proxy push is re-registered
+     */
+    fun registerUnifiedPushForAllAccounts(
+        context: Context,
+        accountManager: UserAccountManager,
+        proxyPushToken: String?
+    ) {
+        CoroutineScope(Dispatchers.IO).launch {
+            val jobs = accountManager.accounts.map { account ->
+                CoroutineScope(Dispatchers.IO).launch {
```
alperozturk96: Nested scopes are unnecessary and also needs to be managed scope.

alperozturk96 (changes requested):

app/src/gplay/java/com/owncloud/android/utils/PushUtils.java
p1gp1g: `deleteLocalData` may not be the best name. It describes "delete user public key, and unsubscribe from the proxy".

The idea is that we can unregister for proxy push notifications in 3 cases:
- When we delete an account
- When we disable proxy-push
- When we hard-reset push subscriptions (and rotate encryption keys)

In the first and last case, we want either to wipe our current keys, and unregister on the proxy. In the 2nd case, that may simply be temporary, we don't need/want to unsubscribe

app/src/main/java/com/nextcloud/client/jobs/UnifiedPushWork.kt
p1gp1g: I see the point, and it should be used in other workers as well; But that may be outside of the scope of this PR ?

app/src/main/java/com/owncloud/android/datamodel/PushConfigurationState.java
p1gp1g: - The first one, `shouldBeDisabled` informs the next push notification worker to disable (proxy) push (where `disabled` will be set to `true`)
- the second one `disabled` when we have successfully disabled (proxy) push notifications

It works like with `shouldBeDeleted`:
- `shouldBeDeleted` inform the next push notification worker to delete push notification data (where `accountPushData` will be wiped)
- once deleted (`accountPushData == null`), it can't be used

app/src/main/java/com/owncloud/android/services/UnifiedPushService.kt
p1gp1g: If we have caught a JSONException, it simply means this is not the push notification to activate web push, but an actual push notification (the subscription is already activated). Should I improve the comment here ?

app/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
p1gp1g: Wouldn't be better to move all the categories to their own file in the same time to avoid no-uniformity ?

app/src/main/java/com/owncloud/android/utils/CommonPushUtils.kt
p1gp1g: I think it's better to not be tied to the activity lifecycle for that. Is it OK for you if I set a `private val scope = CoroutineScope(Dispatchers.IO)` like we have for `FileUploadHelper` ?

app/src/main/java/com/owncloud/android/utils/CommonPushUtils.kt
p1gp1g: There is the `jobs.joinAll()` later, that's to avoid sending requests sequentially

p1gp1g (commented):
Thanks for the review. I'm pushing a few commits.

I tried to answer some of your comments. Do you have a matrix account, or anywhere to communicate directly ? It will be easier to discuss a few points.

app/src/main/java/com/owncloud/android/utils/CommonPushUtils.kt
alperozturk96: This function will be triggered with `class TokenPushEvent` Eventbus. And it is called from FileDisplayActivity.onStart.initFile and onRequestPermissionsResult.PermissionUtil.PERMISSIONS_EXTERNAL_STORAGE

Why it is not better to tied to the activity lifecycle?

`FileUploadHelper.ioScope` should also have cancellation function. We want to avoid this in new PRs.

app/src/main/java/com/owncloud/android/utils/CommonPushUtils.kt
alperozturk96: If registerUnifiedPushForAccount throws for one account rest will be affected as well.

You can use `supervisorScope` and also please handle cancelation of the scope (CoroutineScope(Dispatchers.IO)).

[Documentation supervisorScope](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/supervisor-scope.html)

```
CoroutineScope(Dispatchers.IO).launch {
    supervisorScope {
        val jobs = accountManager.accounts.map { account ->
            launch {
                try {
                    val ocAccount = OwnCloudAccount(account, context)
                    val res = registerUnifiedPushForAccount(context, accountManager, ocAccount)
                    if (res) {
                        PushUtils.setRegistrationForAccountEnabled(account, false)
                    }
                } catch (e: Exception) {
                    Log_OC.e(TAG, "Failed to register push for ${account.name}", e)
                }
            }
        }
        jobs.joinAll()
    }
    proxyPushToken?.let {
        PushUtils.pushRegistrationToServer(accountManager, it)
    }
}
```

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/ui/activity/SettingsActivity.java
alperozturk96: Within scope of PR I wouldn't touch any other categories but in future yes could be.

app/src/main/java/com/owncloud/android/services/UnifiedPushService.kt
alperozturk96: `it simply means this is not the push notification to activate web push, but an actual push notification`

I would like to see something like this in the log.

app/src/main/java/com/nextcloud/client/jobs/UnifiedPushWork.kt
alperozturk96: Yes sure make sense. But 

```
val ocAccount = OwnCloudAccount(accountManager.getAccountByName(accountName), context)
val mClient = OwnCloudClientManagerFactory.getDefaultSingleton().getNextcloudClientFor(ocAccount, context)
```

This needs to be unified in this PR.

app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManager.kt
```diff
@@ -130,6 +131,48 @@ interface BackgroundJobManager {
     fun startMediaFoldersDetectionJob()
 
     fun startNotificationJob(subject: String, signature: String)
+    fun startDecryptedNotificationJob(accountName: String, message: String)
+
+    sealed class WebPushJobData() {
```
alperozturk96: Please put `WebPushJobData` in `owncloud/android/datamodel` package.

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/utils/CommonPushUtils.kt
p1gp1g: This job makes a request that isn't made for its response (which would have been used for the UI), but to change the state on the server. We don't want the job to finish if the user changes the activity

alperozturk96: Okay then so tied to the termination of the app?

1. Assign to a job current implementation and expose via cancel function then call cancel function termination of the app from MainApp.java

2. Internally add a mechanism to cancellation after n minutes and reassessing if its triggered.

What do you think?

p1gp1g: I'm thinking about a scope that forces the job to be run in a timeout. I'm pushing something to show, let me know what you think about after that

app/src/main/java/com/owncloud/android/utils/CommonPushUtils.kt
p1gp1g: Thanks. I'm pushing the change, and I'll use the scope within timeout

app/src/main/java/com/owncloud/android/utils/CommonPushUtils.kt
p1gp1g: Here it is: https://github.com/nextcloud/android/pull/16526/changes/3e3198256dd557214a76d483ba2cb77493316626 what do you think ?

PR #16533, opened 2026-02-17
Adds logs for failing cases

alperozturk96:
Adds logs for failing cases


app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -436,27 +436,49 @@ protected RemoteOperationResult run(OwnCloudClient client) {
         mCancellationRequested.set(false);
         mUploadStarted.set(true);
 
-        updateSize(0);
```
alperozturk96, line 439: I think this sets the `OCUpload` size to `0` but after this line upload not happening and db stuck with this value.

tobiasKaminsky: Please re-add

PR #16602, opened 2026-02-27
Fix recursive folder sync

### LINKED ISSUE #269: Need option "available offline" for folders.
Currently option "make available offline" can be swiched only for files, not for folders.
Need a way to do it ffor folder tree. 
And should be option to turn this off for some of subtrees.

l3pr-org:
## Summary

This PR fixes #269 , the inability to recursively sync folders (an issue that is 10 years old). Previously, when users tried to sync a folder with subfolders recursively, only the immediate children were synced. I spent three days of time I should have been focusing on my programming assignments on this, so please let me know if there is anything wrong with my code instead of just throwing it out. I am happy to collaborate, but I needed this feature so that I can properly sync and listen to my music.

### Root Cause

The issue had three contributing factors:

1. Race Condition: MetadataWorker and FolderDownloadWorker were running without proper dependency, causing FolderDownloadWorker to query the database before metadata was populated.

2. eTag Caching: MetadataWorker was skipping folder content refresh when eTag hadn't changed, leaving the database empty.

3. Single-Level Processing: MetadataWorker only fetched one level of subfolders, not recursively traversing into nested folders like `Artists/ABBA`, for example.

### Changes

- BackgroundJobManagerImpl.kt: Added WorkManager job dependency chain to ensure MetadataWorker runs before FolderDownloadWorker
- MetadataWorker.kt: 
  - Added FORCE_REFRESH flag to bypass eTag caching
  - Implemented BFS-based recursive folder processing to fetch metadata for ALL nested subfolders
- FolderDownloadWorker.kt: Added debug logging and warning when no files found
- UI Components: Added "Download all subfolders" option in file actions with confirmation dialog

### 🏁 Checklist
 
- [x] Tests written, or not not needed (existing tests cover core functionality; new functionality requires instrumented tests with a real Nextcloud server)
- [x] Tested locally on device


l3pr-org:
Going to paste a message from the linked issue here:

> I may have also forgotten to read the contrib guidelines in this case... whoops 😅 I forgot to sign my commits and went past the line length limit for a couple lines of code, but I can fix these when I have a little more time.

>I think even if you (Nextcloud maintainers) take issue with the things I did wrong, we will end up with a working implementation in the end, which is nice.

app/src/main/java/com/nextcloud/client/jobs/folderDownload/FolderDownloadWorker.kt
```diff
@@ -182,9 +191,40 @@ class FolderDownloadWorker(
         return file
     }
 
-    private fun getFiles(folder: OCFile, storageManager: FileDataStorageManager): List<OCFile> =
-        storageManager.getFolderContent(folder, false)
-            .filter { !it.isFolder && !it.isDown }
+    private fun getFiles(folder: OCFile, storageManager: FileDataStorageManager, recursive: Boolean): List<OCFile> {
+        if (recursive) {
```
alperozturk96, line 195: Could you please use `FiledatastorageManager.getAllFilesRecursivelyInsideFolder()` together with the `OCFile.isDown` check instead of introducing a new function?

Using `isDown` should not negatively impact this process; in fact, it helps reduce unnecessary operations.

Also, let’s remove the comments where possible. Ideally, the code should be self-explanatory and clear enough that extensive comments aren’t needed.

app/src/main/java/com/nextcloud/client/jobs/folderDownload/FolderDownloadWorker.kt
```diff
@@ -85,7 +88,13 @@ class FolderDownloadWorker(
 
         return withContext(Dispatchers.IO) {
             try {
-                val files = getFiles(folder, storageManager)
+                val files = getFiles(folder, storageManager, recursiveDownload)
+                
+                // Add warning log when no files found for recursive download
+                if (files.isEmpty()) {
```
alperozturk96, line 94: Why do we need warning log here? It is just extra info if file list is empty.

Please change with info log and remove the emoji along with comment since Log already prints.

app/src/main/java/com/nextcloud/client/jobs/folderDownload/FolderDownloadWorker.kt
```diff
@@ -182,9 +191,40 @@ class FolderDownloadWorker(
...
+        }
+        // Use folder ID to avoid fileExists() check
+        return storageManager.getFolderContent(folder.fileId, false)
+            .filter { !it.isFolder }
+    }
+
+    /**
+     * Recursively get all files in the folder and its subfolders
+     */
+    private fun getAllFilesRecursive(folder: OCFile, storageManager: FileDataStorageManager): List<OCFile> {
```
alperozturk96, line 206: Please remove this and use `FiledatastorageManager.getAllFilesRecursivelyInsideFolder()`

app/src/main/java/com/nextcloud/client/jobs/BackgroundJobManagerImpl.kt
```diff
@@ -795,25 +796,62 @@ internal class BackgroundJobManagerImpl(
         workManager.enqueueUniquePeriodicWork(JOB_INTERNAL_TWO_WAY_SYNC, ExistingPeriodicWorkPolicy.UPDATE, request)
     }
 
-    override fun downloadFolder(folder: OCFile, accountName: String) {
+    override fun downloadFolder(folder: OCFile, accountName: String, recursive: Boolean) {
```
alperozturk96, line 790: Thank you for the PR and your time.

I’d suggest keeping the scope of this PR strictly limited. The changes proposed outside of `FolderDownloadWorker` are quite significant and would require thorough testing.

For now, let’s focus only on `FolderDownloadWorker` with a minimal adjustment, for example, adding the new boolean parameter. From the UI side, we can introduce a separate action for this behavior, and have the worker that uses either the full nested structure or only the first-level depth based on that parameter.

Keeping the scope small will help us review more effectively and move forward with confidence. Once this is merged and validated, we can discuss and plan the next steps separately.

Thanks for understanding and we can test, review other findings in separate PR.

alperozturk96:

Please feel free to reopen it once the changes are addressed. Thank you again for your contribution.

Whiskeysaurus:
> I’ll be closing this PR for now to keep things organized.
> 
> Please feel free to reopen it once the changes are addressed. Thank you again for your contribution.

Certainly not the place to inquire, but if these doesn't meet your standards, is there chance of further investigation of the issue that this attempted to fix? It's confusing how this is not the number one priority for nextcloud core contributors. It's a 10-year bug that renders the android app completely unusable for almost every possible use case. Android users largely don't use nextcloud, and this is the reason.

I'd direct you here:
https://github.com/nextcloud/android/issues/269

PR #15859, opened 2025-10-28
Chore: Improve Gradle Configuration

jimlyas:
## Description
This Merge Include:

### Migrating from Groovy DSL to Kotlin DSL
Migrate from Groovy to Kotlin DSL for the build script

- build.gradle to build.gradle.kts
- jacoco.gradle to jacoco.gradle.kts
- settings.gradle to settings.gradle.kts
- app/build.gradle to app/build.gradle.kts
- appscan/build.gradle to appscan/build.gradle.kts

there's some improvement to like grouping some configuration and setup in one place.

## Generating jacoco report task
Previously, the project run generate report task to trigger the unit test task.

This PR reverse that, so we only need to run the unit test and after it finished, then Gradle will run the task that generate jacoco report.

Another reason being is, when running the unit test task, Android Studio will display Run Test window that help visibility to see which task is passed and which task is failed.

## Version Catalog
- Migrate from extras-based versioning to use version catalog for versioning
- Implement all the convention Gradle plugin from version catalog
- Group dependencies together using bundles

- [x] Tests written, or not not needed

PR #16612, opened 2026-02-28
Fix: Instrumented tests are executed in Drone CI again

### LINKED ISSUE #16350: Many test failures are not catched in the CI

### The problem

See PR #16329, where I systematically provoked failures in automated tests I had recently added.
None of those were catched by the CI
- Screenshot tests failed, but they are broken on master anyways due to some permission issues. And these were no ScreenshotTests, so they were rightfully not included here.
- The Drone job also did not execute them.

This is very dangerous, as it conveys a false sense of security and confidence in the code quality.

It is also quite frustrating, having put effort into stuff developers usually don't like (automating tests), to prevent regressions, and none of it would actually matter.

### The cause

Unclear right now - something seems to be wrong in the test selection setup, where we differentiate between
- Unit tests
- Screenshot tests
- Drone tests

Many tests seem to be picked up in neither category

PhilLab:
The Drone CI job is running ./gradlew createGplayDebugCoverageReport This only executes tests with coverage.

In 5fd2e2942c2b375b86c52256f7c8921395f5594c (PR #15859), the line

    testCoverageEnabled = project.hasProperty("coverage")

was replaced with

    enableUnitTestCoverage = project.hasProperty("coverage")

which only left the unit tests with coverage

Enabling the coverage for androidTest (the instrumented tests) makes Drone run them again.

Closes #16350

### 🏁 Checklist
 
### DIFF
--- a/app/src/androidTest/java/com/owncloud/android/datamodel/UploadStorageManagerTest.java
+++ b/app/src/androidTest/java/com/owncloud/android/datamodel/UploadStorageManagerTest.java
@@ -113,10 +113,12 @@ public class UploadStorageManagerTest extends AbstractIT {
         }
 
         OCUpload[] storedUploads = uploadsStorageManager.getAllStoredUploads();
-        assertEquals(size, uploadsStorageManager.getAllStoredUploads().length);
+        assertEquals(size, storedUploads.length);
+        assertEquals(uploads.size(), storedUploads.length);
 
         for (int i = 0; i < size; i++) {
-            assertTrue(contains(uploads, storedUploads[i]));
+            assertTrue("Upload " + (i+1) + "/" + size + " not found in stored uploads: " + storedUploads[i].getLocalPath(),
+                contains(uploads, storedUploads[i]));
         }
     }
 
@@ -138,7 +140,7 @@ public class UploadStorageManagerTest extends AbstractIT {
 
     private boolean contains(ArrayList<OCUpload> uploads, OCUpload storedUpload) {
         for (int i = 0; i < uploads.size(); i++) {
-            if (storedUpload.isSame(uploads.get(i))) {
+            if (storedUpload.isSame(uploads.get(i), true)) {
                 return true;
             }
         }
--- a/app/src/main/java/com/owncloud/android/db/OCUpload.java
+++ b/app/src/main/java/com/owncloud/android/db/OCUpload.java
@@ -31,6 +31,7 @@ import com.owncloud.android.utils.MimeTypeUtil;
 import java.io.File;
 
 import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /**
@@ -292,12 +293,18 @@ public class OCUpload implements Parcelable {
         return this.hashCode();
     }
 
-    @SuppressFBWarnings("SEO_SUBOPTIMAL_EXPRESSION_ORDER")
+
     public boolean isSame(@Nullable Object obj) {
+        return isSame(obj, false);
+    }
+
+    @SuppressFBWarnings("SEO_SUBOPTIMAL_EXPRESSION_ORDER")
+    @VisibleForTesting
+    public boolean isSame(@Nullable Object obj, boolean ignoreUploadId) {
         if (!(obj instanceof OCUpload other)) {
             return false;
         }
-        return this.uploadId == other.uploadId &&
+        return (ignoreUploadId || this.uploadId == other.uploadId) &&
             localPath.equals(other.localPath) &&
             remotePath.equals(other.remotePath) &&
             accountName.equals(other.accountName) &&


app/build.gradle.kts
```diff
@@ -130,6 +130,7 @@ android {
 
             debug {
                 enableUnitTestCoverage = project.hasProperty("coverage")
+                enableAndroidTestCoverage = project.hasProperty("coverage")
```
PhilLab, line 133: @tobiasKaminsky with this, Drone now actually runs the instrumented tests again. See https://drone.nextcloud.com/nextcloud/android/28585/1/3 as a proof of failure, i.e. broken tests on master which were never detected.

I'll go ahead addressing the test failures. But there is several other things I noticed:
1. We should have Drone only run the instrumentation tests. ~~Currently, drone CI runs everything for which coverage is enabled. Why this is chosen as filter, I don't know. But what this means is that Drone wastes time on the unit tests which we already run elsewhere.~~
    - Unit tests were the only tests Drone ran since #15859. Another weird side-effect of that PR.
    - ~~Easiest for saving drone time would be to disable the coverage for unit tests in the line above, but I guess this is not desired?~~
3. Actually, what about the coverage reporting? It doesn't appear anymore. It looks broken, in several ways:  - I am not seeing the coverage report on any PR I have recently checked. Don't know when this stopped. Many of the pulls on Codecov just show errors: https://app.codecov.io/github/nextcloud/android/pulls
    - The unit test job seems to still upload to Codecov, but prints out wrong links in the log, e.g.https://app.codecov.io/github/***/android/commit/2c1438fba2fda500d3ec2e330d2aa62284ef8e2c ([log](https://github.com/nextcloud/android/actions/runs/22518873081/job/65240710425?pr=16612)) .
    - You need to replace the `***` with `nextcloud`, then it works. Is this redaction of secrets at play - if so, I surely hope, no secret token has simply "nextcloud" as value?
3. Coverage reporting should also be extended by the instrumentation tests. I don't see them uploading anything

scripts/uploadReport.sh
```diff
@@ -51,7 +51,7 @@ if [ -z $USER ] || [ -z $PASS ]; then
 fi
 
 if [ $TYPE = "IT" ]; then
-    FOLDER=app/build/reports/androidTests/connected/flavors/gplay
+    FOLDER=app/build/reports/androidTests/connected/debug/flavors/gplay
```
PhilLab, line 54: @tobiasKaminsky these adjustments are needed so the report files are actually found. But reporting still is broken, which you can see e.g. in the log of https://drone.nextcloud.com/nextcloud/android/28585/1/3:

It does upload the report to https://www.kaminsky.me/nc-dev/android-integrationTests/28585-IT-stable-10-44 but it fails to create the expected Github comment notifying the developer of this.
The error is:
```
{
  "message": "Must have admin rights to Repository.",
  "documentation_url": "https://docs.github.com/rest/issues/comments#create-an-issue-comment",
  "status": "403"
}
```

app/src/androidTest/java/com/nextcloud/utils/UploadDateTests.kt
```diff
@@ -47,10 +49,15 @@ class UploadDateTests {
     fun setup() {
         context = InstrumentationRegistry.getInstrumentation().context
         MockKAnnotations.init(this, relaxed = true)
-        mockkStatic(System::class)
+        mockkStatic(System::currentTimeMillis)
```
PhilLab: @alperozturk96 see the commit message for more details:

    fix(test): Futile attempt at fixing the UploadDateTests crash
    
    The tests themselves run through, but immediately after the android process
    crashes. Bluntly mocking the whole System class causes problems everywhere.
    Even limiting the mocking to a single method still causes the crashes.
    The issue is that mocking System.currentTimeMillis() on Android
    (ART runtime) is fundamentally unsafe, even with a function reference.
    The mock interferes with framework internals (WorkManager, schedulers,
    etc.) that run after the test completes, causing the process crash.

You should also see this test crash on your local machine. The test itself is green, but the whole test run is failed:

Image

And the same now happens on the Drone job, now that it correctly picks up all expected tests.

In the best case, the UploadDateTests **should be a unit test**, instead of an instrumented test. There, I expect the mocking of currentTimeMillis() to not interfere with outside processes. But in the git history I saw that you initially even had that test as unit test and then moved it over. As it is not explained in the commits, I don't know why.
Does it have to do with the context being required for the R.string and for `DateUtils.getRelativeDateTimeString`?
We could mock it using Robolectric, but that would be a new dependency I don't want to introduce into the repo without consultation.

alperozturk96:
@tobiasKaminsky Could you please share your feedback regarding script changes?

app/src/androidTest/java/com/nextcloud/utils/UploadDateTests.kt
PhilLab: @alperozturk96 do you have an idea how to best fix this test? You added it not long ago, maybe your memory is still fresh?

alperozturk96: `DateUtils.getRelativeDateTimeString` calls `android.text.formatDateUtils` and context used in it.

I think this function causing problem.

```
public static CharSequence getRelativeDateTimeString(Context c, long time, long minResolution,
            long transitionResolution, int flags) {
        // Same reason as in formatDateRange() to explicitly indicate 12- or 24-hour format.
        if ((flags & (FORMAT_SHOW_TIME | FORMAT_12HOUR | FORMAT_24HOUR)) == FORMAT_SHOW_TIME) {
            flags |= DateFormat.is24HourFormat(c) ? FORMAT_24HOUR : FORMAT_12HOUR;
        }

        return RelativeDateTimeFormatter.getRelativeDateTimeString(Locale.getDefault(),
                TimeZone.getDefault(), time, System.currentTimeMillis(), minResolution,
                transitionResolution, flags);
    }
```

We need to provide proper alternative of `RelativeDateTimeFormatter.getRelativeDateTimeString` for test. I have to check it. If I get a chance to look at it, I will check it.

And yes let's not introduce another dependency.

Thank you for raising this.

PhilLab: @alperozturk96 I found a way without mocking the clock at all. Plus a small API change I think is required. Take a look at the two latest commits I just pushed, specifically.

By the way: If you do prefer a mockable approach, i.e. injection of a fixed clock into `getRelativeDateTimeString()`, I also have this prepared, on a temporary branch [ph/fix_instrumented_testing_alternative](https://github.com/nextcloud/android/commit/d196a531f69f768db971ef239e637ab55215891e). But I don't think it is necessary.

app/src/androidTest/java/com/nextcloud/utils/AutoRenameTests.kt
```diff
@@ -243,11 +244,32 @@ class AutoRenameTests : AbstractOnServerIT() {
...
+            versionMayor = NextcloudVersion.nextcloud_32.majorVersionNumber
             isWCFEnabled = CapabilityBooleanType.FALSE
         }
-        val filename = "   readme.txt  "
-        val result = AutoRename.rename(filename, capability, isFolderPath = true)
+        var result = AutoRename.rename(filename, capability, isFolderPath = true)
+        assert(result == filename) { "Expected $filename but got $result" }
+
+        // Nextcloud 30-31+: Always applies WCF restrictions, even if flag is set to false
+        capability = capability.apply {
```
alperozturk96: Please remove the inline comment and extract that logic into a separate test function.

* **Nextcloud 30–31:** The test should still attempt to apply WCF restrictions, even when the feature is not enabled. In this case, the JSON (e.g., `forbiddenFilenameCharactersJson`) may simply be empty.
* **Nextcloud 32+:** If the feature is not enabled, the test should skip applying the restrictions.

We already documented in `OCCapability.checkWCFRestrictions()`.

Also, could you clarify why this test was modified? Was it failing previously?

alperozturk96 (changes requested):
Thank you so much 👍

app/src/androidTest/java/com/nextcloud/utils/AutoRenameTests.kt
PhilLab: @alperozturk96 
> Also, could you clarify why this test was modified? Was it failing previously?

Yes, it failed. See the commit message of the commit `fix(test) AutoRenameTests don't fail on stable server anymore` for the explanation. And a failed run can be found here:
- Drone run: https://drone.nextcloud.com/nextcloud/android/28600/1/3
- Uploaded test results: https://www.kaminsky.me/nc-dev/android-integrationTests/28600-IT-stable-12-26

I separated the test into individual test functions, please take a look

scripts/uploadReport.sh
PhilLab: @tobiasKaminsky I am not sure what to do about this, and about the discussion thread above, where I also pinged you.

PhilLab:
@alperozturk96 I fixed the wrong timestamp string. As I only develop after work, I never noticed that the hour of day does not have a trailing zero :smile: 

Also, I fixed the next failure, which was of `UploadStorageManagerTest > largeTest`, see
see https://drone.nextcloud.com/nextcloud/android/28752/1/3 . Added an extra commit with the explanation in the message.

In general, it looks like that the tests cannot complete, not even upload the test failure report, because they are running into a timeout. Maybe not executing the unit tests (see my thread above) will buy us the necessary time

PhilLab:
@alperozturk96 thank you for the collaboration on this. Was it merged now on purpose? Because - see my previous message - the Drone runs are now consistently failing. They run into the 1 hour timeout.

Did you plan to address this in a follow-up MR?

Also @tobiasKaminsky I think the open discussions above are still relevant:
1. https://github.com/nextcloud/android/pull/16612#discussion_r2867653447
2. https://github.com/nextcloud/android/pull/16612#discussion_r2867662449

alperozturk96:
> @alperozturk96 thank you for the collaboration on this. Was it merged now on purpose? Because - see my previous message - the Drone runs are now consistently failing. They run into the 1 hour timeout.
> 
> Did you plan to address this in a follow-up MR?
> 
> Also @tobiasKaminsky I think the open discussions above are still relevant:
> 
> 1. [Fix: Instrumented tests are executed in Drone CI again #16612 (comment)](https://github.com/nextcloud/android/pull/16612#discussion_r2867653447)
> 2. [Fix: Instrumented tests are executed in Drone CI again #16612 (comment)](https://github.com/nextcloud/android/pull/16612#discussion_r2867662449)

Hey this happens for master as well and since these changes not harming something else I think it's okay to merge. We can fix other issues in different PRs.

Thank you for your contribution.

PhilLab:
@alperozturk96
> Hey this happens for master as well

sorry to be annoying but no, the tests running into an 1 hour timeout is only happening since merging this here.

The last merge before that, in https://github.com/nextcloud/android/pull/16622, had an _almost_ successful Drone run: https://drone.nextcloud.com/nextcloud/android/28739/2/1 . You can see it took only 9 minutes for test-stable and 20 minutes for test-master. [The one failure in test-master](https://www.kaminsky.me/nc-dev/android-integrationTests/28739-IT-master-15-58/) I would say is a spurious failure only.

Since merging the PR here, the drone tests finally test what they suppose to test. But they timeout, so they are still not useful. I wonder how to address this :thinking:

app/build.gradle.kts
PhilLab: regarding
> We should have Drone only run the instrumentation tests

This seems to be the case already :thinking: I don't know how this works, but adding coverage to the androidTests somehow removed the unitTests automatically. Quite unexpected, but the outcome is the desired one, i.e. Drone is not executing unit tests anymore.

PR #16618, opened 2026-03-02
fix(upload-db) duplicated entries

alperozturk96:

### DIFF
--- a/app/src/androidTest/java/com/owncloud/android/utils/FileUploadHelperTest.kt
+++ b/app/src/androidTest/java/com/owncloud/android/utils/FileUploadHelperTest.kt
@@ -11,6 +11,7 @@ import com.nextcloud.client.database.entity.UploadEntity
 import com.nextcloud.client.database.entity.toOCUpload
 import com.nextcloud.client.database.entity.toUploadEntity
 import com.nextcloud.client.jobs.upload.FileUploadHelper
+import com.nextcloud.utils.extensions.checkWCFRestrictions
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nextcloud/client/jobs/autoUpload/SyncFolderHelper.kt
```diff
@@ -56,6 +58,9 @@ class SyncFolderHelper(private val context: Context) {
             subFolderRule
         )
 
+        val capability = CapabilityUtils.getCapability(context)
+        result = AutoRename.rename(result, capability)
```
alperozturk96, line 62: Activate auto-rename if exists before auto-upload because after this call DB insertions will start

app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadHelper.kt
```diff
@@ -243,23 +248,69 @@ class FileUploadHelper {
...
+            } else {
+                createOCUpload()
+            }
         }
         backgroundJobManager.startFilesUploadJob(user, uploads.getUploadIds(), showSameFileAlreadyExistsNotification)
     }
 
+    @Suppress("ReturnCount")
+    fun getUploadByPaths(accountName: String, localPath: String, remotePath: String): UploadEntity? {
+        uploadsStorageManager.uploadDao.getUploadByAccountAndPaths(
```
alperozturk96: only use if wcf is enabled

PR #16621, opened 2026-03-03
fix(db): file-deletion

alperozturk96:
### Issue

Child of directory is not gets deleted. `getContentProviderClient()` is always null and `getContentResolver().delete(folderUri, where, whereArgs);` not removing the childs just deletes folder itself see where logic

```
String where = ProviderTableMeta.FILE_ACCOUNT_OWNER + AND + ProviderTableMeta.FILE_PATH + "=?";
String[] whereArgs = new String[]{user.getAccountName(), folder.getRemotePath()};
```
.

### Changes

Replaces content resolver functions with `DAO` functions.
Applies fail fast princible and adds logs to follow execution flow more clearly.
Adds tests.

### DIFF
--- a/app/src/androidTest/java/com/nextcloud/test/FileDeletionTests.kt
+++ b/app/src/androidTest/java/com/nextcloud/test/FileDeletionTests.kt
@@ -274,7 +274,7 @@ class FileDeletionTests : AbstractIT() {
         assertNull(storageManager.getFileById(folder.fileId))
 
         // subdirectories and files are removed
-        val children = storageManager.getFolderContent(folder, false)
+        val children = storageManager.getAllFilesRecursivelyInsideFolder(folder)
         assertTrue(children.isEmpty())
 
         // local folder removed


app/src/androidTest/java/com/nextcloud/test/FileDeletionTests.kt
```diff
@@ -0,0 +1,296 @@
...
+        val result = storageManager.removeFile(file, true, true)
+
+        assertTrue(result)
+
+        assertNull(storageManager.getFileById(file.fileId))
+        assertFalse(File(file.storagePath).exists())
+    }
+
+    @Test
+    fun deleteFolderRecursive() {
```
tobiasKaminsky, line 266: This "only" tests first folder hierarchy.
Please add a real hierarchy / recursive test.

alperozturk96:
Screenshot 2026-03-04 at 11 05 44

PR #16623, opened 2026-03-03
fix(richdocuments): Use filename from DownloadAs postmessage for correct file extension

Rash419:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist
 
### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/RichDocumentsEditorWebView.kt
+++ b/app/src/main/java/com/owncloud/android/ui/activity/RichDocumentsEditorWebView.kt
@@ -169,8 +169,13 @@ class RichDocumentsEditorWebView : EditorWebView() {
                 val url = downloadJson.getString(URL).toUri()
                 when (downloadJson.getString(TYPE)) {
                     PRINT -> printFile(url)
+
                     SLIDESHOW -> showSlideShow(url)
-                    else -> downloadFile(url, fileName)
+
+                    else -> {
+                        val downloadFileName = downloadJson.optString(FILENAME, fileName)
+                        downloadFile(url, downloadFileName)
+                    }
                 }
             } catch (e: JSONException) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/ui/activity/RichDocumentsEditorWebView.kt
```diff
@@ -169,8 +169,13 @@ class RichDocumentsEditorWebView : EditorWebView() {
                 val url = downloadJson.getString(URL).toUri()
                 when (downloadJson.getString(TYPE)) {
                     PRINT -> printFile(url)
+
                     SLIDESHOW -> showSlideShow(url)
-                    else -> downloadFile(url, fileName)
+
+                    else -> {
+                        val downloadFileName = downloadJson.optString(FILENAME, fileName)
+                        downloadFile(url, downloadFileName)
```
alperozturk96, line 177: Hi, thank you for the PR.

I have a quick question regarding the `fileName` handling. Is the `fileName` always incorrect, or is it only missing the extension in some cases? Also, can we confirm whether `downloadJson.optString(FILENAME, fileName)` consistently returns valid data?

Could you please share the steps to reproduce the issue and how you tested this? That would help us better understand the scenario and verify the behavior on our side.

Rash419: Hi,

`fileName` extension  is incorrect when using "Download as"  button to download file into different format. In Collabora Online side, I added https://github.com/CollaboraOnline/online/pull/14344 `filename` property to the DownloadAs postmessage JSON which will filename with correct extension. 

Steps to reproduce:
1. Open writer document
2. Hamburger Menu > Download as -> rtf
3. File will have wrong extension on download

PR #16645, opened 2026-03-10
feat: create encrypted folder action

alperozturk96:
### Changes

Accesses remote operation result data after checking it
Uses stream API to update encryption status of file
Adds insert file functionality to the `OCFileListAdapter`
Notifies changed item instead of full adapter
Extends functionality of create folder operation to provide full flow for directly creating encrypted folder
Converts `OCFileListBottomSheetActions` java interface to Kotlin
Adds create a new encrypted folder action for root directory (mentioned changes above were necessary to achieve full flow as showed in the demo except kt conversion)

### Screenshot

Screenshot_20260310_095354

### Demo

[Screen_recording_20260310_111421.webm](https://github.com/user-attachments/assets/1ceb4911-e785-4b79-921e-abe0c4b9d80b)

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/CreateFolderOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/CreateFolderOperation.java
@@ -59,7 +59,7 @@ public class CreateFolderOperation extends SyncOperation implements OnRemoteOper
 
     protected String remotePath;
     private RemoteFile createdRemoteFolder;
-    private boolean encrypt = false;
+    private volatile boolean encrypt = false;
     private final User user;
     private final Context context;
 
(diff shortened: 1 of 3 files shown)


app/src/main/res/drawable/ic_action_create_encrypted_dir.xml
```diff
@@ -0,0 +1,15 @@
+<!--
+  ~ Nextcloud - Android Client
+  ~
+  ~ SPDX-FileCopyrightText: 2018-2026 Michael Irigoyen
```
alperozturk96, line 4: @AndyScherzinger  I think license is correct https://pictogrammers.com/library/mdi/icon/folder-key-outline ?

AndyScherzinger: Yes, apache2 👍

PR #16652, opened 2026-03-11
feat: navigator

alperozturk96:
### Approach

Introduce `NavigatorActivity` shell activity that hosts fragments via `FragmentContainerView`. All screens that previously had their own activity are migrated to fragments and pushed through `Navigator`. `NavigatorScreen` sealed class represents each destination.

`DrawerActivity` launches `NavigatorActivity` with the target `NavigatorScreen` as a parcelable extra, keeping the call site clean with no per-screen logic.

### Why FileDisplayActivity is not used?

Other activities like `SettingsActivity` and various non-drawer activities also trigger drawer navigation items. Wiring `Navigator` into `FDA` from the start would require every caller to hold a reference to FDA or route through it, which is not possible without a larger refactor. e.g. Navigating from `CommunityActivity` to some other activity.

This way any activity can launch `NavigatorActivity`  via a simple Intent, no `FDA` dependency required. Once all navigation is consolidated into `NavigatorActivity` we can remove it and just use `FDA`

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/navigation/NavigatorScreen.kt
+++ b/app/src/main/java/com/owncloud/android/ui/navigation/NavigatorScreen.kt
@@ -10,7 +10,7 @@ package com.owncloud.android.ui.navigation
 import android.os.Parcelable
 import kotlinx.parcelize.Parcelize
 
-sealed class NavigatorScreen(val tag : String) : Parcelable {
+sealed class NavigatorScreen(val tag: String) : Parcelable {
 
     @Parcelize
     object Community : NavigatorScreen("Community")


app/src/main/java/com/owncloud/android/ui/navigation/Navigator.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.owncloud.android.ui.navigation
+
+import android.view.View
+import androidx.fragment.app.FragmentManager
+import com.owncloud.android.ui.fragment.community.CommunityFragment
+
+class Navigator(private val fragmentManager: FragmentManager, private val fragmentContainer: View) {
+    fun push(screen: NavigatorScreen) {
```
alperozturk96, line 20: Do not push same fragment.

tobiasKaminsky: :+1: Fine to do it in another PR

PR #16671, opened 2026-03-12
Investigating the Drone timeout issue

PhilLab:
## Notes
- No timestamp printing needed. The Drone log UI shows the elapsed seconds on the right, when opening full screen.
- Looking at the prior MR's drone run of test-stable: https://drone.nextcloud.com/nextcloud/android/28763/1/3
   -  assembleGplayDebugAndroidTest: 7.5 minutes
   - waiting for Emulator: 5 seconds
   - installGplayDebugAndroidTest: 26 second
   - waiting for server & deleting old Github comments: 10 seconds
   - createGplayDebugCoverageReport (configuring the tasks): 2 minutes
   - createGplayDebugCoverageReport (actually running the tests): ? - timeout
     - After task configuration, the actual test runs in the logs are only spanning 24 minutes.
   - the last log line `android(AVD) - 16 Tests 417/569 completed. (0 skipped) (0 failed)` appears at second 2058 (**34 minutes total time**). This is shorter than the stated overall time of 60 minutes.
   - So either there is something crashing so hard that the job dangles until timeout. Or the job is stopped too early.
   - The parallel `server` job's last log line is also at around 34 minutes
- Looking at this MR's drone run https://drone.nextcloud.com/nextcloud/android/28833/1/3
  - the last log line `android(AVD) - 16 Tests 439/595 completed. (0 skipped) (7 failed)` appears at second 2973 seconds (49 minutes total time).

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist
 
### DIFF
--- a/app/src/androidTest/java/com/owncloud/android/providers/DocumentsStorageProviderIT.kt
+++ b/app/src/androidTest/java/com/owncloud/android/providers/DocumentsStorageProviderIT.kt
@@ -32,6 +32,7 @@ import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertTrue
 import org.junit.Before
+import org.junit.Ignore
 import org.junit.Test
 import kotlin.random.Random
 
@@ -195,6 +196,7 @@ class DocumentsStorageProviderIT : AbstractOnServerIT() {
 
     @Suppress("MagicNumber")
     @Test(timeout = 5 * 60 * 1000)
+    @Ignore("Problem with endless sleep")
     fun testServerChangedFileContent() {
         // create random file
         val file1 = rootDir.createFile("text/plain", RandomStringGenerator.make())!!
@@ -234,6 +236,7 @@ class DocumentsStorageProviderIT : AbstractOnServerIT() {
     }
 
     @Test
+    @Ignore("Problem with endless sleep")
     fun testServerSuccessive() {
         // create random file
         val file1 = rootDir.createFile("text/plain", RandomStringGenerator.make())!!
--- a/app/src/androidTest/java/com/owncloud/android/ui/preview/PreviewImageActivityIT.kt
+++ b/app/src/androidTest/java/com/owncloud/android/ui/preview/PreviewImageActivityIT.kt
@@ -39,6 +39,7 @@ import org.junit.Ignore
 import org.junit.Test
 import java.io.File
 
+@Ignore("Investigating Drone CI timeout")
 class PreviewImageActivityIT : AbstractOnServerIT() {
     companion object {
         private const val REMOTE_FOLDER: String = "/PreviewImageActivityIT/"


PhilLab:
@alperozturk96 I am trying to experiment with the Drone runs. Extending the timeout apparently requires an approval in the Drone UI, as I changed the .drone.yml file. For security purposes.

I don't have the permission to approve this change - so can you take a look?
image

alperozturk96:
> @alperozturk96 I am trying to experiment with the Drone runs. Extending the timeout apparently requires an approval in the Drone UI, as I changed the .drone.yml file. For security purposes.
> 
> I don't have the permission to approve this change - so can you take a look? image

@tobiasKaminsky fyi

PhilLab:
@tobiasKaminsky @alperozturk96 Yesterday and today, https://drone.nextcloud.com is unreachable. Have we stopped using Drone or is it "just" down?

PhilLab:
> @tobiasKaminsky @alperozturk96 Yesterday and today, https://drone.nextcloud.com is unreachable. Have we stopped using Drone or is it "just" down?

Created ticket for this: #16801

alperozturk96:
We will check this and the created ticket. Thank you.

.drone.yml
```diff
@@ -2,6 +2,7 @@
 kind: pipeline
 type: docker
 name: tests-stable
+timeout: 90
```
tobiasKaminsky: I cannot find any documentation for this: https://docs.drone.io/yaml/docker/
Is this supposed to cancel a job after 90min?

I think we already have a similar approach configured within drone server.

tobiasKaminsky (changes requested):

.drone.yml
tobiasKaminsky: image

drone config as admin

PhilLab: @tobiasKaminsky hm, seems to be a false friend and not a valid attribute indeed - I had simply assumed that Drone would complain about invalid syntax.

I just wanted to see if there is _any_ chance of the current integration tests to succeed after a longer time frame, because running locally the results are different with every run.
But we shouldn't extend the timeout for the whole project, it would not solve the slow execution anyways. Seems we need to follow a different route.

alperozturk96:
@PhilLab Hey, do we still need this PR?

PhilLab:
@alperozturk96 
> @PhilLab Hey, do we still need this PR?

Yes and no.
**Yes**, because we still need to resolve the drone instrumentation runs - they are not able to pass anymore. The only situation when they are green is, when they skip due to no changes. Haven't you noticed this yet?

**No**, because in its current form, it doesn't yet solve the issues. I committed changes such that endlessly waiting tests are skipped so I can get a full picture of the other problematic tests down the line. But still, after one hour not all tests have been executed and it cancels due to timeout. We would need to
1. identify where the time is lost (I suspect needlessly long waits on UI state changes) and fix those
2. Fix the endless waiting tests and "un-ignore" them again

And lastly, a new issue with the Drone integration: Several PRs, e.g. the one here now just say
> continuous-integration/drone/pr _Expected — Waiting for status to be reported_

PR #16696, opened 2026-03-17
fix(oc-file-list-adapter): local file indicator

alperozturk96:


app/src/main/java/com/nextcloud/client/database/dao/FileDao.kt
```diff
@@ -91,20 +91,27 @@ interface FileDao {
...
+        )
+"""
     )
-    suspend fun getSubfiles(
-        parentId: Long,
-        accountName: String,
-        dirType: String = MimeType.DIRECTORY,
-        webdavType: String = MimeType.WEBDAV_FOLDER
-    ): List<FileEntity>
+    fun isFolderFullyDownloaded(parentId: Long, accountName: String): Boolean
```
alperozturk96, line 114: At least one child is exists as file and no child's media_path is null and not directory.

PR #16697, opened 2026-03-17
update agents md

alperozturk96:

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -54,4 +54,4 @@ Avoid creating source files that implement multiple types; instead, place each t
 - Do not use multiple boolean flags to determine states instead use enums or sealed classes.
 - Use modern Java for Java classes. Optionals, virtual threads, records, streams if necessary.
 - Avoid hardcoded strings, colors, dimensions. Use resources.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -50,4 +53,5 @@ Avoid creating source files that implement multiple types; instead, place each t
 - Apply fail fast principle instead of using nested if-else statements.
 - Do not use multiple boolean flags to determine states instead use enums or sealed classes.
 - Use modern Java for Java classes. Optionals, virtual threads, records, streams if necessary.
-
+- Avoid hardcoded strings, colors, dimensions. Use resources.
+- For linting / formatting, use ktlint or detekt for Kotlin, and Android Lint for Java.
```
AndyScherzinger: Works I suppose. You can also just list the gradle tasks to be executed and new issues found to be fixed before committing the changes.

tobiasKaminsky: Yeah, being here more specific might help (at least an human could benefit from it):
```suggestion
- Run lint, spotbugsGplayDebug, detekt, spotlessKotlinCheck and fix findings inside the files that have been changed.
```

AndyScherzinger: The agent will execute it as specified as well 👍

PR #16712, opened 2026-03-20
fix(auto-upload): clean stale upload entities

alperozturk96:
### Issue

When a synced folder is disabled or deleted, its associated `OCUpload` entries in the uploads
table must be cleaned up. Without this, stale upload entries outlive the folder config that
created them, causing `FileUploadWorker` to keep retrying uploads for a folder that no longer
exists or is intentionally turned off, and `AutoUploadWorker` to re-queue already handled files
on its next scan via `FileSystemRepository.getFilePathsWithIds`.

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadHelper.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadHelper.kt
@@ -639,10 +639,30 @@ class FileUploadHelper {
             }
     }
 
-    suspend fun getAutoUploadFolder(files: List<OCFile>, accountName: String): List<SyncedFolderEntity> =
-        files.mapNotNull { file ->
-            getAutoUploadFolderEntity(file, accountName)
+    /**
+     * Splits a list of files into:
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -2172,6 +2172,21 @@ class FileDisplayActivity :
             }
             supportInvalidateOptionsMenu()
             fetchRecommendedFilesIfNeeded(ignoreETag = true, currentDir)
+
+            // clean stale upload entities for auto upload folder
+            if (removedFile.isFolder) {
+                lifecycleScope.launch(Dispatchers.IO) {
+                    val optionalUser = user
```
tobiasKaminsky: if it is an auto upload folder:
- warn user that this will delete configuration and pending uploads
- if yes: remove everything related to that synced folder
- if no: do nothing, also do not delete folder

alperozturk96:
@tobiasKaminsky 

Screenshot_20260326_124941

PR #16744, opened 2026-03-24
feat(sync): download folder including all nested files

### LINKED ISSUE #269: Need option "available offline" for folders.

alperozturk96:
### Issue: https://github.com/nextcloud/android/issues/269
 
Any feedback regarding icons and wording?

@AndyScherzinger  @kra-mo 

<table>
  <tr>
<td>Image 2</td>
    <td>Image 1</td>
    <td>Image 3</td>
  </tr>
</table>

### Demo

[Screen_recording_20260414_100003.webm](https://github.com/user-attachments/assets/7384e9ef-cc49-4f85-a8b8-392440a72c7d)


AndyScherzinger:
Icon looks okay to me, didn't find any thats would be a better option match, maybe @kra-mo has an idea.

Other than that for me the question would be the position in the list. I would put it below the sync item, so it is close in terms of feature. It does however raise the question why there is is sync and sync all.

alperozturk96:
> Icon looks okay to me, didn't find any thats would be a better option match, maybe @kra-mo has an idea.
> 
> Other than that for me the question would be the position in the list. I would put it below the sync item, so it is close in terms of feature. It does however raise the question why there is is sync and sync all.

Sync -> Downloads first hierarchy
Sync all -> Downloads entire folder hiearchy

120e314-git:
> > Icon looks okay to me, didn't find any thats would be a better option match, maybe @kra-mo has an idea.
> > Other than that for me the question would be the position in the list. I would put it below the sync item, so it is close in terms of feature. It does however raise the question why there is is sync and sync all.
> 
> Sync -> Downloads first hierarchy Sync all -> Downloads entire folder hiearchy

Hey, maybe a kinda stupid question: Why you dont name it like "Download entire Folder". It's more specific and the user would know immediatly what this funcion does. 
Also a small, maybe offtopic, question: Wasn't there already a function to download folder earlier? Why was it removed then?

claywd:
> > > Icon looks okay to me, didn't find any thats would be a better option match, maybe @kra-mo has an idea.
> > > Other than that for me the question would be the position in the list. I would put it below the sync item, so it is close in terms of feature. It does however raise the question why there is is sync and sync all.
> > 
> > 
> > Sync -> Downloads first hierarchy Sync all -> Downloads entire folder hiearchy
> 
> Hey, maybe a kinda stupid question: Why you dont name it like "Download entire Folder". It's more specific and the user would know immediatly what this funcion does. Also a small, maybe offtopic, question: Wasn't there already a function to download folder earlier? Why was it removed then?

This could also be a follow up. Personally, I think sync and sync all (which is current nomenclature) makes the most since. Or even "sync recursively" since the sync all options is what is new to the ecosystem here.

app/src/main/res/values/strings.xml
```diff
@@ -326,6 +327,10 @@
     <string name="common_choose_account">Choose account</string>
     <string name="common_switch_account">Switch account</string>
     <string name="common_switch_to_account">Switch to account</string>
+    <string name="sync_all_action_dialog_title">Download entire folder?</string>
+    <string name="sync_all_action_dialog_description">
+    All files and subfolders will be downloaded to your device. This may use significant storage space and take some time depending on your connection.
```
kra-mo: Just a small suggestion :)

```suggestion
    All contained files and folders will be downloaded. This may require a significant amount of storage and take some time depending on your connection.
```

kra-mo (approved):

tobiasKaminsky (dismissed):
- sync and sync all next to each other
- use sync all as "or" in  if (mRemoteFolderChanged || syncAll)  (L142)
- rename function and add documentiation about it
- remove other sync all usages (folder download worker)
- ui: rephrase download to sync in dialog

DanielGlaas:
I want to give feedback about this feature.

I See the New element in the menu 'sync all':
IMG_20260523_230234

I clicked in it and the sync process run very smooth. During the download, the progress information is also nice:
IMG_20260523_230453

What I did not try:
What happens to files that existiert local but not in the Server? Are they uploaded?
Is it detected if these files have been deleted in the Server?

DanielGlaas:
Is a retest necessary?

alperozturk96:
> Is a retest necessary?

Nope thank you for the feedback.

tobiasKaminsky (approved):

PR #16747, opened 2026-03-25
fix(file-delete): local copy

alperozturk96:
### Issue

`removeLocalCopyIfNeeded` was deleting file from local storage however it suppose to delete files under

`/storage/emulated/0/Android/media/com.nextcloud.client/nextcloud/`

not

`/storage/emulated/0/Android/media/Download` or some other local path

### DIFF
--- a/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
+++ b/app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
@@ -935,11 +935,15 @@ public class FileDataStorageManager {
         if (!localPath.equalsIgnoreCase(expectedPath)) {
             Log_OC.w(TAG, "removeLocalCopyIfNeeded: Path mismatch! Expected " + expectedPath
                 + " but found " + localPath + ". Skipping deletion to prevent data loss.");
-            return true;
         }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -913,25 +915,31 @@ public boolean removeFile(OCFile ocFile, boolean removeDBData, boolean removeLoc
...
             return true;
         }
 
-        Log_OC.d(TAG, "removeLocalCopyIfNeeded: deleting local file -> " + localPath);
+        String expectedPath = FileStorageUtils.getDefaultSavePathFor(user.getAccountName(), ocFile);
+        if (!localPath.equalsIgnoreCase(expectedPath)) {
+            Log_OC.w(TAG, "removeLocalCopyIfNeeded: Path mismatch! Expected " + expectedPath
+                + " but found " + localPath + ". Skipping deletion to prevent data loss.");
+            return true;
+        }
```
tobiasKaminsky, line 938: if localpath is wrong, then try to delete via expectedPath.
Please also log that it is a wrong local path, so that we can try to track root cause down.

Issue #591, opened 2017-01-27
User info: show groups


tobiasKaminsky:
#554 by @mario adds the user info screen which is really cool!
As an addition I think it is nice to also see the groups in which the user is part of.
This is also displayed on the web-UI and so the app would be even more consistent to it.

mario:
@schiessle could you please add a list of groups to both endpoints? Also, please a backport to 11.0.2 is more than welcome. Once done, please gimme the object name so I can add support for this to the library.

Thanks!

schiessle:
see: https://github.com/nextcloud/server/pull/3418

AndyScherzinger:
Support is server 11.0.3+
This might need another PR done for the library project.

AndyScherzinger:
@jancborchardt @nextcloud/designers here are two mock ups for integrating the group info the the profile view:

Variant 1
![device-2017-10-06-112015](https://user-images.githubusercontent.com/1315170/31271951-7afdba28-aa89-11e7-9d21-555870e6520c.png)

Variant 2:
![device-2017-10-06-112645](https://user-images.githubusercontent.com/1315170/31271962-81b76ba2-aa89-11e7-9ad4-525a6e1043a1.png)

I'd vote for variant 1 since to my personal data / profile is something different than memberships in groups thus I'd separate it visually.

jancborchardt:
@AndyScherzinger agree with the separation. However can we add the »Groups« icon so the content text aligns? Also, the »Profile« and »Groups« text should be left-aligned on the same line as the content and heading text (moved to the right a bit).

tobiasKaminsky:
Groups should be sorted by name, or?

joshtrichards:
> This might need another PR done for the library project.

Added to `android-library` in nextcloud/android-library#268

joshtrichards:
Related (not yet merged and quite old) PR: #3781

tobiasKaminsky:
Update:
- info is in UserInfo on Android-Library
- use Version 1
- sort groups via name
- split view into Profile and Groups

penguin86:
Hello, I implemented this in PR https://github.com/nextcloud/android/pull/16767
Rationale and screenshots in the PR description.
Should I change something, just let me know.

Have a nice day :)

penguin86:
Update: I reimplemented the feature with a better approach as suggested in https://github.com/nextcloud/android/pull/16767#issuecomment-4148506322 
[...]

PR #16767, opened 2026-03-28
#591 Groups in user info

penguin86:
### 🖼️ Screenshots

Implementation of https://github.com/nextcloud/android/issues/591

Notes: the previous implementation used RecyclerView for the list items. That's a sleek implementation and it's very efficient for long lists due to the efficiency in instantiating only the views needed to cover the screen and recycling them when the user scrolls. But in this case, the list barely consists of 6 elements (as I see, all the groups are concatenated in a single item), and in the vast majority of devices they will be all instantiated and drawn anyway, so it doesn't make a real difference.

Obtaining the required result (see "variant 1" in https://github.com/nextcloud/android/issues/591#issuecomment-334705913) with the two "groups" of items while keeping the current structure is possible (playing with cell backgrounds and different row views for the headers), but in my opinion unnecessarily complex, and less maintainable.

I decided to remove the RecyclerView and opt for a simpler implementation, that should be more readable.
Let me know if there is something that could have been done differently and I'll fix it.

🏚️ Before | 🏡 After
---|---
before|after

### 🏁 Checklist
 
### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/UserInfoActivity.java
+++ b/app/src/main/java/com/owncloud/android/ui/activity/UserInfoActivity.java
@@ -53,6 +53,8 @@ import com.owncloud.android.utils.theme.ViewThemeUtils;
 import org.greenrobot.eventbus.Subscribe;
 import org.greenrobot.eventbus.ThreadMode;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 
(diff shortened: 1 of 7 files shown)


AndyScherzinger:
Hi @penguin86 I'd be fine with the approach for the sake of simplicity for this PR. However in the long(er) run the recycler view would need to be brought back or be a LazyList when migrating to composables I's day or when adding further fields. Because in the long run we would want to show further/all profile fields, like the Talk app does now, see https://github.com/nextcloud/talk-android/pull/5936 - so the simple approach could else have been to go with the non-separated variant and just add another items for the groups or if one would be two implement the two sections to go with a sectioned Recycler view adapter, see for example https://github.com/nextcloud/android/blob/master/app/src/main/java/com/owncloud/android/ui/fragment/contactsbackup/BackupListAdapter.kt#L82-L114 implementing this for the calendar and contacts backup list.

Not saying you'd have to do any of the above mentioned thoughts - just to discuss and elaborate on the approach - also given you can't anticipate we would want to add further fields in the future to mirror what the Talk client can and does display.

CC @tobiasKaminsky @alperozturk96

penguin86:
Hello @AndyScherzinger ,
thanks for the clarification! I was missing the overall view, indeed.
Now it makes sense to me why the original contributor has decided for a RecyclerView.
Thanks to your hint, I was able to come with a better implementation, based on the sectioned RecyclerView (that's very convenient!).
Being it a complete rewrite, in the sake of cleanliness, I prefer to close this PR and open a new one. I'll link the new one in the issue.

PR #16769, opened 2026-03-29
591: Add groups to user info, with sections and using RecyclerView for future expandability

penguin86:
Implementation of https://github.com/nextcloud/android/issues/591

Added groups and implemented sectioning in RecyclerView. Meant for adding further groups if needed in the future.
Based on the suggestions by Andy in https://github.com/nextcloud/android/pull/16767#issuecomment-4148506322

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
before | after_new

### 🏁 Checklist
 

app/src/main/java/com/owncloud/android/ui/adapter/UserInfoAdapter.kt
```diff
@@ -0,0 +1,112 @@
...
+import androidx.annotation.DrawableRes
+import com.afollestad.sectionedrecyclerview.SectionedRecyclerViewAdapter
+import com.afollestad.sectionedrecyclerview.SectionedViewHolder
+import com.nextcloud.android.common.ui.theme.utils.ColorRole
+import com.owncloud.android.R
+import com.owncloud.android.databinding.UserInfoDetailsTableItemBinding
+import com.owncloud.android.databinding.UserInfoDetailsTableItemTitleBinding
+import com.owncloud.android.utils.theme.ViewThemeUtils
+
+class UserInfoAdapter(val context: Context, val mDisplayList: Map<Int, MutableList<UserInfoDetailsItem>>, val viewThemeUtils: ViewThemeUtils) :
```
alperozturk96: Why `mDisplayList` needs to have mutable list? Adapter should just display the list.

Also since this is a new class; `mDisplayList` can we rename it to the `displayList`?

app/src/main/java/com/owncloud/android/ui/adapter/UserInfoAdapter.kt
```diff
@@ -0,0 +1,112 @@
...
+    SectionedRecyclerViewAdapter<SectionedViewHolder>() {
+    companion object {
+        const val SECTION_USERINFO = 0
+        const val SECTION_GROUPS = 1
+    }
+
+    class UserInfoDetailsItem(
```
alperozturk96: `UserInfoDetailsItem` this can be a data class. For item of the adapter better (so that we can compare easily if necessary) to use data class but it's a small thing.

app/src/main/java/com/owncloud/android/ui/adapter/UserInfoAdapter.kt
```diff
@@ -0,0 +1,112 @@
...
+            UserInfoDetailsTableItemBinding.inflate(
+                LayoutInflater.from(parent.context),
+                parent,
+                false
+            ),
+        )
+    }
+
+    override fun onBindViewHolder(
+        holder: SectionedViewHolder?,
```
alperozturk96: `SectionedViewHolder?` we can make this non-nullable.

app/src/main/java/com/owncloud/android/ui/adapter/UserInfoAdapter.kt
```diff
@@ -0,0 +1,112 @@
...
+        section: Int,
+        relativePosition: Int,
+        absolutePosition: Int
+    ) {
+        val item = mDisplayList[section]?.get(relativePosition)
```
alperozturk96: I prefer the extract item position logic to internal enum class.

```kotlin
private enum class ItemPosition {
  SINGLE, FIRST, MIDDLE, LAST;

  fun backgroundRes(): Int = when (this) {
    SINGLE -> R.drawable.rounded_corners_listitem_single_background
    FIRST -> R.drawable.rounded_corners_listitem_first_background
    LAST -> R.drawable.rounded_corners_listitem_last_background
    MIDDLE -> R.drawable.rounded_corners_listitem_center_background
  }
}

private fun resolvePosition(position: Int, count: Int): ItemPosition {
  val isFirst = (position == 0)
  val isLast = (position == count - 1)

  return when {
    isFirst && isLast -> ItemPosition.SINGLE
    isFirst -> ItemPosition.FIRST
    isLast -> ItemPosition.LAST
    else -> ItemPosition.MIDDLE
  }
}
```

So that we can have simpler `onBindViewHolder`

**Usage:**

```kotlin
val count = getItemCount(section)
val positionType = resolvePosition(relativePosition, count)
val uiHolder = holder as UserInfoSectionedViewHolder
val item = mDisplayList[section]?.get(relativePosition)

// Set background
uiHolder.binding.root.setBackgroundResource(
  positionType.backgroundRes()
)
```

alperozturk96 (changes requested):
Hello

Thank you for the PR.

jancborchardt (commented):
Very nice! :)

This is probably a follow-up and should be done separately, but it would be nice to also move it to be in line with how @AndyScherzinger did it for the Talk Android app: https://github.com/nextcloud/talk-android/pull/5943

app/src/main/java/com/owncloud/android/ui/adapter/UserInfoAdapter.kt
penguin86: Hey @alperozturk96 , thank you for spotting the first 3 and thank you for the very sleek item type refactoring!
I committed them! Can I add your name on the contributor list in the adapter?

penguin86:
Hey @jancborchardt , yes I think it's the final intent: in fact, this isn't my first implementation unfortunately, at first I completely missed the big picture 🤦‍♂️ : see comment https://github.com/nextcloud/android/pull/16767#issuecomment-4148506322 in the, now closed, original PR.

app/src/main/java/com/owncloud/android/ui/adapter/UserInfoAdapter.kt
alperozturk96: Yes sure :)

alperozturk96 (approved):

PR #16819, opened 2026-04-09
fix(upload-list): handle conflict actions

### LINKED ISSUE #16049: Auto upload conflict detection not working correctly
### ⚠️ Before posting ⚠️

### Steps to reproduce

1. Enable auto upload on any folder
2. Set to "Skip uploading" if file exists
3. Get notification "Upload failed Conflict" even though the file does not exist

Or

2. Set to "Rename new version" if file exists
3. File gets renamed indefinitely ("$file", "$file (2)", "$file (2) (2)", ...)

### Expected behaviour

The app should either detect that the file does not exist and upload it or detect that the file already exists and do nothing if set to "Skip uploading". If set to "Rename new version" it shoul detect, that the file already exists and rename it once, then stop.

### Actual behaviour

The app doesn't seem to correctly detect if a file conflict occurs. If the file does not exist it still detects a non-existing conflict.

### Android version

16

### Device brand and model

Google Pixel 7a

### Stock or custom OS?

Custom (explain in "additional information")

3.34.1

### Nextcloud server version

31.0.10

### Using a reverse proxy?

Yes

### Android logs

[Nextcloud log afc001edf3d7.txt](https://github.com/user-attachments/files/23904687/Nextcloud.log.afc001edf3d7.txt)

### Server error logs

```bash

```

### Additional information

Using GrapheneOS Build 2025112101.

### LINKED ISSUE #16818: Auto-upload conflicts, yet again -- this is a never ending problem
### ⚠️ Before posting ⚠️

### Steps to reproduce

1. Configure your device's photo storage for auto-upload
2. Take some photos
3. Get a notification about conflicts
4. Click notification to see conflicts
5. Click one that says "Sync conflict, please resolve manually"
6. Wait a few seconds

### Expected behaviour

File should be uploaded to the server.

### Actual behaviour

Toast message saying:

> File not found.  Are you sure that this file exists or has a previous conflict not been resolved?

### Android version

16

### Device brand and model

Pixel 8 Pro

### Stock or custom OS?

Stock

33.0.1

### Nextcloud server version

31.0.9

### Using a reverse proxy?

No

### Android logs

_No response_

### Server error logs

```bash
Nothing
```

### Additional information

I'm not even sure why the Android app is saying there is a conflict as I have checked the server and the file reporting the conflict does **NOT** even exist on the server.

The Android device in question is not kitted to get debug info.  It is a "production" device for a user and _Developer options_ is not an appropriate security model for this device.

Ultimately this seems to be just yet another (of the many) bug(s) with auto-upload.  I wish this feature could _just work_ without constantly being a bug-ridden steaming pile.

alperozturk96:
### Issue

 On Android, we are showing conflicts based on the stored last upload result, but this is wrong because the server can change at any moment. As a result, we may display an incorrect status, and the user might try to open the conflict resolution dialog even though it’s unnecessary, perhaps the file no longer exists on the remote server or has been moved.

### Flow

First check remote path from `OCUpload` via `ReadFileOperation` -> if file not there remove conflict error and allow user to retry upload

If file exists on remote and same ->  if file not there remove conflict error and allow user to retry upload

If file exists and different -> show conflict resolve dialog

### Changes

- Extracts `isSame` logic since its necessary for this PR as well.
- Implements `UploadListAdapterHelper`, `UploadListAdapterAction` since needed for handling conflicts and not bloating the adapter logic. Also we need to implement other actions as well in future.
- Adds correct messages that reflects current state of the upload and action.
- Adds `UploadFileOperationFactory` since huge constructor needs to be repeated for each file

### Demo - if conflict exists

[Screen_recording_20260409_150419.webm](https://github.com/user-attachments/assets/66d0dcfd-bd23-48c7-a60b-723a63ef295a)

### Demo - if conflict not exists anymore (e.g. file moved or removed from server)

[Screen_recording_20260409_150442.webm](https://github.com/user-attachments/assets/d8f5e7d4-d377-46a8-aac3-7c682cde4f71)

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/uploadList/UploadListAdapter.kt
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/uploadList/UploadListAdapter.kt
@@ -10,7 +10,6 @@ package com.owncloud.android.ui.adapter.uploadList
 import android.annotation.SuppressLint
 import android.app.NotificationManager
 import android.content.Context
-import android.os.Looper
 import android.text.format.DateUtils
 import android.view.LayoutInflater
 import android.view.View
@@ -669,43 +668,6 @@ class UploadListAdapter(
         }
 
         notificationManager?.cancel(upload.uploadId.toInt())
-    }
-
-    fun notifyUploadChanged(upload: OCUpload) {
-        for (sectionIndex in uploadListSections.indices) {
-            val section = uploadListSections[sectionIndex]
-
-            val itemIndex = section.items.indexOfFirst { it.uploadId == upload.uploadId }
-
-            if (itemIndex != -1) {
-                val adapterPosition = getAdapterPosition(sectionIndex, itemIndex)
-
-                if (adapterPosition != -1) {
-                    val updatedItems = section.items.toMutableList().apply {
-                        this[itemIndex] = upload
-                    }
-
-                    uploadListSections[sectionIndex] = section.withItems(updatedItems)
-
-                    if (Looper.myLooper() == Looper.getMainLooper()) {
-                        notifyItemChanged(adapterPosition)
-                    } else {
-                        activity.runOnUiThread {
-                            notifyItemChanged(adapterPosition)
-                        }
-                    }
-                }
-                return
-            }
-        }
-    }
-
-    private fun getAdapterPosition(section: Int, relativePosition: Int): Int {
-        var position = 0
-        for (i in 0 until section) {
-            position += uploadListSections[i].items.size + 1
-        }
-        return position + 1 + relativePosition
     }
 
     companion object {
--- a/app/src/main/java/com/owncloud/android/ui/adapter/uploadList/UploadListAdapter.kt
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/uploadList/UploadListAdapter.kt
@@ -10,6 +10,7 @@ package com.owncloud.android.ui.adapter.uploadList
 import android.annotation.SuppressLint
 import android.app.NotificationManager
 import android.content.Context
+import android.os.Looper
 import android.text.format.DateUtils
 import android.view.LayoutInflater
 import android.view.View
@@ -21,7 +22,6 @@ import com.afollestad.sectionedrecyclerview.SectionedRecyclerViewAdapter
 import com.afollestad.sectionedrecyclerview.SectionedViewHolder
 import com.nextcloud.client.account.User
 import com.nextcloud.client.account.UserAccountManager
-import com.nextcloud.client.core.Clock
 import com.nextcloud.client.device.PowerManagementService
 import com.nextcloud.client.jobs.upload.FileUploadHelper
 import com.nextcloud.client.jobs.upload.FileUploadWorker
@@ -40,13 +40,13 @@ import com.owncloud.android.datamodel.ThumbnailsCacheManager
 import com.owncloud.android.datamodel.UploadsStorageManager
 import com.owncloud.android.db.OCUpload
 import com.owncloud.android.db.UploadResult
-import com.owncloud.android.lib.common.operations.OnRemoteOperationListener
-import com.owncloud.android.lib.common.operations.RemoteOperation
-import com.owncloud.android.lib.common.operations.RemoteOperationResult
 import com.owncloud.android.lib.common.utils.Log_OC
-import com.owncloud.android.operations.RefreshFolderOperation
 import com.owncloud.android.ui.activity.FileActivity
 import com.owncloud.android.ui.adapter.progressListener.UploadProgressListener
+import com.owncloud.android.ui.adapter.uploadList.helper.UploadListAdapterHelper
+import com.owncloud.android.ui.adapter.uploadList.helper.UploadListItemOnClick
+import com.owncloud.android.ui.adapter.uploadList.model.UploadListSection
+import com.owncloud.android.ui.adapter.uploadList.model.UploadListType
 import com.owncloud.android.utils.DisplayUtils
 import com.owncloud.android.utils.MimeTypeUtil
 import com.owncloud.android.utils.theme.ViewThemeUtils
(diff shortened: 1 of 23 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1095,7 +1094,10 @@
     <string name="upload_lock_failed">Locking folder failed</string>
     <string name="upload_local_storage_full">Local storage full</string>
     <string name="upload_old_android">Encryption is only possible with &gt;= Android 5.0</string>
-    <string name="upload_sync_conflict">Sync conflict, please resolve manually</string>
+    <string name="upload_sync_conflict_checking">Checking for upload conflicts…</string>
```
alperozturk96, line 1099: @kra-mo Please share your feedback regarding wording :)

kra-mo, line 1099: Do we even need this string? How long can this operation take? If less then a second, I'd say we can skip it.

app/src/main/res/values/strings.xml
```diff
@@ -1095,7 +1094,10 @@
...
+    <string name="upload_sync_conflict_not_exists">Upload has no conflicts anymore, conflict cleared.</string>
```
kra-mo: Could be a bit simpler :)

```suggestion
    <string name="upload_sync_conflict_not_exists">Conflict resolved.</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -1095,7 +1094,10 @@
...
+    <string name="upload_sync_conflict_check_error">Could not check for upload conflicts. Please try again.</string>
```
kra-mo, line 1102: The user does not explicitly choose to check for conflicts so it would be nice if this was some more generic message, not related to conflicts specifically.

kra-mo (dismissed):

app/src/main/res/values/strings.xml
alperozturk96: It depends(imagine slow instance or instance that time is very busy or slow internet connection etc) ... So this can take more than a second. If it's too fast user will not going to see it because we are updating the same snackbar with result of the conflict resolution check. So I would say this is okay to keep.

app/src/main/res/values/strings.xml
alperozturk96: On the upload list screen, users will see these messages only when the last upload failed and has a conflict.

When upload has conflict (this will appear as sub-text before it was "Sync conflict, please resolve manually" which was wrong and confusing and we got a lot of issues/feedback about it):

```
<string name="upload_sync_conflict_check">There might be a conflict from your last upload. Tap to check.
</string> 
```

If user taps the item 

`<string name="upload_sync_conflict_checking">Checking for upload conflicts…</string> `

And after completion the checking process user can see two message: 

`<string name="upload_sync_conflict_not_exists">Conflict resolved.</string> Retry action will appear in the snackbar to retry previously failed upload `

or 

`<string name="upload_sync_conflict_check_error">Could not check for upload conflicts. Please try again.</string> `

I hope this flow is clear. In this case the user does explicitly choose to check for conflicts. Thus we must give correct and user-friendly message.

app/src/main/res/values/strings.xml
alperozturk96: Yes could be but this `has no conflicts anymore` indication refers the previous state thus they know the state of the upload is changed via text as well.

But also they see it from the UI so `Conflict resolved.` can be used as well and its shorter.

app/src/main/res/values/strings.xml
kra-mo: Okay, fair enough ^^

app/src/main/res/values/strings.xml
kra-mo: I see, I see. Okay, makes sense.

kra-mo (approved):

app/src/main/java/com/owncloud/android/ui/adapter/uploadList/UploadListAdapter.kt
```diff
@@ -698,25 +615,6 @@ class UploadListAdapter(
         loadUploadItemsFromDb()
     }
 
-    private fun refreshFolder(view: ItemViewHolder, user: User?, folder: OCFile?, listener: OnRemoteOperationListener) {
-        view.binding.uploadListItemLayout.isClickable = false
-        view.binding.uploadStatus.setText(R.string.uploads_view_upload_status_fetching_server_version)
-        RefreshFolderOperation(
```
alperozturk96, line 704: @tobiasKaminsky  `RefreshFolderOperation` Why this operation is used? We can simple just use `ReadFileOperation` for given remote path and compare, no?

app/src/main/java/com/owncloud/android/ui/adapter/uploadList/helper/UploadListAdapterActionHandler.kt
```diff
@@ -0,0 +1,71 @@
...
+    companion object {
+        private const val TAG = "UploadListAdapterActionHandler"
+    }
+
+    override suspend fun handleConflict(
+        upload: OCUpload,
+        client: OwnCloudClient,
+        storageManager: UploadsStorageManager
+    ): ConflictHandlingResult = withContext(Dispatchers.IO) {
+        val operationResult = ReadFileRemoteOperation(upload.remotePath).execute(client)
```
tobiasKaminsky, line 33: read file -> if exist -> continue show conflict
if file does not exist -> check if parent folder exists -> if exist -> simply upload
if parent folder does not exist -> show error message

app/src/main/java/com/owncloud/android/ui/adapter/uploadList/helper/UploadListAdapterActionHandler.kt
```diff
@@ -0,0 +1,71 @@
...
+
+        if (!operationResult.isSuccess) {
```
tobiasKaminsky, line 35: Do not ask for "retry upload" but simply re-upload conflicted file if possible (when user clicks on it).

tobiasKaminsky (dismissed):

PR #16848, opened 2026-04-14
fix(file-list): browse up

alperozturk96:
### Issue

Very rare app cannot find parent dir

### Changes

Uses file id first if fails uses path
Covers edge cases
Adds tests

### Demo

[Screen_recording_20260414_154306.webm](https://github.com/user-attachments/assets/bc5c75a3-a136-4bcc-8bb6-fe18e92a836b)


tobiasKaminsky (approved):
please also add unit test
maybe it can be reproduced via copy/move if previously folder was a/b/c, but now c does not exist anylonger.

alperozturk96:
**browse up if parent deleted instead browse to root**

[Screen_recording_20260415_143047.webm](https://github.com/user-attachments/assets/7586e40a-8ed2-4e12-a389-b183e834eb57)

PR #16855, opened 2026-04-15
Use built-in Kotlin

tobiasKaminsky:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist
 

app/build.gradle.kts
```diff
@@ -167,12 +165,6 @@ android {
         }
     }
 
-    applicationVariants.configureEach {
-        outputs.configureEach {
-            if (this is ApkVariantOutputImpl) this.outputFileName = "${this.baseName}-${this.versionCode}.apk"
```
tobiasKaminsky, line 172: I think we do not need it.
Since there was no easy migration, I simply removed it.

PR #16865, opened 2026-04-16
feat(unified-share): ui

alperozturk96:
Common Library PR: https://github.com/nextcloud/android-common/pull/939


daniele-verducci:
# Testing results

## Share list

- When a share is added and then the list is reloaded (i.e. with pull-to-refresh), the list is rearranged in a different order
- On small screens, it's impossible to see the list in landscape, and the view can't be scrolled if at least some of the list isn't visible (to reproduce: Pixel 10 emulator, Android 17, default settings):
image

- In landscape, the date dialog isn't useable, the calendar inside it doesn't scroll and it's cut  (to reproduce: Pixel 10 emulator, Android 17, default settings):
image

- If the last item in the list is a share with multiple recipients, there's no way to expand it to change single recipients settings, as the FAB is over the expansion button:
Screenshot_20260907-160331520


## Invited people (Internal)

- It is possible to create two identical shares (same person, same permissions, same settings)


## Anyone (External)

- Links created with Anyone are not working, if opened show "Share not found" (probably server-side issue)

daniele-verducci (approved):
Hey @alperozturk96 ,
done a fair amount of testing and found nothing really blocking (see the previous comment). Thus I approve, but if you should decide to fix a single item in the previous list, I'd suggest it to be the FAB covering the last row expansion button. Maybe it's enough to just add some margin to the last row...

daniele-verducci (approved):
Hey Alper,
all fixed but the non-scrollable-on-landscape issue, but that's common to other PRs (i.e. the new tags management PR) and not directly related to this feature, so I think we can definitely merge!
Thank you!

PR #16903, opened 2026-04-22
feat(encryption): create sub encryption folder

alperozturk96:
User can create sub encrypted folder in normal folder just like web

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/fragment/OCFileListBottomSheetDialog.kt
+++ b/app/src/main/java/com/owncloud/android/ui/fragment/OCFileListBottomSheetDialog.kt
@@ -93,7 +93,9 @@ class OCFileListBottomSheetDialog(
     }
 
     private fun checkCreateEncryptedFolderVisibility() {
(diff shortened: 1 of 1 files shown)


tobiasKaminsky (approved):

app/src/main/java/com/owncloud/android/ui/fragment/OCFileListBottomSheetDialog.kt
```diff
@@ -93,7 +93,7 @@ class OCFileListBottomSheetDialog(
     }
 
     private fun checkCreateEncryptedFolderVisibility() {
-        binding.menuEncryptedMkdir.setVisibleIf(file.isRootDirectory)
+        binding.menuEncryptedMkdir.setVisibleIf(!file.isEncrypted)
```
tobiasKaminsky: Please check first if e2e capibility exists and is enabled

tobiasKaminsky (changes requested):

PR #16905, opened 2026-04-23
improve end-to-end encryption verifications

alperozturk96:

### DIFF
--- a/app/src/main/java/com/owncloud/android/utils/EncryptionUtilsV2.kt
+++ b/app/src/main/java/com/owncloud/android/utils/EncryptionUtilsV2.kt
@@ -150,6 +150,7 @@ class EncryptionUtilsV2 {
         client: OwnCloudClient,
         oldCounter: Long,
         signature: String,
+        rawMetadata: String,
         user: User,
         context: Context,
         arbitraryDataProvider: ArbitraryDataProvider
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/owncloud/android/utils/EncryptionUtilsV2.kt
```diff
@@ -978,31 +876,34 @@ class EncryptionUtilsV2 {
...
-        return certs.any { cert ->
-            runCatching {
-                signer.verify(verifierBuilder.build(cert.publicKey))
-            }.getOrElse {
-                Log_OC.e(TAG, "Exception verifySignedData: $it")
-                false
+    fun verifySignedData(folderMetadata: String, privateKey: String, certs: List<X509Certificate>): Boolean {
+        for (cert in certs) {
+            try {
+                val privateKey = EncryptionUtils.PEMtoPrivateKey(privateKey)
```
tobiasKaminsky: Private key can only be the user's private key, since we do not have the other sharee's private key.

But this then means that we create new signed message with our own private key and check this against the public key.

What we really need to verify the existing signed data from metadata.json:
- User A shares encrypted folder to User B
- User A signs the data with their private key
- User B can verify the signature (stored in metadata.json) via pubic key from User A

tobiasKaminsky (changes requested):

alperozturk96:
replaced with https://github.com/nextcloud/android/pull/16916

PR #16908, opened 2026-04-24
Improve GallerySearchTask 

alperozturk96:
- Uses Coroutines instead of deprecated AsyncTask
- Converts to Kotlin


app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.kt
```diff
@@ -319,18 +319,17 @@ class GalleryFragment :
...
             limit
-        )
-            .execute()
+        ).execute()
     }
 
     private fun loadMoreWhenEndReached(recyclerView: RecyclerView, dy: Int) {
-        if (dy > 0 && !isPhotoSearchQueryRunning) {
-            // scrolled vertical space not bigger than 0 or still searching
+        if (dy <= 0 || isPhotoSearchQueryRunning) {
```
alperozturk96, line 326: @tobiasKaminsky dy is positive if user scrolls down, if user scrolls up it is negative. Thus condition is wrong.

If user scroll up never try to search again because already done or query is active skip search again as well.

app/src/main/java/com/owncloud/android/ui/fragment/GalleryFragment.kt
```diff
@@ -319,18 +319,17 @@ class GalleryFragment :
...
+            Log_OC.d(TAG, "scrolling up or search query already active, do not search gallery")
```
tobiasKaminsky:
```suggestion
// scrolling up or search query already active, do not search gallery
```

tobiasKaminsky (approved):

PR #16916, opened 2026-04-27
fix(end-to-encryption): verify metadata

alperozturk96:
Uses same verification logic just like iOS, Desktop
Adds tests


app/src/main/cpp/cms_verifier.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+extern "C" JNIEXPORT jboolean JNICALL
+Java_com_nextcloud_utils_CmsSignatureVerifier_verifySignedData(
+    JNIEnv* env,
+    jobject /* thiz */,
+    jbyteArray cmsDataArray,
+    jbyteArray messageDataArray,
+    jobjectArray certPemArray
+) {
+    jsize cmsLen = env->GetArrayLength(cmsDataArray);
```
alperozturk96, line 27: @marinofaggiana @mgallien 

Used for https://github.com/nextcloud/end_to_end_encryption_rfc/blob/e185645eb5fd0849f49e308f006b194772cc3002/RFC.md#verifying-the-metadata 

due to `BouncyCastle's` issue with `BER` encoded `CMSSignedData`.

Uses iOS logic.

ndk.env
```diff
@@ -1,2 +1,2 @@
-NDK_VERSION=21.4.7075529
-CMAKE_VERSION=3.18.1
+NDK_VERSION=27.2.12479018
```
alperozturk96: @tobiasKaminsky  fyi

gradle/libs.versions.toml
```diff
@@ -65,6 +65,7 @@ mockitoVersion = "5.22.0"
 mockkVersion = "1.14.9"
 nnioVersion = "0.3.1"
 objenesis = "3.5"
+opensslVersion = "1.1.1q-beta-1"
```
tobiasKaminsky: beta1?
Is a stable version also working?

tobiasKaminsky (dismissed):

gradle/libs.versions.toml
alperozturk96: There is no stable version of it https://mvnrepository.com/artifact/com.android.ndk.thirdparty/openssl

tobiasKaminsky (approved):

PR #16942, opened 2026-04-29
fix gh

alperozturk96:

### DIFF
--- a/app/lint.xml
+++ b/app/lint.xml
@@ -25,6 +25,7 @@
             regexp="screenshot_01_gridView|screenshot_02_listView|screenshot_03_drawer|screenshot_04_accounts|screenshot_05_autoUpload|screenshot_06_davdroid" />
         <ignore path="**/drawable/zdc_*.xml" />
 		<ignore path="**/drawable/iconclose.xml" />
+        <ignore regexp="trashbin_activity_title" />
     </issue>
 
     <issue id="ImpliedQuantity">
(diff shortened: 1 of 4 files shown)


app/lint.xml
```diff
@@ -89,4 +90,8 @@
 	<issue id="Aligned16KB">
         <ignore path="**/arm64-v8a/libmockkjvmtiagent.so" />
     </issue>
-</lint>
+
+    <issue id="OldTargetApi">
```
tobiasKaminsky, line 94: I created another PR to exclude that issue entirely, not only on build.gradle.kts.
I think it is better, since otherwise Lint will also complain for appscan module.

app/lint.xml
```diff
@@ -25,6 +25,7 @@
             regexp="screenshot_01_gridView|screenshot_02_listView|screenshot_03_drawer|screenshot_04_accounts|screenshot_05_autoUpload|screenshot_06_davdroid" />
         <ignore path="**/drawable/zdc_*.xml" />
 		<ignore path="**/drawable/iconclose.xml" />
+        <ignore regexp="trashbin_activity_title" />
```
tobiasKaminsky, line 28: Why is this needed? 
Is it really a false positive check?

tobiasKaminsky (changes requested):

app/lint.xml
alperozturk96: im just testing its not ready for review. probably happened after fragment conversion

PR #16988, opened 2026-05-04
handle(auto-upload): LOCKED

alperozturk96:
### Issue

Server may be busy with previous upload attempt thus next-attempt will throw `LOCKED 423` and user will see wrong information about this such as `Upload failed` however this can be recovered and can be temporary error.

### Changes

- Show correct message instead of error
- Handle LOCKED status and add delay for next upload attempt
- Only show one LOCKED error messages thus this can be same issue for the next upload attempts
- Fixes clear action for `SKIPPED` uploads

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/upload/UploadDelayPolicy.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/upload/UploadDelayPolicy.kt
@@ -7,24 +7,25 @@
 
 package com.nextcloud.client.jobs.upload
 
-import com.nextcloud.utils.TimeConstants
 import kotlin.random.Random
 
 @Suppress("MagicNumber")
-class UploadRetryPolicy {
+class UploadDelayPolicy {
     private var delayInMs: Long = 0
 
     companion object {
+        private const val MAX_DELAY = 120_000L
+        private const val INCREMENT_VALUE = 3500L
         private const val MAX_RANDOM_DELAY = 200L
     }
 
     fun increase() {
-        if (delayInMs >= TimeConstants.ONE_SECOND.times(10)) {
+        if (delayInMs >= MAX_DELAY) {
             return
         }
 
         // random next long used for prevent retrying at the same time if uploads are in parallel
-        delayInMs += (TimeConstants.ONE_SECOND + Random.nextLong(MAX_RANDOM_DELAY))
+        delayInMs += (INCREMENT_VALUE + Random.nextLong(MAX_RANDOM_DELAY))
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/nextcloud/client/jobs/upload/UploadRetryPolicy.kt
```diff
@@ -0,0 +1,35 @@
...
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+package com.nextcloud.client.jobs.upload
+
+import com.nextcloud.utils.TimeConstants
+import kotlin.random.Random
+
+@Suppress("MagicNumber")
+class UploadRetryPolicy {
```
alperozturk96: Change name of it, it's delay policy

app/src/main/res/values/strings.xml
```diff
@@ -1200,7 +1200,8 @@
     <string name="upload_list_delete">Delete</string>
     <string name="create_new">New</string>
     <string name="editor_placeholder" translatable="false">%1$s %2$s</string>
-
+    <string name="upload_locked_title">Upload in progress</string>
+    <string name="upload_locked_message">The server is still working on a previous upload. Retrying…</string>
```
kra-mo: Does knowing this help the user? I feel like it's better not to overload them with technical information. It's fine to just say that the server is busy :)

```suggestion
    <string name="upload_locked_message">The server is busy. Retrying…</string>
```

kra-mo (dismissed):

app/src/main/res/values/strings.xml
alperozturk96: It is good to show message for `LOCKED` case; instead saying it failed which is incorrect and misleading. The server is busy is enough.

app/src/main/java/com/nextcloud/client/jobs/upload/UploadDelayPolicy.kt
```diff
@@ -0,0 +1,36 @@
...
+    private var delayInMs: Long = 0
+
+    companion object {
+        private const val MAX_DELAY = 120_000L
+        private const val INCREMENT_VALUE = 3500L
+        private const val MAX_RANDOM_DELAY = 200L
+    }
+
+    fun increase() {
+        if (delayInMs >= MAX_DELAY) {
```
alperozturk96, line 23: Lock should only take a couple of seconds, maximum.

tobiasKaminsky (approved):

PR #16998, opened 2026-05-06
Whiteboard on Mobile

### LINKED ISSUE #14605: Whiteboard in app
I think having the ability to edit and create whiteboards within the Android mobile app like you can with office files would be very helpful.

tobiasKaminsky:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist
 
### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/activity/WhiteboardWebView.kt
+++ b/app/src/main/java/com/owncloud/android/ui/activity/WhiteboardWebView.kt
@@ -1,0 +1,227 @@
+/*
+ * Nextcloud - Android Client
+ *
+ * SPDX-FileCopyrightText: 2019 Chris Narkiewicz <hello@ezaquarii.com>
+ * SPDX-FileCopyrightText: 2018 Tobias Kaminsky <tobias@kaminsky.me>
+ * SPDX-FileCopyrightText: 2018 Nextcloud GmbH
+ * SPDX-License-Identifier: AGPL-3.0-or-later OR GPL-2.0-only
+ */
+package com.owncloud.android.ui.activity
+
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/owncloud/android/ui/helpers/FileOperationsHelper.java
```diff
@@ -355,6 +356,14 @@ public void openFile(OCFile file) {
             });
         }).start();
     }
+    
+    public void openFileAsWhiteboard(OCFile file, Context context) {
+        Intent collaboraWebViewIntent = new Intent(context, WhiteboardWebView.class);
```
blizzz, line 361: naming? 😁

app/src/main/java/com/owncloud/android/ui/activity/WhiteboardWebView.kt
```diff
@@ -0,0 +1,227 @@
...
+
+    override fun onRestoreInstanceState(savedInstanceState: Bundle) {
+        url = savedInstanceState.getString(EXTRA_URL)
+        super.onRestoreInstanceState(savedInstanceState)
+    }
+
+    override fun onResume() {
+        super.onResume()
+        webView.evaluateJavascript(
+            "if (typeof OCA.RichDocuments.documentsMain.postGrabFocus !== 'undefined') " +
```
blizzz, line 119: 🤔 

P.S.: OK, got it, essentially an mostly unadjusted copy of the richdocs view at this state 🙂

tobiasKaminsky:
Actually it seems we do not need it, but can only rely on generic direct editing implementation :tada:

PR #17018, opened 2026-05-08
fix(oc-file-list): favorite refresh

alperozturk96:
### Issue

Duplicated items is displayed after rename favorite file

### Steps to reproduce

1. Mark file as favorite
2. Open favorite tab
3. Rename favorite file

### Changes

- Fixes the issue by always setting up-to-date cached then remote files properly
- Fixes refresh after rename
- Fixes refresh after delete
- Fixes resetting layout after rename
- Fixes loading state after rename for search task
- Simplifies rename logic via early exits and separate functions
- Add comments to understand flow better

### DIFF
--- a/app/src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
+++ b/app/src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
@@ -338,6 +338,7 @@ public final class AppPreferencesImpl implements AppPreferences {
     @Override
     public String getFolderLayout(FolderLayout layout) {
         if (layout instanceof FolderLayout.Child child) {
+            // keep existing logic for child directories
             return getFolderPreference(context,
                                        userAccountManager.getUser(),
                                        layout.getKey(),
@@ -364,6 +365,7 @@ public final class AppPreferencesImpl implements AppPreferences {
             return;
         }
 
+        // only use new way for root shared, favorite and all files
         preferences.edit()
             .putString(layout.getPrefKey(userAccountManager.getUser()), layoutName)
             .apply();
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -2341,38 +2341,7 @@ class FileDisplayActivity :
...
-                    if (PreviewTextFileFragment.canBePreviewed(renamedFile)) {
-                        startTextPreview(renamedFile, true)
-                    } else {
-                        fileOperationsHelper.openFile(renamedFile)
-                    }
-                }
-            }
-
-            val file = storageManager.getFileById(renamedFile.parentId)
-            if (file != null && file == getCurrentDir()) {
```
alperozturk96, line 2381: `if (file != null && file == getCurrentDir()) {`

Is there any reason to use this condition here @tobiasKaminsky ?

app/src/main/java/com/nextcloud/client/preferences/AppPreferencesImpl.java
```diff
@@ -336,12 +336,12 @@ public String[] getPassCode() {
     }
 
     @Override
-    public String getFolderLayout(OCFile folder) {
+    public String getFolderLayout(OCFile folder, String defaultLayout) {
```
tobiasKaminsky: Please adopt to use it like in " setSortOrder(FileSortOrder.Type type"

tobiasKaminsky (changes requested):

app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
tobiasKaminsky: We need it, yes. If e.g. someone renames a file and directly navigates to another folder.
then we do not need to refresh the other folder.

tobiasKaminsky (dismissed):

tobiasKaminsky (approved):

PR #17025, opened 2026-05-11
fix(share-activity): ui/ux

alperozturk96:
### Issue

User can navigate share detail fragment from `ShareActivity` thus it needs to show exact same UI just like navigated from file listing.

### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
Screenshot_20260511_131504 |  Screenshot_20260521_105916

### Changes

- Fixes UI
- Fixes thumbnail generation
- Allow user to go back properly
- Converts to Kotlin


tobiasKaminsky (dismissed):
Please test with NC Collobara -> sharing (or use sharing option within text/md)

Please "only" fix thumbnail generation and add padding/margin to dialog, but not chang the flow of fragment/dialog, as it might be too risky.

tobiasKaminsky (approved):

PR #17069, opened 2026-05-19
fix: handle negative lock timeout (ETA_INFINITE sentinel) from files_lock

moodyjmz:
## Summary

- When `files_lock` is configured with infinite timeout (the default, `-1` minutes), it sends `nc:lock-timeout=-60` via WebDAV PROPFIND
- `getLockedUntil()` was computing `(lockTimestamp + lockTimeout) * MILLIS_PER_SECOND` without guarding against negative timeout values, resulting in a date in the past
- This caused the UI to show **"Expires: 1 minute ago"** on actively locked files

The guard was changed from `file.lockTimeout == 0L` to `file.lockTimeout <= 0L` so any non-positive timeout (including the `-60` sentinel) correctly returns `null` (no expiry shown).

The root cause fix is server-side in `files_lock` ([see related PR](https://github.com/nextcloud/files_lock)). This is a defensive client-side guard for servers that haven't been updated yet.

## Test plan

- [ ] `FileActionsViewModelTest` — 4 unit tests covering `lockTimeout == 0`, `lockTimeout < 0` (`-60` sentinel), `lockTimestamp == 0`, and positive values
- [ ] Open a file locked with infinite timeout on an unpatched server and verify no expiry label is shown

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/app/src/main/java/com/nextcloud/ui/fileactions/FileActionsViewModel.kt
+++ b/app/src/main/java/com/nextcloud/ui/fileactions/FileActionsViewModel.kt
@@ -9,6 +9,7 @@ package com.nextcloud.ui.fileactions
 
 import android.os.Bundle
 import androidx.annotation.IdRes
+import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/nextcloud/ui/fileactions/FileActionsViewModel.kt
```diff
@@ -125,7 +126,8 @@ class FileActionsViewModel @Inject constructor(
         }
     }
 
-    private fun getLockedUntil(file: OCFile): Long? = if (file.lockTimestamp == 0L || file.lockTimeout == 0L) {
+    @VisibleForTesting
+    internal fun getLockedUntil(file: OCFile): Long? = if (file.lockTimestamp == 0L || file.lockTimeout <= 0L) {
```
alperozturk96, line 130: And for `lockTimestamp` this is not the case?

PR #17095, opened 2026-05-26
improve upload cancellation

alperozturk96:
### Changes

- Preserve user's file selection for upload
- Remove file if already uploaded after cancel upload
- Update DB only from `UploadFileOperation`

### Demo

https://github.com/user-attachments/assets/b86be5be-3650-4e3b-8f06-35fc00a01d21

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/LocalFileListAdapter.java
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/LocalFileListAdapter.java
@@ -39,7 +39,7 @@ import com.owncloud.android.utils.theme.ViewThemeUtils;
 
 import java.io.File;
 import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/owncloud/android/ui/adapter/LocalFileListAdapter.java
```diff
@@ -88,7 +88,7 @@ public LocalFileListAdapter(boolean localFolderPickerMode,
         mContext = context;
         mLocalFolderPicker = localFolderPickerMode;
         this.localFileListFragmentInterface = localFileListFragmentInterface;
-        checkedFiles = new HashSet<>();
+        checkedFiles = new LinkedHashSet<>();
```
alperozturk96, line 91: Preserve user's file selection for upload

PR #17104, opened 2026-05-30
Fix slow file list loading when entering a directory for the first time with 1000+ files.

berlin2123:
### Summary

When syncing a folder with many files for the first time, each remote file triggered a ContentProvider getFileByPath() query. For a directory with 1000+ files, this meant 1000+ individual database lookups — all of which returned null because the files did not exist locally yet.

Skip the per-file fallback when localFilesMap is empty, since no local state exists to merge.

### 🏁 Checklist
 

berlin2123:
I dnot know how to build as a apk, so can not test.

PR #17108, opened 2026-05-31
Kotlin conversion of UploadFileOperation

hannesa2:
I was investigation how to change Wifi upload mechanism, and I found that causing class is still Java 👎
So I converted it in advance to Kotlin without changing anything in the logic. I hope that tests will check, if there is something now faulty


alperozturk96 (changes requested):
Thanks for working on this.

I'd be more comfortable taking an incremental approach, for example, first extracting smaller pieces into separate Kotlin classes and then gradually converting the remaining parts. That would make the changes easier to review, test, and debug if anything goes wrong.

hannesa2:
@alperozturk96 After 5 weeks. you want to multiply the afford here by splitting it in smaller junks.
As this is just the output of 
image

you should re-think your comfortable it's-better-for-me-way

tobiasKaminsky:
@hannesa2 

Alper and me discussed this and to put it in bigger context, we struggle a bit with "automatic" java -> kotlin conversions, especially when it comes to big/central classes that are not easy to test manually (in opposite to UI related classes).
So my/our idea is now to test with your PR
- check code coverage of old java file
- add missing tests
- test tests with your migrated code
- some test

This will help us, not only for this PR, but also for upcoming semi/automated code conversions to see if this approach works.

I wanna thank you for helping with this project and PR :tada:

alperozturk96:
It seems like the following cases are currently missing on master and would be good to have covered before the Kotlin conversion:

* `FileUploaderIT` doesn't seem to be used. Already contains tests for `UploadFileOperation`.
* `UploadIT` covers some `UploadFileOperation` scenarios, but not all combinations.
* The constructor guards in `UploadFileOperation` aren't tested.
* The `encryptedUpload` path isn't covered.
* The early return cases are missing as well (e.g. `TextUtils.isEmpty(getStoragePath())`, `!localFile.exists()`, etc.).
* I also couldn't find tests covering methods like `obtainNewOCFileToUpload`, `getNewAvailableRemotePath`, `existsFile`, and `createNewOCFile`.

Since this class is responsible for all uploads (AutoUpload, normal upload, etc.).

hannesa2:
> I wanna thank you for helping with this project and PR 🎉

@tobiasKaminsky I want to tell my my frustration. I made this, I didn't changed anything on the logic and it was working for me. But as it is not yours, you all put the brake and hesitate to accept it. -Ok, it's your choice-

But what does this mean for me ? It was just an intermediate step for me to solving an other (later) issue. But as this is not accepted, and dedicated to be pointless and to die, I stopped my afford to Nextcloud complete and unfortunately I've to keep on ownCloud. But the good thing for, there is the situation the same bad one.

Looks like a common behaviour "I/we know it better", so it's better to do nothing instead of moving forward.

Btw, you have at least a lot of CI test, but as they fail and nobody takes care about it and **ignores them**, this all CI checks are complete pointless and just waste of time of everyone.
image

PR #17139, opened 2026-06-05
Test CI

hannesa2:
It's just to see the result of the CI without any change at all


hannesa2:
@alperozturk96 @AndyScherzinger 
No clue who is the right person here, I just pick the lastest 2 merger.

As it looks like the test is broken and nobody takes care about it 
image
It looks like, you simply ignore this error. Ok, this one way to handle it, but what's the point of this test, when it's broken.
I really like this screenshot compare mechanism, but when it's just an ignored error you could delete it complete.

PR #17140, opened 2026-06-05
Add a favorite icon to the search screen

AlexViseven:
It was added displaying of favorite icon while searching files/folders
image

### DIFF
--- a/app/src/main/java/com/owncloud/android/ui/adapter/UnifiedSearchItemViewHolder.kt
+++ b/app/src/main/java/com/owncloud/android/ui/adapter/UnifiedSearchItemViewHolder.kt
@@ -61,6 +61,7 @@ class UnifiedSearchItemViewHolder(
         bindTextView(binding.title, entry.title)
         bindTextView(binding.subline, entry.subline)
         bindLocalFileIndicator(entry)
+        bindFavoriteIndicator(entry)
 
         val entryType = entry.getType()
         bindThumbnail(entry, entryType)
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/owncloud/android/ui/adapter/UnifiedSearchItemViewHolder.kt
```diff
@@ -85,6 +86,11 @@ class UnifiedSearchItemViewHolder(
         binding.localFileIndicator.setVisibleIf(showLocalFileIndicator)
     }
 
+    private fun bindFavoriteIndicator(entry: SearchResultEntry) {
+        val file = storageManager.getFileByRemotePath(entry.remotePath())
```
alperozturk96: I can see multiple DB read for same path

`val file = storageManager.getFileByRemotePath(entry.remotePath())`

Please only read beginning of the ` fun bind(entry: SearchResultEntry)` once then use it below for others.

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/ui/adapter/UnifiedSearchItemViewHolder.kt
AlexViseven: Applied

alperozturk96 (changes requested):
Thanks for the update, seems like in this folder section's alignment is not correct.

Image

alperozturk96:
Hey

Is this PR still active?

AlexViseven:
@alperozturk96 I don't see any such problem when launching. Could you please adjust the margins in the file unified_search_item (element @+id/favorite_action")? Probably it would be enough just to remove margins

alperozturk96:
replaced via: https://github.com/nextcloud/android/pull/17568

PR #17145, opened 2026-06-06
Remove unused class

hannesa2:
The class is simply not used


alperozturk96 (commented):
@tobiasKaminsky Any reason to keep this? As far as I can see not used at all.

tobiasKaminsky:
@alperozturk96 please fix the usage of it, instead of simply removing it.

alperozturk96:
> @alperozturk96 please fix the usage of it, instead of simply removing it.

https://github.com/nextcloud/android/pull/17268/changes/d18a7dd836a120b0779cdd10ee889b8d0b426aae

PR #16578, opened 2026-02-24
Screenshot tests were failing, JUnit downgrade needed

tobiasKaminsky:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist
 
- [ ] Tests written, or not not needed

PR #17146, opened 2026-06-06
Remove screenshot test action

hannesa2:
close #17145 
as the gradle task simply not exists


tobiasKaminsky:
Here we try to revive screenshots: https://github.com/nextcloud/android/pull/16578
So I think we should follow the other PR first, and only if this does not work, we should remove the CI tasks.
Closing it thus for now.

PR #17243, opened 2026-06-23
🎨 Tint leading list icons in dark mode

AndyScherzinger:
SVG activity/notification icons decode into a `PictureDrawable`, which ignores tint, so the leading icons stayed dark-on-dark in dark mode. Now rasterized into a tintable `BitmapDrawable` and themed via `viewThemeUtils`, so the tint applies in both themes.

AI disclosure: assisted by ClaudeCode (claude-opus-4-8).


nickvergessen (approved):
Works 👍 and code looks reasonable 🤷

tobiasKaminsky (approved):

PR #17267, opened 2026-07-01
fix(e2ee): refresh during crud operations

alperozturk96:
### Issue

Refreshing folder causing crash during crud operations in encrypted folder. Local counter should be only updated after completion of operation.

### Changes

- Consolidates counter updates
- Updates counter after `serializeAndUploadMetadata` and unlock folder
- Better naming


app/src/main/java/com/owncloud/android/operations/RemoveRemoteEncryptedFileOperation.kt
```diff
@@ -183,9 +183,7 @@ class RemoveRemoteEncryptedFileOperation internal constructor(
             encryptionUtilsV2.removeFileFromMetadata(fileName, metadata)
         }
 
-        parentFolder.setE2eCounter(metadata.metadata.counter)
         val storageManager = FileDataStorageManager(user, context.contentResolver)
-        storageManager.saveFile(parentFolder)
```
alperozturk96, line 188: Should be after `serializeAndUploadMetadata`

app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
```diff
@@ -904,11 +904,7 @@ private void updateMetadataForV2(DecryptedFolderMetadataFile metadata, Encryptio
             e2eData.getIv(),
             e2eData.getEncryptedFile().getAuthenticationTag(),
             e2eData.getKey(),
-            metadata,
-            getStorageManager());
-
-        parentFile.setE2eCounter(metadata.getMetadata().getCounter());
```
alperozturk96, line 898: Should be after `serializeAndUploadMetadata`

app/src/main/java/com/owncloud/android/utils/EncryptionUtilsV2.kt
```diff
@@ -487,23 +486,17 @@ class EncryptionUtilsV2 {
 
         metadataFile.metadata.files[encryptedFileName] = decryptedFile
         metadataFile.metadata.counter++
-        ocFile.setE2eCounter(metadataFile.metadata.counter)
```
alperozturk96, line 489: `addFileToMetadata` has one caller and that already updates counter. No need to double it.

app/src/main/java/com/owncloud/android/utils/EncryptionUtilsV2.kt
```diff
@@ -487,23 +486,17 @@ class EncryptionUtilsV2 {
...
         encryptedFileName: String,
         fileName: String,
-        metadataFile: DecryptedFolderMetadataFile,
-        ocFile: OCFile,
-        fileDataStorageManager: FileDataStorageManager
+        metadataFile: DecryptedFolderMetadataFile
     ): DecryptedFolderMetadataFile {
         metadataFile.metadata.folders[encryptedFileName] = fileName
         metadataFile.metadata.counter++
-        ocFile.setE2eCounter(metadataFile.metadata.counter)
```
alperozturk96, line 504: `addFolderToMetadata` has one caller and that already updates counter. No need to double it.

alperozturk96: Also aligns with other functions such as `removeFolderFromMetadata`

app/src/main/java/com/owncloud/android/utils/EncryptionUtilsV2.kt
alperozturk96: Also aligns with other functions such as `removeFolderFromMetadata`

tobiasKaminsky (approved):

app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -532,6 +535,14 @@ private void synchronizeData(List<Object> folderAndFiles) {
 
         Object object = null;
         if (mLocalFolder.isEncrypted()) {
+            final var workManager =  WorkManager.getInstance(mContext);
+            final var tag = BackgroundJobManagerImpl.JOB_FILES_UPLOAD + user.getAccountName();
+            boolean isWorkScheduled = WorkManagerExtensionsKt.isWorkScheduled(workManager, tag);
+            if (isWorkScheduled) {
+                Log_OC.d(TAG, "Upload worker running for " + user.getAccountName() + "; deferring metadata/counter update to next refresh");
+                return;
```
tobiasKaminsky: Please check if the counter is really correctly updated?
I assume that counter is not correctly updated for upload work and thus it works only if we do the operations one after another.
Please double check.

tobiasKaminsky (changes requested):

app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
alperozturk96: 1. User tries to upload 5 encrypted files
2. `UploadFileOperation` updates counter one by one after each successful upload
3. User during uploads do the pull-to-refresh
4. Refresh folder operation uses `mLocalFolder` and sends newly fetch counter via download metadata e.g. 23
5. The refresh downloads metadata still at 23, but the upload already pushed the local counter to 22
6. Crashes because after pull to refresh upload operation at that time didn't push the latest value yet because it received 23 at that time.

This is not related to the updating logic, this is related to the timing and async execution.

Solution:

1. Don't let the refresh verify metadata for a folder while an upload still in progress
2. ?

PR #17268, opened 2026-07-01
Fix CI tests

tobiasKaminsky:
### 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
B | A

### 🏁 Checklist
 
- [ ] ⛑️ Tests (unit and/or integration) are included or not needed
- [ ] 📅 Milestone is set
- [ ] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI


alperozturk96:
I will check those later

Screenshot 2026-07-06 at 08 08 58

alperozturk96:
Seems like only following exception needs to be fixed:

`androidx.test.espresso.base.RootViewPicker$RootViewWithoutFocusException: Waited for the root of the view hierarchy to have window focus and not request layout for 10 seconds. If you specified a non default root matcher, it may be picking a root that never takes focus. Root:`

Screenshot 2026-07-13 at 16 27 40

andreBurnt:
Confirming the impact of this fix from the outside, plus one thing worth checking on API 33+.

That raw `GrantPermissionRule.grant(WRITE_EXTERNAL_STORAGE, POST_NOTIFICATIONS)` was the single largest failure cluster in the CI history I analyzed (2026-04-10 to 2026-07-09): 11,061 failure events across 414 test ids, which is 67% of all failure events in the window. Both CI emulator lanes (API 28 and 29) hit it on every run. I reproduced it on my own bench across three API levels (28 + 34 emulators, physical API 31), 31 iterations, zero passes. `grantStorageAndNotification()` gates POST_NOTIFICATIONS behind `SDK_INT >= TIRAMISU`, so it removes the dominant source of red in the suite.

One thing worth checking, on API 33+. The new rule still grants `WRITE_EXTERNAL_STORAGE` on both branches. The manifest caps that permission at `maxSdkVersion="29"` (still present on master), so on my API-34 bench the grant threw `Package com.nextcloud.client has not requested permission WRITE_EXTERNAL_STORAGE`. CI runs API 28/29, where the storage grant is valid, so this does not affect the lanes you're fixing. But if you ever add an API 33+ instrumentation lane, the storage half would throw there. A cheap guard: request `WRITE_EXTERNAL_STORAGE` only when `SDK_INT <= 29`.

Happy to share the repro loop and the cluster breakdown if useful.

alperozturk96:
> master-IT test failed: https://www.kaminsky.me/nc-dev/android-integrationTests/17268-IT-master-12-28

We will check these with different PR.

PR #17302, opened 2026-07-07
fix(conflict-resolve) check

alperozturk96:
### Issue

Successful file uploads was showing conflict resolution notification.

### How to reproduce?

1. Create a custom auto upload configuration
2. Use a voice recorder app to record audio files to the chosen custom folder

### Changes

- Unifies usage of `isSameFileOnRemote`
- Adds eTag for `OCUpload` for eTag comparison against newly fetched `RemoteFile`. `OCFile` cannot be used because eTag field of `OCFile` not consistent and gets updated from multiple places. eTag of `OCUpload` only gets updated from one place.

### DIFF
--- a/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadHelper.kt
+++ b/app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadHelper.kt
@@ -23,11 +23,14 @@ import com.nextcloud.client.jobs.BackgroundJobManager
 import com.nextcloud.client.network.Connectivity
 import com.nextcloud.client.network.ConnectivityService
 import com.nextcloud.client.notifications.AppWideNotificationManager
+import com.nextcloud.utils.TimeConstants
 import com.nextcloud.utils.extensions.checkWCFRestrictions
+import com.nextcloud.utils.extensions.getBitmapSize
+import com.nextcloud.utils.extensions.getExifSize
 import com.nextcloud.utils.extensions.getUploadIds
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadHelper.kt
```diff
@@ -578,22 +588,77 @@ class FileUploadHelper {
...
+            }
+            return result
         }
+
+        Log_OC.d(TAG, "local file different than remote file")
+
         return false
     }
 
+    private fun isETagUnchangedSinceLastUpload(remoteFile: RemoteFile, localPath: String, remotePath: String): Boolean {
```
alperozturk96, line 620: `isSameFileOnRemote` collision check relied only on `RemoteFile.isSame() (size + creation/modification timestamps + image dimensions)`, which is not reliable.

I added an etag column to `OCUpload`, written once at upload success `UploadFileOperation` copies file.getEtag() from the post-upload and never mutated afterward.

The `etag` of the `OCFile` is not stable it is always mutated from somewhere else and causing issue to compare.

For example `A.jpg` uploaded before I must have `eTag` and `eTagOnServer` value but when I check database `eTag` field is empty. I need simple comparison `eTag` is changed or not.

Image

@tobiasKaminsky CC

tobiasKaminsky (changes requested):
Let us first try to find out why etag on ocfile is empty.

PR #17438, opened 2026-08-05
fix(file-list): list directory 🚀

### LINKED ISSUE #14855: SyncBroadcastReceiver in FileDisplayActivity Triggers Multiple Times Causing Repeated Directory Listing

alperozturk96:
### Comparasion

Operation | Before | After
-- | -- | -- 
Cold Start | 4 | 2 
Directory Swap | 2 | 1 
Go Back to Root | 3 | 2
 
### Fixes

- Listing files unnecessarily. Which fixes unnecessary api calls, re-filling adapter, thumbnail generation etc.

https://github.com/nextcloud/android/issues/14855

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
@@ -304,8 +304,7 @@ public class RefreshFolderOperation extends RemoteOperation {
             sendLocalBroadcast(EVENT_SINGLE_FOLDER_CONTENTS_SYNCED, mLocalFolder.getRemotePath(), result);
         }
 
-
-        final var remoteFiles = ResultParser.dataOfType(result, RemoteFile.class);
+        final var remoteFiles = ResultParser.list(result, RemoteFile.class);
         if (!remoteFiles.isEmpty() && !mSyncFullAccount && !mOnlyFileMetadata) {
             // this needed because if file has new share or share is removed, eTag is not changing.
             // that's why another separate EVENT_SINGLE_FOLDER_SHARES_SYNCED introduced before.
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/owncloud/android/ui/activity/FileDisplayActivity.kt
```diff
@@ -227,6 +227,8 @@ class FileDisplayActivity :
     private var mPlayerConnection: PlayerServiceConnection? = null
     private var lastDisplayedAccountName: String? = null
 
+    private var listFragmentJustCreated = false
```
alperozturk96, line 233: fixes first time app launch multiple listing directory

app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
```diff
@@ -134,6 +134,11 @@ public class RefreshFolderOperation extends RemoteOperation {
      */
     private boolean mRemoteFolderChanged;
 
+    /**
+     * 'True' means that the sharees of at least one child of the folder changed
+     */
+    private boolean sharesChanged;
```
alperozturk96, line 140: Fixes refresh folder + shares event trigger call for same directory. However only once enough and only second event needs to be fired actually shares updated in local storage.

AndyScherzinger (approved):

AndyScherzinger (approved):

daniele-verducci (approved):

PR #17465, opened 2026-08-12
perf(file-list): initial load, refresh 🚀 

alperozturk96:
### Changes

- Update user profile and oc version after refresh events thus no need to wait those calls since they are not needed for file list
- Remove hardcoded delay to execute refresh folder operation use instead `onWindowFocusChanged`

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/RefreshFolderOperation.java
@@ -263,10 +263,10 @@ public class RefreshFolderOperation extends RemoteOperation {
             return new RemoteOperationResult<>(ResultCode.FILE_NOT_FOUND);
         }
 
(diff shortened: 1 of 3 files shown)


daniele-verducci (approved):
Hello, first PR review for me, so I'll be a little more verbose.
I checked the changes and seem to match the description on the first post. Installed the app from qr code and tested on a device, I can't say if the difference in speed is perceptible, but it definitely feels snappy and I found no problems.

alperozturk96:
> Hello, first PR review for me, so I'll be a little more verbose. I checked the changes and seem to match the description on the first post. Installed the app from qr code and tested on a device, I can't say if the difference in speed is perceptible, but it definitely feels snappy and I found no problems.

This PR removes delay of the refresh action and sync logic. So a bit faster refresh we got after this PR.

Speed difference based on execution of these two blocker API call. If instance and/or connection slow, speed will be more noticeable.

```
updateOCVersion(client);
updateUserProfile();
```

PR #17470, opened 2026-08-13
fix(media): video player returning from background

alperozturk96:
### Steps to reproduce

1. Open video file
2. Put in full screen
3. Put app in background
4. Put app in foreground
5. Video not starts where it left off before


daniele-verducci (approved):
Tested and works perfectly on my device

PR #17471, opened 2026-08-13
fix(navigation): search fragment to root

alperozturk96:
### Steps to reproduce

1. Download a nested folder
2. Open On-device
3. Navigate to the sub folder
4. Press back button
5. Instead of the root of the On-device it still shows sub-folder


daniele-verducci (approved):
Could reproduce the issue on prod and can confirm this fixes it!

PR #17474, opened 2026-08-13
fix(unified-search): handle slow server search for loader spinner

alperozturk96:
### Issue

In slow server instance spinner always showed and files are loading in main thread.

### Changes

Loading spinner has max ceiling value
Loads the file in IO thread
Bind function directly takes `OCFile` thus no need for each bind database operation


daniele-verducci (approved):
It seems good to me. I'm not sure how to simulate a slow server, so I tested throttling down the connection and the search part works as intended on timeout. Also checked with an encrypted folder and the more button is correctly hidden.

PR #17479, opened 2026-08-14
feat(empty-state): oome

alperozturk96:
### Issue

Parsing big folder causing OOME. Needs to be handled in client side.

### Screenshot

Screenshot 2026-08-14 at 14 56 35


app/src/main/res/values/strings.xml
```diff
@@ -15,6 +15,7 @@
     <string name="bottom_navigation_menu_media_label">Media</string>
     <string name="file_list_error_headline">Poor connection</string>
     <string name="file_list_error_description">Check your internet connection or try again later</string>
+    <string name="file_list_out_of_memory_description">Could not able to load content</string>
```
daniele-verducci:
```suggestion
    <string name="file_list_out_of_memory_description">Could not load content</string>
```

daniele-verducci (approved):

PR #17503, opened 2026-08-19
Improve media provider

alperozturk96:
### Changes

- Converts to Kotlin
- Simplifies logic
- Clean up logs


daniele-verducci (approved):
Difficult to check all the code, so I tested it on my phone that has a lot of media folders, checked on the autoupload screen comparing with the released app and they match ✅

Issue #12967, opened 2024-05-03
Track auto upload files when original file is "kept in original folder"


ktmaul:
### Is your feature request related to a problem? Please describe.
The On Device feature is currently the only way for me to (relatively) quickly clear local files to save space on my device. However, auto upload files using the **kept in original folder** option do not track the original source file in a way that is compatible with the On Device feature. In order to remove these files, I must manually delete them after checking to see if NextCloud uploaded them.

### Describe the solution you'd like
I would greatly appreciate if files synchronized via auto upload would be mapped to the original file in a way that allows the file to show up in the On Device feature. This way, I can choose when to delete files that have been uploaded regardless of their current location. Once the file no longer exists on the device (whether from cleanup or deletion by another app), the mapping can be discarded and future downloads can use the NextCloud data folder.

### Describe alternatives you've considered
I have considered using the **deleted** option for original files, but that risks an error resulting in lost files which defeats the purpose of NextCloud as a backup. I prefer to manually review the upload errors to resolve problems and then clear the space myself.

I have attempted to use the **moved to app folder** option, but it complicates the use of other apps when interacting with the original files. For example, I might use an app similar to Google Photos as my primary way to interact with on-device media, while using NextCloud to backup to a private cloud. That app may expect to find media in the typical system folders and - while it *might* be able to access media in other folders - the user experience is significantly diminished.

### Additional context
N/A

enoch85:
My solution to the problem is:

1. Set overwrite existing file
2. Delete it from phone when uploaded.

I agree that this would be a great feature!

daniele-verducci:
From a technical standpoint, it would be useful to have that information on the database (knowing that a certain local file corresponds to a certain uploaded file, even if the file is ouside of the application's own folders).
The problem is, it can't be mapped using the same db column we already use for the manually uploaded files ("media_path"), because it's technically wrong (I tried, as a proof of concept, in this discarded PR: https://github.com/nextcloud/android/pull/17522 ).
The right way would probably be adding another column "for future proofing" (this is worthless until it's filled with data). This would benefit cases like the functionality that frees up space deleting already autouploaded files: https://github.com/nextcloud/android/pull/17516
This wouldn't necessarily display the green tick mark on the file like the manually uploaded/downloaded files, that should be discussed.

PR #17522, opened 2026-08-25
Fix missing local path for autouploaded files

daniele-verducci:
Files uploaded by autoupload do not have their filelist.media_path column populated in the db.
That is not the case for other workflows (i.e. manually uploading a single file).
This field allows them to be retrieved using FileDao.getFileByLocalPath(), in other words makes possible to reconstruct a file's origin.
The consequences of this change are still to be discussed, so marking as Draft.

### 🏁 Checklist
 
- [ ] ⛑️ Tests (unit and/or integration) are included or not needed
- [ ] 📅 Milestone is set
- [ ] 🌸 PR title is meaningful (if it should be in the changelog: is it meaningful to users?)

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI

### DIFF
--- a/app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
+++ b/app/src/main/java/com/owncloud/android/operations/UploadFileOperation.java
@@ -1410,7 +1410,6 @@ public class UploadFileOperation extends SyncOperation {
 
             default:
                 Log_OC.d(TAG, "DEFAULT local behaviour will be handled");
-                mFile.setStoragePath("");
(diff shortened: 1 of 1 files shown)


daniele-verducci:
After discussing it with @tobiasKaminsky , we found this is technically wrong, that field is meant for the files in the app's own paths and would have unwanted consequences if filled with external paths.
So I discard this PR, the discussion can continue on the relevant issue: https://github.com/nextcloud/android/issues/12967

PR #17539, opened 2026-08-31
Fix media view not being refreshed when selecting a different media folder

daniele-verducci:
How to reproduce:
- Open app
- Go to media view
- 3 dots menu -> select "Set media folder"
- Choose another media folder
- The view is not refreshed*

* = this may not happen the first time, as the check is successful if there have been changes in the media folder. If you can't reproduce, retry.

### 🏁 Checklist
 
- [x] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
- [x] 📅 Milestone is set

## 🤖 AI (if applicable)


daniele-verducci:
@alperozturk96 I'm not sure this is the correct solution, can you take a look? I saw you worked on the same files previously.
The problem is that currently the check to decide if refresh the view is done on the return value of `parseMedia()`, but that one, as far as I can understand, refreshes the local files based on the remote ones and returnes true _only if there have been changes_. The result is placed in an `emptySearch` variable, so maybe it's just a misunderstanding. I think emptySearch must be true only if there are no files at all. Is it correct?
(note that the media view doesn't have a search feature, so the search is always called with an empty query)

alperozturk96 (approved):

PR #17564, opened 2026-09-07
refactor(thumbnail-loader): to player thumbnail loader

alperozturk96:
Renames the `ThumbnailLoader` to `PlayerThumbnailLoader` so that we don't get confusion between `ThumbnailGenerator`.


daniele-verducci (approved):
Hey thank you for improving code readability, definitely approve!

PR #17565, opened 2026-09-07
fix(player): 3-dot menu

alperozturk96:
### Changes

- Fixes visibility of the 3-dot menu in video preview
- Adds missing actions


daniele-verducci (approved):
Nice! Tested and working for me (the only item I couldn't test in the menu is the `Stream with...`, I think I need a DLNA TV to test it).

PR #17566, opened 2026-09-07
fix(notification): empty state

alperozturk96:
### Issue

When a user has a notification with a primary and secondary action and presses one of them, the empty state of the notification fragment is not set correctly.

### Changes

Fix the empty state of the notification list.


daniele-verducci (approved):
Tested, riproduced on master & working here 👍

PR #17567, opened 2026-09-07
fix(upload): limited file access

alperozturk96:
### Steps to reproduce

1. Clean install of the app
2. Storage permissions popup select "Media read-only"
3. Select "Allow limited access" for some specific files
4. Tap FAB button that opens local file listing
5. Select same media previously allowed and observe while select / unselect / select all media file
6. Tap upload button

**Actual Result:**

1. When selecting/deselecting media files or using the **Select All** option, the UI appears to hang. In some cases, media files cannot be manually selected or deselected.
2. Clicking the **Upload** button is unresponsive, and the upload process does not start. Only the **Cancel** button works.
3. The **Upload** option is also not clickable when accessed from the footer section.

**Expected Result:**
With **Read-Only** and **Limited Access** permissions, users should be able to upload files without any issues.

### Changes

- Fixes UI lag
- If user selects limited file access do not shows storage permission banner

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -49,9 +49,10 @@
     -->
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" /> <!-- Runtime permissions introduced in Android 13 (API level 33) -->
     <uses-permission android:name="android.permission.READ_MEDIA_IMAGES"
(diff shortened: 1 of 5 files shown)


daniele-verducci (approved):
Hey Alper, I couldn't reproduce the issue on the master branch, but I tested this one and seems to work properly. I tested on Android 16 granting different levels of permissions and on Android 10, that doesn't support for granular permissions, and it worked as I would expect.

PR #17569, opened 2026-09-07
improve: fetch-template

alperozturk96:
### Changes

- Uses `TemplateFilenameState` instead of booleans
- `hasUserInteracted` guard helps to prevent error shown before the user types
- Empty template list now dismisses the dialog with "Error retrieving templates" instead of leaving an unusable dialog open
- Converts `FetchTemplateOperation` to Kotlin


daniele-verducci (approved):
Oh I noted the interaction on the creator popup was a little rough, very nice improvement! Thx!

Issue #17589, opened 2026-09-08
Offline upload conflicts resolution overwrites the remote file even if the user select to keep both


daniele-verducci:
### ⚠️ Before posting ⚠️

- [x] This is a **bug**, not a question or an enhancement.
- [x] I've [searched for similar issues](https://github.com/nextcloud/android/issues) and didn't find a duplicate.
- [x] I've written a clear and descriptive title for this issue, not just "Bug" or "Crash".
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Steps to reproduce

1. Create a folder
2. Upload a file (e.g. a text file)
3. Go offline (e.g. airplane mode)
4. Change the file (e.g. add some characters)
5. Upload a second time the file
6. Go online
7. A conflict message is shown
8. Choose to keep BOTH files
9. The online file is overwritten by the new uploaded file (only one file present in the folder at the end)

### Expected behaviour

The folder should contain two files: the remote one and the newly uploaded, renamed (as it happens if following the same steps without ever going offline).

### Actual behaviour

The online file is overwritten by the new uploaded file (only one file present in the folder at the end)

### Android version

17

### Device brand and model

Emulator on Android Studio

### Stock or custom OS?

Stock

### Nextcloud android app version

35.0.0 RC1

### Nextcloud server version

Nextcloud Hub 26 Spring (36.0.0 dev)

### Using a reverse proxy?

No

### Server error logs

```bash

```

### Additional information

I _think_ it's related to ConflictsResolveActivity having its own flow for offline uploads (see showOfflineOperationConflictDialog()) and passing Decision.KEEP_BOTH_FOLDER instead of KEEP_BOTH to the worker

daniele-verducci:
This should be fixed by https://github.com/nextcloud/android/pull/17590

PR #17590, opened 2026-09-08
Fix conflicts for offline uploads when uploading same file again

daniele-verducci:
Fixes the following two problems:

## Conflict message being shown for offline uploads when the offline-uploaded file is the same as the remote

Fixes the conflict message being shown for offline uploads when the offline-uploaded file is the same as the remote. This makes the behavior aligned with what happens online.

**How to repro:**
- Create a folder
- Upload a file
- Go offline (e.g. airplane mode)
- Upload a second time the same, unmodified file
- Go online
- A conflict message is shown

Now execute the same actions in a new folder without going offline:
- A conflict message is NOT shown

This is because the offline upload erases the etag_on_server field for the already existing db entry.
This PR fixes this issue.

## Remote file overwritted by local file when selecting "keep both" after offline upload

This PR does fix this this second issue: https://github.com/nextcloud/android/issues/17589

**How to repro:**
- Create a folder
- Upload a file
- Go offline (e.g. airplane mode)
- Upload a second time the same, unmodified file
- Go online
- A conflict notification is shown
- Click resolve, note the dialog: it says folder conflict, and the dates and sizes are wrong
- Select both versions and click ok
- Notice the remote version is overwritten by the local one. At the end you are left with a single file in the folder instead of two.

### 🏁 Checklist
 
- [x] 🔙 Backport requests are created or not needed: `/backport to stable-xx.x`
- [x] 📅 Milestone is set

## 🤖 AI (if applicable)


daniele-verducci:
@tobiasKaminsky should this be in milestone 35 or 36?

app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -230,14 +233,52 @@ public OfflineOperationEntity addCreateFolderOfflineOperation(String path, Strin
...
+        }
+
+        OCFile file = new OCFile(remotePath);
         file.setMimeType(mimeType);
         file.setCreationTimestamp(createdAt);
         file.setModificationTimestamp(modificationTimestamp);
         saveFileWithParent(file, MainApp.getAppContext());
     }
 
+    private boolean fileIsTheSame(OCFile ocFile, File localFile) {
```
alperozturk96: Could you please use the `FileExtensions.kt`?

app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -230,14 +233,52 @@ public OfflineOperationEntity addCreateFolderOfflineOperation(String path, Strin
...
+    public void createPendingFile(
+        String remotePath,
+        String mimeType,
+        long createdAt,
+        long modificationTimestamp,
+        String localPath
+     ) {
+        final OCFile existingFile = getFileByRemotePath(remotePath);
+        if (existingFile != null) {
+            final File localFile = new File(localPath);
```
alperozturk96: You can also move this to the `FileExtensions.kt` and use `localPath.toFile()` function.

Then if block fully comes from the extension file.

app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
```diff
@@ -343,6 +343,13 @@ class FileUploadWorker(
                 Log_OC.e(TAG, "credentials are no longer valid, stopping uploads")
                 break
             }
+
+            if (result.isSuccess && result.resultData is String) {
```
alperozturk96: What `result.resultData` can be? Always `String` or something else also possible?

app/src/main/java/com/nextcloud/client/jobs/upload/FileUploadWorker.kt
```diff
@@ -343,6 +343,13 @@ class FileUploadWorker(
...
+                // Update the remote etag to avoid conflict when uploading same file, both in online and offline uploads
+                val file = operation.storageManager.getFileByRemotePath(upload.remotePath)
+                file?.etagOnServer = result.resultData.toString()
+                operation.storageManager.saveFile(file)
```
alperozturk96: `UploadFileOperation.handleLocalBehaviour(temporalFile, expectedFile, originalFile, client);` already saves the file why we are checking here in worker again?

alperozturk96 (changes requested):

alperozturk96 (approved):

daniele-verducci:
Hey @alperozturk96 , as anticipated I fixed also the missing case, so I re-requested your review. You can check just the last 4 (new) commits.
Basically, **before the last 3 commits** :

- Create a folder
- Upload a file
- Go offline (e.g. airplane mode)
- Upload a second time the same, unmodified file
- Go online
- A conflict notification is shown
- Click resolve, note the dialog: it says **folder** conflict, and the dates and sizes are wrong
- Select both versions and click ok
- Notice the remote version is overwritten by the local one. At the end you are left with a single file in the folder instead of two.

So, the last commits fixes the dialog content and the overwrite. If approved, will close https://github.com/nextcloud/android/issues/17589 .

Thank you!

app/src/main/java/com/nextcloud/client/jobs/offlineOperations/repository/OfflineOperationsRepository.kt
```diff
@@ -73,34 +73,41 @@ class OfflineOperationsRepository(private val fileDataStorageManager: FileDataSt
...
-                            }
+                            updateOperationPath(newPath, nextOperation)
                         }
                     }
                 }
             }
             .forEach { dao.update(it) }
     }
 
+    private fun updateOperationPath(newPath: String, nextOperation: OfflineOperationEntity): OfflineOperationEntity? =
```
alperozturk96:
```
private fun updateOperationPath(newPath: String, nextOperation: OfflineOperationEntity): OfflineOperationEntity? {
        if (newPath == nextOperation.path) return null

        val updatedType = when (val currentType = nextOperation.type) {
            is OfflineOperationType.CreateFile -> currentType.copy(remotePath = newPath)
            is OfflineOperationType.CreateFolder -> currentType.copy(path = newPath)
            else -> currentType
        }

        return nextOperation.apply {
            type = updatedType
            path = newPath
        }
    }
```

app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -230,8 +232,22 @@ public OfflineOperationEntity addCreateFolderOfflineOperation(String path, Strin
...
+        String remotePath,
+        String mimeType,
+        long createdAt,
+        long modificationTimestamp,
+        String localPath
+     ) {
+        final OCFile existingFile = getFileByRemotePath(remotePath);
+        final File localFile = FileExtensionsKt.toFile(localPath);
+        if (FileExtensionsKt.isTheSameAs(existingFile, localFile)) {
+            // In case the same file was already uploaded, do not overwrite it to avoid triggering a conflict
```
alperozturk96: Just a tiny thing :) No need to have another comment since we have `Log_OC.i`. You can update it to cover both.

app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -353,13 +366,13 @@ public void keepOfflineOperationAndServerFile(OfflineOperationEntity entity, OCF
...
         if (parentFolder == null) return;
 
-        DateFormatPattern formatPattern = DateFormatPattern.FullDateWithHours;
-        String currentDateTime = DateExtensionsKt.currentDateRepresentation(new Date(), formatPattern);
-
-        String newFolderName = oldFileName + " - " + currentDateTime;
-        String newPath = parentFolder.getDecryptedRemotePath() + newFolderName + OCFile.PATH_SEPARATOR;
-        moveLocalFile(file, newPath, parentFolder.getDecryptedRemotePath());
-        offlineOperationsRepository.updateNextOperations(entity);
+        final String newPath = UploadFileOperation.getNewAvailableRemotePath(
```
alperozturk96: Good improvement with `getNewAvailableRemotePath`. 

`moveLocalFile` was unnecessary here since file is already there.

alperozturk96 (changes requested):

daniele-verducci:
Hey @alperozturk96 , thanks for the review and the suggestion! I applied them, now I wait for the CI to run.

app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
```diff
@@ -340,10 +356,7 @@ public void renameOfflineOperation(OCFile file, String newFolderName) {
         moveLocalFile(file, newPath, parentFolder.getDecryptedRemotePath());
     }
 
-    @SuppressLint("SimpleDateFormat")
-    public void keepOfflineOperationAndServerFile(OfflineOperationEntity entity, OCFile file) {
-        if (file == null) return;
-
+    public void keepOfflineOperationAndServerFile(OfflineOperationEntity entity, OCFile file, OwnCloudClient client) {
```
alperozturk96, line 359: `OCFile file` this can be null thus removing `if (file == null) return;` will cause `NPE`. Since it is called from `ConflictResolveActivity.kt` via nullable server file.

app/src/main/java/com/owncloud/android/ui/activity/ConflictsResolveActivity.kt
```diff
@@ -201,7 +201,11 @@ class ConflictsResolveActivity :
 
     private suspend fun keepBothFolder(offlineOperation: OfflineOperationEntity?, serverFile: OCFile?) {
         offlineOperation ?: return
-        fileDataStorageManager.keepOfflineOperationAndServerFile(offlineOperation, serverFile)
+        fileDataStorageManager.keepOfflineOperationAndServerFile(
+            offlineOperation,
+            serverFile,
+            clientRepository.getOwncloudClient()
```
alperozturk96: `clientRepository.getOwncloudClient()` this is nullable please handle it.

app/src/main/java/com/owncloud/android/datamodel/FileDataStorageManager.java
alperozturk96: `getNewAvailableRemotePath` adds " (2)" and offline folder remote path ends with `/`

Thus for example new path will be `/parent/folder/ (2)` instead `/parent/folder (2)/`. Also `getNewAvailableRemotePath` uses `remotePath.lastIndexOf('.')` thus it will cause more problem for example: original path: `/my.dir/folder/` , renamed path: `/my (2).dir/folder/`.

Please do not modify `getNewAvailableRemotePath` and use new function for the all type of the offline operations.

app/src/main/java/com/nextcloud/utils/extensions/FileExtensions.kt
```diff
@@ -111,3 +117,25 @@ fun String.getBitmapSize(): Pair<Int, Int>? = try {
...
+    val remoteName = this.fileName
+    val localSize = localFile.length()
+    val remoteSize = this.fileLength
+    val localCreated = attr.creationTime().toMillis() / MS_IN_SECOND // Unix time in milliseconds
+    val localModified = attr.lastModifiedTime().toMillis() / MS_IN_SECOND // Unix time in milliseconds
+    val remoteCreated = this.creationTimestamp // Unix time in seconds!
+    val remoteModified = this.modificationTimestamp / MS_IN_SECOND // Unix time in milliseconds
+    remoteName == localName &&
+        remoteSize == localSize &&
+        remoteCreated == localCreated &&
```
alperozturk96, line 138: How come `remoteCreated` and `localCreated` time can be same? We receive the remote file from server then write into our DB. 

Scenario 1:

File created from other client, remote creating date written in the server's DB and Android client writes that and compare against the potentially exists local file this can be same file but creating time can be different.

Scenario 2:

File created from Android client does server stores Android's creation time exactly or stores based on server's creation time?

app/src/main/java/com/owncloud/android/ui/dialog/conflict/ConflictResolveDialogFactory.kt
```diff
@@ -22,23 +24,34 @@ import java.io.File
...
-            headline = context.getString(R.string.conflict_folder_headline),
-            description = context.getString(R.string.conflict_message_description_for_folder),
+            headline = context.getString(R.string.choose_which_file),
+            description = context.getString(R.string.conflict_message_description),
             localFile = context.conflictFileData(
-                titleId = R.string.prefs_synced_folders_local_path_title,
-                timestamp = (leftFile.createdAt ?: 0L) * SECONDS_TO_MILLIS,
-                fileLength = UNKNOWN_FOLDER_SIZE
+                titleId = R.string.conflict_local_file,
+                timestamp = (localFile?.lastModified() ?: 0L),
```
alperozturk96: Before it was using `OfflineOperationEntity` now only uses `java.io.File` for only `OfflineOperationType.CreateFile` this will cause issue:

1. Displaying invalid information for example timestamp will be `1/1/1970` if local file is null and file length will be`0 B`.

2. This need to be covered for other type of the offline operation `RenameFile`, `RemoveFile` since now via this PR all other 3 used as `null`.

alperozturk96 (changes requested):

app/src/main/java/com/owncloud/android/ui/dialog/conflict/ConflictResolveDialogFactory.kt
daniele-verducci: Hey Alper,
that's right, I fixed it restoring the old values for the other cases.
Anyway, in the original version, the fileLength was always UNKNOWN_FOLDER_SIZE, so 0L.
I'm not sure how to obtain that value in the other two cases, because in case of RemoveFile we have only the remote path and in case of Rename we have the OCFile, and it's not sufficient to obtain the local path (local path column "media_path" is almost always empty in the db).

app/src/main/java/com/nextcloud/utils/extensions/FileExtensions.kt
daniele-verducci: Good question. Not sure about this, i looked into it with the debugger:
- file created from other client and overwritten by android client while offline: remoteCreated = 0, localCreated = 1789640828
- file created from android client, then re-uploaded while offline: remoteCreated = 1789392966, localCreated = 1789392966
(This is called by createPendingFile, so it's never called if the app is online)

I checked why remoteCreated isn't populated: it is, but it's already 0 in RemoteFile

alperozturk96: Was the created from other client same?

app/src/main/java/com/nextcloud/utils/extensions/FileDataStorageManagerExtensions.kt
```diff
@@ -215,21 +215,28 @@ fun getRemotePathForConflictResolution(client: OwnCloudClient, remotePath: Strin
...
+        else if (fileName.indexOf(".") > 0) "."
+        else ""
+    var nameFirstPart =
+        if (separator.isNotEmpty()) fileName.substringBeforeLast(separator)
+        else fileName
+    var nameLastPart =
+        if (separator.isNotEmpty()) fileName.substringAfterLast(separator, "")
+        else "" // Extension or path separator
+    if (nameLastPart.isNotEmpty() || isFolder) nameLastPart = "$separator$nameLastPart"
     val regex = Regex("""(.*)\((\d+)\)$""", RegexOption.MULTILINE)
```
alperozturk96, line 229: `    val regex = Regex("""(.*)\((\d+)\)$""", RegexOption.MULTILINE)`

Can we move out of the function thus no need to create every time again?

app/src/main/java/com/nextcloud/utils/extensions/FileExtensions.kt
daniele-verducci: I'll remove Created check and write a test on the other fields

PR #17593, opened 2026-09-09
refactor(conflict-dialog): dialog determination

alperozturk96:
### Changes

- Uses `kotlinx.parcelize.Parcelize`
- Makes clear distinction between offline and normal via sealed class
- Uses factory class to keep code more clear 
- Uses only conflict package


daniele-verducci (approved):
Thank you for the refactoring! Definitely more readable and mantainable! I tested it and it seems to match the previous behavior. Approved on my side 👍

PR #17608, opened 2026-09-10
Fix subfolders in encrypted folders

daniele-verducci:
How to repro:
- create encrypted folder
- create encrypted subfolder inside the one just created
- app crashes

How to repro 2:
- enter encrypted folder
- enter encrypted subfolder inside the previous one
- app crashes

### 🏁 Checklist
 
- [ ] 📅 Milestone is set

## 🤖 AI (if applicable)


app/src/main/java/com/nextcloud/utils/e2ee/E2EEKeyInspector.kt
```diff
@@ -102,23 +103,29 @@ class E2EEKeyInspector @Inject constructor(
...
+            arbitraryDataProvider
         )
-
-        val user = metadataFile.users.find { it.userId == userId }
+        val user = metadata.users.find { it.userId == userId }
             ?: throw IllegalStateException("cannot find current user in metadata")
 
         return try {
-            EncryptionUtils.decryptStringAsymmetricV2(user.encryptedMetadataKey, privateKey)
+            EncryptionUtils.decryptStringAsymmetricV2(user.decryptedMetadataKey, privateKey)
```
alperozturk96: Why we are trying to decrypt the decrypted metadata key here?

app/src/main/java/com/nextcloud/utils/e2ee/E2EEKeyInspector.kt
```diff
@@ -102,23 +103,29 @@ class E2EEKeyInspector @Inject constructor(
...
             } else {
                 decryptsMetadataV1(metadata.metadata, privateKey, folder.localId)
             }
         }
 
-    private fun decryptsMetadataV2(serializedMetadata: String, privateKey: String, userId: String): Boolean {
-        val metadataFile = EncryptionUtils.deserializeJSON(
-            serializedMetadata,
-            object : TypeToken<EncryptedFolderMetadataFile>() {}
+    private fun decryptsMetadataV2(ocFile: OCFile, privateKey: String, client: OwnCloudClient): Boolean {
```
alperozturk96, line 112: Previously `GetMetadataRemoteOperation` was already given the metadata why now we need to `retrieveTopMostMetadata` and also eventually call `decryptFolderMetadataFile` ?

alperozturk96 (commented):
Could you please share the crash-log and explain your approach?

alperozturk96 (changes requested):

daniele-verducci:
> Could you please share the crash-log and explain your approach?

Hey Alper, here's the crashlog:
```

2026-09-14 12:06:21.801  4737-4737  AndroidRuntime          com.nextcloud.client                 E  FATAL EXCEPTION: main
                                                                                                    Process: com.nextcloud.client, PID: 4737
                                                                                                    java.lang.NullPointerException: Attempt to invoke interface method 'java.util.Iterator java.lang.Iterable.iterator()' on a null object reference
                                                                                                    	at com.nextcloud.utils.e2ee.E2EEKeyInspector.decryptsMetadataV2(E2EEKeyInspector.kt:117)
                                                                                                    	at com.nextcloud.utils.e2ee.E2EEKeyInspector.access$decryptsMetadataV2(E2EEKeyInspector.kt:33)
                                                                                                    	at com.nextcloud.utils.e2ee.E2EEKeyInspector$canDecryptFolderMetadata$2.invokeSuspend(E2EEKeyInspector.kt:105)
                                                                                                    	at com.nextcloud.utils.e2ee.E2EEKeyInspector$canDecryptFolderMetadata$2.invoke(E2EEKeyInspector.kt:8)
                                                                                                    	at com.nextcloud.utils.e2ee.E2EEKeyInspector$canDecryptFolderMetadata$2.invoke(E2EEKeyInspector.kt:4)
                                                                                                    	at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatched(Undispatched.kt:66)
                                                                                                    	at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn(Undispatched.kt:43)
                                                                                                    	at kotlinx.coroutines.BuildersKt__Builders_commonKt.withContext(Builders.common.kt:497)
                                                                                                    	at kotlinx.coroutines.BuildersKt.withContext(dex-id-4119e2f2b95ce035797d115b10131393af91e27e:1)
                                                                                                    	at com.nextcloud.utils.e2ee.E2EEKeyInspector.canDecryptFolderMetadata(E2EEKeyInspector.kt:86)
                                                                                                    	at com.nextcloud.utils.e2ee.E2EEActionResolver$checkFolderMetadataKey$2.invokeSuspend(E2EEActionResolver.kt:44)
                                                                                                    	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
                                                                                                    	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
                                                                                                    	at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:124)
                                                                                                    	at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:89)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:586)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:798)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:717)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:704)
                                                                                                    	Suppressed: kotlinx.coroutines.internal.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@5e01fcc, Dispatchers.Main.immediate]
```

This happens when opening a **subfolder** on an encrypted folder. The crash is caused by the users array being empty/null (the user is used to retrieve the key).
I checked on the web watching the requests and I saw the users array being empty there as well, so I figured out they are using the parent encrypted folder's key.
I then replaced the piece of code that obtains the user array from the current folder with one that retrieves the topmost encrypted folder (that always has the users array populated) and uses its key.

I can say it works, but I'm not sure it's the correct fix. What do you think about it?

app/src/main/java/com/nextcloud/utils/e2ee/E2EEKeyInspector.kt
daniele-verducci: GetMetadataRemoteOperation retrieves the metadata for the current folder, but I saw with the debugger that the users array is present only if the folder is the root encrypted folder. If we are in a subfolder, it's not. I think this isn't a server bug because it's working in the web version (checked the calls and the answer is the same, just in XML instead of JSON, so they must be keeping the root encrypted folder's key and using it for the subfolders, I believe)

app/src/main/java/com/nextcloud/utils/e2ee/E2EEKeyInspector.kt
daniele-verducci: I see, this doesn't make sense. The result is not used, so I believe this call in the previous code was used to throw an exception if the key isn't valid. Is it a correct interpretation?

In that case, maybe I can get directly the key with retrieveTopMostMetadataKey(). It should throw an exception as well.
But then should I verify the key in any way?

daniele-verducci:
@alperozturk96 Can you check last commit? Could it be a solution?

app/src/main/java/com/nextcloud/utils/e2ee/E2EEKeyInspector.kt
```diff
@@ -102,24 +103,27 @@ class E2EEKeyInspector @Inject constructor(
...
-
-        val user = metadataFile.users.find { it.userId == userId }
-            ?: throw IllegalStateException("cannot find current user in metadata")
+    private fun decryptsMetadataV2(ocFile: OCFile, privateKey: String, client: OwnCloudClient): Boolean {
+        val userId = client.userId
 
         return try {
-            EncryptionUtils.decryptStringAsymmetricV2(user.encryptedMetadataKey, privateKey)
-            true
+            val metadataKey = EncryptionUtilsV2().retrieveTopMostMetadataKey(
```
alperozturk96, line 116: I saw this comment in the master. @tobiasKaminsky I think using `retrieveTopMostMetadataKey` directly here is okay. What do you think?

` // we are in a subfolder, decrypt information is in top most encrypted folder`

daniele-verducci:
Completed, but do not merge until the encrypted folder tests are implemented (in a separate PR).

alperozturk96:
> Completed, but do not merge until the encrypted folder tests are implemented (in a separate PR).

You can put this PR in a draft then activate again later :)

Screenshot 2026-09-16 at 12 26 50

PR #17616, opened 2026-09-11
fix(connectivity-service): capability change check

alperozturk96:
### Issue

`onCapabilitiesChanged` gets called even for stable same connection thus to prevent too much call throttling applied.


daniele-verducci (approved):
Tested checking the logs and playing with the emulator connection, seems to work as intended. Thank you!

PR #17632, opened 2026-09-14
fix(upload): pause

alperozturk96:
### How to reproduce?

1. Try upload a big file to the root
2. Pause it
3. Create new folder
4. Try upload big file to the newly created folder

#### Expected

First upload should be paused, second upload should be uploaded

#### Actual

Both files are uploaded and notification disappears

### Changes

- Fixes global pause action so that current uploads can be paused
- Fixes pause action


daniele-verducci (changes requested):
Hey Alper, I tested it but the paused upload is listed as "cancelled":

image

To be sure I'm testing it the right way, here what I've done:
- upload big file
- pause upload **from notification**
- create folder
- upload **same big file** in folder
- go to menu -> uploads to check progression: first upload is shown as **cancelled**, second one is uploading correctly

Here I should wee the first one as paused and the second one running if I interpreted the first post right.


If I do the same test, but pausing from the Uploads list instead of the notification:
- upload big file
- pause upload **from the Uploads list**
- create folder
- upload same big file in folder
- in Uploads list both are paused
- if I un-pause from there, the first resumes upload and, after finished, the second one starts

This last behavior seems correct to me.

alperozturk96:
> Hey Alper, I tested it but the paused upload is listed as "cancelled":
> 
> image
> To be sure I'm testing it the right way, here what I've done:
> 
> * upload big file
> * pause upload **from notification**
> * create folder
> * upload **same big file** in folder
> * go to menu -> uploads to check progression: first upload is shown as **cancelled**, second one is uploading correctly
> 
> Here I should wee the first one as paused and the second one running if I interpreted the first post right.
> 
> If I do the same test, but pausing from the Uploads list instead of the notification:
> 
> * upload big file
> * pause upload **from the Uploads list**
> * create folder
> * upload same big file in folder
> * in Uploads list both are paused
> * if I un-pause from there, the first resumes upload and, after finished, the second one starts
> 
> This last behavior seems correct to me.

Both behaves same for me. I didn't understand what is wrong? Upload listed as cancelled because user wanted to pause it.

Wording, UI/UX wise it's not good but this is out of scope for this PR. Could you please share your screen recording that shows bug?

### Demo

https://github.com/user-attachments/assets/0f49e2b3-9ebb-46bc-83c4-dccd5dd79e39

daniele-verducci:
Thanks for the video. There was a little of a misunderstanding here: I tought that the "pause" button in the notification was functionally equal to the main pause button in the upload window (the top-right one):

image

While I considered the "X" button a "cancel" button.

So, considering this, now it make sense and functionally is ok (= it works as intended), but I think it's still misleading (but as you said, out of scope for this PR)

daniele-verducci (approved):

PR #17657, opened 2026-09-16
perf(startup): splash screen

alperozturk96:
### Changes

1400 ms faster cold start

### Before

[Screen_recording_20260916_133720.webm](https://github.com/user-attachments/assets/0dedbad2-0ae7-402a-9db2-f383021b5d7e)

### After

[Screen_recording_20260916_133829.webm](https://github.com/user-attachments/assets/1b132038-de1a-493a-9e90-c77c66350943)


app/src/main/java/com/nmc/android/ui/LauncherActivity.kt
```diff
@@ -56,30 +56,39 @@ class LauncherActivity : BaseActivity() {
...
+        }
+
+        startActivity(Intent(this, nextScreen))
+        finish()
     }
 
     companion object {
-        const val SPLASH_DURATION = 1500L
+        private val SPLASH_DURATION = 1500.milliseconds
+        private val NO_SPLASH_DURATION = 100.milliseconds
```
alperozturk96, line 92: 100 ms needed so that we can show app icon.

tobiasKaminsky: @jancborchardt since you raised the issue :)

jancborchardt (approved):
Very nice! Can we get it even faster? :)

alperozturk96:
> Very nice! Can we get it even faster? :)

This is just the beginning. After this PR, I’ll investigate the internal state changes between the loading and loaded states. The splash screen can be faster for the non-branded flow around 1.4 seconds. For the branded flow, it will remain the same so that the custom logo and other branding elements can be displayed. Do you think 1.5 seconds is enough for the branded flow? I would suggest setting it to 300 ms for the branded flow and 100 ms for the non-branded flow.

PR #17667, opened 2026-09-17
perf(file-list): avatar loading

alperozturk96:
### Issue

Scrolling up and down causing to re-generate avatars

### Changes

- Doing all job in background thread and render only in main thread
- Adds unified share calls to the missing places


daniele-verducci (approved):
Hey Alper, this seems good to me, so I approved.
Anyway, I found a bug that has NOT been introduced in this one (already present on master) that may be related, see video:

https://github.com/user-attachments/assets/8062255b-08ea-4e7f-b442-e4b5de029f4b

Basically, at start the sharee icons are flickering and then only the first sharee is shown. When I move to Recent or Shared, then all the sharees are shown in the icon, and if I go back to files view, it remains like that until I reload the list.

PR #17692, opened 2026-09-22
fix(auth): polling

alperozturk96:


daniele-verducci (changes requested):
Hey Alper, I'm sorry, it didn't work

Image

